PR #116, opened 2016-11-21
New style of phishing playbook with email parsing and single response

slavikm:
Lior, please make sure it works and add additional response steps with other integrations. Also, make sure to use the right sandbox as available.


liorkol:
@slavikm 
Added all the required steps and ran enough tests-fixes cycles to be satisfied with it.
If it looks good go ahead and merge when ready.

Scripts/script-CheckSender.yml
```diff
@@ -1,20 +1,31 @@
...
-  Array.prototype.push.apply(res, rep);
-  res.push({Contents: 'no', Type: entryTypes.note, ContentsFormat: formats.text});
-  return res;
-type: javascript
+  import re
+  email = ''
+  if 'email' in demisto.args():
+      email = demisto.args()['email']
+  else:
+      sender = re.search(r".*From\w*:.*\b([A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,})\b", demisto.incidents()[0]['details'], re.I)
```
slavikm: I'm not liking it. I think we should just receive a parameter of where to look and then pass the relevant context parameter.Let's not hard-code the demisto.incidents()[0]['details'] here. So, 2 parameters - email and body which you'll regex the email from.

PR #118, opened 2016-11-22
Sandblast integration

Spivaka:
fixes https://github.com/demisto/server/issues/3620
@aviadl please review
@liorkol / @JonathanMeler - You can start writing scripts

A few words about the integration:
Checkpoint Threat Emulation (or Sandblast) is a somewhat equivalent service to FireEye / PANW Wildfire - It allows you to upload files, query hashes (sha1/sha256/md5) and download reports (pdf/xml) if generated. 
The service also allow you to query the quota usage of your account.
TE has a cloud instance which is set as the default URL, but can also have a in premise device that has it's own URL (can be overridden manually).


aviadl:
@Spivaka Please make a short video with a demo of this integration and place it in the integrations channel in slack

Integrations/integration-Sandblast.yml
```diff
@@ -0,0 +1,249 @@
...
+  required: false
+script:
+  script: |-
+    var server = params.server;
+    var insecure = params.insecure;
+    var token = params.token;
+    var version = params.version;
+    var requestTemplate = JSON.stringify({request:[{features:['te']}]});
+
+    var base = server +'/tecloud/api/' + version + '/file/';
```
aviadl: I think you need to sanitize the server URL.
Check if it ends with a '/' or not ,,,

Integrations/integration-Sandblast.yml
```diff
@@ -0,0 +1,249 @@
...
+            }
+            // Extract file name from response
+            var fileName = command + '_at_' + currentTime.getTime();
+            if (res.Headers && res.Headers['content-disposition'] && res.Headers['content-disposition'].length === 1) {
+                var matches = /'filename=\"(.*)\"'/.match(res.Headers['content-disposition'][0]);
+                if (matches) {
+                    fileName = matches[1];
+                }
+            }
+            return {Type: 9, FileID: saveFile(res.Body), File: fileName,  Contents: 'we must have contents for an entry'};
```
aviadl: Putting this content here, will make it load in a search.
Lets place something valid. like file name
BTW we do we have to have content ?

Integrations/integration-Sandblast.yml
```diff
@@ -0,0 +1,249 @@
...
+        if (!args.features) {
+            args.features = 'all';
+        }
+        if (args.features !== 'extraction' && args.features !== 'av') {
+            var feature = {};
+                var keys = Object.keys(teArgs);
+                for (var i = 0; i < keys.length; i++) {
+                    if (args[keys[i]]) {
+                        if (teArgs[keys[i]] === 0) {
+                            console.log(args[keys[i]])
```
aviadl: remove console.log

Integrations/integration-Sandblast.yml
Spivaka: Otherwise the entry does not save correctly

PR #119, opened 2016-11-22
IncidentToContext: includes custom fields

JonathanMeler:
fixes: https://github.com/demisto/server/issues/4261


liorkol:
@JonathanMeler 
1. I see the custom fields don't appear directly in context, instead they are under a key called "CustomFields". I would want them to be directly available under `keyname`

2. Please retain the static structure of the script - it is needed for those items to appear under the `source` menu. Let's at least keep the items we know statically in there.

PR #120, opened 2016-11-22
Yml validator

meirwah:


meirwah:
first iteration @aviadl , only checks yaml's are valid yaml's

aviadl:
Without seeing the validator code it is hard to see :)

PR #121, opened 2016-11-22
Zip up each agent tools dir with "tools-{name}.zip" convention

liorkol:


liorkol:
@aviadl  Please review

liorkol:
@aviadl Done. Check it out :)

PR #132, opened 2016-11-29
Moved to using appendContext in Extract and Check scripts

liorkol:
@JonathanMeler Have a look here and give comments if you have any
Also look at how appendContext is used, we'll want to keep using this in future scripts.

@slavikm Welcome to have a look as well if you've got the time

### DIFF
--- a/Scripts/script-CheckIPs.yml
+++ b/Scripts/script-CheckIPs.yml
@@ -1,6 +1,6 @@
 commonfields:
   id: CheckIPs
-  version: 1
+  version: -1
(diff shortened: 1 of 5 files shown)


Scripts/script-CheckIPs.yml
```diff
@@ -38,16 +38,15 @@ script: |-
           if positiveIp(r):
               badIps.append(ip)
               res.append(shortIp(r))
+
```
slavikm: Extra line?

PR #134, opened 2016-11-29
Update script-CommonIntegration.yml

Spivaka:


Scripts/script-CommonIntegration.yml
```diff
@@ -19,6 +19,15 @@ script: |-
           }
           return query;
   };
+  
+  function rplaceInTemplates(input, args) {
```
liorkol: Is this being used in an integration somewhere? We need to fix the typo there as well..

Spivaka:
@liorkol fixed the typo. This is not used but we will need it for 2 main scenarios:
1. URL that include fields. e.g., go/to/:my-id/and-fetch/:this-id. We can use it to replace my-id and this-id with user input fields.
2. SOAP: <Login><User>:username</User><Password>:password</Password></Login>

The latter is common for many integrations (Algosec) and if we want to usu BYOI to write them this is essential.

PR #155, opened 2016-12-12
Functions name changed + support new api name

JonathanMeler:


liorkol:
Solved on another PR that branches from master.

PR #161, opened 2016-12-17
align reputation to new indicator repository

meirwah:
align reputation to new indicator repository


meirwah:
@liorkol plz review

PR #175, opened 2017-01-01
Add files via upload

yuvalbenshalom:
trend micro integration #https://github.com/demisto/server/issues/2266

### DIFF
--- a/Integrations/integration-trendMicroDsm.yml
+++ b/Integrations/integration-trendMicroDsm.yml
@@ -1,18 +1,18 @@
 commonfields:
-  id: TrendMicro DSM
+  id: Trend Micro
   version: -1
-name: TrendMicro DSM
-display: TrendMicro DSM
+name: Trend Micro
+display: Trend Micro
 category: IT service Management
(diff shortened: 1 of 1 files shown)


Integrations/integration-trendMicroDsm.yml
```diff
@@ -0,0 +1,327 @@
...
+description: Trend Micro Deep Security
```
Spivaka: I would check with dan if this is the description he wants (usually it's uses trend micro...)

Integrations/integration-trendMicroDsm.yml
```diff
@@ -0,0 +1,327 @@
...
+  name: proxy
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Fetch incidents
+  name: isFetch
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Incident type
```
Spivaka: sould be removed as it can be content for builds lower than 2.0 that doesn't support this feature.

Integrations/integration-trendMicroDsm.yml
```diff
@@ -0,0 +1,327 @@
...
+  name: incidentType
+  defaultvalue: ""
+  type: 13
+  required: false
+script:
+  script: |2+
+
+    var username = params.credentials.identifier;
+    var password = params.credentials.password;
+    var server = params.server.replace(/[\/]+$/, '');
```
Spivaka: should probably add this utility in the common integration script.

Integrations/integration-trendMicroDsm.yml
```diff
@@ -0,0 +1,327 @@
...
+
+
+    function getFilterContent(filterName, args) {
+        var filterContent = '';
+        filterArgsDict = filterContentDict[filterName];
+        var argNames = Object.keys(args);
+        for (var i = 0; i < argNames.length; i++) {
+            if (argNames[i] in filterArgsDict){
+            filterContent = filterContent + filterArgsDict[argNames[i]];
+            }
```
Spivaka: += is supported in js

Integrations/integration-trendMicroDsm.yml
```diff
@@ -0,0 +1,327 @@
...
+        var match = responseDict[method].exec(responseXML);
+        if (match && match[1]) {
+            return match[1];
+        }
+        throw method +' failed';
+    }
+
+    // Wrap host IDs array if exists
+    if ('hostIDs' in args) {
+            args.hostIDs = '<urn:hostIDs>' + args.hostIDs.split(',').join('</urn:hostIDs><urn:hostIDs>') + '</urn:hostIDs>';
```
Spivaka: usage here is not clear... would move from the function region

PR #179, opened 2017-01-05
add get-instance-info command to AWS integration

meirwah:
add get-instance-info command to AWS integration


meirwah:
@JonathanMeler plz review

PR #180, opened 2017-01-05
QRadar - Limit the amount of data fetched in one call

Spivaka:


Spivaka:
You are right about the test. will limit.
about the call - whoever calls it can limit. don't think we should hard-code limit it

aviadl:
@Spivaka what about this pr ?

aviadl:
@Spivaka what is this waiting for ?

liorkol:
@Spivaka @aviadl Ping :)

PR #188, opened 2017-01-12
Koodous integration

Diviei:
Basic integration with the Koodous API


meirwah:
Looks good @Diviei just changed the category to existing one....

PR #191, opened 2017-01-18
protect from empty indicators field in StixParser

meirwah:
protect from empty indicators field in StixParser


liorkol:
@meirwah Merge when ready

PR #195, opened 2017-01-22
Venafi integration

Spivaka:
for https://github.com/demisto/server/issues/4927


aviadl:
@spivaka that was quick

PR #203, opened 2017-01-30
ParseCSV

shadejinx:
This file will take a CSV file that has been uploaded to a War Room and allow the parsing of IP addresses, domains and hashes.


liorkol:
@shadejinx Looks very useful
I will review it as soon as I can

PR #204, opened 2017-01-31
change StixParser Docker image to demisto instead of meirwah

meirwah:
change StixParser Docker image to demisto instead of meirwah


meirwah:
do not merge until server PR is merged

PR #208, opened 2017-02-01
wrapper script for copy-from' command

aviadl:
Will also set file metadata into context, if key is provided
Need to validate that the depends on parameters are set correctly
also need to validate that the fromversion is set correctly

### DIFF
--- a/Scripts/script-RACopyFrom.yml
+++ b/Scripts/script-RACopyFrom.yml
@@ -1,0 +1,54 @@
+commonfields:
+  id: RACopyFrom
+name: RACopyFrom
+fromversion: 2.0.1
+script: |-
+  var files = executeCommand("copy-from", {"file":args.filePath})
+  var err = null;
+  var res = null
+  for (i = 0; i < files.length; i++) {
+      if (files[i].Type === entryTypes.file) {
+          res = files[i];
+      } else if (files[i].Type === entryTypes.error) {
+          err = files[i]
+      }
+  }
+  if (err) {
+      return err;
+  }
+
+  if (args.contextKey) {
+      setContext(args.contextKey+"/MD5", res.FileMetadata.MD5);
+      setContext(args.contextKey+"/SHA1", res.FileMetadata.SHA1);
+      setContext(args.contextKey+"/SHA256", res.FileMetadata.SHA256);
+      setContext(args.contextKey+"/SHA512", res.FileMetadata.SHA512);
+      setContext(args.contextKey+"/SSDeep", res.FileMetadata.SSDeep);
+      setContext(args.contextKey+"/Type", res.FileMetadata.Type);
+  }
+
+  return res;
+type: javascript
+tags: []
+comment: |
+  Copy a file using remote access integration.
+  Will place file attributes on contextKey param:
+  Assuming contextKey parameter is foo, the following keys will be set:
+  foo/MD5
+  foo/SHA1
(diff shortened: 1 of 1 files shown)


aviadl:
Need to make script accept a list of files (csv or array)
Need to allow limiting to a specific instance
Need to allow fetching from multiple instances

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
...
+  Copy a file using remote access integration.
+  Will place file attributes on contextKey param:
+  Assuming contextKey parameter is foo, the following keys will be set:
+  foo/MD5
+  foo/SHA1
+  foo/SHA256
+  foo/sha512
+  foo/SSDeep
+  foo/Type
+enabled: true
```
liorkol: Not needed. Add "system: true" instead

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
...
+      setContext(args.contextKey+"/SHA1", res.FileMetadata.SHA1);
+      setContext(args.contextKey+"/SHA256", res.FileMetadata.SHA256);
+      setContext(args.contextKey+"/SHA512", res.FileMetadata.SHA512);
+      setContext(args.contextKey+"/SSDeep", res.FileMetadata.SSDeep);
+      setContext(args.contextKey+"/Type", res.FileMetadata.Type);
+  }
+
+  return res;
+type: javascript
+tags: []
```
liorkol: Add tags

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
+commonfields:
+  id: RACopyFrom
```
liorkol: I suggest renaming to RemoteaccessCopyFrom. Much more descriptive and not that long

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
...
+name: RACopyFrom
+fromversion: 2.0.1
+script: |-
+  var files = executeCommand("copy-from", {"file":args.filePath})
+  var err = null;
+  var res = null
```
liorkol: Missing some ; (in a few other lines too)

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
...
+- name: filePath
+  required: true
+  default: true
+  description: File to copy to war room
+- name: contextKey
+  description: Context key to store file details on
+scripttarget: 0
+dependson: 
+  must:
+  - remoteaccess
```
liorkol: These are commands, so it needs to be "copy-from"

Scripts/script-RACopyFrom.yml
```diff
@@ -0,0 +1,54 @@
+commonfields:
+  id: RACopyFrom
+name: RACopyFrom
+fromversion: 2.0.1
+script: |-
+  var files = executeCommand("copy-from", {"file":args.filePath})
```
liorkol: Script should take and forward "using" arg. Pretty quick and worth it for the added power

aviadl:
@gshriki you have modified this script on you end.
Please validate that it fits the comments @liorkol has wrote and place it in this PR

meirwah:
closing this...

PR #210, opened 2017-02-02
Add files via upload

yuvalbenshalom:


liorkol:
@Spivaka @yuvalbenshalom Please merge when ready. Need 2.0.1 to include usable Salesforce..

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+configuration:
+- display: Credentials
+  name: credentials
+  defaultvalue: ""
+  type: 9
+  required: true
+- display: Consumer Key
```
Spivaka: shouldn't the secret and the key be an authentication type as they are bounded together?
usually we use camel case notation and not _ in params

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+        var response = http(
+                    authUrl,
+                    {
+                        Method: 'POST',
+                        Headers: {'Content-Type': ['application/x-www-form-urlencoded']},
+                        Body: body,
+                    },
+                    false
+                );
+            if (response.StatusCode < 200 || response.StatusCode >= 300) {
```
Spivaka: indention - should be one tab left

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+    function getNewToken() {
+        var request = {
+                grant_type: 'password',
+                client_id: params.client_id,
+                client_secret: params.client_secret,
+                username: params.credentials.identifier,
+                password: params.credentials.password
+            };
+
+        var body = encodeToURLQuery(request).substr(1);
```
Spivaka: Is that worth adding as a call inside common integration to be reused? not sure, do it if you think so also

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    function sendRequest(method, url, body, token) {
+        var headers = {};
+        if (token) {
+            headers['Authorization'] = ['Bearer ' + token];
+        }
+        if(method == 'POST' || method == 'PATCH') {
+            headers['Content-Type'] = ['application/json'];
+        }
+            return http(
```
Spivaka: indention...

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+                        Headers: headers,
+                        Body: body,
+                    },
+                    false
+                );
+        }
+
+    function sendRequestInSession(method, uri, body) {
+        var contextData = getIntegrationContext();
+        if (!contextData || !contextData.access_token) {
```
Spivaka: View box for better handling in case of two updates in the same time.

I prefer KISS - keep it simple stupid and always authenticate

Integrations/Integration-Salesforce.yml
```diff
@@ -0,0 +1,174 @@
...
+            response = sendRequest(method, contextData.instance_url + uriPrefix + uri, body, contextData.access_token);
+        }
+        if (response.StatusCode < 200 || response.StatusCode >= 300) {
+                throw 'Failed to run command uri: ' + uri + ', request status code: ' + response.StatusCode + ' and Body: ' + response.Body + '.';
+        }
+        return response;
+    }
+
+
+    if (command == 'test-module') {
```
Spivaka: I would get the token any how and use it

PR #218, opened 2017-02-08
Get output from process

orius123:
In case of error


orius123:
I suggest we extract the logic of spawning a process to a util, and handle exceptions and error code better there.

PR #219, opened 2017-02-09
Update integration-jira.yml

meirwah:
add fetch incidents for Jira
handle missing / suffix in url


meirwah:
@Spivaka can you review?

PR #228, opened 2017-02-14
support tags in threatconnect integration

meirwah:
support tags in threatconnect integration


meirwah:
@Spivaka can u plz review?

Spivaka:
the tag and get indicator doesn't seem to send anything. is that ok??

meirwah:
where do you see that? point to code plz...

meirwah:
@Spivaka ?

Spivaka:
Maybe I am not following but does indicators.retrieve() sends something? if yes, than forget what I wrote. If not I don't understand in which line the message goes out

meirwah:
indicators.retrieve()  sends an GET request, to get iindicators

PR #232, opened 2017-02-21
IntSights integration

Spivaka:
This integration doesn't fetch incidents from IOCs yet as there is no IOC filtering at IntSights side.
We should resume and add it when the new API is out.
fixes https://github.com/demisto/server/issues/5739


Integrations/integration-IntSights.yml
```diff
@@ -0,0 +1,547 @@
...
+        'intsights-remediate': '/alerts/resolve-request/%AlertID%',
+        'intsights-whitelist': '/alerts/report-white-list/%AlertID%',
+        'intsights-assign': '/alerts/assign/%AlertID%',
+        'intsights-unassign': '/alerts/unassign/%AlertID%',
+        'intsights-tag': '/alerts/add-tag/%AlertID%',
+        'intsights-untag': '/alerts/remove-tag/%AlertID%',
+        'intsights-comment': '/alerts/comment/%AlertID%',
+        'intsights-alert-preview': '/alerts/preview/%AlertID%',
+        'intsights-get-alert-db-file': '/alerts/db-file/%AlertID%',
+        'intsights-get-
```
liorkol: Should move to common

liorkol: should remove this and add system:true

Integrations/integration-IntSights.yml
```diff
@@ -0,0 +1,533 @@
...
+        'intsights-search-ioc': 'iocs/search',
+        'intsights-ioc-download-csv': 'iocs/download-csv'
+    };
+
+    var methodDictionary = {
+        'login': 'POST',
+        'logout': 'GET',
+        'intsights-get-accounts': 'GET',
+        'intsights-get-alerts': 'GET',
+        'intsights-
```
aviadl: Need to remove this console.log

PR #237, opened 2017-02-27
Create integration-censys.yml

shadejinx:
This python script will provide API access to the Search and View commands offered by Censys.io. The output is a voluminous JSON blob and needs to be sliced-and-diced with proper automation scripts.


liorkol:
Awesome idea! We can expand and build some very cool stuff with this.

PR #243, opened 2017-03-04
CVE Information Integration

gshriki:
added CVE information integration including scripts and reputation hypersearch for CVSS score

Fixes https://github.com/demisto/server/issues/2117


slavikm:
@gshriki if you're already touching this, do you mind adding SHA1 and SHA256 to the objects (same idea as md5 just with 40 chars for sha1 and 64 chars for sha256). The same scripts and capabilities should just work.

gshriki:
@slavikm - done

PR #246, opened 2017-03-06
Cylance PROTECT - Configuring the filter according to user needs

Spivaka:
fixes https://github.com/demisto/server/issues/6126
Time field is configurable
filters can be added with JSON format


Integrations/integration-Cylance_Protect.yml
```diff
@@ -4,12 +4,13 @@ commonfields:
...
 description: Manage end points using Cylance protect
-detaileddescription: For more information regarding thee integrations specific fields,
-  please advise the Cylance API Guide for version 1.7.
+detaileddescription: |-
+  For more information regarding thee integrations specific fields, please advise the Cylance API Guide for version 1.7.
```
aviadl: You have here a typo, please ask Dan for specific text

Integrations/integration-Cylance_Protect.yml
```diff
@@ -192,6 +220,7 @@ script:
   type: javascript
   commands:
   - name: cp-get-list
+    deprecated: false
```
aviadl: remove the deprecated fields

Integrations/integration-Cylance_Protect.yml
```diff
@@ -41,15 +42,24 @@ configuration:
   defaultvalue: ""
   type: 8
   required: false
+- display: Time Field
+  name: timeField
+  defaultvalue: last_found
```
Spivaka: @liorkol this is the correct default value otherwise duplications will be created. if a specific customer wants to change its now possible.

PR #254, opened 2017-03-15
fix RSA Netwitness SA integration

anara123:
Was missing break in switch. And it cause all the commands run one after another


anara123:
@liorkol review it and merge it ASAP

PR #277, opened 2017-04-05
Create Integration-Wildfire.yml

shaniacht1:
fixes https://github.com/demisto/server/issues/6463


Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,261 @@
...
+
+    // The command input arg holds the command sent from the user.
+    switch (command) {
+      // This is the call made when pressing the integration test button.
+      case 'test-module':
+          var res = http(
+              url,
+              {
+                  Method: 'POST',
+                  Body: 'apikey=37ddffe5490c01835b6645fb6ecfe9c9&format=xml&hash=7f638f13d0797ef9b1a393808dc93b94',
```
Spivaka: A couple of things.
1. It should be the user API key in order to test his env.
2. Our apikey shouldn't be available on this public repo...

Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,261 @@
...
+                  Headers: headers
+              }
+          );
+
+          if(res.StatusCode==200)
```
Spivaka: you can just use the report for the known hash and it will throw an error for you

Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,261 @@
...
+                  EntryContext: context
+          };
+        }
+        break;
+
+      case 'WildFireUpload':
+          var url = params.server + urlDict.upload;
+          var result = sendMultipartRequest(url, headers, args.upload, params.token);
+          var fileinfo = result.wildfire['upload-file-info'];
+          var context = parseFile(fileinfo);
```
Spivaka: can be in one line with 226

Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,261 @@
...
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: API Key
+  name: token
+  defaultvalue: ""
+  type: 4
+  required: true
+- display: Do not validate server certificate (insecure)
+  name: secure
```
shaniacht1: Sorry, I didn't quite get what you meant

Integrations/Integration-Wildfire.yml
shaniacht1: About this one - I needed the fileinfo for lines 227 and 228. I figured creating a var would make it cleaner. Is it wrong?

Integrations/Integration-Wildfire.yml
Spivaka: Field means to skip validation which is nor secure, so a proper name should be insecure (as when it's true, you are not secured)

Spivaka:
Please also add this description (from our current wildfire integration) to the detailed description so it will be on the flip card (when pressing the ? at the integration params window):
Don’t have a WildFire API key?
Go to your Wildfire account,
Login to: https://wildfire.paloaltonetworks.com/wildfire/account
Select the “Account” tab
Copy your API Key

Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,204 @@
...
+    switch (command) {
+      case 'test-module':
+          url += urlDict.report;
+          var res = http(
+              url,
+              {
+                  Method: 'POST',
+                  Body: 'apikey='+params.token+'&format=xml&hash=7f638f13d0797ef9b1a393808dc93b94',
+                  Headers: headers
+              }
```
Spivaka: This request should also include the insecure and proxy

Integrations/Integration-Wildfire.yml
```diff
@@ -0,0 +1,222 @@
...
+                      HumanReadable: md, /*todo*/
+                      ReadableContentsFormat: formats.markdown,
+                      EntryContext: context
+              };
+
+    }
+
+  type: javascript
+  commands:
+  - name: WildFireReport
```
Spivaka: You can use outputs instead of creating an entry context and returning it, but in order to be used in playbooks you should add the output to context path File

PR #279, opened 2017-04-06
SEPM Integration

nitzanshapira:


Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -33,13 +32,51 @@ configuration:
...
+        'fullPathName',
+        'numberOfPhysicalComputers',
+        'numberOfRegisteredUsers',
+        'policySerialNumber',
+        'policyDate',
+        'description',
+        'created'
+    ]
+
     var fixUrl = function(base) {
```
Spivaka: all this code for server = params.server.replace(/\/$/, '') + '/';?

Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -52,6 +89,16 @@ script:
...
+    }
+
+    var buildCommandXml = function(data) {
+        return '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:com="http://command.client.webservice.sepm.symantec.com/"> \
+                <soapenv:Header/><soapenv:Body>{0}</soapenv:Body></soapenv:Envelope>'.format(data);
+    }
+
     var parseResponse = function(resp) {
       if (resp.StatusCode === 200) {
           try {
```
Spivaka: response is in json while requests are in xml? really?

Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -66,8 +113,7 @@ script:
         }
         throw err;
       }
-
-    };
+    }
 
     var doAuth = function(qParams) {
       url = fixUrl(qParams.server) + 'sepm/api/v1/identity/authenticate';
```
Spivaka: warning... bad naming in the way...

Spivaka: that could be in one line instead of 10...

Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -91,43 +137,303 @@ script:
...
-          },
-          qParams.insecure
-      );
+        res = http(
+            url,
+            {
+                Method: 'GET',
+                Headers: {'Authorization': ['Bearer ' + token]}
+            },
+            qParams.insecure
```
Spivaka: what about proxy?

Integrations/integration-SymantecEndpointProtection.yml
nitzanshapira: Yes :/

nitzanshapira: Some of the commands of the API send JSON (e.g. "authenticate" or "computers") and some send XML (e.g. the Command ones). However, I'm parsing the result as JSON anyway and handle it in a united way.

Integrations/integration-SymantecEndpointProtection.yml
nitzanshapira: TBH this code was already there (by Lior) and it worked so I didn't touch it. I fixed the naming (not using "u, p, d" anymore). Regarding the authentication parameter, currently it is mandatory so it's not necessary to check, but I think it's better to leave it anyway in case we change that.

nitzanshapira:
@Spivaka - I fixed your comments (as well as a big refactoring - removing the use of qArgs, qParams) - please take a look.

nitzanshapira:
p.s. (to @JonathanMeler ) The current SEP scripts are out-of-date. They should be removed. Should I remove them from the repository?

JonathanMeler:
@liorkol can we make them deprecated?

nitzanshapira:
@aviadl suggested that we delete them, as they are not backward compatible (and basically useless, since Lior re-defined everything anyway).

PR #280, opened 2017-04-06
cisco-meraki

asafshen:
fixes https://github.com/demisto/server/issues/6334

implementation of most API calls

### DIFF
--- a/Integrations/integration-cisco-meraki.yml
+++ b/Integrations/integration-cisco-meraki.yml
@@ -107,8 +107,9 @@ script:
     var apiKey = params.apiKey;
 
     var sendRequest = function(url, method, body) {
-        var requestUrl = merakiUrlPrefix + url;
+
         var requestMethod = method || 'GET';
+        var requestUrl = merakiUrlPrefix + url;
 
         var httpParams = {
             Method: requestMethod,
@@ -116,7 +117,7 @@ script:
                 'X-Cisco-Meraki-API-Key': [apiKey],
                 'Content-Type': ['application/json'],
             },
-            body: body
+            Body: body
         };
 
         var res = http(
@@ -124,10 +125,11 @@ script:
             httpParams,
             false,
             false,
-            true // no redirect
+            true // no redirect (e.i. request is not automatically redirect - instead it will return 302)
         );
 
-        if (res.StatusCode === 302) {
+        // catch redirect response
+        if (!res || res.StatusCode === 302) {
             var redirectUrl = res.Headers.Location;
 
             res = http(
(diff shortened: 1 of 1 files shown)


Integrations/integration-cisco-meraki.yml
```diff
@@ -14,10 +14,13 @@ configuration:
...
-
-    var merakiUrlPrefix = 'https://dashboard.meraki.com/api/v0/';
-
+  script: |
+    /**
+       * flatten nested object to top level fields
+       * @param {Object | Array<Object>} ob - object to flatten
+       * @returns {String} the flatten object
+       * { "a": "bla", "b": { "c": "hello", "d": "there" } } => { "a": "bla", "b.c" = "hello", "b.d": "there" }
+    */
```
Spivaka: should be in common server for a generic usage

Integrations/integration-cisco-meraki.yml
```diff
@@ -38,12 +41,10 @@ script:
         return toReturn;
     };
 
-    // ------------------------ mics -------------------- //
-
     /**
        * Converts a demisto table in JSON form to a Markdown table
        * @param {String} name - the name of the table
-       * @param {Object} table - the JSON table - Array of objects with the same keys
+       * @param {Object | Array<Object>} table - the JSON table - Array of objects with the same keys
```
Spivaka: Also look like a function for general use (and a well written one, :+1:) so please move to common

Integrations/integration-cisco-meraki.yml
```diff
@@ -102,28 +103,39 @@ script:
...
+    var apiKey = params.apiKey;
+
     var sendRequest = function(url, method, body) {
         var requestUrl = merakiUrlPrefix + url;
-        var apiKey = params.apiKey;
         var requestMethod = method || 'GET';
 
-
         var httpParams = {
             Method: requestMethod,
```
Spivaka: kine 111 should be here, no other usage of method. probably the same for line 110

Integrations/integration-cisco-meraki.yml
```diff
@@ -102,28 +103,39 @@ script:
...
             Headers: {
                 'X-Cisco-Meraki-API-Key': [apiKey],
                 'Content-Type': ['application/json'],
             },
+            body: body
```
Spivaka: add insecure and proxy here

Spivaka: I think body should be Body in order to be marshalled

Integrations/integration-cisco-meraki.yml
```diff
@@ -102,28 +103,39 @@ script:
...
         var res = http(
             requestUrl,
-            httpParams
+            httpParams,
+            false,
+            false,
+            true // no redirect
         );
 
+        if (res.StatusCode === 302) {
```
Spivaka: How can we be redirected if the no redirect is on by default?

Integrations/integration-cisco-meraki.yml
```diff
@@ -135,7 +147,7 @@ script:
             // body is empty - than just return true (action succeed)
```
Spivaka: condition doesn't match comment... fix one of them (probably the code as the comment makes sense)

Integrations/integration-cisco-meraki.yml
```diff
@@ -135,7 +147,7 @@ script:
...
             return true;
         } else {
-             var body = res && JSON.parse(res.Body);
+             var body = JSON.parse(res.Body);
```
Spivaka: I think the older version was better, would also add res && res.Body

Integrations/integration-cisco-meraki.yml
asafshen: the 'no-redirect' means that http-client is not automatically redirects the request - instead it returns 302/3/4

maybe we should change its name if its not clear. anyway I will add better doc on this integration

PR #283, opened 2017-04-09
Fixed utility tag

dorsha:


Scripts/script-InviteUser.yml
```diff
@@ -5,7 +5,7 @@ name: InviteUser
 script: return executeCommand('@', args);
 type: javascript
 tags:
-- util
+- Utility
```
meirwah: Is this really a utility script @liorkol ????

PR #295, opened 2017-04-13
Okta new arch

gshriki:
Integration with new model.
Deprecated all scripts - all are implemented within integration.


slavikm:
@gshriki please convert the files and remove the dos enters (^M)

Integrations/integration-Okta.yml
```diff
@@ -2,7 +2,6 @@ commonfields:
   id: okta
   version: -1
 name: okta
-system: true
 display: okta
```
slavikm: We removed system? Is this final?

Integrations/integration-Okta.yml
```diff
@@ -193,6 +192,105 @@ script:
       return creds;
     };
 
+    var usersDict = {
```
slavikm: Like it on one line - `var usersDict = {};` - also missing ';'

Integrations/integration-Okta.yml
```diff
@@ -193,6 +192,105 @@ script:
...
+
+    }
+
+    var identLevel = function(count) {
+        var ret='';
+        for (var i=0;i<count;i++) {
+            ret += '     ';
+        }
+        return ret;
+    }
```
slavikm: `;` after every function because it's a statement

Integrations/integration-Okta.yml
```diff
@@ -337,6 +474,15 @@ script:
...
     - name: userid
       description: User ID of requested user
+    - name: verbose
+      auto: PREDEFINED
+      predefined:
+      - "true"
+      - "false"
+      description: Print all details of user
+    outputs:
+    - contextPath: Account
```
slavikm: Need to specify the leafs (fields) for the account with a separate description for each - otherwise not usable in playbook.

Integrations/integration-Okta.yml
```diff
@@ -415,21 +561,24 @@ script:
...
+      description: OKTA, ACTIVE_DIRECTORY,LDAP, FEDERATION, or SOCIAL
     - name: providerName
       description: Name of provider
     - name: groupIds
       description: Ids of groups that user will be immediately added to at time of
         creation (Do not include default group)
     - name: activate
       description: Executes activation lifecycle operation when creating the user
+    outputs:
+    - contextPath: Account
```
slavikm: Same as above

Integrations/integration-Okta.yml
gshriki: Not sure what is the comment here. Do we need system:true or not?
@aviadl?

aviadl: i am adding this automatically when populating via content update.
but you will need to add fromversion in that case

gshriki: OK. Can you point me to an example?

aviadl: example for what ?
Place this file in your res directory, 
change the assetid in the content-descriptor file
and restart the server
okta should be available, and treated as system (no drill in)

gshriki: example of how to specify fromversion... That's what you're asking me to add, no?

Integrations/integration-Okta.yml
gshriki: removed it altogether... not needed...

Integrations/integration-Okta.yml
gshriki: done

Integrations/integration-Okta.yml
gshriki: done

Integrations/integration-Okta.yml
gshriki: added

PR #302, opened 2017-04-14
added depends on a nexpose command vulnerability-list

gshriki:
Added the dependency on the Nexpose scripts to make sure we have the vulnerability-list command (part of Nexpose integration), so Nexpose scripts won't show up in the CLI when there is no Nexpose instance set up.

### DIFF
--- a/Scripts/script-NexposeEmailParser.yml
+++ b/Scripts/script-NexposeEmailParser.yml
@@ -81,5 +81,7 @@ args:
 - name: defaultNexposeSeverity
   description: Severity to be set on triggered incidents
 scripttarget: 0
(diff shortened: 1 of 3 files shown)


meirwah:
not sure about this.
adding @Spivaka 
I think Email Parser does not need depends on.
because this is the exact workaround when you want to fetch from email instead of Nexpose directlly

gshriki:
Yes... That was my thought as well. But think of all the customers who do not have Nexpose... Now they have to see that in the autocomplete, and other places (Task Library). Maybe we can have a different approach - disable those by default, and then customers can enable if desire... Thoughts?

PR #308, opened 2017-04-15
Is greater then condition

meirwah:


slavikm:
Actually, please change the name to IsGreaterThan with an `a`

PR #309, opened 2017-04-15
add count array size

meirwah:


Scripts/script-CountArraySize.yml
```diff
@@ -0,0 +1,26 @@
+commonfields:
+  id: CountArraySize
+  version: -1
+name: CountArraySize
+script: |-
+  var arr = args.array;
+  if (Array.isArray(arr)) {
+      return {Type: entry.note, Contents: arr.length, ContentsFormat: formats.text, EntryContext: {ArraySize: arr.length}}
```
aviadl: I dont think that this should be part of context ...
you will get this way really weird stuff
Maybe add as argument if and where to place in context

slavikm: I would just use `setContext` to always override it

PR #310, opened 2017-04-15
add getTime script

meirwah:


Scripts/script-GetTime.yml
```diff
@@ -0,0 +1,27 @@
+commonfields:
+  id: GetTime
+  version: -1
+name: GetTime
+script: |-
+  var d = new Date();
+  if ((args.date) && (args.date.length > 0)){
```
slavikm: Personal preference is not to add extra brackets when not required

Scripts/script-GetTime.yml
```diff
@@ -0,0 +1,27 @@
...
+      d =  new Date(args.date);
+  }
+  return {Type: entryTypes.note, Contents: String(d), ContentsFormat: formats.text,
+      EntryContext: {TimeNowUnix: String(d.getTime()), TimeNow: String(d)}};
```
slavikm: Why converting to `String` a perfectly good long?

PR #317, opened 2017-04-17
adding cisco umbrella investigation threatgrid will follow later based…

trorabaugh:
… on requests. This also uses the new arch.


slavikm:
Nice.
When specifying the outputs - you need to document the leafs and not the object. Also, just returning the entire json to context might be an overkill - you want to expose only the interesting parts. The user already has the `Contents` as the raw response so if he wants something else from it he can get it via `Extend` property.

trorabaugh:
will do

meirwah:
@trorabaugh what's the statues of this?

liorkol:
Being merged in another PR

PR #325, opened 2017-04-17
Bugfix/new wildfire

liorkol:
fixes https://github.com/demisto/server/issues/6463


Integrations/Integration-Wildfire.yml
```diff
@@ -148,19 +131,36 @@ script:
               return 'Something went wrong.\n Status code: ' + res.StatusCode + '\nBody: '+ JSON.stringify(res) + '.';
           break;
 
-      case 'WildFireReport':
+      case 'wildfire-report':
```
slavikm: Just verifying that this was the same command as before - we cannot change the commands or at least we need to support both.

Integrations/Integration-Wildfire.yml
```diff
@@ -148,19 +131,36 @@ script:
...
 
         var bodyXML = 'apikey='+params.token+'&format=xml&hash='+hash;
         var resXML = sendRequest(url, bodyXML, headers);
         var result = JSON.parse(x2j(resXML));
         var report = result.wildfire.task_info.report;
         var file_info = result.wildfire.file_info;
 
-        var context = parseFile(result.wildfire.file_info);
+        var context = { dbotScore: {indicator: hash, type: 'hash', vendor: 'Wildfire', score: 0 }};
+        var body = result.wildfire.file_info;
```
slavikm: You already have file_info a few lines above

Integrations/Integration-Wildfire.yml
```diff
@@ -148,19 +131,36 @@ script:
...
-        var hash = args.md5 || args.sha256;
+        var hash = args.md5 || args.hash;
         if(!hash)
-            throw 'Missing md5/sha256';
+            throw 'Missing md5/hash argument.';
 
         var bodyXML = 'apikey='+params.token+'&format=xml&hash='+hash;
         var resXML = sendRequest(url, bodyXML, headers);
         var result = JSON.parse(x2j(resXML));
         var report = result.wildfire.task_info.report;
```
slavikm: Please use `dq` to access nested elements - otherwise it will throw an error if the element does not exist

Integrations/Integration-Wildfire.yml
```diff
@@ -191,43 +193,67 @@ script:
...
           var md = resToMD(result.wildfire['upload-file-info'],'Wildfire Report');
 
           return {Type: entryTypes.note,
                       Contents: result,
                       ContentsFormat: formats.json,
-                      HumanReadable: md, /*todo*/
-                      ReadableContentsFormat: formats.markdown,
-                      EntryContext: context
+                      HumanReadable: md,
+                      ReadableContentsFormat: formats.markdown
```
slavikm: Default, no need to specify

Integrations/Integration-Wildfire.yml
```diff
@@ -191,43 +193,67 @@ script:
...
   type: javascript
   commands:
-  - name: WildFireReport
+  - name: wildfire-report
     arguments:
     - name: md5
-    - name: sha256
+      description: MD5 hash to check
+    - name: hash
+      description: SHA256 hash to check
```
slavikm: Also SHA1, no?

liorkol: No..

Integrations/Integration-Wildfire.yml
liorkol: Ok.

Integrations/Integration-Wildfire.yml
liorkol: Verified. wildfire-report is the old command

Integrations/Integration-Wildfire.yml
liorkol: Ok. Changing.

PR #327, opened 2017-04-17
Fixed from version for migrated integrations

aviadl:


Integrations/integration-VirusTotal.yml
```diff
@@ -645,3 +645,4 @@ script:
       description: Datetime token in format YYYYMMDDHHMISS that can be used for paging
     description: Private API. Retrieve comments for a given resource
 hidden: false
+fromversion: 2.5.0
```
dorsha: there is a red icon near the 2.5.0 (for new line missing).
Hopes it won't cause issues

aviadl: Didnt cause me issues when i tested it on a specific integration

PR #331, opened 2017-04-18
utility to add specific entry comments to context

gshriki:
Use case: analyst puts some comments on a manual task he is doing. Later in playbook a task takes those comments and adds them to a sent email. (SureScripts)


liorkol:
@slavikm @gshriki Declare outputs in yaml?

PR #337, opened 2017-04-19
Converting AWS Integration to new format

maiorrulz:
[https://github.com/demisto/server/issues/6456](url)

### DIFF
--- a/Integrations/integration-aws.yml
+++ b/Integrations/integration-aws.yml
@@ -42,168 +42,232 @@ script:
     var apiVersion = '2016-09-15';
 
     var sendRequest = function(method, url) {
-            var res = http(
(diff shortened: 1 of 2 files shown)


Integrations/integration-aws.yml
```diff
@@ -42,168 +42,232 @@ script:
     var apiVersion = '2016-09-15';
```
meirwah: Do we need here ```fromVersion: 2.5``` ? @aviadl ?

aviadl: Don't honk so, what is the limitation here ?

meirwah: So @maiorrulz you are good to go

PR #340, opened 2017-04-19
Create script-WhileLoop.yml

meirwah:
add whileloop script


meirwah:
@slavikm if you can review too, plz do...

PR #342, opened 2017-04-19
Create playbook-ATD-sandbox

meirwah:


meirwah:
@liorkol plz review this, it's important

meirwah:
@Spivaka can you review?

Spivaka:
I think you should verify that the context is filled after getting the report.
I also don't understand where the File.ID is coming from.
Rest looks great!

PR #343, opened 2017-04-19
Update script-CommonServer.yml

shaniacht1:
Changed 1 line in tabltToMd to make sure that value '0' is not treated as null

### DIFF
--- a/Scripts/script-CommonServer.yml
+++ b/Scripts/script-CommonServer.yml
@@ -116,7 +116,7 @@ script: |-
           t.forEach(function(entry){
               var vals = [];
               headers.forEach(function(h){
-                  vals.push(!!entry[h] ? stringEscapeMD(formatCell(entry[h])) : ' ');
+                  vals.push((!!entry[h] || entry[h]==0) ? stringEscapeMD(formatCell(entry[h])) : ' ');
               });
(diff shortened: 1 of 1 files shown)


Scripts/script-CommonServer.yml
```diff
@@ -116,7 +116,7 @@ script: |-
           t.forEach(function(entry){
               var vals = [];
               headers.forEach(function(h){
-                  vals.push(!!entry[h] ? stringEscapeMD(formatCell(entry[h])) : ' ');
+                  vals.push((!!entry[h] || entry[h]==0) ? stringEscapeMD(formatCell(entry[h])) : ' ');
```
slavikm: If I'm not mistaken, default JS is to try and convert to int using this syntax and if the entry is a string it might throw. Please verify.

shaniacht1: I checked it again: '0G' == 0 returns false, while '00' == 0 returns true (obviously also '0'==0), none of them throw.

Spivaka: @slavikm do you want to test something else here or we can merge?

slavikm: nope - looks good.

PR #351, opened 2017-04-20
Added Docx Reports

Bars92:
related to: https://github.com/demisto/server/issues/3284


Reports/report-docx-criticalAndHighIncidentsReport.json
```diff
@@ -0,0 +1,374 @@
+{
+  "id": "docxCriticalAndHighIncidentsReport",
+  "name": "Critical and High incidents",
+  "description": "All critical and high severity incidents that may need the analyst attention.",
```
dorsha: please add to all descriptions:
(DOC format).

Like with the CSV:
image

Bars92: Isn't this redundant let alone confusing ? This details is the details of the report. The symbol near it shows clearly what format this is going to be. I think we should remove it from the description at all..

dorsha: A bit redundant but users may not pay attention to the icon... your choice

PR #356, opened 2017-04-21
Create script-Base64Encode.yml

meirwah:
Base64Encode script

### DIFF
--- a/Scripts/script-Base64Encode.yml
+++ b/Scripts/script-Base64Encode.yml
@@ -1,0 +1,26 @@
+commonfields:
+  id: Base64Encode
+  version: -1
+name: Base64Encode
+script: |-
+  var result = Base64.encode(args.input);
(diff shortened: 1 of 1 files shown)


Scripts/script-Base64Encode.yml
```diff
@@ -0,0 +1,26 @@
+commonfields:
+  id: Base64Encode
+  version: -1
+name: Base64Encode
+script: |-
+  var result = Base64.encode(args.input);
```
slavikm: What's Base64 - are we exposing this somewhere?
screen shot 2017-04-21 at 2 02 37 pm

meirwah: It should be in the commons
You have last content version?

slavikm: Yup, I did not have the latest. It is in `Common`.

PR #359, opened 2017-04-23
Bring QRadar to the new standard

JonathanMeler:
fixes https://github.com/demisto/server/issues/6328


Scripts/script-QRadarFullSearch.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          if (q_status && q_status in ['WAIT', 'EXECUTE', 'SORTING']){
+              log('search id is {0} search status is {1}'.format(search_id,q_status));
+          }
+          else if (q_status && q_status == 'COMPLETED'){
+              search_done = true;
+          }
+          else{
+              error = true;
+              return {"Type" : entryTypes.error, "ContentsFormat" : formats.text, "Contents" : 'An Error occurred during the search process. search status={0}.'.format(status)};
```
Spivaka: No need for "Type" in JS, Type: is enough (sets it as a string key automatically)

PR #362, opened 2017-04-23
Use wildfire report

orius123:
Use `wildfire-report` command


Spivaka:
why is this needed if you changed to file in wildfire?

orius123:
because `file` is deprecated

PR #379, opened 2017-04-24
Tanium update to new standard

JonathanMeler:
https://github.com/demisto/server/issues/6462


Scripts/script-TaniumApprovePendingActions.yml
```diff
@@ -65,11 +65,12 @@ script: |
   var res = executeCommand('tn-get-object', getObjArgs);
   var table = {
```
Spivaka: probably good to rename this to entry

Scripts/script-TaniumAskQuestion.yml
```diff
@@ -19,8 +19,10 @@ script: |-
 
   var table = {
```
Spivaka: same here

Scripts/script-TaniumAskQuestion.yml
```diff
@@ -90,6 +92,9 @@ script: |-
       if (table.Contents.length === 0) {
          return 'No Results';
       }
+      var title = 'name' in args ? args.name : "Tanium ask question";
```
Spivaka: We usually use args.name and not name in args (this will return undefined if name doesn't exist in object, which is equivalent to false)

Scripts/script-TaniumDeployAction.yml
```diff
@@ -41,7 +42,19 @@ script: |-
         'target-group': targetGroup,
       };
   var res = executeCommand('tn-deploy-package', taniumArgs);
-  return res
+  if (isError(res[0])) {
+      return res;
+  }
+
+  var table = {
```
Spivaka: same here, entry is better

Scripts/script-TaniumFindRunningProcesses.yml
```diff
@@ -5,8 +5,11 @@ name: TaniumFindRunningProcesses
 script: |-
   var table = {
```
Spivaka: and here

Scripts/script-TaniumShowPendingActions.yml
```diff
@@ -69,8 +69,10 @@ script: |
   var res = executeCommand('tn-get-object', getObjArgs);
   var table = {
```
Spivaka: +1

PR #380, opened 2017-04-24
Crowdstrike host - handle no results case

liorkol:
fixes https://github.com/demisto/server/issues/6888

### DIFF
--- a/Integrations/integration-CrowdStrikeHost.yml
+++ b/Integrations/integration-CrowdStrikeHost.yml
@@ -58,7 +58,7 @@ script:
             params.insecure,
             params.useproxy
         );
-        if (result.StatusCode < 200 && result.StatusCode > 299) {
+        if (result.StatusCode < 200 || result.StatusCode > 299) {
             throw 'Failed to perform request ' + path + ', request status code: ' + result.StatusCode + ', body: ' + result.Body;
         }
         if (result.Body === '') {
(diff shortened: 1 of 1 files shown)


Integrations/integration-CrowdStrikeHost.yml
```diff
@@ -71,7 +71,11 @@ script:
             throw 'Error parsing reply - ' + result.Body + ' - ' + ex;
         }
         if (obj.errors && obj.errors.length > 0) {
-            throw JSON.stringify(obj.errors);
+            if (dq(obj, 'errors.[0].code') === 404 && dq(obj, 'errors.[0].message').indexOf('Resource Not Found') > -1) {
```
Spivaka: I think this will fail in line 62 as the code will be 404 with this body... would check and handle differently in case of 404 is status code

liorkol: Well.. tested successfully, but due to the older bug we saw :) I'll push a fix

Integrations/integration-CrowdStrikeHost.yml
```diff
@@ -58,22 +58,35 @@ script:
             params.insecure,
             params.useproxy
         );
-        if (result.StatusCode < 200 && result.StatusCode > 299) {
```
Spivaka: @slavikm this piece of code is AMAZING 💯

slavikm: Hey, in quantum physics an int can be in two states at once :-)

PR #381, opened 2017-04-24
Playbook iochunt upgrade

liorkol:
And adding two subplaybooks (Crowdstsrike and Bit9)
Tested all 3


dorsha:
@aviadl - please review.

Scripts/script-ReadFile.yml
```diff
@@ -0,0 +1,33 @@
+commonfields:
+  id: eccea40d-921c-4a52-875a-738855531f58
```
aviadl: static ID
Also please limit by size, if someone will load a few Giga file here we would die

Scripts/script-ReadFile.yml
```diff
@@ -0,0 +1,33 @@
...
+name: ReadFile
+script: |-
+  res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
+  with open(res[0]['Contents']['path'], 'r') as f:
+      data = f.read()
+  if data:
+      message = 'Read %d bytes from file.' % (len(data))
+      demisto.results({"Type" : entryTypes["note"],
+                       "ContentsFormat" : formats["text"],
+                       "Contents" : message,
```
aviadl: Put the data on the Contents, so one can extend the context with value fetched from file to other places

Playbooks/playbook-CrowdStrike_Rapid_IOC_Hunting.yml
```diff
@@ -0,0 +1,398 @@
+id: playbook16
```
aviadl: version should be -1 (on all tasks)
Change all conditions to else

Playbooks/playbook-Carbon_black_Protection_Rapid_IOC_Hunting.yml
```diff
@@ -0,0 +1,89 @@
+id: playbook17
```
aviadl: version should be -1

PR #388, opened 2017-04-24
Changed incName to name in scripts

dorsha:


aviadl:
@dorsha I dont think that we need this now ...

PR #391, opened 2017-04-25
add 'no' condition to crowdstrike in phishing  playbook

meirwah:


meirwah:
@liorkol plz review

meirwah:
Why? why we need to add 'else' always ?

liorkol:
Following our conversation, I am approving this.
Playbook will stop if task doesn't return yes / no or if there's an error. User should be able to see that their incident is paused and why.

PR #397, opened 2017-04-25
Fix to md to tbl branch

maiorrulz:


dorsha:
Merged :)

dorsha:
the thing is that there were a bug.. if you don't provide any header it will never work.. it is a bit critical

dorsha:
So users must now provide headers to this function

PR #399, opened 2017-04-26
Fix categories

Spivaka:
fixes https://github.com/demisto/server/issues/6944


Integrations/integration-mcafeeDam.yml
```diff
@@ -3,7 +3,7 @@ commonfields:
   version: -1
 name: McAfeeDAM
 display: McAfeeDAM
-category: Authentication
+category: End Point
```
aviadl: DAM is not endpoint ...
@slavikm , @gshriki  how would you categorize it ?

Spivaka:
@aviadl please review and merge

PR #400, opened 2017-04-26
QRadar full search - bug fix

JonathanMeler:


Scripts/script-QRadarFullSearch.yml
```diff
@@ -25,7 +25,7 @@ script: |-
 
   //submit query, retrive  search_id
   var query_res = executeCommand("qradar-searches", args);
-  if (isError(query_res)){
+  if (isError(query_res[0])){
```
aviadl: are you sure you will always get a single entry ?

JonathanMeler: yes. that's how the API works, you enter a single query and get it's id

PR #405, opened 2017-04-26
Added a small util to retrieve and split lists

slavikm:
and then used it in the extract scripts to ignore whitelist

### DIFF
--- a/Scripts/script-CommonServer.yml
+++ b/Scripts/script-CommonServer.yml
@@ -548,6 +548,23 @@ script: |-
       return [];
   }
 
+  /**
+   * Get list data which is JSON format into an object that can be an array as well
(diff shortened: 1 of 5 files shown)


Scripts/script-CommonServer.yml
```diff
@@ -534,6 +534,20 @@ script: |-
       return res;
   }
 
+  /**
+   * Get list data which is comma-separated into an array
+   * @param {String} listName - the name of the list
+   * @returns {Array} the list as an array
+   */
+  function getCSVListAsArray(listName) {
```
aviadl: add an optional format parameter,
Since lists maybe managed via scripts and then be in proper json format

slavikm: I will add another function to get list as object

PR #406, opened 2017-04-27
Fix logos

Spivaka:
fixes https://github.com/demisto/server/issues/6956


Spivaka:
Should be easier than Hebrew!

PR #410, opened 2017-04-27
Req.txt

meirwah:


meirwah:
I'm adding a requirement.txt file to keep track of python dependencies we have in python

PR #413, opened 2017-04-28
Create Integration-Samanage.yml

shadejinx:
This is the first version of my Samanage integration. Currently it only exposes hardware and user queries. I will be adding case query and modification in the next version.


Integrations/Integration-Samanage.yml
```diff
@@ -0,0 +1,119 @@
...
+    - name: param
+      required: true
+      auto: PREDEFINED
+      predefined:
+      - name
+      - email
+    - name: query
+      required: true
+    description: Query Samanage for user information
+  dockerimage: aviadlich/python-requests
```
meirwah: you can switch this to demisto/python
Our official image includes python requests

Integrations/Integration-Samanage.yml
```diff
@@ -0,0 +1,119 @@
+commonfields:
+  id: Samanage
+  version: -1
+name: Samanage
+display: Samanage
+category: Data Enrichment
```
meirwah: Is it really Data Enrichment ? isn't it  "IT Services" or "Case Management"?

shadejinx: Probably. My current use case is data enrichment, but I have plans to add minimal case management in the future.

Integrations/Integration-Samanage.yml
shadejinx: excellent, I will update my scripts to include.

gshriki:

PR #414, opened 2017-04-28
Create script-SamGetHostSummary

shadejinx:
This script uses the Samanage interface to get a summary table of information about a specific hostname


gshriki:

PR #415, opened 2017-04-28
Create script-SamGetUser

shadejinx:
This script uses the Samanage Integration to get User information from a hostname. The script can handle multiple hostnames via CSV.


gshriki:

PR #416, opened 2017-04-30
Proxy in guidance is string and not boolean

Spivaka:


Spivaka:
fixes https://github.com/demisto/server/issues/7018

PR #419, opened 2017-05-01
Threatexchange 2.5

meirwah:


meirwah:
@slavikm @aviadl plz review

PR #420, opened 2017-05-01
Fix extract URl to provide unique results

aviadl:
fixes https://github.com/demisto/server/issues/7008


aviadl:
@slavikm Please review it after all, since the readable output of the script will be unique now as well, which is better

PR #423, opened 2017-05-01
Replace Script-GetUser.yml with Script-GetUserFromHost.yml

shadejinx:
Changed the name to be more descriptive. Fixed loop bug that prevented CSV input.


gshriki:

PR #426, opened 2017-05-03
Arcsight integration

nitzanshapira:
Fixes https://github.com/demisto/server/issues/5742.

Note: creating XML files ("update case") only in a local or shared directory. Remote execution via "executeCommand" is not supported in an integration.
If needed, it can be added as a separate Script.


Spivaka:
@liorkol python is not my speciality

aviadl:
@slavikm can you please decide with @liorkol who will review this you or him ?

slavikm:
got it.

Integrations/integration-ArcSightXML.yml
```diff
@@ -0,0 +1,193 @@
...
+        case_data = [{'type': key, 'value': value} for key, value in case_data.iteritems()]
+
+        return case_data
+
+
+    def create_incidents(all_labels, all_details):
+        incidents = []
+        for labels, details in zip(all_labels, all_details):
+            incident = {
+                'Name': 'ArcSight File Event',
```
liorkol: Please use the case name instead

Integrations/integration-ArcSightXML.yml
```diff
@@ -0,0 +1,193 @@
...
+    SECURITY_EVENT_END_TAG = '</SecurityEvent>'
+
+    LIST_CASE_ENTRIES = ['caseEvents', 'childOf', 'ownedBy']
+    TIME_CASE_ENTRIES = ['detectionTime', 'estimatedStartTime']
+
+    SYSTEM = ''
+    URI_PREFIX = ''
+    TOREPLACE = '\\"\\<\\>\\(\\)'
+
+    XML_TEMPLATE = """<?xml version="1.0" encoding="UTF-8"?>
```
liorkol: This template did not work, please make it smaller and stick to the format I gave you.

Integrations/integration-ArcSightXML.yml
```diff
@@ -0,0 +1,193 @@
...
+        return case_data
+
+
+    def create_incidents(all_labels, all_details):
+        incidents = []
+        for labels, details in zip(all_labels, all_details):
+            incident = {
+                'Name': 'ArcSight File Event',
+                'Labels': labels,
+                'Details': details
```
liorkol: Should not put large XMLs in Details field. Need another solution for raw inputs in a way that user can browse them if they really want to, but don't interfere when users are trying to look at incident fields.

Integrations/integration-ArcSightXML.yml
```diff
@@ -0,0 +1,193 @@
...
+
+        return case_data
+
+
+    def create_incidents(all_labels, all_details):
+        incidents = []
+        for labels, details in zip(all_labels, all_details):
+            incident = {
+                'Name': 'ArcSight File Event',
+                'Labels': labels,
```
liorkol: Not all case fields are interesting to stick in labels. Lots of noise here.

Integrations/integration-ArcSightXML.yml
```diff
@@ -0,0 +1,193 @@
...
+
+
+    def parse_arcsight_xml(filepath):
+        with open(filepath, 'rb') as f:
+            xml_content = f.read()
+
+        tree = ET.parse(filepath)
+        root = tree.getroot()
+
+        security_events = root.findall('SecurityEvent')
```
liorkol: Lots of useful data is buried in this XML. Users need a way to specify a list of interesting arcsight fields *inside the base events (not the case)* that will be extracted (e.g. sourceUserName)

Integrations/integration-ArcSightXML.yml
nitzanshapira: Please explain, what format?

Integrations/integration-ArcSightXML.yml
nitzanshapira: Are you sure? in this case the incident name will be: _system_case_Event_Export_Request_4-3-2017_14-36-31.255, not very meaningful, is it?

Integrations/integration-ArcSightXML.yml
nitzanshapira: There is no specification so I just put all, as we discussed. Please see the attached screenshots. I can't think of any field that I know is certainly redundant. I think that it's better to keep all of them. What do you think?
If you can specify exactly which fields you think should be removed it will be best.
image
image

Integrations/integration-ArcSightXML.yml
nitzanshapira: Any hint? 
The details field is textual, isn't it?

Integrations/integration-ArcSightXML.yml
nitzanshapira: First, this is a redundant line (113), the "security_events" parameter is unused. I removed it.
Second, I agree that there is more information inside the "SecurityEvent" object which can be interesting .However, this is a "fetch incident"  action. How should the user define these attributes? A list of parameters when creating the instance? or maybe this information should be included automatically in the context of every incident created for an inside case?

nitzanshapira:
@liorkol please see my comments. Some things need to be clarified before implementing features which may be unwanted.

liorkol:
@nitzanshapira Please get incident name from case name (in actual customer setup it was a good indicative name), and then we can merge.
Regarding the other items - I will follow up later.

PR #432, opened 2017-05-04
Threat grid integration

Spivaka:
fixes https://github.com/demisto/server/issues/6233


Spivaka:
2nd commit fixes https://github.com/demisto/server/issues/7092

Integrations/integration-Threat_Grid.yml
```diff
@@ -0,0 +1,922 @@
...
...zRxthkAvSieGnAXk5A0a+wOt6vyfM3k20g8ySgkze352u5w1yPLT5t8O/hCq9RNcVSRp6+WxL2BP5mHJEljNlg3AkmSxsqiYmHxGSyGkyRppAZRGsKMG0biFcYgSWMj601NnpqwGE6SRizvTUy+29UdpI4LblPU2TeAZYYhSdJ4sUOcJEnjY1FRzf8ccGejkCRpxDb0Ox9BsbgkkZ0Uid09ICTNGrvD7Yosn5h8V8BjjELS3J+v7Q63Db1i4+SHI/EPRiEJWF2U+VdL4mjgIuOQJGk82CFOkqTxsKCo5n8Ki+EkSdL4uFckHmYMkjQWUr5i3ZsTHGcUkjRSea+/5D1YDCfp6g4psnQmsK9RSJI0HiyIkyRp9OYXU5v+G7i7UUiSNHqDy5wqA/MiZW8xBkmzyu5wOyvlp617S4p4ilFIGs752u5w13k+npp8Z8DDjULStVhV5IPPgV3WJUkaBy6ZKknSaPWLNP1RgnsYhSRJY2KJU+WC6acBN/FgkDSrPnq2GVyXS+dDVtMvSupUkCIIaiAB8RpIFsNJ0jCuU5fOZ/6CAVWVqFOQBURkkALgddipU9K2JG5ZMP25Eu4JbDAQSZJGx9feJUkanV6xsfdRgvsbhSRJGiO7Ay8yBkkaC6+C9HxjkKRxOB/zXGOQtB3uUGTpk8B8o5AkaXR87V2SpNHIev0l7wni741CkqQxE1WnN79I9QuBvTwQJM2qPRaYwY77V+CFxiBpqFwu9dq8xPOxpB109yJNf7SEY4CBcUiSNHwWxEmSNFtWX8evbzyXeXWi7lXUdUZGEKl+TUR6uKFJkjReBnXe9QgOgHimR4IkjdxxkF5mDJI0cs8mxSuMQdIOC+7f6y153wAeCVQGIknScFkQJ0nS8D2DSM8zBkmSxlDa0unNz8lfDMk2TpJml93hdtQ9gzgFSEYhaajsDndNDwvS641B0s6KxMOKjWt/l7LsWXVdk+WJQZVDymH9CgOSJGkOZUYgSdJQPSCINxiDJEkaQzdMpMcagySN1M2I+iNAzygkaaRum4JTsThZ0q5K6ZmAndglSRoyO8RJkjQ8t0tUH5p5/UuSJI2l6fmd3fSiP3guMN+DQNKssjvcjti/TtlnCXYzCklDZ3e4qzow4NOQFhqFpNkQEW8A1gP/bRqSJA2HHeIkSRqOg6L2RpokSeNssKXT74ztBul4jwJJG
```
slavikm: nice!

slavikm: You got this right :-)

slavikm: Not sure about the Contents - I think it should be empty for Type: 9

slavikm: I don't like that returnObject is defined below as almost global - should probably pass it in, no?

slavikm: Indent?

slavikm: I prefer using `"` on the outside so you don't need to escape

slavikm: If there is no array you can simply do:
```
addMalicious(ec, outputPaths.file, {MD5: md5[0], Malicious: {Vendor: 'ThreatGrid', Description: 'Sample has score of ' + returnObject.Score + ' which is higher than treshold (80)'});
```

slavikm: Do you think you should add this somewhere to the context? Maybe to IP?

slavikm: Same as above - otherwise it will be hard to use these samples in playbook

slavikm:
I have no idea why git placed my comments all in the same place

Spivaka:
it's not usable right now... I can't see to which things you relate

Integrations/integration-Threat_Grid.yml
slavikm: Line 44 - regarding the map function...

slavikm: Status code check :-)

slavikm: For file return, I think we don't need to return contents on line 80

slavikm: Line 151

slavikm: Line 316

slavikm: Line 322

slavikm: Line 324

slavikm: Line 366

slavikm: Line 392

Spivaka: :1st_place_medal:

Spivaka: I am quite sure we do, cause without it it want save as file correctly

Spivaka: The spec there is a bit missing yet... we should add it when we have more dta from customers

Spivaka: same - we don't understand it enough now... we should either consult customers or let Lior define it

PR #445, opened 2017-05-07
add important to threat grid

Spivaka:


Integrations/integration-Threat_Grid.yml
```diff
@@ -785,6 +785,13 @@ script:
       description: For malicious files, the vendor that made the decision
     - contextPath: File.Malicious.Description
       description: For malicious files, the reason for the vendor to make the decision
+    important:
+    - contextPath: File.Malicious.Vendor
```
aviadl: I think it should be:
File(val.Malicious)
See other integrations as examples.
And then you should remove the second important entry

Spivaka:
@aviadl please approve and merge

PR #447, opened 2017-05-07
Remedy on demand

asafshen:
- fixes: https://github.com/demisto/server/issues/3106
- remedy-on-demand integration
- playbook-test for remedy-on-demand integration
- add `expectedValue` to VerifyContext script

implements the following commands:
pasted image at 2017_05_04 12_34 pm


Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+  name: url
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Port
+  name: port
+  defaultvalue: "8008"
+  type: 0
+  required: true
+- display: Username
```
Spivaka: add proxy and insecure (always)

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+        };
+
+        var res = http(
+            url,
+            {
+                Method: 'POST',
+                Headers: {
+                    'Content-Type': ['application/x-www-form-urlencoded']
+                },
+                Body: encodeToURLQuery(body).replace(/^\?/, '')
```
Spivaka: if you use regex here, also please use it in line 28 mutating the address.
Either use string mutators or regex, but using both is confusing

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+            }
+        );
+
+        if (!res || res.StatusCode < 200 || res.StatusCode >= 300) {
+            throw 'Request Failed'
+                + '\nurl: ' + url
+                + '\nStatus code: ' + res.StatusCode
+                + '.\nBody: ' + JSON.stringify(res, null, 2);
```
Spivaka: didn't you create a call for that above (stringify?)

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+            throw 'Request Failed'
+                + '\nurl: ' + url
+                + '\nStatus code: ' + res.StatusCode
+                + '\nBody: ' + JSON.stringify(res);
+        }
+
+        return res;
+    };
+
+    var login = function() {
```
Spivaka: no need for logout as well?

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+                Headers: {
+                    'Content-Type': ['application/json'],
+                    'Authorization': ['AR-JWT ' + token]
+                },
+                Body: body
+            }
+        );
+
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            throw 'Request Failed'
```
Spivaka: if there is a need to logout, make sure you logout here in case of an error

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+               "Last_Name" : lastName,
+               "Description" : description,
+               "Status" : status,
+               "Reported Source": source,
+               "Service_Type" : serviceType,
+               "Impact" : impact,
+               "Urgency" : urgency
+         }
+        };
+        var res = sendRequest(url, token, "POST", JSON.stringify(body));
```
Spivaka: to many stringify and stringify overrides...

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+            // human-readable
+            ReadableContentsFormat: formats.markdown,
+            HumanReadable: tableToMarkdown(name, contents, headers),
+            // context
+            EntryContext: context
+        };
+    }
+
+    // mutetor that removes all falsly fields
+    var filterEmptyFields = function(obj) {
```
Spivaka: currently this removes undefined, empty string, null and false.
I am not sure you want to remove false. If so, rename cause it doesn't remove empty fields (false is not empty)

Integrations/integration-Remedy-On-Demand.yml
```diff
@@ -0,0 +1,448 @@
...
+
+        sendRequest(url, token, "PUT", JSON.stringify(body));
+        return getIncident(incID, 'Updated incident:');
+    }
+
+    switch (command) {
+        case 'test-module':
+            fetchIncidents();
+            return 'ok';
+        case 'remedy-incident-create':
```
Spivaka: don't we want to fetch incidents to Demisto from the integration?

PR #448, opened 2017-05-08
Net witness integ

maiorrulz:


Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -49,24 +49,115 @@ script:
...
+        return Object.keys(obj).length === 0;
+    }
+
+    var mapTypeToValue = function (fields) {
+        var result = {};
+        var TYPE = 'type';
+        var VALUE = 'value';
+        if(fields && Array.isArray(fields)) {
+            fields.forEach(function (element) {
+                if (element.hasOwnProperty(TYPE) && element[TYPE]) {
```
Spivaka: why do we need both hasOwnProperty and element[TYPE]?

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -49,24 +49,115 @@ script:
...
+        }
+        return result;
+    }
+
+    var buildHrAndContext = function (data) {
+        //Check if we have values
+        var retVal = {md:[], context: {Netwitness:{Events:[]}}};
+        var typeToValue;
+
+        if((typeof data) ==='object' && Array.isArray(data)) {
```
Spivaka: Would create an inner function to handle a single case, and call it instead of duplicating the code
see ThreatGrid as ref

Spivaka: don't mix typeof and .isType. choose one method and go with it (Array.isArray, instanceof Array, typeof () === 'array')
Beside I don't understand this condition as it checks data is an array and an object...

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -344,11 +439,95 @@ script:
...
+        case  'netwitness-query':
+            return buildMdAndContext(doReq('GET', commandToPath[command], args, args.responseType))
+        case 'netwitness-packets':
+            return parseDownloadResponse(doReq('GET', commandToPath[command], args, args.responseType));
         default:
             return parseResponse(doReq('GET', commandToPath[command], args, args.responseType));
     }
   type: javascript
   commands:
+  - name: netwitness-msearch
```
Spivaka: no outputs?

Spivaka: doesn't seem you have added outputs and important to any of the commands
this is important so it can be viewed and used inside playbooks

Spivaka:
fixes https://github.com/demisto/server/issues/6555

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
maiorrulz: Because It check if value of "TYPE" property is not undefined, not empty string and not null

PR #458, opened 2017-05-10
checkValue script - for conditional tasks

gshriki:
Checking value of a field/label/context - helpful in conditional tasks. Similar to Print, with few diffs:
1. Better name for playbook authoring
2. Value argument is not mandatory (so if input is not set, we return false)

### DIFF
--- a/Scripts/script-checkValue.yml
+++ b/Scripts/script-checkValue.yml
@@ -10,8 +10,9 @@ type: javascript
 tags:
 - Utility
 - Condition
-comment: 'Gets a value and return it. This is to be used in playbook conditional tasks
-  - get a value from incident field, label or context, and act accordingly. '
+comment: "Gets a value and return it. This is to be used in playbook conditional tasks
+  - get a value from incident field, label or context, and act accordingly. \nIf an
+  array is returned. the first value will be the decision making value."
 enabled: true
 args:
 - name: value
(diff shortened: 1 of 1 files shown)


Scripts/script-checkValue.yml
```diff
@@ -0,0 +1,20 @@
...
+script: |-
+  //check if value is provided, otherwise return false
+  //needed for custom fields which if not set initially will be null
+  return (args.value) ? args.value : false;
+type: javascript
+tags:
+- Utility
+- Condition
+comment: 'Gets a value and return it. This is to be used in playbook conditional tasks
+  - get a value from incident field, label or context, and act accordingly. '
```
aviadl: maybe add in the details using DT ?
If not you can merge

aviadl:
array is indeed a problem, but it is different from exists in the sense that it returns the value, so it can be used in a switch case, and not only `yes`, `no`

gshriki:
@slavikm - don't think this should not take arrays, as the intention is to use it in conditional tasks...

slavikm:
@gshriki but what happens if you're passed an array? You should solve that case, no? Otherwise the behavior is a bit weird. On the other hand, maybe it's the normal behavior - just to return all the values and so the first value that matches a branch will be chosen...

gshriki:
I'll run some tests to understand better the experience. Please do not merge yet.

gshriki:
So if an array is returned, the first value will be the decision making value. Added clarification on that in the script description. Ready to be merged.

PR #459, opened 2017-05-10
cisco umbrella fixes

asafshen:
- parse url to use only domain (support `domain/other/path`)
- fix stringify typo

### DIFF
--- a/Integrations/integration-Cisco-umbrella.yml
+++ b/Integrations/integration-Cisco-umbrella.yml
@@ -84,7 +84,7 @@ script:
             hostname = url.split('/')[2];
         }
         else {
-            hostname = url.split('/')[0];
+            hostname = url ? url.split('/')[0] : '';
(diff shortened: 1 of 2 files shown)


Integrations/integration-Cisco-umbrella.yml
```diff
@@ -77,18 +77,29 @@ script:
...
     var baseURL = params.baseURL;
 
     var domainArg = (args.domain) ? args.domain : '';
-    domainArg = domainArg.replace(/^https?:\/\//i, '');
-
-    // removes 'http://'/'https://'' prefix
-    var domainWithoutProtocol = function(domain) {
-        return domain && domain.replace(/^https?:\/\//i, '');
-    }
+    domainArg = extractDomainName(domainArg);
```
Spivaka: verify it works ok without the domain arg

PR #460, opened 2017-05-10
data itself is not an array and is a hash

trorabaugh:


slavikm:
@trorabaugh let's discuss tomorrow - not sure I understand this script...

trorabaugh:
not mine, existing script, was throwing out error because there was some code that popped off an array, data is a hash not array

slavikm:
I would deprecate it - all of it exists directly in integration.

meirwah:
so @slavikm @trorabaugh no need for this fix anymore?

aviadl:
@slavikm @trorabaugh can you please handle this ?

Scripts/script-VirustotalIsMalicious.yml
```diff
@@ -1,6 +1,6 @@
 commonfields:
   id: dbbdc2e4-6105-4ee9-8e83-563a4b991a89
-  version: 12
```
Spivaka: should be -1

Spivaka:
@slavikm @trorabaugh this is needed?
Either merge or close

trorabaugh:
It was noticed for customer who was using this. I recommended the raw command, and just patched this one for the customer. I don't remember off the top of my head. If you want to close it thats fine but the command was breaking because it was not always an array.

gshriki:
let's fix but also deprecate.

Spivaka:
as mentu
ioned, unattended pr will be closed.

PR #461, opened 2017-05-10
if any column is "0" then it does not appear to work

trorabaugh:
If any column is 0 then the if will fail. This should be if X is not None: blah blah blah. The reason is that if the col is set to 0 then it will fail.

### DIFF
--- a/Scripts/script-ParseCSV.yml
+++ b/Scripts/script-ParseCSV.yml
@@ -65,7 +65,7 @@ script: |
               next(csvdata)
 
           for row in csvdata:
-              if parseip is not None:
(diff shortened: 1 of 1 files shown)


Scripts/script-ParseCSV.yml
```diff
@@ -65,23 +65,23 @@ script: |
               next(csvdata)
 
           for row in csvdata:
-              if parseip:
+              if parseip is not None:
```
slavikm: Just need to make sure this works from playbook task because playbook tasks might pass parameters as empty string ('') and so this will be true even if user did not specify.

PR #467, opened 2017-05-14
Email parse - Push the new Labels to incident labels and not context key

JonathanMeler:
fixes https://github.com/demisto/server/issues/7219

old parser just put labels into context key, new version change the labels at the incident level

### DIFF
--- a/Scripts/script-ParseEmailFile.yml
+++ b/Scripts/script-ParseEmailFile.yml
@@ -485,17 +485,22 @@ script: |-
 
       temp_res = [{'ContentsFormat': formats['markdown'], 'Type': entryTypes['note'], 'Contents': md}]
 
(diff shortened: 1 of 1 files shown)


slavikm:
I'm not sure that this is always the requested behavior. I would definitely like to keep / know who forwarded the email, no?

JonathanMeler:
This is how it's being done in the original scripts. it just used the context keys 'Labels/Email/*' instead of setting the incident labels properly.

This script is used to parse an email into an investigation, the users can set the original aside and use it later (like it was done in the original phishing playbook)

slavikm:
@JonathanMeler please verify with existing playbooks. Also, might be problematic to change the script behavior as customers are already using it. The best way might be to deprecate this and add a new script that does as you say + stores the original data in another set of incident labels.

PR #471, opened 2017-05-15
1. Add delimiter to table, default is ','

maiorrulz:
Google App Play book
Carbon black protect playbook


Spivaka:
build fails... check it and show me in your env tomorrow

PR #475, opened 2017-05-17
added ArcSight ESM integration

anara123:
resolves https://github.com/demisto/server/issues/7242
resolves https://github.com/demisto/server/issues/7259


Integrations/integration-ArcSightESM.yml
```diff
@@ -0,0 +1,418 @@
...
+                server + url + encodeToURLQuery(args),
+                {
+                    Method: 'GET',
+                    Headers: {'Content-Type': ['application/x-www-form-urlencoded']}
+                },
+                params.insecure,
+                params.proxy
+            );
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            if (!login) {
```
Spivaka: you can check that the args have auth token that means we are not at the log in and cancel that input to this function

Spivaka:
also please create a go PR to remove arcsight and reconfigure the integrations in BYOI (there is an example in the upgrade file)

PR #487, opened 2017-05-22
GRR fixs

JonathanMeler:
Added old commands as depricated


Integrations/integration-GRR.yml
```diff
@@ -282,6 +282,26 @@ script:
                 urlArgs.offset = args.offset;
             }
             return getFlows('GET', server + '/api/clients/' + args.client_id + '/flows' + encodeToURLQuery(urlArgs), 'get flow',null,"items","Flows");
+        case 'grr_set_flows'://deprecated
```
Spivaka: do it with two cases pointing to the same code
case (old_command):
case (new_command):
// do something common here

PR #490, opened 2017-05-22
deprecate splunk scripts

meirwah:


meirwah:
@liorkol @JonathanMeler we need deprecate splunk scripts righT?

PR #494, opened 2017-05-23
Create integration-CheckpointFirewall

shaniacht1:


asafshen:
@shaniacht1 can you give in the description the relevant issue?

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+            var res = sendRequest('logout', '{}');
+            sid = undefined;
+            return JSON.parse(res.Body).message;
+        }
+        return 'Already logged out';
+    };
+
+    var add_host = function(){
+        sendRequest('add-host', JSON.stringify({name:args.ip, 'ip-address':args.ip}));
+        sendRequest('publish', '{}');
```
asafshen: can you explain what is the `publish` http is for?

shaniacht1: every change you do in checkpoint (add, set etc.) needs a publish command to save the outcome. Without it objects get locked by the session and access to them is locked.

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+        var temp = [];
+        var result;
+        var res;
+        try{
+            sendRequest('add-host', JSON.stringify({name:args.ip, 'ip-address':args.ip}));
+        }
+        catch(err){
+
+        }
+        sendRequest('publish', '{}');
```
asafshen: second parameter should be `{}` instead of `'{}'` - not crucial

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+        case 'checkpoint-show-access-rule-base':
+            answer = show_access_rule_base();
+            break;
+        case 'checkpoint-set-rule':
+            answer = set_rule();
+            break;
+        case 'checkpoint-delete-rule':
+            answer = delete_rule();
+            break;
+        // case 'CheckpointFWBackupStatus':     - script
```
asafshen: remove comment :)

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+
+    var delete_rule = function(){
+        if(args.uid || args.name || args.rule_number){
+            var body = {
+                uid : args.uid,
+                name: args.name,
+                'rule-number': args.rule_number,
+                layer: args.layer
+            };
+            res = sendRequest('delete-access-rule', JSON.stringify(body));
```
asafshen: you can use `var res = ...` instead and move `return JSON.parse(res.Body);` inside the body
(we should try to avoid global variables - without `var` declaration)

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+            logout();
+            throw 'Delete rule requires at least one of the following arguments - uid, name, rule_number';
+        }
+        return JSON.parse(res.Body);
+    };
+
+    var show_access_rule_base = function(){
+        var result = sendRequest('show-access-rulebase', JSON.stringify({name:args.name}));
+        var res = JSON.parse(result.Body);
+        var md = tblToMd('Access rulebase table',res.rulebase, Object.keys(res.rulebase[0]));
```
asafshen: isn't the 3rd argument redundant? (`Object.keys(res.rulebase[0])`)

Scripts/script-CPShowBackupStatus.yml
```diff
@@ -58,7 +59,8 @@ comment: Connect to a checkpoint firewall appliance using SSH and retrieve statu
   for backup tasks. The user account being used to access the device must be set to
   use the SSH shell and not the built in Checkpoint CLI. Consult the Checkpoint documentation
   for instructions on how to do this.
-system: true
+enabled: true
+deprecated: true
```
asafshen: this script is deprecated & yet we changed it?

Scripts/script-CPCreateBackup.yml
```diff
@@ -39,18 +45,19 @@ script: |
...
 type: python
 tags:
 - checkpoint
 comment: Connect to a checkpoint firewall appliance using SSH and trigger a task to
   create a configuration backup of the device. The user account being used to access
   the device must be set to use the SSH shell and not the built in Checkpoint CLI.
   Consult the Checkpoint documentation for instructions on how to do this.
-system: true
+enabled: true
+deprecated: true
```
asafshen: this script is deprecated & yet we changed it?

shaniacht1: I changed it so it would add the correct things to context, just in case anyone uses it. It's the same code as CheckpointFWCreateBackup

Integrations/integration-CheckpointFirewall
```diff
@@ -0,0 +1,707 @@
...
+      description: Last date of hits.
+    - contextPath: CheckpointFWRule.Hits.Level
+      description: Level of hits.
+    - contextPath: CheckpointFWRule.Hits.Percentage
+      description: Percentage of hits
+    - contextPath: CheckpointFWRule.Hits.Value
+      description: Value of hits.
+    description: Block one or more IP addresses using Checkpoint Firewall.
+  - name: checkpoint
+    arguments: []
```
shaniacht1: Since this !checkpoint should work with any given arg, and required args override the others (bug) I had to remove 'command' from the arguments. The 'checkpoint' command makes sure that 'args.command' is defined.

PR #498, opened 2017-05-24
ATDDetonate - Poll through ATD detonate process, URL and Files

JonathanMeler:
input: `atd-file-upload` input and `atd-get-report` reportType input

### DIFF
--- a/Scripts/script-ATDDetonate.yml
+++ b/Scripts/script-ATDDetonate.yml
@@ -1,0 +1,159 @@
+commonfields:
+  id: ATDDetonate
+  version: -1
+name: ATDDetonate
+script: |-
+  import json
+  from time import sleep
+
+  res = []
+  timeout = 960
+  dArgs = demisto.args()
+
+  if 'timeout' in demisto.args():
+      timeout = int(demisto.args()['timeout'])
+  interval = 10
+  if 'interval' in demisto.args():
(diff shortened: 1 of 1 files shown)


Spivaka:
@JonathanMeler either fix and merge or delete

JonathanMeler:
@Spivaka @aviadl In my opinion we can merge (I need someone to approve)
regarding aviad questions:
1. why playbook - this should be callable from the CLI as well
2. The context exists, it's created from the base-api calls
3. should I just copy the context/important fields from the base-api call?

aviadl:
if you call an api call from a script nothing is being placed into the context, it is the script responsibility to place it.
So you should declare it and handle it in the script (getting the data as is form the api call and forwarding it is good enough, but still needs to be done.

JonathanMeler:
I'm returning the result of the api as is:
' demisto.results(demisto.executeCommand('atd-get-report',{'taskId':taskID, 'type' : reportType}))'
The api call is from the integration, it contains human-readable and context.

I'll add declarations to the script

Spivaka:
@JonathanMeler add and I'll approve and merge

Spivaka:
@JonathanMeler ???

JonathanMeler:
@Spivaka sorry for the delay. please review the changes (the output was copied from the original integration api-command)

PR #510, opened 2017-05-28
Passivetotal url command

gshriki:
Added a url command to PassiveTotal which is an alias to pt-malware,
then removed the executeCommand for pt-malware in the DataURLReputation, as it will anyway be executed in the url command...


Integrations/integration-PassiveTotal.yml
```diff
@@ -484,6 +489,60 @@ script:
...
+    - contextPath: DBotScore.Indicator
+      description: The indicator we tested
+    - contextPath: DBotScore.Type
+      description: The type of the indicator
+    - contextPath: DBotScore.Vendor
+      description: Vendor used to calculate the score
+    - contextPath: DBotScore.Score
+      description: The actual score
+    important:
+    - contextPath: File(val.Malicious)
```
aviadl: Does URL populate File and IP ?
I think that you mixed here some stuff

gshriki: No it will not. When running with a URL, pt-malware only adds relevant context entries. 
This is how pt-malware is designed, so I just duplicated it

gshriki:
fixed

PR #511, opened 2017-05-28
Fix tableToMD when there is a single header

aviadl:
Fixed for both python and js
fixes https://github.com/demisto/server/issues/7396


dorsha:
I would also add what anar encountered, so it will support array of strings.. but can be another PR

PR #514, opened 2017-06-01
Update integration-Vectra.yml

shaniacht1:


Integrations/integration-Vectra.yml
```diff
@@ -158,7 +158,7 @@ script:
             }
         }
 
-        md = tblToMd('Detection table', res, keys);
+        md = tblToMd('Detection table', res, ['id','type_vname','category','src_ip','t_score','c_score','first_timestamp','tags']);
```
Spivaka: what is this fixing? why are the keys empty/missing?

PR #515, opened 2017-06-01
Get file name for upload

Spivaka:
fixes https://github.com/demisto/server/issues/7454


Spivaka:
Please also validate that this indeed solves the issue

PR #517, opened 2017-06-01
Added expiry parameter for RSA NetWitness Packets and Logs

dorsha:
fixes: https://github.com/demisto/server/issues/7464

### DIFF
--- a/Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
+++ b/Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
@@ -40,6 +40,11 @@ configuration:
   defaultvalue: "false"
   type: 8
   required: false
+- display: Expiration time
+  name: expiry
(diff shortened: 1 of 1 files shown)


Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -219,6 +224,10 @@ script:
         var responseType = resType ? ResponseTypes[resType] : 'application/json';
         var fullUrl = url + path + parametersUrl + '&force-content-type=' + responseType;
 
+        if (params.expiry) {
```
anara123: Do we want expire as param? If yes then all the commands in this instance will have the same expire.

dorsha: Yes, all requests accept it, I think it is fine

PR #522, opened 2017-06-03
fixed human readable markdown

anara123:
fixes https://github.com/demisto/server/issues/7463

### DIFF
--- a/Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
+++ b/Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
@@ -2,7 +2,6 @@ commonfields:
   id: RSA NetWitness Packets and Logs
   version: -1
 name: RSA NetWitness Packets and Logs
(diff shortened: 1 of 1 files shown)


Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -2,7 +2,6 @@ commonfields:
   id: RSA NetWitness Packets and Logs
   version: -1
 name: RSA NetWitness Packets and Logs
-system: true
```
dorsha: I think you should keep it true

anara123: Actually no. The implementation already supports it. It makes things much simpler. Playbook, script, BYOI is system only if it comes from content. All the rest are not system. It just we need to update all the scripts/integrations/playbooks in the content repo

PR #523, opened 2017-06-04
Assign analyst to incident utility

dorsha:
If isRandom is 'yes' (default) the analyst is picked randomly from the available users, according to the provided roles (if no roles provided, will fetch all users).

If isRandom is 'no' the less busy analyst will be picked to be the incident owner.

Please run it as well and do some tests (I think I covered everything, but need to verify)

### DIFF
--- a/Scripts/script-AssignAnalystToIncident.yml
+++ b/Scripts/script-AssignAnalystToIncident.yml
@@ -57,6 +57,6 @@ args:
   - machine-learning
   - top-user
   - less-busy-user
-  description: (default: random) You can pick how to assign the owner - by random,
-    machine learning, top owner or the less busy user.
+  description: '(default: random) You can pick how to assign the owner - by random,
+    machine learning, top owner or the less busy user.'
 scripttarget: 0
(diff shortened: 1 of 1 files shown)


Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -0,0 +1,55 @@
...
+
+  res = demisto.executeCommand('getUsers', { 'roles': ','.join(rolesList) });
+  usersList = list(map((lambda u: u['username']), res[0]['Contents']))
+
+  userToAssign = None
+  if isRandom:
+      userToAssign = choice(usersList)
+  else:
+      usersMap = dict.fromkeys(usersList, 0)
+      assignedIncidents = demisto.executeCommand("getIncidents", { 'status': '1' })[0]['Contents']['data']
```
aviadl: 1. you will get at most 100 incidents with this implementation.
2. might be that all returned incidents will not be assigned.

dorsha:
@slavikm we will change it to use the machine learning (we gonna expose it as service module).
Don't merge it.

dorsha:
Changed to javascript and use machine learning, ready for merge (along with the server and client).
Please review.

image

PR #525, opened 2017-06-04
Fixes in representation of Netwitness results

maiorrulz:


maiorrulz:
Anar please run a couple of queries of netwitness-sdk-search/query  to ensure that it returns in right way

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -98,9 +99,15 @@ script:
...
+        groupObject = groupObject.filter(function (element) {
+            return element != undefined || element != null;
+        });
 
-        if(retVal.md.length > 0) {
-            var hr = tblToMd(command, retVal.md,'', '<br>');
+        if(groupObject.length > 0) {
+            retVal.md = groupObject;
+            retVal.context.NetWitness.Events = groupObject;
+            var hr = tblToMd(command, retVal.md);
```
anara123: use `tableToMarkdown` instead of `tblToMd`

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -63,32 +63,33 @@ script:
...
             });
         }
-        return result;
     }
 
     var buildHrAndContext = function (data) {
         //Check if we have values
         var retVal = {md:[], context: {NetWitness:{Events:[]}}};
-        var typeToValue;
+        var groupObject = [];
```
anara123: you call it `groupObject` but it is an array. Call it something more simple `queryResults` or something similar.

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -63,32 +63,33 @@ script:
...
-        var TYPE = 'type';
+    var fillgroupObject = function (fields, groupObject) {
+        var TYPE  = 'type';
         var VALUE = 'value';
+        var GROUP = 'group';
         if(fields && Array.isArray(fields)) {
             fields.forEach(function (element) {
-                if (element[TYPE] && element[VALUE]) {
-                    result[element[TYPE]] = element[VALUE];
+                if (element[TYPE] && element[VALUE] && element[GROUP]) {
```
anara123: In case that there is no group key you will not add the element. But I think it will be better if the code will handle those cases if group not exists

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
maiorrulz: I prefer tbToMd(style issue)

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
maiorrulz: I thought about it, but there are too many cases we have to cover(If first have a group and second don't have a group, if there are elements with same group). we have to decide if we trust group field or covering all use cases

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
maiorrulz: fixed

Integrations/integration-RsaNetWitnessPacketsAndLogs.yml
```diff
@@ -98,9 +99,15 @@ script:
...
+        //May be empty elements which should be filtred
+        queryResults = queryResults.filter(function (element) {
+            return element != undefined || element != null;
+        });
 
-        if(retVal.md.length > 0) {
-            var hr = tblToMd(command, retVal.md,'', '<br>');
+        if(queryResults.length > 0) {
+            retVal.md = queryResults;
+            retVal.context.NetWitness.Events = queryResults;
```
anara123: `groupObject` will not always be an event, you use `buildHrAndContext` also to parse `netwitness-search` results too, and those are not events. Events should be only those who return from `netwitness-query`

maiorrulz: ,Search can be any entity If you know how it fits better, then you should write design. And I think this what was decided by me and Lior or Jonathan

PR #531, opened 2017-06-05
Fixed the ID and name of the newwitness scripts.

dorsha:
Must change the name as well.


dorsha:
yes but take a look at this one:
image

You did not change the ID before and this one of the script that cause us problems... so it is probably related to the name.

PR #538, opened 2017-06-07
CommonServer - add underscore to camel case converter

shaniacht1:


Scripts/script-CommonServer.yml
```diff
@@ -203,6 +203,18 @@ script: |-
...
   }
 
+  var tblToMd = tableToMarkdown;
+
+  /**
+   * Converts underscore case strings to camel case
+   * @param {String} string - string to be converted - i.e. hello_world
+   * @returns {String} - converted string - i.e. HelloWorld
+  */
+  var undrscrToCamelCase = function(string){
```
dorsha: There is a typo in the function name, should be `underscoreToCamelCase`

Scripts/script-CommonServer.yml
```diff
@@ -203,6 +203,18 @@ script: |-
...
+      string = '_'+string;
```
dorsha: In JavaScript (and other languages), it is not a good practice to mutate the function input, so please change the assignment to another variable.

Scripts/script-CommonServer.yml
shaniacht1: done

Scripts/script-CommonServer.yml
shaniacht1: done

PR #539, opened 2017-06-07
esm v10

asafshen:
In progress

### DIFF
--- a/Integrations/integration-ESM-v10.yml
+++ b/Integrations/integration-ESM-v10.yml
@@ -43,13 +43,14 @@ configuration:
   type: 8
   required: false
 script:
-  script: |+
+  script: |
     import base64
     import json
     import re
     import sys
     import requests
     import time
+    import datetime
 
     verify =  demisto.params()['insecure'] == False
 
@@ -76,6 +77,14 @@ script:
             md = md + ' | '.join(row['values']) + '\n'
 
         return md
+
+    def severity_to_level(severity):
+        if severity > 65:
+            return 3
+        elif severity > 32:
+            return 2
+        else:
+            return 1
 
     class NitroESM(object):
 
@@ -206,20 +215,6 @@ script:
             #fetch result
             res = self.fetch_results(resultID)
 
-            '''
-            # result mock
-            res = {
-                "columns": [
-                    {"name": "Alert.ID"},
-                    {"name": "Alert.LastTime"}
-                ],
-                "rows": [
-                    {"values": [ "32886", "03\/01\/2016 13:01:54"]},
-                    {"values": [ "32906","03\/01\/2016 13:01:54"]}
-                ]
-            }
-            '''
-
             table = search_result_to_md(res)
             return {
                 'ContentsFormat': formats['json'],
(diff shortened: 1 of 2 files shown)


Integrations/integration-PostgreSQL.yml
```diff
@@ -62,7 +62,7 @@ script:
         conn = pgsqlConnect()
         try:
             cur = conn.cursor()
-            cur.execute(demisto.args()['query'])
+            cur.execute(demisto.args()['query'])Å
```
aviadl: @asafshen you have here a typo, please fix it

Integrations/integration-ESM-v10.yml
```diff
@@ -0,0 +1,384 @@
...
+
+            cmd = 'alarmGetTriggeredAlarmsPaged?pageSize=500&pageNumber=1&triggeredTimeRange=' + time_range
+            user = self.get_user()
+
+            query = json.dumps({ 'assignedUser' : user})
+
+            res = self.cmdquery(cmd, query)
+
+            '''
+            # mock of the result
```
meirwah: emmm.. not sure you want to leave this here...

Integrations/integration-ESM-v10.yml
```diff
@@ -0,0 +1,384 @@
...
+        sys.exit(0)
+    if demisto.command() == 'esm-fetch-fields':
+        res = esm.featchAllFields()
+        demisto.results(res)
+        sys.exit(0)
+    if demisto.command() == 'esm-search':
+        args = demisto.args()
+        res = esm.search(
+            demisto.get(args, 'timeRange'),
+            json.loads(demisto.get(args, 'filters')),
```
meirwah: if filters arg is empty, you get empty json here?

asafshen: we get an error if filter is empty

Spivaka:
Remove go module in the server as well

meirwah:
Wait, the Go impl supports version 9.X, while this integration supports version 10.x only.
So don't remove until you can support both versions in BYOI (or keep things as are)

asafshen:
@meirwah @Spivaka 
Please review 
also updated the description of nitro integration (support up to 9.X)

Integrations/integration-ESM-v10.yml
```diff
@@ -0,0 +1,460 @@
...
+description: Run queries and receive alarms from Intel Security ESM
```
meirwah: add here, for ESM version 10.x ONLY

Integrations/integration-ESM-v10.yml
```diff
@@ -0,0 +1,460 @@
...
+        sys.exit(0)
+    if demisto.command() == 'fetch-incidents':
+
+        last_run = demisto.getLastRun()
+        start = last_run and last_run['value']
+
+        if (not start):
+            start = (datetime.datetime.now() - datetime.timedelta(minutes=1)).isoformat()
+
+        end = datetime.datetime.now().isoformat()
```
meirwah: if ESM if in a different time zone then the demisto server, you will have an issue here.
you should add a config param to take time -/+ the time zone delta

asafshen: @meirwah will check that against our instance
I noticed that in go implementation we use zulu time..

why shouldn't we do this here also?

PR #540, opened 2017-06-08
Anomali ThreatStream integration

nitzanshapira:
Fixes https://github.com/demisto/server/issues/4705.
Including @dorsha fixes.


Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -0,0 +1,433 @@
...
+        'threatstream-intelligence': {
+            method: 'GET',
+            url: 'intelligence/',
+            version: 'v2',
+        },
+    }
+
+    var sendRequest = function(method, url, version, args, headers, multipart) {
+        url = base + version + '/' + url;
+        if (method === 'GET') {
```
Spivaka: we use anything else than get?
do we still user multipart?
I think not so the code shouldn't include those cases

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -0,0 +1,433 @@
...
+            var cred = {
+                username: args.username,
+                api_key: args.api_key,
+            };
+            delete(args.username);
+            delete(args.api_key);
+            url += encodeToURLQuery(cred);
+        }
+
+        console.log(url);
```
Spivaka: remove

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -0,0 +1,433 @@
...
+
+        var file_id = args.file_id;
```
Spivaka: Also can be removed

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -0,0 +1,433 @@
...
+
+    var buildEmptyEntryContext = function() {
+        return {Type: entryTypes.note, Contents: null, ContentsFormat: formats.json, EntryContext: {}};
+    }
+
+    var isStandardCommand = function(command) {
+        return (command == 'ip' || command == 'file' || command == 'domain');
+    }
+
+    var convertKeysToPascalCase = function(dict) {
```
Spivaka: Shouldn't be a part of common?

Spivaka:
fixes https://github.com/demisto/server/issues/4705

Integrations/integration-AnomaliThreatStream.yml
nitzanshapira: At the moment we don't (I also removed the "import" functions which you implemented), but we may do so in the future. @NeelRust asked for the "upload file" feature (currently not implemented) which will require the "post" and "multipart" features.
I can remove it, but it works now and will probably be used in the future. Whatever you think.

PR #543, opened 2017-06-08
Light cyber magna -Added new commands

dem-evgeny:
Added new commands from the new API:
lcm-host-installed-executables
lcm-host-loaded-modules
lcm-host-opened-ports
lcm-host-processes
lcm-host-processes-internet-connections
lcm-host-suspicious-artifacts


Spivaka:
shouldn't we upgrade this while we are at it?
also have this issue:
https://github.com/demisto/server/issues/5752

gshriki:
Agreed.
@dem-evgeny - please work with @JonathanMeler and @liorkol to define the right war room entries and context for all commands and upgrade the integration.

PR #547, opened 2017-06-08
Improved the assign analyst to incident script

dorsha:
1. Fixed description
2. Added the ability to specify username to assign

### DIFF
--- a/Scripts/script-AssignAnalystToIncident.yml
+++ b/Scripts/script-AssignAnalystToIncident.yml
@@ -4,28 +4,30 @@ commonfields:
 name: AssignAnalystToIncident
 system: true
 script: |
-  var userToAssign = null;
+  var userToAssign = args.username;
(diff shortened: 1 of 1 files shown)


Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -4,28 +4,30 @@ commonfields:
 name: AssignAnalystToIncident
 system: true
 script: |
-  var userToAssign = null;
+  var userToAssign = args.username;
```
aviadl: Dont  you want to verify that the user exists ?

dorsha: So if the setOwner does not find the user it returns an error, so I just use the isError to verify that and provides an appropriate message:

image

dorsha: Check the last commit

PR #550, opened 2017-06-11
Symantec Endpoint Protection test playbook

nshap:
Fixes https://github.com/demisto/server/issues/7576
Previously a POC, now a proper playbook.


Scripts/script-Sleep.yml
```diff
@@ -1,11 +1,9 @@
 commonfields:
   id: Sleep
-  version: -1
+  version: 7
```
Spivaka: should be minus 1

Scripts/script-Sleep.yml
```diff
@@ -1,11 +1,9 @@
...
 name: Sleep
-script: |
-  import time
-  time.sleep(int(demisto.args()['seconds']));
-type: python
+script: wait(parseInt(args.seconds));
```
Spivaka: great!

Scripts/script-Sleep.yml
```diff
@@ -16,5 +14,4 @@ args:
   default: true
   description: Amount of seconds to sleep
 scripttarget: 0
-dependson: {}
-timeout: 6000s
+timeout: 1h40m0s
```
Spivaka: why?

nshap: It was 6000s in the Python version (I don't know why - I think it's arbitrary), which is 1 hour and 40 minutes.

PR #551, opened 2017-06-12
add description to ArcSight ESM XML

meirwah:
add description to ArcSight ESM XML


meirwah:
@nshap plz review, and modify description according...

PR #553, opened 2017-06-13
ParseCSV - change outputs

shaniacht1:
fixes https://github.com/demisto/server/issues/6047

### DIFF
--- a/Scripts/script-ParseCSV.yml
+++ b/Scripts/script-ParseCSV.yml
@@ -113,19 +113,27 @@ script: |
           old_iplist = list(demisto.get( demisto.context(), 'ips')) if demisto.get( demisto.context(), 'ips' ) else []
           iplist = list(set(iplist) - set(old_iplist))
           if len(iplist) > 0:
-              context["IP"] = iplist
+              context["IP"] = []
+              for ip in iplist:
+                  context["IP"].append({"Address" : ip})
 
       if domainlist:
           old_domainlist = list(demisto.get( demisto.context(), 'domains')) if demisto.get( demisto.context(), 'domains' ) else []
           domainlist = list(set(domainlist) - set(old_domainlist))
           if len(domainlist) > 0:
-              context["Domain"] = domainlist
+              context["Domain"] = []
+              for domain in domainlist:
+                  context["Domain"].append({"Name": domain})
 
       if hashlist:
           old_hashlist = list(demisto.get( demisto.context(), 'hashes')) if demisto.get( demisto.context(), 'hashes' ) else []
           hashlist = list(set(hashlist) - set(old_hashlist))
           if len(hashlist) > 0:
-              context["Hash"] = hashlist
+              context["File"] = []
+              for hash in hashlist:
+                  if len(hash)==32: context["File"].append({"MD5" : hash})
+                  if len(hash)==64: context["File"].append({"SHA256" : hash})
+                  if len(hash)==40: context["File"].append({"SHA1" : hash})
 
       demisto.results( { "Type" : entryTypes["note"], "ContentsFormat" : formats["text"], "Contents" : content ,"HumanReadable" : md, "EntryContext": context })
   else:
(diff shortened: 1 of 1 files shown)


shaniacht1:
Do not review yet

shaniacht1:
List of changes:
* added outputs
* changed the way the context is set
* changed contextEntry names (i.e. from 'ips' to 'IP' 
* changed output to war room to show table #6047

shaniacht1:
Do not merge yet

Scripts/script-ParseCSV.yml
```diff
@@ -81,50 +81,53 @@ script: |
           if has_header:
               next(csvdata)
 
+          md = '### Parsed Data Table\n'+('IPs |' if 'ips' in dArgs else '') + ('Domains |' if 'domains' in dArgs else '') + ('Hashes |' if 'hashes' in dArgs else '') + '\n'
+          md += ('- |' if 'ips' in dArgs else '') + ('- |' if 'domains' in dArgs else '') + ('- |' if 'hashes' in dArgs else '') + '\n'
```
Spivaka: would set the condition above and reuse it instead of calculating it more time

Scripts/script-ParseCSV.yml
```diff
@@ -81,50 +81,53 @@ script: |
...
+                  md += (row[parsedomain] + '|' if row[parsedomain] else ' |')
                   has_dot = '.' in row[parsedomain]
                   no_spaces = ' ' not in row[parsedomain]
-
-                  if is_unique and has_dot and no_spaces:
+                  if has_dot and no_spaces:
                       domainlist.append(row[parsedomain])
 
               if parsehash is not None and parsehash != '':
-                  is_unique = row[parsehash] not in hashlist
```
meirwah: so no need for is_unique anymore?

Scripts/script-ParseCSV.yml
```diff
@@ -149,6 +153,12 @@ args:
     0)
 - name: parseAll
   description: parses and converts all the rows in csv into json and puts into context.
+outputs:
+- contextPath: IP
+  description: IP address found in parsed file
+- contextPath: Domain
+  description: Domain found in parsed file
+- contextPath: Hash
```
meirwah: is this the right output?
shouldn't it be File.MD5 ?

Scripts/script-ParseCSV.yml
shaniacht1: no because dq does it anyway when adding to context

Scripts/script-ParseCSV.yml
```diff
@@ -81,50 +81,61 @@ script: |
...
-          hashcount = len(hashlist)
-          appendContext('hashes', hashlist)
-
-      if iplist or domainlist or hashlist:
-          demisto.results('Parsed CSV: %s, %d IPs, %d Domains, %d Hashes added to the current context' % ( file, ipcount, domaincount, hashcount ))
+          if len(hashlist) > 0:
+              context["File"] = []
+              for hash in hashlist:
+                  if len(hash)==32: context["File"].append({"MD5" : hash})
+                  if len(hash)==64: context["File"].append({"SHA256" : hash})
```
Spivaka: I don't think this is unique like this
should have val.MD5 == obj.MD5

Scripts/script-ParseCSV.yml
```diff
@@ -81,50 +81,61 @@ script: |
...
       if domainlist:
-          old_domainlist = list(demisto.get( demisto.context(), 'domains' )) if demisto.get( demisto.context(), 'domains' ) else []
+          old_domainlist = list(demisto.get( demisto.context(), 'domains')) if demisto.get( demisto.context(), 'domains' ) else []
           domainlist = list(set(domainlist) - set(old_domainlist))
-          domaincount = len(domainlist)
-          appendContext('domains', domainlist)
+          if len(domainlist) > 0:
+              context["Domain"] = []
+              for domain in domainlist:
+                  context["Domain"].append({"Name": domain})
```
Spivaka: same here for unique

PR #555, opened 2017-06-14
Added the correct report decoder for the openDuration field

dorsha:
fixes: https://github.com/demisto/server/issues/7628


dorsha:
@Bars92 / @aviadl - please review

PR #559, opened 2017-06-14
Create script-LoadJSON.yml

meirwah:
load a json string to object script


meirwah:
This uses default python lib, no need to run it on docker...

PR #566, opened 2017-06-16
LR fetch incidents return on empty value

aviadl:
In case there were no incidents found on the return a json of an empty array.
so there will be no errors in the server


dorsha:
Should merge also did 2.6.0?

PR #567, opened 2017-06-16
Added custom filters in get-instance-info

gshriki:
Added the ability to add custom filter values in the get-instance-info command


Integrations/integration-aws.yml
```diff
@@ -214,6 +213,17 @@ script:
             urlArgs[filter+'.Value'] = keyName;
             filterid = filterid + 1;
         }
+        if ((customFilters) && (customFilters.length > 0)) {
```
dorsha: Syntax wise, I would change it to be like this:

```
if (customFilters && customFilters.length > 0) {
        customFilters.split(',').forEach(function (customFilter) {
            filter = 'Filter.' + filterid.toString();
            urlArgs[filter+'.Name'] = customFilter.split("=")[0];
            urlArgs[filter+'.Value'] = customFilter.split("=")[1];
            filterid = filterid + 1;
        });
    }
```

gshriki: fixed

gshriki: fixed

Integrations/integration-aws.yml
```diff
@@ -413,3 +424,4 @@ script:
     - name: headers
       description: Headers to display in Human readable format
     description: Get EC2 EBS Volume Information
+hidden: false
```
meirwah: not needed i think

Integrations/integration-aws.yml
```diff
@@ -4,7 +4,6 @@ commonfields:
 name: Amazon Web Services
 display: Amazon Web Services
 category: IT Services
-system: true
```
meirwah: you remove this on propuse  ?

dorsha: It is not needed anymore

gshriki: the system removed it when I downloaded the script. It's not needed anymore...  (anything that comes from us is automatically tagged as system as far as I know)

PR #573, opened 2017-06-19
update threatconnect to support proxy

meirwah:
update threatconnect to support proxy


meirwah:
haven't been tested yet(no instance)

Spivaka:
:+1: 
merge after testing

PR #578, opened 2017-06-20
Moloch integration

nshap:
Fixes https://github.com/demisto/server/issues/6629.

* Adding human readable and entry context for the moloch_sessions_json command. Entry context is entered under the Moloch.Sessions object.
* Adding numToIp which converts a unit32 to an IP address string.

If you want to test it you can use the public instance (https://demo.molo.ch) and run !moloch_sessions_json expression="ip.src=1.1.1.1".


Integrations/integration-Moloch.yml
```diff
@@ -1,8 +1,8 @@
 commonfields:
-  id: Moloch
+  id: Moloch2
```
Spivaka: should be moloch

Integrations/integration-Moloch.yml
```diff
@@ -30,6 +30,7 @@ script:
     var insecure = params.insecure;
 
     var sendRequest = function(url, queryName) {
+        console.log('sendRequest: ' + url);
```
Spivaka: this is probably for debugging

Integrations/integration-Moloch.yml
```diff
@@ -58,29 +59,94 @@ script:
...
+        for (var i = 0, objects = response.data; i < objects.length; i++) {
+            var output = {};
+
+            output['ID'] = objects[i].id;
+            output['Index'] = objects[i].index;
+
+            if (objects[i].pr in prToName) {
+                output['Protocol'] = prToName[objects[i].pr];
+            } else {
+                output['Protocol'] = objects[i].pr;
```
Spivaka: would do the else codr to begin with and then the if
a = b
if c
a = d

Integrations/integration-Moloch.yml
```diff
@@ -58,29 +59,94 @@ script:
...
+            output['Source Bytes'] = objects[i].by1;
+            output['Source Databytes'] = objects[i].db1;
+            output['Destination IP'] = numToIp(objects[i].a2);
+            output['Destination Port'] = objects[i].p2;
+            output['Destination Packets'] = objects[i].pa2;
+            output['Destination Bytes'] = objects[i].by2;
+            output['Destination Databytes'] = objects[i].db2;
+
+            outputs.push(output);
+            contexts.push(convertKeysToPascalCase(output));
```
Spivaka: does it handle spaces gracefully?

Integrations/integration-Moloch.yml
```diff
@@ -58,29 +59,94 @@ script:
...
             if (sendRequest(urlDict['moloch_connections_json'], 'Test')) {
                 return 'ok';
             }
+
             return 'not cool';
+
         default:
             var currentTime = new Date();
+
+            if (!args.date) {
```
Spivaka: I think default values are handled at moloch side

nshap: They are not, that's why I added this code.
When searching for values in the UI, the URL by default contains "date=-1".
If not defining this value, then no results will be returned for any request (I tried it myself).

Integrations/integration-Moloch.yml
nshap: Yep. Used it other integrations as well. The code is in CommonServer, you can take a look.

PR #582, opened 2017-06-21
Fix some utilities

anara123:


anara123:
@Spivaka

Scripts/script-AreValuesEqual.yml
```diff
@@ -3,10 +3,17 @@ commonfields:
...
-  answer = 'yes' if demisto.args()['left'] == demisto.args()['right'] else 'no'
-  demisto.setContext('AreValuesEqual', answer)
-  demisto.results(answer)
-type: python
+  var left = (typeof args.left === 'string') ? args.left : args.left + '';
+  var right = (typeof args.right === 'string') ? args.right : args.right + '';
+  var answer = 'no';
+  if (left === right) {
+    answer = 'yes';
+  } else {
```
Spivaka: no need for else here

Scripts/script-Set.yml
```diff
@@ -5,7 +5,20 @@ name: Set
...
+  var result = {
+      Type: entryTypes.note,
+      Contents: ec,
+      ContentsFormat: formats.json,
+      HumanReadable: 'Key ' + args.key + ' set'
+  };
+
+  if (args.append) {
+      result.EntryContext = ec;
+  } else if (!args.append || args.append === 'false') {
```
Spivaka: why do you need both conditions?

Spivaka:
@anara123 please fix and merge

Scripts/script-Set.yml
anara123: In case append was not passed or false string was passed. 
But I need to switch the order between them, in case someone pass `append=false`

PR #586, opened 2017-06-22
Create script-Append.yml

meirwah:
add append script

### DIFF
--- a/Scripts/script-Append.yml
+++ b/Scripts/script-Append.yml
@@ -1,0 +1,33 @@
+commonfields:
+  id: Append
+  version: -1
+name: Append
+script: |
+  result = demisto.args()["prefix"] + demisto.args()["suffix"]
+  res = {
+      'ContentsFormat': formats['text'],
+      'Type': entryTypes['note'],
+      'Contents': result,
+      'ReadableContentsFormat': formats['text'],
+      'HumanReadable': result,
+      'EntryContext': {demisto.args()['key']: result}
+  }
+  demisto.results(res)
+type: python
+tags:
+- Utility
+comment: 'will append prefix string to suffix string, and return to context under
+  key given'
+enabled: true
+args:
+- name: prefix
+  required: true
+  default: true
+  description: left side string
+- name: suffix
+  required: true
+  description: right side string
+- name: key
+  required: true
+  description: key to place result in context
+scripttarget: 0


Scripts/script-Append.yml
```diff
@@ -0,0 +1,33 @@
...
+  res = {
+      'ContentsFormat': formats['text'],
+      'Type': entryTypes['note'],
+      'Contents': result,
+      'ReadableContentsFormat': formats['text'],
+      'HumanReadable': result,
+      'EntryContext': {demisto.args()['key']: result}
+  }
+  demisto.results(res)
+type: python
```
dorsha: Because is performance I would do this in JavaScript

Scripts/script-Append.yml
```diff
@@ -0,0 +1,33 @@
...
+name: Append
+script: |
+  result = demisto.args()["prefix"] + demisto.args()["suffix"]
+  res = {
+      'ContentsFormat': formats['text'],
+      'Type': entryTypes['note'],
+      'Contents': result,
+      'ReadableContentsFormat': formats['text'],
+      'HumanReadable': result,
+      'EntryContext': {demisto.args()['key']: result}
```
dorsha: Should be with output and not context.

Scripts/script-Append.yml
```diff
@@ -0,0 +1,33 @@
...
+  version: -1
+name: Append
+script: |
+  result = demisto.args()["prefix"] + demisto.args()["suffix"]
+  res = {
+      'ContentsFormat': formats['text'],
+      'Type': entryTypes['note'],
+      'Contents': result,
+      'ReadableContentsFormat': formats['text'],
+      'HumanReadable': result,
```
dorsha: I would not add entry in the war room only for append strings, too noisy and not relevant for the investigation.
Should have output only

meirwah: you have any example for python script with output?

Scripts/script-Append.yml
meirwah: you have any example for python script with output?

Scripts/script-Append.yml
meirwah: disagree on that, I think readability is more important here

dorsha: If it is that important to add garbage to the war room, please provide better human readable  message to the war room, no one will understand that.

dorsha: And what how war room readability related to performance?

Scripts/script-Append.yml
dorsha: The same as in JavaScript

Scripts/script-Append.yml
meirwah: code readability

Scripts/script-Append.yml
dorsha: script-UnzipFile.yml and there are many other python examples in the content repo

PR #588, opened 2017-06-22
Minor issues found in cloud lock

Spivaka:
fixes https://github.com/demisto/server/issues/7527

### DIFF
--- a/Integrations/integration-CloudLocki.yml
+++ b/Integrations/integration-CloudLocki.yml
@@ -290,3 +290,4 @@ script:
       description: Activity vendor name
     - contextPath: CloudLock.Activities.Vendor.Service
       description: Activity vendor service
+hidden: false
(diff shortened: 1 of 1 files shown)


Integrations/integration-CloudLocki.yml
```diff
@@ -40,6 +40,20 @@ script:
         'cloudlock-get-users': {
             url: 'users',
             method: 'GET',
+            extended: true,
```
asafshen: I think you need to add `innerPath` to this command

Integrations/integration-CloudLocki.yml
```diff
@@ -50,7 +64,18 @@ script:
         },
         'cloudlock-get-user-apps': {
             url: 'users/%email%/apps',
-            method: 'GET'
+            method: 'GET',
+            extended: true,
```
asafshen: I think you need to add `innerPath` to this command

Spivaka:
not exactly, but I'll fix it in another way (no dq if no inner path, take response as is)

PR #590, opened 2017-06-23
Allow users to set multiple keys/values in one go

slavikm:


Scripts/script-SetMultipleValues.yml
```diff
@@ -0,0 +1,26 @@
+commonfields:
+  id: SetMultipleValues
+  version: -1
+name: SetMultipleValues
+script: |
+  keys = [k.strip() for k in demisto.args()['keys'].split(',')]
```
Spivaka: would add a check that the amount keys matchs the amount values

slavikm: I think this is the expected behavior:
```
slavik@madjet-2:[dcli](master): python
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> zip([1,2,3,4], [5,6,7])
[(1, 5), (2, 6), (3, 7)]
```
You want to throw an error? Why?

PR #594, opened 2017-06-25
added email rasterize. minor fixes in url rasterize

gshriki:
added rasterize-email body
handling URLs without http prefix

### DIFF
--- a/Integrations/integration-rasterize.yml
+++ b/Integrations/integration-rasterize.yml
@@ -11,14 +11,14 @@ configuration: []
 script:
   script: |-
     import os, subprocess,sys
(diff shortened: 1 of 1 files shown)


Integrations/integration-rasterize.yml
```diff
@@ -16,15 +15,50 @@ script:
         # This is the call made when pressing the integration test button.
         demisto.results('ok')
         sys.exit(0)
+    if demisto.command() == 'rasterize-email':
+        import os, subprocess,sys
```
slavikm: Just move the import to the top as both branches use it...

PR #595, opened 2017-06-25
Use correct path on extract script

aviadl:
This will prevent duplications if script runs twice


aviadl:
Probably, but since i didnt see we do with it anything i didnt bother to (and you also didnt have a shortcut for it :) )

PR #612, opened 2017-07-02
Added to use templates when creating new incident

dem-evgeny:
Added to use templates when creating new incident


Integrations/integration-ServiceNow.yml
```diff
@@ -3,7 +3,6 @@ commonfields:
   version: -1
 name: ServiceNow
 display: ServiceNow
-fromversion: 2.5.0
```
Spivaka: I don't think you intended to remove this.
Didn't we have SN prior to 2.5?
if so, we should add another integration with to version 2.5, before those changes

Integrations/integration-ServiceNow.yml
```diff
@@ -96,11 +95,17 @@ script:
         }
     };
 
-    var getBody = function() {
+    var getBody = function(templateName) {
         var body = {};
+        if (templateName !== undefined) var tmpRes = SNGetTemplate(templateName);
```
Spivaka: don't use such if ... declaration statements, this is confusing as coders will search declarations in the left side
write it in the following manner
var tmpRes;
if (...) {
tmpRes = ...
}

Integrations/integration-ServiceNow.yml
dem-evgeny: fixed

Integrations/integration-ServiceNow.yml
dem-evgeny: fixed

Integrations/integration-ServiceNow.yml
```diff
@@ -96,11 +96,22 @@ script:
         }
     };
 
-    var getBody = function() {
+    var getBody = function(templateName) {
         var body = {};
+        var tmpRes;
+
+        if (templateName){
```
Spivaka: checking a !== undefined can be replaced with a, cause if a has a value if (a) return true, else false (also when a is false or empty string)

PR #613, opened 2017-07-02
multi search in ews

Spivaka:


dorsha:
@aviadl - I think we can add tests in the new test framework to what we did.. should be easy.

PR #616, opened 2017-07-03
Check that the content type is correct

Spivaka:


Integrations/integration-Cylance_Protect.yml
```diff
@@ -78,6 +78,11 @@ script:
         if (res.StatusCode < 200 || res.StatusCode >= 300) {
             throw 'Request to Cylance protect Failed.\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res.Body) + '.';
         }
+        // If we have a body, it should have application/json content type.
+        if (res.Body &&
+            (!res.Headers['Content-Type'] || !res.Headers['Content-Type'][0] || res.Headers['Content-Type'][0].indexOf('application/json') === -1)) {
+            throw 'Invalid content type, expected application/json. Check your url!';
```
meirwah: maybe return here the body as string?

Spivaka: we wanna throw here... but I'll add the body to the error message

PR #632, opened 2017-07-07
Add packet mail integration for checking IP reputation

aviadl:


aviadl:
There is no icon, so going out there with the default

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
+commonfields:
+  id: PacketMail
+  version: -1
+name: PacketMail
+display: PacketMail.net
+category: Authentication
```
dorsha: Authentication?

aviadl: fixed

gshriki:
Default icon is lame.
Why not something like this?
image

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
...
+- display: API Key
+  name: apikey
+  defaultvalue: ""
+  type: 4
+  required: false
+- display: Use Proxy
+  name: proxy
+  defaultvalue: "true"
+  type: 8
+  required: false
```
Spivaka: Add insecure as a param and use it in the http call.
I would also add the server address and use the initial one below as default

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
...
+            },
+            true,
+            params.proxy
+        );
+
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            throw 'Request Failed.\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res) + '.';
+        }
+
+        return JSON.parse(res.Body);
```
Spivaka: add response content type check for json. some cases server return 200 with error in the html body.

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
...
+
+    // The command input arg holds the command sent from the user.
+    switch (command) {
+        // This is the call made when pressing the integration test button.
+        case 'test-module':
+            res = sendRequest(initialurl + '8.8.8.8' + '?apikey='+params.apikey);
+            if (res.query_result === 'Success') {
+                return 'ok';
+            } else {
+                return res;
```
Spivaka: way else? just reurn in this case

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
...
+            }
+            ec = {};
+            if (maliciousLists. length > 0) {
+                addMalicious(ec, outputPaths.ip,{
+                        Address: args.ip,
+                        Malicious: {Vendor: 'PacketMail', Description: 'Found in malicious lists: ' + maliciousLists}});
+            }
+            return {Contents: res, HumanReadable: md, ContentsFormat: formats.json, Type: entryTypes.note, EntryContext: ec};
+        default:
+            // You can use args[argName] or args.argName to get a specific arg. args are strings.
```
Spivaka: remove comments

Integrations/integration-PacketMail.yml
```diff
@@ -0,0 +1,118 @@
...
+            md = '';
+            maliciousLists = '';
+            var keys = Object.keys(res);
+            keys = keys.sort();
+            for (i = 0 ; i < keys.length; i++) {
+                if (md.length > 0) {
+                    md += '\n\n';
+                }
+                if (res[keys[i]] instanceof Object) {
+                    if (malLists.indexOf(keys[i]) >= 0) {
```
Spivaka: I don't understand this. You check it in the response and only if in the param mal list you mark it as malicious? if it's planned thats fine but this is a strange behaviour

Integrations/integration-PacketMail.yml
aviadl: The content type returned is text :(
So not going to check for it

aviadl:
@Spivaka fixed all your comments but one, please review again

Integrations/integration-PacketMail.yml
Spivaka: so try-catch-rethrow?

PR #633, opened 2017-07-09
added dimensions and type for email

gshriki:
added dimensions and type arguments to rasterize-email, like we have for rasterize


aviadl:
@slavikm are you reviewing this ?

PR #636, opened 2017-07-11
Add impersonation address as param to EWS

Spivaka:


Integrations/integration-Office_EWS.yml
```diff
@@ -590,7 +597,6 @@ script:
                 }
             }
             args['mailbox-search-scope'] = args['mailbox-search-scope'].map(function(a) { return a.trim(); });
-            args['impersonating-mailbox'] = params.credentials.identifier + '@' + params.domain;
```
dorsha: what is `impersonating-mailbox`? don't we need it anymore?

PR #637, opened 2017-07-11
isError script

gshriki:
A script to check whether an entry is an error

### DIFF
--- a/Scripts/script-isError.yml
+++ b/Scripts/script-isError.yml
@@ -3,23 +3,33 @@ commonfields:
   version: -1
 name: isError
 script: |-
-  res = executeCommand("getEntry", {"id":args.entryId});
-  if (isError(res[0])) {
-      return 'yes';
-  } else {
-      return 'no';
+  entryIds = (Array.isArray(args.entryId)) ? args.entryId : [args.entryId];
+  for(var entryIdx=0; entryIdx < entryIds.length; ++entryIdx) {
+      res = executeCommand("getEntry", {"id":entryIds[entryIdx]});
+      for (var resIdx=0; resIdx < res.length; ++resIdx) {
+          if (isError(res[resIdx])) {
+              return 'yes';
+          }
+      }
   }
+  return 'no';
 type: javascript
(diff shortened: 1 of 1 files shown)


Scripts/script-isError.yml
```diff
@@ -0,0 +1,25 @@
+commonfields:
+  id: isError
+  version: -1
+name: isError
+script: |-
+  res = executeCommand("getEntry", {"id":args.entryId});
```
aviadl: Should make arg support arrays.
And then work here with multiple entryIds

Scripts/script-isError.yml
```diff
@@ -0,0 +1,25 @@
...
+comment: Check whether a given entry is an error. Use ${lastCompletedTaskEntries.[0]}
+  to check the previous task entry.
+enabled: true
+args:
+- name: entryId
+  required: true
+  default: true
+  description: Entry to check
+scripttarget: 0
+releaseNotes: "New script to check whether a given entry is an error"
```
aviadl: No need for releaseNotes on new content (only for content that was not there in previous versions)

Scripts/script-isError.yml
```diff
@@ -0,0 +1,25 @@
...
+      return 'no';
+  }
+type: javascript
+tags:
+- Condition
+- Utility
+comment: Check whether a given entry is an error. Use ${lastCompletedTaskEntries.[0]}
+  to check the previous task entry.
+enabled: true
+args:
```
aviadl: Missing output

Scripts/script-isError.yml
```diff
@@ -0,0 +1,25 @@
...
+  if (isError(res[0])) {
+      return 'yes';
+  } else {
+      return 'no';
+  }
+type: javascript
+tags:
+- Condition
+- Utility
+comment: Check whether a given entry is an error. Use ${lastCompletedTaskEntries.[0]}
```
aviadl: lastCompletedTaskEntries without the `.[0]` (see my previous comment)

PR #638, opened 2017-07-11
Splunk proxy

meirwah:


Integrations/integration-SplunkPy
```diff
@@ -70,12 +80,55 @@ script:
...
     service = client.connect(
             host=demisto.params()['host'],
             port=demisto.params()['port'],
             username=demisto.params()['authentication']['identifier'],
             password=demisto.params()['authentication']['password'])
 
+    proxy = demisto.params()['proxy']
+    if proxy:
+        try:
+            service = client.connect(
```
asafshen: I think it is more readable if instead that `handler` function returns `request` function -
We can call `handler` function (maybe we can call it `create_proxy`)
And use the `request` function as the `handler` param (maybe we can call it `handler` instead)

Integrations/integration-SplunkPy
```diff
@@ -40,12 +42,20 @@ configuration:
   defaultvalue: ""
   type: 13
   required: false
+- display: 'Proxy - in format: localhost:8080'
```
asafshen: usually, the user uses another ip as a proxy..?
if so - it should reflect in the example..

Integrations/integration-SplunkPy
meirwah: Used similar code to the original python SDK example:
https://github.com/splunk/splunk-sdk-python/blob/master/examples/handlers/handler_proxy.py

I prefer to keep it like this, so it can be referenced easily

asafshen: Thanks for explaining :)

PR #644, opened 2017-07-13
Fix phishing playbook

aviadl:
1. Set sender properly into context
2. Use the sender from incident label instead of context, since context can contain many other emails as well


aviadl:
@slavikm looked at the flow, and there is nothing else that is pushing data to account.email.address
So the flow can stay parallel.
Fixed also account enrichment playbook to look at the correct DT path

PR #645, opened 2017-07-16
Light cyber magna

dem-evgeny:
Added the commands:
lcm-host-opened-ports
lcm-host-suspicious-artifacts
lcm-host-processes
lcm-host-loaded-modules
lcm-host-processes-internet-connections
lcm-host-installed-executables

Because we don't have LightCyber env. tested it only at Customers env.


Integrations/integration-LightCyberMagna.yml
```diff
@@ -49,6 +53,7 @@ script:
     var buildArgs = function(args) {
         sArgs = Object.keys(args).map(function(key) {
             return [key, args[key]].map(encodeURIComponent).join("=");
+        log(JSON.stringify(sArgs));
```
aviadl: Use functions to log to log file

Spivaka:
Again, 7 days and out.

PR #651, opened 2017-07-17
increase test polling timeout

asafshen:
increase polling timeout


asafshen:
@dorsha waiting for a green build

PR #652, opened 2017-07-18
esm version 10

asafshen:
support versions newer than ESM 10.0
(they have changed a bit the api return value


asafshen:
@meirwah - We still don't have ESM test-playbook(takes time).

This is a hot-fix for a costumer with ESM-v10

PR #655, opened 2017-07-19
Add canGetSamples: true to integrations that can fetch

Spivaka:


Integrations/integration-Cylance_Protect.yml
```diff
@@ -6,6 +6,7 @@ display: Cylance Protect
 category: End Point
 system: true
 fromversion: 2.0.1
+canGetSamples: true
```
asafshen: use all lowercase instead (`cangetsamples`)

PR #662, opened 2017-07-22
rename commands

asafshen:
related to https://github.com/demisto/server/issues/5526#issuecomment-317138186


Spivaka:
we should understand if anyone is using the names that you changed, if so we need to add a third option and deprecate the old one

Spivaka:
I think we can and should live without this.

PR #663, opened 2017-07-22
Match regex including group support

slavikm:
Simple utility to match regex in text (or object) including extraction of groups

### DIFF
--- a/Scripts/script-MatchRegex.yml
+++ b/Scripts/script-MatchRegex.yml
@@ -1,0 +1,45 @@
+versionedfields: {}
+commonfields:
+  id: MatchRegex
+  version: -1
+name: MatchRegex
+script: |-
+  var data = (typeof args.data) === 'string' ? args.data : JSON.stringify(args.data);
+  var flags = args.flags ? args.flags : 'im';
+  var r = new RegExp(args.regex, flags);
+  var m = r.exec(data);
+  if (m && m.length > 0) {
+      var val = m[0];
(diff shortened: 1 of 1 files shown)


Scripts/script-MatchRegex.yml
```diff
@@ -0,0 +1,45 @@
...
+script: |-
+  var data = (typeof args.data) === 'string' ? args.data : JSON.stringify(args.data);
+  var flags = args.flags ? args.flags : 'im';
+  var r = new RegExp(args.regex, flags);
+  var m = r.exec(data);
+  if (m && m.length > 0) {
+      var val = m[0];
+      var group = parseInt(args.group);
+      if (args.group && m.length > group) {
+          val = m[group];
```
meirwah: maybe check here ```m.length > group```

Scripts/script-MatchRegex.yml
```diff
@@ -0,0 +1,45 @@
...
+      }
+      var ec = {};
+      if (args.out) {
```
meirwah: will change ```out``` to ```contextKey```, will already use the convention in some places...

Scripts/script-MatchRegex.yml
slavikm: I'm checking in the if, no?

Scripts/script-MatchRegex.yml
slavikm: ok, contextKey it is

PR #667, opened 2017-07-24
Light cyber magna

dem-evgeny:


Integrations/integration-LightCyberMagna.yml
```diff
@@ -494,4 +510,62 @@ script:
...
+    - name: sort
+      defaultValue: final_score
+    - name: host_id
+      required: true
+      default: true
+      description: Host id as presented in LightCyber Magna
+    description: Returns all installed autoruns programs in this host
   isfetch: true
+hidden: false
+releaseNotes: "Added the commands lcm-host-autoruns, lcm-host-processes-internet-connections, lcm-host-loaded-modules, lcm-host-processes, lcm-host-processes, lcm-host-suspicious-artifacts, lcm-host-opened-ports"
```
aviadl: - > `Added the following commands`
@gshriki what do you say ?

What about context & important to any of these commands ?

Integrations/integration-LightCyberMagna.yml
```diff
@@ -494,4 +510,62 @@ script:
...
+    description: Shows the used ports in this host
+  - name: lcm-host-suspicious-artifacts
+    arguments:
+    - name: sort
+      defaultValue: final_score
+    - name: host_id
+      required: true
+      default: true
+      description: Host id as presented in LightCyber Magna
+    description: Returns all the suspicious artifacts of this host
```
aviadl: of this host id

Integrations/integration-LightCyberMagna.yml
dem-evgeny: This is an old integration that doesn't push info to context.
I added some commands that added before at some customer.
We currently don't have LightCyber product, so if I will change all the integration to add info to the context I can't test it.

PR #673, opened 2017-07-25
DataXreputation manual override

gshriki:
Changing all dataXreputataion to take manual reputation before checking threat intel services.

Please review but do not merge. Once changes are agreed upon, will redo the changes in the other scripts.

### DIFF
--- a/Scripts/script-DataURLReputation.yml
+++ b/Scripts/script-DataURLReputation.yml
@@ -16,7 +16,7 @@ script: |-
         }
       }
   }
-  return score < 0 ? 0 : score; 
+  return score < 0 ? 0 : score;
 type: javascript
 tags:
 - reputation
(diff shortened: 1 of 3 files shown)


Scripts/script-DataURLReputation.yml
```diff
@@ -4,19 +4,26 @@ commonfields:
 name: DataURLReputation
 script: |-
   var score = -1;
-  var rep = executeCommand('url', {url: args.input});
-  var scores = dq(rep, 'EntryContext.DBotScore.Score');
-  for (var i = 0; scores && i < scores.length; i++) {
-    if (scores[i] && score < scores[i]) {
-      score = scores[i];
-    }
+  var res = executeCommand("findIndicators", {"query": "type:URL and value:" + args.input})
```
aviadl: add `manualScore:T` to filter only for manual score
wrap value with a single bracket `'`

aviadl:
@gshriki do you think you can add a test for this ?

Nitzperetz:
Need to add forceCalc param - to calculate reputation even if indicator was manually set.

PR #676, opened 2017-07-25
Panorama byoi

shaniacht1:
fixes #7161

added panorama BYOI
added dynamic block list command
added script dynamic block list
minor changes to release notes test


Integrations/integration-Panorama.yml
```diff
@@ -0,0 +1,148 @@
...
+                'file' : file
+              },
+              params.insecure,
+              params.proxy
+            );
+
+            if (res.StatusCode !==200) {
+              throw 'Request Failed.\nStatus code: ' + res.StatusCode +' '+errorDict[res.StatusCode]+ '.\nBody: ' + JSON.stringify(res) + '.';
+            }
+            return JSON.parse(x2j(res.Body));
```
Spivaka: we also had a case where the response was 200, but the xml itself has an error. So we need to check this scenario as well and throw

Integrations/integration-Panorama.yml
```diff
@@ -0,0 +1,153 @@
+versionedfields: {}
+commonfields:
+  id: Panorama
+  version: -1
+name: Panorama
+display: Panorama
```
Spivaka: add from version 3.0.0

shaniacht1: so from now I should write the actual version instead of -1?

dorsha: no.. you need to add:
`fromversion: 3.0.0`

shaniacht1: done

PR #679, opened 2017-07-26
IntSights integration

nshap:
Fixes https://github.com/demisto/server/issues/6767.


Integrations/integration-IntSights.yml
```diff
@@ -1,13 +1,15 @@
...
 description: Use IntSights to manage and fetch alerts
+detaileddescription: |-
```
Spivaka: Remove it from here, not the point of detailed description

Integrations/integration-IntSights.yml
```diff
@@ -206,6 +461,23 @@ script:
                 Contents: fileName
             };
             break;
+
+        case 'intsights-send-mail':
+            // convert comma-separated to array
+            args.Emails = args.Emails.split(',');
```
Spivaka: this won't do nothing

nshap: It converts a string such as "a,b,c" to a list [a, b, c]. It works. What do you mean by won't do anything?

PR #688, opened 2017-07-27
Malwr integration

nshap:
Fixes https://github.com/demisto/server/issues/5629.
@Spivaka 

Due to the bug described here: https://github.com/demisto/server/issues/8246, this is a Python script. It receives two mandatory arguments, "commandAction" and "commandArgument". The first one can be Submit/Status/Result, and the second one is either a file ID in Demisto, or the analysis ID returned from the "Submit" command.
Username and password are optional.

It uses the Python API for Malwr, with little changes (mostly code removal).


Scripts/script-Malwr.yml
```diff
@@ -0,0 +1,262 @@
...
+
+      res, soup = malwr.submit_sample(filepath=filepathResult[0]['Contents']['path'])
+      if 'analysis_link' in res:
+          analysisId = res['analysis_link'].split('/')[-2]
+
+          message = 'File submitted: %s%s\n' % (MAIN_URL, res['analysis_link'])
+          message += 'MD5: %s\n' % (res['md5'], )
+          message += 'Analysis ID: %s' % (analysisId, )
+
+          entry['Contents'] = str(soup)
```
meirwah: soup here is in json format? if not, you need to convert it to json string before putting in contents

Scripts/script-Malwr.yml
```diff
@@ -0,0 +1,262 @@
+versionedfields: {}
+commonfields:
+  id: 20bd44d4-9abf-4913-8903-c323baee80ab
```
meirwah: id should be the name of the script

Scripts/script-Malwr.yml
```diff
@@ -0,0 +1,262 @@
...
+  version: 151
```
meirwah: version should be -1

Scripts/script-Malwr.yml
nshap: soup is the HTML page, e.g.:
image

How can you convert this to JSON?

meirwah: >>> import json, xmljson
>>> from lxml.etree import fromstring, tostring
>>> xml = fromstring('<p id="1">text</p>')
>>> json.dumps(xmljson.badgerfish.data(xml))
'{"p": {"@id": 1, "$": "text"}}'

Spivaka: @meirwah isn't this for xml?

meirwah: ohh soup, I was sure it was soap... never mind then

Spivaka:
this misses comments and release notes so test fail

PR #690, opened 2017-07-28
rasterize - added proxy settings

gshriki:


Integrations/integration-rasterize.yml
```diff
@@ -5,21 +5,32 @@ name: Rasterize
...
+  required: false
 script:
   script: |-
     import os, subprocess,sys
     reload(sys)
     sys.setdefaultencoding("utf-8")
+    proxy = demisto.params()["proxy"]
+
+    if proxy:
+        http_proxy = os.environ["http_proxy"]
```
slavikm: Is this going to work? Is the proxy reflected inside of docker?

gshriki: yes it does.
checked:
2017-07-29 12:56:00.0946 info rasterize proxy settings: --proxy=127.0.0.:80 (source: /Users/shriki/Demisto/src/src/github.com/demisto/server/services/automation/pythonrunner.go:363)

PR #692, opened 2017-07-28
Changed reports created by to DBot

dorsha:


dorsha:
I guess it gonna fail because of the release notes can't handle reports...

PR #698, opened 2017-07-30
fix json parsing for reputations as well

aviadl:


content-descriptor.json
```diff
@@ -1 +1,11 @@
...
\ No newline at end of file
+{
+  "installDate": "0001-01-01T00:00:00Z",
+  "assetId": "REPLACE_THIS_WITH_CI_BUILD_NUM",
+  "releaseNotes": "## Release Notes for version 2.5.0",
```
meirwah: shouldn't this be 2.6.1?

content-descriptor.json
```diff
@@ -1 +1,11 @@
...
+  "modified": "REPLACE_THIS_WITH_RELEASE_DATE",
+  "ignoreGit": false,
+  "releaseDate": "REPLACE_THIS_WITH_RELEASE_DATE",
+  "version": -1,
+  "release": "2.5.0",
```
meirwah: shouldn't this be 2.6.1?

content-descriptor.json
aviadl: Doesnt really matter as we create it from flat each build according to the yml file

content-descriptor.json
aviadl: Doesnt really matter as we create it from flat each build according to the yml file

meirwah:
@aviadl to merge?

PR #713, opened 2017-08-01
Fix escapeMD to escape \r

maiorrulz:
Issue 
https://github.com/demisto/server/issues/8061


meirwah:
@aviadl @Spivaka @dorsha plz review this too, since this is in a sensitive place.
I know we already merge this, but if you think this is a mistake, we can revert

meirwah:
@maiorrulz please add test for this in server side

dorsha:
Looks fine (even \r is not exactly a new line, but it is ok)

PR #718, opened 2017-08-02
Service now ticket type

dem-evgeny:


Spivaka:
The build fails. Otherwise looks fine

dem-evgeny:
@Spivaka It fails because of other tests.
But thanks for the approval

PR #720, opened 2017-08-02
Initial addition - fetch will be added after discussions with TCell

trorabaugh:


Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+description: TCell is a cloud based web application firewall
+configuration:
+- display: Server URL (e.g. https://192.168.0.1)
+  name: server
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Enter API Key (Full Access)
```
Spivaka: We usually use "API Key". look at wildfire as an example

Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+  name: api_key
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Do not validate server certificate (insecure)
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: ""
```
Spivaka: Proxy has a display name in other integrations

Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+
+
+    var tcellListApps = function() {
+        var respBody = sendRequest(methodDictionary['tcell-list-apps'], urlDictionary['tcell-list-apps'],'', '');
+        return {
+            Type: entryTypes.note,
+            Contents: respBody,
+            ContentsFormat: formats.json,
+            HumanReadable: tableToMarkdown('TCell Apps', respBody.apps),
+            EntryContext: {'tcell_apps':respBody.apps}
```
Spivaka: we usually manipulate the data to extract the main things out of it and not pushing all the to context

Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+            Type: entryTypes.note,
+            Contents: respBody,
+            ContentsFormat: formats.json,
+            HumanReadable: tableToMarkdown('TCell App', [respBody.app]),
+            EntryContext: {'tcell_app_info': [respBody.app]}
+        };
+    }
+
+    var tcellListAgents = function() {
+        var urlPath = urlDictionary['tcell-list-agents'].replace('%APP_ID%', args.appid);
```
Spivaka: we have a "replaceInTemplates" and "replaceInTemplatesAndRemove" that does a very similar task. please use it

Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+            Type: entryTypes.note,
+            Contents: respBody,
+            ContentsFormat: formats.json,
+            HumanReadable: tableToMarkdown('TCell Agents', respBody.agents),
+            EntryContext: {'tcell_agents_list': respBody.agents}
+        };
+    }
+
+    var tcellAgentData = function() {
+        var urlPath = urlDictionary['tcell-agent-data'].replace('%APP_ID%', args.appid).replace('%AGENT_ID%',args.agentid);
```
Spivaka: here you would be able to use replaecInTemplates(urlDictionary['tcell-agent-data'], args) instead of the replaces

Integrations/integration-TCell.yml
```diff
@@ -0,0 +1,635 @@
...
+                reqObj,
+                params.insecure,
+                proxy,
+                true
+            );
+            if (res.StatusCode < 200 || res.StatusCode >= 300) {
+                throw 'Action failed, request status code: ' + res.StatusCode + ' and Body: ' + res.Body + '.';
+            }
+
+            if (res.Headers['Content-T
```
Spivaka: why so many white spaces?

Spivaka: remove it if you are not going to use it

Spivaka: we usually provide full information on data pushed to context. we first extract what ever is important for the context, and than fully describe it

Integrations/integration-TCell.yml
trorabaugh: Do you mean you use the words API Key not Enter API Key (Full Access)?

Integrations/integration-TCell.yml
trorabaugh: I actually copied this from another integration.

Integrations/integration-TCell.yml
trorabaugh: I would generally do that however, as stated this is the initial version. Their API's maybe changing thus, what is being pushed into the context may also be changing.

Integrations/integration-TCell.yml
trorabaugh: Great, it might be really useful to have some of this in the contribution document as well along with the standard for how things should be structured.

Integrations/integration-TCell.yml
trorabaugh: It should have converted to 4 tabspaces if not I will figure out why exactly. I generally use 4 tab spaces. This might have been the culprit of a newly downloaded sublime versus previous IDE's I've used.

trorabaugh: That is the difficult portion about this at the moment. Determining what is important for the context and fully describing it is difficult as it is an Alpha version and their API is also being updated so this may actually change in a later version.

Integrations/integration-TCell.yml
trorabaugh: Is there an advantage of using replaceInTemplates('<entry>%entry%</entry>', { entry: entry }) over simply using .replace?

Spivaka: Yes,
1. It will require less replaces as it will replace all %key1%, %key2% in values in a single call.
2. Code consistency across out product.

trorabaugh:
Waiting on TCell

dorsha:
Closing for now

PR #725, opened 2017-08-03
Remove unused args

Spivaka:
fixes https://github.com/demisto/server/issues/8386


Scripts/script-UnzipFile.yml
```diff
@@ -2,6 +2,7 @@ commonfields:
   id: UnzipFile
   version: -1
 name: UnzipFile
+releaseNotes: "Remove unused arguments"
```
aviadl: Change to "-"

PR #728, opened 2017-08-03
Protectwise - upgraded integration and deprecated scripts

liorkol:
fixes https://github.com/demisto/server/issues/6457


Integrations/integration-ProtectWise.yml
```diff
@@ -1,25 +1,18 @@
...
 description: Cloud based Security Network DVR
-detaileddescription: |-
-  Integration with ProtectWise cloud product.
-  Please note you can use either token (acquired by api call : https://api.protectwise.com/momfodhxhz/#access-token-token) or user email/password for authentication.
+detaileddescription: Integration with ProtectWise cloud product.
```
aviadl: Did you deliberately remove all the detailed description? if adding back make sure to use proper markdown syntax for URLs

Integrations/integration-ProtectWise.yml
```diff
@@ -63,96 +98,45 @@ script:
...
+            params.insecure,
+            params.proxy
             );
         if (res.StatusCode !== 200) {
-            throw 'Failed to execute download Event Pcap request:' + res.StatusCode + ', body: ' + res.Body;
+            throw 'Failed to execute ' + url + ' : ' + res.StatusCode + ', body: ' + res.Body;
         }
-        return saveFile(res.Body);
+
+        return (raw ? res.Body : JSON.parse(res.Body));
```
aviadl: You should wrap parse with try catch, so we can return proper error if body was not a proper json
Look at other places you do parse as well

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+        for (var i = 0; i<keys.length; i++) {
+            val = event[keys[i]];
+            if (TIME_FIELDS.indexOf(keys[i]) > -1) {
+                val = convertTimestampToString(val);
+            }
+            labels.push({'type': keys[i], 'value': String(val)});
+        }
+        return {
+            "name": event.message,
+            "labels": labels,
```
aviadl: Need to add rawJSON so it will be able to  go through mapping properly

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
     switch (command) {
         case 'test-module':
-            var token = getToken();
            if (token && token.length > 0){
                 return true;
             }
             return false;
+        case 'fetch-incidents':
+            var lastRun = getLastRun();
+            var startTime = (lastRun && lastRun.startTime ? lastRun.startTime : (new Date().getTime()- 1000 * 60 ));
```
aviadl: This is last minute back, usually we give it a but more

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+                        }
+                    }
+                } else {
+                    skip = false;
+                }
+                if(!skip) {
+                    incidents.push(createIncidentFromEvent(res[i]));
+                }
+            }
+            setLastRun({startTime: startTime});
```
aviadl: You are not incrementing the lastRun, or i am missing something

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+            return JSON.stringify(incidents);
         case 'sensors':
-            var token = getToken();
-            return {Type: 1, Contents: sensors(args.sensorId,token), ContentsFormat: 'json'};
+        case 'protectwise-show-sensors':
+            var url = params.url+'sensors';
```
aviadl: In most integrations we protect ourselves from cases placed / didnt place the `/`

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+                   }
+               });
+            });
+
+            return {
+                Type: entryTypes.note,
+                Contents: raw,
+                ContentsFormat: formats.json,
+                HumanReadable: tableToMarkdown('Protectwise Event Search', items, eventHeaders),
+                EntryContext: {'Protectwise.Event(val.id == obj.id)': items}
```
aviadl: Dont you have here a lot of events ? what ID are you referring to ?

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+                   }
+               });
+            });
+
+            return {
+                Type: entryTypes.note,
+                Contents: raw,
+                ContentsFormat: formats.json,
+                HumanReadable: tableToMarkdown('Protectwise Observation Search', items, headers),
+                EntryContext: {'Protectwise.Observation(val.id == obj.id)': items}
```
aviadl: look at the comment above

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+                ContentsFormat: formats.json,
+                HumanReadable: md,
+                EntryContext: {'Protectwise.Event(val.id == obj.id)': {'id': res.id, 'PCAPSize': res.estimatedSize}}
+            };
         case 'observation-pcap-download':
-            var token = getToken();
-            return {Type: 3, FileID: observationEventPcap(args.id,args.sensorId,args.filename,token), File: args.filename, Contents: 'we must have contents for an entry'};
+        case 'protectwise-observation-pcap-download':
+            var filename = (args.filename && args.filename.length > 0) ? args.filename : ( args.sensorId+'-'+args.id + '.pcap');
+            var url = params.url+'pcaps/observations/'+args.sensorId+'/'+args.id + '?filename=' + filename;
```
aviadl: you need to escape filename (we have a function for that in common

Integrations/integration-ProtectWise.yml
```diff
@@ -201,101 +187,300 @@ script:
...
+            return {
+                Type: entryTypes.note,
+                Contents: res,
+                ContentsFormat: formats.json,
+                HumanReadable: md,
+                EntryContext: {'Protectwise.Observation(val.id == obj.id && val.sensorId == obj.sensorId)': {'id': args.id, 'sensorId': args.sensorId, 'PCAPSize': res.estimatedSize}}
+            };
         case 'get-token':
-            return getToken();
+            return token;
```
aviadl: above you used the function getToken
why did you remove it here ?

Integrations/integration-ProtectWise.yml
liorkol: Removed the URL. Token no longer an option to use instead of password.

Integrations/integration-ProtectWise.yml
liorkol: Updated to 5mins

Integrations/integration-ProtectWise.yml
liorkol: Good idea. Done

Integrations/integration-ProtectWise.yml
liorkol: Right.. it's an array of events. But it merges on a per-item basis
Seemed to work at least when I tested, let me know if you think I am misunderstanding

Integrations/integration-ProtectWise.yml
liorkol: getToken always gets called, before checking which command it is - so no reason to call it again, just return the token we already got

Integrations/integration-ProtectWise.yml
liorkol: Done

liorkol:
@aviadl Fixed and tested, merge when ready

Integrations/integration-ProtectWise.yml
```diff
@@ -230,10 +239,19 @@ script:
...
         case 'fetch-incidents':
             var lastRun = getLastRun();
-            var startTime = (lastRun && lastRun.startTime ? lastRun.startTime : (new Date().getTime()- 1000 * 60 ));
-            var data = eventsSearch(startTime , new Date().getTime() , params.eventType, params.killChainStage, params.threatLevel,
+            var now = (new Date()).getTime();
+            if (!lastRun || !lastRun.time) {
+                // First time, retrieve events from the last 10 min
+                lastRun = {time: now - 10 * 60 * 1000};
+            }
+            setLastRun({time: now});
```
aviadl: You should set last run, only if fetch was successful
Else you may lose timeframes
If you intend to have it this way, then ignore

PR #729, opened 2017-08-04
Removed required parameter

dorsha:
fixes: https://github.com/demisto/server/issues/8387

### DIFF
--- a/Integrations/integration-Guidance_Encase_Endpoint.yml
+++ b/Integrations/integration-Guidance_Encase_Endpoint.yml
@@ -20,7 +20,7 @@ configuration:
   name: port
   defaultvalue: ""
   type: 0
-  required: true
+  required: false
 - display: Use system proxy settings
   name: proxy
   defaultvalue: ""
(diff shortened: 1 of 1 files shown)


Integrations/integration-Guidance_Encase_Endpoint.yml
```diff
@@ -20,7 +20,7 @@ configuration:
   name: port
   defaultvalue: ""
   type: 0
-  required: true
+  required: false
```
aviadl: So you need a default port if you make it non required

dorsha: OK, what will be the default port?

aviadl: I have no idea :) @orius123  ?

orius123: Maybe 8700. Saw it somewhere as default and it is as good as anything else..

dorsha: We will leave it required.
We didn't found the original field that was not required.

PR #743, opened 2017-08-07
Set default address for send email in phishing playbook

Nitzperetz:


dorsha:
@Nitzperetz you have conflicts now

PR #746, opened 2017-08-07
Added the ability to validate file extensions

dorsha:
And fixed two files.

fixes: https://github.com/demisto/server/issues/8494


asafshen:
& quick :)

PR #748, opened 2017-08-08
AreValuesEqual - making arguments non-mandatory

gshriki:
In order to handle correctly situations where one of the arguments is empty/missing - returning no instead of raising an error in the playbook and awaiting users intervention

### DIFF
--- a/Scripts/script-AreValuesEqual.yml
+++ b/Scripts/script-AreValuesEqual.yml
@@ -33,8 +33,7 @@ outputs:
 - contextPath: no
   description: if left is not equal to right
 scripttarget: 0
-comment: Check whether the values provided in arguments are equal. If either of the
-  arguments are missing, no is returned.
+comment: Check whether the values provided in arguments are equal. If either of the arguments are missing, no is returned.
 dependson: {}
 timeout: 0s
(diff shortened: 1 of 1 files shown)


Scripts/script-AreValuesEqual.yml
```diff
@@ -3,13 +3,17 @@ commonfields:
   version: -1
 name: AreValuesEqual
 script: |
+  if (!args.left || !args.right) {
```
dorsha: if both empty you should return yes.

Scripts/script-AreValuesEqual.yml
```diff
@@ -19,18 +23,18 @@ tags:
...
   description: if left is equal to right
 - contextPath: no
   description: if left is not equal to right
 scripttarget: 0
-comment: Check whether the values provided in arguments are equal.
+comment: Check whether the values provided in arguments are equal. If either of the
+  arguments are missing, no is returned.
 dependson: {}
 timeout: 0s
+releaseNotes: Arguments are not mandatory anymore. If an argument is missing (or provided with an empty value), no is returned.
```
dorsha: change the releaseNotes

dorsha:
@gshriki - fix description

PR #766, opened 2017-08-15
Release notes reset automation

yaakovi:
Fixes: https://github.com/demisto/server/issues/8575
added a stand-alone script that removes all release notes from json\yml files

### DIFF
--- a/release_notes_clear.py
+++ b/release_notes_clear.py
@@ -105,12 +105,12 @@ def main(root_dir):
     for folder in yml_folders_to_scan:
         print 'scanning directory: %s' % (folder, ),
         changed = remove_releaseNotes_folder(os.path.join(root_dir, folder), '*.yml')
-        print 'Done: removed %d files' % (changed, )
+        print 'Done: removed %d from files' % (changed, )
 
     for folder in json_folders_to_scan:
         print 'scanning directory: %s' % (folder, ),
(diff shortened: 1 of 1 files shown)


release_notes_clear.py
```diff
@@ -0,0 +1,117 @@
+# remove all releaseNotes from files in: Itegrations, Playbooks, Reports and Scripts.
+# Note: using yaml will destroy the file structures so filtering as regular text-file.
+import os
```
asafshen: maybe mark that to use the script, you should be in the root directory & run `python release_notes_clear.py`

release_notes_clear.py
```diff
@@ -0,0 +1,117 @@
...
+    yml_folders_to_scan = ['Integrations', 'Playbooks', 'Scripts', ] # yml
+    json_folders_to_scan = ['Reports', ] # json
+
+    for folder in yml_folders_to_scan:
+        print 'scanning directory: %s' % (folder, ),
+        changed = remove_releaseNotes_folder(os.path.join(root_dir, folder), '*.yml')
+        print 'Done: removed %d from files' % (changed, )
+
+    for folder in json_folders_to_scan:
+        print 'scanning directory: %s' % (folder, ),
```
asafshen: lets be consistent with capital case in the start of print

PR #772, opened 2017-08-16
Deprecate SendURLDetailsByEmail and add a script to count ads

slavikm:
Related to https://github.com/demisto/server/issues/8346


Scripts/script-URLNumberOfAds.yml
```diff
@@ -0,0 +1,47 @@
...
+args:
+- name: url
+  required: true
+  default: true
+  description: URL to load and count ads on
+- name: easylist
+  description: Optional easylist to get ads from
+  defaultValue: "https://easylist.github.io/easylist/easylist.txt"
+outputs:
+- contextPath: Ads.URL
```
aviadl: Shouldnt it be URL.Ads
and also add URL.data
so it will merge with all the other url stuff we are doing ?

Scripts/script-SendURLDetailsByEmail.yml
```diff
@@ -91,6 +91,7 @@ tags:
 comment: Get all details about URL (reputation, number of ads, etc.) and send them
   via email
 system: true
+deprecated: true
```
aviadl: This script needs to be totally removed
Not just deprecated

Scripts/script-URLNumberOfAds.yml
slavikm: Don't want to mix the lists because we don't want to iterate on the ads normally

PR #777, opened 2017-08-17
fix issue in TableToMarkdown

meirwah:


slavikm:
@meirwah please make sure in this case to fix `ToTable` as well to use `dq`.

meirwah:
@slavikm we are dropping this solution in the end, please see @maiorrulz comments on the issue

PR #779, opened 2017-08-17
Test playbook witout integration

asafshen:
- ability to run test playbook without integration
- enlarge test playbook global timeout


asafshen:
@meirwah please review when possible

PR #784, opened 2017-08-20
Table to mark down reverse and add playbook

maiorrulz:


Tests/scripts/ensurefilenames.sh
```diff
@@ -32,7 +32,6 @@ ensureFilename Reports report- .json
 ensureFilename Scripts script- .yml
 ensureFilename Scripts script- .yml
 ensureFilename Misc reputations .json
-ensureFilename TestPlaybooks playbook- .yml
```
meirwah: are you sure you want to delete this??? @asafshen ?

maiorrulz: Yes, spoke with Asaf,

PR #785, opened 2017-08-20
Flat object before passing to tableToMardown

maiorrulz:


meirwah:
@maiorrulz did you validated this fix? look like something that can break many integrations

Scripts/script-ToTable.yml
```diff
@@ -5,7 +5,8 @@ commonfields:
 name: ToTable
 script: |-
   var headers = args.columns ? args.columns.split(',') : null;
-  return {Type: entryTypes.note, Contents: tableToMarkdown(args.title, args.data, headers), ContentsFormat: formats.markdown};
+  var flattenObject = treeToFlattenObject(args.data);
```
slavikm: `ToTable` receives an array and needs to flatten the elements while keeping an array to pass to `tableToMarkdown`

slavikm:
What is the difference between: `treeToFlattenObject` and `flattenFields`?

maiorrulz:
I think they doing the same almost the same, except of treeToFlattenObject can flat array of objects, and if it is array of primitives, function returns array which can be printed in table

Scripts/script-ToTable.yml
maiorrulz: What is use case of toTable, I never used it, it is better to understand when you are making fix :-)

slavikm: I added it recently to allow taking an array in the context and printing it nicely. It's basically a wrapper script for `tableToMarkdown`.

meirwah:
@maiorrulz can we verify this PR?

aviadl:
@meirwah @maiorrulz any reason that this is not merged ?

meirwah:
@aviadl they were concerns this may break some integrations, asked @maiorrulz to double check to be sure

maiorrulz:
I didn't find any integration in our content which uses **toTable** function.

PR #788, opened 2017-08-21
Do not escape dot on table to markdown

Nitzperetz:


dorsha:
@slavikm you added it, so you should give the answers :)

slavikm:
Hmmm... If I added it, it means we needed this for something. So, I'm afraid this will break other stuff.

aviadl:
@dorsha isn't this related to the fix we did, where we saw all the chars that needs to be escaped in MD?

dorsha:
A different one

meirwah:
@Nitzperetz what is this pending on?

PR #789, opened 2017-08-22
bug fix -returned json instead of file

shaniacht1:


Integrations/integration-Threat_Grid.yml
```diff
@@ -308,6 +309,9 @@ script:
             args,
             filename
         );
+        if (filename){
+            return rawResponse;
```
asafshen: I think its not necessary to call `sendRequest` if `!filename`
What happen if `!filename`?
Can you elaborate on this bug?

shaniacht1: If you look at line 80 (the end of the sendRequest) you'll see that if filename is set, it returns a file entry, otherwise it returns the rawResponse.
The issue was, that the lines following the sendRequest were parsing the data (expecting raw data). The file entry does not need to be parsed and should be returned to the war room as is.

PR #791, opened 2017-08-24
added html format

shaniacht1:
related to https://github.com/demisto/server/issues/8564


dorsha:
@shaniacht1 need to add for the python common as well:

image

PR #793, opened 2017-08-27
add tc-get-incident-associate-indicators command

meirwah:


meirwah:
@Spivaka can you review this?

PR #797, opened 2017-08-28
change ConfiguredModules to modules

michalgold:
change ConfiguredModules to modules


michalgold:
Please note that this was not tested.

PR #801, opened 2017-08-31
Adding support for mixed CR/LF in ParseEmailFiles

azollman:
When raw messages are retrieved from a server, there sometimes have mixed line terminations. These messages are still parsable by Python's email module, but the filetype string is not recognized in this Script.


dorsha:
@azollman you need to add `releaseNotes` tag to explain your changes. For example:

image

PR #802, opened 2017-08-31
Added createEntry function

shaniacht1:
fixes https://github.com/demisto/server/issues/8688
added the following functions:
createEntry
createContext
mapObjFunction
toArray
isFunction
dqQueryBuilder
easyDQ


Scripts/script-CommonServer.yml
```diff
@@ -230,6 +230,169 @@ script: |-
...
+              res.push(createContextSingle(data[j]));
+          }
+          return res;
+      }
+      return createContextSingle(data);
+  }
+
+  function isFunction(functionToCheck) {
+   var getType = {};
+   return functionToCheck && getType.toString.call(functionToCheck) === '[object Function]';
```
slavikm: This looks a bit ugly and also not indented... I think you can:
```js
return functionToCheck && typeof functionToCheck === 'function';
```

shaniacht1: @slavikm changed

PR #809, opened 2017-09-03
Archer - create/update record with non string types

shaniacht1:
fixes: https://github.com/demisto/server/issues/8876


Spivaka:
did you checked it with our instance?

PR #815, opened 2017-09-06
add utiliy func scoreToReputation to common

meirwah:


Scripts/script-CommonServer.yml
```diff
@@ -1103,6 +1103,21 @@ script: |-
       email: 'Account.Email(val.Address && val.Address == obj.Address)'
   };
 
+
+  // Converts score (in number format) to human readable reputation format
+  function scoreToReputation(score) {
```
dorsha: use `===` in all places.
And don't we have constants for these numbers?

Scripts/script-CommonServer.yml
```diff
@@ -1103,6 +1103,21 @@ script: |-
...
+      }
+      if (score == 2){
+          return 'Suspicious';
+      }
+      if (score == 1){
+          return 'Good';
+      }
+      return 'None';
+  };
+
```
dorsha: I would also remove this and use your new func:

image

meirwah: don't understand the remark here.. what to remove?

dorsha: The entire function since you have it in common now

meirwah: that script is deprecated, only for version <=3.1, so can't use that func there

PR #824, opened 2017-09-07
The ability to extract urls from query string

dorsha:
image

fixes: https://github.com/demisto/server/issues/8705


dorsha:
Will add test playbook soon.

dorsha:
Added test playbook for ExtractURL

PR #825, opened 2017-09-07
IsValueInArray - check if a value is in an array

gshriki:
Renamed script
Added test playbook


dorsha:
@gshriki - you need to update the Tests/conf.json with the test playbook

PR #826, opened 2017-09-08
nmap support

slavikm:
Support running nmap scans


Integrations/integration-nmap.yml
```diff
@@ -0,0 +1,98 @@
...
+            if len(host.hostnames):
+                tmp_host = host.hostnames.pop()
+                h['Hostname'] = tmp_host
+            else:
+                tmp_host = host.address
+
+            h['Address'] = host.address
+            h['Status'] = host.status
+            svc = []
+            md += "- Nmap scan report for {0} ({1})".format(tmp_host, host.address) + '\n'
```
aviadl: If tmp_host == host address this line will look wierd

Integrations/integration-nmap.yml
```diff
@@ -0,0 +1,98 @@
...
+      description: The protocol of the service
+    - contextPath: NMAP.Scan.Hosts.Services.State
+      description: The state of the service
+    - contextPath: NMAP.Scan.Hosts.Services.Banner
+      description: Any captured banner from the service
+    - contextPath: NMAP.Scan.Hosts.Services.Service
+      description: The service name
+    description: Scan targets with the given parameters
+    execution: true
+  dockerimage: demisto/nmap
```
aviadl: You need to add this image to requirements.txt 
Also to the zip file we now have with all docker images

Integrations/integration-nmap.yml
```diff
@@ -0,0 +1,98 @@
...
+            svc = []
+            md += "- Nmap scan report for {0} ({1})".format(tmp_host, host.address) + '\n'
+            md += "- Host is {0}.".format(host.status) + '\n'
+            md += 'Port | State | service\n-|-|-\n'
+            for serv in host.services:
+                pserv = "{0} / {1} | {2} | {3}".format(
+                        str(serv.port),
+                        serv.protocol,
+                        serv.state,
+                        serv.service)
```
aviadl: service might contain chars that needs to be escaped, as they might be MD special chars

Integrations/integration-nmap.yml
```diff
@@ -0,0 +1,98 @@
...
+                if len(serv.banner):
+                    pserv += " ({0})".format(serv.banner)
```
aviadl: same for banner
Also we dot have a header for banner, is this on purpose?

Integrations/integration-nmap.yml
slavikm: This is the nmap style reporting but I see what you mean

Integrations/integration-nmap.yml
slavikm: Changed to use standard `tableToMarkdown`

Integrations/integration-nmap.yml
slavikm: Why requirements.txt? It's not in our standard image - it's a separate image - or maybe I'm missing something...

aviadl: You are

PR #830, opened 2017-09-10
Fixed fetch incidents of cases:

yaakovi:
get all new cases associated with integration user with ID over integration parameter


yaakovi:
on small configuration time period (10s) between two consecutive schedule jobs (fetch cases), the max sessions can be reached (around 100). each session timeout is around 5 minutes so its not a big deal in production and every day use.

PR #836, opened 2017-09-12
Added the ability to provide demisto lib location

dorsha:


gshriki:
maybe add some code to handle user's input without trailing slash?

PR #838, opened 2017-09-12
skip hp-service test

asafshen:
skip hp-service-test since it instance was shut-down


meirwah:
@asafshen what has changed here? looks the same...

PR #840, opened 2017-09-13
Generate Google auth URL for given scopes

slavikm:


Scripts/script-GoogleAuthURL.yml
```diff
@@ -0,0 +1,46 @@
+commonfields:
+  id: GoogleAuthURL
+  version: -1
+name: GoogleAuthURL
+script: |
+  var scopes = argToList(args.scopes);
+  var res = http('https://demistobot.demisto.com/google-oauth' + encodeToURLQuery({scopes: scopes.join(',')}),
```
meirwah: you are doing join on scopes, but isn't it already in comma sep format? according to desc?

slavikm: It might be or it might be an array - I treat it generically using `argToList`

PR #843, opened 2017-09-15
EWS Move to 'Messaging' category

slavikm:


slavikm:
@aviadl but also pulls things in (mail listener). What else would you place it in? It's definitely not SIEM.

aviadl:
Deleting branch, will be fixed in another PR

PR #845, opened 2017-09-17
Create CyberArk AIM integration

orius123:
Command for querying CyberArk vault through CCP REST API.
Vault Module that can fetch credentials.

### DIFF
--- a/Integrations/integration-cyberark_AIM.yml
+++ b/Integrations/integration-cyberark_AIM.yml
@@ -67,7 +67,7 @@ script:
             );
         if ((allowOtherStatuses && res.StatusCode !== 400) && (res.StatusCode < 200 || res.StatusCode >= 300)) {
             if (res.StatusCode === 404) {
-                return 'no results found';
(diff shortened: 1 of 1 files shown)


meirwah:
@orius123 is this ready?

orius123:
Yes.
Pending review.

Integrations/integration-cyberark_AIM.yml
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    Method: method,
+                    Body: method !== 'GET' ? JSON.stringify(args) : ''
+                },
+                insecure,
+                proxy
+            );
+        if ((allowOtherStatuses && res.StatusCode !== 400) && (res.StatusCode < 200 || res.StatusCode >= 300)) {
+            if (res.StatusCode === 404) {
+                return 'no results found';
```
Spivaka: You expect an object if success, but string in case 0f 404, might be better to return undefined and check it instead of a string

PR #846, opened 2017-09-18
Add string replace

noaco:


Scripts/script-StringReplace.yml
```diff
@@ -0,0 +1,82 @@
...
+      setContext('StringReplace.Result', val);
+      return {
+          Type: entryTypes.note,
+          Contents: val,
+          ContentsFormat: formats.text
+      };
+  }
+  return {
+      Type: entryTypes.note,
+      Contents: 'No regex match to replace',
```
anara123: return the original string. There can be cases where user just wants to replace some string, and if there is nothing to replace then return the original value

anara123: And don't forget to setContext here too

anara123:
@noaco You also have conflicts, please resolve the conflicts then merge

Scripts/script-StringReplace.yml
```diff
@@ -0,0 +1,78 @@
...
+  description: The regex to find matches to be replaced with a new value
+- name: newValue
+  required: true
+  description: The new value to replace the regex match. Pass '' to remove regex match.
+- name: replaceAll
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: The boolean value to indicate if 'g' flag is passed to regex constructor
```
dorsha: I would explain what does g means.. something like, true to replace all, false to replace only the first occurrence.
In most cases users are not programmers.
Same for the other arguments (i and m)

Scripts/script-StringReplace.yml
```diff
@@ -0,0 +1,78 @@
...
+  //if no match found - return the original string
+  setContext('StringReplace.Result', val);
+  return {
+      Type: entryTypes.note,
+      Contents: val,
+      ContentsFormat: formats.text
+  };
+type: javascript
+tags:
+- Utilities
```
dorsha: Should be `Utility` not `Utilities`, otherwise you won't see this in the Tasks Library

Scripts/script-StringReplace.yml
```diff
@@ -0,0 +1,78 @@
...
+  if (args.multiLine === 'true') {
+      arr.push('m');
+  }
+
+  var newValue = args.newValue === "''" ? "" : args.newValue;
+  var r = new RegExp(args.regex, arr.join(''));
+  var val = args.data.replace(r, newValue);
+
+  //if no match found - return the original string
+  setContext('StringReplace.Result', val);
```
dorsha: Using SetContext is risky. Encountered several time a race condition with it. I would just return it as output for the next tasks.

anara123: We already think about it. Just return the result is not enough. We want to have all the script some output so the users can easily use that as input. But you right about race condition, but in other hand if we use `EntryContext` instead, then at the first run it will set output to `StringReplace.Result` and at the second run it will make it array, which is not user friendly (user must modify the output from source). This is an open issue, something that we need to decide how we want to make it user friendly and usable

anara123:
@noaco fix @dorsha 's notes, then merge

dorsha:
Closing for now, too many days open (keeping the branch).

Tests/conf.json
```diff
@@ -78,14 +78,16 @@
...
             "playbookID": "TestHPServiceManager",
             "timeout": 400
         },
         {
+            "playbookID": "TestStringReplace"
+        },
+        {
+            "playbookID": "TestHttpPlaybook"
+        },
             "name": "VxStream",
```
anara123: you are missing curly braces here

PR #847, opened 2017-09-18
add StringLength script

noaco:


Scripts/script-StringLength.yml
```diff
@@ -0,0 +1,28 @@
+commonfields:
+  id: StringLength
+  version: -1
+name: StringLength
+script: |+
+  if (typeof args.str !== 'string') {
+      return {
+          Type: entryTypes.error,
+          Contents: 'Argument must be of type \'string\'. Not {0} type'.format(typeof args.str)
```
anara123: `format` is python function if I am not mistaken. There is no such in JS. You don't have to check `args.str` to be string because all the args are strings when they passed from the CLI.

Scripts/script-StringLength.yml
```diff
@@ -0,0 +1,28 @@
...
+          Type: entryTypes.note,
+          Contents: args.str.length,
+          ContentsFormat: formats.txt
+      };
+  }
+
+type: javascript
+tags:
+- Utilities
+comment: Returns the length of  the string passed as argument
```
anara123: `of  the` you have 2 spaces here. make it 1

Scripts/script-StringLength.yml
```diff
@@ -0,0 +1,28 @@
...
+  id: StringLength
+  version: -1
+name: StringLength
+script: |+
+  if (typeof args.str !== 'string') {
+      return {
+          Type: entryTypes.error,
+          Contents: 'Argument must be of type \'string\'. Not {0} type'.format(typeof args.str)
+      };
+  } else {
```
dorsha: no need for the else since you return inside the first if

Scripts/script-StringLength.yml
```diff
@@ -0,0 +1,28 @@
...
+      return {
+          Type: entryTypes.note,
+          Contents: args.str.length,
+          ContentsFormat: formats.txt
+      };
+  }
+
+type: javascript
+tags:
+- Utilities
```
dorsha: Should be Utility

Scripts/script-StringLength.yml
```diff
@@ -0,0 +1,28 @@
...
+comment: Returns the length of  the string passed as argument
+enabled: true
+args:
+- name: str
+  required: true
+  description: The string to evaluate
+scripttarget: 0
```
dorsha: I think we should define output here

Spivaka:
please check why tests fail

anara123:
McAfeeESMTest test is unstable. There is an open issue on it https://github.com/demisto/server/issues/9027

PR #853, opened 2017-09-20
MAR integration

asafshen:
fixes: https://github.com/demisto/server/issues/7570

Also sent a different version to @gshriki (works at current build version - because we use var from common)


asafshen:
@meirwah thanks,

Regarding the dbot score - MAR doesn't return any score upon `file`/`ip`

meirwah:
In that case @asafshen maybe it's not a good idea to call them !ip/!url/!file commands, not sure..
@liorkol @Spivaka  WDYT?

Integrations/integration-McAfee-MAR.yml
```diff
@@ -0,0 +1,462 @@
+commonfields:
+  id: McAfee Active Response
+  version: -1
+name: McAfee Active Response
+display: McAfee Active Response
+category: Authentication
```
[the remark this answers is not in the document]
NeelRust: Category should be `Endpoint`. One word.

aviadl:
@meirwah is right, if commands do not have reputation, they should not use the standard reputation commands `ip` , `file` etc

asafshen:
Thanks. Removed`!ip` & `!file`  & change category

Integrations/integration-McAfee-MAR.yml
```diff
@@ -0,0 +1,332 @@
...
+      - Software
+      - Startup
+      - UsbConnectedStorageDevices
+      - UserProfiles
+      - WinRegistry
+      description: The collector to query
+    - name: outputs
+      description: Comma separated fields (outputs) to view - run `mar-collectors-list`
+        command for all collectors outputs, shows all outputs if not supplied
+      isArray: true
```
meirwah: are you sure it should be array??
cause you ask here for  Comma separated ....

Integrations/integration-McAfee-MAR.yml
```diff
@@ -0,0 +1,332 @@
...
+  required: true
+- display: Client certificates content (see `client.crt` in instructions)
+  name: cert_file
+  defaultvalue: ""
+  type: 12
+  required: true
+- display: Client private key content (see `client.key` in instructions)
+  name: private_key
+  defaultvalue: ""
+  type: 12
```
meirwah: you may want to change this to type 14
Which is ConfigDataTypeTextAreaEncrypted

Integrations/integration-McAfee-MAR.yml
```diff
@@ -0,0 +1,331 @@
...
+
+    def search_result_to_table(search_result):
+        items = search_result['items']
+        table = [extract_item_output(item) for item in items]
+        return table
+
+    def search(collector, outputs, filter_by, filter_operator, filter_value):
+        config = get_client_config()
+        with DxlClient(config) as client:
+            x = client.connect()
```
Spivaka: x??

asafshen:
Waiting for our MAR instance to enabled again so we can add the playbook test

Integrations/integration-McAfee-MAR.yml
asafshen: x mark the spot

PR #860, opened 2017-09-25
Update UnzipFile To Use fileResult

Bars92:
fixes demisto/server#7577


Bars92:
Approved work with docker and without.

PR #861, opened 2017-09-25
ExportToCSV - wrap values where needed

gshriki:
Wrap values that have commas with quotes

### DIFF
--- a/Scripts/script-ExportToCSV.yml
+++ b/Scripts/script-ExportToCSV.yml
@@ -41,7 +41,7 @@ script: |-
               }          
           }
 
-          str += line + '\r\n';
+          str += line + '\n';
       }
 
       return str;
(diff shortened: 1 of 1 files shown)


Scripts/script-ExportToCSV.yml
```diff
@@ -32,7 +32,13 @@ script: |-
                   line += ',';
               }
 
-              line += array[i][header];
+              var value = array[i][header] + '';
+              value = value.trim();
+              if (value && value.indexOf(',') > -1) {
+                  line += '"' + value + '"';
```
aviadl: There are more csv escaping rules you need to follow:
https://tools.ietf.org/html/rfc4180

Scripts/script-ExportToCSV.yml
```diff
@@ -32,7 +32,13 @@ script: |-
...
+              } else {
+                  line += value;
+              }          
           }
 
           str += line + '\r\n';
```
aviadl: why do we need \r as well ?

PR #868, opened 2017-09-26
add more tests for splunk

meirwah:
add DumpJSON script


TestPlaybooks/playbook-TestSplunk.yml
```diff
@@ -1,14 +1,14 @@
 id: Splunk-Test
-version: -1
+version: 8
```
asafshen: shouldn't it be `-1`?

PR #875, opened 2017-10-01
Clrear release notes

asafshen:
clear release notes & update circle sha1


circle.yml
```diff
@@ -8,7 +8,7 @@ test:
     - ./Tests/yamlvalidator
 
     # build content descriptor
-    - ./setContentDescriptor.sh $CIRCLE_BUILD_NUM aed8890ca84a0c3d64900d2dc565b9fe60d15fab 3.0.0
```
meirwah: 3.0.0 should change too?

asafshen: no, we update this every version branch (same as server)..

I actually thought about this a bit..I think it will be more notable for us (developers) that it need change. and also less confusing for user if the next RN will be 3.0.0 than that last version..

PR #877, opened 2017-10-02
added sentinelone byoi integration

shaniacht1:
fixes https://github.com/demisto/server/issues/6759


Integrations/integration-SentinelOne.yml
```diff
@@ -0,0 +1,2709 @@
...
+            for (var arg in commadArgs) {
+                if (queryMap[arg]) {
+                    queryArgs[arg] = commadArgs[arg];
+                    delete commadArgs[arg];
+                }
+            return queryArgs;
+            }
+        }
+
+        var sendRequest = function(fullurl, method, headers, body) {
```
Spivaka: we should take care of logout in case of failure and I am not sure we are doing that. please verify

shaniacht1: i will add it tonight

NeelRust: @shaniacht1 Change category to `Endpoint` - One word

PR #880, opened 2017-10-08
show in table values which aren't null/undefined

shaniacht1:


Scripts/script-CommonServer.yml
```diff
@@ -2,6 +2,7 @@ commonfields:
   id: CommonServer
   version: 1
 name: CommonServer
+releaseNotes: "Fixed display of false values in TblToMD"
```
dorsha: Change the text in the release notes to tableToMarkdown

Scripts/script-CommonServer.yml
```diff
@@ -206,7 +207,12 @@ script: |-
           t.forEach(function(entry){
               var vals = [];
               headers.forEach(function(h){
-                    vals.push((!!entry[h] || entry[h]===0) ? stringEscapeMD(formatCell(entry[h], cellDelimiter), true) : ' ');
+                  if(entry[h] === null || entry[h] === undefined){
+                      vals.push(' ');
+                  }
+                  else {
```
dorsha: else should be in the same line of the closing (line 212).
Also, space should be added after the if and before the {

PR #883, opened 2017-10-08
OpsGenie Integrations

gshriki:
OpsGenie new integration to allow getting on call users from specific schedules


Integrations/integration-OpsGenie.yml
```diff
@@ -0,0 +1,141 @@
...
+detaileddescription: OpsGenie integration allows querying for the current on-call
+  users for a given Schedule. The integration requires an API key, that can
+  be acquired from the OpsGenie interface, under Integrations / API.
+configuration:
+- display: API Key
+  name: apiKey
+  defaultvalue: ""
+  type: 4
+  required: true
+- display: User system proxy configuration
```
Spivaka: insecure?
+ usually the order is
server url
api key
proxy and insecure

Integrations/integration-OpsGenie.yml
```diff
@@ -0,0 +1,141 @@
...
+            params.proxy
+        );
+
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            logError('OpsGenie error: Request failed for ' + requestUrl + '. Returned body is ' + JSON.stringify(res));
+            throw 'OpsGenie: Request Failed.\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res) + '.';
+        }
+
+        var jsonBody = '';
+        try {
```
Spivaka: better to just return inside the try

Integrations/integration-OpsGenie.yml
```diff
@@ -0,0 +1,141 @@
...
+            sendRequest(url);
+            return 'ok';
+
+        case 'opsgenie-get-on-call':
+            return getOnCall(args.schedule);
+
+        case 'opsgenie-get-user':
+            return getUser(args.userID);
+
+        default:
```
Spivaka: can be removed, can't happen

Integrations/integration-OpsGenie.yml
```diff
@@ -0,0 +1,141 @@
...
+        baseURL += '/';
+    }
+
+    var serviceURLs = {
+        schedules: 'schedules',
+        account: 'account',
+        users: 'users'
+    };
+
+    var sendRequest = function(url, param) {
```
Spivaka: you can remove params here as its not passed inside the call but a global

Integrations/integration-OpsGenie.yml
gshriki: it's a cloud service, so no need insecure. will change order.

Integrations/integration-OpsGenie.yml
gshriki: I think it's better to keep so if someone adds a command without adding the case, it's easier to see.

Integrations/integration-OpsGenie.yml
gshriki: This is `param` and not `params`. it's a way to pass parameters to the API. 
But I'm not using it with the current APIs implemented, so removed,.

Integrations/integration-OpsGenie.yml
gshriki: done

gshriki:
@Spivaka - ping.

PR #885, opened 2017-10-08
Archer adding group parser

maiorrulz:


Integrations/integration-ArcherRSA.yml
```diff
@@ -434,16 +434,32 @@ script:
...
+                        group.forEach(function (gr) {
+                            groupValue = groupValue ? groupValue + ',' + gr['-name'] : gr['-name']
+                        })
+                    }
+                }
+                var toPush = {
                     FieldId: field['-id'],
                     FieldName: mapping[field['-id']].Name,
                     FieldType: field['-type'],
-                    Value: field['-value'],
```
dorsha: Keep the line and change it to:
`Value: field[groupValue || '-value']`

Integrations/integration-ArcherRSA.yml
```diff
@@ -434,16 +434,32 @@ script:
...
                     LevelId: field['-id']
-                });
+                }
+                if (groupValue) {
```
dorsha: remove these lines

Integrations/integration-ArcherRSA.yml
maiorrulz: Did you mean - ` Value: groupValue || field['-value'] ` ?

dorsha: Right

PR #887, opened 2017-10-10
Common server enhancements

yaakovi:
minor changes in the CommonServer:
1) bug fix in function cleanObject.
2) enhancement to createEntry function: broke function interface but no one started using it yet (a new function recently added by Shani)

### DIFF
--- a/Scripts/script-CommonServer.yml
+++ b/Scripts/script-CommonServer.yml
@@ -45,7 +45,7 @@ script: |-
           }
           return res;
       }
-      cleanSingleObject(obj);
+      return cleanSingleObject(obj);
   }
 
   /**
--- a/Scripts/script-CommonServer.yml
+++ b/Scripts/script-CommonServer.yml
@@ -376,14 +376,14 @@ script: |-
    * * dest-key/orig-key should be sparated with '.' (e.g. 'File.Malicious')
    * * dest-key is also used for war room table headers
    * * humanReadable states whether field should appear in war room table. Default is true.
-   * @param {function} headerTransform - transform mark down headers to readable format.
    * @param {function} MDfilter - (optional) filters out fields that shouldn't be displayed in warroom (returns true/false). Default uses humanReadable field.
+   * @param {function} headerTransform - transform mark down headers to readable format.
    * @returns {Object} - the war room entry
   */
-  var createEntry = function(result, translator, headerTransform, MDfilter) {
+  var createEntry = function(result, translator, MDfilter, headerTransform) {
       /* filters out all fields where humanReadable is false */
-      var filter = function(obj){
-          if(obj.humanReadable !== false){
+      var filter = function(obj) {
+          if(obj.humanReadable !== false) {
               return true;
           }
           return false;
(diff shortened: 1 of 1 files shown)


Scripts/script-CommonServer.yml
```diff
@@ -369,17 +376,21 @@ script: |-
    * * dest-key/orig-key should be sparated with '.' (e.g. 'File.Malicious')
    * * dest-key is also used for war room table headers
    * * humanReadable states whether field should appear in war room table. Default is true.
+   * @param {function} headerTransform - transform mark down headers to readable format.
    * @param {function} MDfilter - (optional) filters out fields that shouldn't be displayed in warroom (returns true/false). Default uses humanReadable field.
    * @returns {Object} - the war room entry
   */
-  var createEntry = function(result , translator, MDfilter){
+  var createEntry = function(result, translator, headerTransform, MDfilter) {
```
dorsha: it might break compatibility with previous versions, you should add the `headerTransform` argument at the end of the function (even it seems no one is using it)

Scripts/script-CommonServer.yml
```diff
@@ -391,7 +402,7 @@ script: |-
       var content = translator.innerPath ? dq(result, translator.innerPath) : result;
       var translatedContext = mapObjFunction(translator.data) (content);
       var translatedMD = mapObjFunction(translator.data, filter) (content);
-      entry.HumanReadable = tableToMarkdown(translator.title,translatedMD, undefined, undefined, dotToSpace);
+      entry.HumanReadable = tableToMarkdown(translator.title, translatedMD, undefined, undefined, headerTransform);
```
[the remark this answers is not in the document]
dorsha: Just need to do:
`headerTransform = dotToSpace` at the function signature.

Scripts/script-CommonServer.yml
yaakovi: i think it should be before MDfilter because it will be more usable than the MDfilter.
the MDfilter is a function to filter data shown to the user but the basic filtering most users will want could be easier to use with the translator parameter.

in this case, every call the the function will have to skip the MDfilter parameter using undefined to access the headerTransform:
createEntry(result, {...}, undefined, dotToSpace)
instead of
createEntry(result, {...}, dotToSpace)

Scripts/script-CommonServer.yml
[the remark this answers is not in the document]
yaakovi: i get a syntax error doing that, are you sure this version of JS support default values?
i fixed it inside the function for now...

Scripts/script-CommonServer.yml
dorsha: right, but what about backward compatibility? There might be users that already using this function and you might break their usage.

Scripts/script-CommonServer.yml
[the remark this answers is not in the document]
dorsha: I think that is should be supported, but it is ok to do this inside with `||`.
`headerTransform || dotToSpace`

Scripts/script-CommonServer.yml
yaakovi: as i mentioned, no one is yet to use this function (except Shani who is starting now).
otherwise i would have never suggested it.

now it's in mode "speak now or forever hold your peace"

dorsha: How can you be sure about that?
The function was implemented at 31/08, and we released two content versions since then.
It is out there.. and you can't be sure what users are doing, you can guess and estimate, but can be sure 100%.

@gshriki - your call

gshriki: Guys it is not true nobody is using it. We had customer questions and it is also included in one of our help center articles. 
I don’t know where this “speak now...” mode is coming from - but I’m speaking. 
WE SHOHLD NOT BREAK BACKWARD COMPAT. 
And this should be a strict rule which I find strange that I need to say.

yaakovi: got it, forgot this is open-sourced to clients...
changed it

PR #892, opened 2017-10-12
Update integration-ArcSightESM.yml

anara123:
- make proxy of type boolean.

### DIFF
--- a/Integrations/integration-ArcSightESM.yml
+++ b/Integrations/integration-ArcSightESM.yml
@@ -50,7 +50,7 @@ configuration:
 - display: Use system proxy settings
   name: proxy
   defaultvalue: ""
-  type: 0
+  type: 8
   required: false
 - display: Fetch incidents
   name: isFetch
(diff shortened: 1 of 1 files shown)


Integrations/integration-ArcSightESM.yml
```diff
@@ -50,7 +50,7 @@ configuration:
 - display: Use system proxy settings
   name: proxy
   defaultvalue: ""
-  type: 0
+  type: 8
```
dorsha: Change default value to be false or true

Integrations/integration-ArcSightESM.yml
```diff
@@ -325,7 +325,7 @@ script:
                 Body: body
             },
             params.insecure,
-            params.useproxy
+            params.proxy
```
dorsha: Was it never worked?

anara123: It seems not. But because we don't have proxy env so it never tested

PR #893, opened 2017-10-13
Splunkpy edit ne and create jobs

meirwah:


Integrations/integration-SplunkPy.yml
```diff
@@ -57,6 +57,60 @@ script:
...
+        Update some notable events.
+
+        Arguments:
+        sessionKey -- The session key to use
+        comment -- A description of the change or some information about the notable events
+        status -- A status (only required if you are changing the status of the event)
+        urgency -- An urgency (only required if you are changing the urgency of the event)
+        owner -- A nowner (only required if reassigning the event)
+        eventIDs -- A list of notable event IDs (must be provided if a search ID is not provided)
+        searchID -- An ID of a search. All of the events associated with this search will be modified unless a list of eventIDs are provided that limit the scope to a sub-set of the results.
```
asafshen: I have not seen any place that we pass `searchID`

I guess that this code was taken from somewhere?

meirwah: yep, will support in next version of this.. for now, not sure it will be needed

PR #898, opened 2017-10-13
UnzipFile should not read and copy unzipped data

slavikm:
Fixed #https://github.com/demisto/server/issues/9312


Scripts/script-CommonServerPython.yml
```diff
@@ -293,7 +293,14 @@ script: |-
           f.write(data)
       return {'Contents': '', 'ContentsFormat': formats['text'], 'Type': entryTypes['file'], 'File': filename, 'FileID': temp}
       
+
+  def file_result_existing_file(filename):
+      """ Create a file result from an existing file by moving it into the right file name """
+      temp = demisto.uniqueFile()
+      os.rename(filename, demisto.investigation()['id'] + '_' + temp)
```
aviadl: This will not work in some setups

slavikm: why?

PR #902, opened 2017-10-15
Urlscan private submit

dem-evgeny:


Integrations/integration-UrlScan.yml
```diff
@@ -454,4 +490,4 @@ script:
     - contextPath: DBotScore.Score
       description: The actual score
     description: Submit an url to scan
-hidden: false
+releaseNotes: "urlscan-submit now supports private submit."
```
gshriki: submit -> submits

Integrations/integration-UrlScan.yml
```diff
@@ -437,7 +466,14 @@ script:
...
     - name: timeout
       description: How much seconds to wait to the scan id result
-      defaultValue: "10"
+      defaultValue: "30"
+    - name: public
+      auto: PREDEFINED
+      predefined:
+      - "yes"
+      - "no"
+      description: Does the submit will be public or private
```
gshriki: Set submit as public or private

PR #903, opened 2017-10-15
show dbotscore 0 when no result is returned

noaco:
fixes [demisto/server/#9189](https://github.com/demisto/server/issues/9189)

### DIFF
--- a/Integrations/integration-ThreatExchange.yml
+++ b/Integrations/integration-ThreatExchange.yml
@@ -93,14 +93,15 @@ script:
         }
         var res = doReq('GET', 'malware_analyses', argsForCall);
         var data = res.obj.data;
-        if (data.length === 0) {
-            return {Type: entryTypes.note, Contents: res.body, ContentsFormat: formats.json,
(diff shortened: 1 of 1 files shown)


Bars92:
@noaco missing releaseNote

Integrations/integration-ThreatExchange.yml
```diff
@@ -175,7 +179,9 @@ script:
         var data = res.obj.data;
         if (data.length === 0) {
             return {Type: entryTypes.note, Contents: res.body, ContentsFormat: formats.json,
-                HumanReadable: 'ThreatExchange does not have details about ' + url + '\n'};
+                HumanReadable: 'ThreatExchange does not have details about ' + url + '\n',
+                EntryContext: {DBotScore: [{Indicator: url, Type: 'url', Vendor: 'ThreatExchange', Score: 0}]}
+            };
         }
         var ec = {};
```
Bars92: I would use the `ec` array and insert before return so the return would be almost the same for all.

anara123:
@noaco You need to add `releaseNotes` in yml in order the build to pass

PR #910, opened 2017-10-17
Create integration-DigitalShadows.yml

shaniacht1:
fixes https://github.com/demisto/server/issues/9231

Some commands are still missing context


dorsha:
deletedBy_role and requestedBy_role are deifned twice and cause UI errors:

image

image

image

dorsha:
Also, command names are too long:
image

shaniacht1:
@dorsha 
What do you suggest to do with the command names? We used to add the shortened name of the integration but then we stopped. And I'm afraid if I shorten the actual name of the command it will make the usage of the command unclear. 
Ideas?

dorsha:
@NeelRust please suggest what do you want to do here.

rishibhargava:
My recommendation is to shorten the name of brand but have the description start with digital shadows for each command so that search works great in war room. Can we complete this by Sunday. We have an event on Tuesday with digital shadows.

Regards,
Rishi

dorsha:
Commands names should start with: `ds-`

PR #925, opened 2017-10-20
Allow conversion from XML files

slavikm:
Simple copy from existing ConvertXmlToJson but from files

### DIFF
--- a/Scripts/script-ConvertXmlFileToJson.yml
+++ b/Scripts/script-ConvertXmlFileToJson.yml
@@ -33,4 +33,5 @@ args:
   - "True"
   - "False"
   description: Print json result to warrom
+  defaultValue: "True"
 scripttarget: 0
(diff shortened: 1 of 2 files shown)


Scripts/script-ConvertXmlFileToJson.yml
```diff
@@ -0,0 +1,36 @@
...
+  var xmlJson = x2j(xml);
+  try {
+    xmlJson = JSON.parse(xmlJson);
+  } catch(err) {}
+
+  if (args.contextKey && args.contextKey.length > 0) {
+      setContext(args.contextKey,xmlJson);
+  }
+  if (args.verbose && args.verbose.length > 0 && args.verbose == 'True'){
+      return xmlJson;
```
Spivaka: I think we should set a default value to this otherwise running it without anything wouldn't return results

slavikm: Yeah, I think the same but the original script did not so I thought maybe there is some smarts here I don't understand...

slavikm: Will fix both scripts

slavikm:
@Spivaka ok?

PR #926, opened 2017-10-22
QRadar fixes

Spivaka:
Add get closing reasons command and require close reason id when closing a qradar offense
fixes https://github.com/demisto/server/issues/9464


Integrations/integration-QRadar.yml
```diff
@@ -408,5 +419,28 @@ script:
...
+      - "true"
+      - "false"
+      description: If true, deleted closing reasons are included in the response
+    - name: fields
+      description: Use this parameter to specify which fields you would like to get
+        back in the response
+    - name: filter
+      description: This parameter is used to restrict the elements in a list base
+        on the contents of various fields
+    - name: Range
```
meirwah: Really needed ? Range?

Spivaka: it is part of the api... not sure its needed

PR #927, opened 2017-10-22
CyberArk commands fixes & password encrypted

Bars92:


Integrations/integration-cyberark_AIM.yml
```diff
@@ -179,7 +179,7 @@ script:
     };
 
     var getAccountDetails = function(args) {
-        var result = sendRequestWithToken('GET', credentialsUrl, args, true);
+        var result = sendRequestWithToken('GET', credentialsUrl, args);
```
orius123: Y did U remove it again? I use the error message to parse what account are there in the folder.

Integrations/integration-cyberark_AIM.yml
```diff
@@ -256,6 +256,13 @@ script:
             delete args.keywords;
             return getAccountDetails(args);
         default:
+            args.AppID = appid;
```
orius123: if it's the default case you can delete the `fetch-incidents` one. nope?

Integrations/integration-cyberark_AIM.yml
Bars92: Thats `sendRequestWithToken` last arg is used to get raw data or parsed. since `reset-credentials` does not return data.

Integrations/integration-cyberark_AIM.yml
Bars92: good point!

Integrations/integration-cyberark_AIM.yml
orius123: ok.

Integrations/integration-cyberark_AIM.yml
Bars92: @orius123 `fetch-credentials` has false on HumanReadable is it necessary?

orius123: don't think so but need to validate.

PR #932, opened 2017-10-23
iDefense: Added new integration

yaakovi:
Fixes: demisto/server#9269


Integrations/integration-iDefense.yml
```diff
@@ -0,0 +1,411 @@
...
+  name: url
+  defaultvalue: https://api.intelgraph.verisign.com/
+  type: 0
+  required: true
+- display: API Token
+  name: api_token
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Do not validate server certificate (insecure)
```
Spivaka: proxy?

Integrations/integration-iDefense.yml
```diff
@@ -0,0 +1,411 @@
...
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+script:
+  script: |-
+    // handle '/' at the end of the url
+    var base_url = params.url.slice(0, params.url.length - params.url.match('/*$')[0].length) + '/rest/';
+
+    function mergeObjects(objs) {
```
Spivaka: move to common

Integrations/integration-iDefense.yml
```diff
@@ -0,0 +1,411 @@
...
+        case 'ip':
+            return get_ip(args.ip);
+        case 'url':
+            return get_url(args.url);
+        case 'uuid':
+            // still having issues with this command.
+            return get_uuid(args.uuid);
+        // This is the call made when pressing the integration test button.
+        case 'test-module':
+            return 'ok';
```
Spivaka: unless I missed something, doesn't seem that you really check something here before returning ok

Integrations/integration-iDefense.yml
yaakovi: what do you mean?

Spivaka: we have a flag to use system demisto proxy ettings or general one. you can see it at almost all of other integrations

Spivaka:
you have conflicts, resolve them and merge

PR #939, opened 2017-10-27
Opsgenie additional commands

gshriki:
add opsgenie-get-schedules and opsgenie-get-schedule-timelines commands to get further information from opsgenie


dorsha:
@gshriki lets go over this together tomorrow

Integrations/integration-OpsGenie.yml
```diff
@@ -38,8 +38,9 @@ script:
         users: 'users'
     };
 
-    var sendRequest = function(url) {
-        var requestUrl = baseURL + url;
+    var sendRequest = function(url, param) {
```
dorsha: you added the `param` parameter as second argument.
I checked all `sendRequest` calls, and it seems that you never pass the second argument... am I missing something? If not in use, please remove.

Integrations/integration-OpsGenie.yml
```diff
@@ -63,24 +64,30 @@ script:
...
-    var getOnCall = function(schedule) {
-        var url = serviceURLs.schedules + '/' + encodeURIComponent(schedule) + '/on-calls?scheduleIdentifierType=name';
+    var getOnCall = function(schedule, date) {
+        var urlParams = {scheduleIdentifierType:'name'};
+        if (date) {
+            urlParams.date = new Date(date).toISOString();
+        }
+        var url = serviceURLs.schedules + '/' + encodeURIComponent(schedule) + '/on-calls' + encodeToURLQuery(urlParams);
+
+        console.log(url);
```
dorsha: remove this console.log

Integrations/integration-OpsGenie.yml
```diff
@@ -93,27 +100,65 @@ script:
...
+        } else {
+            md = 'No schedules\n';
+        }
+        return {Type: entryTypes.note, Contents: res.data, EntryContext: ec, ContentsFormat: formats.json, HumanReadable: md};
+    };
+
+    var getScheduleTimeline = function(schedule) {
+        var url = serviceURLs.schedules + '/' + schedule + '/timeline?identifierType=name';
+        var res = sendRequest(url);
+        var rotations = res.data.finalTimeline.rotations;
```
dorsha: you are checking below for `(res && res.data)` but you access `res.data` without checking.. you might have issue here. 
Just move this line inside the if statement

dorsha:
@gshriki let's fix and merge. should be quick.

Integrations/integration-OpsGenie.yml
gshriki: you are right. had it FFU, but no need for now.

Integrations/integration-OpsGenie.yml
gshriki: removed

Integrations/integration-OpsGenie.yml
gshriki: yep... fixed.

PR #941, opened 2017-10-27
CrowdStrike FalconHost UploadIOC fix

gshriki:
Fixing upload IOC command.
Please do not merge yet.

### DIFF
--- a/Integrations/integration-CrowdStrikeHost.yml
+++ b/Integrations/integration-CrowdStrikeHost.yml
@@ -94,7 +94,7 @@ script:
     };
 
     var uploadIOC = function() {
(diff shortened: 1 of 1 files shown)


Integrations/integration-CrowdStrikeHost.yml
```diff
@@ -94,6 +94,7 @@ script:
     };
 
     var uploadIOC = function() {
+        args.expiration_days = Number(args.expiration_days)
```
slavikm: parseInt?

gshriki: not sure it makes a difference, but fixed...

PR #943, opened 2017-10-29
ContextSearchForString

Itay4:
Regarding issue https://github.com/demisto/server/issues/9404
Note that it depends on fix done to flattenFields in https://github.com/demisto/content/pull/936


Scripts/ContextSearchForString.yml
```diff
@@ -0,0 +1,45 @@
+commonfields:
+  id: bb2d479b-2bd4-40cb-84ac-9d7b620f060e
```
aviadl: ID incorrect

Scripts/ContextSearchForString.yml
```diff
@@ -0,0 +1,45 @@
...
+  version: 72
```
aviadl: Version incorrect

Scripts/ContextSearchForString.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  return searchString(contextObject);
+
+
+type: javascript
+tags: []
+comment: |
+  Searches for string in a path in context. If path is null, string will be searched in full context.
+enabled: true
+args:
```
aviadl: Missing output

Scripts/ContextSearchForString.yml
dorsha: id should be the same as the name

Scripts/ContextSearchForString.yml
dorsha: version should be -1

Scripts/script-ContextSearchForString.yml
```diff
@@ -0,0 +1,51 @@
...
+script: |+
+  var path = args.path;
+  var str = args.str;
+  var flat={};
+
+  function searchString(obj) {
+      flattenFields(obj,undefined,flat);
+      keysArr = Object.keys(flat);
+      for (i = 0; i < keysArr.length - 1; i++) {
+          if (flat[keysArr[i]] == str) {
```
Spivaka: ===

Spivaka: read about differences between == and === in javascript 
https://stackoverflow.com/questions/523643/difference-between-and-in-javascript

PR #944, opened 2017-10-29
Revert "Hypersearch script deprecation"

dorsha:
Reverts demisto/content#942

fixes: https://github.com/demisto/server/issues/9559


aviadl:
These should be deprecated with our new architecture !

dorsha:
Yes, but you merged it before it was merged and cause the build to fail.
Next time should be merged after the server merged.

PR #949, opened 2017-10-30
Clear release notes

asafshen:
Clear version 3.0.4 release notes

Please also merge to master


circle.yml
```diff
@@ -8,7 +8,7 @@ test:
     - ./Tests/yamlvalidator
 
     # build content descriptor
-    - ./setContentDescriptor.sh $CIRCLE_BUILD_NUM d9d2de23493ea67fba166ac1b130d412cf1b97ff 3.0.0
+    - ./setContentDescriptor.sh $CIRCLE_BUILD_NUM 606050fdb73897821ed2f48d67686fe424e95247 3.0.0
```
orius123: don't we need to update the version?

dorsha: @asafshen ?

asafshen: No. We have [3.0.4 branch](https://github.com/demisto/content/tree/3.0.4) with an update version (same treat as our server repo).

PR #950, opened 2017-10-30
IsItPhishing integration

Itay4:
New IsItPhishing integration
Fixes https://github.com/demisto/server/issues/9528

### DIFF
--- a/Integrations/integration-IsItPhishing.yml
+++ b/Integrations/integration-IsItPhishing.yml
@@ -8,7 +8,7 @@ image: data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAHgAAAATCAYAAABbV8lLAAAAAXN
 description: Collaborative web service that provides validation on whether a URL is
   a phishing page or not by analyzing the content of the webpage
 configuration:
-  - display: Server URL (e.g. https://192.168.0.1)
+- display: Server URL (e.g. https://192.168.0.1)
   name: url
   defaultvalue: https://ws.isitphishing.org
   type: 0
--- a/Integrations/integration-IsItPhishing.yml
+++ b/Integrations/integration-IsItPhishing.yml
@@ -8,21 +8,21 @@ image: data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAHgAAAATCAYAAABbV8lLAAAAAXN
 description: Collaborative web service that provides validation on whether a URL is
   a phishing page or not by analyzing the content of the webpage
 configuration:
-- display: Customer's name
-  name: name
-  defaultvalue: ""
-  type: 12
-  required: true
-- display: Customer's License
-  name: license
-  defaultvalue: ""
-  type: 12
-  required: true
-- display: Server URL
+  - display: Server URL (e.g. https://192.168.0.1)
   name: url
   defaultvalue: https://ws.isitphishing.org
   type: 0
   required: false
+- display: Customer's name
+  name: name
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Customer's License
+  name: license
+  defaultvalue: ""
+  type: 0
+  required: true
 - display: Use system proxy settings
   name: proxy
   defaultvalue: "true"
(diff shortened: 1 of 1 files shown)


Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+  type: 8
+  required: false
+- display: Do not validate server certificate (insecure)
+  name: insecure
+  defaultvalue: "false"
+  type: 8
+  required: false
+script:
+  script: |+
+    var auth = 'Bearer ' + Base64.encode(params.name + ':' + params.license);
```
Spivaka: please use btoa (or atob, nut sure which way) instead of this base 64 encode which is very slow

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+description: Collaborative web service that provides validation on whether a URL is
+  a phishing page or not by analyzing the content of the webpage
+configuration:
+- display: Customer's name
```
Spivaka: are those fields that long they need the long text type?
and usually, URL is first

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+  name: name
+  defaultvalue: ""
+  type: 12
+  required: true
+- display: Customer's License
+  name: license
+  defaultvalue: ""
+  type: 12
+  required: true
+- display: Server URL
```
Spivaka: provide an example in the display name (you may see other integrations that do so like wildfire)

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+  defaultvalue: "false"
+  type: 8
+  required: false
+script:
+  script: |+
+    var auth = 'Bearer ' + Base64.encode(params.name + ':' + params.license);
+
+    var sendRequest = function(method, api, body) {
+        var url = params.url;
+        if (url[url.length - 1] === '/') {
```
Spivaka: while will take care of more than one / at the end of the string.
you can also use regex to eliminate them

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+            url = url.substring(0, url.length - 1);
+        }
+
+        var requestUrl = url + '/' + api;
+        var res = http(
+            requestUrl + encodeToURLQuery(body),
```
Spivaka: rename body as this is not body but url query

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            throw 'Request Failed.\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res) + '.';
+        }
+
+        return res;
+    };
+
+    var isPhishing = function(url, force, smart, area, timeout) {
+        var body = {
+            name: params.name,
```
Spivaka: the name and license are send twice, once in header and once in the url args?

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+      description: Set true to force checks on URLs that may cause collateral damage
+        to the end user, or false to ignore the argument
+      defaultValue: "true"
+    - name: area
+      description: The regional area to force using a proxy
+    - name: timeout
+      description: Timeout in milliseconds. Default value set to 10000, with a minimum
+        value of 1000. Once timeout is reached, TIMEOUT response is returned
+    outputs:
+    - contextPath: URL was identified as spam.
```
Spivaka: I don't think you use contextPath correctly here. @shaniacht1 please guide him

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+        if (!area) {
+            delete body.area;
+        }
+        if (!timeout) {
+            delete body.timeout;
+        }
+        var res = sendRequest('POST', 'check', body);
+        var resBody = res.Body.trim();
+        switch (resBody){
+            case 'SPAM':
```
Spivaka: you need to push to context, a simple string is not usable for the next task

Integrations/integration-IsItPhishing.yml
```diff
@@ -0,0 +1,166 @@
...
+    - name: smart
+      description: Set true to force checks on URLs that may cause collateral damage
+        to the end user, or false to ignore the argument
+      defaultValue: "true"
+    - name: area
+      description: The regional area to force using a proxy
+    - name: timeout
+      description: Timeout in milliseconds. Default value set to 10000, with a minimum
+        value of 1000. Once timeout is reached, TIMEOUT response is returned
+    outputs:
```
Spivaka: outputs should document what you are pushing to context. so you first need to push something to context and than document it...

PR #953, opened 2017-10-30
splunk fetch configure time zone support

meirwah:


orius123:
why +180 and not +3 and we will multiply by 60?
and you should validate so we wouldn't get exceptions IMHO

aviadl:
Because there are time zones in half hours as well

orius123:
Than you can treat it as a float. it will work with 180 it's just less clear

aviadl:
This is how our system works in all places

Spivaka:
@meirwah are you going to merge this?

PR #960, opened 2017-11-02
DeleteContext update

erezh31:
change delete context to throw an error when no key or all=yes specified


dorsha:
Missing release notes

PR #961, opened 2017-11-02
Fix random filename variable

slavikm:


gshriki:
It's deprecated...

slavikm:
Yup, it's deprecated in favor of this one.

dorsha:
Still it has a bug.... but ok

dorsha:
And other people can copy this bug to other scripts (like slavik did)

slavikm:
@dorsha I deprecated it and started with the code from there - don't think anyone else will do it.

dorsha:
ok, lets leave it to the fate 🤞

PR #967, opened 2017-11-05
3.1.0

asafshen:
Update version number in content release to 3.1.0
(this version will not be release. only for demisto 3.1.0 release)

@dorsha we need to make sure will take content for master please

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,7 +8,7 @@ test:
     - ./Tests/yamlvalidator
 
     # build content descriptor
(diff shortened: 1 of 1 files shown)


gshriki:
What will be the experience for customers with 3.0.1 or older versions?

asafshen:
@gshriki nothing
They won't get any content update,  since we wont publish a release
(Only master version will be affected, this will be used when build 3.1.0)

aviadl:
@asafshen but we will release in a week , what will you do then

asafshen:
@aviadl we will update version number before
even if not - we will have the same problem we have today (costumers with 2.5/2.6 still have content update - isn't them?)

PR #974, opened 2017-11-06
ATDBug fix - returns an error on unsupported files

JonathanMeler:
fixes https://github.com/demisto/server/issues/9166#issuecomment-342090977

### DIFF
--- a/Scripts/script-ATDDetonate.yml
+++ b/Scripts/script-ATDDetonate.yml
@@ -2,6 +2,7 @@ commonfields:
   id: ATDDetonate
   version: -1
 name: ATDDetonate
+releaseNotes: 'Returns an error on unsupported files'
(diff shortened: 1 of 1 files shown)


Scripts/script-ATDDetonate.yml
```diff
@@ -33,6 +34,11 @@ script: |-
       else:
           demisto.results({ "Type" : entryTypes["error"], "ContentsFormat" : formats["text"], "Contents" : "Coudn't extract TaskID from upload" })
           sys.exit(0)
+
+  if taskID == "-1":
```
Spivaka: can happen for other reasons?

JonathanMeler: That's the meaning according to the API

PR #984, opened 2017-11-10
Executed commands script

gshriki:
Script to list all executed commands in war room - by playbook, by user, or both

@NeelRust FYI


Scripts/script-listExecutedCommands.yml
```diff
@@ -0,0 +1,64 @@
+commonfields:
+  id: listExecutedCommands
+  version: -1
+name: listExecutedCommands
+script: |-
+  var fixNewLines = function(str) {
+      var tmp = str.split('\n').join('<br>');
```
dorsha: You can just do 
`return str.split('\n').join('<br>').split('\r').join(' ');`

Also, why not using replace?

Scripts/script-listExecutedCommands.yml
```diff
@@ -0,0 +1,64 @@
...
+  }
+  if (userCommands.length > 0) {
+      md += tableToMarkdown('User Executed Commands', userCommands, ['Time', 'Entry ID', 'User', 'Command']) + '\n';
+  }
+  if (md === '') {
+      md = 'No commands found\n';
+  }
+  return {ContentsFormat: formats.markdown, Type: entryTypes.note, Contents: md};
+type: javascript
+tags: []
```
dorsha: Do you want to tag it?

Scripts/script-listExecutedCommands.yml
```diff
@@ -0,0 +1,64 @@
...
+comment: Lists  executed commands in War Room
+enabled: true
+args:
+- name: source
```
dorsha: missing description for source

TestPlaybooks/playbook-listExecutedCommands-Test.yml
```diff
@@ -0,0 +1,243 @@
...
+      scriptName: AreValuesEqual
+      type: condition
+      iscommand: false
+      brand: ""
+    nexttasks:
+      "no":
+      - "8"
+    scriptarguments:
+      left: ${withData=val.trim()}
+      right: No commands found
```
orius123: Shouldn't it find results in this case (The playbook `Print` task)?

gshriki: So first time I'm calling it in the playbook, i use the source=manual, so it shouldn't find. The second time I run it, I use source=playbook. and the playbook checks those conditions.

orius123: and you navigate only on the "no" condition. saw it just now.

dorsha:
@gshriki please resolve conflict, and merge or handle this PR.

dorsha:
@gshriki please merge from master and merge

PR #993, opened 2017-11-14
Test multi integration

asafshen:
fixes: https://github.com/demisto/server/issues/9784
Support playbook that requires multiple integrations
Also changing Wiki (and review (very shortly) on next demo)

### DIFF
--- a/Tests/test_content.py
+++ b/Tests/test_content.py
@@ -115,7 +115,8 @@ def main():
 
         test_message = 'playbook: ' + playbook_id
         if integrations:
-            test_message = test_message + ' with integration(s): ' + ','.join(integrations_conf)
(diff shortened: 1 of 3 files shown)


Tests/test_content.py
```diff
@@ -67,51 +67,67 @@ def main():
...
-        if 'params' in integration:
-            integration_params = integration['params']
-        else:
-            # get from secret conf
-            secret_integration_match = [item for item in secret_integrations if item["name"] == integration_name]
-            if len(secret_integration_match) > 0:
-                integration_params = secret_integration_match[0].get('params')
+        playbook_id = t['playbookID']
+
+        integrations_conf = t.get('integrations', None)
```
orius123: You can change the default to an empty array, instead of the `if ... is None`

PR #998, opened 2017-11-15
ContextGetPathForString script

Itay4:
Fixes https://github.com/demisto/server/issues/9404

### DIFF
--- a/Scripts/script-ContextGetPathForString.yml
+++ b/Scripts/script-ContextGetPathForString.yml
@@ -1,0 +1,46 @@
+commonfields:
+  id: ContextGetPathForString
+  version: -1
+name: ContextGetPathForString
+script: |+
+  var path = args.path
+  var str = args.str;
+  var flat = {};
+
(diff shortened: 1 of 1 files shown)


Scripts/script-ContextGetPathForString.yml
```diff
@@ -0,0 +1,46 @@
...
+name: ContextGetPathForString
+script: |+
+  var path = args.path
+  var str = args.str;
+  var flat = {};
+
+  function searchString(obj) {
+      flattenFields(obj,undefined,flat);
+      keysArr = Object.keys(flat);
+      for (i = 0; i < keysArr.length - 1; i++) {
```
orius123: why `keysArr.length - 1` and not `keysArr.length`?
It will not look for the last key. right?

orius123:
@asafshen do we do tests for this kind of scripts?

asafshen:
@orius123 we can and we should
(@Itay4 you can run playbook no integrations dependencies)

PR #999, opened 2017-11-16
Added a task to pretty print JIRA issue

dorsha:
Includes ticket details, status, timeline, attachments and comments.

image

### DIFF
--- a/Scripts/script-JIRAPrintIssue.yml
+++ b/Scripts/script-JIRAPrintIssue.yml
@@ -5,7 +5,7 @@ name: JIRAPrintIssue
 script: |-
   var response = executeCommand('jira-get-issue', { issueId: args.issueId });
 
-  if (!response || response.length === 0 || !response[0].Contents) {
+  if (!response || response.length === 0 || !response[0].Contents || isError(response[0])) {
       return 'JIRA issue not found (' + args.issueId + ')';
   }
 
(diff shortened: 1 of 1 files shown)


Scripts/script-JIRAPrintIssue.yml
```diff
@@ -0,0 +1,82 @@
...
+      });
+  }
+
+  return { Type: entryTypes.note, Contents: issue, ContentsFormat: formats.json, HumanReadable: md };
+type: javascript
+tags: []
+comment: Pretty print JIRA issue into the incident war room
+enabled: true
+args:
+- name: issueId
```
asafshen: Shouldn't it be required arg?

Scripts/script-JIRAPrintIssue.yml
```diff
@@ -0,0 +1,82 @@
...
+      comments.map(function(comment, i) {
+          md += '\n**Comment ' + (i + 1) + '** From: ' + comment.author.displayName + ' (' + comment.author.emailAddress + ') - ' + comment.created + ':\n';
+          md += comment.body;
+          md += '\n\n --- \n\n';
+      });
+  }
+
+  return { Type: entryTypes.note, Contents: issue, ContentsFormat: formats.json, HumanReadable: md };
+type: javascript
+tags: []
```
asafshen: we have `jira` tag. I think we should is it

Scripts/script-JIRAPrintIssue.yml
```diff
@@ -0,0 +1,82 @@
+commonfields:
+  id: JIRAPrintIssue
+  version: -1
+name: JIRAPrintIssue
+script: |-
+  var response = executeCommand('jira-get-issue', { issueId: args.issueId });
+
+  if (!response || response.length === 0 || !response[0].Contents) {
+      return 'JIRA issue not found (' + args.issueId + ')';
```
asafshen: I ran it with random Id and it did not catch it
instead it throws `Cannot access member 'summary' of undefined` (probably from line 18)

dorsha:
@asafshen done.

PR #1001, opened 2017-11-16
Auto enable phishtank & urlscan

aviadl:


gshriki:
Wow. didn't know it's all part of the script itself... great!!!

Spivaka:
as long as default values are correct

PR #1002, opened 2017-11-18
Handle stringify error in qradar

Spivaka:
return response so it can be analyzed


Integrations/integration-QRadar.yml
```diff
@@ -448,4 +453,4 @@ script:
       description: Number of results in return
     description: Get closing reasons
   isfetch: true
-releaseNotes: 'Adding proxy config support for Qradar'
\ No newline at end of file
+releaseNotes: 'Printing message body in case of stringify error'
```
meirwah: don't you need to add to releaseNotes that was there? instead of override?

dorsha: No, and tests failed because of macafee, already opened issue

Spivaka:
@meirwah / @dorsha please approve and merge

PR #1007, opened 2017-11-19
Adding file detonate to VxStream

Itay4:
Fixes https://github.com/demisto/server/issues/9748


Integrations/integration-VxStream.yml
```diff
@@ -284,6 +288,30 @@ script:
         return res;
     }
 
+    function detonateFile(entryId, environmentId) {
+
+        var hash = submitFile(entryId, environmentId).substring(10);
+
+        var delay = 60000;
```
slavikm: Delay and total wait time should be arguments to the command. You can provide sensible defaults but why hardcode?

Integrations/integration-VxStream.yml
```diff
@@ -284,6 +288,33 @@ script:
...
+
+        sleep(delayTime);
+
+        while (waitTime<timeOut) {
+            res = scan(hash);
+            if (res.response[0]) {
+                return scanToEntry(res);
+            } else {
+                waitTime = waitTime + delayTime;
+                sleep(delayTime)
```
Spivaka: we have a wait function in jshttp that doesn't do busy waiting, use it instead of your implementation

Integrations/integration-VxStream.yml
```diff
@@ -284,6 +288,33 @@ script:
...
+            }
+        }
+        return 'Timeout';
```
Spivaka: maybe throw an error?

Integrations/integration-VxStream.yml
```diff
@@ -472,10 +468,10 @@ script:
       required: true
       description: 'environment ID to submit file to(get all via vx-get-environments) '
     - name: delay
-      description: Delay wait time between calls (in milliseconds)
+      description: Delay wait time between calls (in seconds)
       defaultValue: "3000"
```
dorsha: 3000 seconds is the default? You meant 3 seconds?

Integrations/integration-VxStream.yml
```diff
@@ -472,10 +468,10 @@ script:
...
     - name: timeout
-      description: Total wait time (in milliseconds)
+      description: Total wait time (in seconds)
       defaultValue: "60000"
```
dorsha: 60000?

Integrations/integration-VxStream.yml
```diff
@@ -284,6 +288,29 @@ script:
...
+
+        delayTime = parseInt(delay);
+        timeOut = parseInt(timeout);
+        var waitTime = delayTime;
+
+        wait(delayTime);
+
+        while (waitTime<timeOut) {
+            res = scan(hash);
+            if (res.response[0]) {
```
dorsha: I would check also if res.length > 0, so you won't get out of range exception/ 
`if (res.response.length > 0 && res.response[0])`

Integrations/integration-VxStream.yml
```diff
@@ -428,3 +459,19 @@ script:
...
     description: Retrieve result data upon a file. NOTE - This command returns a file
       (also - minimum required authorization is `default`)
+  - name: vx-detonate-file
+    arguments:
+    - name: entryId
+      required: true
+      description: War-room entry ID of sample file
+    - name: environmentID
+      required: true
+      description: 'environment ID to submit file to(get all via vx-get-environments) '
```
dorsha: Change this:
`'environment ID to submit file to(get all via vx-get-environments) '`
to
`Environment ID to submit file to (get all IDs via vx-get-environments)`

Integrations/integration-VxStream.yml
```diff
@@ -428,3 +459,19 @@ script:
       description: 'environment ID to submit file to(get all via vx-get-environments) '
     description: Retrieve result data upon a file. NOTE - This command returns a file
       (also - minimum required authorization is `default`)
+  - name: vx-detonate-file
+    arguments:
+    - name: entryId
+      required: true
+      description: War-room entry ID of sample file
+    - name: environmentID
+      required: true
```
asafshen: @michalgold we consider to make `environmentID` optional - and if its not provided - it will choose the first environment ID that it finds. What do you think?

michalgold: I think that we could make it optional and when it's not provided use environment ID 100 as default. 
image
If they change this/environment ID 100 is not found, let's use the first windows environment ID we can find.

Spivaka:
@slavikm anything else?

PR #1009, opened 2017-11-21
get demisto license from conf repo

asafshen:
Fixes CircleCI failing build 

Should also be merged with [content-test-conf branch](https://github.com/demisto/content-test-conf/tree/add-lic)

### DIFF
--- a/release_notes.py
+++ b/release_notes.py
@@ -276,6 +276,10 @@ def createFileReleaseNotes(fileName, deleteFilePath):
         names = fileName.split("\t")
         changeType = names[0]
         fullFileName = names[1]
+        fileType = fullFileName.split("/")[0]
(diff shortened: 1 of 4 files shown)


release_notes.py
```diff
@@ -276,6 +276,15 @@ def createFileReleaseNotes(fileName, deleteFilePath):
         names = fileName.split("\t")
         changeType = names[0]
         fullFileName = names[1]
+
```
asafshen: @aviadl @orius123 

When I renamed `download_secret_conf_file.sh` to `download_demisto_conf.sh` - this script failed
python could not open the `.sh` file in `r` mode

I refactor this a bit to skip anything that doesn't have `fileTypeMapping` (anything that is not script/integration/playbook/report

PR #1010, opened 2017-11-21
Support tanium proxy settings

aviadl:
Also added default value for port


dorsha:
@aviadl the tanium playbook fails, maybe related to the param you added, please check.

PR #1012, opened 2017-11-22
Fixed compilation and syntax in CommonServer

dorsha:


dorsha:
It was compilation error, there is no logic here to test.. not sure how it passed any test.

PR #1017, opened 2017-11-22
Human readable json

guydemi:
fixes: https://github.com/demisto/server/issues/9474

Added a utility to format a JS object.

Example:
* The stringified object:
> "{"testField":"testString","testArray":["string",12,[1,2,3],{"foo":"bar"}],"testNum":42,"testObj":{"a":"b","c":4,"e":["f","g",8]}}".

* Represented in JSON:
screen shot 2017-11-22 at 16 26 04

* Will be formatted as:
screen shot 2017-11-22 at 19 04 40


Scripts/script-CommonServer.yml
```diff
@@ -1365,6 +1365,69 @@ script: |-
...
+            res += "  ";
+        }
+        return res;
+    }
+
+    /* Run all object keys and invoke the recursive helper function */
+    Object.keys(obj).forEach(function(key) {
+        var value = obj[key];
+        var str = jsonToHumanReadableRecursive(value);
+        obj[key] = str;
```
dorsha: create another object and return it, don't mutate the input parameter

Scripts/script-CommonServer.yml
```diff
@@ -1365,6 +1365,69 @@ script: |-
...
 
+  /**
+   * Converts all object (and array) values of a given object to human-readable strings
+   * @param {Object} obj - the object to be altered
+   */
+  function jsonToHumanReadable(obj) {
+
+    /* Helper function that rucurses an object, formatting it's values as strings */
+    var jsonToHumanReadableRecursive = function(obj, depth) {
+        var res = "";
```
dorsha: In JavaScript the notation is '' and not "".
Please replace in all places.

Scripts/script-CommonServer.yml
```diff
@@ -1365,6 +1365,69 @@ script: |-
...
+            }
+        });
+
+        return res;
+    }
+
+    /* Helper function that handles the padding of new lines by depth */
+    var indent = function(depth) {
+        var res = "";
+        if (undefined === depth) {
```
dorsha: no need for this condition, just do below:
`for (var i = 0; i < (depth || 0); i++) {`

(It is not good practice to change function arguments).
Do the same in other places.

Scripts/script-CommonServer.yml
dorsha: And the results should be string (not object)

dorsha:
Also:
`jsonToHumanReadable('lala');`
`jsonToHumanReadable(111);`
`jsonToHumanReadable(new Date());`

etc.

Scripts/script-CommonServer.yml
guydemi: As far as cloning the object, I agree!

Regarding the 2nd comment:
The recursive function should always return a string, but the wrapper function needs to return an object (containing the fields as strings) so the it could be converted to markdown.

For example, if in our case a string that represents the whole object is returned, the next function (which is resToMD), will not create the table, but instead create 1 big cell containing the entire string.
Let's look at the original issue's image (https://user-images.githubusercontent.com/3792355/31864214-5a745966-b762-11e7-8e4b-c053f7643aba.png), I've tried to keep it exactly the same, just without the stringified objects (In other words, 1 big object containing string fields), so the same markdown will be applied.

dorsha:
Also make sure to handle null obj

Scripts/script-CommonServer.yml
dorsha: 👍

orius123:
I think the array inside `testArray` should be indented.

orius123:
and wrap indexes with []. i.e. 
or something that will show that `foo: bar` and the array are also vars of  the array.
```
[
string
12
[
  1
  2
  3
]
foo: bar
]
```

dorsha:
@orius123 don't think we need to wrap with [].
If we wanted to be wrapped we could just use the JSON.stringify(obj, null, 2) - which do the same

Scripts/script-CommonServer.yml
```diff
@@ -1365,6 +1365,63 @@ script: |-
...
+   */
+  function jsonToHumanReadable(obj) {
+
+    /* Helper function that rucurses an object, formatting it's values as strings */
+    var jsonToHumanReadableRecursive = function(obj, depth) {
+        var res = '';
+
+        /* Return non objects (or dates) as strings */
+        if (!obj || typeof obj !== 'object' || !Object.keys(obj).length) {
+              return indent(depth || 0) + obj + '\n';  // this will always convert to string type
```
orius123: wouldn't it be easier to just init depth once instead of trying `(depth || 0)` all around?

orius123: if `!obj` and it's undefined wouldn't it print 'undefined'?

guydemi: Yes it will, but that is the original value, the one that is actually there in the object, on purpose or not, I don't think we are here to "fix" it (or hide it, in our case), I think we should reflect the object as is (as far as possible, string-wise...)

orius123: ok.

orius123:
@guydemi can you drop some screenshots after the fixes?

guydemi:
screen shot 2017-11-26 at 13 32 36
screen shot 2017-11-26 at 13 33 48

guydemi:
I've added a new Test Playbook for CommonServer, which currently only tests this utility function.
@dorsha

PR #1018, opened 2017-11-22
ATD bugfixes

Itay4:
Fixes https://github.com/demisto/server/issues/9879


Integrations/integration-ATD.yml
```diff
@@ -499,7 +510,7 @@ script:
       - txt
       - xml
       - zip
-      - json
+      - table
```
dorsha: need to keep json for backward compatibility (for users that already used it in their playbooks)

PR #1019, opened 2017-11-22
add fetch indicator script

erezh31:


Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,114 @@
+commonfields:
+  id: 283b5229-4577-4be4-8a4e-006001543341
```
meirwah: change this to script name FetchAnonymousInfo

Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,114 @@
...
+  version: 85
+name: FetchAnonymousInfo
+script: |-
```
meirwah: add description to script (fetch incident and indicator data annoymized)

Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,114 @@
+commonfields:
+  id: 283b5229-4577-4be4-8a4e-006001543341
+  version: 85
```
meirwah: version should be -1

Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,115 @@
+commonfields:
+  id: FetchAnonymousInfo
+  version: -1
+name: FetchAnonymousInfo
```
orius123: Would change to FetchSystemData anonymous is optional

Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,115 @@
...
+      incident_info = {key:incident[key] for key in INCIDENT_SAFE_FIELDS}
+      for field in INCIDENT_HASH_VALUES:
+          incident_info[field] = hash_object(incident[field])
+
+      if incident['labels'] is not None:
+          incident_info['labels'] = []
+          for label in incident['labels']:
+              incident_info['labels'].append({'type': label['type'], 'value': hash_object(label['value'])})
+      else:
+          incident_info['labels'] = None
```
orius123: Do you have to set it if it doesn't exist?

Scripts/script-FetchAnonymousInfo.yml
```diff
@@ -0,0 +1,115 @@
...
+  cmd_run = False
+  if (demisto.args().has_key('fromDate')):
+      from_date = demisto.args()['fromDate']
+  else:
+      from_date = (datetime.datetime.now().date() - datetime.timedelta(days=365)).isoformat()
+
+  if demisto.args()['incidents'] == 'yes':
+      cmd_run = True
+      res = demisto.executeCommand("getIncidents", {'query': 'created:>"%s"' % from_date})
+      incidents_data = map(parse_incident, res[0]['Contents']['data'])
```
orius123: Don't you need to check that you got results first?
same for indicators.

Scripts/script-FetchAnonymousInfo.yml
erezh31: Not really, but I want it will have the same structure as the original incident object (and there the key exist)

Scripts/script-FetchAnonymousInfo.yml
erezh31: amm I don't think so.. because I think of that as warning when someone use the script with both arguments are no (so he does something wrong). if there is not data it's ok

dorsha:
@meirwah can you approve?

meirwah:
we moved this script to different repo, closing this PR

PR #1020, opened 2017-11-22
ADGetGroupMembers empty field fix

shaniacht1:
Script would break if memberOf field in answer was empty.

### DIFF
--- a/Scripts/script-ADGetGroupMembers.yml
+++ b/Scripts/script-ADGetGroupMembers.yml
@@ -10,7 +10,7 @@ script: |+
           endpoint['Type'] = 'AD'
           endpoint['ID'] = demisto.get(l,'dn')
           endpoint['Hostname'] = demisto.get(l,'name')
-          endpoint['Groups'] = demisto.get(l,'memberOf').split('<br>')
+          endpoint['Groups'] = demisto.get(l,'memberOf').split('<br>') if demisto.get(l,'memberOf') else demisto.get(l,'memberOf')
           for attr in set(argToList(attrs)) - set(['dn','name','memberOf']):
               endpoint[attr.title()] = demisto.get(l,attr)
           endpoints.append(endpoint)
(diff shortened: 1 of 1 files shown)


Scripts/script-ADGetGroupMembers.yml
```diff
@@ -10,7 +11,7 @@ script: |+
           endpoint['Type'] = 'AD'
           endpoint['ID'] = demisto.get(l,'dn')
           endpoint['Hostname'] = demisto.get(l,'name')
-          endpoint['Groups'] = demisto.get(l,'memberOf').split('<br>')
+          endpoint['Groups'] = demisto.get(l,'memberOf').split('<br>') if demisto.get(l,'memberOf') else demisto.get(l,'memberOf')
```
asafshen: whats the point of assigning `Groups` fields in `demisto.get(l,'memberOf')` if its falsely?
I recommend:
1. assign `None` to it
OR
2. Don't assign this field at all:
```
if demisto.get(l,'memberOf'):
    endpoint['Groups'] = demisto.get(l,'memberOf').split('<br>')
```

dorsha: Also, the one liner syntax is very hard to understand

Scripts/script-ADGetGroupMembers.yml
```diff
@@ -2,6 +2,7 @@ commonfields:
   id: ADGetGroupMembers
   version: -1
 name: ADGetGroupMembers
+releaseNotes: "fixed script failing on empty fields"
```
asafshen: please change to `Fixed script failure when user/computer has no groups`

Scripts/script-ADGetGroupMembers.yml
orius123: I think it's readable (this is python ternary operation)
`demisto.get(l,'memberOf').split('<br>') if demisto.get(l,'memberOf') else ''` or `else None`
or
`demisto.get(l,'memberOf') and demisto.get(l,'memberOf').split('<br>')`

dorsha: I know it is python ternary.. but it is still hard to understand.. you can write everything in one liner..

PR #1023, opened 2017-11-23
Bugfix/catch json parse error intsights

Spivaka:
fixes demisto/server#9899


Integrations/integration-IntSights.yml
```diff
@@ -475,14 +484,17 @@ script:
...
         default:
-            result = JSON.parse(
-                sendRequest(
-                    methodDictionary[command],
-                    getUrl(command, args),
-                    createBody(command, args),
-                    loginCookies
-                ).Body
-            );
+            res = sendRequest(
```
orius123: You need to parse first in order to access `.Body`

PR #1025, opened 2017-11-26
MWG: fixed logout in case of exceptions

yaakovi:
fixes: https://github.com/demisto/server/issues/9910


Integrations/integration-McAfeeWebGateway.yml
```diff
@@ -334,3 +359,4 @@ script:
       required: true
       description: Entry Position in table
     description: Delete an entry from a list
+releaseNotes: added force logout in case of errors
```
dorsha: Sentence should start in capital letter

PR #1034, opened 2017-11-30
skip failing tests

asafshen:
related to https://github.com/demisto/server/issues/9985
related to https://github.com/demisto/server/issues/9984

Also fixed issue in `AssignAnalystToIncident` (fallback to roles)

### DIFF
--- a/Scripts/script-AssignAnalystToIncident.yml
+++ b/Scripts/script-AssignAnalystToIncident.yml
@@ -98,3 +98,4 @@ args:
 - name: email
   description: When specified, the user of provided email will be assigned as the incident owner (optional)
 scripttarget: 0
+releaseNotes: "-"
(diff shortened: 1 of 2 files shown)


asafshen:
after consulting with @aviadl, the `AssignAnalistToIncident` issue was probably caused due to a problem in our server
reverted it

dorsha:
@asafshen ?

asafshen:
@dorsha 
image
this playbook fails due to a regression in the server

I can make a patch in `AssignAnalistToIncident` script that will fix this. But after consulting with @aviadl - we decided not to do so.

cannot merge until this fix (because merge is block until build passes)
do you think we should handle it in other way?

dorsha:
No, need to fix the problem.. just make sure we have issue for this.

dorsha:
@asafshen still failing

asafshen:
@dorsha?

PR #1039, opened 2017-12-03
Context getters fix

Itay4:


orius123:
@Itay4 merge after build passes

PR #1043, opened 2017-12-05
IsIPInRanges fix bug

JonathanMeler:
use new lib - `netaddr`

fixes: https://github.com/demisto/etc/issues/9988


aviadl:
@JonathanMeler in commonServer we have the following function:
isIPInSubnet
I think it should do the work, and then we will have a consistent logic
Please check it, and change the script to call this function (JS)

dorsha:
@JonathanMeler - please see @aviadl comment and close the PR.

JonathanMeler:
@aviadl, changed to Javascript

dorsha:
I improved the script syntax (and performance).

PR #1048, opened 2017-12-06
Splunk test with different  query

meirwah:


TestPlaybooks/playbook-TestSplunk.yml
```diff
@@ -1,14 +1,14 @@
 id: Splunk-Test
-version: -1
+version: 3
```
asafshen: Need to revert to `-1`

PR #1049, opened 2017-12-06
SandBlast fix to pass tests

Itay4:
fixes https://github.com/demisto/etc/issues/10047


anara123:
@Itay4 merge when the build is passes

PR #1051, opened 2017-12-07
Fixed sandblast backward compatibility

dorsha:
fixes: https://github.com/demisto/etc/issues/10047

### DIFF
--- a/Integrations/integration-Sandblast.yml
+++ b/Integrations/integration-Sandblast.yml
@@ -18,7 +18,7 @@ configuration:
   type: 0
   required: true
 - display: Use API Key
-  name: prem
+  name: useApiKey
   defaultvalue: "true"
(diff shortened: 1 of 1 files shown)


Integrations/integration-Sandblast.yml
```diff
@@ -523,4 +525,4 @@ script:
     description: Use the Quote API to have a client application get the current license
       and quota status of the API Key that you use
 hidden: false
-releaseNotes: 'Added an option to use the integration without API key (for example, when using on premises)'
\ No newline at end of file
+releaseNotes: '-'
```
aviadl: Why not add release notes?

dorsha: we already had release notes from yesterday push

PR #1054, opened 2017-12-09
fix Carbon Black Response test playbook

anara123:
Don't know why till now it worked. But most probably something changed.
Same session was exported to the context, so the step cb-close-session was executed twice. The first run it closed the session, but second time the session was already closed so we got exception 404


anara123:
@dorsha can't merge because it seems that because of IceBerg the build is unstable

dorsha:
We must fix all unstable tests ASAP

anara123:
fixes https://github.com/demisto/etc/issues/10037

PR #1055, opened 2017-12-10
Send message on online users in Demisto

gshriki:
a script that sends message to online Demisto users

### DIFF
--- a/Scripts/script-SendMessageToOnlineUsers.yml
+++ b/Scripts/script-SendMessageToOnlineUsers.yml
@@ -16,14 +16,19 @@ script: |-
       }
       return ret;
   };
+  var medium = (args.medium) ? args.medium.toLowerCase() : 'email';
   var slackEnabled = commandAvailable('slack-send');
   var emailEnabled = commandAvailable('send-mail');
 
   if (!slackEnabled && !emailEnabled) {
       throw 'No available medium to send message';
   }
-
-  var medium = (args.medium) ? args.medium.toLowerCase() : 'email';
+  if (medium === 'email' && !emailEnabled) {
+    throw 'Email integration not available';
+  }
+  if (medium === 'slack' && !slackEnabled) {
+    throw 'Slack integration not available';
+  }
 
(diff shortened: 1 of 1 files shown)


Scripts/script-SendMessageToOnlineUsers.yml
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  return entries;
+type: javascript
+tags: []
+comment: Send message to Demisto online users over Email, Slack or both.
+enabled: true
+args:
+- name: medium
+  auto: PREDEFINED
```
aviadl: Add a default value ?

Scripts/script-SendMessageToOnlineUsers.yml
```diff
@@ -0,0 +1,72 @@
...
+          instanceKeys.forEach(function (key) {
+              if (allModules[key].state === 'active') {
+                  ret = true;
+              }
+          });
+      }
+      return ret;
+  };
+  var slackEnabled = commandAvailable('slack-send');
+  var emailEnabled = commandAvailable('send-mail');
```
aviadl: Do you want to return here an error in case the medium selected is not enabled?

gshriki: If it is all - than I'll only send to the enabled ones. If it is specific - there will be an error anyway.
I will add an error if none of the mediums are available.

Scripts/script-SendMessageToOnlineUsers.yml
gshriki: done

Scripts/script-SendMessageToOnlineUsers.yml
aviadl: But you will have multiple errors of the same type, which will only create confusion.
So why not create one error here (**not** for case of **all**)

gshriki: got you. agree. added.

PR #1056, opened 2017-12-10
Add entitlement changes

Nitzperetz:
Deprecate `addOneTimeEntitlement` command, Use `addEntitlement` instead.
`addEntitlement` Receives a boolean argument - `persistent`.


Nitzperetz:
@gshriki , updated.

gshriki:
I thought we agreed we are not going to release the EmailAskUser3.5.yml, no?
We should have only one file, and it should have the fromVersion.
And it all should be merged AFTER we have 3.5 out

Nitzperetz:
@gshriki , but then when upgrading content on v3.1.0, the scripts would not work. Let's discuss.

Nitzperetz:
@asafshen , @aviadl , please approve and merge.

Nitzperetz:
@aviadl , fixed.
I think we can consider removing the persistent parameter (meaning all entitlements would be persistent). It was used in order to add `one-time` entitlement as `assOnetimeEntitlement` command is now deprecated.
Any objections?

aviadl:
@Nitzperetz the persistent parameter should be an argument.
That has a default.
Default should actually be `false` to support backward compatibility

Nitzperetz:
Updated.

Nitzperetz:
@aviadl  please review and merge.

PR #1061, opened 2017-12-11
add langDetect script

gshriki:


dorsha:
Nice! I would call the script LanguageDetect.

Scripts/script-langDetect.yml
```diff
@@ -0,0 +1,36 @@
+commonfields:
+  id: langDetect
```
orius123: I think script name should start with capital letter

gshriki:
agree with both. will fix

PR #1062, opened 2017-12-11
Okta unlock user command

aviadl:


Integrations/integration-Okta.yml
```diff
@@ -296,6 +297,12 @@ script:
                 return resp;
             }
             break;
+        case 'okta-deactivate-user':
+            var userId = getUserId(args, params);
```
orius123: Did you mean to write `okta-unlock-user` here?

PR #1063, opened 2017-12-11
typo in qradar

meirwah:
typo in qradar


Integrations/integration-QRadar.yml
```diff
@@ -453,3 +453,4 @@ script:
       description: Number of results in return
     description: Get closing reasons
   isfetch: true
+releaseNotes: fix typo
```
dorsha: Change to "-" not that important for release notes

PR #1067, opened 2017-12-12
moved tanium to nightly

shaniacht1:
fixes https://github.com/demisto/etc/issues/10087


anara123:
@shaniacht1 please merge when the build is green

PR #1068, opened 2017-12-12
Fixed okta-unlock-user command

dorsha:


dorsha:
Opened issue to add test playbook

PR #1072, opened 2017-12-13
EWS with folder ID

aviadl:
Fixed folder id param for fetch incidents

### DIFF
--- a/Integrations/integration-Office_EWS.yml
+++ b/Integrations/integration-Office_EWS.yml
@@ -843,6 +843,14 @@ script:
                 lastTime: new Date(new Date().getTime() - (10 * 60 * 1000)).getTime()
             }
         }
+        if (lastRun.folderId !== params.folder) {
+          //Folder was changed, we need to reset lastRun
+          lastRun = {
+              folderId: params.folder,
+              lastId: '',
+              lastTime: new Date(new Date().getTime() - (10 * 60 * 1000)).getTime()
+          }
+        }
         var pageSize = params.pageSize ? parseInt(params.pageSize) : defaultPageSize;
         //Will fetch emails according to time stamp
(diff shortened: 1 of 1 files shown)


aviadl:
@michalgold Please check this branch as well.
Lets check if we can adjust our test playbook for this use case as well

Integrations/integration-Office_EWS.yml
```diff
@@ -843,6 +843,14 @@ script:
                 lastTime: new Date(new Date().getTime() - (10 * 60 * 1000)).getTime()
             }
         }
+        if (lastRun.folderId !== params.folder) {
+          //Folder was changed, we need to reset lastRun
+          lastRun = {
+              folderId: params.folder,
```
orius123: Why do you need to put the folder ID on the `lastRun` and not just taking it from the params?

aviadl: Since, if it was changed i want to clear the lastRun and start over again

orius123: I understand why you restart the `lastTIme` field but why putting the `params.folder` in the last run object?

aviadl: Since then i will not know to compare it with the previous run.
I want to know if the previous run was on the same folder or not

Integrations/integration-Office_EWS.yml
```diff
@@ -904,7 +912,7 @@ script:
         }
         fetchArgs.last_time = new Date(lastRun.lastTime).toISOString();
         fetchArgs['target-mailbox'] = getDefaultTargetMailbox();
-        if (params.folder) {
+        if (lastRun.folderId) {
```
orius123: It means you set `folder` just after `folderId` was changed.

aviadl: No, last run is loaded from the server each time.
And stored at the flow.
So for the first time, i will load a lastRun that will not have a folderId and params will have it, then i will put it in, and will save.
The next time, they will have the same value, so lastRun will have a folderId and we are good

What am i missing?

orius123: What I am missing is why should it solve the issue. Also today if params had it we would get into this if. So how does getting it from the last run instead of the param make it any better?

aviadl: The problem is that we checked the param object
But took the data from the lastParam object
Now hey are in sync

orius123: ok. told you that I missed something.

PR #1076, opened 2017-12-14
QRadar offense notes

Spivaka:


Integrations/integration-QRadar.yml
```diff
@@ -455,5 +473,33 @@ script:
     - name: range
       description: Number of results in return
     description: Get closing reasons
+  - name: qradar-create-note
+    arguments:
+    - name: offense_id
+      required: true
+      default: true
```
orius123: Does it make sense to make an arg default if we have more than one arg required? You will have to use arg names in the cli anyways

anara123: I think this is ok

Spivaka: it will just b the first one to pop up

PR #1077, opened 2017-12-15
fix VMRay to support file upload from war room

meirwah:
fixes: https://github.com/demisto/etc/issues/10107


Integrations/integration-VMRay.yml
```diff
@@ -104,4 +105,4 @@ script:
       description: Job sample id
     description: Get all jobs details about a specific sample.
   dockerimage: demisto/vmray
-hidden: false
+releaseNotes: 'Fix VMRay fetching file issue'
```
dorsha: The release notes should be more descriptive for the user. Hard to understand what was fixed and how it relevant for the user

PR #1078, opened 2017-12-15
GRR Integration - Added valid CSRFToken retrieval function

secdevopsai:
The getCSRFToken function enables 'POST' requests to send valid CSRF Token. grr-set-flows now properly adds valid CSRF token to header

i.e)
`!grr-set-flows client_id=C.120694fab000fff flow="{\"runner_args\":{\"flow_name\":\"Netstat\",\"output_plugins\":[]},\"args\":{}}"`

### DIFF
--- a/TestPlaybooks/playbook-McAfeeWebGatewayTest.yml
+++ b/TestPlaybooks/playbook-McAfeeWebGatewayTest.yml
@@ -1,5 +1,5 @@
 id: McAfeeWebGatewayTest
-version: 32
+version: 38
 name: McAfeeWebGatewayTest
 starttaskid: "0"
 tasks:
@@ -39,8 +39,9 @@ tasks:
       '#none#':
       - "7"
     scriptarguments:
-      all: "yes"
-      key: ""
+      all:
+        simple: "yes"
+      key: {}
     view: |-
       {
         "position": {
@@ -64,7 +65,8 @@ tasks:
       '#none#':
       - "9"
     scriptarguments:
-      list_id: ${MWG.Lists.[0].Name}
+      list_id:
+        simple: ${MWG.Lists.[0].Name}
     view: |-
       {
         "position": {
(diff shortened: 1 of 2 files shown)


Integrations/integration-GRR.yml
```diff
@@ -34,8 +34,21 @@ script:
...
+        var res = http(
+                url,
+                {
+                    Method: 'GET',
+                    Username: username,
+                    Password: password
+                },
+                insecure
+            );
+        return res.Cookies[0].Value;
```
dorsha: Just add verification that the request was succeeded 
```
    throw 'Failed to get CSRF token, request status code: ' + res.StatusCode + ' and Body: ' + res.Body + '.';
}
```

dorsha: Is the CSRF token will always be the first in the cookies array? I would look for the correct key and then take the value (to be on the safe side).

secdevopsai: Complete

secdevopsai: Complete

secdevopsai:
@dorsha changes completed. Let me know if there's anything else..

Integrations/integration-VMRay.yml
```diff
@@ -74,8 +74,7 @@ script:
         # and stores it internaly. To get the filename you need an automation.
         # You can use demisto.args()[argName] to get a specific arg. args are strings.
         file = demisto.args()['file_id']
-        path = demisto.getFilePath(file)
```
orius123: I think this a bad merge that overrides a fix. @meirwah right?

meirwah: true, @Spivaka plz fix this merge

Spivaka:
fixes https://github.com/demisto/etc/issues/10038

Integrations/integration-VMRay.yml
secdevopsai: accidental, fixing..

PR #1083, opened 2017-12-18
ignore multi-line html code

gshriki:
ignoring multiline `script` and `style` blocks

### DIFF
--- a/Scripts/script-EmailAskUserResponse.yml
+++ b/Scripts/script-EmailAskUserResponse.yml
@@ -7,6 +7,7 @@ script: |
 
   if (res[0].Type==entryTypes.note) {
       text = res[0].Contents;
+      text = text.replace(/\n/g,'__NL__').replace(/<script .+?<\/script>/g,'').replace(/<style .+?<\/style>/g,'').replace(/__NL____NL__/g,'').replace(/__NL__/g,'\n');
       response = text.replace(/<(?:.|\n)*?>/gm, '').trim("\n").split("\n")[0].trim();
       setContext('EmailAskUserResponse', response);
       return response;
@@ -26,3 +27,4 @@ args:
   default: true
   description: Entry ID where EmailAskUser will complete when user replies
 scripttarget: 0
+releaseNotes: "Ignore multi-line scripts and style sections in email body"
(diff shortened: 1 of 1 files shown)


Scripts/script-EmailAskUserResponse.yml
```diff
@@ -7,6 +7,7 @@ script: |
 
   if (res[0].Type==entryTypes.note) {
       text = res[0].Contents;
+      text = text.replace(/\n/g,'__NL__').replace(/<script .+?<\/script>/g,'').replace(/<style .+?<\/style>/g,'').replace(/__NL____NL__/g,'').replace(/__NL__/g,'\n');
```
orius123: you handle just one new line or 2? are those the only options?
Code-wise i think you should do new assignment for each replacement for readability and can even add comments
```
text = text.replace(/<script .+?<\/script>/g,'');
text = text.replace(/<style .+?<\/style>/g,'')
...
```

dorsha:
We need to think here about the performance and must have test for this.

Scripts/script-EmailAskUserResponse.yml
dorsha: I don't agree about the new assignment for each line, that's the whole purpose of chaining... It should be wel documented.

dorsha: And I would put each replace in new line of course

orius123: can also chain with new lines, just meant for it to be more readable

gshriki: Like this you mean?
```
    text = text.replace(/\n/g,'__NL__')
                .replace(/<script .+?<\/script>/g,'')
                .replace(/<style .+?<\/style>/g,'')
                .replace(/__NL____NL__/g,'')
                .replace(/__NL__/g,'\n');
```

orius123: yup

gshriki:
performance wise - I don't think it's a concern as this script will not run that often (only on responses from EmailAskUser)

orius123:
And what about a test?

gshriki:
Think this should be part of the EmailAskUser overall test.

PR #1085, opened 2017-12-19
Fix parse csv

anara123:
fixes https://github.com/demisto/etc/issues/10125


dorsha:
@anara123 ?

anara123:
Trying to figure our why the test fails on timeout. Have no clue yet

PR #1092, opened 2017-12-21
support arcsight 6.1 6.7 and 6.9

anara123:
fixes https://github.com/demisto/etc/issues/10143


dorsha:
@anara123 ?

anara123:
I have session with customer which has similar problem, so I want to test the fix in his environment. If it will fix the problem, I will close the PR

anara123:
Now it is ready to be merged

Issue #1095, opened 2017-12-24
"filtered" array is not used anywhere


jeremynielson:
This still extracts for processing the w3.org and microsoft URLs that should be filtered.  I suspect that it's because the filtered array is not used anywhere...

asafshen:
Hi @jeremynielson,

I would like to get more details on this issue please
Which integration's command/script are you running?

Thanks

dorsha:
Fixed

PR #1096, opened 2017-12-24
Remove unused variable

asafshen:
remove unused `filter` variable from ExtractDomain & ExtractURL

we are now filter only by indicators whitelist


asafshen:
Fixes: https://github.com/demisto/content/issues/1095

PR #1099, opened 2017-12-25
Zendesk additions

gshriki:
add user
get article 
showing organization details in ticket details


Integrations/integration-Zendesk.yml
```diff
@@ -1,4 +1,4 @@
-commonfields:
```
dorsha: revert

Integrations/integration-Zendesk.yml
```diff
@@ -140,6 +140,12 @@ script:
         }
     };
 
+    var getOrganizationDetails = function(id) {
+        var res = sendRequest('GET', 'organizations/' + id + '.json');
+        var org = res.organization;
```
dorsha: too many assignments.
Just do:
`return res.organization `
Or even 
`sendRequest('GET', 'organizations/' + id + '.json') .organization`

Integrations/integration-Zendesk.yml
```diff
@@ -394,19 +392,28 @@ script:
...
+    var addUser = function() {
+        var body = {
+                user: {
+                    name: args.name,
+                    email: args.email
+                }
+        };
+
+        res = sendRequest('POST', 'users/create_or_update.json', JSON.stringify(body));
+        md = 'Zendesk user added successfuly.\n';
```
dorsha: I think it is better to write the name/email of that user. will be clearier after.

Integrations/integration-Zendesk.yml
```diff
@@ -430,10 +437,16 @@ script:
...
         var ec = {}
         var usersEc = {}
+        var org = {};
         var newContext = {};
 
         md += 'Requester: __' + getUserName(resDetails.ticket.requester_id) + '__\n';
         usersEc[resDetails.ticket.requester_id]=getUserName(resDetails.ticket.requester_id);
+        if (resDetails.ticket.organization_id) {
+            org = getOrganizationDetails(resDetails.ticket.organization_id);
+            md += 'Organization: __' + org.name + '__ (' + org.tags.join(', ') +')\n';
```
dorsha: please share screenshot of this org results, not sure I got how will it look like.

Also, will you always have `org.tags`? can it be null somehow? If so, just do:
`(org.tags || []).join(', ')`  - will handle this

Integrations/integration-Zendesk.yml
```diff
@@ -511,17 +522,23 @@ script:
...
             Users: usersEc,
+            Org: org,
             CustomFields: fieldsEc
         }
-
         return({ContentsFormat: formats.json, Type: entryTypes.note, Contents: res, HumanReadable: md, EntryContext: ec})
     };
 
+    var getArticle = function() {
+        var locale = (args.locale) ?  args.locale : 'en-us';
```
dorsha: just do:
`var locale = args.locale || 'en-us';`

Integrations/integration-Zendesk.yml
```diff
@@ -511,17 +522,23 @@ script:
...
+        var id = args.articleID;
```
dorsha: no need for this assignment, just use it after in the URI

Integrations/integration-Zendesk.yml
```diff
@@ -511,17 +522,23 @@ script:
...
+        var url = 'help_center/' + locale + '/articles/' + id + '.json';
+        res = sendRequest('GET', url);
+        return ( {ContentsFormat: formats.html, Type: entryTypes.note, Contents: res.article.body} );
```
dorsha: just pay attention that here you expose to XSS attacks.. but because our server handles XSS we should be fine. - please verify that.

Integrations/integration-Zendesk.yml
```diff
@@ -864,4 +882,23 @@ script:
...
+    arguments:
+    - name: articleID
+      required: true
+      default: true
+      description: Article ID to get
+    - name: locale
+      description: Locale. Defaults to en-us
+    description: Get Zendesk Help Center article
   isfetch: true
+releaseNotes: Added zendesk-add-user for adding end users. Added zendesk-get-article to get help center article
```
dorsha: I would also mention about the org, and put it in new line if you can..

Integrations/integration-Zendesk.yml
```diff
@@ -511,17 +522,23 @@ script:
             ec.Ticket = newContext;
         }
 
-
         var res = {
             TicketDetails: resDetails,
             TicketComments: resComments,
             Users: usersEc,
+            Org: org,
```
dorsha: I would call it Organization and not Org.

Integrations/integration-Zendesk.yml
gshriki: yes. you are right. had there more logic which I removed so it stayed... will do the return res.organization.

Integrations/integration-Zendesk.yml
gshriki: added

Integrations/integration-Zendesk.yml
gshriki: will fix.
With tags:
image

without tags:
image

Integrations/integration-Zendesk.yml
gshriki: will fix

Integrations/integration-Zendesk.yml
gshriki: fixed

Integrations/integration-Zendesk.yml
gshriki: fixed

Integrations/integration-Zendesk.yml
gshriki: good point. will check

gshriki: Checked.
Looks like zendesk themselves do not allow XSS:
> Zendesk Support utilizes Ruby on Rails framework security controls to limit exposure to OWASP Top 10 security flaws. These include inherent controls that reduce our exposure to Cross Site Scripting (XSS), Cross Site Request Forgery (CSRF), and SQL Injection (SQLi), among others.

In addition, I checked manually and we also handle it well.

PR #1101, opened 2017-12-26
Added valid CSRFToken retrieval function

secdevopsai:
The getCSRFToken function enables 'POST' requests to send valid CSRF Token. grr-set-flows now properly adds valid CSRF token to header

i.e)
`!grr-set-flows client_id=C.120694fab000fff flow="{\"runner_args\":{\"flow_name\":\"Netstat\",\"output_plugins\":[]},\"args\":{}}"`


secdevopsai:
Added this new PR ... Couldn't get the other [(1078)](https://github.com/demisto/content/pull/1078) to revert an accidental commit

Integrations/integration-GRR.yml
```diff
@@ -34,8 +34,30 @@ script:
...
 
+    var getCSRFToken = function(url) {
+        var res = http(
+                url,
+                {
+                    Method: 'GET',
+                    Username: username,
+                    Password: password
+                },
+                insecure
```
orius123: Add also proxy param

Integrations/integration-GRR.yml
```diff
@@ -34,8 +34,30 @@ script:
     var server = params.server + ':' + params.port;
     var insecure = params.insecure;
 
+    var getCSRFToken = function(url) {
+        var res = http(
+                url,
```
orius123: Will it work with any URL?

secdevopsai: Any URL that is referenced in this integration, yes.  Function was created while referencing GRR Python API Client (https://github.com/google/grr/blob/master/api_client/python/grr_api_client/connectors/http_connector.py)

Integrations/integration-GRR.yml
```diff
@@ -34,8 +34,30 @@ script:
...
+                    Password: password
+                },
+                insecure
+            );
+        if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            throw 'Failed to get CSRF token, request status code: ' + res.StatusCode + ' and Body: ' + res.Body + '.';
+        }
+
+        for (var ix in res.Cookies) {
+            if (res.Cookies[ix].Name.toLowerCase() == 'csrftoken') {
```
dorsha: Should be `===`

Integrations/integration-GRR.yml
Spivaka: I think its safer to use the server url rather than sending GET to each url, you might get a 405 error

secdevopsai: @Spivaka - That's speculative... The client source above shows if the CSRF variable is not present then it will perform GET request on specified URL endpoint.  I'm following GRR implementation on this one -- Ignoring it resulted in the generation of this PR

Integrations/integration-GRR.yml
orius123: And what about proxy? (was requested in the other PR)

secdevopsai: @orius123 - GRR is an incident response tool that is typically installed within an organization's internal infrastructure.  Additionally, I don't see any references on the GRR documentation for allowing GRR clients to proxy IR data back to GRR server.

orius123: if it's always in the same network, and no way you will need proxy to access the GRR server, than it's cool, but if in some cases you might need it, why not having it here? 
don't see how it can hurt us.. @anara123 WDYT?

anara123: I think it is always better to have it there just in case

secdevopsai: Sorry for delay... Added proxy @anara123

dorsha:
@secdevopsai next time it is better to clone the repo instead of fork, since the build is not running when forking

dorsha:
@secdevopsai what you can do, is to create a branch in demisto content and merge this fork into it instead directly to the master.

You can change the base branch here:

image

image

orius123:
@dorsha we need to add it to the contributing.md guide

orius123:
@dorsha @anara123 can we merge it?

dorsha:
Yes, after handling the branch. Can't be merge to master from fork

PR #1103, opened 2017-12-27
Creating user type indicator when using ADGetUser automation

meirwah:


meirwah:
@JonathanMeler please review

PR #1107, opened 2017-12-27
Add hostname indicator

meirwah:


Scripts/script-ADGetComputer.yml
```diff
@@ -187,3 +191,4 @@ outputs:
   description: The Endpoint hostname
 - contextPath: Endpoint.Groups
   description: The groups the endpoint is part of
+releaseNotes: "Adding hostname indicator to ADGetComputer automation and making it enhancement script"
```
aviadl: Also mention that the default parameter has changed

PR #1112, opened 2017-12-31
Script samples for field triggers 

Itay4:
Fixes https://github.com/demisto/etc/issues/9440


anara123:
@Spivaka please review

Scripts/script-emailFieldTriggered.yml
```diff
@@ -0,0 +1,21 @@
...
+script: |-
+  var users = executeCommand('getUsers', {});
+  for (var i = 0; i < users[0].Contents.length; i++) {
+      if (incidents[0].owner === users[0].Contents[i].name) {
+          var email = users[0].Contents[i].email;
+          break;
+      }
+  }
+  var body = 'Hello,\nField ' + args.name + ' was changed from ' + args.old + ' to ' + args.new + ' in incident ' + incidents[0].name + '.\n--DBot';
+  var subject = 'Field ' + args.name + ' was triggered';
```
anara123: I think it is better to send incident link instead of incident name.

Scripts/script-emailFieldTriggered.yml
```diff
@@ -0,0 +1,21 @@
...
+  executeCommand('SendEmail', {'to':email, 'subject':subject, 'body':body});
```
anara123: check if email exists, then send email. Because incident can be without  owner and then email will be undefined

anara123: check this scenario incident without owner

Scripts/script-emailFieldTriggered.yml
Itay4: It is not possible to do at this time, as the url is not exposed to scripts.
There's an open issue for it.

PR #1114, opened 2017-12-31
Add hostname indicator to splunk search results

meirwah:


Integrations/integration-SplunkPy.yml
```diff
@@ -215,14 +215,19 @@ script:
 
         reader = results.ResultsReader(oneshotsearch_results)
         res = []
+        dbot_scores = []
         for item in reader:
             if isinstance(item, results.Message):
                 demisto.results({"Type": 1, "ContentsFormat": "json", "Contents": json.dumps(item.message)})
             elif isinstance(item, dict):
+                if demisto.get(item, 'host'):
+                    dbot_scores.append({'Indicator': item['host'], 'Type': 'hostname', 'Vendor': 'Splunk', 'Score': 0, 'isTypedIndicator': True})
```
asafshen: only 0 `Score`?

meirwah: yes, from splunk hostname is not necessary malicious

Integrations/integration-SplunkPy.yml
```diff
@@ -409,3 +414,4 @@ script:
     description: 'Create new search job in Splunk '
   dockerimage: demisto/splunksdk:1.0
   isfetch: true
+releaseNotes: 'Add hostname indicator to splunk search results'
```
asafshen: do you think `Added hostname reputation to Splunk search results` is better?

PR #1119, opened 2018-01-03
Added new "Add Attribute" command and new "Create MISP Event" command

dem-evgeny:
Added new "Add Attribute" command and new "Create MISP Event" command


Integrations/integration-MISP.yml
```diff
@@ -417,6 +423,92 @@ script:
...
+
+
+    var getNowTime = function() {
+        var time = new Date();
+        time = addMinutes(time,time.getTimezoneOffset());
+        var month = "0" + (time.getMonth()+1);
+        var day = "0" + time.getDate();
+        var hours = "0" + time.getHours();
+        var minutes = "0" + time.getMinutes();
+        var seconds = "0" + time.getSeconds();
```
yuvalbenshalom: what is the reason for this? looks like you assign value day /hours/ minutes/second but never use them.

Integrations/integration-MISP.yml
```diff
@@ -417,6 +423,92 @@ script:
...
+
+        var body = { 'Event': {
+            'date': getNowTime(),
+            'threat_level_id': '1',
+            'published': args.published,
+            'info': args.info,
+            'analysis': '0',
+            'distribution': '0',
+            'Attribute': [currentArgs] } };
+
```
yuvalbenshalom: are you sure about all of those constants?  
'threat_level_id': '1'
 'analysis': '0'
 'distribution': '0'

maybe some of them needs to be arguments to the command?

Integrations/integration-MISP.yml
dem-evgeny: I do use it here:
var body = { 'Event': {
            'date': getNowTime(),
            'threat_level_id': '1',
            'published': args.published,
            'info': args.info,
            'analysis': '0',
            'distribution': '0',
            'Attribute': [currentArgs] } };

Integrations/integration-MISP.yml
dem-evgeny: Yes, I checked it at the customer

dem-evgeny: Fixed

Integrations/integration-MISP.yml
dem-evgeny: Fixed

PR #1124, opened 2018-01-04
removed if __name__ == '__main__' from all the integration scripts

anara123:


anara123:
@yuvalbenshalom I forgot releaseNotes

PR #1125, opened 2018-01-05
Enable fetch incidents to intsights

dorsha:
Everything is already there.. but `isfetch: true` was missing.


Spivaka:
we decided not to implement it as intsights doesn't have a way to query IOC to a certain point at time or ordered IDs. so we either filter on our side or wait for them to add it.

dorsha:
So why do we have fetch-incidents  implemented? Just in case?
According to the customer it seems that it used to work before.

dorsha:
image

PR #1127, opened 2018-01-07
python common server: support both python 2.7 and python 3.5

yaakovi:
fixes: https://github.com/demisto/etc/issues/8273

python common server:
main change is adding a string type list (different between the two versions)

test script:
generalized the tests script to support additional testing of the common server in the future.


yaakovi:
last changes are for replacing the demisto.dt back to dq function.

Scripts/script-CommonServerPython.yml
```diff
@@ -1,6 +1,6 @@
 commonfields:
   id: CommonServerPython
-  version: 1
+  version: 200
```
anara123: change this to -1

PR #1133, opened 2018-01-09
Add 'GeneratePassword' script.

gsamuelhays:
The Generate Password script is used to generate relatively secure
passwords for use by other scripts/automations.


aviadl:
@gsamuelhays created a local branch and added some fixes as listed bellow:
Remove required fields - Not needed if we have defaults  …
Fixed script ID - should be like script name
Remove default_sane - not needed if we have default values, will make the use easier
Change default debug to false
Some lintting fixes

We will also add a test playbook, so this script will be tested as part of our build process
Please  review this PR, and see that i didnt brake anything
https://github.com/demisto/content/pull/1136

PR #1134, opened 2018-01-09
Reports Time Zone

shaniacht1:


dorsha:
@shaniacht1 do we need to add `fromVersion: 3.5.0`? Will it still work for users without the sane-reports fix?

shaniacht1:
@dorsha 
For docx and pdf it will work as before
for csv it will add a space:
screen shot 2018-01-09 at 11 08 52

dorsha:
We can live with that

PR #1143, opened 2018-01-10
Forcepoint unstable test fix

Itay4:
Fixes https://github.com/demisto/etc/issues/10286


anara123:
@Itay4 the test is still failing

PR #1146, opened 2018-01-11
New widget by entries

aviadl:


Widgets/widget-InvestigationActivity.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "investigation-activity",
+  "version": -1,
+  "name": "Investigation Activity Timeline",
+  "dataType": "entries",
+  "widgetType": "line",
+  "size": 5,
+  "query": "",
+  "isPredefined": false,
```
orius123: I think it should be set to true and the user will be able to dup it if he wants.
all other content widgets are set to true ATM

PR #1147, opened 2018-01-13
Indicator widgets

aviadl:


Widgets/widget-IndicatorActivity.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "indicators-activity",
+  "version": -1,
+  "name": "Indicators Activity",
+  "dataType": "indicators",
+  "widgetType": "line",
+  "size": 1000,
```
orius123: Why the limit. We return just groups.

aviadl: Limits to number of points loaded

Widgets/widget-IndicatorManualUsers.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "indicators-manual-user",
+  "version": -1,
+  "name": "Manual Reputation Indicators by User",
+  "dataType": "indicators",
+  "widgetType": "bar",
+  "size": 1000,
```
orius123: Can we have more than 1000 users? Maybe we should put a reasonable default in the server instead

Widgets/widget-IndicatorActivity.json
orius123: Than we should add it to all line charts

aviadl:
@orius123 fixed user widget size

PR #1148, opened 2018-01-13
Some more incidents widgets

aviadl:
By role
Number of incidents in error


aviadl:
@asafshen this will also fix the bad release notes fix
Please merge ASAP
You will still need to make sure the script is fixed

PR #1149, opened 2018-01-13
fix reputation format

asafshen:
fix reputation format after bad clear with `release-notese-clrear.py`


asafshen:
closing, will fix with - https://github.com/demisto/content/pull/1148

PR #1154, opened 2018-01-14
Domain fix2

meirwah:


meirwah:
Different approach, plz review @erezh31

PR #1160, opened 2018-01-14
add incidents-by-close-reason widget

meirwah:


meirwah:
plz review again @aviadl

PR #1164, opened 2018-01-16
playbook new schema

asafshen:
Playbook new schema (with conditionals, filters & outputs)


asafshen:
Tested against `DBot Indicator Enrichment - Generic` from demo2

@JonathanMeler FYI

PR #1167, opened 2018-01-16
FilesToContext

gshriki:
Re-exposing the File context key. Useful when using DeleteContext...

### DIFF
--- a/Scripts/script-FilesToContext.yml
+++ b/Scripts/script-FilesToContext.yml
@@ -1,0 +1,42 @@
+commonfields:
+  id: FilesToContext
+  version: -1
+name: FilesToContext
+script: |
+  var entries = executeCommand("getEntries", {});
+  var fileEntries = dq({entries:entries},'entries(val.FileMetadata)');
+
+  var files = [];
+
+  fileEntries.forEach(function(fileEntry) {
+      var entry = {
+          EntryID: fileEntry.ID,
+          Name: fileEntry.File,
+          Size: fileEntry.FileMetadata.Size,
+          Info: fileEntry.FileMetadata.Info,
+          Type: fileEntry.FileMetadata.Type,
+          MD5: fileEntry.FileMetadata.MD5,
+          SHA1: fileEntry.FileMetadata.SHA1,
+          SHA256: fileEntry.FileMetadata.SHA256,
+          SSDeep: fileEntry.FileMetadata.SSDeep,
+      };
+      entry.Extention = (entry.Name.indexOf('.') > -1) ? entry.Name.split('.')[entry.Name.split('.').length-1] : '';
+      files.push(entry);
+  });
+
+  if (files.length === 0) {
+      return 'No files';
+  } else {
+      return {
+          Type: entryTypes.note,
+          Contents: files,
+          ContentsFormat: formats.json,
+          HumanReadable: 'File key set with ' + files.length + ' file(s)',
+          EntryContext: {File: files}
+      };
+  }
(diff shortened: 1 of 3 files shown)


dorsha:
This is the same mistake we did with the expose incident and with the expose modules. No?

dorsha:
If this is not for playbooks run maybe you should write it in the description

gshriki:
not sure about your first comment. what do you mean mistake?

Scripts/script-FilesToContext.yml
```diff
@@ -0,0 +1,42 @@
...
+script: |
+  var entries = executeCommand("getEntries", {});
+  var fileEntries = dq({entries:entries},'entries(val.FileMetadata)');
+
+  var files = [];
+
+  fileEntries.forEach(function(fileEntry) {
+      var entry = {
+          EntryID: fileEntry.ID,
+          Name: fileEntry.File,
```
asafshen: Please add Set context outputs as well (with types if known)

Scripts/script-FilesToContext.yml
```diff
@@ -0,0 +1,42 @@
...
+
+  if (files.length === 0) {
+      return 'No files';
+  } else {
+      return {
+          Type: entryTypes.note,
+          Contents: files,
+          ContentsFormat: formats.json,
+          HumanReadable: 'File key set with ' + files.length + ' file(s)',
+          EntryContext: {File: files}
```
asafshen: I think key should be instead `outputPaths.file` instead `File`, so it will be merge with existing context values  if exist

TestPlaybooks/playbook-FilesToContext-Test.yml
```diff
@@ -0,0 +1,167 @@
...
+  "3":
+    id: "3"
+    taskid: dcbc1e1b-03af-4e34-8953-88e5aa091c76
+    type: condition
+    task:
+      id: dcbc1e1b-03af-4e34-8953-88e5aa091c76
+      version: -1
+      name: Check filename
+      description: Check if File context key is populated correctly with filename
+      scriptName: AreValuesEqual
```
asafshen: this playbook was written on an "old" version (not master)
I suggest you to build it on a new version (and you can use our built it conditions instead `AreValuesEqual` script 

Not a must of course, but recommended :)

dorsha:
I don't want it to be used in playbooks (since it is not needed). Users can easily do that mistake. Need to clarify this somehow.

TestPlaybooks/playbook-FilesToContext-Test.yml
```diff
@@ -0,0 +1,167 @@
...
+      }
+  "4":
+    id: "4"
+    taskid: 82a0c2fa-7513-4a45-84cb-458f3d70aabd
+    type: regular
+    task:
+      id: 82a0c2fa-7513-4a45-84cb-458f3d70aabd
+      version: -1
+      name: Close
+      description: Close investigation
```
asafshen: Very responsible, but not need to close the investigation since built process delete the incident (FYI)

Tests/conf.json
```diff
@@ -194,6 +194,9 @@
             "integrations": "McAfee ESM-v10",
             "playbookID": "McAfeeESMTest",
             "timeout": 500
+        },
+        {
```
asafshen: I think you accidentally located this on `"skipped"` section

gshriki:
Closing this for now as I do not think we agree it is needed. We can discuss and decide later.

PR #1168, opened 2018-01-17
File Enrichment - Generic playbook

Itay4:
Fixes https://github.com/demisto/etc/issues/10316


JonathanMeler:
@Itay4 you need to add a test playbook

PR #1175, opened 2018-01-18
Mcafee tie playbook update

guydemi:
updated to use the wizard instead of dt when possible,
added outputs,
added test playbook

relates to: https://github.com/demisto/etc/issues/8532
relates to: https://github.com/demisto/etc/issues/10411


Playbooks/playbook-IOC_Hunt_-_TIE.yml
```diff
@@ -0,0 +1,303 @@
+id: IOC Hunt - TIE
+version: -1
+name: IOC Hunt - TIE
```
anara123: leave the old id. For backward compatability

anara123:
Your test fails because you missing descriptions, please make sure the build is green

anara123:
@guydemi also attach screenshot of the IOC Hunt - TIE playbook and the screenshot of test playbook.

anara123:
image
Why you add face hostname?

asafshen:
@guydemi 
we have made some changes in playbook structure - please download the playbook again from an updated demisto env (demo2/your local env)

Playbooks/playbook-IOC_Hunt_-_TIE.yml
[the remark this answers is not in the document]
guydemi: so what will happen when less than 3.5.0?
maybe duplicate and use 2 different playbooks? (one unchanged - for backward compatibility, and one new with "fromVersion")

guydemi:
@anara123, I have added the fake hostname because we don't have a working ePo instance, so hostname will never return while testing

guydemi:
ioc_hunt_-_tie_sun_jan_21_2018
ioc_hunt_-_tie_-_test_sun_jan_21_2018

Tests/conf.json
```diff
@@ -165,6 +165,9 @@
         },
         {
             "playbookID": "endpoint_enrichment_-_generic_test"
+        },
+        {
+            "playbookID": "IOC Hunt - TIE - Test"
```
anara123: You need to add which integration playbook will use `McAfee Threat Intelligence Exchange`

PR #1178, opened 2018-01-20
Added 3.5 sub-folder to Playbooks

dorsha:
image


aviadl:
@dorsha why did we do this?
If at all, it makes sense to do it the other way around ...

PR #1180, opened 2018-01-20
upgrade 3.5 playbook

asafshen:
upgrade 3.5 playbook to the new schema
also zip bundle all sub-folders


dorsha:
@asafshen I moved the playbook into the main Playbooks folder

PR #1182, opened 2018-01-21
fix error message of assigned user failure.

aviadl:
Now it will contain the exact error message the server is sending


aviadl:
@dorsha please review again

PR #1183, opened 2018-01-21
Zip file

gshriki:
Zip file into the war room. useful when sending files over email.


Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,31 @@
...
+  import zipfile
+  import os
+
+  res = demisto.executeCommand('getFilePath', {'id': demisto.args()["EntryID"]})
+  os.system('cp ' + res[0]['Contents']['path'] + ' ' + res[0]['Contents']['name'])
+
+  zipFilename = res[0]['Contents']['name'] + '.zip'
+  zipfile.ZipFile(zipFilename, mode='w').write(res[0]['Contents']['name'])
+
+  with open(zipFilename, 'rb') as f:
```
dorsha: Isn't it heavy? what if the file size in 1-2 GB? am I missing something?

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,31 @@
+commonfields:
+  id: ZipFile
+  version: -1
+name: ZipFile
+script: |
+  import zipfile
+  import os
+
+  res = demisto.executeCommand('getFilePath', {'id': demisto.args()["EntryID"]})
+  os.system('cp ' + res[0]['Contents']['path'] + ' ' + res[0]['Contents']['name'])
```
dorsha: what if res is empty or null? can it be?

dorsha: You are using `res[0]['Contents']['name']`  many times, please use variable

dorsha: You don't want to support directories, correct?

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,31 @@
...
+tags:
+- Utility
+- file
+comment: Zip a file and upload to war room
+enabled: true
+args:
+- name: EntryID
+  required: true
+  description: entry id of file to zip
+outputs:
```
dorsha: remove the outputs since they are empty.

dorsha:
@gshriki what is the status of this issue?

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,31 @@
...
+  demisto.results(fileResult(zipFilename,data))
+type: python
+tags:
+- Utility
+- file
+comment: Zip a file and upload to war room
+enabled: true
+args:
+- name: EntryID
+  required: true
```
orius123: make also default

gshriki:
Sorry guys. All your comments are valid. 
Will get back to it when things cool down a bit, fix and re-PR...

hirshbergshachar:
@gshriki I'll fix it

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,132 @@
...
+  fileEntryID = ''
+  zipName = None
+  # preparation for later version which will support several inputs
+  # if demisto.args().has_key('fileName'):
+  #   entries = demisto.executeCommand('getEntries', {})
+  #   for entry in entries:
+  #       fn = demisto.get(entry, 'File')
+
+  #       is_text = type(fn) in [unicode, str]
+  #       is_correct_file = demisto.args().get('fileName', '').lower() == fn.lower()
```
anara123: remove all these comments

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,132 @@
...
+  #           if demisto.args().has_key('fileName'):
+  #               demisto.results({
+  #                   'Type': entryTypes['error'],
+  #                   'ContentsFormat': formats['text'],
+  #                   'Contents': demisto.args().has_key('fileName') + ' not such file in war room'
+  #               })
+  #           sys.exit(0)
+
+  if demisto.args().has_key('entryID'):
+    fileEntryID = demisto.args().get('entryID')
```
anara123: entryID is required field, so there is no need to check `has_key` you can just assume that there is such key

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,132 @@
...
+  if not fileEntryID:
+    demisto.results({
+        'Type': entryTypes['error'],
+        'ContentsFormat': formats['text'],
+        'Contents': 'You must set an entryID when using the zip script'
+    })
+    sys.exit(0)
+
+  if not zipName:
+    demisto.results({
```
anara123: if the zipName is not provided then by default add zip to its name and zip.
Example: if the file we zip called `test.txt` then the by default it will become `test.txt.zip`

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,132 @@
...
+        'Contents': 'You must set a zipName when using the zip script'
+    })
+    sys.exit(0)
+
+  res = demisto.executeCommand('getFilePath', {'id': fileEntryID})
+  if res[0]['Type'] == entryTypes['error']:
+    demisto.results({
+        'Type': entryTypes['error'],
+        'ContentsFormat': formats['text'],
+        'Contents': 'Failed to get the file path for entry: ' + fileEntryID
```
anara123: Print also the original error message for better troubleshooting. `res[0]['Contents']`

Scripts/script-ZipFile.yml
```diff
@@ -0,0 +1,132 @@
...
+
+  demisto.results(fileResult(zipName,file_data))
+
+  results = []
+  results.append(
+    {
+        'Type': entryTypes['note'],
+        'ContentsFormat': formats['json'],
+        'Contents': { 'zippedFiles': zipName },
+        'EntryContext': { 'zippedFiles': zipName, 'File(val.EntryID=="' + fileEntryID + '").zipped': True },
```
anara123: When we output to context like here `zippedFiles`, our conventions are to EntryContext with capital case
`zippedFiles` => `ZippedFiles`
Also you need to add corresponding output
image

Tests/conf.json
```diff
@@ -284,6 +284,13 @@
         },
         {
             "playbookID": "GeneratePassword-Test"
+        }, 
+        {
+            "playbookID": "ZipFile-Test"
+        },
+        {
+            "integrations": "VxStream",
+            "playbookID": "VxStream Test"
```
anara123: remove VxStream test

anara123:
@hirshbergshachar lets add password protect

PR #1185, opened 2018-01-21
fix to the new playbook schema

erezh31:


erezh31:
@JonathanMeler plz approve & merge

PR #1186, opened 2018-01-21
align widget ID's according to name

meirwah:


aviadl:
@meirwah i think that you need to fix them on dashboards as well

meirwah:
You asked for it!!!

Widgets/widget-ActiveIncidentsBySeverity.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "6b25afb6-7a95-4878-a24c-24cf067eb38a",
+  "id": "active-incidents-by-severity",
```
dorsha: The dashboard is still using the old ID.

image

Widgets/widget-UnassignedIncidentsNumber.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "63de68a7-1da4-489e-8939-b8566da5fc58",
+  "id": "unassigned-incidents",
```
orius123: should add `-number`

Widgets/widget-ActiveIncidentsBySeverity.json
orius123: The widgets in the dashboard are copies of the widgets when we created them, they don't update if we update the widgets.

PR #1191, opened 2018-01-21
Fix AlienVault  outputs

asafshen:
IP.Malicious -> File.Malicious


dorsha:
@asafshen what is the status of this PR?

PR #1192, opened 2018-01-21
Post processing script to generate summary

aviadl:
Can be triggered when investigation is closed

### DIFF
--- a/Scripts/script-generateSummaryReport.yml
+++ b/Scripts/script-generateSummaryReport.yml
@@ -1,0 +1,23 @@
+commonfields:
+  id: GenerateInvestigationSummaryReport
+  version: -1
+fromversion: 3.5.0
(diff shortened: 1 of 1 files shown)


Scripts/script-generateSummaryReport.yml
```diff
@@ -0,0 +1,23 @@
...
+enabled: true
+args:
+- name: name
+  default: true
+  description: Report name to generate (should correspond with report type)
+  defaultValue: Investigation Summary
+- name: type
+  description: Report type to generate (should correspond with report name)
+  defaultValue: pdf
+- name: incidentId
```
dorsha: do we really need the incidentId argument for post-processing?

aviadl: No, but this way this script can be used without pre processing as well

PR #1201, opened 2018-01-22
delete .replace from all fetch incidents

anara123:
fixes https://github.com/demisto/etc/issues/10431

### DIFF
--- a/Integrations/integration-Cylance_Protect.yml
+++ b/Integrations/integration-Cylance_Protect.yml
@@ -284,7 +284,7 @@ script:
         var incident = {
             name: threat.priority + ' ' + threat.name,
             labels: [],
-            rawJSON: JSON.stringify(threat),
+            rawJSON: JSON.stringify(threat)
         };
         delete(threat.name);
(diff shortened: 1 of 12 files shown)


Spivaka:
we added it on purpose AFAIR cause otherwise we had parsing errors... did you check it still work with those integrations?

liorkol:
Here is the json field that broke classification
 (only some incidents had quotes in the value so at first there seemed to
be no pattern.. Some incidents just weren't being classified)

"u_soc_custom_string_2": "\u003cResource URI=" / All Rules / Real - time
Rules / Redacted / Corporate / Security Domains / Insider Threat / MyDetec
","

liorkol:
@anar if you want to add a content test and you need the full json for this
please ping me, I will get it

anara123:
@Spivaka This is what I thought. But later I noticed that without replace it works fine. Didn't tested the fetch on all the integrations. But will check on few of them, just to make sure.
@liorkol Removing the `.replace` fixed the fetch problem in servicenow, right?

PR #1204, opened 2018-01-23
Fix playbook tests

erezh31:
fixes:
https://github.com/demisto/etc/issues/10434
fixes:
https://github.com/demisto/etc/issues/10434


erezh31:
fix https://github.com/demisto/etc/issues/10443

PR #1205, opened 2018-01-23
 Take into account emails for domain extraction

meirwah:


Scripts/script-ExtractDomain.yml
```diff
@@ -155,8 +199,14 @@ args:
 - name: tldList
   description: Name of Demisto list to hold known TLD list, so will be ignored when
     looking for domains
+- name: extractFromEmails
+  auto: PREDEFINED
+  predefined:
```
aviadl: Need to add default values `true` to markAsIndicator and `calcReputation` - then you will not need the logic at the top of the script

Scripts/script-ExtractDomain.yml
```diff
@@ -13,10 +25,76 @@ script: |-
           reg = reg.substring(1);
           return str.match(reg);
       }
-
```
aviadl: For regex match is the logic with the `~` prefix still valid at 3.5?

Scripts/script-ExtractDomain.yml
```diff
@@ -13,10 +25,76 @@ script: |-
...
       return false;
   };
 
+  var addToContextAndMD = function(domains,ec,md) {
```
aviadl: add spaces between parameters `domains, ec, md`

Scripts/script-ExtractDomain.yml
meirwah: It works with backward too

meirwah:
@aviadl plz review fixes

PR #1209, opened 2018-01-23
Take into account indicators (from indicators DB) for IsMalicious

meirwah:


meirwah:
@gshriki this should fix something we talked about

Scripts/script-IsMaliciousIndicatorFound.yml
```diff
@@ -27,6 +27,11 @@ script: |-
   if (ips && (!Array.isArray(ips) || ips.length > 0)) {
       return 'yes';
   }
+  var indicatorsRes = executeCommand("findIndicators", {"query":"(reputation:Bad or reputation:Suspicious)  and incident.id:"+investigation.id,"size":1});
```
aviadl: I think `investigationIDs` would work much faster

meirwah:
@aviadl can you review fix?

PR #1215, opened 2018-01-24
Fix content b ug

erezh31:


meirwah:
CI stills fails:
https://www.calcalist.co.il/local/articles/0,7340,L-3730356,00.html?ref=ynet

PR #1221, opened 2018-01-25
fixed task to new tasks indexing

Bars92:


Widgets/widget-LateTasks.json
```diff
@@ -1,11 +1,11 @@
 {
-  "id": "b2e42da2-79b3-4265-9a2a-60365eec4443",
+  "id": "late-tasks",
   "version": -1,
   "fromVersion": "3.5.0",
   "name": "Late Tasks by User",
   "dataType": "tasks",
   "widgetType": "bar",
-  "query": "late",
+  "query": "currentLate",
```
orius123: I think that now that we have relative time support in queries `dueDate:>now` should also work.
but this will also do the job

Bars92: Do we prefer to use relative and deprecate this?

orius123: Test it. If it works I think it is more consistent

PR #1222, opened 2018-01-26
Generate report script output

aviadl:
Now places data to context
Also changed script name & id had typo
No problem to change ID since it is not released yet


dorsha:
@aviadl build stil fails

PR #1223, opened 2018-01-27
Added the ability to assign current user as owner

dorsha:
image


dorsha:
Not sure, because the Header is "Improved Scripts" so we just list the changes.
Having "Added the.. " in each bullet will look odd... it is not commit messages or PR titles. 
Take a look at our product release notes (or any other products), we don't write: "Added the ability..." for each bullet.

asafshen:
Ok make sence

PR #1228, opened 2018-01-28
Deprecate VirustotalIsMalicious script

aviadl:


liorkol:
@meirwah Sent you answer separately

PR #1237, opened 2018-01-29
Fix url enrichment test

liorblob:
Also changed the playbook, make sure to look at it in the review


liorblob:
Should be more stable.

liorblob:
Skipped for now, further investigation required

dorsha:
@liorblob you referenced the PR to itself, I guess you wanted to reference to the issue..

And skipping is not the best solution for this problem.. we must make sure we enable it asap, so we won't have untested flows.

liorblob:
Agreed, I'll look at it tomorrow.

PR #1241, opened 2018-01-29
 Added the ability to provide tags and occurred time to the AddEvidence automation

dorsha:


dorsha:
@aviadl any idea why build fails?

PR #1242, opened 2018-01-30
Fix typo

Spivaka:
fixes https://github.com/demisto/etc/issues/10562


aviadl:
fixed already on another PR

PR #1243, opened 2018-01-30
move proofpoint test to nightly

meirwah:


dorsha:
@meirwah what was the problem? usage?

meirwah:
Not sure

dorsha:
we might have real problem.. no?

meirwah:
Let's open an issue an explore in parallel..
Not worth blocking content build on that

dorsha:
It seems like a real issue.

PR #1245, opened 2018-01-30
Entity enrichment generic

liorblob:
Please note that the playbook test takes quite a while, and there's a risk of timeout(on 240 currently). Maybe we should remove one input from it. Meanwhile I will run numerous builds. 
You can also view the playbook + test in demo2.


JonathanMeler:
fixes: https://github.com/demisto/etc/issues/10311

PR #1246, opened 2018-01-31
JAMF rewrite to Python

secdevopsai:
JAMF rewritten in Python.  Prepping for more functions such as checking FileVault encryption status

### DIFF
--- a/Integrations/integration-jamf.yml
+++ b/Integrations/integration-jamf.yml
@@ -20,12 +20,12 @@ configuration:
   required: false
 - display: Do not validate server certificate (insecure)
   name: insecure
-  defaultvalue: ""
+  defaultvalue: "true"
   type: 8
   required: false
 - display: Use system proxy settings
   name: proxy
-  defaultvalue: "true"
+  defaultvalue: "false"
   type: 8
   required: false
 script:
--- a/Integrations/integration-jamf.yml
+++ b/Integrations/integration-jamf.yml
@@ -53,18 +53,20 @@ script:
         endpoint = "/JSSResource/computers"
         try:
             contents = requests.get(base_url + endpoint,auth=auth,headers=HEADERS).json()
-            contents = [{"ID": item["id"], "Name":item["name"]} for item in contents['computers']]
+            ec = {}
+            ec["JAMF"] = [{"ID": item["id"], "Name":item["name"]} for item in contents['computers']]
 
-            demisto.results({ 'Type' : entryTypes['note'], 'ContentsFormat' : formats['json'], 'Contents' : contents, "HumanReadable": tableToMarkdown("table", contents)})
+            demisto.results({ 'Type' : entryTypes['note'], 'ContentsFormat' : formats['json'], 'Contents' : contents, "EntryContext": ec,"HumanReadable": tableToMarkdown("table", ec["JAMF"])})
         except Exception,e:
-            demisto.results({ 'Type' : entryTypes['error'], 'ContentsFormat' : formats['text'], 'Contents' : str(e) })
+            demisto.results({ 'Type' : entryTypes['error'], 'ContentsFormat' : formats['text'], 'Contents' : "Unable to complete command.  Please check credentials and/or proxy settings" })
             sys.exit(0)
 
     def jamf_get_computers_match(match):
         endpoint = "/JSSResource/computers/match/{match}"
         try:
             contents = requests.get(base_url + endpoint.format(match=match),auth=auth,headers=HEADERS).json()
-            contents = [
+            ec = {}
+            ec["JAMF"] = [
                 {"ID": item["id"],
                 "Name": item["name"],
                 "AltMAC": item["alt_mac_address"],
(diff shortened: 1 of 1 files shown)


Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
     - contextPath: Jamf.Computers.Name
       description: Computer name
+    description: Get a list of managed computers
   - name: jamf-get-computers-match
-    description: Get a list of managed computers that matches inputs query
     arguments:
     - name: match
       required: true
-      description: Match criteria. e.g. to retrieve all computers associated with a user use $UserID
+      description: Match criteria. e.g. to retrieve all computers associated with
```
anara123: Add example, link to docs which give an example of `match criteria`

Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
+    def jamf_get_computers():
+        endpoint = "/JSSResource/computers"
+        try:
+            contents = requests.get(base_url + endpoint,auth=auth,headers=HEADERS).json()
+            ec = {}
+            ec["JAMF"] = [{"ID": item["id"], "Name":item["name"]} for item in contents['computers']]
+
+            demisto.results({ 'Type' : entryTypes['note'], 'ContentsFormat' : formats['json'], 'Contents' : contents, "EntryContext": ec,"HumanReadable": tableToMarkdown("Computers", ec["JAMF"])})
+        except Exception,e:
+            demisto.results({ 'Type' : entryTypes['error'], 'ContentsFormat' : formats['text'], 'Contents' : "Unable to complete command.  Please check credentials and/or proxy settings" })
```
anara123: image
This error is misleading. You should show the real error exception to the user. 
I executed `!jamf-get-computers-match match="name=Shai-k"` for example, and I got `"Unable to complete command.  Please check credentials and/or proxy settings"`. After removing the `try except` I got 
```
File "<string>", line 462, in tableToMarkdown
 IndexError: list index out of range
```

Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
+                "Email": item["email"],
+                "EmailAddress": item["email_address"],
+                "Position": item["position"],
+                "Username": item["username"],
+                "Realname": item["realname"],
+                "Room": item["room"]}
+                for item in contents["computers"]
+            ]
+
+            demisto.results({ 'Type' : entryTypes['note'], 'ContentsFormat' : formats['json'], 'Contents' : contents, "EntryContext": ec,"HumanReadable": tableToMarkdown("Computers", ec["JAMF"])})
```
anara123: in case the `computers=[]` tableToMarkdown throws exception, need to check this case too.

anara123: Also change in all places from one line JSON/dict to 
```
demisto.results({ 
                'Type' : entryTypes['note'], 
                'ContentsFormat' : formats['json'], 
                'Contents' : contents, 
                "EntryContext": ec,
                "HumanReadable": tableToMarkdown("Computers", ec["JAMF"])
})
```

anara123: Be consistent, either use `'` single quote, or double `"`

Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
+        except Exception,e:
+            demisto.results({ 'Type' : entryTypes['error'], 'ContentsFormat' : formats['text'], 'Contents' : "Unable to complete command.  Please check credentials and/or proxy settings" })
+            sys.exit(0)
+
+    def jamf_get_computers_match(match):
+        endpoint = "/JSSResource/computers/match/{match}"
+        try:
+            contents = requests.get(base_url + endpoint.format(match=match),auth=auth,headers=HEADERS).json()
+            ec = {}
+            ec["JAMF"] = [
```
anara123: In the previous version the context path was `Jamf.Computers`, for backward compatibility the same should be here.

anara123: You also should do the `Jamf.Computers(val.ID==obj.ID)` otherwise you will have duplicate objects in the context

Integrations/integration-jamf.yml
anara123: This taken from the docs
`Name, mac address, etc. to filter by. Match uses the same format as the general search in Jamf Pro. For instance, admin* can be used to match computer names that begin with admin`

Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
+            sys.exit(0)
+
+    def jamf_computers_hardware(computer_id):
+        endpoint = "/JSSResource/computers/id/{computer_id}/subset/hardware"
+
+        try:
+            contents = requests.get(base_url + endpoint.format(computer_id=computer_id),auth=auth,headers=HEADERS).json()
+            ec = {}
+            ec["JAMF"] = {
+                "Make": contents["computer"]["hardware"]["make"],
```
anara123: your code should be more protective code. don't assume that `hardware` or `make` field will be there. Because if not, the code will throw exception. Instead you can use `dt`, like `"Make": dt(contents, "computer.hardware.make")`

Integrations/integration-jamf.yml
```diff
@@ -30,213 +30,164 @@ configuration:
...
-        return function(obj) {
-            if (obj instanceof Array) {
-                var res = [];
-                for (var j=0; j < obj.length; j++) {
-                    var current = transformSingleObj(obj[j]);
-                    if (current) {
-                        res.push(current);
-                    }
+
+            ec["JAMF"]["Storage"] = [
```
anara123: Same as above should be `Jamf.Computers(val.ID==obj.ID)` and make sure also that when calling this command twice, the `Storage` stay single object in the context

Itay4:
Closing and keeping branch alive.

PR #1253, opened 2018-01-31
cbp playbook test move

michalgold:


anara123:
fixes https://github.com/demisto/etc/issues/10588

PR #1256, opened 2018-02-01
Rerun test playbook if failed

anara123:
- run test playbook second time if for the first time it failed
- increase the timeout for getting incident (from 3 seconds to 10)
- pull docker images before the tests started to run


anara123:
fixes https://github.com/demisto/etc/issues/10499
fixes https://github.com/demisto/etc/issues/10498

PR #1274, opened 2018-02-04
Change Condition for DBotScore

JonathanMeler:
Using built-in conditional instead of DT
calculate_severity_-_generic_copy_sun_feb_04_2018


Playbooks/playbook-Calculate_Severity_-_Generic.yml
```diff
@@ -2,6 +2,7 @@ id: Calculate Severity - Generic
 version: -1
 name: Calculate Severity - Generic
 fromversion: 2.5.0
+releaseNotes: "-"
```
aviadl: From version is wrong

dorsha:
@aviadl please review again

PR #1275, opened 2018-02-05
Fixed root object of inputs

aviadl:
It will be inputs.<object name>
Will give more flexibility on array iteration.
Fixed also on playbook creation


asafshen:
please note that this is a bit fragile 
if someone copy the playbook & go into those tasks & save again - it will break

aviadl:
We fixed it in the product as well

PR #1277, opened 2018-02-05
CB Defense fixes

hirshbergshachar:
Fixed "change_device_status" command and added mandatory flag in add_rule and replace_policy where needed.


anara123:
@noaco please review

noaco:
@hirshbergshachar what about set-policy command? 
If I remember correctly, the command required a small fix.

hirshbergshachar:
@noaco Michal said that "set-policy" is not a main use case and asked to open an issue so she could look at that more thoroughly and see if we can take it off. The opened issue can be seen here -
 [https://github.com/demisto/etc/issues/10698](url). Therefore, at the moment, the version is supporting all of CBD's important usages.

PR #1279, opened 2018-02-05
Update integration-Zscaler.yml

K35537:
Changed 'zcaler' to 'zscaler'


dorsha:
We will need to keep zcaler for backward compatibility.

PR #1280, opened 2018-02-05
Update script-AddEvidence.yml

K35537:
Added a default timestamp of when the evidence is added to the board. I haven't found a good use case for manually entering a time for evidence to be found. Removed the 'occurred' arguments from the script.

Perhaps this could also just be added as a default option, while still allowing for an argument of 'occurred' to be entered.

### DIFF
--- a/Scripts/script-AddEvidence.yml
+++ b/Scripts/script-AddEvidence.yml
@@ -13,13 +13,19 @@ script: |-
   entryIDs = (Array.isArray(entryIDs)) ? entryIDs : [entryIDs];
   entryTags = (Array.isArray(entryTags)) ? entryTags.join(',') : entryTags;
   
-  // Get the month, day, and year. Sample output: "2008-01-02T15:04:05"
+  // Add zeros to single-digit date values for proper formatting.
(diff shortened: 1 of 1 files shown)


aviadl:
@K35537 Your code does look good, BUT the `when` date on an evidence, is not necessarily `now`
It should resemble the date that this `action` actually happened, it can be a few days back, but you got the evidence only now.
Think about old log files for example

K35537:
@aviadl That makes sense. Perhaps this is my unfamiliarity with Demisto, but could this be implemented as a default timestamp, unless an optional time is manually specified?

aviadl:
@K35537 This is actually a nice idea, maybe we can add a boolean whether to add `now` in case occurred was not supplied or not
Do you want to go for it?

dorsha:
In the next version of Demisto, if date is not specified it will set Now as the time

K35537:
I've tested this as working now by checking to see if the 'occurred' argument is present.

If it's empty, it will default to the current time.

dorsha:
Fixed.

PR #1282, opened 2018-02-06
Move VxStream test from skipped to tests

Itay4:
Move the test from skipped to tests after fixing test and changing the integration name to CrowdStrike.


Itay4:
Test fails cause of `crowdstrike-search` command, as described here: https://github.com/demisto/etc/issues/10721

Itay4:
@dorsha There are a few issues with VxStream we are working on with them. For now, it is better to keep the test skipped.

dorsha:
OK, closing, keeping the branch open.

PR #1283, opened 2018-02-06
increase timeout

erezh31:
fixes: https://github.com/demisto/etc/issues/10650

### DIFF
--- a/Scripts/script-CBLiveGetFile.yml
+++ b/Scripts/script-CBLiveGetFile.yml
@@ -4,7 +4,7 @@ commonfields:
 name: CBLiveGetFile
 script: |-
   import time
-  secTimeout = 60
+  secTimeout = 120
 
   # find sensor ID
   if not demisto.get(demisto.args(), 'ip') and not demisto.get(demisto.args(), 'hostname'):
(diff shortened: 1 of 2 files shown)


meirwah:
@anara123 @erezh31 can we merge this ?

Scripts/script-CBLiveGetFile.yml
```diff
@@ -147,4 +147,5 @@ dependson:
   - cb-command-create
   - cb-command-info
   - cb-file-get
-timeout: 0s
+timeout: 300ns
```
anara123: ns=nanoseconds ?

anara123:
@meirwah I run the build 3 times, if the these tests will not fail then we will merge

erezh31:
@anara123 not sure it's smart to do it parallel... the query go to the same carbon black env but let's see

Scripts/script-CBLiveGetFile.yml
erezh31: not sure... was wondering too. but this is the value when you change the automation timeout and download the yml.... suppose to in seconds

anara123:
it seems your tests passed in all 3 builds

PR #1285, opened 2018-02-06
Anomali ThreatStream DBot Score support

Itay4:
Fixes https://github.com/demisto/etc/issues/10206

### DIFF
--- a/Integrations/integration-AnomaliThreatStream.yml
+++ b/Integrations/integration-AnomaliThreatStream.yml
@@ -251,10 +251,10 @@ script:
         }
 
         entry.EntryContext.DBotScore = {
-                                        Indicator: args.value,
-                                        Type: args.type,
-                                        Vendor: 'ThreatStream',
-                                        Score: dbotScore
+            Indicator: args.value,
+            Type: args.type,
+            Vendor: 'ThreatStream',
+            Score: dbotScore
         };
 
         var entries = [entry];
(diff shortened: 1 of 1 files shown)


anara123:
@meirwah can you review it?

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -243,6 +243,20 @@ script:
...
+        var dbotScore = 1;
+        if (output['Threat Score'] >= args.threshold) {
+            dbotScore = 3;
+        } else if (output['Threat Score'] > 0) {
+            dbotScore = 2;
+        }
+
+        entry.EntryContext.DBotScore = {
+                                        Indicator: args.value,
+                                        Type: args.type,
```
meirwah: args.type ?
Those should be consts ...

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -243,6 +243,20 @@ script:
...
+                                        Vendor: 'ThreatStream',
```
anara123: fix the indentation

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -243,6 +243,20 @@ script:
             entry.EntryContext[contextKey] = contexts;
         }
 
+        var dbotScore = 1;
+        if (output['Threat Score'] >= args.threshold) {
```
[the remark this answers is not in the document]
Itay4: According to their documentation, 'Threat Score' will always be returned from every query.

Integrations/integration-AnomaliThreatStream.yml
Itay4: This is not really an argument. If you would look at the switch-case, you would find out that each 'type' of command is set into 'args'.

PR #1294, opened 2018-02-07
Fixed version

dorsha:


dorsha:
@asafshen changed, please check again.
Can you please update the wiki release page with this step? since it is not written there.

asafshen:
done

PR #1296, opened 2018-02-10
move phishing and detonate playbooks available in 3.1.0

anara123:
@dorsha @asafshen we need to release new content release.


Bars92:
No complain from no one so far. not so important IMO

anara123:
@Bars92 But we still need to release new content release

PR #1299, opened 2018-02-12
Removed a from version

dorsha:
@asafshen this cause us issues with our envs. let's talk about that.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,7 @@ test:
     - ./Tests/yamlvalidator
 
     # build content descriptor
-    - ./setContentDescriptor.sh $CIRCLE_BUILD_NUM 95f762855ccbbd6598d8b1d052eabe3d5e31610a 18.2.3a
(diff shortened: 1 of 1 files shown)


aviadl:
@asafshen please merge or explain your objections

asafshen:
Didn't know that it causes issues with our envs
what kind of issues?

I don't have major objections
the reason behind this is to allow people to upload content as a zip without messing up the next content release

lets take the following scenario:
- we release version 18.2.3
- after content release we clear rn and update version to 18.2.4
- someone wants to use current content master version in his env - so he/she takes the zip artifact from master & upload it. now its content version is 18.2.4
-  now when we actually release 18.2.4 (two weeks after) - this env will not get the latest version because this env already have 18.2.4

* I don't mid that to happen on local envs. I'm more concern with demo2/someone sending master zip co costumer (happened few times) 

maybe we should update to current number version - I'll check that it works later this week

PR #1300, opened 2018-02-12
Fix Type for hashes when no results are returned

Bars92:
In the VT Integration, if you scan a file hash and no results are returned,
the context adds an entry with Type 'file', however in all other cases it
adds Type 'hash' to the context. This fixes the behavior to consistently apply
'hash' in all cases.

related https://github.com/demisto/content/pull/1263


dorsha:
release notes missing

PR #1302, opened 2018-02-12
Add extract domain playbook test

meirwah:


meirwah:
@anara123 can you review

PR #1306, opened 2018-02-13
Proofpoint TAP integration

anara123:


anara123:
resolves https://github.com/demisto/etc/issues/5969

anara123:
[ProofPoint.zip](https://github.com/demisto/content/files/1719735/ProofPoint.zip)

Scripts/script-CommonServer.yml
```diff
@@ -8,6 +8,27 @@ script: |-
   // This script will be appended to each server script before being executed.
   // Place here all common functions you'd like to share between server scripts.
 
+  /**
```
shaniacht1: @anara123 better to do it same as String.prototype.format and with the documentation format. Otherwise it won't show in the script helper

anara123: fixed

Integrations/integration-ProofpointTAP.yml
```diff
@@ -0,0 +1,280 @@
...
+  name: isFetch
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Incident type
+  name: incidentType
+  defaultvalue: ""
+  type: 13
+  required: false
+- display: Server URL (TAP API URL)
```
Itay4: The parameters order is a bit odd. In my opinion, `Server URL` and `API Version` should be above `Credentials`. Also, there should be example for `Server URL` in its description.

Integrations/integration-ProofpointTAP.yml
```diff
@@ -0,0 +1,280 @@
...
+    switch(command) {
+        case 'test-module':
+            // the test will get the last 10 minutes events
+            // we just make sure the response is 200 and the username password are valid.
+            var now = new Date();
+            var last10minutes = new Date(now.getTime() - 10 * MINUTE_IN_MILLIS);
+            try {
+                getEventsRequest(null, dateToISO8601(last10minutes), null, null);
+            } catch (err) {
+                throw 'Test failed. Failed to fetch last 10 minutes events. Original error: ' + err
```
Itay4: Missing semicolon

Integrations/integration-ProofpointTAP.yml
```diff
@@ -0,0 +1,280 @@
...
+
+            var events = getEventsRequest(null, sinceTime, params.threatType, params.threatStatus);
+            var incidents = [];
+            if (events) {
+                // convert events to demisto incidents
+                if (events.messagesDelivered && events.messagesDelivered.length > 0) {
+                    for (var i = 0; i < events.messagesDelivered.length; i++) {
+                        var event = events.messagesDelivered[i];
+                        event.type = 'messages delivered';
+                        var incident = {
```
Itay4: Are there no `labels` ?

Integrations/integration-ProofpointTAP.yml
```diff
@@ -0,0 +1,280 @@
...
+            throw 'Request Failed.\nRequest URL: ' + requestUrl +'\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res) + '.';
+        }
+
+        return JSON.parse(res.Body);
+    }
+
+    function getEvents() {
+        var events = getEventsRequest(args.interval, args.sinceTime, args.threatType, args.threatStatus);
+
+        return {
```
Itay4: Why are there no HR and EC returned?

Scripts/script-CommonServer.yml
shaniacht1: @anara123 is it possible to also remove the "if"?(same as in string format)
The auto-doc doesn't recognize it as a function and so it's not being added to the doc.

anara123: it is better not be removed, this how polyfill should be written. But it can be done

Integrations/integration-ProofpointTAP.yml
anara123: This integration was intended to only fetch incidents.

Integrations/integration-ProofpointTAP.yml
```diff
@@ -0,0 +1,384 @@
+commonfields:
+  id: Proofpoint TAP
+  version: -1
+name: Proofpoint TAP
+display: Proofpoint TAP
+category: Network Security
```
NeelRust: @anara123 the category should be `Email Gateway`

Integrations/integration-ProofpointTAP.yml
anara123: no need labels

PR #1307, opened 2018-02-13
ExtractDomain to calcReputation be false by default

meirwah:


aviadl:
@meirwah I think we should leave the script as is.
Fixed the flow in the playbook, see other PR

JonathanMeler:
was fixed in https://github.com/demisto/content/pull/1321

PR #1320, opened 2018-02-15
Added description to CustomFields argument

dem-evgeny:


dem-evgeny:
Fixes: https://github.com/demisto/etc/issues/10520

dem-evgeny:
@gshriki A little bit problematic because if for example the name itself will have u_ so it could have u_u_blabla...

Integrations/integration-ServiceNow.yml
```diff
@@ -828,7 +828,8 @@ script:
     - name: template
       description: Template name to use as a base to create new incidents.
     - name: custom_fields
-      description: 'Custom fields in this format: fieldname=value;fieldname=value...'
+      description: 'Custom fields in this format: fieldname=value;fieldname=value...
+        (You don''t need to add "u_" in the name of the custom field. It adds it automatically.'
```
anara123: shouldn't `don''t`  be `don\'t`?

dem-evgeny: I wrote 
You don't need to add "u_" in the name of the custom field. It adds it automatically.
Demisto added the duplicate ''

gshriki:
ok. so why would the user add it? where does he see it?

dem-evgeny:
@gshriki If it is a power user that is looking in the JSON and not in the GUI. Only in the JSON it adds "u_". I had some user like this recently.

gshriki:
Agreed not to merge

PR #1323, opened 2018-02-15
McAfeeDAMbugfix

GavrielFilippov:
bug in human readble response, first line from the output was missing and additionally some table was corrupted


GavrielFilippov:
@gshriki the only thing ive changed is lines 204, 208 and release notes, why do i see more changes ?
version ? im on 3.1 with latest content

PR #1327, opened 2018-02-18
check if module is active - changed logic + cbp bug fix

guydemi:
check if module is active - changed logic
fixed a bug in cbp where no computer ids were found

### DIFF
--- a/Playbooks/playbook-Search_Endpoints_By_Hash_-_TIE.yml
+++ b/Playbooks/playbook-Search_Endpoints_By_Hash_-_TIE.yml
@@ -37,10 +37,10 @@ tasks:
       }
   "1":
     id: "1"
-    taskid: aa03817d-efd2-4b56-84ae-d785dae24cd9
+    taskid: f96c54b2-b8fa-48b8-8a56-56555a749c10
     type: condition
     task:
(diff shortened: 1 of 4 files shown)


Playbooks/playbook-Search_Endpoints_By_Hash_-_TIE.yml
```diff
@@ -37,10 +37,10 @@ tasks:
...
   "1":
     id: "1"
-    taskid: 58adf1e3-dbee-46b1-86b3-c8f9c3e2dff8
+    taskid: aa03817d-efd2-4b56-84ae-d785dae24cd9
     type: condition
     task:
-      id: 58adf1e3-dbee-46b1-86b3-c8f9c3e2dff8
+      id: aa03817d-efd2-4b56-84ae-d785dae24cd9
       version: -1
       name: Is TIE enabled?
```
JonathanMeler: Missing the TIE and ePO checks

guydemi: TIE still exists, see in the file itself (it doesn't show in the diff because the same text snippet was already present), epo is indeed missing - added now

PR #1330, opened 2018-02-19
McAfee-ePObugfix

GavrielFilippov:
bug in epo-command when result is a string, the output is not presented


Integrations/integration-epo.yml
```diff
@@ -386,4 +387,4 @@ script:
   - name: epo-command
     arguments: []
     description: Execute ePO command
-hidden: false
+releaseNotes: "bug fix for epo-command when result is a string"
```
gshriki: PLease rephrase release notes:
'Better handling of string-type outputs by ePO'

PR #1331, opened 2018-02-19
Change categories

asafshen:
changed categories naming by @dansarel2 requests


asafshen:
We will still have 3 “Data enrichment” go integrations (whois, AD query & Threat Central)
will change them in 3.6 release

PR #1333, opened 2018-02-19
PostgreSQL - closing connection after executing the statement

gshriki:


Integrations/integration-PostgreSQL.yml
```diff
@@ -84,7 +84,7 @@ script:
             table.append(tableRow)
 
         demisto.results( {'ContentsFormat': formats['json'], 'Type': entryTypes['note'], 'Contents': contents, "ReadableContentsFormat": formats['table'], "HumanReadable": table} );
-
+        conn.close()
```
dorsha: You should also do cur.close() before

image

dorsha:
You should also close the connection in the Test mode........ 

image

dorsha:
and not only in the command...
Please create a function for closing the connection, so everyone will call it if needed, it will be better for maintenance later.

slavikm:
As usual, for languages that support it, the right way to do this is using `finally`.

dorsha:
Beside the above comments, another issue I found with this fix is that if you have **syntax** error, the connection was not closed as well..

Anyhow, I fixed all of my code review comments and issues.

PR #1340, opened 2018-02-21
Remedy AR integration

Itay4:
Fixes https://github.com/demisto/etc/issues/10710


Integrations/integration-Remedy_AR.yml
```diff
@@ -0,0 +1,150 @@
...
+- display: Credentials
+  name: credentials
+  defaultvalue: ""
+  type: 9
+  required: true
+- display: Do not validate server certificate (insecure)
+  name: insecure
+  defaultvalue: "false"
+  type: 8
+  required: false
```
anara123: what about proxy?

Integrations/integration-Remedy_AR.yml
```diff
@@ -0,0 +1,150 @@
...
+                verify=USE_SSL,
+                data=data,
+                headers=headers
+            )
+            if res.status_code not in (200, 204):
+                raise Exception('Your request failed with the following error: ' + res.reason)
+        except Exception, e:
+            LOG(e)
+            raise
+        return res
```
anara123: don't you need to return JSON here like `res.json()`?

Integrations/integration-Remedy_AR.yml
```diff
@@ -0,0 +1,150 @@
...
+            res = requests.request(method,
+                BASE_URL + url_suffix,
+                verify=USE_SSL,
+                data=data,
+                headers=headers
+            )
+            if res.status_code not in (200, 204):
+                raise Exception('Your request failed with the following error: ' + res.reason)
+        except Exception, e:
+            LOG(e)
```
anara123: No need to `LOG(e)` here, because the exception will be printed in the `try catch` at the bottom

Integrations/integration-Remedy_AR.yml
Itay4: Nope, I need the result as is for generating the token.

anara123:
@Itay4 Also you should redirect users to Remedy On Demand integration in case they will look for other commands like create incident

dorsha:
@Itay4 / @anara123 please handle this PR.

dorsha:
@Itay4 don't forget the documentation ,didn't see it in the issue nor here.

dorsha:
@Itay4 don't forget the documentation (couldn't find it)

Itay4:
Documentation - https://github.com/demisto/etc/issues/10845

Integrations/integration-Remedy_AR.yml
```diff
@@ -0,0 +1,162 @@
...
+            'EntryContext': context
+        }
+        return entry
+
+    ''' EXECUTION CODE '''
+    auth = login()
+    token = auth.content
+    DEFAULT_HEADERS['Authorization'] = 'AR-JWT ' + token
+
+    LOG('command is %s' % (demisto.command(), ))
```
dorsha: do we want this log? I think it can be removed.

Itay4: It's good to have, cause in case of error, we'll be able to know in which command it occured. We are using it in all recent Python integrations.

PR #1343, opened 2018-02-22
Archer integration improvements

hirshbergshachar:
Archer now gets a record's inner records in getRecord & fetch-incidents
fixes https://github.com/demisto/etc/issues/10752


anara123:
@dorsha what is the status of this PR?
@hirshbergshachar Please delete the `fetchFields` param, we always should return all the fields. And I also think we should remove `fetchByField`, this should be always be Date Created

Integrations/integration-ArcherRSA.yml
```diff
@@ -1030,16 +1374,29 @@ script:
         }
     }
 
+    function reLogin(token){
+        destroyAuthToken(token);
+        return getAuthToken();
+    }
+
+    function test() {
```
anara123: You need to do some request, run some command so that the test be reliable

hirshbergshachar:
@anara123

anara123:
It should be merged soon.waiting for buid to be green

PR #1347, opened 2018-02-22
Malware Investigation - Generic

JonathanMeler:
### Malware Investigation - Generic
malware_investigation_-_generic_thu_feb_22_2018

### Malware Investigation - Generic - Setup
malware_investigation_-_generic_-_setup_thu_feb_22_2018


JonathanMeler:
fixes: https://github.com/demisto/etc/issues/7809

PR #1354, opened 2018-02-24
parse the _raw part of splunk event

meirwah:
fixes https://github.com/demisto/etc/issues/10869


meirwah:
I explained this to @NeelRust before too, this will not hold for all cases... that is why I preferred to avoid it.

It worked for dow POC.. but will not work for others...

If you think it's wrong approach I will close this PR.
and maybe use this code as an optional automation to run in playbook, WDYT?

slavikm:
@meirwah I think we should try and do the right thing by default - best effort. If it does not work, user can always disable it, no?

meirwah:
i made it optional...
 by default it's disabled

orius123:
If we parse `_raw` we can handle also non-notable events. no?

orius123:
I have code that I used for two customers that parse `_raw` from comma separated list with `:` for splitting pairs. don't know if we can know what the format will be..

Integrations/integration-SplunkPy.yml
```diff
@@ -412,5 +444,16 @@ script:
...
+      description: The Raw data of splunk event, in string format
+      defaultValue: ${Splunk.Result._raw}
+    outputs:
+    - contextPath: Splunk.Raw.Parsed
+      description: The raw event data parsed
+      type: unknown
+    description: Parse the Raw part of the event
   dockerimage: demisto/splunksdk:1.0
   isfetch: true
+releaseNotes: ''
```
orius123: mentions the new command

Integrations/integration-SplunkPy.yml
```diff
@@ -412,5 +444,16 @@ script:
     - contextPath: Splunk.Job
       description: The created jon SID
     description: 'Create new search job in Splunk '
+  - name: splunk-parse-raw
+    arguments:
+    - name: raw
```
orius123: make default

Integrations/integration-SplunkPy.yml
```diff
@@ -67,6 +72,20 @@ script:
     import urllib3
     urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
+    def rawToDict(raw):
+        result = {}
+        raw = raw.strip("}")
```
orius123: Since format may defer add some checks / error handling, or wrap in a try catch

meirwah: strip doesn't fail (throw error)

meirwah:
@orius123 can you review this?

PR #1362, opened 2018-02-25
PGSQL - set to auto commit

gshriki:
Setting auto commit to true

### DIFF
--- a/Integrations/integration-rasterize.yml
+++ b/Integrations/integration-rasterize.yml
@@ -41,7 +41,7 @@ script:
             friendlyName = 'email.pdf'
 
         f = open('htmlBody.html','w')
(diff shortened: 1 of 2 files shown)


Integrations/integration-rasterize.yml
```diff
@@ -41,7 +41,7 @@ script:
             friendlyName = 'email.pdf'
 
         f = open('htmlBody.html','w')
-        f.write(html)
+        f.write('<html style="background:white";>' + html + '</html>')
```
dorsha: I think it is not related, try to checkout from master

gshriki: no idea why is this here. will fix

PR #1365, opened 2018-02-25
fix incidents-top-close-analysts widget

meirwah:


meirwah:
fixes https://github.com/demisto/etc/issues/10889

PR #1366, opened 2018-02-25
Support for conf file (remove hypersearch)

asafshen:
Add server configuration support

server PR - https://github.com/demisto/server/pull/10439
content test PR - https://github.com/demisto/content-test-conf/pull/49

this will pass only after run on green server master
NOTE - merge together with content-test PR


asafshen:
will re-poen after stabilizing

PR #1368, opened 2018-02-27
Fix integration properties

yaakovi:
fixes: https://github.com/demisto/etc/issues/4702


Integrations/integration-OpenPhish.yml
```diff
@@ -3,7 +3,8 @@ commonfields:
   version: -1
 name: OpenPhish
 display: OpenPhish
-category: Data Enrichment & Threat Intelligence
+releaseNotes: "Added DBotScore 0 in case there is no score from OpenPhish"
```
dorsha: Why release notes?

Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -2,7 +2,7 @@ commonfields:
   id: Symantec Endpoint Protection
   version: -1
 name: Symantec Endpoint Protection
-display: Symantec Endpoint Protection
+display: Symantec Endpoint Protection 14
```
dorsha: 14?

yaakovi: that's what @NeelRust said in the issue

PR #1369, opened 2018-02-27
Splunk test parse raw

meirwah:


meirwah:
@asafshen plz review

Integrations/integration-SplunkPy.yml
```diff
@@ -82,8 +82,11 @@ script:
           single_key_val = key_val.split("=")
           if len(single_key_val)>1:
               val = single_key_val[1]
+              val = val.strip("\\")
               val = val.strip("\"")
-              result[single_key_val[0]] = val
+              val = val.strip("\\")
```
asafshen: is the double line `val = val.strip("\\")` intended?

PR #1376, opened 2018-03-01
stable Test-IsMaliciousIndicatorFound

meirwah:


meirwah:
fixes https://github.com/demisto/etc/issues/10942

PR #1377, opened 2018-03-01
added the Mattermost equivalent of SlackAskUser

guydemi:
Also updated SendMessageToOnlineUsers with Mattermost


guydemi:
@aviadl after discussing with @gshriki, the newly added script will have dependson mattermost-send and the old script will have the same behaviour as if slack didn't exist (informative error message)

Scripts/script-MattermostAskUser.yml
```diff
@@ -0,0 +1,57 @@
...
+- name: persistent
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: Indicates wether to use one-time entitlement or a persistent one
+  defaultValue: "false"
+scripttarget: 0
+runonce: false
+releaseNotes: "Added the Mattermost equivalent for SlackAskUser"
```
gshriki: Since this is a new script, I do not think we need release notes, as it will take the script description.
Please confirm.
If we take the release notes it should be what we have in description and not what is there now...

PR #1379, opened 2018-03-04
The report file pushed to WarRoom as a file and not as a link.

dem-evgeny:
Now, the report file pushed to WarRoom as a file and not as a link.


dem-evgeny:
@anara123 @Spivaka Can anyone of you approve it?

PR #1380, opened 2018-03-05
support for new file handling

yuvalbenshalom:
removing the investigation id added to the random file name

### DIFF
--- a/Scripts/script-CommonServerPython.yml
+++ b/Scripts/script-CommonServerPython.yml
@@ -1,6 +1,7 @@
 commonfields:
   id: CommonServerPython
   version: -1
+  "fromVersion": "3.6.0"
 name: CommonServerPython
 script: |-
   # Common functions script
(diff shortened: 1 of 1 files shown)


Scripts/script-CommonServerPython.yml
```diff
@@ -1,6 +1,7 @@
 commonfields:
   id: CommonServerPython
   version: -1
+  "fromVersion": "3.6.0"
```
yaakovi: this will cause error for all content release until version 3.6.
this file will be missed from the content.zip and crash all (or at least most of ) python integrations/scripts.

possible solutions:
1) duplicate the original file with "toversion"
2) remove this label and apply the change for all content release
** however other code may depend on the current name convention

@dorsha @anara123

anara123:
@yuvalbenshalom what is the issue that related to this PR?

dorsha:
@yuvalbenshalom is this relevant?

yuvalbenshalom:
@dorsha no, we don't need it. It was solved without changing the content

PR #1388, opened 2018-03-06
Cybereason isolate and un-isolate machines commands

Itay4:
Fixes https://github.com/demisto/etc/issues/10102
Work in progress


anara123:
@Itay4 solve the conflict and merge

PR #1389, opened 2018-03-06
tableToMarkdown fix with "|" char

erezh31:
fixes:
https://github.com/demisto/etc/issues/9053

### DIFF
--- a/Scripts/script-CommonServerPython.yml
+++ b/Scripts/script-CommonServerPython.yml
@@ -254,7 +254,7 @@ script: |-
   def formatCell(data, is_pretty=True):
       l_format = '\n' if is_pretty else ''
       if type(data) in STRING_TYPES:
-          return data
+          return data.replace("|","\\|")
       elif isinstance(data, dict):
           return l_format.join([k + ': ' + flattenCell(data[k], is_pretty) for k in data])
       else:


Scripts/script-CommonServerPython.yml
```diff
@@ -254,7 +254,7 @@ script: |-
   def formatCell(data, is_pretty=True):
       l_format = '\n' if is_pretty else ''
       if type(data) in STRING_TYPES:
-          return data
+          return data.replace("|","\\|")
```
yaakovi: this will only solve the "|" character, what about the rest of them? the issue mentions other characters as well

meirwah:
Also compare to JS impl(of escaping): https://github.com/demisto/content/blob/master/Scripts/script-CommonServer.yml#L181

Scripts/script-CommonServerPython.yml
erezh31: I know... but I don't think we want to escape other chars, and I don't want to do many changes

erezh31:
@meirwah it's more visual, so testing need to do manual.
I did a basic test - seems work

anara123:
image
image
It seems that all the values are one row slided up. ID should 1 for example. I think this issue is related to this PR

anara123:
@erezh31 And I agree with @meirwah that we need here a test playbook, which will test the markdown. 
Also I got another exception when I pass empty array `[]` to tableToMarkdown.

erezh31:
I think this PR is not relevant. @yaakovi said that he made a fix that includes that

anara123:
@erezh31 @yaakovi can we merge this pr?

erezh31:
I think @yaakovi already made a change so we don't need it.
@yaakovi?

erezh31:
not necessary due to:
https://github.com/demisto/content/pull/1471

PR #1390, opened 2018-03-06
rename pyEWS

erezh31:


Integrations/integration-EWSv2.yml
```diff
@@ -1,8 +1,8 @@
 commonfields:
-  id: pyEWS
+  id: EWSv2
   version: -1
-name: pyEWS
-display: pyEWS
+name: EWSv2
+display: EWSv2
```
asafshen: I think EWS-v2 or EWS V2 will look better

PR #1395, opened 2018-03-07
fixes case that parseNotableEventsRaw not setup yet

meirwah:


Integrations/integration-SplunkPy.yml
```diff
@@ -1,8 +1,8 @@
 commonfields:
-  id: SplunkPy
+  id: SplunkPyx
   version: -1
-name: SplunkPy
-display: SplunkPy
+name: SplunkPyx
```
orius123: Change back

Integrations/integration-SplunkPy.yml
```diff
@@ -168,11 +168,12 @@ script:
         incident["occurred"] = event["_time"]
         incident["rawJSON"] = json.dumps(event)
         labels = []
-        isParseNotableEventsRaw = demisto.params()['parseNotableEventsRaw']
-        if isParseNotableEventsRaw:
-            rawDict = rawToDict(event['_raw'])
-            for rawKey in rawDict:
-                labels.append({'type':rawKey,'value':rawDict[rawKey]})
+        if demisto.get(demisto.params(),'parseNotableEventsRaw'):
```
orius123: I think you can just add a `default` for `get` function

meirwah: is it supported?
Anyway it doesn't matter since default is false

meirwah:
@orius123 plz review so we can release content

PR #1404, opened 2018-03-12
fix sort on single element issue

liorblob:
if a single element is returned, convert it to array.
Solves https://github.com/demisto/etc/issues/10861


Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -523,3 +527,4 @@ script:
       type: number
     description: Checks the reputation of the given IP
 hidden: false
+releaseNotes: "Fix parse result sort issue"
```
anara123: release notes are unnecessary here. `releaseNotes: "-"`

Integrations/integration-Remedy_AR.yml
```diff
@@ -181,3 +181,4 @@ script:
       type: string
     description: Retrieves server details
 hidden: false
+releaseNotes: "-"
```
anara123: no need. remove this change and pull from master

anara123:
@liorblob open new issue for documentation and put here a link for that issue

liorblob:
Documentation: https://github.com/demisto/etc/issues/11057
Need to add outputs to the commands

anara123:
@liorblob You mean some commands missing outputs?

Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -266,6 +266,10 @@ script:
     }
 
     var parseResult = function(command, result) {
+        if(!(result['objects'] instanceof Array)){
```
asafshen: what about `parseReputationResult`?
```
var objects = dq(result, 'objects');
objects.sort(function(x, y) {return x.created_ts < y.created_ts});
```

also - I think we should handle a case that the return value is not  an object, and not trust the api to return an object (`res.Body`)

liorblob: `parseReputationResult` is always called from `parseResult`, the bug was actually there but the source is from `parseResult`. And I didn't fully understand - did you mean the `objects` field in the result not being returned at all or the result itself not being an object?

asafshen: thanks for the `parseReputationResult` clarification

regrading the second thing - if `result` (which is the body of the returned request) is not an object - we will also get an exception here

liorblob: Added checks for `result`

PR #1406, opened 2018-03-13
[ZipFile] ability to create a password protected zip file

hirshbergshachar:
Solving issue https://github.com/demisto/etc/issues/10911

### DIFF
--- a/Scripts/script-ZipFile.yml
+++ b/Scripts/script-ZipFile.yml
@@ -6,18 +6,24 @@ script: |
   import zipfile
   import os
   from os.path import isfile
+
   try: # in order to support compression of the file
       import zlib
       compression = zipfile.ZIP_DEFLATED
   except:
       compression = zipfile.ZIP_STORED
+  import pyminizip
+
   filePath = None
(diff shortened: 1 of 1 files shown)


Scripts/script-ZipFile.yml
```diff
@@ -6,19 +6,25 @@ script: |
   import zipfile
   import os
   from os.path import isfile
+
   try: # in order to support compression of the file
       import zlib
       compression = zipfile.ZIP_DEFLATED
   except:
       compression = zipfile.ZIP_STORED
+  import pyminizip
```
anara123: Can't we do zip with password in python native way? If not, do you created docker image which uses `pyminizip`  package?

hirshbergshachar: If by the native way you refer to Python's zipfile, it doesn't support encryption at the moment: "It supports decryption of encrypted files in ZIP archives, but it currently cannot create an encrypted file". 

I've created the docker image only locally, it is not created automatically at new content release?

anara123:
Fix the build

PR #1410, opened 2018-03-15
Add output

K35537:
Added output as a Boolean with the contextPath of "in_whitelist"

This gives you a True/False to pivot off of once the whitelist has been checked.

### DIFF
--- a/Scripts/script-CheckWhitelist.yml
+++ b/Scripts/script-CheckWhitelist.yml
@@ -9,10 +9,11 @@ script: |
   else:
       whitelist = res[0]['Contents']
       if demisto.args()['item'] in whitelist:
-          answer = 'yes'
+          answer = True
       else:
-          answer = 'no'
+          answer = False
       demisto.results(answer)
+      demisto.setContext('in_whitelist', answer)
 type: python
 tags:
 - whitelist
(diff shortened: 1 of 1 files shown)


Bars92:
Hello @K35537! Thanks for the PR!
Im afraid we will have to still return 'yes', 'no' as a result to allow for backward compatibility. But the context is a good idea. I suggest something like this:
```
contextKey = "WhitelistCheck"
contextValue = demisto.get(demisto.context(), contextKey)
contextDic = { demisto.args()['item']: answer }
if contextValue:
    contextValue.update(contextDic)
    demisto.setContext(contextKey, contextValue)
else:
    demisto.setContext(contextKey, contextDic)
```

This will add to `WhitelistCheck` in the context the indicator and the results while keeping old results. Let me know if you find this ok and we can fix this PR.

K35537:
@Bars92 that sounds good to me!

Bars92:
@K35537 thanks for the PR! we will continue it in https://github.com/demisto/content/pull/1412.

K35537:
@Bars92  would you like me to delete this branch?

Bars92:
yes @K35537, thanks you!

PR #1414, opened 2018-03-16
Extract HTML tables data into context

slavikm:


anara123:
@slavikm Please add test playbook

slavikm:
@anara123 please verify my test playbook is according to our standard

TestPlaybooks/playbook-TestExtractHTMLTables.yml
```diff
@@ -0,0 +1,445 @@
...
+      iscommand: false
+      brand: ""
+    nexttasks:
+      '#default#':
+      - "5"
+      "yes":
+      - "6"
+    scriptarguments:
+      value:
+        simple: '${HTMLTables=>val[0] == ''a'' && val[1] == ''1'' ? ''x'' : ''''}'
```
anara123: `'${HTMLTables=>val[0] == ''a'' && val[1] == ''1'' ? ''x'' : ''''}'` should be `'${HTMLTables=val[0] == ''a'' && val[1] == ''1'' ? ''x'' : ''''}'`
replace `=>` to `=` in all the playbook

TestPlaybooks/playbook-TestExtractHTMLTables.yml
```diff
@@ -0,0 +1,445 @@
...
+        }
+      }
+  "5":
+    id: "5"
+    taskid: 50da60bb-61ac-4ad6-8f89-8f43eb32cb9f
+    type: regular
+    task:
+      id: 50da60bb-61ac-4ad6-8f89-8f43eb32cb9f
+      version: -1
+      name: Bad
```
anara123: This Bad task is regular manual task, so if the playbook will reach here, nothing will happen. Playbook should be in error state in order to fail.
So you can add automation script like Equals 0=1 (always false)

anara123:
@slavikm Your test is not working. And you should add the playbook to conf.json

TestPlaybooks/playbook-TestExtractHTMLTables.yml
slavikm: Why??? `=>` is the correct form in this case because I want the array

TestPlaybooks/playbook-TestExtractHTMLTables.yml
slavikm: Will do - did not remember if we actually need to fail

TestPlaybooks/playbook-TestExtractHTMLTables.yml
anara123: You right

PR #1418, opened 2018-03-21
Custom filters and transfomers

asafshen:
examples for custom filters & transformers
- InRange (number filter)
- ReverseList (entire-list transformer - e.i. get the whole list as an argument rather then works on each argument)
- StripChars - (string transformer)strip chars from prefix/suffix of the string

### DIFF
--- a/Scripts/script-InRange.yml
+++ b/Scripts/script-InRange.yml
@@ -1,0 +1,51 @@
+commonfields:
+  id: InRange
+  version: -1
+name: InRange
+script: |
+  INVALID_ARG = "INVALID_ARG"
+  def to_float(s):
+      try:
+          return float(s)
+      except:
+          return INVALID_ARG
+
+  # "1,8" => will return a tupple (1, 8)
+  def parse_range(rangeStr):
+      splitted = rangeStr.split(",")
+      if len(splitted) < 2:
+          return INVALID_ARG, INVALID_ARG
+
+      # parse
+      return to_float(splitted[0]), to_float(splitted[1])
+
+  leftArg = demisto.args()["left"]
+  rightArg = demisto.args()["right"]
+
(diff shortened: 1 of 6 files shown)


dorsha:
@meirwah ?

Scripts/script-InRange.yml
```diff
@@ -0,0 +1,51 @@
...
+enabled: true
+args:
+- name: left
+  required: true
+  description: value to check if in range (e.g. 4)
+- name: right
+  required: true
+  description: range to check against with the form of from,to (e.g. 2,8)
+scripttarget: 0
+runonce: false
```
meirwah: no need for this right?

Scripts/script-ReverseList.yml
```diff
@@ -0,0 +1,31 @@
...
+  the argument as a list  (note the \"entirelist\" tag)\n "
+enabled: true
+args:
+- name: value
+  required: true
+  default: true
+  description: List to reverse
+  isArray: true
+scripttarget: 0
+runonce: false
```
meirwah: no need for this right?

Scripts/script-StripChar.yml
```diff
@@ -0,0 +1,26 @@
...
+enabled: true
+args:
+- name: value
+  required: true
+  description: text to trim (e.g. "~!!~www.mydomain.com~!~!~")
+- name: chars
+  required: true
+  description: charecters to stripped from the text (e.g. "!~")
+scripttarget: 0
+runonce: false
```
meirwah: no need for this right?

Scripts/script-InRange.yml
asafshen: it imported automatically with the script
did we remove this feature?

PR #1423, opened 2018-03-22
better handling of non utf chars

gshriki:
Added section to strip non-utf chars, so script doesn't fail
When non utf chars are stripped - original is stored as file so it can be parsed and get IOC extracted from

### DIFF
--- a/Scripts/script-ParseEmailFiles.yml
+++ b/Scripts/script-ParseEmailFiles.yml
@@ -33,7 +33,7 @@ script: |-
   import olefile as OleFile
   import random
   import string
-  
+
   reload(sys)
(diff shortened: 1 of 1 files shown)


Scripts/script-ParseEmailFiles.yml
```diff
@@ -476,23 +477,38 @@ script: |-
           elif a.shortFilename:
               fname = a.shortFilename
           if fname and a.data:
-              fname = fname.encode('utf-8')
+              fname = fname.encode('utf-8').strip()
```
aviadl: we probably want to do it here as well, no ?

PR #1424, opened 2018-03-23
Add Intezer integration

Nitzperetz:
`analyze-by-sha256` implemented


Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+        while (res.status !== SUCCESS_STATUS && tries < maxRetries) {
+            // this means that status is 'in_progress' or 'queued'
+            tries++;
+            wait(delay); // wait {delay} seconds
+            res = getAnalysis(analysisId);
+        }
+
+        // if polling result did not work, return an error message
+        if (res.status !== SUCCESS_STATUS) {
+            throw 'Failed to pool sha256 analysis for ' + hash + ', Try to change maxRetries or delay arguments\n';
```
gshriki: not sure we want to fail in case we don't have .a response yet. @slavikm - wdyt? what did you do in other services? I remember you did some polling as well...

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+                Method: method,
+                Body: method == 'POST' ? JSON.stringify(parameters) : ''
+            },
+            false,
+            params.useproxy
+        );
+
+        // validate response
+        if (result.StatusCode < 200 || result.StatusCode > 299) {
+            switch (result.StatusCode) {
```
gshriki: In all errors - if we get additional details from the service - let's also print them. Especially in the 5xx codes and the default...

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+                dbotScore = 3;
+                var malFile = {};
+                addMalicious(malFile, outputPaths.file, {
+                    SHA256: hash,
+                    Malicious: {Vendor: 'Intezer', Metadata: res.result}
+                });
+                ec[outputPaths.file].push(malFile[outputPaths.file]);
+            } else if (verdict === 'suspicious') {
+                dbotScore = 2;
+            } else if (verdict === 'trusted' || verdict === 'neutral') {
```
gshriki: is neutral really good from the service point of view? or is it unknown?

Integrations/integration-Intezer.yml
gshriki: would also add service name to errors, so it is clear (sometimes when you do `!file` it's hard to know which service failed.

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+      required: true
+      default: true
+      description: Hash of the file to query. Supports SHA256
+    - name: codeItemType
+      auto: PREDEFINED
+      predefined:
+      - file
+      - memory_module
+      - fileless_code
+      description: '(String, Optional) - If the code item is a regular file, a memory
```
gshriki: Just - `Item type - File (file), Memory Module (memory_module) or Memory Fileless Code (fileless_code)`.

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+        case 'test-module':
+            // send is available
+            return sendRequest('GET', IS_AVAILABLE_URL).obj.is_available;
+        case 'file':
+            return analyzeSHA256(args.file, args.codeItemType, args.maxRetries, args.delay);
+        default:
+    }
+  type: javascript
+  commands:
+  - name: file
```
gshriki: in all argument description - we don't need the type and optional/mandatory at the beginning. optional/mandatory is shown in the interface. type should be derived from the description.

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+      - memory_module
+      - fileless_code
+      description: '(String, Optional) - If the code item is a regular file, a memory
+        module or a memory fileless code. Values: file,memory_module,fileless_code'
+      defaultValue: file
+    - name: maxRetries
+      description: (Number) - maximum number of retries trying to get an analysis
+      defaultValue: "60"
+    - name: delay
+      description: (Number) - delay in seconds between requests for analysis result
```
gshriki: Delay in seconds between polling of analysis results

Integrations/integration-Intezer.yml
```diff
@@ -0,0 +1,221 @@
...
+      auto: PREDEFINED
+      predefined:
+      - file
+      - memory_module
+      - fileless_code
+      description: '(String, Optional) - If the code item is a regular file, a memory
+        module or a memory fileless code. Values: file,memory_module,fileless_code'
+      defaultValue: file
+    - name: maxRetries
+      description: (Number) - maximum number of retries trying to get an analysis
```
gshriki: Number of retries for polling the analysis report

Integrations/integration-Intezer.yml
Nitzperetz: @gshriki , this is after waiting for [delay * maxRetries].

Integrations/integration-Intezer.yml
Nitzperetz: image

Integrations/integration-Intezer.yml
Nitzperetz: Yes, worked on that with someone from their team.

Nitzperetz:
@gshriki , fixed comments.
About the response errors - I took the descriptions from the documentation.
About the error after the polling, I think it's ok, let me know what you think about it.

PR #1429, opened 2018-03-27
New artifacts - layouts, classifiers & incident-fields

asafshen:
fixes: https://github.com/demisto/etc/issues/11010

Support for layouts, classifiers & incident-fields artifacts

major change (code addition) is in the releaes notes


release_notes.py
```diff
@@ -243,14 +261,123 @@ def modifiedReleaseNotes(self, cnt):
...
+
+        return self.getReleaseNotes(cnt)
+
+    def modifiedReleaseNotes(self, cnt):
+        rn = cnt.get("releaseNotes", "")
+
+        if len(rn) == 0:
+            return None
+
+        if len(rn) > 0 and rn == "-":
```
yaakovi: the first condition is always True here.

PR #1430, opened 2018-03-27
fixed server_get_artifact script

yaakovi:
get the latest successful server build

PR #1431, opened 2018-03-28
change version 3.6 to 3.5.1

meirwah:


dorsha:
@meirwah build fails.

meirwah:
it fails on issue that should be solved in: https://github.com/demisto/content/pull/1430

PR #1433, opened 2018-03-28
Properly escape URLs on url command

aviadl:


gshriki:
Fixes https://github.com/demisto/etc/issues/11263

PR #1434, opened 2018-03-28
Carbon black test fix + server_get_artifact fix

Bars92:
server_get_artifact fix from https://github.com/demisto/content/pull/1430#pullrequestreview-107768666.

removed carbon black test and fixes server get artifact changes.


meirwah:
@anara123 I think we should merge this.. to make build green again.. 
CB tests are not working as expected .. need to fix them in separate PR

PR #1435, opened 2018-03-29
Extract fields splunk

meirwah:
Support Splunk fetch incident to extract custom fields from _raw of notable events


NeelRust:
It works slavik. we tested in our own and one external env

PR #1439, opened 2018-04-01
Where is the egg?

dorsha:
from version 3.6.0, scriptNotVisible and defaultEnabled


Integrations/integration-WhereIsTheEgg.yml
```diff
@@ -0,0 +1,66 @@
+commonfields:
+  id: Where is the egg?
+  version: -1
+name: Where is the egg?
+display: Where is the egg?
+fromVersion: 3.6.0
```
asafshen: change to `fromversion`

dorsha: done

PR #1440, opened 2018-04-02
Preempt test and fixes

bakatzir:


Integrations/integration-Preempt.yml
```diff
@@ -456,22 +451,22 @@ script:
       required: true
       description: ID of the user
     outputs:
-    - contextPath: Endpoint.Hostname
+    - contextPath: Preempt.Endpoint.Hostname
```
anara123: Endpoint should stay Endpoint

PR #1441, opened 2018-04-02
tableToMarkdown - if the response was empty than in previous version it resulted an err…

barkatzir42:
…or, so I fixed it.


bakatzir:
@anara123  please notice that i've added the relevant tests to the playbook.

bakatzir:
Fixed in #1516

PR #1447, opened 2018-04-03
added missing characters to url regex

Bars92:
added some characters that can appear in regex to better find urls.


dorsha:
Need to add tests, I think we already have one for it.

Bars92:
Closing for better fix in client.

PR #1453, opened 2018-04-04
Cylance Protect v2 enhancements

Itay4:
Fixes https://github.com/demisto/etc/issues/9869
- Added fetch incidents
- Made context adjustments according to standard
- Fixed translation of Cylance score from API raw float score to UI score
- Fixed additional minor bugs


Itay4:
Fixes https://github.com/demisto/etc/issues/11359

PR #1457, opened 2018-04-05
Fix AD queries

yaakovi:
fixes: https://github.com/demisto/etc/issues/11030


Scripts/script-ADGetUser.yml
```diff
@@ -256,7 +264,6 @@ args:
   description: mail attribute of desire user
 - name: userAccountControlOut
   auto: PREDEFINED
-  description: Get translated user acount control flags
```
dorsha: Why did you remove the description?

Scripts/script-ADGetUser.yml
```diff
@@ -302,6 +309,8 @@ outputs:
   description: The indicator's score
   type: number
 scripttarget: 0
+runonce: false
 dependson:
   must:
   - ad-search
+releaseNotes: "add escaping to filter"
```
dorsha: Release notes should start with capital and should be explain better the fix (users will not understand this..)

Scripts/script-ADGetUser.yml
```diff
@@ -32,6 +32,13 @@ script: |-
   if demisto.get(demisto.args(), 'limit'):
       limit = demisto.args()['limit']
 
+
+  def escapeSpecialCharacters(s):
+      for c in r'\()':
```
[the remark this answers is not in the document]
dorsha: why not using regex group replace?

Scripts/script-ADGetUser.yml
yaakovi: there were two description lines, when downloading the integration it removed the first automatically.

dorsha: 👍

Scripts/script-ADGetUser.yml
```diff
@@ -32,6 +34,11 @@ script: |-
   if demisto.get(demisto.args(), 'limit'):
       limit = demisto.args()['limit']
 
+
+  def escapeSpecialCharacters(s):
```
asafshen: what happens if there are two agencent brackets?
image
  
it seems that only the second `(` is missing escape characters..isn't it?

anara123:
@asafshen @yaakovi what is status here?

PR #1465, opened 2018-04-08
add transformer JoinIfSingleElementOnly

meirwah:
@asafshen plz review, I need it for default mapping

### DIFF
--- a/Scripts/script-JoinIfSingleElementOnly.yml
+++ b/Scripts/script-JoinIfSingleElementOnly.yml
@@ -22,7 +22,7 @@ args:
 - name: value
   required: true
   default: true
-  description: The array to join is has single element
+  description: The array to join if it has single element
   isArray: true
 scripttarget: 0
 releaseNotes: 'add transformer JoinIfSingleElementOnly'
--- a/Scripts/script-JoinIfSingleElementOnly.yml
+++ b/Scripts/script-JoinIfSingleElementOnly.yml
@@ -16,7 +16,7 @@ tags:
 - transformer
 - general
 comment: Return the single element in case the array has only 1 element in it, otherwise
-  return the whole array.
+  return the whole array
 enabled: true
 args:
 - name: value
--- a/Scripts/script-JoinIfSingleElementOnly.yml
+++ b/Scripts/script-JoinIfSingleElementOnly.yml
@@ -1,0 +1,28 @@
+commonfields:
+  id: JoinIfSingleElementOnly
+  version: -1
+name: JoinIfSingleElementOnly
+script: |-
+  value = demisto.args()["value"]
+  res = value
+  if isinstance(value, list):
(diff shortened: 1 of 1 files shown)


Scripts/script-JoinIfSingleElementOnly.yml
```diff
@@ -0,0 +1,28 @@
...
+          res = value[0]
+
+  demisto.results(res)
+type: python
+tags:
+- entirelist
+- transformer
+- general
+comment: Return the single element in case the array has only 1 element in it, otherwise
+  return the whole array.
```
asafshen: lets remove the "." suffix :)

Scripts/script-JoinIfSingleElementOnly.yml
```diff
@@ -0,0 +1,28 @@
...
+enabled: true
+args:
+- name: value
+  required: true
+  default: true
+  description: The array to join is has single element
+  isArray: true
+scripttarget: 0
+releaseNotes: 'add transformer JoinIfSingleElementOnly'
```
asafshen: fyi no need for release notes in new files

Scripts/script-JoinIfSingleElementOnly.yml
```diff
@@ -0,0 +1,28 @@
...
+- transformer
+- general
+comment: Return the single element in case the array has only 1 element in it, otherwise
+  return the whole array.
+enabled: true
+args:
+- name: value
+  required: true
+  default: true
+  description: The array to join is has single element
```
asafshen: "is" => "if it" ?

Scripts/script-JoinIfSingleElementOnly.yml
```diff
@@ -0,0 +1,28 @@
+commonfields:
+  id: JoinIfSingleElementOnly
+  version: -1
+name: JoinIfSingleElementOnly
```
asafshen: "Join" has the connotation of joining list with a separator (python `",".join(somelist)` / js `somelist.join(",")`

maybe we should call it `ExtractSingleElementFromList` / `GetSingleElementFromList` .. wdyt?

meirwah:
@asafshen will explain the use case:
when I map fields from EWS, some fields can have values, that sometimes are arrays and other times arrays of size 1.
For instance EmailTo: 
can be [ email1 ]
or
[ email1 , email2 ]

when it's [ email1 ] I want the value to be `email1`
but when its array of size 2, I want it to be still an array (so other tasks will run it as array)

makes sense?

Scripts/script-JoinIfSingleElementOnly.yml
meirwah: neither, because when it's not single element, it stays an array.

Open to different name.. but don't have a good idea

asafshen:
@meirwah 👍

PR #1470, opened 2018-04-10
Phishing default mapping

meirwah:


Classifiers/classifier-EWS_v2.json
```diff
@@ -0,0 +1,297 @@
+{
+	"id": "c9332555-348b-4d91-85b4-e345cdba12c8",
+	"version": -1,
+	"modified": "2018-04-08T19:42:30.856069216Z",
+	"defaultIncidentType": "",
```
asafshen: Shouldn't be `"Phishing"`?

Classifiers/classifier-EWS_v2.json
```diff
@@ -0,0 +1,297 @@
...
+						]
+					}
+				}
+			},
+			"dontMapEventToLabels": false
+		}
+	},
+	"unclassifiedCases": {},
+	"custom": true,
+	"transformer": {
```
asafshen: if we know that all should map to "Phishing" we can give an empty `transformer`  (also try for `keyTypeMap` field)

IncidentFields/incidentfields-phishing.json
```diff
@@ -0,0 +1,812 @@
+[
```
asafshen: please change this file name to `incidentfields` (like reputation)

Classifiers/classifier-EWS_v2.json
meirwah: Where do you set this up in UI?  haven't seen that

asafshen: incident type in the instance settings:
image

and than just put an empty classification key

(a bad flow, we have an issue for that - https://github.com/demisto/etc/issues/11403)

meirwah: very

PR #1474, opened 2018-04-10
fixed 'IncidentSet' fromVerison from 3.6 to 3.5.1

guydemi:
title


Bars92:
There is no need.. fixed in https://github.com/demisto/content/pull/1431

PR #1477, opened 2018-04-10
Revert "Phishing default mapping"

meirwah:
Reverts demisto/content#1470


meirwah:
won't have time to fix it today..
Will try tomorrow ...

PR #1486, opened 2018-04-13
URL reputations with cleaning of escaped

slavikm:
@aviadl several things to notice:
1. I did not understand our previous URL regex as it allowed things like `www.(example.com)` so I changed it to something I think is valid with host prefix, port option and then whatever.
2. Added script to extract actual URL from ProofPoint URLs
3. Same script also tries to normalize if the URL does not have `http` prefix will add it
4. Added `contextPath` and `contextValue`
5. Added support for escaped URLs via `hxxp` and `[.]` and `&amp;`

Overall, I'm not sure about (3) and also the regex needs another eye. Will be happy to explain.


anara123:
@slavikm Does it solves also this PR https://github.com/demisto/etc/issues/11396?

aviadl:
@anara123 this regex will catch correctly the stuff you mentioned on the issue
But, it doesnt use the scripts you mention, so it is not relevant

PR #1491, opened 2018-04-16
Easy Vista Integration

Itay4:
Fixes https://github.com/demisto/etc/issues/11221


Itay4:
Documentation: https://github.com/demisto/etc/issues/11483

PR #1496, opened 2018-04-22
Support release notes for incident fields

asafshen:
encapsulate incident fields in an object in order to support release notes 

server PR: https://github.com/demisto/server/pull/10739

also changed a bit the order of release notes:
 - integration release notes to be before scripts release notes
 - moved reputation to the bottom


meirwah:
@asafshen build seems to fail

asafshen:
@meirwah I don't think it related to the PR

will check that though

PR #1501, opened 2018-04-24
SQS

bakatzir:
added sqs integration
fixes [etc/issues#11077](https://github.com/demisto/etc/issues/11077)


anara123:
@bakatzir what is the status here? fix the build and merge

PR #1503, opened 2018-04-25
Aquatone

meirwah:


Scripts/script-AquatoneDiscover.yml
```diff
@@ -0,0 +1,34 @@
+commonfields:
+  id: AquatoneDiscover
+  version: -1
+name: AquatoneDiscover
+script: |-
+  import subprocess
+  import time
+
+  domain = demisto.args()['domain']
+  res = subprocess.check_output(['aquatone-discover','--domain',domain])
```
anara123: If the process fails and the exit status is 1, then we get
image

I think we should better handle this and print the original output

PR #1504, opened 2018-04-25
fix playbook close incident if duplicate found

erezh31:
fixes:
https://github.com/demisto/etc/issues/11533


erezh31:
@meirwah plz merge (failed on EWS, not related to this PR)

PR #1510, opened 2018-04-29
Cybereason enhancements and fixes

Itay4:
- Improved the query sent in `query-connections` command.
- Added outputs to `is-probe-connected` command.
- Removed `login` command as it is unnecessary.


anara123:
fixes https://github.com/demisto/etc/issues/11311

anara123:
fixes https://github.com/demisto/etc/issues/11038
@Itay4 right?

PR #1511, opened 2018-04-29
New transformers

asafshen:
Added two transformers:
 - `TimeStampToDate` - Converts UNIX Epoch time to date that is support by demisto 
 - `WhereFieldEquals` - Filter list by certain field (need to done in order to filter nested levels, such as "Email.Headers"

And tests for both..

Also created another artifact - `content_test.zip` - that contains content with all `TestPlaybooks` artifacts 

will also show the second one in the demo next week, I think it can help the SEs...

### DIFF
--- a/Scripts/script-WhereFieldEquals.yml
+++ b/Scripts/script-WhereFieldEquals.yml
@@ -1,0 +1,55 @@
+commonfields:
+  id: WhereFieldEquals
+  version: -1
+name: WhereFieldEquals
+script: |+
+  var value = args.value;
+  var field = args.field;
+  var equalTo = args.equalTo;
+  var getField = args.getField;
+
+  if (!Array.isArray(value)) {
+      value = [value];
+  }
+
+  var res = value.filter(function(item) {
+     return item && item[field] == equalTo; // using '==' and not '===' to handle number equality better
+  });
+
+  if (getField) {
+      if (typeof getField !== 'string') {
+          logInfo('WhereFieldEquals - got invalid getField [' + getField + ']. Ignoring');
+      } else {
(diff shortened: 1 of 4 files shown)


Scripts/script-WhereFieldEquals.yml
```diff
@@ -0,0 +1,55 @@
...
+
+  var res = value.filter(function(item) {
+     return item && item[field] == equalTo; // using '==' and not '===' to handle number equality better
+  });
+
+  if (getField) {
+      if (typeof getField !== 'string') {
+          logInfo('WhereFieldEquals - got invalid getField [' + getField + ']. Ignoring');
+      } else {
+           res = res.map(function(item) { return item[getField]});
```
dorsha: make sure res is not nil.. if the above filter returned nothing

Scripts/script-WhereFieldEquals.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: value
+  required: true
+  default: true
+  description: List to apply the transformer on
+- name: field
+  required: true
+  description: Attribute in the collection's items check equality against 'equalTo'
+- name: equalTo
+  required: true
+  description: Value to filter all items in the collec
```
dorsha: change collec to collection

Scripts/script-WhereFieldEquals.yml
asafshen: I think `filter` cannot return null/undefined - it always returns an array (perhaps empty one)

dorsha:
@Bars92 you can use the `WhereFieldEquals` transformer (from the next content update).
Or upload it with a different name and id.

PR #1517, opened 2018-04-30
Added Ping results as output

dorsha:
I kept the (weird) `ping_result` default context key for backward compatibility.
Also changed the `verbose` argument to be predefined.

### DIFF
--- a/Scripts/script-Ping.yml
+++ b/Scripts/script-Ping.yml
@@ -16,10 +16,10 @@ script: |-
 
       verbose = demisto.get(demisto.args(),'verbose')
       if verbose and verbose.lower() == 'false':
-          demisto.results({'ContentsFormat': 'json', 'Type': entryTypes['note'], 'Contents': res, 'EntryContext': { "Ping.Result": res }})
+          demisto.results({'ContentsFormat': 'json', 'Type': entryTypes['note'], 'Contents': res, 'EntryContext': { 'Ping': res }})
       else:
           data = [ { 'ret_code' : res['ret_code'], 'destination' : res['destination'], 'max_rtt' : res['max_rtt'] , 'avg_rtt' : res['avg_rtt'],  'destination_ip' : res['destination_ip'] } ]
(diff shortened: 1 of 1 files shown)


Scripts/script-Ping.yml
```diff
@@ -16,10 +16,10 @@ script: |-
 
       verbose = demisto.get(demisto.args(),'verbose')
       if verbose and verbose.lower() == 'false':
-          demisto.results(res)
+          demisto.results({'ContentsFormat': 'json', 'Type': entryTypes['note'], 'Contents': res, 'EntryContext': { "Ping.Result": res }})
```
aviadl: Use single quote on Ping.Result
I think we have consts for entry format as well

dorsha: Didn't find any const

dorsha:
The result is not big, you can see it above. Just a table. I can break it down multiple outputs, it will be easier to use after

aviadl:
So lets break it, this way we can also declare on the integration level what is the output

dorsha:
Unfortunately we must keep the setContext for the entire result.. it is not big.. but messy. Unless we say that we don't want to support backward. Or deprecate this script and create new one.

aviadl:
@gshriki please approve to break compatibility

gshriki:
approved

aviadl:
@dorsha so do you want to leave it this way? or create the context explicitly, i tend to the second option (after Shriki has approved)

Scripts/script-Ping.yml
```diff
@@ -16,10 +16,10 @@ script: |-
 
       verbose = demisto.get(demisto.args(),'verbose')
       if verbose and verbose.lower() == 'false':
```
dorsha: @aviadl I also don't think we need to verbose argument.. this is why we have the raw-response... this script is a bit strange.. :)

gshriki:
Can we document breaking backward compatibility as part of the release notes, so customers can see they might have issues after upgrading?

dorsha:
Yes

dorsha:
@aviadl completely changed the script, please review again.

PR #1530, opened 2018-05-02
Fix epo-command

asafshen:
When you declaring command arguments - Demisto will ignore arguments that are not declared

I have extend the description of the command instead noting the command

we should also support this from Demisto server IMO


asafshen:
we will also create another release

asafshen:
Also added @aviadl's request for default parameters

PR #1532, opened 2018-05-02
release notes improvements

asafshen:
fixes: https://github.com/demisto/etc/issues/11593

also changed function names and variables from camel-case to snake-case (python convention)

tested heavily

output example: https://github.com/asafshen/my-company-content/releases/tag/18.5.0a


dorsha:
I don't like the italic, really annoying.

asafshen:
@dorsha I'll discuss this with Alex (he still doesn't have github access I think)

asafshen:
@yaakovi please review again

release_notes.py
```diff
@@ -372,142 +409,139 @@ def modifiedReleaseNotes(self, cnt):
...
 }
 
 
-def parseChangeList(filePath):
-    with open(filePath, 'r') as f:
+def parse_change_list(file_path):
+    with open(file_path, 'r') as f:
         data = f.read()
         return data.split("\n")
     return []
```
yaakovi: unreachable code: if success, the first return will execute. otherwise, an exception will be raised

release_notes_clear.py
```diff
@@ -107,7 +107,7 @@ def remove_releaseNotes_folder(folder_path, files_extension):
 
 
 def main(root_dir):
-    yml_folders_to_scan = ['Integrations', 'Playbooks', 'Scripts', ] # yml
+    yml_folders_to_scan = ['Integrations', 'Playbooks', 'Scripts', 'TestPlaybooks'] # yml
```
yaakovi: since when do we have releasenotes on test playbook?

release_notes.py
```diff
@@ -9,301 +9,337 @@
...
-    def loadData(self, data):
+    def load_data(self, data):
         return yaml.safe_load(data)
 
-    def getHeader(self):
+    def get_header(self):
         return "Scripts"
 
-    def addedReleaseNotes(self, cnt):
+    def added_release_notes(self, cnt):
```
yaakovi: why is this different from other implementation of the function in other classes (PlaybookContent, ReportContent...)

release_notes_clear.py
asafshen: we don't - but devs put it anyway

release_notes.py
asafshen: Do you mean script release notes?

taken from script comment (equivalent to description)

PR #1535, opened 2018-05-03
add fields and layout for malware

meirwah:


Layouts/layout-details-Malware.json
```diff
@@ -0,0 +1,459 @@
...
+						"version": 0,
+						"modified": "0001-01-01T00:00:00Z",
+						"fieldId": "incident_malwarefamily",
+						"isVisible": true
+					},
+					{
+						"id": "",
+						"version": 0,
+						"modified": "0001-01-01T00:00:00Z",
+						"fieldId": "incident_src",
```
asafshen: this field is not shown after content update - and we know why:
- in this PR - we added "Malware" to `incident_src` associated types 
- by design - we don't override `associated fields` (because the user

for now - we need to duplicate `incident_src` in order use it here (otherwise it will be accessible only for users which this field populated from content only now)

in the long-run - we need to solve this:
 - disable users from editing content fields properties - then we can override (imo best solution)
 - smart-merge (e.i. allow user only to add fields)

@aviadl any suggestions?

Layouts/layout-details-Malware.json
```diff
@@ -0,0 +1,459 @@
...
+				"name": "System Affected",
+				"type": "",
+				"isVisible": true,
+				"readOnly": false,
+				"fields": [
+					{
+						"id": "",
+						"version": 0,
+						"modified": "0001-01-01T00:00:00Z",
+						"fieldId": "incident_dest",
```
asafshen: same note as above

Layouts/layout-details-Malware.json
```diff
@@ -0,0 +1,459 @@
...
+						"isVisible": true
+					},
+					{
+						"id": "",
+						"version": 0,
+						"modified": "0001-01-01T00:00:00Z",
+						"fieldId": "incident_destntdomain",
```
asafshen: same note as above

Classifiers/classifier-Crowdstrike-Streaming.json
```diff
@@ -0,0 +1,56 @@
...
+				"File Path": {
+					"simple": "event.FilePath",
+					"complex": null
+				},
+				"User": {
+					"simple": "event.UserName",
+					"complex": null
+				},
+				"Vendor Product": {
+					"simple": "\"CrowdStrike\"",
```
asafshen: 👍

Classifiers/classifier-SplunkPy.json
```diff
@@ -28,6 +28,19 @@
 				}
 			},
 			"dontMapEventToLabels": false
+		},
+		"Malware": {
+			"internalMapping": {
```
asafshen: we can add "Vendor Product" -> "Splunk"  here as well

Classifiers/classifier-Crowdstrike-Streaming.json
```diff
@@ -0,0 +1,56 @@
...
+					"complex": null
+				}
+			},
+			"dontMapEventToLabels": false
+		}
+	},
+	"unclassifiedCases": {},
+	"custom": true,
+	"transformer": {
+		"simple": "metadata.eventType",
```
asafshen: if all events should map to "Malware" - we can (and perhaps should) remove the transformer. `defaultIncidentType` is sufficient (we need to improve the UX on that manner)

Classifiers/classifier-SplunkPy.json
meirwah: the vendor is not splunk.. because the vendor is the system the found the malware (like symantec endpoint) splunk is just the log mgmt

Layouts/layout-details-Malware.json
asafshen: absolutely agree, but today's behavior doesn't support this

meirwah: @aviadl is more for the smart-merge
@asafshen can you open issue to fix that?

asafshen: https://github.com/demisto/etc/issues/11804

PR #1538, opened 2018-05-03
Archer added command to get user's id by username

hirshbergshachar:


hirshbergshachar:
@anara123 should we close this? The new code is included in the Archer REST API PR

PR #1547, opened 2018-05-07
Phish.AI: Bugfix with api key usage

yevgenypats:


Integrations/integration-PhishAI.yml
```diff
@@ -7,15 +7,21 @@ category: Data Enrichment & Threat Intelligence
...
+  name: apiKey
   defaultvalue: ""
   type: 0
   required: false
+- display: Use system proxy settings
+  name: proxy
+  defaultvalue: "false"
+  type: 8
+  required: false
 script:
```
hirshbergshachar: Need to add an "insecure" param

Integrations/integration-PhishAI.yml
```diff
@@ -7,15 +7,21 @@ category: Data Enrichment & Threat Intelligence
...
   script: |2
 
     var TIMEOUT = 20 * 1000;
+    var DONT_CHECK_CERTIFICATE = false;
```
hirshbergshachar: No need if we have an "insecure" param

Integrations/integration-PhishAI.yml
```diff
@@ -41,7 +47,8 @@ script:
                 },
                 Body: JSON.stringify({'url': url})
             },
-            "false"
+            DONT_CHECK_CERTIFICATE,
```
hirshbergshachar: Change to params.insecure

Integrations/integration-PhishAI.yml
```diff
@@ -54,20 +61,22 @@ script:
...
                     Method: 'GET',
                     Headers: {
-                        'Accept': ['application/json']
+                        'Accept': ['application/json'],
+                        'Authorization': [apiKey]
                     }
                 },
-                "false"
+                DONT_CHECK_CERTIFICATE,
+                params.proxy
```
hirshbergshachar: Change to params.insecure

anara123:
@hirshbergshachar I think in this case when the Phish.AI service is SAAS, we can always assume that certificate could and should be validated. I think this can be ok to leave it constant

PR #1552, opened 2018-05-08
Playbook Malware Investigation - Generic: Add support for auto-extract:

JonathanMeler:
- Remove Extract Indicators sub-playbook

malware_investigation_-_generic-dev_tue_may_08_2018


anara123:
@JonathanMeler please make sure that there is no version problem, that we don't see 2 same playbooks

PR #1556, opened 2018-05-08
add indicators thresholds as instance parameters

noaco:
[etc/issues/10058](https://github.com/demisto/etc/issues/10058)

### DIFF
--- a/Integrations/integration-AnomaliThreatStream.yml
+++ b/Integrations/integration-AnomaliThreatStream.yml
@@ -32,6 +32,21 @@ configuration:
   defaultvalue: ""
   type: 8
   required: false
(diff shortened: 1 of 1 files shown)


Integrations/integration-AnomaliThreatStream.yml
```diff
@@ -32,6 +32,21 @@ configuration:
   defaultvalue: ""
   type: 8
   required: false
+- display: File Threshold
```
anara123: Add description of threshold range for all thresholds , and explain the numbers

noaco:
will be handheld on https://github.com/demisto/etc/issues/12043

PR #1564, opened 2018-05-10
FireEye HX fixes

noaco:


Integrations/integration-FireEyeHX.yml
```diff
@@ -905,7 +905,11 @@ script:
             raise ValueError('Failed to parse response body - unexpected response structure from the server.')
```
hirshbergshachar: The code below is ok. 

I have a problem with row 893. The dictionary comprehension there would also remove 0 values since the truth value of 0 is False. I don't know what are values that we can get there but if 0 is an option it is worth changing it.

noaco: It can only return a list. If no results are found, an empty list is returned.

PR #1565, opened 2018-05-10
FireEye HX fixes

noaco:


noaco:
@hirshbergshachar not the same command

PR #1567, opened 2018-05-13
Cylance Protect v2 DBotScore support

Itay4:
Related to https://github.com/demisto/etc/issues/10058
Updated documentation: https://github.com/demisto/etc/issues/11055


Integrations/integration-Cylance_Protect_v2.yml
```diff
@@ -173,6 +179,15 @@ script:
         ''' Translates API raw float (-1 to 1) score to UI score (-100 to 100) '''
         return score*100
 
+    def translate_score(score, threshold):
+        if score > 0:
```
noaco: Note that the threshold should determine if the indicator is malicious. 
If the score is higher or equal to the threshold then the DBot score should be 3 (malicious). 
Usually, if the score is in the range of threshold/2-threshold, the DBot score given is 2 (suspicious).
I don't know if this matches your case, but please consider this and match the logic.

Itay4: It is not the case here, as for malicious/suspicious threats - the threat score will be negative.

Integrations/integration-Cylance_Protect_v2.yml
```diff
@@ -733,14 +781,26 @@ script:
         page_size = demisto.args()['pageSize'] if 'pageSize' in demisto.args() else None
 
         lst = get_list_request(demisto.args()['listTypeId'], page, page_size)['page_items']
+        dbot_score_array = []
         for threat in lst:
+            dbot_score = 0
             score = threat.get('cylance_score', None)
             if score:
                 threat['cylance_score'] = normalize_score(threat['cylance_score'])
+                threshold = demisto.args().get('threshold')
```
anara123: you should do `demisto.args().get('threshold', FILE_THRESHOLD)`

Integrations/integration-Cylance_Protect_v2.yml
```diff
@@ -978,6 +1038,10 @@ script:
       required: true
       default: true
       description: The device ID
+    - name: threshold
+      description: If threat score is less or equal than the threshold, then file
```
anara123: explain here the threshold range or point to some docs

PR #1570, opened 2018-05-15
add fetch incidents limitation to RSA Netwitness

noaco:
fixes [etc/issue#11686#issuecomment-388550973](https://github.com/demisto/etc/issues/11686#issuecomment-388550973)

### DIFF
--- a/Integrations/integration-RSANetWitness_v11.1.yml
+++ b/Integrations/integration-RSANetWitness_v11.1.yml
@@ -575,6 +575,8 @@ script:
     def fetch_incidents():
 
         lastRun = demisto.getLastRun()
+        #limit each fetch to 100 incidents
+        limit = 100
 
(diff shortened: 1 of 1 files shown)


anara123:
@noaco can we limit the paging to 10?

noaco:
@anara123  It is possible to limit the page size to 10 and by that limit the data returned from each request. 
But, if we don't limit the incidents to be returned we will just pile up incidents before outputting them to the war room, and we will use a lot of requests to do so.
The timestamp for each fetch is set by the latest incident. This could have been a great solution to limit results per fetch without losing data if the incidents were returned in ascending order (by date)... but since the incidents are returned in descending order (and no option to change that) we need to find a different solution.

anara123:
@noaco lets limit the page_size to 10 and limit the total incidents to 100

PR #1584, opened 2018-05-18
Resolved issue with tc-tag-indicator

secdevopsai:
When running this command I was getting an error.  After revewing tc-get-indicators I noticed we weren't passing in the owners filter. This fixed the issue


anara123:
You have conflict with https://github.com/demisto/content/pull/1747/files. What is the difference between those two? If no difference then close the PR, otherwise solve the conflict.

PR #1589, opened 2018-05-22
Fix issue where splunk-notable-event-edit taking proxy settings when not needed

meirwah:


meirwah:
@anara123 can you plz review?

PR #1590, opened 2018-05-22
Better error when sha256/md5 are invalid

Spivaka:
fixes https://github.com/demisto/etc/issues/11777

### DIFF
--- a/Integrations/integration-Wildfire.yml
+++ b/Integrations/integration-Wildfire.yml
@@ -150,9 +150,9 @@ script:
         var hash = md5 || argHash || file;
         if(!hash)
             throw 'Missing md5/hash argument.';
-        if (hash.length === 40) {
-            // SHA1 not supported.
-            return 'No results';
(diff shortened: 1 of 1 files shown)


Integrations/integration-Wildfire.yml
```diff
@@ -150,9 +150,9 @@ script:
         var hash = md5 || argHash || file;
         if(!hash)
             throw 'Missing md5/hash argument.';
-        if (hash.length === 40) {
-            // SHA1 not supported.
-            return 'No results';
+        if ((argHash && !/^[A-Fa-f0-9]{64}$/.test(argHash)) ||
+           (md5 && !/^[A-Fa-f0-9]{32}$/.test(md5))) {
+            throw 'Invalid SHA256. Only SHA256 and MD5 are supported.';
```
anara123: better change the error to `Invalid hash. Only SHA256 and MD5 are supported`

PR #1602, opened 2018-05-23
add Round transformer

meirwah:


dorsha:
For the specific use we can send this script

meirwah:
don't think we should be afraid of performance

dorsha:
Still, best practice is to use native go and not start vm or docker for each transformer/filter. 
We might have 10+ transformers and filter within a single tasks, and for the entire playbook we might come up with 1000+ transformers and filters for a single run. If we run 10-30 playbooks in parallel it will cause issues for sure. 
I added the ability to create transformers and filters within script for the reason that I wanted a quick solution for customers, for the long term the solution is to use native go.

Scripts/script-Round.yml
```diff
@@ -0,0 +1,17 @@
+commonfields:
+  id: Round
+  version: -1
+name: Round
+script: demisto.results(int(round(float(demisto.args()["value"]))))
+type: python
+tags:
+- transformer
+- number
```
anara123: what number tag means?

Scripts/script-Round.yml
```diff
@@ -0,0 +1,17 @@
+commonfields:
+  id: Round
+  version: -1
+name: Round
+script: demisto.results(int(round(float(demisto.args()["value"]))))
```
anara123: you better set the result for some output, instead of just return the result to warroom.
And add some description whether the round will be always rounded down 3.2 => 3 and 3.9 => 3?

anara123:
@dorsha I think it is better to add transformers in content because content release is every 2 weeks, and each server release we can move those transformers to be implemented in go and remove them from content

Scripts/script-Round.yml
dorsha: @anara123 it is a transformer, there is no outputs and it doesn't go to the war room
@asafshen already added floor and ceil as well

Scripts/script-Round.yml
dorsha: Numeric transformer

dorsha:
Yes, @asafshen already pushed the transformer to the server as well.

Scripts/script-Round.yml
anara123: then why we need round if we have ceil and floor?
And those transformers already available at 3.6?

dorsha: Round is very important, their are cases that you want to round to the closet number, ceil and floor will not help

anara123:
@dorsha so we merge or close the issue?

dorsha:
Lets close for now.

PR #1605, opened 2018-05-24
A script to hide fields on `new incident`

aviadl:
fixes https://github.com/demisto/etc/issues/11765


NeelRust:
Nice one..Really like the comments.

PR #1607, opened 2018-05-24
QRadar offense per call is undefined when instance already exist

Spivaka:
fixes https://github.com/demisto/etc/issues/11781

### DIFF
--- a/Integrations/integration-QRadar.yml
+++ b/Integrations/integration-QRadar.yml
@@ -534,3 +534,4 @@ script:
     description: Retrieve a note for an offense
   isfetch: true
   runonce: false
+releaseNotes: "Fixed upgrade issue for fetch incidents"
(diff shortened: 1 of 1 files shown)


Integrations/integration-QRadar.yml
```diff
@@ -64,7 +64,7 @@ script:
     var server = params.server+'/';
     var insecure = params.insecure;
     var query = params.query;
-    var offensesPerCall = parseInt(params.offensesPerCall);
+    var offensesPerCall = parseInt(params.offensesPerCall || 50);
```
omercnet: approved, but why is this needed? there's a default value for this parameter

dorsha: For already configured instances. Must always think about upgrade.

omercnet: right, that's why the default is there
if there's no configured value, the default should be used - why is it not being used after an upgrade?

dorsha: Because that's the way it works at the moment

dorsha: https://demisto.slack.com/archives/C1X1QDL6B/p1527163184000039

omercnet: so let's fix it :)

PR #1617, opened 2018-05-29
Added handling of ItemAttachment EWSObjects

BenJoParadise:
Changed ews-get-attachment: when calling on an ItemAttachment, will save it to the warroom as an EML file, along with all of it's attachments.


BenJoParadise:
Adds feature: https://github.com/demisto/etc/issues/11654

PR #1619, opened 2018-05-29
Fixed ascii encoding problem by encoding all output strings to utf-8

BenJoParadise:
fixes https://github.com/demisto/etc/issues/11565


anara123:
can't reproduce in our lab

PR #1621, opened 2018-05-30
Fix SEPM14 insecure label

omercnet:
fixes https://github.com/demisto/etc/issues/11857


Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -643,3 +643,4 @@ script:
       description: The last-updated date.
     description: Retrieves the client content.
 hidden: false
+releaseNotes: Fix insecure label
```
dorsha: Put "-" instead

dorsha:
@omercnet why did you merge? Please fix the comment

PR #1622, opened 2018-05-30
remove releasenotes for SEPM14

omercnet:


Integrations/integration-SymantecEndpointProtection.yml
```diff
@@ -643,4 +643,4 @@ script:
       description: The last-updated date.
     description: Retrieves the client content.
 hidden: false
-releaseNotes: Fix insecure label
+releaseNotes: -
```
dorsha: You might need to change to "-" if build fails

PR #1624, opened 2018-05-30
FindSimilarIncidents fix

erezh31:


Scripts/script-FindSimilarIncidents.yml
```diff
@@ -96,50 +95,34 @@ script: |-
...
 
   def get_incidents_by_keys(similar_incident_keys, incident_time, incident_id, hours_back, ignore_closed, max_number_of_results):
-      similar_keys_query = " or ".join(map(lambda t: '%s:%s' % (t[0], t[1]), similar_incident_keys.items()))
+      similar_keys_query = " and ".join(map(lambda t: '%s:"%s"' % (t[0], t[1]), similar_incident_keys.items()))
 
       incident_time = parse_datetime(incident_time)
       max_date = incident_time
       min_date = incident_time - timedelta(hours=hours_back)
-      hours_back_query = 'occurred:>="%s" and occurred:<="%s"' % (min_date.isoformat(), max_date.isoformat())
+      hours_back_query = '{0}:>="{1}" and {0}:<="{2}"'.format(TIME_FIELD, min_date.isoformat(), max_date.isoformat())
```
meirwah: so now you can support both   occurred and created time?

Scripts/script-FindSimilarIncidents.yml
```diff
@@ -373,3 +329,4 @@ outputs:
 scripttarget: 0
 timeout: 300ns
 runonce: false
+releaseNotes: "Fix bugs"
```
meirwah: maybe elaborate here a bit

Scripts/script-FindSimilarIncidents.yml
erezh31: yes

Scripts/script-FindSimilarIncidents.yml
```diff
@@ -167,42 +150,50 @@ script: |-
...
+              'time': time}
 
 
   def is_text_equal_by_x_different_words(text1, text2, number_of_different_words, separator=' '):
       if not isinstance(text1, basestring):
           text1 = str(text1)
       if not isinstance(text2, basestring):
           text2 = str(text2)
-      if number_of_different_words < 0:
+      text1 = text1.lower()
```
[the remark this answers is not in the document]
erezh31: it's different logic.. TFIDF is also statistics based, here is for only rule based. it's possible to use both in a playbook if some wants to

PR #1626, opened 2018-05-30
Darktrace - new integration

erezh31:
@gshriki asked @dansarel2, the API is documented in the product so we can publish it.
Also we use the same function call as in this example:
https://github.com/hutchris/darktrace


gshriki:
@dansarel2 - Last time we talked we were not sure if ok to publish as it is based on private APIs.
However, I think we can publish, as the APIs are also published in the python client we based it on.
wdyt?

erezh31:
already talked with @dansarel2

Integrations/integration-Darktrace.yml
```diff
@@ -0,0 +1,163 @@
...
+
+    def add_label(labels, label_key, value):
+        if value:
+            labels.append({'type': label_key, 'value': str(value)})
+
+    def parse_event_to_incident(event):
+        incident = {
+            'name': 'Darktrace - ' + event['model']['now']['name'],
+            'details': json.dumps(event),
+            'occurred': from_unixtime_mili(event['time']).isoformat() + '.000Z'
```
anara123: rawJSON is missing

erezh31:
@anara123 added the rawJSON field.
We don't have instance - so can't write test playbook and anyway this integration has only fetch incidents command...

PR #1628, opened 2018-05-30
Rasterize test btn

aviadl:


aviadl:
fixes demisto/etc#11615

PR #1638, opened 2018-06-04
Fixed "error_message undefined" issue. 

BenJoParadise:
The raise should only occur in the "except" block.
This resolves the issue experienced by slack user (customer) @mohan1704

### DIFF
--- a/Integrations/integration-EWSMailSender.yml
+++ b/Integrations/integration-EWSMailSender.yml
@@ -251,7 +251,7 @@ script:
             else:
                 demisto.info(error_message + "\nFull debug log:\n" + debug_log)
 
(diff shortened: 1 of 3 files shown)


anara123:
ok. But still add test playbook.
Do you see the entry email sent confirmation?

BenJoParadise:
Pending merge of https://github.com/demisto/content-test-conf/pull/93

BenJoParadise:
@anara123 I see it in my setup, i'm thinking i'll send him a debug integration to see why the entry is not returning.

BenJoParadise:
@anara123 He sees the message now after he restarted the Demisto service.
I've added the test, please review.

PR #1639, opened 2018-06-04
fixes in McAfee ATD, support in detonate commands and test playbook

bakatzir:
fixes etc/issues#8986
fixes etc/issues#8991
fixes etc/issues#9165
fixes etc/issues#9789


JonathanMeler:
@bakatzir please commit the playbook (for ATD), I'll add it to the master playbook)

bakatzir:
There is an issue regarding the integration that I wasn't able to solve yet.
when giving the detonate-file command an input if more than 1 file via a test playbook,
the dq command (in the fileUploadRaw function in the code) doesn't return the filename
although it does work if I run the command from the commandline in demisto.

bakatzir:
our license is out of date, @shai-kappel  is handling it.

PR #1646, opened 2018-06-05
print both email body parts

gshriki:
Print both email body parts - Text and HTML in the war room entry response.
Fixes https://github.com/demisto/etc/issues/11934

do not merge until final confirmation


gshriki:
Most emails are heavily formatted and styled, so I think it will be a very bad looking entry. We can debate on rasterize, but I think that's up to the playbook...

anara123:
sorry didn't noticed the first message to not merge

PR #1647, opened 2018-06-05
Make strings parse properly text file

aviadl:
fixes https://github.com/demisto/etc/issues/11931
test playbook to follow


slavikm:
Who needs to extract strings from text files? They are text files, no? You can just review content?

aviadl:
You called this `strings`, this is how `strings` works ...

PR #1653, opened 2018-06-07
don't override labels in CS streaming mapping

meirwah:


Classifiers/classifier-Crowdstrike-Streaming.json
```diff
@@ -43,7 +43,7 @@
 					"complex": null
 				}
 			},
-			"dontMapEventToLabels": false
+			"dontMapEventToLabels": true
```
asafshen: @meirwah why?

meirwah: because CS streaming have a set of labels the integration creates (parsed).
If I make this false, those labels will be overriden by junk labels.

asafshen: got it,
we need to change this behavior so existing labels won't be overridden (I'll open an issue)

meirwah: yes

PR #1658, opened 2018-06-08
JSTableToMarkdown array values

liorblob:
master currently fails


TestPlaybooks/script-VerifyTableToMarkDown.yml
```diff
@@ -32,6 +36,16 @@ script: |-
       throw "Simple Data failed to be converted to RIGHT Human Redable fotmat. Got: " + h;
   }
 
+  h = tableToMarkdown("Simple Array", JSON.parse(simpleArray), 'Array');
```
anara123: why you define simpleArray as a string and then `JSON.parse` it? Why not define it as array from the beginning

TestPlaybooks/script-VerifyTreeToFlatObject.yml
```diff
@@ -0,0 +1,31 @@
+commonfields:
+  id: c5cb179f-d6d2-4d87-8857-b224689d5b00
+  version: 15
+name: VerifyTreeToFlatObject
+script: |+
+  var dataWithNestedObject = '{'+
```
anara123: same here with, why not just define the object

TestPlaybooks/script-VerifyTableToMarkDown.yml
liorblob: I was being consistent with the `VerifyTableToMarkdown` script, I assume there was a reason for it

anara123: I don't see a valid reason reading the script. Please change this in all the places in the script to be object

PR #1659, opened 2018-06-09
tableToMarkdown support array of strings

anara123:


dorsha:
Do the same also for the JS implementation, they should be aligned, otherwise users will not trust the content

anara123:
@dorsha sure will do. @liorblob was working on that change, that is why I assigned him to review

dorsha:
👍

liorblob:
@anara123 @dorsha https://github.com/demisto/content/pull/1658 should solve the issue in JS

anara123:
@liorblob sure. But you can review this PR and approve in the mean time

anara123:
@liorblob that is how I check if the header is string, if yes then it single string, otherwise it should be list

anara123:
closing for now. DON'T DELETE THE BRANCH

PR #1662, opened 2018-06-10
Extract domain formatting

meirwah:
fixes https://github.com/demisto/etc/issues/11939


meirwah:
@aviadl can you review again?

Scripts/script-ExtractDomainFromUrlFormat.yml
```diff
@@ -0,0 +1,32 @@
...
+name: ExtractDomainFromUrlAndEmail
+script: |-
+  from tld import get_tld
+  from validate_email import validate_email
+  import re
+
+
+  inp = demisto.args()['input']
+
+  is_email = validate_email(inp)
```
aviadl: You are running this script only on urls

meirwah: both urls and emails

PR #1664, opened 2018-06-11
Crowdstrike test playbook made generic

hirshbergshachar:
Playbook is now looking for devices instead of checking a specific device.


TestPlaybooks/playbook-CrowdStrike_Endpoint_Enrichment_-_Test.yml
```diff
@@ -1,24 +1,21 @@
 id: crowdstrike_endpoint_enrichment_-_test
 version: -1
 name: CrowdStrike Endpoint Enrichment - Test
-description: ""
-fromversion: "3.5.0"
```
anara123: Do you removed fromversion on purpose? If not then return it

PR #1666, opened 2018-06-11
Fix set date field

meirwah:


meirwah:
fix set date field by meirwah pull request 1666 demisto content 2018-06-11 20-09-03

PR #1668, opened 2018-06-11
scripts I have created during POCs 

vik-patel:
adding a few numeric transformers which I created for POCs


dorsha:
All of them already part of the product.

vik-patel:
does not look like I did this correctly, do I need a new pull request?

dorsha:
You will have them in the next product release. Meanwhile use the scripts you created

PR #1672, opened 2018-06-13
Fix categorization false positive

omercnet:
fixes https://github.com/demisto/etc/issues/12026

### DIFF
--- a/Integrations/integration-Cisco-umbrella.yml
+++ b/Integrations/integration-Cisco-umbrella.yml
@@ -194,8 +194,8 @@ script:
             }
         }
 
-        if (categorization.status === 1) {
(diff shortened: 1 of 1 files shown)


dorsha:
Maybe we should also handle 1 results?
Also, add release notes

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -1094,3 +1094,4 @@ script:
       database. There are several excellent tools online such as http://regexr.com
       to help if you’re not familiar with building RegEx.
 hidden: false
+releaseNotes: "Fixed categorization false positive."
```
omercnet: @dorsha ^

omercnet:
I thought I did add release notes

We handle the 1 result by returning the categories in the response, is there a good place in the context to add them that's not Malicious?

PR #1673, opened 2018-06-13
ipinfo.io - Added support to use API token for paid plans

omercnet:
fixes https://github.com/demisto/etc/issues/11915


Itay4:
Did you test it? I assume we don't have an instance.

omercnet:
yes this is tested
API keys are also available free without a plan

Integrations/integration-Ipinfo.yml
```diff
@@ -32,7 +37,7 @@ script:
     }
 
     var base = 'http://ipinfo.io/';
-    var jsonSuffix = '/json';
+    var jsonSuffix = '/json?token=' + params.token;
```
Itay4: Should it not be added to `ipinfo_field` command endpoint URL as well?

omercnet: yes, fixed

Integrations/integration-Ipinfo.yml
```diff
@@ -50,7 +55,7 @@ script:
             }
             return reply;
         case 'ipinfo_field':
-            return sendRequest(base+args.ip+'/'+args.field, false);
+            return sendRequest(base+args.ip+'/'+args.field+'?token='+args.token, false);
```
Itay4: There is no `token` argument, should be `params.token` .
Please create a global `token` variable and use it in both URL addresses.

Integrations/integration-Ipinfo.yml
```diff
@@ -32,7 +37,8 @@ script:
     }
 
     var base = 'http://ipinfo.io/';
-    var jsonSuffix = '/json';
+    var token = params.token;    
+    var jsonSuffix = '/json?token=' + token;
```
Spivaka: you will have undefined there when no token

Spivaka: you should use ?ternary condition here

Integrations/integration-Ipinfo.yml
```diff
@@ -37,8 +37,8 @@ script:
     }
 
     var base = 'http://ipinfo.io/';
-    var token = params.token;    
-    var jsonSuffix = '/json?token=' + token;
+    var token = params.token ? "?token=" + params.token : "";
```
Spivaka: we use ' and not "

PR #1676, opened 2018-06-13
Scripts to convert warroom JSON to CSV

vik-patel:


anara123:
@vik-patel `FILEToCSV` seems like `ParseCSV`
and `JSONToCSV` seems like `ExportToCSV`

vik-patel:
@anara123 ... these are different ... 

FILEtoCSV takes a json warroom output file (when output is  too large) and converts to CSV file.

I could not get ExportToCSV to work for me because JSON was too large to push into context ... thus I created JSONtoCSV to take an entryid can convert to a CSV file

NeelRust:
name for the former shd be `JSONFiletoCSV`

NeelRust:
@anara123 @vik-patel can you change the name of the file please? the name is really confusing otherwise.

vik-patel:
will do ... change file name, add test playbook and pull for master

Itay4:
@vik-patel What is the status of this?

vik-patel:
@anara123 how do I create test playbooks for this? ... need to have a JSON entry and a JSON file entry

Itay4:
@vik-patel You can upload JSON file to https://github.com/demisto/content/tree/master/TestData and then use HTTP script to download it to the war-room and use its entry ID.

yardensade:
So I have tried to test the script with an example valid json file:
{ "dns-admin@google.com": { "totalResults": 2, "offset": 0, "moreDataAvailable": true, "limit": 2, "sortField": "domain name [default]", "domains": [ { "domain": "googlerightsflow.net", "current": true }, { "domain": "mycloudaudit.net", "current": true } ] }, "admin@google.com": { "totalResults": 2, "offset": 0, "moreDataAvailable": true, "limit": 2, "sortField": "domain name [default]", "domains": [ { "domain": "arabaoyunu.us", "current": false }, { "domain": "aterisalonu.com", "current": false } ] } }

And I got this error:

AttributeError: 'unicode' object has no attribute 'items'

(since d in dictlist was a string basically)

When I've provided a valid json file which is demisto.result from my own integration, it gave me this error:

ValueError: Expecting property name: line 1 column 3 (char 2)

I'm not currently sure if this script is supposed to work in a specific use-case or not, if it does I think the script details should reflect that and also provide some sort of example here that works.

The second script provides me with the same results essentially,
if the JSON itself is a list and not dict I get another type of error:

> AttributeError: 'list' object has no attribute 'items'

Also it seems like there are some indentation issues in the code here which I had to correct for the script to work, should take a look into it.

vik-patel:
@yardensade ... yes, these scripts were created to handle single-level dictlist json ... I will see if I can add support for more complex/multilevel json ... if not, I will update description/examples to reflect what it can handle

anara123:
@vik-patel I think single level dict should be ok in this case, because we convert JSON to CSV. So just add proper documentation

Scripts/script-JSONFileToCSV.yml
```diff
@@ -0,0 +1,83 @@
...
+- contextPath: File.SHA1
+  description: SHA1 hash of the file (only in case of report type=json)
+- contextPath: File.SHA256
+  description: SHA256 hash of the file (only in case of report type=json)
+scripttarget: 0
+timeout: 600ns
+runonce: false
+releaseNotes: "-"
+tests:
+- Forgive me for my sins but I did not create any test
```
anara123: @jochman why  you couldn't add test?

jochman: I'm working on it, will add it asap

PR #1681, opened 2018-06-14
Handling edge case of None type in getUtfString

hirshbergshachar:


hirshbergshachar:
@anara123 Added test playbook

PR #1686, opened 2018-06-16
handling usernames with backslash

gshriki:
Handling usernames with backslash


gshriki:
Fixes https://github.com/demisto/etc/issues/12068

PR #1694, opened 2018-06-18
change proxy

liorblob:


dorsha:
Make sure it works after upgrade, in case the instance already saved. 
You can do it in the integration level by passing false/true instead of the string.

liorblob:
@dorsha I think it works for both `"true"` and `true`, but I added a conversion just in case

PR #1696, opened 2018-06-18
Fixes issue with EWS Search and Delete

JonathanMeler:


Playbooks/playbook-Search_And_Delete_Emails_-_EWS.yml
```diff
@@ -2,6 +2,7 @@ id: search_and_delete_emails_-_ews
 version: -1
 name: Search And Delete Emails - EWS
 fromversion: 3.6.0
+releaseNotes: "Add mailbox id to delete function"
```
aviadl: Note should be:
Added targe-mail-box to delete task

Scripts/script-BuildEWSQuery.yml
```diff
@@ -2,7 +2,13 @@ commonfields:
   id: BuildEWSQuery
   version: -1
 name: BuildEWSQuery
+releaseNotes: "Add paramter for stripping the subject from prefixes"
```
aviadl: typo on parameter

Scripts/script-BuildEWSQuery.yml
```diff
@@ -2,7 +2,13 @@ commonfields:
...
 script: |-
+  import re
```
aviadl: Since you have a script for this
You can easily right a test playbook for this specific script

Scripts/script-BuildEWSQuery.yml
```diff
@@ -2,7 +2,13 @@ commonfields:
...
+
+  # Regex for removing forward/replay prefixes
+  p = re.compile('([\[\(] *)?(RE|FWD?) *([-:;)\]][ :;\])-]*|$)|\]+ *$', re.IGNORECASE)
```
aviadl: This regex seems to be removing data from middle of line as well, but i might be wrong here

PR #1698, opened 2018-06-18
Fix home dashboard to handle properly users that contain `\` in their…

aviadl:
… name

related to https://github.com/demisto/etc/issues/12069

### DIFF
--- a/Dashboards/dashboard-Home-old.json
+++ b/Dashboards/dashboard-Home-old.json
@@ -1,8 +1,8 @@
 {
 	"id": "my-home-dashboard",
 	"version": -1,
-	"fromVersion": "4.0.0",
-	"releaseNotes": "Support for users with \\",
+	"fromVersion": "3.5.1",
+	"toVersion": "4.0.0",
 	"period": {
 		"byTo": "",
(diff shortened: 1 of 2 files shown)


Dashboards/dashboard-Home-old.json
```diff
@@ -0,0 +1,275 @@
+{
+	"id": "my-home-dashboard",
```
dorsha: mistake?

Widgets/widget-MyTasks-old.json
```diff
@@ -0,0 +1,29 @@
+{
+  "id": "my-tasks",
```
dorsha: mistake #2?

Dashboards/dashboard-Home-old.json
aviadl: No, should have the same ID as the other one, since they are targeted for different versions

Widgets/widget-MyTasks-old.json
aviadl: No, should have the same ID as the other one, since they are targeted for different versions

PR #1702, opened 2018-06-19
Fixed error when parsing incident timestamp during fetch_incident

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/12020

Converted FETCH_SEVERITIES from global constant to parameter


Integrations/integration-SymantecMSS.yml
```diff
@@ -102,10 +107,24 @@ script:
         return proxy
 
 
+    def load_severities():
+        POSSIBLE_SEVERITIES = ["Emergency", "Critical", "Warning", "Informational"]
+        severities_list = None
+        try:
+            severities_list = demisto.params()["severities"].replace(" ", "").split(",")
+        except:
+            raise Exception("Error parsing severities parameter.")
```
anara123: in this case we better return an error entry rather than raise an Exception.
```
def return_error_and_exit(message, err):
        if err is not None:
            demisto.error(err)

        demisto.results({
            'Type': entryTypes['error'],
            'Contents': message
        })
        sys.exit(0)

# usage:
return_error_and_exit("Error parsing severities parameter.")
```

PR #1708, opened 2018-06-20
scripts - deprecate checkwhitelist + add filterbywhite lists

shaniacht1:
fixes https://github.com/demisto/etc/issues/12389


Scripts/script-CheckWhitelist.yml
```diff
@@ -41,3 +41,5 @@ outputs:
 scripttarget: 0
 dependson: {}
 timeout: 0s
+deprecated: true
+releaseNotes: "deprecated CheckWhitelist script"
```
dorsha: Should start with capital letter and you should mention the new script to use instead

Scripts/script-FilterByWhitelist.yml
```diff
@@ -0,0 +1,57 @@
...
+  default: true
+  description: The value to look for on the whitelist
+- name: whitelistname
+  description: Name of whitelist to check
+  defaultValue: Indicators Whitelist
+outputs:
+- contextPath: Whitelist.In
+  description: List of items in the whitelist
+- contextPath: Whitelist.NotIn
+  description: List of items not in the whitelist
```
dorsha: I would check the wording with @kirbles19  - for all descriptions

kirbles19: - Line 35 should be `HumanReadableFormat` not `HomanReadableFormat`.
- Line 42: Check whether the specified item is whitelisted. Default list is the Demisto Indicators Whitelist.
- Line 47: Value might be the argument name, but what is it a value of? an IP? or are there multiple valid values?
- Line 53: List of items on the whitelist (Try to stay consistent with in/on. We used on in Line 47, so better to keep it the same throughout)
Line 55: List of items not on the whitelist

shaniacht1: @kirbles19 regarding line 47 - it could be anything.
The main usage of this command is to search the whitelist (so value=indicator), but it could also be used to search any list and then the value could be anything

anara123:
It seems that the reason to replace checkwhitelist with filterbywhitelist is because array input. So I think performance  wise it is better to implement filterbywhitelist to support array `value` input.

shaniacht1:
@anara123 it's not that. it's because the outputs of the previous one are not useable in a playbook.

shaniacht1:
@anara123 changed

PR #1712, opened 2018-06-21
Fix git hash

asafshen:
from last content update - release notes was not created properly:
https://10140-60525392-gh.circle-artifacts.com/0/artifacts/release-notes.txt

they seems to be empty because git diff fails:
image

it seems that there is a glitch in the specific git hash we compare against
I changed the git hash, added missing release notes (we didn't catch them)


asafshen:
@ana
@yaakovi

please approve and merge asap

PR #1717, opened 2018-06-24
Splunkpy search

meirwah:
fixes https://github.com/demisto/etc/issues/11917

### DIFF
--- a/Scripts/script-SplunkPySearch.yml
+++ b/Scripts/script-SplunkPySearch.yml
@@ -1,0 +1,34 @@
+commonfields:
+  id: SplunkPySearch
+  version: -1
+name: SplunkPySearch
+script: |-
+  query = demisto.args()['query']
+  rows = demisto.args()['rows']
+
+  query = query + ' | head ' + rows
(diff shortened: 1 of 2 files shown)


Scripts/script-SplunkPySearch.yml
```diff
@@ -0,0 +1,34 @@
...
+name: SplunkPySearch
+script: |-
+  query = demisto.args()['query']
+  rows = demisto.args()['rows']
+
+  query = query + ' | head ' + rows
+  res = demisto.executeCommand('splunk-search',{'using-brand': 'splunkpy', 'query': query})
+
+  if (res and len(res)>0 and res[0]['Contents']):
+      demisto.results({"Type": 1, "Contents": res[0]['Contents'], "ContentsFormat": "json", "EntryContext": {},
```
dorsha: Is it only for display? no need outputs?

meirwah: I saw that other enhancements scripts don't update context.
They just show outputs

dorsha: If it used only for enhancements it is ok

PR #1718, opened 2018-06-24
add context canvas connections

erezh31:


asafshen:
@erezh31 validation looks good 👍

PR #1723, opened 2018-06-25
Condition tag added to ContextContains

saharsabban:
Condition tag added to ContextContains automation for the ability to use it as conditional task in playbooks.


aviadl:
Why do we even need this script now ?
@dorsha , @gshriki

dorsha:
Don't think so, can be marked as deprecated I think

saharsabban:
Customer confirmed Exists script works as desired and closed ticket.

PR #1728, opened 2018-06-26
Ews untitled/empty attachments

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/12048

@anara123 most of the fixes are only reachable from `fetch-incidents`, but there were a few fixes from `ews-get-attachments`, should i create a test playbook for those?


anara123:
@BenJoParadise yes

BenJoParadise:
Can't create a test email because outlook blocks adding empty files as email attachments.

aviadl:
Use another mail client, such as the go client that we have, res sender

BenJoParadise:
I managed to send an email with an empty attachment, and discover a problem.

Every time we invoke `.content` on an attachment (e.g. to check that it isn't empty), exchangelib tries to `base64decode` the data, but since the data is empty, it throws an exception.

I've opened an issue in their git: https://github.com/ecederstrand/exchangelib/issues/436#issuecomment-401165596

Until this is resolved, the only workaround (without actually changing the exchangelib code) is to wrap every one of those with try/except.

PR #1736, opened 2018-06-27
Remove special support for EWS

JonathanMeler:
Remove special support for EWS
Remove special playbook for custom fields (Use "setIncident" instead)

fixes: https://github.com/demisto/etc/issues/12112

process_email_-_generic_dev2_wed_jun_27_2018


anara123:
@JonathanMeler add test

PR #1737, opened 2018-06-27
Threat Grid File Detonation + URL Detonation

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/9266
Improved the human readable output of threat-grid-get-threat-summary-by-id
Added DBot Score to threat-grid-get-threat-summary-by-id


BenJoParadise:
Added @JonathanMeler as approver before merging.

gshriki:
@BenJoParadise - can we add the threat-grid playbook selection to the detonate command? this came up from a customer recently.

gshriki:
From Citco:
> The Threat Grid integration (threat-grid-upload-sample) does not have an option to specify the (Threat Grid) playbook to be used. 
Without specifying the playbook, it uses the default "random mouse movement'. This is fairly useless as it won't trigger anything. For example, a document I've sent through Demisto testing Threat Grid's sandboxing many, many times has always come back clean. I just went into Threat Grid interactively and clicked on a link in that document and it came back with a score of 90 after launching dropper.

>The threat-grid-upload-sample command also includes options "os" and "osver" in addition to "vm", which is confusing. 
>According to the Threat Grid API documentation, you use "vm" to specify which image to use (which in turn selects the OS). there is no "os" or "osver" parameter in their submission API.

PR #1739, opened 2018-06-28
Now supports cloud Metadefender

dem-evgeny:


Integrations/integration-opswat-metadefender.yml
```diff
@@ -39,42 +39,82 @@ configuration:
   defaultvalue: "66"
   type: 0
   required: true
+- display: It is cloud
```
Itay4: I would change it to something which looks a bit more professional, like `Cloud based`

Integrations/integration-opswat-metadefender.yml
```diff
@@ -19,7 +19,7 @@ configuration:
   defaultvalue: ""
   type: 4
   required: false
-- display: Trust any certificate (unsecure)
```
Itay4: Why did you change it? Should be as it was

Integrations/integration-opswat-metadefender.yml
```diff
@@ -39,42 +39,82 @@ configuration:
   defaultvalue: "66"
   type: 0
   required: true
+- display: It is cloud
+  name: cloud
+  defaultvalue: "false"
```
Itay4: Instances which are already defined won't get the default value, you should handle it in the code.

Integrations/integration-opswat-metadefender.yml
```diff
@@ -39,42 +39,82 @@ configuration:
...
+  type: 8
+  required: false
 script:
   script: |
     var sendRequest = function(url,param, fileId) {
         var requestUrl = params.url + url + param;
```
Itay4: Please handle the case that the URL given in the parameters is followed by `/`

Integrations/integration-opswat-metadefender.yml
```diff
@@ -39,42 +39,82 @@ configuration:
...
+                            Accept: ['application/json']
+                        }
                     }
-                }
-                );
+                    );
+            }
         }
 
+        return res;
```
Itay4: Remove this

Integrations/integration-opswat-metadefender.yml
```diff
@@ -39,42 +39,82 @@ configuration:
...
+                    requestUrl,
+                    {
+                        Method: 'GET',
+                        Headers: {
+                            Accept: ['application/json']
+                        }
                     }
-                }
-                );
+                    );
```
Itay4: Please lint properly

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Changed

dem-evgeny:
@Itay4 "Did you test it?"
Strange question. Yes, I did. Not only I but SE as well.

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Because it should be: "Do not validate server certificate (insecure)"

Integrations/integration-opswat-metadefender.yml
dem-evgeny: I already handled it:
"if (params.cloud) "
It will go to else

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Removed

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Added

Integrations/integration-opswat-metadefender.yml
Itay4: No, we changed it.

Integrations/integration-opswat-metadefender.yml
Itay4: `params.cloud` will be `undefined` for existing instances, as they won't get the default value you set for it. You should handle that before.

dem-evgeny: And if it will be undefined to where it will go? To else...

Itay4: Yes, but it's bad practice.

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Changed

Integrations/integration-opswat-metadefender.yml
dem-evgeny: Chnaged

dem-evgeny:
@Itay4 thanks :)

PR #1747, opened 2018-07-02
Iterate over TC Owners when tagging indicator

secdevopsai:
tc-tag-indicator This function would fail sometimes because it does not go through each owner in TC


asafshen:
@secdevopsai looks good

I'm not familiar with the threat connect python implementation  @meirwah can you please review also?

Integrations/integration-ThreatConnect.yml
```diff
@@ -297,6 +298,12 @@ script:
         indicators = tc.indicators()
         filter1 = indicators.add_filter()
         filter1.add_indicator(demisto.args()['indicator'])
+        owners_arr = []
```
meirwah: You should provide a way to override this with argument:
```
owners = ''
        if demisto.get(demisto.args(), 'owners'):
            owners = demisto.args()['owners']
        if len(owners) > 0:

```

secdevopsai: @meirwah - added

Itay4:
@meirwah Can you please review ?

PR #1750, opened 2018-07-03
Top malicious ratio indicators

erezh31:


meirwah:
@erezh31 build fails:
```
Failed: Scripts/script-TopMaliciousRatioIndicators.yml failed
<SchemaError: error code 2: Schema validation failed:
 - Key 'fromVersion' was not defined. Path: ''.: Path: '/'>
```

erezh31:
@asafshen I updated "script" schema validation. can you verify it? I think it was a bug

asafshen:
@erezh31 the schema validation is case sensitive (we cannot configure this to be insensitive)

I recommend to:
 - revete script schema yml
 - in your script - change the from/to version to match the schema (`fromversion` `toversion`)

BTW -- ideally, it should be identical in all artifacts, unfortunately it is not

erezh31:
@meirwah CI failed on some playbook, not related to the script
can you plz review and merge?

Scripts/script-TopMaliciousRatioIndicators.yml
```diff
@@ -0,0 +1,39 @@
+commonfields:
+  id: TopMaliciousRatioIndicators
+  version: -1
+fromversion: "4.0.0"
+name: TopMaliciousRatioIndicators
+script: |
+  from datetime import datetime, timedelta
+  import json, random
+
+  MAX_INDICATORS = 10000
```
meirwah: make this an argument

Scripts/script-TopMaliciousRatioIndicators.yml
```diff
@@ -0,0 +1,39 @@
...
+  MIN_NUMBER_OF_INVS = 3
```
meirwah: make this an argument

Scripts/script-TopMaliciousRatioIndicators.yml
```diff
@@ -0,0 +1,39 @@
...
+  demisto.results(json.dumps({"total": len(sorted_indicators), "data": sorted_indicators}))
+type: python
+tags: []
+comment: |-
+  Find the top malicious ratio indicators.
+  Malicious ratio is defined by the ratio between the number of "bad" incidents divided by the number of total number of incidents that the indicators appears in.
+enabled: true
+scripttarget: 0
+runonce: false
+releaseNotes: "Add new script"
```
meirwah: make this adding malicious ratio indicators script

meirwah:
@erezh31 plz fix and merge

Scripts/script-TopMaliciousRatioIndicators.yml
erezh31: It's for dashboard use - we can't really set custom argument. we still want to add this?

Scripts/script-TopMaliciousRatioIndicators.yml
erezh31: It's for dashboard use - we can't really set custom argument. we still want to add this?

Scripts/script-TopMaliciousRatioIndicators.yml
meirwah: if this is a dashboard script, aren't you missing:
```
tags:
- widget
```
?

erezh31:
@meirwah added the widget and tested together.
plz review the widget and changes

meirwah:
fix build first @erezh31 : 
image

erezh31:
@meirwah fixed

PR #1757, opened 2018-07-05
Zoom support within Demisto

slavikm:
@rishibhargava started and I'm cleaning up...
@meirwah we need to create a docker image for jwt but in I would like to make sure our default image has both jwt and BeautifulSoup as they are common requirements. 

Fixes: demisto/etc#12513


anara123:
@slavikm the build is red. missing description

DeanArbel:
Fixes: https://github.com/demisto/etc/issues/12513

DeanArbel:
It seems the zoom-fetch-recording command doesn't work:
image
image
image

PR #1758, opened 2018-07-05
add systemAssociatedTypes to fields schema

asafshen:
related to https://github.com/demisto/etc/issues/11804


asafshen:
right, but it will save few failed tests I think :)

anara123:
@asafshen merge from master

PR #1760, opened 2018-07-05
Access Investigation - Generic

JonathanMeler:
New playbooks:
* Access Investigation - Generic
* Access Investigation - QRadar

Updated playbooks:
* IP Enrichment - Generic

New script:
* IPToHost

Updated script:
* EmailAskUser


JonathanMeler:
fixes: https://github.com/demisto/etc/issues/12171

PR #1761, opened 2018-07-05
use the added command

liorblob:
depends on https://github.com/demisto/web-client/pull/3917


liorblob:
solves https://github.com/demisto/etc/issues/11266

liorblob:
@anara123 Argument is now deprecated

PR #1765, opened 2018-07-05
Postgres fix error

liorblob:


anara123:
@liorblob add here related issue link

liorblob:
solves https://github.com/demisto/etc/issues/10914

Integrations/integration-PostgreSQL.yml
```diff
@@ -71,17 +71,27 @@ script:
...
         try:
             cur = conn.cursor()
             cur.execute(demisto.args()['query'])
-            rows = cur.fetchall()
-            colnames = [desc[0] for desc in cur.description]
+            # try to fetch rows
+            try:
+                colnames = [desc[0] for desc in cur.description]
+                rows = cur.fetchall()
+            except:
```
anara123: catch the error and ignore only errors which return `no results to fetch`
if the command failed because other reason, then raise the original error.

PR #1766, opened 2018-07-05
Fixed move-between-mailboxes using impersonation

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/11896


aviadl:
@BenJoParadise nice solution !

PR #1768, opened 2018-07-09
Crowdstrike falcon intel v2 support

yardensade:


yardensade:
Tested migration from from old integration to new with both v1 and v2 api keys

PR #1773, opened 2018-07-10
new widget should be predefined

meirwah:


orius123:
We should add a validation for it

PR #1778, opened 2018-07-10
malicious ration reputation script

erezh31:


erezh31:
@meirwah plz review

Scripts/script-MaliciousRatioReputation.yml
```diff
@@ -0,0 +1,48 @@
...
+      }
+      demisto.results({
+          'Type': entryTypes['note'],
+          'EntryContext': ec,
+          'Contents': 'Malicious ratio for %s is %.2f' % (indicator['value'], mr_score),
+          'ContentsFormat': formats['text']
+      })
+type: python
+tags:
+- reputation
```
meirwah: can it be also enhancement script?

erezh31:
@meirwah ready to review. changed to work well as reputation script (and also can used as script part of playbook...)

Scripts/script-MaliciousRatioReputation.yml
```diff
@@ -0,0 +1,55 @@
...
+                  'Score': 2, #suspicious
+                  'Vendor': 'DBot-MaliciousRatio',
+                  'Indicator': indicator['value']
+              }
+              entry = {
+                  'Type': entryTypes['note'],
+                  'EntryContext': ec,
+                  'Contents': ec['DBotScore']['Score'],
+                  'ContentsFormat': formats['text'],
+                  'HumanReadable': 'Malicious ratio for %s is %.2f' % (indicator['value'], mr_score),
```
meirwah: add here the score result, is suspcious / malicous/ good

erezh31: actually we said the Malicious ratio will set only to suspcious, and the mr_score is the  Malicious ratio itself (value between 0-1)

PR #1782, opened 2018-07-11
updating default value of API version to 2.0

omercnet:
1.6 is no longer available, 2.0 is the default version in the hosted environment


dorsha:
Missing release notes

PR #1784, opened 2018-07-11
Convert Incident fields to array

asafshen:
Convert incident fields to array to support on 3.1.0

from `content_new.zip`
image


asafshen:
@aviadl its done before moving artifacts so I think its ok

also from the tests:
image

PR #1786, opened 2018-07-11
Added script to remove a user from investigation team

gshriki:
Can be used by users in the Incidents table (Run Command button) to leave incidents. Can also be used to remove users from investigations during a playbook.

### DIFF
--- a/Scripts/script-DemistoLeaveInvestigation.yml
+++ b/Scripts/script-DemistoLeaveInvestigation.yml
@@ -1,0 +1,62 @@
+commonfields:
+  id: DemistoLeaveInvestigation
+  version: -1
+name: DemistoLeaveInvestigation
+script: |-
+  var body = {};
+
+  if (args.InvestigationID) {
+      var invID = args.InvestigationID;
+      var res = executeCommand("getIncidents", {id: invID});
+      if (isError(res[0])) {
+          return res[0].Contents;
+      }
+      var owner = res[0].Contents.data[0].owner;
+
+  } else {
+      var invID = incidents[0].id;
+      var owner = incidents[0].owner;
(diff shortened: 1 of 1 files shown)


Scripts/script-DemistoLeaveInvestigation.yml
```diff
@@ -0,0 +1,62 @@
...
+      }
+      var owner = res[0].Contents.data[0].owner;
+
+  } else {
+      var invID = incidents[0].id;
+      var owner = incidents[0].owner;
+  }
+
+  if (args.User) {
+      var user = args.User;
```
aviadl: Declare user once before the condition, and then use same object

Scripts/script-DemistoLeaveInvestigation.yml
```diff
@@ -0,0 +1,62 @@
...
+      var uri = '/investigation/' + invID + '/deleteuser/' + user;
+
+      var res = executeCommand('demisto-api-post', {uri: uri, body: body});
+      if (isError(res[0])) {
+          return res[0].Contents;
+      }
+
+      return 'Successfully removed ' + user + '.';
+  } else {
+      return 'User ' + user + ' is the owner.';
```
aviadl: Dont you want an error in that case?

gshriki: no. this can be run with the RunCommand on a lot of incidents (to clear left pane), so don't want to have errrors...

Itay4:
@gshriki @aviadl What is the status of this?

dorsha:
Why do we need this? We already added this ability as part of the product. I won't merge this, better to use the built in command.

PR #1787, opened 2018-07-11
DeleteContext - added the ability to provide keys to keep

gshriki:
Useful when want to keep File or other important keys that are auto-populated


Scripts/script-DeleteConext.yml
```diff
@@ -44,4 +47,8 @@ args:
   - "yes"
   - "no"
   description: if you choose yes - all context will be deleted
+- name: keysToKeep
+  description: Context keys (comma-separated) to keep when deleting all context
```
aviadl: I would mention that this is only first level keys

meirwah:
@aviadl @gshriki can be merged?

gshriki:
sec. let me add what aviad suggested (missed that)

PR #1788, opened 2018-07-11
Ews readable errors

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/12355

Per @gshriki's request:
- When clicking "Test Module" on EWSv2, if the test fails, only the error message will be printed now instead of all the debugging info (stack trace, full log, exceptions etc). 
- Since the default instance is configured for office365, the **authentication type** field now reflects this default.


Integrations/integration-EWSv2.yml
```diff
@@ -37,7 +37,7 @@ configuration:
   required: true
 - display: Authentication Type (NTLM or Basic or Digest). Use Basic for Office365.
   name: authType
-  defaultvalue: NTLM
+  defaultvalue: Basic
```
anara123: Do we on purpose changed the defaultvalue from NTLM to Basic?

PR #1791, opened 2018-07-11
Update script-MatchRegex.yml

LPettay:
The previous version only returns the first match (even using the 'g' flag), so some looping was required to return all matches. I also added an output.

### DIFF
--- a/Scripts/script-MatchRegex.yml
+++ b/Scripts/script-MatchRegex.yml
@@ -19,13 +19,13 @@ script: |-
           if (args.contextKey) {
               ec[args.contextKey] = val;
           }
+          vals.push(val);
       }
   }
+  setContext('MatchRegex.results', vals)
   if (vals.length) {
-      setContext('MatchRegex.results', vals)
       return {Type: entryTypes.note, Contents: vals, ContentsFormat: formats.text, EntryContext: ec};
   }
-  setContext('MatchRegex.results', vals)
   return {Type: entryTypes.note, Contents: 'Regex does not match', ContentsFormat: formats.text};
 type: javascript
 tags:
--- a/Scripts/script-MatchRegex.yml
+++ b/Scripts/script-MatchRegex.yml
@@ -4,7 +4,7 @@ commonfields:
 name: MatchRegex
 script: |-
   var data = (typeof args.data) === 'string' ? args.data : JSON.stringify(args.data);
-  var flags = args.flags ? args.flags : 'im';
+  var flags = args.flags ? args.flags : 'gim';
   var r = new RegExp(args.regex, flags);
   var m;
   var vals = [];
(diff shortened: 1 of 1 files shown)


Scripts/script-MatchRegex.yml
```diff
@@ -6,37 +6,51 @@ script: |-
...
+              ec[args.contextKey] = val;
+          }
       }
-      var ec = {};
-      if (args.contextKey) {
-          ec[args.contextKey] = val;
-      }
-      return {Type: entryTypes.note, Contents: val, ContentsFormat: formats.text, EntryContext: ec};
   }
+  if (vals.length) {
```
BenJoParadise: val is never added into vals, causing vals to always be empty.

Scripts/script-MatchRegex.yml
```diff
@@ -6,37 +6,51 @@ script: |-
...
 tags:
 - Utility
-comment: Extract regex data from given text - supports groups as well
+comment: Extract regex data from given text - supports groups and looping as well.
+  May need to provide 'g' in flags argument.
 enabled: true
 args:
 - name: data
   required: true
-  default: true
```
BenJoParadise: For backwards compatibility, let's keep this as the default for the script.

Scripts/script-MatchRegex.yml
```diff
@@ -6,37 +6,51 @@ script: |-
...
   description: The regex to match and extract
 - name: group
-  description: The matching group to return (if not provided the full match will be returned), group value should start at 1
+  description: The matching group to return (if not provided the full match will be
+    returned), group value should start at 1
 - name: contextKey
   description: The context key to populate with the result
 - name: flags
-  description: The regex flags to match. Default is - i.
+  description: The regex flags to match. Default is -i.
```
BenJoParadise: The default is -im
(i know it was like that before your edit, but if you're already fixing this line, might as well fix that part too, thanks!)

Scripts/script-MatchRegex.yml
```diff
@@ -6,37 +6,51 @@ script: |-
...
+  defaultValue: 
+outputs:
+- contextPath: MatchRegex.results
+  description: List of Regex matches
+  type: string
 scripttarget: 0
+runonce: false
```
BenJoParadise: At the bottom of the file, please add a release note (without preceding tabs).
You can use the example below, or write your own (just make sure it's clear to other users):
`releaseNotes: "Added the option to return all matches."`

Scripts/script-MatchRegex.yml
LPettay: Whoops, I had some dedup code that I realized wouldn't be ideal for every scenario, that's where I was adding into vals before I got rid of it.

Scripts/script-MatchRegex.yml
```diff
@@ -6,24 +6,32 @@ script: |-
...
   var r = new RegExp(args.regex, flags);
-  var m = r.exec(data);
-  if (m && m.length > 0) {
-      var val = m[0];
-      var group = parseInt(args.group);
-      if (args.group && m.length > group) {
-          val = m[group];
+  var m;
+  var vals = [];
+  while ((m = r.exec(data)) !== null) {
```
BenJoParadise: !MatchRegex_copy data="asd" regex="asd" causes infinite loop, i suspect it happens because r.exec() always returns a value without the 'g' flag.

LPettay: I've fixed this with the latest changes

PR #1795, opened 2018-07-12
IsIPinRanges - better handling of ip ranges array

GavrielFilippov:
customer had a problem where a spaces or new lines in the provided list caused the command to return wrong output


Scripts/script-IsIPInRanges.yml
```diff
@@ -20,7 +20,7 @@ script: |
   }
 
   var isInRange = ipranges.some(function(iprange) {
-      return isIPInSubnet(ip, iprange);
+      return isIPInSubnet(ip, iprange.replace(/\n/g,'').replace(/\r/g,'').replace(/ /g,''));
```
aviadl: This will work, but why not create one regex for this [\n\g ] ?

Itay4:
@aviadl Can you please review?

Scripts/script-IsIPInRanges.yml
```diff
@@ -41,7 +41,7 @@ tags:
 - ip
 - Condition
 comment: Returns yes if ip is in one of the ranges provided, returns no otherwise.
-system: true
+enabled: true
```
aviadl: change this line back to system:true

Scripts/script-IsIPInRanges.yml
```diff
@@ -60,3 +60,5 @@ outputs:
 - contextPath: IP.InRange
   description: Is the IP is in the input ranges? (could be 'yes' or 'no)
 scripttarget: 0
+runonce: false
```
aviadl: remove this line

Scripts/script-IsIPInRanges.yml
```diff
@@ -60,3 +60,5 @@ outputs:
...
+releaseNotes: "Better handling of IP ranges array"
```
aviadl: I think that this release note is incorrect
At lease this is not what your fix is doing

Itay4:
@GavrielFilippov What is the status of this?

PR #1803, opened 2018-07-15
Postgres: Replaced username/password fields with Credentials store

omercnet:
This is breaking change when upgrading an existing instance

fixes https://github.com/demisto/etc/issues/12430

### DIFF
--- a/Integrations/integration-PostgreSQL.yml
+++ b/Integrations/integration-PostgreSQL.yml
@@ -39,17 +39,22 @@ script:
           conn.close()
 
     def pgsqlConnect():
-        connectionString = "dbname='" + demisto.params()['dbname'] +"' ";
-        connectionString +="user='" + demisto.params()['credentials']['identifier'] +"' ";
-        connectionString +="host='" + demisto.params()['host'] +"' ";
-        connectionString +="port='" + str(demisto.params()['port']) +"' ";
-        connectionString +="password='" + demisto.params()['credentials']['password'] +"'";
+        params = demisto.params()
+        username = params.get('username') or params.get('credentials')['identifier']
+        password = params.get('password') or params.get('credentials')['password']
+
+        connectionString = "dbname='" + params.get('dbname') + "' ";
+        connectionString +="user='" + username + "' ";
+        connectionString +="host='" + params.get('host') + "' ";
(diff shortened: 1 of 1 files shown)


gshriki:
@omercnet - let's either add a new integration, or provide a new YML to the customer. I do not want to break BC

dorsha:
I think there is a simple way to handle it. Let's see tomorrow.

dorsha:
No need for new integration or something like that.

gshriki:
even better!

gshriki:
@omercnet - discussed with Dorsha. Need to create logic in the integration itself to look for the old parameters if the new ones are not there. this way the integration can be backward compatible.

dorsha:
1. It can be mandatory 
2. The backward check is only for old users which already have saved instance of the integration with the user/password fields

Integrations/integration-PostgreSQL.yml
```diff
@@ -44,17 +39,22 @@ script:
           conn.close()
 
     def pgsqlConnect():
-        connectionString = "dbname='" + demisto.params()['dbname'] +"' ";
-        connectionString +="user='" + demisto.params()['user'] +"' ";
-        connectionString +="host='" + demisto.params()['host'] +"' ";
-        connectionString +="port='" + str(demisto.params()['port']) +"' ";
-        connectionString +="password='" + demisto.params()['password'] +"'";
+        params = demisto.params()
+        username = params.get('username') or params.get('credentials')['identifier']
```
omercnet: @dorsha something like this?

dorsha: Yes, just need to verify it works (and understand the using problem)

omercnet:
tested locally and it works
the "using" issue is solved when recreating the instance

I'm guessing it's related to the  upgrade error issue but I'm guessing it doesn't apply to content update, is there a way to test/simulate a content update ?

dorsha:
Yes will send you the details soon

anara123:
@dorsha @omercnet what is the status of the pr?

omercnet:
Blocked by https://github.com/demisto/etc/issues/12504

Itay4:
Closing and keeping branch alive for now, until https://github.com/demisto/etc/issues/12504 is resolved.

PR #1810, opened 2018-07-16
Email ask user should support now PB id as well

aviadl:
keep old script for backward compatibility


anara123:
@aviadl merge from master

PR #1813, opened 2018-07-17
Update process email + phishing layout

JonathanMeler:
* And `HTML Rendered Image` MD field to phishing layout
* Add Base64 output to the `rasterize-email` command
* Update rasterized image to the Phishing summary page

TODO:
* Remove the HTML field mapping in the relevant integrations

Fixes: https://github.com/demisto/etc/issues/12327

### DIFF
--- a/IncidentFields/incidentfields.json
+++ b/IncidentFields/incidentfields.json
@@ -1,4 +1,5 @@
 {
+	"releaseNotes": "Add HTML Image field",
 	"incidentFields": [{
 		"id": "incident_attachmenthash",
 		"version": 1,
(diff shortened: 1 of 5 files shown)


aviadl:
I tend to agree

JonathanMeler:
The base64 is truncated in the context when opened, so the browser won't get stuck when viewed by user. Also the Image created is around 10k in size, so I don't think it will burden the BE.

How can I had the tag filtering field? I'll give it a chance

@liorkol @trorabaugh what do you guys think?

JonathanMeler:
I meant how can I show those specific entries in the Summary page?

slavikm:
You can add to the layout `Incident Timeline` with filter on the tag - not sure if you can specify any filter but I think you can.

PR #1814, opened 2018-07-17
Passive Total - added proxy and insecure parameters

gshriki:
Added Proxy and Insecure parameters to the HTTP call


Integrations/integration-PassiveTotal.yml
```diff
@@ -46,7 +56,9 @@ script:
                 Username: params.Username,
                 Password: params.APIKey,
                 Body: body ? body : ''
-            }
+            },
+            args.insecure,
+            args.proxy
```
dorsha: Add `|| false` (for existing instances without these params configured)

slavikm:
@gshriki can you review my additions - I've taken over this PR :-)
Fixed URL to call enrichment as well as added domain and IP support. Also, cleaned a bit and added more context data where relevant.
Fixes https://github.com/demisto/etc/issues/12063

Generally, should add a lot of the new API functionality at http://api.passivetotal.org/api/docs/#api-Enrichment-GetV2Enrichment

PR #1816, opened 2018-07-17
Python3 Docker for TextFromHTML

secdevopsai:
Proposing that we add the Python3 docker for this automation script.  Python 2.7 frequently returns errors for decoding strings as input


slavikm:
@secdevopsai probably related to the changes to the way strings are treated in Python (unicode vs. bytes). Without looking at the code, we might need to treat re as unicode (`ur'reg-string'`) and also specify re.U flag...

secdevopsai:
@Itay4  - I added the changes you requested.  Can you review ?

Thanks
-Ron

PR #1821, opened 2018-07-18
check proxy parameter before client.connect()

pvacey:
Attempting to connect to the splunk server before checking for the proxy parameter causes a connection timeout if the splunk server is not accessible without using a proxy.


meirwah:
@pvacey Thank you very much for this fix contribution! highly appreciated !

I merged this PR code in a different PR: https://github.com/demisto/content/pull/1824
Just added release notes to it.

👍

PR #1822, opened 2018-07-18
Fixed unnamed attachments bug

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/12519

### DIFF
--- a/Integrations/integration-EWSv2.yml
+++ b/Integrations/integration-EWSv2.yml
@@ -185,6 +185,11 @@ script:
             return dict((k, v) for k, v in d.items() if v is not None)
         return d
 
+    def get_attachment_name(attachment_name):
+        if not attachment_name:
+            return 'demisto_untitled_attachment'
+        return attachment_name
+
 
     def get_entry_for_object(title, context_key, obj, headers=None):
         if len(obj) == 0:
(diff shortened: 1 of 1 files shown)


Integrations/integration-EWSv2.yml
```diff
@@ -185,6 +185,11 @@ script:
             return dict((k, v) for k, v in d.items() if v is not None)
         return d
 
+    def get_attachment_name(attachment_name):
+        if not attachment_name:
```
hirshbergshachar: @BenJoParadise if the attachment name would be 0 we will return 'demisto_untitled_attachment' instead of 0. Please fix that scenario if relevant.

Integrations/integration-EWSv2.yml
```diff
@@ -185,6 +185,11 @@ script:
             return dict((k, v) for k, v in d.items() if v is not None)
         return d
 
+    def get_attachment_name(attachment_name):
+        if not str(attachment_name):
```
hirshbergshachar: @BenJoParadise if attachment_name is None this will cause it to become 'None' and then the condition would be false instead of true.

I think a better solution here would be something like:
`if not (attachment_name == 0 and attachment_name):`

PR #1831, opened 2018-07-22
Consider fromversion field in release notes generation

Itay4:
Fixes https://github.com/demisto/etc/issues/12385


release_notes.py
```diff
@@ -165,13 +169,23 @@ def load_data(self, data):
     def get_header(self):
         return "Scripts"
 
-    def added_release_notes(self, cnt):
+    def added_release_notes(self, cnt, demisto_version, first_release):
+        if first_release:
+            from_version = cnt.get("fromversion") or cnt.get("fromVersion")
+            if from_version:
+                if LooseVersion(from_version) == LooseVersion(demisto_version):
```
asafshen: if `from_version < demisto_version` we also want to include release notes (true all `added_release_notes`)

we have a lot of code duplication here
I recommend create this logic before calling `added_release_notes` / `modified_release_notes`

release_notes.py
```diff
@@ -165,13 +169,23 @@ def load_data(self, data):
     def get_header(self):
         return "Scripts"
 
-    def added_release_notes(self, cnt):
+    def added_release_notes(self, cnt, demisto_version, first_release):
+        if first_release:
```
asafshen: I think something is wrong with the logic here

if `first_release` is false - you still want to compare version against `demisto_version`

release_notes.py
```diff
@@ -317,7 +371,7 @@ def added_release_notes(self, cnt):
 
         return add_dot(rn) + "\n"
 
-    def modified_release_notes(self, cnt):
+    def modified_release_notes(self, cnt, demisto_version, first_release):
```
asafshen: we want this logic to be on all artifacts

PR #1839, opened 2018-07-23
Repopulate files

slavikm:
Simple script to repopulate File context from file entries after DeleteContext is called.


anara123:
Why would you need such script?

slavikm:
@anara123 see description - when you debug on a given incident and do `DeleteContext` and you want to repopulate `File` data

PR #1845, opened 2018-07-24
Add generic dedup playbook

erezh31:
improve dedup scripts


Scripts/script-FindSimilarIncidentsByText.yml
```diff
@@ -8,16 +8,16 @@ script: |-
   TEXT_FIELDS = set(map(lambda x: x.lower(), demisto.args()['textFields'].split(',')))
```
meirwah: all the changes in this file won't break backward?

Scripts/script-FindSimilarIncidents.yml
```diff
@@ -108,7 +108,7 @@ script: |-
   def get_incidents_by_keys(similar_incident_keys, incident_time, incident_id, hours_back, ignore_closed, max_number_of_results):
       similar_keys_query = " and ".join(map(lambda t: '%s:"%s"' % (t[0], t[1]), similar_incident_keys.items()))
       incident_time = parse_datetime(incident_time)
-      max_date = incident_time
+      max_date = incident_time + timedelta(hours=hours_back)
```
meirwah: all the changes in this file won't break backward?

erezh31: No.. it's just expand the query to search across more time (like the other scripts)

Scripts/script-FindSimilarIncidentsByText.yml
erezh31: no.. changed const in the code to arguments with the same default value as the consts

PR #1848, opened 2018-07-24
QRadar: Fix pagination

omercnet:
Pagination didn't properly go over all the outstanding incidents


Integrations/integration-QRadar.yml
```diff
@@ -235,13 +235,13 @@ script:
             var totalOffenses = res.length;
             var lastCallOffenses = res.length;
             while (startTime && lastCallOffenses >= offensesPerCall) {
-                var min = res.length;
-                var max = min + offensesPerCall;
+                var from = totalOffenses + 1;
+                var to = min + offensesPerCall;
```
hirshbergshachar: @omercnet does the new code work? it seems in this line that we are calling the variable `min`, but it doesn't exist anymore...

omercnet: thanks! fixed

anara123:
@omercnet we can do kind of mock inside the code itself. lets sit on it together

omercnet:
this isn't the latest fix - waiting for customer to test before pushing here

PR #1854, opened 2018-07-25
Create script-TestHTTPApi.yml

omercnet:
Test an HTTP endpoint and verify it returns a valid response.
The response can be verified to have a specific HTTP response code and a specific JSON field value


anara123:
@omercnet please add test playbook.

omercnet:
will stick with existing HTTP command instead

PR #1855, opened 2018-07-26
PassiveTotal integration tags parameter bug fix

Itay4:
Fixes https://github.com/demisto/etc/issues/12667

### DIFF
--- a/Integrations/integration-PassiveTotal.yml
+++ b/Integrations/integration-PassiveTotal.yml
@@ -86,6 +86,12 @@ script:
         params.proxy = true;
     }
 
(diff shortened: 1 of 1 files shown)


anara123:
@gshriki please review

Itay4:
@gshriki We could, but it would require changing some of the functionality added in https://github.com/demisto/content/pull/1814
Anyway, this pull request was created so such cases as you described won't happen, so I think we're good.
Do you still want it changed?

gshriki:
that's what I'm talking about.
think @slavikm mentioned the tag is not mandatory, but i see it is..

anara123:
@slavikm please review

PR #1865, opened 2018-07-29
added servicenow test playbook

yardensade:


anara123:
@yardensade add here screenshot of the test playbook

yardensade:
@anara123 
Here we go, ain't much to it :)
image

anara123:
add verifycontext task, which should verify if the command exports the
correct output

yardensade:
Current PR is closed until further advance with issue:
https://github.com/demisto/etc/issues/12678

PR #1879, opened 2018-07-31
JIRA Integration: Add support for creating sub tasks

pingunaut:
By allowing to specify a parent, one can create sub tasks of an existing JIRA issue


pingunaut:
Implemented like described in [Jira REST API examples](https://developer.atlassian.com/server/jira/platform/jira-rest-api-examples/#creating-a-sub-task)

Regards

PR #1888, opened 2018-08-02
Support same key splunk parse

meirwah:


Integrations/integration-SplunkPy.yml
```diff
@@ -6,6 +6,7 @@ display: SplunkPy
...
 description: Run queries on Splunk servers
+releaseNotes: "Fix issue where you use parseNotableEventsRaw did not handle same key in _raw event part"
```
Nitzperetz: No need for quotes

PR #1904, opened 2018-08-05
Fix QRadar Pagination

omercnet:
qradar returns offenses sorted desc on start_time and there's no way to change sorting.
if we get `offernsesPerCall` offenses it means we (probably) have more than that so we
start looking for the end of the list by doubling the page position until we're empty.
then start binary search back until you find the end of the list and finally return
`offensesPerCall` from the end.


DeanArbel:
fixes: https://github.com/demisto/etc/issues/12883
The latest change was tested using a mock server containing the data provided in the above issue.

omercnet:
@anara123 please review and approve

PR #1911, opened 2018-08-06
EWS Mail Sender - Added the ability to omit attachNames

gshriki:
Using the entry ID to get the filename if the argument is not provided


aviadl:
@gshriki test playbook ?
You can do it using email ask user flow, send an email that will be parsed by the same playbook, and then you can see that it had a proper file

gshriki:
Closing for now until we can handle it.

PR #1912, opened 2018-08-07
SCADAfence integration (#1910)

hirshbergshachar:
Integrations to provide access to SCADAfence CNM data for:
getting alerts
getting asset details
getting asset connections
getting asset traffic statistics
updating alerts status in SCADAfence CNM


Integrations/integration-SCADAfence_CNM.yml
```diff
@@ -0,0 +1,467 @@
...
+  type: 4
+  required: true
+- display: API auth key
+  name: APIKey
+  defaultvalue: 
+  type: 4
+  required: true
+- display: API url
+  name: APIUrl
+  defaultvalue: https://172.28.128.10/externalApi
```
anara123: why not to add `externalApi` in the integration itself?

Integrations/integration-SCADAfence_CNM.yml
```diff
@@ -0,0 +1,467 @@
...
+        get_alerts('Critical', None)
+        demisto.results('ok')
+        sys.exit(0)
+
+    elif demisto.command() == 'scadafence-getAlerts':
+
+        data = get_alerts(demisto.args()['severity'], None)
+        md = '## Following alerts were found\n'
+        md += tableToMarkdown('Alerts are: ', data)
+        # tableToMarkdown('header', [{}, {}, {}...])
```
anara123: remove this comment

Integrations/integration-SCADAfence_CNM.yml
hirshbergshachar: @AlexzxelA is the `externalApi` path constant? If so, perhaps we can set only the server URL here.

AlexzxelA: it is sure possible

hirshbergshachar: @anara123 @AlexzxelA concatenating it to the API url now.

hirshbergshachar:
@anara123 we have test env in place, all good here.

PR #1916, opened 2018-08-07
Changed default auth according to office365 defaults. Fixed undefined…

BenJoParadise:
… IS_TEST_MODULE. Fixed attachments requiring names.

Fixes https://github.com/demisto/etc/issues/12788
Fixes https://github.com/demisto/etc/issues/12787

### DIFF
--- a/Integrations/integration-EWSMailSender.yml
+++ b/Integrations/integration-EWSMailSender.yml
@@ -30,7 +30,7 @@ configuration:
   required: true
 - display: Authentication Type (NTLM or Basic or Digest). Use Basic for Office365.
   name: authType
-  defaultvalue: NTLM
+  defaultvalue: Basic
   type: 0
   required: true
 - display: Trust any certificate (unsecure)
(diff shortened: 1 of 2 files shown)


Integrations/integration-EWSMailSender.yml
```diff
@@ -141,10 +143,20 @@ script:
...
-            raise Exception("attachIDs and attachNames lists should be the same length")
+        file_entries_for_attachments = []
+        attachments_names = []
+        if attachIDs:
+            file_entries_for_attachments = attachIDs.split(",")
+            if attachNames:
+                attachments_names = attachNames.split(",")
+            else:
+                for att_id in file_entries_for_attachments:
+                    att_name = demisto.dt(demisto.context(), "File(val.EntryID === '%s').Name" % att_id)
```
gshriki: Is this the right way to do this?
You also get the name with the `getFilePath` response... the file might not be in the context...

PR #1919, opened 2018-08-08
Common server python fix

yaakovi:


yaakovi:
added to master by https://github.com/demisto/content/pull/1920

PR #1924, opened 2018-08-08
Add failed instances script

Nitzperetz:
Executes test for all integration instances available and returns a detailed table about failed integration instances.

Later on could be embedded within a playbook which could use in a job. (send email about results..)

Please review.


Nitzperetz:
@gshriki , it works on latest master. Please review.

Scripts/script-FailedInstances.yml
```diff
@@ -0,0 +1,72 @@
+commonfields:
+  id: FailedInstances
+  version: -1
+fromversion: 4.0.0
```
anara123: is there specific reason why this script is `fromversion: 4.0.0`?

Nitzperetz: @anara123 , yes, it uses the test module feature which is only supported in v4.

PR #1926, opened 2018-08-08
Virus Total test playbook - removed task to avoid api limit

hirshbergshachar:


hirshbergshachar:
Will move to nightly instead

PR #1929, opened 2018-08-09
Fix content version comparison

asafshen:
Fix content version comparison - because artifact with `fromversion: 4.0.0` were added  to `content_yml.zip`


anara123:
@asafshen what is the status here?

PR #1930, opened 2018-08-09
Cybereason integration

Itay4:
- Fixes https://github.com/demisto/etc/issues/12485
- Rewriting the integration in Python
- Adding malops fetch and commands
- Fixes https://github.com/demisto/etc/issues/13146


Itay4:
@JonathanMeler Please add new functionality to relevant playbooks.

JonathanMeler:
@Itay4 let's go over the integration so I could understand the use-cases.

PR #1931, opened 2018-08-09
Removed ATD Detonate subplaybook task

BenJoParadise:
ATD Detonate is not available in 3.6


asafshen:
@BenJoParadise 
please change `toversion` to `3.6.0`
image

PR #1932, opened 2018-08-09
Salesforce Ask User

Itay4:


Integrations/integration-Salesforce.yml
```diff
@@ -206,7 +206,7 @@ script:
             }
         }
         return createEntry(raw_info, {
-            contextPath: 'SalesForce.Lead(val.ID && val.ID == obj.ID)',
+            contextPath: 'SalesForce.Lead(val.ID && val.ID == obj.ID',
```
yaakovi: you have a bug here, put it back

Integrations/integration-Salesforce.yml
```diff
@@ -588,6 +600,7 @@ script:
                 lastRun.last_case_time = dateToString(new_time, "%Y-%m-%dT%H:%M:%S.%fZ");
 
                 setLastRun(lastRun);
+                logInfo(lastRun.last_case_time);
```
yaakovi: don't leave log records without proper description

Integrations/integration-Salesforce.yml
```diff
@@ -626,7 +639,6 @@ script:
                         var messageSegments = data.body.messageSegments;
                         for (var j = 0; j < messageSegments.length; j++)
                         if (messageSegments[j].type === 'Text') {
-                            // Found the relavent comment (there's only one), so we return it
```
yaakovi: why did you remove this comments?

PR #1934, opened 2018-08-09
MailSender support array

yaakovi:
fixes https://github.com/demisto/etc/issues/12853


aviadl:
@slavikm you screwed it, you should review it :)

PR #1937, opened 2018-08-09
Common Server python add regex formatting for hash,ip,email

bakatzir:
Common Server python add regex formatting for hash,ip,email


bakatzir:
new PR at : � https://github.com/demisto/content/pull/1945

PR #1944, opened 2018-08-12
Added support in missing username while assigning analyst

ronykoz:
fixes https://github.com/demisto/etc/issues/11998


ronykoz:
@aviadl I basically copied the Random choice option to the end of the switch block.
If there is no role it's supposed to get all the users. And in case the response returned error than it will be addressed at the following code segment, same as it addresses it in case the user selected "random" choice option.

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -11,6 +11,12 @@ script: |
   var emailToAssign = args.email;
   var userToAssign = args.username;
 
+  var getUserException = {
```
dorsha: You have indentations issues here.

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -11,6 +11,12 @@ script: |
...
+                          ContentsFormat: formats.text,
+                          Type: entryTypes.error,
+                          Contents: "Error from getUsers."
```
dorsha: You should add the actual error as well, users should understand what happened.

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -28,37 +34,60 @@ script: |
 
   assignBy = args.assignBy || 'random';
 
+  function chooseUser(usersRes){
```
dorsha: add space before the `{`

dorsha: change function name to `pickRandomUser`

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -28,37 +34,60 @@ script: |
...
-                    userToAssign = res.userLeastLoad;
-                    break;
-            }
-    }
+      switch(assignBy) {
+          case 'online':
+              var usersRes = executeCommand('getUsers', { roles: args.roles, online: true });
+              if (isError(usersRes[0])) {
+                  return getUserException;
+              } else {
```
dorsha: no need for else

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -28,37 +34,60 @@ script: |
...
+                  chooseUser(usersRes);
+              }
+              break;
+          case 'current':
+              var usersRes = executeCommand('getUsers', { current: true });
+              if (isError(usersRes[0])) {
+                  return getUserException;
+              } else {
```
dorsha: no need for else

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -28,37 +34,60 @@ script: |
...
+                  chooseUser(usersRes);
+              }
+              break;
+          case 'random':
+              var usersRes = executeCommand('getUsers', { roles: args.roles });
+              if (isError(usersRes[0])) {
+                  return getUserException;
+              } else {
```
dorsha: no need for else

Scripts/script-AssignAnalystToIncident.yml
dorsha: rename to getUserError

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -28,37 +34,60 @@ script: |
...
+                      userToAssign = res.topOwner;
+                      break;
+                  case 'less-busy-user':
+                      userToAssign = res.userLeastLoad;
+                      break;
+              }
+      }
+  }
+
+  if (!userToAssign) {
```
dorsha: Not sure about this part, if someone asks for online users and got empty response you might return here users which the caller is not expected.
It also breaks compatibility.

Scripts/script-AssignAnalystToIncident.yml
ronykoz: Fixed

ronykoz: Done

Scripts/script-AssignAnalystToIncident.yml
ronykoz: Done

Scripts/script-AssignAnalystToIncident.yml
ronykoz: Done

Scripts/script-AssignAnalystToIncident.yml
ronykoz: Done

Scripts/script-AssignAnalystToIncident.yml
ronykoz: Done

Scripts/script-AssignAnalystToIncident.yml
ronykoz: You are right, moved it to the end of the default of the first switch. This way we will run this function only in case the getOwnerSuggestion command returned an empty value.

dorsha:
@ronykoz next time don't forget to mention after fixing the code review, otherwise we won't be aware.
Good job.

PR #1949, opened 2018-08-13
Mcafee ATD add credentials from new major

bakatzir:
fix https://github.com/demisto/etc/issues/12772

DO NOT MERGE - until we'll know in which version update instance will be.


bakatzir:
@anara123  We talked with @gshriki and the issue is that for costumers that already have a configured 
instance, it will break.

regardless, i'll split PR and deal with the detonation deprecation in a different one, it's not really related.

anara123:
Currently in order to upgrade integration (new mandatory field) we must write code in go.

PR #1950, opened 2018-08-13
Update IsValueInArray script - try to parse string array

Nitzperetz:
1. Try to parse string array
2. Support numeric values

Following commands would not work as expected from CLI:
1. !IsValueInArray array=`["Phishing", "Malware"]` value=Malware
2. !IsValueInArray array=`[1, 2]` value=1


Nitzperetz:
Merged in another PR.

PR #1951, opened 2018-08-13
Pre commit hook

hirshbergshachar:


Tests/validate_release_notes.py
```diff
@@ -0,0 +1,92 @@
...
+def get_modified_files(files_string):
+    all_files = files_string.split('\n')
+    modified_files_list = []
+    for f in all_files:
+        file_data = f.split()
+        if not file_data:
+            continue
+        file_status = file_data[0]
+        file_path = file_data[1]
+        if file_status == "M" and not re.match(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE) \
```
asafshen: what about other modified statuses:
image

Tests/validate_release_notes.py
```diff
@@ -0,0 +1,92 @@
...
+
+def validate_file_release_notes(file_path):
+    data_dictionary = None
+    with open(file_path, "r") as f:
+        if file_path.endswith(".json"):
+            data_dictionary = json.load(f)
+        elif file_path.endswith(".yaml") or file_path.endswith('.yml'):
+            data_dictionary = yaml.safe_load(f)
+
+    if data_dictionary and data_dictionary.get('releaseNotes', None) is None:
```
asafshen: not critical but you can omit the `get` second argument

Tests/validate_release_notes.py
```diff
@@ -0,0 +1,92 @@
...
+except ImportError:
+    pip.main(['install', 'pyyaml'])
+    import yaml
+import json
+import sys
+import re
+from subprocess import Popen, PIPE
+
+# file types regexes
+INTEGRATION_REGEX = "(integration-).*(.yml)"
```
asafshen: we have this regex duplicate here, lets see if we can remove this duplication
I would call this script/logic from `validate_schema.py` (if `is_local` is true maybe)

Tests/validate_schema.py
```diff
@@ -1,14 +1,22 @@
...
         sys.exit(1)
 
     file_path = argv[0]
     schema_path = argv[1]
+    
+    is_local = False
+    if len(argv) == 3:
+        is_local = argv[2]
+
+    if is_local:
```
asafshen: note that if you pass `"False"`  string to this - it will go inside the if..
in python
```
if "False":
     print "I'm in"
```
will print "I'm in"

.hooks/pre-commit
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+
+# case insensitivity for regex 
+shopt -s nocasematch
+
+# file types regexes
+INTEGRATION_REGEX="(integration-).*(.yml)"
+PLAYBOOK_REGEX="^(?!testPlaybooks).*(playbook-).*(.yml)"
```
asafshen: need to make sure we don't have `yaml` artifacts (scripts, integrations, playbooks)

Tests/schemas/test-playbook.yml
```diff
@@ -0,0 +1,211 @@
+type: map
```
asafshen: we might be able to avoid all this duplication

we can use the `pykwalify` validation with schema data (and not path)
image

so we can read the playbook file, and change the relevant part instead duplicate this file

dorsha:
And update about this one?

hirshbergshachar:
@dorsha hoping to get to it today/tomorrow

.hooks/pre-commit
hirshbergshachar: We don't have - validated

hirshbergshachar:
@asafshen ready for review

Tests/scripts/validate_files_structure.py
```diff
@@ -0,0 +1,204 @@
...
+        print "File " + file_path + " is missing releaseNotes, please add."
+        return False
+    
+    return True
+
+def validate_schema(file_path, matching_regex=None):
+    if matching_regex is None:
+        for regex in CHECKED_TYPES_REGEXES:
+            if re.match(regex, file_path, re.IGNORECASE):
+                matching_regex = regex
```
asafshen: maybe add `break`?

Tests/scripts/validate_files_structure.py
```diff
@@ -0,0 +1,204 @@
...
+DASHBOARDS_DIR = "Dashboards"
+WIDGETS_DIR = "Widgets"
+INCIDENT_FIELDS_DIR = "IncidentFields"
+LAYOUTS_DIR = "Layouts"
+CLASSIFIERS_DIR = "Classifiers"
+MISC_DIR = "Misc"
+CONNECTIONS_DIR = "Connections"
+
+# file types regexes
+INTEGRATION_REGEX = "{}.*integration-.*.yml".format(INTEGRATIONS_DIR)
```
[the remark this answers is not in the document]
hirshbergshachar: @asafshen `%` is the old Python formatting, using `{}` has some advantages such as giving specific order by using `{0} {2} {1}.format(first,second,third)`

Tests/scripts/validate_files_structure.py
hirshbergshachar: @asafshen good catch

Tests/scripts/validate_files_structure.py
```diff
@@ -0,0 +1,204 @@
...
+        c = Core(source_file=file_path, schema_files=[SCHEMAS_PATH + REGEXES_TO_SCHEMA_DIC.get(matching_regex) + '.yml'])
+        try:
+            c.validate(raise_exception=True)
+            return True
+        except Exception as err:
+            print 'Failed: %s failed' % (file_path,)
+            print err
+            return False
+
+    print file_path + " doesn't match any of the known supported file prefix/suffix, please make sure that its naming is correct."
```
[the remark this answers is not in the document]
hirshbergshachar: yes

Tests/scripts/validate_files_structure.py
```diff
@@ -0,0 +1,204 @@
...
+    all_files = files_string.split('\n')
+    modified_files_list = []
+    for f in all_files:
+        file_data = f.split()
+        if not file_data:
+            continue
+        file_status = file_data[0]
+        file_path = file_data[1]
+        if file_status in ACTIONABLE_FILE_STATUSES:
+            modified_files_list.append(file_path)
```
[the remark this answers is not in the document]
hirshbergshachar: @asafshen tried those, couldn't replicate - ended up with 'M' or 'A' & 'D'
image
image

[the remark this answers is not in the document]
asafshen: need to handle same way as `release_notes.py` handle those statuses

Tests/scripts/validate_files_structure.py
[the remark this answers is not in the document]
yaakovi: @hirshbergshachar noted, thanks

PR #1957, opened 2018-08-14
added timeout because sometime it isn't enough

bakatzir:
Test is not stable,
sometimes it times out so increased a little the timeout


bakatzir:
will do both changes in https://github.com/demisto/content/pull/1956/files

PR #1968, opened 2018-08-16
Bug fix, presenting in summary To instead of Subject

JonathanMeler:
Change it to the right field


JonathanMeler:
fixes: https://github.com/demisto/etc/issues/12948

PR #1971, opened 2018-08-16
Fix parse csv

liorblob:
solves https://github.com/demisto/etc/issues/12730


liorblob:
the order of the csv columns were switched sometimes (inconsistently), maybe worth opening another issue

PR #1972, opened 2018-08-16
Content tests moved to skipped (RSA packets and logs + Joe Security + VirtusTotal)

liorblob:
Temporary until we get data in the instance
@NeelRust

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -80,10 +80,6 @@
             "timeout": 240
         },
         {
-            "integrations": "RSA NetWitness Packets and Logs",
-            "playbookID": "rsa_packets_and_logs_test"
-        },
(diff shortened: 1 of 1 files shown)


dorsha:
Skip that? @liorblob make sure to have proper commit message (and PR header as well). So it will be easier to understand after.

liorblob:
@dorsha you're right, but it's temporary until we get data in the instance. Will add explanation.

DeanArbel:
VirusTotal and Joe Security were moved to skipped due to the fact they're constantly failing at the moment because we are out of quota (rate limit).
Joe Security should be returned to nightly at September.
VirusTotal needs to be further investigated before returned to the test pool.

PR #1974, opened 2018-08-17
Update script-RepopulateFiles.yml

xamiel:
Swapping File.Info and File.Type

When an incident is first created the File attributes are as follows:
Info: application/pdf
Type: PDF document, version 1.7

If you delete the context with !DeleteContext and then run !RePopulateFiles, this gets turned around.
Since !RePopulateFiles is usually run on the Playground, this makes testing File.Info or File.Type not work properly on the Playground when compared to the context in an Incident.

Unless/Until the context data between the Playground an Incident matches, this seems like a good interim fix.


xamiel:
It does appear however that the initial Incident data has it wrong/backwards.
Type should be the short/curt version; application/pdf
Info should be the long version; PDF document, version 1.7

PR #1975, opened 2018-08-17
Automation for showing all scheduled entries per incident

dorsha:


gshriki:
This is nice. I did something similar for Nike, for clearing schedule. Maybe we should add this capability in this script, as an argument?

PR #1978, opened 2018-08-19
Ignore errors that indicate o365 maintenance operations

BenJoParadise:
fixes https://github.com/demisto/etc/issues/12975

Also changes default version to match other defaults for office 365.

### DIFF
--- a/Integrations/integration-EWSv2.yml
+++ b/Integrations/integration-EWSv2.yml
@@ -95,7 +95,7 @@ script:
     logging.basicConfig(stream=log_stream, level=logging.DEBUG)
 
     from exchangelib.errors import ErrorItemNotFound, ResponseMessageError, TransportError, \
-        ErrorFolderNotFound, ErrorToFolderNotFound, ErrorMailboxStoreUnavailable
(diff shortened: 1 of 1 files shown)


Integrations/integration-EWSv2.yml
```diff
@@ -1220,6 +1220,20 @@ script:
         debug_log = log_stream.getvalue()
         error_message_simple = ""
         error_message = ""
+
+        # Office365 regular maintenance case
+        if (isinstance(e, ErrorMailboxStoreUnavailable) or isinstance(e, ErrorMailboxMoveInProgress)) and 'outlook.office365.com' in EWS_SERVER:
+            log_message = "Office365 is undergoing load balancing operations. As a result, the service is temporarily unavailable."
+            demisto.info(log_message)
```
hirshbergshachar: @BenJoParadise why do we log it? How long can the maintenance take? If it's few moments I guess we would simply fetch in the next minute (the fetch part here is ok) but otherwise, we will see the result from error_message_simple anyway

BenJoParadise: They don't say how long maintenance can take.
Moved the logging to the fetch_incidents case.

PR #1979, opened 2018-08-19
Ews error counter

BenJoParadise:
Fixes https://github.com/demisto/etc/issues/12980 for fetch-incidents


Integrations/integration-EWSv2.yml
```diff
@@ -143,6 +143,7 @@ script:
     LAST_RUN_TIME = "lastRunTime"
     LAST_RUN_IDS = "ids"
     LAST_RUN_FOLDER = "folderName"
+    ERROR_COUNTER = "errors"
```
hirshbergshachar: @BenJoParadise please change the value to `errorCounter`/`errorAmount` - something that would match the naming

Integrations/integration-EWSv2.yml
```diff
@@ -715,24 +716,39 @@ script:
...
+                LAST_RUN_TIME: start_time.ewsformat(),
+                LAST_RUN_FOLDER: folder_name,
+                LAST_RUN_IDS: ids,
+                ERROR_COUNTER: 0
+            }
+
+            demisto.setLastRun(new_last_run)
+            return incidents
+
+        except RateLimitError:
```
hirshbergshachar: @BenJoParadise where is `RateLimitError` defined? I don't see it here

Integrations/integration-EWSv2.yml
```diff
@@ -715,24 +716,39 @@ script:
...
 
-        demisto.setLastRun(new_last_run)
-        return incidents
+        try:
+            last_emails = fetch_last_emails(account, folder_name, last_run.get(LAST_RUN_TIME), last_run.get(LAST_RUN_IDS))
+
+            ids = []
+            incidents = []
+            for item in last_emails:
+                if item.message_id:
```
hirshbergshachar: @BenJoParadise if there's no message_id this line won't raise an error? Asking because of the dot notation

Integrations/integration-EWSv2.yml
```diff
@@ -715,24 +716,39 @@ script:
...
+                ERROR_COUNTER: 0
+            }
+
+            demisto.setLastRun(new_last_run)
+            return incidents
+
+        except RateLimitError:
+            if LAST_RUN_TIME in last_run:
+                last_run[LAST_RUN_TIME] = last_run[LAST_RUN_TIME].ewsformat()
+            demisto.info(str(last_run))
```
hirshbergshachar: @BenJoParadise why do we log it here? And if you decide to log please add something indicative such as `EWS Last run was + str(last_run)`

Integrations/integration-EWSv2.yml
```diff
@@ -715,24 +716,39 @@ script:
...
+            if ERROR_COUNTER not in last_run:
+                last_run[ERROR_COUNTER] = 0
+            last_run[ERROR_COUNTER] += 1
```
hirshbergshachar: @BenJoParadise did you verify that this part work? I don't remember if last_run dictionary returns as strings or ints, if it's strings you will encounter an error here

Integrations/integration-EWSv2.yml
```diff
@@ -715,24 +716,39 @@ script:
...
+            demisto.setLastRun(last_run)
+            if last_run[ERROR_COUNTER] > 2:
+                raise
```
hirshbergshachar: @BenJoParadise please raise an indicative error here such as `EWS encountered more than two errors in fetch....`

Integrations/integration-EWSv2.yml
BenJoParadise: It's an object property. The object will always have this property, but it might be None, in which case the predicate will be false.

Integrations/integration-EWSv2.yml
BenJoParadise: Verified, this works (i can see the number incrementing in the indicator)

Integrations/integration-EWSv2.yml
BenJoParadise: It is important to raise the exact same error, it is indicative of a configuration, connectivity or authentication error.
Also, raising a different exception will overwrite the stack trace.

Integrations/integration-EWSv2.yml
BenJoParadise: Fixed in aef2a3d

Integrations/integration-EWSv2.yml
BenJoParadise: Fixed in 7dd3d15

PR #1980, opened 2018-08-20
Fixed DeleteContext to handle sub-playbook

dorsha:
fixes: https://github.com/demisto/etc/issues/12982


Scripts/script-DeleteConext.yml
```diff
@@ -30,7 +31,11 @@ script: |
           return "Context cleared";
       }
   } else {
-      return executeCommand('delContext', { key: args.key });
+      var key = args.key;
+      if (isSubPlaybookKey) {
+        key = 'subplaybook-${currentPlaybookID}.' + key;
```
aviadl: Are you sure that it will be replaced correctly this way?
Basically we need something like:
${currentPlaybookID='subplaybook-'+val+'.' + key}

Scripts/script-DeleteConext.yml
```diff
@@ -46,8 +51,15 @@ args:
...
+  auto: PREDEFINED
+  predefined:
+  - "yes"
+  - "no"
+  description: Choose yes if the context key is inside a sub-playbook
 - name: keysToKeep
   description: Context keys (top-level keys) to keep when deleting all context. Use comma to specify multiple values 
 scripttarget: 0
 runonce: false
+releaseNotes: "-"
```
aviadl: Why not add release notes ?

Scripts/script-DeleteConext.yml
dorsha: Yes, replaced correctly

Scripts/script-DeleteConext.yml
dorsha: Don't want to expose for old versions

PR #1992, opened 2018-08-21
IntSights integration: Add possibility to close alert

pingunaut:
Add the possibility to close an IntSights alert


pingunaut:
Sorry...needs to be reworked

PR #1993, opened 2018-08-21
IntSights Integration: add possibility to close alert

pingunaut:
IntSights Integration: add possibility to close alert


anara123:
@pingunaut Thank you for the contribution

PR #1995, opened 2018-08-21
IsEmailAddressInternal script context fix

bakatzir:
fixes: https://github.com/demisto/etc/issues/12799

handled context for script to update and not append values


bakatzir:
added test playbook for the script.

PR #1997, opened 2018-08-21
Added threat-grid-get-samples-state command output

BenJoParadise:
Related: https://github.com/demisto/etc/issues/8989

The ThreatGrid.Sample.Status output was not documented in the command outputs.


BenJoParadise:
It was a playbook that tested the detonate-file command, which no longer exists because we moved to using the polling playbook

BenJoParadise:
@yaakovi @anara123 any other comments? can we merge?

PR #2000, opened 2018-08-21
added Endpoint in context and outputs for cs-device-details command.

bakatzir:
fixes - https://github.com/demisto/etc/issues/10369


bakatzir:
Added a test and opened a documentation issue,
will update the outputs as @kirbles19 will say.

PR #2001, opened 2018-08-22
Account enrichment playbook

bakatzir:
fixes: https://github.com/demisto/etc/issues/12851


asafshen:
@bakatzir can you please merge from master?

bakatzir:
Removed the input text that uses filters in the old file, left it in the new playbook.
Please check also that the from/to version is ok

bakatzir:
I rephrased the releaseNotes.

PR #2004, opened 2018-08-22
Common server python

yardensade:
added dbotscore to output paths on common server python script


yaakovi:
@bakatzir 
the same vendor can have multiple scores, for example:
email scan that contains a url in the body and has an attachment will result with 2 DBotScores.
ATD is enabled, it should scan the file and the URL and add DBotScore for each of them.
if we filter by vendor alone they will override each other.

yardensade:
@bakatzir  ^ Exactly
If you only use vendor there is no point using dbotscore really,
the indicator is the main part here to get the dbotscore, atleast in the integrations I've worked on.

PR #2006, opened 2018-08-22
Return vt from skipped

bakatzir:
After Merging, Do not close PR and delete branch until we check in nightly that we don't exceed the quota.


bakatzir:
opening a new PR

PR #2007, opened 2018-08-22
Added ews-get-folder command

DeanArbel:
solves: https://github.com/demisto/etc/issues/12964


Integrations/integration-EWSv2.yml
```diff
@@ -1162,6 +1166,12 @@ script:
             }
         }
 
+    def get_folder(folder_path, target_mailbox=None):
+        account = get_account(target_mailbox or ACCOUNT_EMAIL)
+        folder = get_folder_by_path(account, folder_path)
+        folderDict = parse_item_as_dict(folder, target_mailbox)
+
+        return get_entry_for_object("Folder %s" % (folder_path,), CONTEXT_UPDATE_FOLDER, folderDict)
```
BenJoParadise: The context output is inconsistent with the context output of other folder-related commands.
For example, ews-find-folders.

This is the current implementation, with an example output of ews-find-folders as well:
playground - war room 2018-08-26 14-10-58

This is the expected context output of the ews-get-folder command:
playground - war room 2018-08-26 14-14-19

PR #2010, opened 2018-08-23
Update script-MatchRegex.yml (#1791)

BenJoParadise:
* Update script-MatchRegex.yml

The previous version only returns the first match (even using the 'g' flag), so some looping was required. I also added an output.

* Update script-MatchRegex.yml

Fixed version and ID

* Update script-MatchRegex.yml

Fixed some minor code-review comments.

* Update script-MatchRegex.yml

Fixed adding 'val' to 'vals' and removed redundant 'setContext'

* Update script-MatchRegex.yml

* Fixed Infinite Looping
* Added 'g' as default flag
* Added conditional to return a single string if only one match is found, otherwise a list of matches (this is to mirror the original MatchRegex functionality)

* Update script-MatchRegex.yml


anara123:
@BenJoParadise  what is the status here?

PR #2014, opened 2018-08-23
Fix Extract Indicators from file bug

JonathanMeler:
No condition was set for else


JonathanMeler:
fixes https://github.com/demisto/etc/issues/12553

PR #2027, opened 2018-08-26
Fixed 'in-cap-upload-public-key' entry, it was missing in the JS URL …

npanone:
…Dictionary

Reorganized the urlDict and commands to match and correspond to the Incapsula API Documentation layout
Added Account Management API Calls
Added Site Management - Rules API Calls
Added Site Management - Data Centers API Calls
Added Infrastructure Protection Test Alert API Calls


anara123:
@npanone Looks good, thank you for the contribution

PR #2036, opened 2018-08-27
use new content aws spot instances

Bars92:
- pass server config from other repo.
- use spot instance aws to invoke aws instances.

saved about 15-20 minutes for now. Should maybe try higher performance machines, after that should be faster


.circleci/config.yml
```diff
@@ -26,6 +26,9 @@ jobs:
               sudo npm install -g jsdoc-to-markdown
               sudo pip install git+https://github.com/rahulrrixe/parinx.git@master
               chmod +x ./Tests/scripts/*
+      - add_ssh_keys:
+          fingerprints:
+              - "02:df:a5:6a:53:9a:f5:5d:bd:a6:fc:b2:db:9b:c9:47"
```
dorsha: Why not taking from conf? we should not have this here, remove this since we are in public repo, I think you should also delete the commit history....

Tests/scripts/instance.json
```diff
@@ -0,0 +1,18 @@
...
+      "Ebs": {
+        "DeleteOnTermination": true,
+        "VolumeType": "gp2",
+        "VolumeSize": 100,
+        "SnapshotId": "snap-0bc9018122add8925"
+      }
+    }
+  ],
+  "UserData": "ZmlsZTovL1Rlc3RzL3NjcmlwdHMvc2h1dGRvd25faW5zdGFuY2Uuc2g=",
+  "SecurityGroupIds": ["sg-ad9b98d1"]
```
dorsha: I think we have here sensitive data as well.. regions, etc.....

.circleci/config.yml
Bars92: there is no other way.. this step doesn't work with variables and its just a fingerprint for pem

Tests/scripts/instance.json
Bars92: we had it here already :/ it might be right for future tho. i can pull it from other repo after this

dorsha: I know we had it.. this is not good..

dorsha:
Approved. lets merge so other can work and fix these comments in another PR right away. 
Also, build fails.

Tests/scripts/instance.json
Bars92: ill do it tomorrow 👍

Bars92:
@dorsha only carbon black failed before

Tests/scripts/create_instance.sh
```diff
@@ -6,18 +6,36 @@ echo "Start create_instance script"
...
+
+if [ -z "$REQUEST_ID" ]
+then
+    echo "Failed setting up request for spot-instance."
+    exit 1
+fi
+
+MACHINE_STATE=""
+while [ "$MACHINE_STATE" != "fulfilled" ] ; do
+    echo "Waiting for machine to be ready ($REQUEST_ID)."
```
asafshen: maybe add an timeout/limit for this while loop

PR #2037, opened 2018-08-27
Creating whitelisted indicator programmatically using demisto REST API

gshriki:
DemistoAddWhitelist to allow programmatically adding indicators to the whitelisting list


gshriki:
OK. so closing this for now.

PR #2042, opened 2018-08-28
Add PagerDuty Script test

ronykoz:
Fix https://github.com/demisto/etc/issues/13114


ronykoz:
There are no users defined on the CircleCI instance so we can't validate the scripts of PagerDuty

PR #2044, opened 2018-08-28
select sensor to quarantine dynamically from online sensors

noaco:
https://github.com/demisto/etc/issues/13070#event-1809390533


anara123:
@noaco make sure that the test passes. Because currently the test is running nightly. It means it will not run in regular build

noaco:
@anara123 Cb Response test was updated and should pass.
Live Response test is not fixed yet, Arian is getting the sensors back online for the test.

PR #2045, opened 2018-08-28
Get instance details from other repo + add max retries

Bars92:
added max tries for create instance.
added instance conf from other repo.


dorsha:
Don't you need to remove the conf from here?

Tests/scripts/create_instance.sh
```diff
@@ -20,11 +20,14 @@ then
 fi
 
 MACHINE_STATE=""
-while [ "$MACHINE_STATE" != "fulfilled" ] ; do
-    echo "Waiting for machine to be ready ($REQUEST_ID)."
+TRY_COUNT=1
+MAX_TRIES=10
+while [ "$MACHINE_STATE" != "fulfilled" ] && [[ $TRY_COUNT -le $MAX_TRIES ]]; do
+    echo "Waiting for machine to be ready ($REQUEST_ID). try # $TRY_COUNT"
+    sleep 10
```
asafshen: not super critical - but do you think we should wait first?
if usually the first time succeed - lets wait after.. its only 10 secs but it will happen every build

Bars92:
@dorsha, @asafshen fixed.

PR #2046, opened 2018-08-28
Created GRR test and updated property "Credentials" -> "Username"

DeanArbel:
Implements: https://github.com/demisto/etc/issues/7328

### DIFF
--- a/Integrations/integration-GRR.yml
+++ b/Integrations/integration-GRR.yml
@@ -17,7 +17,7 @@ configuration:
   defaultvalue: "8000"
   type: 0
   required: false
-- display: Credentials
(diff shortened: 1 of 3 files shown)


Integrations/integration-GRR.yml
```diff
@@ -17,7 +17,7 @@ configuration:
   defaultvalue: "8000"
   type: 0
   required: false
-- display: Credentials
+- display: Username
```
aviadl: @anara123 would this cause backward compatibility issues ?

anara123: Changing the display name will not break comptibility

PR #2052, opened 2018-08-30
Email parsing fix.

ronykoz:
fixes https://github.com/demisto/etc/issues/13090


Scripts/script-ParseEmailFiles.yml
```diff
@@ -557,7 +568,7 @@ script: |-
               'Subject': subj,
               'HTML': html,
               'Text': text,
-              'Headers': headers,
+              'Headers': str(header_dict.items()),
```
erezh31: removed the "str"
store it as  list of `{'name': header.name, 'value': header.vale}`

Scripts/script-ParseEmailFiles.yml
```diff
@@ -520,7 +521,17 @@ script: |-
           fileData = emlFile.read()
           if b64:
               fileData = b64decode(fileData)
-          headers = fileData.split('\r\n\r\n')[0]
+
+          parser = HeaderParser()
+          headers = parser.parsestr(fileData)
+
+          header_dict = {
```
erezh31: no need

Playbooks/playbook-Process_Email_-_Generic.yml
```diff
@@ -319,12 +328,117 @@ tasks:
...
+              regex:
+                value:
+                  simple: .+\('Message-ID', '<
+              replaceWith: {}
       emailreceived: {}
-      emailreplyto: {}
-      emailreturnpath: {}
-      emailsenderip: {}
-      emailsize: {}
+      emailreplyto:
```
erezh31: On all those field, use Demisto filters and transformers to extract the fields. see EWSv2 mapping (you can just copy the logic)

erezh31:
you have email process generic playbook test that failed
plz fix

erezh31:
please verify the playbook is still working when the extra custom fields on email does not exist in the env

PR #2055, opened 2018-08-30
Adding DemistoDeleteIncidents command

ronykoz:
Added a command to delete multiple incidents using one command.

### DIFF
--- a/Scripts/script-DemistoDeleteIncidents.yml
+++ b/Scripts/script-DemistoDeleteIncidents.yml
@@ -1,0 +1,38 @@
+commonfields:
+  id: DemistoDeleteIncidents
+  version: -1
+name: DemistoDeleteIncidents
+script: |-
+  var body = {
+      ids: args.ids.toString().split(","),
+      all: false,
+      filter: {}
+  };
+
+  var res = executeCommand('demisto-api-post', {uri: '/incident/batchDelete', body: body});
+  if (isError(res[0])) {
+      throw res[0].Contents;
+  }
+
+  var response = res[0].Contents.response;
+  var md = tableToMarkdown('Demisto delete incidents', response);
+
+  return {
+      ContentsFormat: formats.json,
+      Type: entryTypes.note,
+      Contents: response,
+      HumanReadable: md
+  };
+type: javascript
+tags:
+- DemistoAPI
+comment: Delete incidents from Demisto (note action is irreversible)
+enabled: true
+args:
+- name: ids
+  required: true
+  default: true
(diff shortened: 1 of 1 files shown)


anara123:
WE should move both commands to rest api integration. And deprecate the scripts

ronykoz:
For some reason it took a different commit message

ronykoz:
@anara123 added those new commands and deprecated the old script

Integrations/integration-DemistoRESTAPI.yml
```diff
@@ -119,6 +119,34 @@ script:
...
     };
 
+    var deleteIncident = function(ids_to_delete) {
+        var body = {
+            ids: ids_to_delete,
+            all: false,
+            filter: {}
+        };
+
+        var args = {
```
anara123: no need to define args and then pass args.uri and args.body, you can just pass the uri as a string and the body already you have defined a variable above

Integrations/integration-DemistoRESTAPI.yml
```diff
@@ -148,6 +176,12 @@ script:
             }
             var desc = args.description || '';
             return ({Type: entryTypes.file, FileID: res.Path, File: filename, Contents: desc});
+        case 'demisto-delete-incident':
+            var ids = [args.id];
+            return deleteIncident(ids);
+        case 'demisto-delete-incidents':
+            var ids = args.ids.toString().split(",");
```
anara123: instead use `argToList(args.ids)`

Integrations/integration-DemistoRESTAPI.yml
anara123: and it is a bad practice to shadow a variable. `args` is a global variable of demisto

ronykoz:
@anara123 could you review this?

yaakovi:
why do we need two seperate commands?
The "delete incident**s**" supports single incident as well as multiple.

ronykoz:
@yaakovi You are right, my bad, removed the command for deleting only one incident

Integrations/integration-DemistoRESTAPI.yml
```diff
@@ -119,6 +119,29 @@ script:
         }
     };
 
+    var deleteIncident = function(ids_to_delete) {
```
yaakovi: the function deletes multiple incidents so the name should be plural: `deleteIncidents`

ronykoz: Fixed

PR #2056, opened 2018-08-30
Remove HTML mapping from default phishing mapping

JonathanMeler:
Removed from:
* EWSv2
* Mail-listener

Didn't remove from:
* Gmail - `body` field is mapped to `Email Body`, field that's not of type `html` and is not rendered (links aren't clickable)

fixes https://github.com/demisto/etc/issues/12327

### DIFF
--- a/Classifiers/classifier-EWS_v2.json
+++ b/Classifiers/classifier-EWS_v2.json
@@ -1,4 +1,5 @@
 {
+	"releaseNotes": "-",
 	"id": "c9332555-348b-4d91-85b4-e345cdba12c8",
 	"version": -1,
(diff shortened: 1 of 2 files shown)


Classifiers/classifier-EWS_v2.json
```diff
@@ -1,4 +1,5 @@
 {
+	"releaseNotes": "-",
```
aviadl: I think that this should be documented in RN

JonathanMeler: I'll let @kirbles19 know about it

aviadl: @JonathanMeler you simply need to document your change on this line ...

PR #2058, opened 2018-09-01
Delete Context keysToKeep enhancements

Itay4:
Fixes https://github.com/demisto/etc/issues/13050

### DIFF
--- a/Scripts/script-DeleteConext.yml
+++ b/Scripts/script-DeleteConext.yml
@@ -20,24 +20,39 @@ script: |
                       Type: entryTypes.error};
   }
   if (shouldDeleteAll) {
-      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.split(',') : [];
+      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.trim(' ').split(',') : [];
+      var keysToKeepObj = {};
+      var value;
+      for (var i = 0; i < keysToKeep.length; i++) {
+          value = dq(invContext, keysToKeep[i]);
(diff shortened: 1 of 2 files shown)


Scripts/script-DeleteConext.yml
```diff
@@ -20,24 +20,39 @@ script: |
                       Type: entryTypes.error};
   }
   if (shouldDeleteAll) {
-      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.split(',') : [];
+      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.trim(' ').split(',') : [];
```
dorsha: no need `(' ')` in trim, just `.trim()`

dorsha: need to do the trim after splitting, in case of:
`aaa , bbb,   ccc`
use map

Scripts/script-DeleteConext.yml
```diff
@@ -20,24 +20,39 @@ script: |
                       Type: entryTypes.error};
   }
   if (shouldDeleteAll) {
-      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.split(',') : [];
+      var keysToKeep = (args.keysToKeep) ? args.keysToKeep.split(',').map(function(item) { return item.trim(); }) : [];
```
anara123: You are using `.map`, then add `fromversion:3.5.0`

dorsha: why? map it is not es6..

PR #2059, opened 2018-09-02
Create script-AsciiEncode (#2057)

anara123:
Due to an issue in SplunkPy, I came across the need to encode some data into ascii before calling splunk-submit event. There weren't any automations around this, so I made one myself. I'm sure there will be other times in the future where someone could use this.


NeelRust:
@anara123 this script should be renamed as `encodeToASCII`

anara123:
@NeelRust ok, will change the name

PR #2075, opened 2018-09-03
Adjust integration logos according to standard

Itay4:
Fixes https://github.com/demisto/etc/issues/13193


dorsha:
Some of the logos are actually bigger, is that ok?

PR #2078, opened 2018-09-03
Threatconnect url command fix

yardensade:
fixes: https://github.com/demisto/etc/issues/13166

lowercase url input on url command to stop the request from failing


yardensade:
This won't be merged, a solution for a specific use-case.

PR #2081, opened 2018-09-03
Threat Grid detonation response (report) changed to infoFile type

DeanArbel:
relates to: https://github.com/demisto/etc/issues/8986


bakatzir:
@DeanArbel Please note that changing the entryType to 9 puts the data in the context but in order to be able to use it in playbooks you have to add it in the context as well.
You can look at ATD/Joe to see how we did it there

PR #2082, opened 2018-09-03
Adding phishing test

ronykoz:
fixes: https://github.com/demisto/etc/issues/8174


Scripts/script-CompleteManualTask.yml
```diff
@@ -0,0 +1,29 @@
+commonfields:
+  id: CompleteManualTask
+  version: -1
+name: CompleteManualTask
```
anara123: This looks like script that is relevant only for tests, then move it to be under `TestPlaybooks ` folder

ronykoz: Done

PR #2101, opened 2018-09-05
Deprecate whileloop scripts, should use native loops instead

bakatzir:
fixes: https://github.com/demisto/etc/issues/12939


Scripts/script-WhileNotMdLoop.yml
```diff
@@ -1,6 +1,6 @@
 commonfields:
-  id: 7b5c080e-f3b1-411a-83b0-e1f53c21bef8
-  version: 29
+  id: WhileNotMdLoop
```
dorsha: don't change these so it won't break stuff

Scripts/script-WhileNotExistLoop.yml
```diff
@@ -1,6 +1,6 @@
 commonfields:
-  id: f99a85a6-c572-4c3a-8afd-5b4ac539000a
-  version: 8
+  id: WhileNotExistLoop
```
dorsha: don't change these so it won't break stuff

PR #2109, opened 2018-09-06
Servicenow python

liorblob:


liorblob:
I ran the test playbook and it passed, although it's not in the build because the instance keeps hibernating. doc: https://github.com/demisto/etc/issues/12838

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+        path = 'table/' + ticket_type
+
+        return send_request(path, 'get', params = query_params)
+
+
+    def upload_file_command():
+        ticket_type = get_table_name(demisto.args().get('ticket_type', None))
+        ticket_id = demisto.args()['id']
+        file_id = demisto.args()['file_id']
+        file_name = demisto.args().get('file_name',
```
yaakovi: use `getFilePath` to get the name of the file: the function returns a dict with name, path as keys.

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+                                    demisto.dt(demisto.context(), "File(val.EntryID=='"+file_id+"').Name"))
+        file_name = file_name[0] if isinstance(file_name, list) else file_name
```
yaakovi: when will it be a list?

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+                return 'Cannot find computer'
+            computer = res['result'][0]
+        else:
+            computer = res['result']
+
+        if computer['u_code'] != computer_name:
+            return 'Computer not found'
+
+        hr = {
+            'Id': computer['sys_id'],
```
yaakovi: should be ID

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+        if not res or 'result' not in res:
+            return 'Cannot find groups'
+
+        hr_groups = []
+        context_groups = []
+
+        for group in res['result']:
+            if group['name'] == group_name:
+                hr_groups.append({
+                    'Id': group['sys_id'],
```
yaakovi: should be ID

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+            'sysparm_query': 'name=' + group_name
+        }
+
+        return send_request(path, 'get', params=query_params)
+
+    def fetch_incidents():
+        query_params = {}
+        incidents = []
+        sysparm_query = demisto.params().get('sysparm_query', None)
+        sysparm_limit = demisto.params().get('fetch_limit', DEFAULTS['limit'])
```
yaakovi: should be DEFAULTS['fetch_limit']

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+        query_params['sysparm_limit'] = sysparm_limit
+        if sysparm_query:
+            query_params['sysparm_query'] += '^' + sysparm_query
+
+        path = 'table/' + ticket_type
+
+        res = send_request(path, 'get', params = query_params)
+
+        for result in res['result']:
+            incident = result
```
yaakovi: why?

Integrations/integration-ServiceNowPython.yml
```diff
@@ -0,0 +1,1728 @@
...
+  required: false
+- display: Default ticket type of commands. Can be incident, sc_request, problem or
+    change_reuquest. Commands also take ticket_type argument
+  name: ticket_type
+  defaultvalue: incident
+  type: 0
+  required: false
+- display: ServiceNow API Version (e.g. 'v1')
+  name: api_version
+  defaultvalue: ""
```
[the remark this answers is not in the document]
liorblob: The version isn't necessary, blank can be a default value

Integrations/integration-ServiceNowPython.yml
liorblob: If there's more than one file in the context

Integrations/integration-ServiceNowPython.yml
liorblob: backward compatibility, can add ID in addition

Integrations/integration-ServiceNowPython.yml
liorblob: backward compatibility, can add ID in addition

Integrations/integration-ServiceNowPython.yml
liorblob: The file can be given as an argument, not necessarily from context

PR #2111, opened 2018-09-06
EWSv2: Fixed RateLimitError being thrown while fetching incidents.

BenJoParadise:
Fixed a case where RateLimitError would not be caught when fetching incidents.


anara123:
@BenJoParadise is this still relevant?

PR #2114, opened 2018-09-06
Fix find similar incidents

erezh31:
fixes https://github.com/demisto/etc/issues/13141


anara123:
@aviadl please review

PR #2120, opened 2018-09-10
Wrong context data when executing VT 'file' command with multiple hashes

sagylell:
When sending multiple hashes to the 'file' command, the response of the context was written always in the first entry.


Itay4:
@sagylell VT test is on nightly, can you please make sure it passes locally ?

PR #2126, opened 2018-09-12
increase limit for server build search

yaakovi:


yaakovi:
moved jask to skip, will return after the instance fix

Tests/conf.json
```diff
@@ -624,6 +620,10 @@
             "integrations": "AlphaSOC Wisdom",
             "playbookID": "AlphaSOC-Wisdom-Test"
         },
+        {
+            "integrations": "Jask",
```
asafshen: why move to skipped?

PR #2131, opened 2018-09-16
Expanded error messages for McAfee ePO

DeanArbel:
fixes: https://github.com/demisto/etc/issues/13028


DeanArbel:
For example:
image

PR #2133, opened 2018-09-16
Instance shutdown fix for aws circle

Bars92:
https://github.com/demisto/content-test-conf/pull/161


Tests/scripts/shutdown_instance.sh
```diff
@@ -1,3 +0,0 @@
-#!/bin/bash
-#in case circle didn't terminate instance for some reason
-shutdown +150
```
meirwah: why are we removing this @Bars92 ?

Bars92: This is embedded in the https://github.com/demisto/content-test-conf/blob/master/instance.json file under user data instead of a script

PR #2138, opened 2018-09-17
Table to markdown python string array

yaakovi:
continue of https://github.com/demisto/content/pull/1659


TestPlaybooks/script-TestPyCommonServer.yml
```diff
@@ -225,6 +225,35 @@ script: |-
...
+
+
+      table_string_array = tableToMarkdown('tableToMarkdown test with string array', ['foo', 'bar', 'katz'], ['header_1'])
+      expected_string_array_tbl = '''### tableToMarkdown test with string array
+  |header_1|
+  |-|
+  |foo|
+  |bar|
+  |kat|
+  '''
```
liorblob: should we add test for simple values + simple value header?

PR #2142, opened 2018-09-18
fix to fetch incidents pagination issue

DeanArbel:
relats to: https://github.com/demisto/etc/issues/13315


dorsha:
@DeanArbel I already created PR for reverting the revert:
https://github.com/demisto/content/pull/2141
You can close this and use the other one

PR #2144, opened 2018-09-19
ADGetUser - Output formatting 

LeonBirk:
UserAccountControl Flags:
Instead of creating a row for each possible flag and set true/false values, merge the flags that are actually set in one row.

image
image

instead of
image
image

### DIFF
--- a/Scripts/script-ADGetUser.yml
+++ b/Scripts/script-ADGetUser.yml
@@ -92,7 +92,7 @@ script: |-
               if userAccountControlOut:
                   accountControlFlags = ""
                   for flag, mask in UserAccountControlFlags.iteritems():
-                      if simulated_flag & mask != 0:
+                      if int(m['Contents'][0]['UserAccountControl'],10) & mask != 0:
                           accountControlFlags += flag + ", "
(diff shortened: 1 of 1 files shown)


Scripts/script-ADGetUser.yml
```diff
@@ -90,8 +90,12 @@ script: |-
               if 'UserAccountControl' not in m['Contents'][0] or m['Contents'][0]['UserAccountControl'] == "":
                   continue
               if userAccountControlOut:
+                  accountControlFlags = ""
```
dorsha: I would do something like this instead of string manipulation:

```
              if userAccountControlOut:
                  accountControlFlags = []
                  for flag, mask in UserAccountControlFlags.iteritems():
                      if int(m['Contents'][0]['UserAccountControl'],10) & mask != 0:
                          accountControlFlags.append(flag)
                  m['Contents'][0]['UserAccountControl'] = ', '.join(accountControlFlags)
```

LeonBirk: @dorsha yeah looks better, thanks!

PR #2152, opened 2018-09-20
SNDBOX plugin

tomerf-sndbox:


tomerf-sndbox:
Hi Shai,

In order to download please use the private user we provided you, as the
public user doesn't support downloads.

PR #2153, opened 2018-09-20
Upgrade to Venafi Integration - Typos and new 'venafi-get-details' command

cr4zyg047:
Added a new command which queries the Venafi API with the guid alone, no other arguments, to obtain the full metadata within the cert store. Command drops the contents to a simple markdown table and writes contents to the Venafi.**Certificates**.Details context point.

[Breaking change] Also fixed instances of the typo '**Certificats**' which will break users who rely on the output from this command.

### DIFF
--- a/Integrations/integration-Venafi.yml
+++ b/Integrations/integration-Venafi.yml
@@ -216,4 +216,3 @@ script:
     - name: guid
       required: true
     description: Use a certificate guid to extract more details from the cert store.
-  runonce: false
--- a/Integrations/integration-Venafi.yml
+++ b/Integrations/integration-Venafi.yml
@@ -83,8 +83,19 @@ script:
                 Contents: raw,
                 ContentsFormat: formats.json,
                 ReadableContentsFormat: formats.markdown,
-                EntryContext: {'Venafi.Certificats(val.ID==obj.ID)': certs},
-                HumanReadable: tableToMarkdown('Venafi certificats query response', certs),
+                EntryContext: {'Venafi.Certificates(val.ID==obj.ID)': certs},
+                HumanReadable: tableToMarkdown('Venafi certificates query response', certs),
+            };
+            return entry;
+        case 'venafi-get-details':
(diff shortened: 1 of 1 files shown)


Itay4:
Thanks for contributing to Demisto!

I agree that this change is required, but if we do it the way you suggested, we will break backward compatibility for user who use the current context path.

In order to handle that, we should output to both paths, the old one with the typo and the one you suggested.

Also, note that there is difference between the context paths of the command you modified you added - one uses `Certificates` while the other uses `Certificate`

In order to get them under the same path in the context, they should be the same - I suggest using `Certificate`

Please see my comments in the code as well.

Integrations/integration-Venafi.yml
```diff
@@ -83,8 +83,19 @@ script:
...
+            };
+            return entry;
+        case 'venafi-get-details':
+            var res = sendRequest('GET', '/vedsdk/certificates/' + args.guid, undefined, auth.APIKey);
+            entry = {
+                Type: entryTypes.note,
+                Contents: raw,
+                ContentsFormat: formats.json,
+                ReadableContentsFormat: formats.markdown,
+                EntryContext: {'Venafi.Certificate.Details(val.ID==obj.ID)': res},
```
Itay4: This will use the `ID` key in the `Details` object as a unique field, while you want the `ID` key from the `Certificate` object to be the unique field.

Integrations/integration-Venafi.yml
```diff
@@ -185,18 +196,23 @@ script:
...
-    - contextPath: Venafi.Certificats.SchemaClass
+    - contextPath: Venafi.Certificates.SchemaClass
       description: Certificate schema
-    - contextPath: Venafi.Certificats.ID
+    - contextPath: Venafi.Certificates.ID
       description: Certificate ID (GUID)
     description: Get Venafi certificates query. All dates are in the 2016-11-12T00:00:00.0000000Z
       format. Additional fields can be used in the query by adding them in a key=value
       manner
+  - name: venafi-get-details
```
Itay4: Please add the command outputs.

PR #2159, opened 2018-09-25
Added command aws-s3-upload-file.

tsymbalov:
Fix: https://github.com/demisto/etc/issues/13109
Added command aws-s3-upload-file


tsymbalov:
@anara123 
Docs updated.

PR #2166, opened 2018-09-26
add depends on (must or should) for scripts that lacked it

bakatzir:


bakatzir:
reference: https://github.com/demisto/etc/issues/13027

PR #2178, opened 2018-10-02
update epo-command and test PB

bakatzir:
fixes: https://github.com/demisto/etc/issues/13371
@nickdemisto - I thought that instead of adding a script I can add the same functionality in the integration command itself.
I've checked it on my env and it worked, and it's also going to be tested in the test playbook.
PLEASE HAVE A LOOK AND LET ME KNOW WHAT YOU THINK.

I've already updated the documentation accordingly. reference: https://github.com/demisto/etc/issues/13060

### DIFF
--- a/Integrations/integration-epo.yml
+++ b/Integrations/integration-epo.yml
@@ -252,7 +252,7 @@ script:
               return result.objectId;
             }
              // If reached here then the response is an empty "OK:", which means VSEContentUpdateDemisto was not found in the server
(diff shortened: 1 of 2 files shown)


nickdemisto:
Bar, so far, everything looks great.  I will download and review to confirm.
Thank you!
Nick

bakatzir:
@nickdemisto 
reverting this PR due to https://github.com/demisto/etc/issues/13637
will hand out a solution in a new command shortly.

PR #2181, opened 2018-10-02
Adding bashrc files

ronykoz:


Utils/.server-rc
```diff
@@ -0,0 +1,48 @@
...
+    export GOPATH=$HOME/dev/go
+}
+
+function setup_server_repo() {
+    if [[ "$1" == "--help" ]]; then
+      echo "Usage: setup_server_repo <Server path in git>"
+    else
+      mkdir $HOME/dev/go/src
+      mkdir $HOME/dev/go/src/github.com
+      mkdir $HOME/dev/go/src/github.com/demisto
```
DeanArbel: It's a shame mkdir -p is not safe, otherwise you could do this in one line

PR #2184, opened 2018-10-03
Returning Tests from skip

ronykoz:


DeanArbel:
relates to: https://github.com/demisto/etc/issues/13435

yardensade:
RTIR -> for unknown reason only on tests run same command twice which leads to failure in other command due to context changes.
Fix -> always sample the first command's context so the duplicate won't cause error.
Could not duplicate the situation on local env via playbook or via running the command.

yardensade:
https://github.com/demisto/etc/issues/13435

PR #2199, opened 2018-10-04
Clean skipped tests

noaco:


DeanArbel:
relates to: https://github.com/demisto/etc/issues/13435

PR #2204, opened 2018-10-04
Add slack notifications

ronykoz:
fixes https://github.com/demisto/etc/issues/13494


ronykoz:
Added a failed test section, testing it now, then need to update the channel to which I'm posting the message

ronykoz:
@anara123 could you review this?

PR #2210, opened 2018-10-07
FalconHost - new commands: cs-detection-search, cs-detection-details

bakatzir:
fixes: https://github.com/demisto/etc/issues/13493


bakatzir:
@josh-demisto 
Edited the HR, added all outputs in the context and prettified them - please let me know if it looks ok.
also added the first_behavior argument.

bakatzir:
in 3rd commit - CR from @anara123 changes:
rename output hierarchy, handle duplicates, throw->type.error, arrtomd->tabletomarkdown
add test to test PB

PR #2211, opened 2018-10-07
Adding SignalSciences integration

ronykoz:
Fixes https://github.com/demisto/etc/issues/12784


ronykoz:
Still have to check couple of minor points with @prasens33 but pretty much done

PR #2212, opened 2018-10-07
Cisco Umbrella context changes due to context override.

yardensade:
Do not merge yet, pending review.


Integrations/integration-Cisco-umbrella.yml
```diff
@@ -1,11 +1,11 @@
 commonfields:
-  id: Cisco Umbrella Investigate
+  id: Cisco Umbrella Investigate - Python
```
ronykoz: Why Python?

yardensade: To differentiate from the old integration (JS) that might still be active for clients.

Itay4: Why don't we replace the old one?

aviadl: We should make sure to be fully backward compatible, and then we should replace the old integration all together

yardensade: Honestly the only reason NOT to, is due to the old integration outputting several keys to context that are simply wrong, if someone uses/used it AS IS, it might not work for him.
Example: Domain.FirstSeean I fixed to Domain.FirstSeen (I couldn't not keep it with a mistake for the new integration)

yardensade: It is fully backward compatible, I will simply "duplicate" the old keys that are wrong to be supported but not declared in the settings output.

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+    ''' GLOBALS/PARAMS '''
+
+    API_TOKEN = demisto.params()['APIToken']
+    BASE_URL = demisto.params()['baseURL']
+    USE_SSL = True if demisto.params().get('insecure') else False
+    DEFAULT_HEADERS = {
+        'Authorization': 'Bearer {}'.format(API_TOKEN),
+        'Accept': 'application/json'
+    }
+    # remove proxy if not set to true in params
```
ronykoz: Why shouldn't we move all the proxy related stuff to CommonServer?

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
-
-        //find & remove port number
-        hostname = hostname.split(':')[0];
-
-        return hostname;
-    }
+    ''' HELPER FUNCTIONS '''
+
+    def extract_domain_name(url):
+        # Remove protocol, removes directory after the host name, returns only the host name.
```
ronykoz: No need for this comment, if you want just use function documentation

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+
+    def http_request( api_endpoint, params_dict=None, method='GET',data_list=None):
+        req_params = {}
+        # request does not accept lists, only string/dict
+        if data_list and isinstance(data_list, list):
+            data_list = json.dumps(data_list)
+        if params_dict:
+            req_params.update(params_dict)
+        url = BASE_URL + api_endpoint
+        LOG('running %s request with url=%s\theaders=%s\nparams=%s\tdata=%s' % (method, url, DEFAULT_HEADERS, json.dumps(req_params), data_list))
```
ronykoz: Remove DEFAULT_HEADERS it contains the api token of the user

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
-        return hostname;
-    }
+    ''' HELPER FUNCTIONS '''
+
+    def extract_domain_name(url):
+        # Remove protocol, removes directory after the host name, returns only the host name.
+        return url.split("//")[-1].split("/")[0]
+
+
+    def http_request( api_endpoint, params_dict=None, method='GET',data_list=None):
```
ronykoz: why don't we use this function in commonServer as well? all integrations seem to have it

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+
+            res.raise_for_status()
+            return res.json()
+
+        except Exception, e:
+            LOG(e)
+            raise
+
+
+    def string_to_table_header(string):
```
ronykoz: Change name to, `format_string_to_table_header_format`

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        # Checks if string, change underscore to space, capitalize every word. example: "one_two" to "One Two" (title() could work too)
+        if type(string) in STRING_TYPES:
+            return " ".join(word.capitalize() for word in string.replace("_"," ").split())
+        else:
+            return_error('The key is not a string: {}'.format(string))
+
+
+    def string_to_context_key(string):
```
ronykoz: Change name to `format_string_to_context_key_format`

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        # Checks if string, removes underscores, capitalize every word. example: "one_two" to "OneTwo" (title() could work too)
+        if type(string) in STRING_TYPES:
+            return "".join(word.capitalize() for word in string.split('_'))
+        else:
+            return_error('The key is not a string: {}'.format(string))
+
+
+    def date_to_ts(date):
```
ronykoz: date_to_timestamp

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+            if date[19]=='+':
+                ts+=timedelta(hours=int(date[20:22]),minutes=int(date[22:24]))
+            elif date[19]=='-':
+                ts-=timedelta(hours=int(date[20:22]),minutes=int(date[22:24]))
+            ts = time.mktime(ts.timetuple())
+            return str(int(ts) * 1000)
+        return date
+
+
+    def ts_to_date(ts):
```
ronykoz: timestamp_to_date

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        if ts:
+        # Gets a timestamp (either str or int, either in seconds or milliseconds) and converts it to a date.
+            ts = str(ts)
+            if len(ts) > 10:
+                ts = ts[:10]
+            ts = int(ts)
+            return datetime.utcfromtimestamp(ts).strftime('%Y-%m-%dT%H:%M:%S')
+        return ts
```
ronykoz: Why are you returning ts?

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        elif sr == -100:
+            DBotScore = 3
+        return DBotScore
+
+
+    ''' INTERNAL FUNCTIONS '''
+
+    def get_co_occurences(domain):
+        # Build & Send request
+        endpoint_url = '/recommendations/name/' + domain + '.json'
```
ronykoz: I think it would be better to make a CONST out of it, use `'/recommendations/name/{0}.json'`
and then here you can use `.format(domain)`

Integrations/integration-Cisco-umbrella.yml
yardensade: @liorblob does not approve of it.

liorblob: @yardensade sure, make me the bad guy

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
 
-    // "one_two" to "OneTwo"
-    var keyToContextKey = function(s) {
-        return s && keyToTableKey(s).replace(/\s+/g, '');
-    }
+            context[outputPaths['domain']] = domain_context
+
+        # Build Output
+        results.append ({
+            # type
```
ronykoz: remove those comments

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+            # context
+            'EntryContext': context
+        })
+        # Send table/context output to war-room
+        return results
+
+
+    def get_domain_categorization(domain):
+        # Build & Send request
+        endpoint_url = '/domains/categorization/' + domain + '?showLabels'
```
ronykoz: CONST, same as above

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+    def get_domain_search_command():
+        # Initialize
+        contents = []
+        context = {}
+        headers = []
+        results = []
+        # Get vars
+        regex = demisto.args()['regex']
+        start = demisto.args().get('start','')
+        # Fetch data
```
ronykoz: You can use blank lines to separate different code segments, that would be more readable this way

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown('Search Results:', contents, headers),
+            # context
+            'EntryContext': context
+        })
+        # Send table/context output to war-room
+        return results
+
+
+    def get_domain_search(regex, start):
```
ronykoz: What does it do?

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        res = http_request(endpoint_url)
+
+        # Validate and assign response
+        categorization = res.get(domain, [])
+        if not categorization:
+            return False
+        return categorization
+
+
+    def get_domain_search_command():
```
ronykoz: You probably can split this function so it will be more readable

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+        # Domain reputation + [whois -> whois nameservers -> whois emails] + domain categorization
+        results.append ({
+        # type
+        'Type': entryTypes['note'],
+            # contents
+            'ContentsFormat': formats['json'],
+            'Contents': [contents, whois, name_servers, emails, domain_categorization_table],
+            # human-readable
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown('Umbrella Investigate Domain Reputation for: ' + domain, contents, headers) +
```
ronykoz: You can do it outside of here(all the +++)

Integrations/integration-Cisco-umbrella.yml
```diff
@@ -31,495 +31,1834 @@ configuration:
...
+                }
+            else:
+                params = {
+                    'nameServerList': nameservers_string,
+                }
 
-    var getDomainTags = function(domain) {
-        var url = baseURL + '/domains/' + encodeURIComponent(domain) + '/latest_tags';
-        var tags = sendRequest(url);
+        res = ht
```
ronykoz: Perhaps you should create a CONST list for each command and the use `demisto.command() in LIST`

Integrations/integration-Cisco-umbrella.yml
yardensade: In case ts is not parseable I still use it in my call, sometimes users might want to pass date time string like "2 days ago", its good for me.

Integrations/integration-Cisco-umbrella.yml
yardensade: You mean CONST for endpoints basically? not sure @anara123 would approve

Integrations/integration-Cisco-umbrella.yml
yardensade: Did similar thing in minemeld back then, @anara123 is not a fan of it.

Integrations/integration-Cisco-umbrella.yml
yardensade: we could/should

PR #2221, opened 2018-10-08
office365 compliance search - new script added

bakatzir:
fix: https://github.com/demisto/etc/issues/13278

printscreen of entry ex:
screen shot 2018-10-08 at 18 56 39


BenJoParadise:
Redesigning this using polling architecture.

PR #2228, opened 2018-10-09
Add Awake Security integration

Gabriel439:
This adds support for the following commands:

* `awake-lookup-{ip,domain,email,device}`
    * Retrieves details for the corresponding entity in our knowledge graph
* `awake-query-{domains,devices,activities}`
    * Allows Demisto to use our query API to retrieve and filter entities en
      masse
    * In addition to using our query language, these commands also provide
      Demisto arguments to filter by common predicates for ease of playbook
      integration
* `awake-pcap-download`
    * Allows Demisto to retrieve our PCAP data as a file
    * Supports filtering by host/port/timespan for ease of playbook integration
* `domain`, `ip`, and `email` and commands
    * These are to integrate with the common idiom of searching for certain
      entities across multiple installed integrations
    * These are thin wrappers around `awake-lookup-{domain,device}`
    * Note: We do not yet provide support for `!file` (i.e. retrieving matches
      to files by hash)
* Fetch incidents support
    * This is a preliminary wrapper around our threat behavior API to retrieve
      all matches to specified behaviors since the last time bucket
    * Note: The context provided to threat behavior matches is essentially empty
      due to limitations in our current public API, which can only detect the
      presence matches, but does not provide details about the matches.

We reserve the `AwakeSecurity` field of the `Context` for returning
machine-readable data and also export human readable output for a subset of
the data in markdown tables.

### DIFF
--- a/Integrations/integration-Awake_Security.yml
+++ b/Integrations/integration-Awake_Security.yml
@@ -1,0 +1,562 @@
+commonfields:
+  id: Awake Security
+  version: -1
+name: Awake Security
+display: Awake Security
+category: Network Security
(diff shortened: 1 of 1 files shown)


Gabriel439:
I've completed everything that I intended to add.  The only thing I'm not sure of is if I'm adding the Incident metadata for the `Query`/`StartTime`/`EndTime` in the correct place

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+        # Our product does not yet assign meaning
+        # to these scores since unusual behavior is
+        # not necessarily malicious, but empirically
+        # partioning the scores by quartile appears to
+        # map closely enough to Demisto's classification
+        # scheme for now.
+        if percentile <= 25:
+            # Typical range for people doing ordinary
+            # office work
+            score = 0
```
bakatzir: dbot dcore 0 means unknown

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+            "Type": 1,
+            "ContentsFormat": "json",
+            "Contents": json.dumps(machineReadable),
+            "HumanReadable": humanReadable,
+            "ReadableContentsFormat": "markdown",
+        }
+
+        demisto.results(results)
+
+    def toDBotScore(indicator_type, percentile, lookup_key):
```
bakatzir: maybe it's better instead of hardcoding the thresholds for the awake to dbot score, to add 2 params to the integration:
lower threshold
high threshold
we can give them any initial values we want.
and the dbotscoregiven can be according to them.
WDYT?

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+        contents = lookup("ip", lookup_key)
+
+        humanReadableFields = [
+            "deviceCount",
+            "activityCount",
+            "ipFirstSeen",
+            "ipLastSeen",
+        ]
+
+        # Note: No DBotScore for IP addresses as we do not score them.
```
bakatzir: we still want to indicate that we enriched the ip, even if we can't generate a score using awake, so here dbotscore 0 is a good option

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+            "application",
+            "numberSimilarDevices",
+            "numberSessions",
+            "firstSeen",
+            "lastSeen",
+            "duration",
+            "deviceId",
+        ]
+
+        if "notabilityPercentile" in contents:
```
bakatzir: i'll check it thrilly when we'll have access, but i think dbotscore 0 might be a better option here

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+            "intelSources",
+            "numAssociatedDevices",
+            "numAssociatedActivities",
+            "approxBytesTransferred",
+            "protocols",
+            "firstSeen",
+            "lastSeen",
+        ]
+
+        if "notability" in contents:
```
bakatzir: i'll check it thrilly when we'll have access, but i think dbotscore 0 might be a better option here

Integrations/integration-Awake_Security.yml
```diff
@@ -0,0 +1,607 @@
...
+            "activeIP",
+            "nSimilarDevices",
+            "ipCount",
+            "applicationCount",
+            "protocols",
+            "firstSeen",
+            "lastSeen",
+        ]
+
+        if "deviceScore" in contents:
```
bakatzir: i'll check it thrilly when we'll have access, but i think dbotscore 0 might be a better option here

PR #2229, opened 2018-10-09
Update Threat Grid for new API

saharm1:


Integrations/integration-Threat_Grid.yml
```diff
@@ -1,7 +1,7 @@
 commonfields:
-  id: Threat Grid
+  id: Threat Grid New
```
Itay4: Why did you add `New` to `name` and `id` fields? As far as I understand this should keep backward compatibility.

DeanArbel:
Updated documentation: https://github.com/demisto/etc/issues/13749

DeanArbel:
documentation: https://github.com/demisto/etc/issues/13749

DeanArbel:
relates to: https://github.com/demisto/etc/issues/8989

PR #2232, opened 2018-10-10
Fixed proxy settings.

yardensade:


Integrations/integration-Cylance_Protect_v2.yml
```diff
@@ -1,8 +1,8 @@
 commonfields:
-  id: Cylance Protect v2
+  id: Cylance Protect v2 1
```
Itay4: You have some `1`'s here :)

yardensade:
Fixes:
https://github.com/demisto/etc/issues/13533

PR #2235, opened 2018-10-10
fix proxy configuration in FE integrations

noaco:


liorblob:
@noaco I forgot `del os.environ['https_proxy']`, please add

PR #2244, opened 2018-10-11
VirusTotal Private API - handle proxy

liorblob:


Integrations/integration-VirusTotal-Private_API.yml
```diff
@@ -61,6 +61,12 @@ script:
     # disable insecure warnings
     requests.packages.urllib3.disable_warnings()
 
+    if not demisto.params().get('proxy', False):
```
Itay4: Parameter name is `useProxy` and not `proxy`

liorblob: nice catch :)

PR #2247, opened 2018-10-11
Leave all investigations Script

gshriki:
Script to be executed by user to leave all active investigations - will clear out the incidents left pane


Nitzperetz:
@gshriki , a user could own the incident and not be part of the incident users.
If the aim is to clear the left pane, you need to make sure to remove all the incidents the user is part of the users field or the owner field.

Also, shouldn't the script use "incident" terminology rather than "investigation"?

gshriki:
I’m checking few things so will review that too. Another aspect I’m checking is restricted. I want to make sure restricted investigation is not left ‘orphan’ in this case.

PR #2252, opened 2018-10-15
mcafee atd - handle errors gracefully and detonate-url fix

bakatzir:
fixes: https://github.com/demisto/etc/issues/13566
fixes: https://github.com/demisto/etc/issues/13608


Integrations/integration-McAfee-ATD.yml
```diff
@@ -586,53 +582,59 @@ script:
         return uriSuffix;
 
     def detonate(submitType ,sample, timeout, report_type, threshold, fileName):
-        sys.exit(0)
-        result = file_upload(submitType, sample, fileName)
+        result = file_upload(submitType, sample, fileName, None, None, None, None, None, None, None, None)
```
liorblob: Why send all these `None` args?

bakatzir: it uses the same uploadfile command of the basic command that has many arguments, and are not needed for the detonate command.
not sending them results a python error.

anara123:
@bakatzir please deprecate all the detonate commands in ATD integration with `deprecated: true`

anara123:
also there is `integration-ATD.yml` and `integration-McAfee-ATD.yml`
`integration-ATD.yml` has no fromversion, toversion, please add

bakatzir:
@anara123  will do. there is still no full instance to work with.

bakatzir:
Outdated, closing this PR to avoid all the conflicts and opening a new one.

PR #2256, opened 2018-10-15
Exabeam BYOI in Python

slavikm:
@anara123 need to clean and finish. Most of the implementation is done. Test playbook should be added once we have the access. Can use a bit of refactoring into more functions.


bakatzir:
fix: https://github.com/demisto/etc/issues/6761

PR #2263, opened 2018-10-16
Update validate file structure 

ronykoz:
Updating validate file structure to validate only changed files


ronykoz:
@hirshbergshachar why should it break the git hook, I have changed the pre-commit hook to the new format, are there any other changes I need to do?

PR #2264, opened 2018-10-16
bugfix/move_urlscan_default_enabled_to_false

Spivaka:
fixes https://github.com/demisto/etc/issues/13476


Itay4:
Not sure I understand how this fixes the issue, is the way urlscan calculates DBotScore wrong?

Spivaka:
the issue was to remove it from being default enabled

anara123:
@Spivaka merge from master

PR #2268, opened 2018-10-16
Integration QRadar

DeanArbel:
relates to: https://github.com/demisto/etc/issues/13296


Integrations/integration-QRadar.yml
```diff
@@ -42,296 +38,554 @@ configuration:
...
+    import sys, os, requests, traceback, datetime, json
+    from requests.exceptions import HTTPError
+
+    ''' GLOBAL VARS '''
+    SERVER = demisto.params()['server'][:-1] if demisto.params()['server'].endswith('/') else demisto.params()['server']
+    USERNAME = demisto.params().get('credentials').get('identifier')
+    PASSWORD = demisto.params().get('credentials').get('password')
+    TOKEN = demisto.params().get('token')
+    USE_SSL = not demisto.params().get('insecure', False)
+    CUSTOM_HEADERS = demisto.args().get('headers')
```
anara123: arg shouldn't be global var.
Not all the commands has `headers` argument I suppose

Integrations/integration-QRadar.yml
```diff
@@ -42,296 +38,554 @@ configuration:
...
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    # disable insecure warnings
+    requests.packages.urllib3.disable_warnings()
+
+    # Header names transformation maps
+    # Format: {'OldName': 'NewName'}
+
+    OFFENSES_NAMES_MAP = {'follow_up': 'Followup', 'id': 'ID', 'description': 'Description', 'source_address': 'SourceAddress', 'destination_address': 'DestinatioAddress', 'start_time': 'StartTime', 'event_count': 'EventCount', 'magnitude': 'Magnitude', 'last_updated_time': 'LastUpdatedTime', 'offense_type': 'OffenseType'}
```
anara123: Format the json/dict `OFFENSES_NAMES_MAP` and `SINGLE_OFFENSE_NAMES_MAP`, all of them

anara123: between two variables add 1 or 2 spaces, this way it will be easier to read

Integrations/integration-QRadar.yml
```diff
@@ -42,296 +38,554 @@ configuration:
...
+    SEARCH_ID_NAMES_MAP = {'search_id': 'ID'}
+    ASSET_PROPERTIES_NAMES_MAP = {'Unified Name': 'Name', 'CVSS Collateral Damage Potential': 'AggregatedCVSSScore', 'Weight': 'Weight'}
+    ASSET_PROPERTIES_ENDPOINT_NAMES_MAP = {'Primary OS ID': 'OS'}
+    FULL_ASSET_PROPERTIES_NAMES_MAP = {'Compliance Notes': 'ComplianceNotes', 'Compliance Plan': 'CompliancePlan', 'CVSS Collateral Damage Potential': 'CollateralDamagePotential', 'Location': 'Location', 'Switch ID': 'SwitchID', 'Switch Port ID': 'SwitchPort', 'Group Name': 'GroupName', 'Vulnerabilities': 'Vulnerabilities'}
+    NOTE_NAMES_MAP = {'id': 'ID', 'note_text': 'Text', 'create_time': 'CreateTime', 'username': 'CreatedBy'}
```
anara123: it seems that you use it only in 2 commands, it would be better if you will define it inside the command itself, because those maps are command specific

Integrations/integration-QRadar.yml
```diff
@@ -42,296 +38,554 @@ configuration:
...
-        if (filter) {
-            urlArgs.filter = filter;
-        }
-        if (fields) {
-            urlArgs.fields = fields;
-        }
-        var headers = {'Content-Type': ['application/json']};
-        if (range) {
-            headers.Range = ['items=' + range];
+    def get_entry_for_assets(title, obj, human_readable_obj, headers=None):
```
anara123: if you create helper method for specific command then better write this function near that command

Integrations/integration-QRadar.yml
```diff
@@ -42,296 +38,554 @@ configuration:
...
+        return get_entry_for_object('QRadar Search Results', res, CUSTOM_HEADERS, 'QRadar.Search(val.ID === {0}).Result'.format(search_id))
+
+    def get_assets_command():
+        res = send_get_assets(demisto.args().get('range'), demisto.args().get('filter'), demisto.args().get('fields'))
+        res, human_readable_res = create_assets_result(res)
+        return get_entry_for_assets('QRadar Assets', res, human_readable_res, CUSTOM_HEADERS)
+
+    def get_asset_by_id_command():
+        _filter = "id=" + str(demisto.args().get('asset_id'))
+        res = send_get_assets(_filter=_filter)
```
anara123: you don't need to write `send_xxx`, we just do `get_assets`

anara123: and get_assets should usually return `assets`, so it should be 
`assets = get_assets(...)`
or 
`raw_assets = get_assets(...)`

Integrations/integration-QRadar.yml
DeanArbel: This will break the structure used for the script, where all the utilities funcs are together, but I will do so and also add comments to each method to make it clearer

DeanArbel:
documentation: https://github.com/demisto/etc/issues/13764

PR #2270, opened 2018-10-17
Create instance notifier

ronykoz:
fixes https://github.com/demisto/etc/issues/13635


ronykoz:
Do not merge yet, want to check that first

PR #2272, opened 2018-10-17
Better error handling in Archer

hirshbergshachar:


Integrations/integration-ArcherRSA.yml
```diff
@@ -353,8 +353,11 @@ script:
             rawResponse = result.Body;
             break;
         }
-        var response = JSON.parse(x2j(rawResponse));
-
+        try {
+            var response = JSON.parse(x2j(rawResponse));
+        } catch (err) {
+            throw "Problem with login, please make sure that you have the right credentials. Original server response: " + rawResponse + " Error was: " + err
```
anara123: return error entry instead of throwing error

hirshbergshachar: @anara123 doesn't fit here, this is an inner function.

PR #2282, opened 2018-10-19
Converting skip lists to dictionaries

ronykoz:
Adding an option to add skip reason to the skip integrations and tests

fixes https://github.com/demisto/etc/issues/13670


ronykoz:
I still have to give description/reason for each of the skips so If you know what is the reason for some of the skips please let me know so i can fill that in

anara123:
@ronykoz no need, lets keep it this way for now

PR #2288, opened 2018-10-21
fix FireEyeDetonateFile script

noaco:
- add arguments ti set analysis type and prefetch
- use conditional json loads

fixes: https://github.com/demisto/etc/issues/13365


Scripts/script-FireEyeDetonateFile.yml
```diff
@@ -43,10 +40,11 @@ script: |+
           resp = demisto.executeCommand('fe-submit', bArgs)
           if isError(resp[0]):
               demisto.results(resp)
-              sys.exit(0)
+              sys.exit(1)
```
anara123: sys.exit(0) exit with 0

PR #2290, opened 2018-10-21
Add outputs to mcafee atd detonate playbooks

bakatzir:
fix: https://github.com/demisto/etc/issues/13687

added outputs to the playbooks from the get-report command.
since we have no instance the tests were skipped.
but only addition of outputs - so to your judgement.


liorblob:
need to make sure it works, guess we're waiting for the instance...

Playbooks/playbook-Detonate_URL_-_McAfee_ATD.yml
```diff
@@ -1,4 +1,4 @@
-id: detonate_url_-_mcafee_atd
+id: Detonate URL - McAfee ATD
```
anara123: revert the id

PR #2291, opened 2018-10-21
modify authentication test for CB Live Response 

noaco:
fixes https://github.com/demisto/etc/issues/13275

### DIFF
--- a/Integrations/integration-Carbon_Black_Enterprise_Live_Response.yml
+++ b/Integrations/integration-Carbon_Black_Enterprise_Live_Response.yml
@@ -2654,3 +2654,6 @@ script:
       description: The destination path of the file.
     description: Save a file to a specific path on the endpoint.
   runonce: false
+releaseNotes: "fix the authentication test"
+tests:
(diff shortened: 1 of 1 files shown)


Integrations/integration-Carbon_Black_Enterprise_Live_Response.yml
```diff
@@ -206,6 +206,12 @@ script:
         if (res.StatusCode === 200) {
             return 'ok';
         }
+        // 405 - Method Not Allowed.
+        // This error can raise when using CB Defence; This URL (which is used to list sessions) is theoretically not available for CB Defence. In practice it works for some cases.
+        // 401 is the status code used for 'UNAUTHORIZED' error
+        if (CB_PRODUCT === 'Defense' || res.StatusCode === 405) {
```
anara123: I think the condition should `AND` `if (CB_PRODUCT === 'Defense' && res.StatusCode === 405)`

noaco: its an error. will fix it.

PR #2293, opened 2018-10-21
Support giving min vendor count for IsMaliciousIndicatorFound automation

meirwah:
fixes https://github.com/demisto/etc/issues/13627


Itay4:
@meirwah Do we still need that?

PR #2298, opened 2018-10-22
add EICAR.exe to test data repository

bakatzir:


anara123:
@bakatzir I think it is better to call it `EICAR` no?

PR #2301, opened 2018-10-22
FE iSIGHT return DBot score for indicator when no data found

noaco:
fixes https://github.com/demisto/etc/issues/13319

### DIFF
--- a/Integrations/integration-isight.yml
+++ b/Integrations/integration-isight.yml
@@ -165,7 +165,19 @@ script:
         var records = basicSearch('ip', ip);
 
         if (!records) {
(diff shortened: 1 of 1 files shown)


Integrations/integration-isight.yml
```diff
@@ -442,4 +478,6 @@ script:
       - malware
       - other
     description: Submission of malware and other files for community sharing
-hidden: false
+releaseNotes: "return DBot score for indicator when no data is found"
+tests:
+  - Forgive me for my sins but I did not create any test
```
anara123: I forgive you :)

PR #2305, opened 2018-10-23
Sumologic fix and enhancement

Itay4:
Fixes https://github.com/demisto/etc/issues/13497 https://github.com/demisto/etc/issues/13649

### DIFF
--- a/Integrations/integration-SumoLogic.yml
+++ b/Integrations/integration-SumoLogic.yml
@@ -278,3 +278,5 @@ script:
     description: Search SumoLogic for matching records to the given query
   isfetch: true
 releaseNotes: "Added support for equal sign in query and headers arguments to search command"
+tests:
+  - Forgive me for my sins but I did not create any test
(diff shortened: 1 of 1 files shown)


Integrations/integration-SumoLogic.yml
```diff
@@ -204,16 +204,28 @@ script:
...
+            if (httpIndex != -1) {
+                var httpSubstring = query.substring(httpIndex)
+                var whitespaceIndex = httpSubstring.indexOf(' ')
+                var url = httpSubstring.substring(0, whitespaceIndex)
+                url = url.replace(/=/g, '\\\\=')
+                var beforeHttp = query.substring(0, httpIndex)
+                var afterWhitespace = query.substring(httpIndex + whitespaceIndex)
+                query = beforeHttp + url + afterWhitespace
+            }
+            var headers = 'headers' in args ? argToList(args.headers) : null;
```
yaakovi: use `undefined` instead of `null`

PR #2308, opened 2018-10-23
deprecate CB Live Response older version commands

noaco:
The following will be deprecated:

- cb-command-create
- cb-command-create-and-wait

This is a 'generic command' to execute specific commands on the endpoind. We added explicit commands in the last integration update.
e.g.  _cb-command-create name="directory-list" session=[sessionID] object=[path]_ is now available as _cb-directory-listing path=[path] sensor=[sensorID]_

- cb-command-cancel

To cancel a 'pending' command. Not usable if 'cb-command-create' is not available.

### DIFF
--- a/Integrations/integration-Carbon_Black_Enterprise_Live_Response.yml
+++ b/Integrations/integration-Carbon_Black_Enterprise_Live_Response.yml
@@ -1682,6 +1682,7 @@ script:
     description: Archive the given session (If the session has no content it will
       fail)
   - name: cb-command-cancel
+    deprecated: true
     arguments:
     - name: session
       required: true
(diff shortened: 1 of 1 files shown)


anara123:
@noaco isn't there more commands that we want to deprecate in this integration? Like getting file from an endpoint?
And open an issue for this one and link it here

noaco:
@anara123 

The following file operation are already deprecated:
1. 'cb-file-upload' (uploads a file from cb server to the endpoint)
2. 'cb-file-get' (uploads a file from the endpoint to cb server)
Both will require to run an additional commands 'cb-command-create' and 'cb-file-info' to push/get a file from cb server from/to the war room.
Both were replaced with commands that can stand alone. 

The following commands are file operation on files found on cb server.
1. 'cb-file-info' returns information on a file in cb server
2. 'cb-delete-file' will delete a file from cb server 
3. 'cb-list-files' listing files on cb server for a given session

Not sure if we want to deprecate them. Maybe 'cb-file-info'.

PR #2309, opened 2018-10-23
fixed emailtocount and emailto fields

yaakovi:
fixes: https://github.com/demisto/etc/issues/13522

### DIFF
--- a/Playbooks/playbook-Process_Email_-_Generic.yml
+++ b/Playbooks/playbook-Process_Email_-_Generic.yml
@@ -810,3 +810,5 @@ outputs:
 - contextPath: File
   description: The File object
   type: unknown
+tests:
+  - Forgive me for my sins but I did not create any test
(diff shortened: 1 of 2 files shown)


Playbooks/playbook-Process_Email_-_Generic.yml
```diff
@@ -437,12 +437,17 @@ tasks:
...
+            args:
+              separator:
+                value:
+                  simple: ','
       emailtocount:
         complex:
           root: Email
           accessor: To
           transformers:
+          - operator: uniq
```
yaakovi: to avoid same address count as two recipients.

Scripts/script-ParseEmailFiles.yml
```diff
@@ -35,7 +35,6 @@ script: |-
   import olefile as OleFile
   from email.parser import HeaderParser
   from email.parser import Parser as EmailParser
-  import re
```
yaakovi: this code is taken from somewhere open-sources so it should not be changes (all code under license)

PR #2312, opened 2018-10-24
Gmail block mail

yaakovi:
fixes: https://github.com/demisto/etc/issues/11082


anara123:
@yaakovi merge after docs are ready

PR #2314, opened 2018-10-24
flattenCell Fix

compsecmonkey:
Adding additional logic in flattenCell to attempt to catch encoding issues

### DIFF
--- a/Integrations/integration-epo.yml
+++ b/Integrations/integration-epo.yml
@@ -389,92 +389,57 @@ script:
         }
     }
 
-    function parseCommandArgs(command, commandArgs) {
-        //commandArgs should be in the format of:  keyName1:keyValue1, keyName2:KeyValue2
-        if (!commandArgs) {
-            return null;
-        }
-        var commandArgsDict = {};
-        commandArgsDict["command"] = command;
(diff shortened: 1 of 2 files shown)


Itay4:
Thanks for contributing to Demisto!
I've tried running some special characters and they seem to be encoded as expected, could you please share the problematic characters ?

compsecmonkey:
I will PM you a sampling of the data; but its any unicode character that is out of ascii's range. Example of the stack trace when this happens:

`Error from Scripts is : Script failed to run: 
   File "/tmp/pyrunner/_script_docker_python_loop.py", line 451, in <module>
    exec(code, sub_globals, sub_globals)
   File "<string>", line 39, in <module>
   File "<string>", line 968, in flattenTable
   File "<string>", line 955, in flattenRow
   File "<string>", line 955, in <dictcomp>
   File "<string>", line 627, in formatCell
   File "<string>", line 647, in flattenCell
   File "<string>", line 647, in <genexpr>
 UnicodeEncodeError: 'ascii' codec can't encode characters in position 15-20: ordinal not in range(128)
] (2604) (2603)`

PR #2321, opened 2018-10-28
Falconhost fix offset

liorblob:


Integrations/integration-CrowdStrikeHost.yml
```diff
@@ -114,7 +114,7 @@ script:
     // add args to a if exists
     function add(a, k, f) {
         if (args[k]) {
-            var parts = argToList(args[k]);
+            var parts = argToList(args[k].toString());
```
anara123: This solution is not good enough. I think it only covers part of the problem.
There should be other arguments of type list. Pass list to some of the arguments and make sure it works properly

liorblob: @anara123 my thought was that arguments from CLI always pass as string, so if I always convert to string it should work normally. Anyway, here's a test I did: 
image

liorblob:
@anara123 handled number specifically

PR #2322, opened 2018-10-28
Fail a commit if added a test but not to conf.json

ronykoz:
fixes https://github.com/demisto/etc/issues/13787


anara123:
@ronykoz resolve conflicts

PR #2327, opened 2018-10-29
URLScan Enhancements

amshamah419:
1. Deprecated IP, File, and URL command since the integration does not accept them.
2. Formatted outputs to be easily accessible for customers
3. Migrated integration to Python


Integrations/integration-UrlScan.yml
```diff
@@ -16,496 +15,640 @@ configuration:
...
       - public
       - private
-      description: Does the submit will be public or private
+      description: Will the submission be public or private
       defaultValue: public
+    - name: limit
+      description: Limits the results in the war room
+      defaultValue: "20"
     outputs:
-    - contextPath: URL.Data
```
Itay4: Why did you remove all these outputs? Please add them back

Tests/conf.json
```diff
@@ -49,10 +49,7 @@
           "playbookID": "devo_test_playbook"
         },
         {
-            "integrations": [
-                "VirusTotal",
-                "urlscan.io"
-            ],
+          "integrations": "VirusTotal",
```
Itay4: Why are you using `VirusTotal` and no `urlscan` ?

amshamah419: Switched to urlscan.

PR #2338, opened 2018-10-30
Clear release notes - 18.10.3

Itay4:


asafshen:
@Itay4 I believe  that the git hash should change to latest on this branch
(build fails)

PR #2341, opened 2018-10-30
Fix separator line in tableToMarkdown function

gshriki:
Putting `---` for separator line as the standard requires


anara123:
```
TestCommonPython failed with error/s
Playbook TestCommonPython has failed:
	- Command: !TestPYCommonServer test_type="TBLMD" extend-context=""
	- Body: failure
	- Body: Playbook **TestCommonPython** execution error
```
@gshriki your test is failing

PR #2357, opened 2018-11-04
fix exceed virustotal api rate

jochman:
giving wait time between api calls to ensure not exceed api rate

fixes: https://github.com/demisto/etc/issues/13527


yardensade:
@anara123 @jochman 
This test right now is written poorly.
This is simply not going to prevent the test from failing again.

Easiest example, right now there are 5 calls for !url (with same domain even), each call demands a sleep of 60 seconds.
This means this test is going to run for well over 10 minutes, while in the meantime it might be ran from other masters/tests/local machines as well, easily reaching the limit of the service.

Is there any justification for creating so much calls from an already very fragile service? My personal feeling is not.

Also:

- Why is there a sleep after first delete context?
- Why is there a sleep and delete context after last command?
- Why is there a sleep after 2 first "set"s to context which are not related to the service?
- Why are we using so many verifyContext scripts instead of the faster transformers?

I would recommend:
- cleaning the test from unneeded sleeps.
- reducing the total amount of calls to the minimum possible
- since its free credentials I think we should use a specific set of creds for the sake of tests only.

PR #2361, opened 2018-11-04
InfoArmor Vigilante ATI

anara123:


anara123:
resolves https://github.com/demisto/etc/issues/13778

anara123:
docs https://github.com/demisto/etc/issues/14196

PR #2366, opened 2018-11-04
Flatten cell test

orenzohar:
Merge from https://github.com/demisto/content/pull/2314 to master


Scripts/script-CommonServerPython.yml
```diff
@@ -529,9 +529,16 @@ script: |-
       if isinstance(data, STRING_TYPES):
           return data
```
anara123: what about the case if the data is NOT list, but string with special character in it?

anara123:
@orenzohar your tests is failing in circle

Scripts/script-CommonServerPython.yml
orenzohar: In that case there's no need to flatten anything, and the calling functions seems to know how to handle unicode.

Itay4:
@orenzohar What is the status of this?

anara123:
```
------ Test playbook: TestCommonPython start ------
TestCommonPython failed with error/s
Playbook TestCommonPython has failed:
	- Command: !TestPYCommonServer test_type="XML"
	- Body: Error from Scripts is : Script failed to run: 
   
   File "<string>", line 307
     demisto.results(toEntry('special character failure - flatten_cell))
                                                                       ^
 SyntaxError: EOL while scanning string literal
] (2604) (2603)
	- Body: Playbook **TestCommonPython** execution error
```

PR #2370, opened 2018-11-05
vt-priavte-api - added quota error message

bakatzir:
fix: https://github.com/demisto/etc/issues/13878

### DIFF
--- a/Integrations/integration-VirusTotal-Private_API.yml
+++ b/Integrations/integration-VirusTotal-Private_API.yml
@@ -100,7 +100,12 @@ script:
             )
             res.raise_for_status()
 
-            return res.json()
+            if res.status_code == 200:
+                return res.json()
+            # 204 HTTP status code is returned when api rate limit has been exceeded
+            elif res.status_code == 204:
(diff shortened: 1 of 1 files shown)


bakatzir:
this is quite instructive in general. seems like they recommend to raise for status and then check statuses 200 and 204 separately.

screen shot 2018-11-05 at 13 33 41

Integrations/integration-VirusTotal-Private_API.yml
```diff
@@ -100,7 +100,12 @@ script:
             )
             res.raise_for_status()
 
-            return res.json()
+            if res.status_code == 200:
+                return res.json()
+            # 204 HTTP status code is returned when api rate limit has been exceeded
+            elif res.status_code == 204:
+                return_error("You've reached your API call quota. Contact your VirusTotal representative.")
```
anara123: Just `You've reached your API call quota.` is enough

bakatzir: talked with alex, it's what he recommended.

PR #2372, opened 2018-11-05
SMG-integration added message on failed login

jochman:
fixes https://github.com/demisto/etc/issues/13613

### DIFF
--- a/Integrations/integration-Symantec_Messaging_Gateway.yml
+++ b/Integrations/integration-Symantec_Messaging_Gateway.yml
@@ -95,7 +95,15 @@ script:
         }
         login_url = 'login.do'
         login_response = http_request('post', login_url, cookies=cookies, data=data)
-        jsession = login_response.cookies.get_dict()['JSESSIONID']
+
+        # check if status is valid
+        login_response.raise_for_status()
+
+        #if JSESSIONID doesn't exist - creds may be invalid
+        try:
+            jsession = login_response.cookies.get_dict()['JSESSIONID']
+        except:
+            raise Exception("Failed to login. Username or password may be invalid")
 
         # Add Jsession ID to the cookies
         COOKIES['JSESSIONID'] = jsession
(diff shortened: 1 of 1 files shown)


Integrations/integration-Symantec_Messaging_Gateway.yml
```diff
@@ -95,7 +95,15 @@ script:
...
-        jsession = login_response.cookies.get_dict()['JSESSIONID']
+
+        # check if status is valid
+        login_response.raise_for_status()
+
+        #if JSESSIONID doesn't exist - creds may be invalid
+        try:
+            jsession = login_response.cookies.get_dict()['JSESSIONID']
+        except:
+            raise Exception("Failed to login. Username or password may be invalid")
```
BenJoParadise: Instead of catching a KeyError, check if the key is there.

Integrations/integration-Symantec_Messaging_Gateway.yml
```diff
@@ -95,7 +95,15 @@ script:
         }
         login_url = 'login.do'
         login_response = http_request('post', login_url, cookies=cookies, data=data)
-        jsession = login_response.cookies.get_dict()['JSESSIONID']
+
+        # check if status is valid
+        login_response.raise_for_status()
```
Itay4: This is unnecessary as `http_request` handles that

Integrations/integration-Symantec_Messaging_Gateway.yml
```diff
@@ -95,7 +95,15 @@ script:
...
+
+        #if JSESSIONID doesn't exist - creds may be invalid
+        try:
+            jsession = login_response.cookies.get_dict()['JSESSIONID']
```
Itay4: This could fail for other reasons, please check if the cookies contain `JSESSIONID`

Integrations/integration-Symantec_Messaging_Gateway.yml
Itay4: Use `return_error` to raise exceptions and use `'` instead of `"`

PR #2375, opened 2018-11-05
Handling PSEOF exception

jochman:
fixes https://github.com/demisto/etc/issues/13757


Scripts/script-ReadPDFFile.yml
```diff
@@ -4,12 +4,33 @@ commonfields:
...
 script: |-
   import pdfx
+  from pdfminer.psparser import PSEOF
   import sys
   reload(sys)
   sys.setdefaultencoding('utf-8')
 
   entryID = demisto.args()['entryID']
 
+  def mark():
```
BenJoParadise: The name can be more descriptive, for example:
```suggestion
  def mark_suspicious():
```

Scripts/script-ReadPDFFile.yml
```diff
@@ -29,9 +50,18 @@ script: |-
   else:
       path = demisto.get(res[0],'Contents.path')
       if path:
-          pdf = pdfx.PDFx(path)
+          try:
+              pdf = pdfx.PDFx(path)
+          except PSEOF:
+              mark()
```
BenJoParadise: Same as above comment
```suggestion
              mark_suspicious()
```

BenJoParadise:
Also, you should add a test for this, please consult @ronykoz if you need help doing this.

ronykoz:
@jochman you should add the pdf file in another PR, otherwise your test won't be able to load the file from git, as it is not there yet. Afterwards you need to add the TestPlaybook for this script.
And I also want you to add a regular PDF and make sure it is not marked as a suspicious one(Another check for the TestPlaybook).

PR #2378, opened 2018-11-05
Rasterize image command

gshriki:
Taking an images entry and provide a secure image (using rasterize process)


Integrations/integration-rasterize.yml
```diff
@@ -56,6 +56,45 @@ script:
...
         sys.exit(0)
 
+    if demisto.command() == 'rasterize-image':
+        res = demisto.getFilePath(demisto.args()['EntryID'])
+        # demisto.results({'res':res})
+        # res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
+        with open(res['path'], 'r') as f:
+            data = f.read()
+        b64 = base64.b64encode(data)
+        html = '<img src="data:image/png;base64, ' + b64 + '">'
```
slavikm: Why PNG and not use file magic to get actual type? It can be JPEG or GIF as well.

Integrations/integration-rasterize.yml
```diff
@@ -56,6 +56,45 @@ script:
...
+
+        return_code = 0
+        error_message = ''
+
+        friendlyName = 'image.png'
+        f = open('htmlImage.html','w')
+        f.write('<html style="background:white";>' + html + '</html>')
```
slavikm: `;` should be inside the style and not after `"`. Also, I would add `<body>` tags just to be nice.

Integrations/integration-rasterize.yml
```diff
@@ -56,6 +56,45 @@ script:
...
+        f.close()
+
+        proxy_flag = ""
+        if proxy:
```
slavikm: Why do we care about proxy? We are not exiting to network, no?

Integrations/integration-rasterize.yml
```diff
@@ -56,6 +56,45 @@ script:
...
+        command = ['phantomjs', proxy_flag, '/usr/local/bin/rasterize.js', 'htmlImage.html', friendlyName]
+        if demisto.get(demisto.args(), 'width') and demisto.get(demisto.args(), 'height'):
+            command.append(demisto.gets(demisto.args(), 'width') + '*' + demisto.gets(demisto.args(), 'height'))
+        try:
+            error_message = subprocess.check_output(command)
+        except Exception as e:
+            return_code = -1
+            error_message = e.output
+
+        if return_code == 0:
```
slavikm: I think there is a function in common that does this - `file_result_existing_file`

slavikm: Still does most of it and saves lines, no? Just say `res = file_result_ex.....` and then change the type and return, no?

PR #2387, opened 2018-11-06
Alexa Ranking Integration

amshamah419:
Uses Amazon's Alexa rankings as an indicator of whether or not a domain is trustworthy.

Threshold is user configurable.

Unlisted domains (unranked) are flagged as "Bad" via a DBot score of 3
Low ranking domains are flagged as "Suspicious" via a DBot score of 2
Trusted domains are flagged as "Good" via a DBot score of 1


amshamah419:
Addresses demisto/etc#13824

amshamah419:
I believe it is pending design. The integration itself is very simple, we just need to determine how we want to use the data we get back.

Itay4:
@NeelRust What's the plan with this one?

PR #2388, opened 2018-11-06
Panorama - more commands, deprecate scripts, add outputs,HR, and fw vs panorama commands compliance issues.

bakatzir:
@anilagr  @hassanainkapadia  FYI
fix: https://github.com/demisto/etc/issues/13623
fix: https://github.com/demisto/etc/issues/13676
fix: https://github.com/demisto/etc/issues/14610
reference: https://github.com/demisto/etc/issues/5840

and move it to python!

- deprecate these scripts:
screen shot 2018-11-06 at 16 58 55
and add support in the integration with relevant outputs.
- [x] blockip

- [x] edit `Block IP - Generic` to use panorama-block-ip

- [x] panorama move

- [x] panoramacommit (added panorama-commit)

- [x] panoramadynamicaddresslist - how to deprecate? talked with shriki, since it didn't work - will probably delete.
- [x] panorama-dynamic-block command - how to deprecate? talked with shriki, since it didn't work - will probably delete.
- [x] added panorama-create/edit/get/delete-address-group

- [x] panoramaconfig (added panorama-create-rule)

- [ ] panoramapcaps

- [x] add block application command @shilonoa (via panorama-custom-block-rule)

- [x] add block address group command (via panorama-custom-block-rule)

- [x] add panorama-get-applications to create full flow

- [x] add panorama-block-url capabilities - done via panorama-custom-block-rule
- [x] add panorama-create/delete-custom-url-category

- [x] add a command to add/remove sites in Custom URL category. See under Objects->Custom Objects->URL Category. The command should allow addition/deletion of "Sites" in a particular "Name" list.

- [x] add panorama-enable-rule (by rulename)

- [x] add panorama-disable-rule (by rulename)

- [x] FW vs panorama products code. maybe split integrations? let's decide @anara123 - talked with dan and amichai, decide to keep all in same integration, will need to make xpath better. i'll think of a solution. 
- [x] add handling of URL Filter object get/create/edit/delete
- [x] add test PB for FW
- [x] add test PB for Panorama
- [ ] add documentation


bakatzir:
was able to block ip and application using commands. hope everyone is excited as i am.
screen shot 2018-11-07 at 21 29 47

bakatzir:
@anara123  basically you can look at the error in circleci, it is handled now.
so it's even good that the test failed now on quota.

PR #2391, opened 2018-11-06
DemistoUploadFileToIncident Automation

secdevopsai:
This automation will leverage Entry ID to upload file from context as attachment to a specified incident ID


secdevopsai:
@anara123 - Thanks for feedback -- Explained very well.  I added a new commit to this PR

PR #2394, opened 2018-11-07
Added camelize function to CommonServerPython

BenJoParadise:


BenJoParadise:
Tested the camelize function after the commit that states that it's untested.

ronykoz:
@anara123 you are ok with adding such a function, right?
It will make our lives easier when converting human readable to context format

Scripts/script-CommonServerPython.yml
```diff
@@ -1234,4 +1256,4 @@ dependson: {}
 timeout: 0s
 tests:
   - TestPYCommonServer
-releaseNotes: "Add djb2 hash function"
```
ronykoz: Not sure this should be removed

BenJoParadise: It's not my release note, i think this goes to Alex after every merge (it's there from the last merge of this file to Master).

Not sure though, how can we know for sure?

ronykoz: These releaseNotes are staying there until Itay removes them after the release, thus these release notes are there for the next release, you can check with him to make sure of that

Itay4: @ronykoz is right. In this cases, just add numbering and append your release notes.

BenJoParadise:
@anara123 i followed @ronykoz steps, can you please force merge?

PR #2300, opened 2018-10-22
Tenable.io

BenJoParadise:
Reopened in https://github.com/demisto/content/pull/2397

PR #2397, opened 2018-11-07
Tenable.io

BenJoParadise:
fixes https://github.com/demisto/etc/issues/12037
merged changes from https://github.com/demisto/content/pull/2300


Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+            results.append(get_entry_for_object('Assets', 'TenableIO.Assets', assets, SCAN_REPORT_HOSTS_HEADERS))
+            if 'remediations' in scan_details and 'remediations' in scan_details['remediations'] and len(scan_details['remediations']['remediations']) > 0:
+                remediations = replace_keys(scan_details['remediations']['remediations'], REMEDIATIONS_NAMES_MAP)
+                results.append(get_entry_for_object('Remediations', 'TenableIO.Remediations', remediations, SCAN_REPORT_REMEDIATIONS_HEADERS))
+        return results
+
+    def get_vulnerability_details(plugin_id, date_range):
+        info = send_vuln_details_request(plugin_id, date_range)
+        if 'error' in info:
+            return { 'Type': entryTypes['error'], 'ContentsFormat': formats['text'], 'Contents': info['error'] }
```
anara123: format the dictionary. 
`{ 'Type': entryTypes['error'], 'ContentsFormat': formats['text']`
=>
```
{
        'Type': entryTypes['error'],
.....
}
```

Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+
+    # Command methods
+    def test_module():
+        send_scan_request()
+        return 'ok'
+
+    def get_scans(folder_id, last_modification_date):
+        if last_modification_date:
+            last_modification_date = int(time.mktime(datetime.strptime(last_modification_date[0:len('YYYY-MM-DD')], "%Y-%m-%d").timetuple()))   # str(YYYY-MM-DD) to int(timestamp)
+        response = send_scan_request(folder_id=folder_id, last_modification_date=last_modification_date)
```
anara123: we usually call the wrapper function
`get_scans_command` and the inner function `get_scans`
this way it is more clear.
please change all the functions accordingly

Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+            if ignore_license_error and res.status_code in (403, 500):
+                return
+            err_msg = get_scan_error_message(res, scan_id)
+            if demisto.command() != 'test-module':
+                err_msg = {
+                "Type": entryTypes["error"],
+                "ContentsFormat": formats["text"],
+                "Contents": err_msg
+                }
+            demisto.results(err_msg)
```
anara123: use `return_error` everywhere where you return an error

Integrations/integration-Tenable_io.yml
anara123: same use return_error

Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+            indicator = asset_id
+        info = send_asset_vuln_request(asset_id, date_range)
+        if 'error' in info:
+            return { 'Type': entryTypes['error'], 'ContentsFormat': formats['text'], 'Contents': info['error'] }
+        vulns = replace_keys(info['vulnerabilities'], ASSET_VULNS_NAMES_MAP)
+        return get_entry_for_object('Vulnerabilities for asset {}'.format(indicator), 'TenableIO.Vulnerabilities', vulns, ASSET_VULNS_HEADERS)
+
+    def get_scan_status(scan_id):
+        scan_details = send_scan_request(scan_id)
+        scan_status = { 'Id': scan_id, 'Status': scan_details['info']['status'] }
```
anara123: same format the dictionary

Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+        return get_entry_for_object('Scan status for {}'.format(scan_id), 'TenableIO.Scan', scan_status)
+
+
+    # Command selector
+    if demisto.command() == 'test-module':
+        demisto.results(test_module())
+    elif demisto.command() == 'tenable-io-list-scans':
+        demisto.results(get_scans(demisto.getArg('folderId'), demisto.getArg('lastModificationDate')))
```
anara123: the args we extract inside of the `_command` function and pass them explicitly to inner functions

Integrations/integration-Tenable_io.yml
```diff
@@ -0,0 +1,673 @@
...
+            return formatEpochDate(v) if isinstance(v, int) and any(s in k.lower() for s in ('date', 'time')) else v
+
+        if isinstance(d, list):
+            return map(convert_dict_readable_dates, d)
+        if isinstance(d, dict):
+            return {k: convert_dict_readable_dates(convert_epoch_to_date(k, v)) for k, v in d.iteritems()}
+        return d
+
+    def get_entry_for_object(title, context_key, obj, headers=None, remove_null=False):
+        # Python program to illustrate the intersection
```
anara123: remove the comment

Integrations/integration-Tenable_io.yml
BenJoParadise: The inner function is a wrapper for requests.request() and is used in many other functions, so `get_scans` is not a good name for it. There is also no need to wrap it in a `get_scans` function, because it will be a one-liner that is used only once.

I will change the "command" functions, but i'm keeping the name of the inner function, unless there is a better name that can be used in all calls to the function.

PR #2401, opened 2018-11-08
deprecate CBFindHash

noaco:
need to add test before merge


noaco:
reviewed with Jonatan - this script should be deprecated.

PR #2406, opened 2018-11-08
Common Server Python common code

Itay4:
Added `http_request` function and common integration code


anara123:
@Itay4 @yaakovi why do we need this function? `requests` do the same job.
The only reason we create `http_request` function in the integration is to contain integration specific things like cookies, headers, authentication

Itay4:
@anara123 For the same reason we have `http_request` in every integration. We could also use just `requests` in every integration, but we have our standard on what to do with exceptions, so we can just use that function I added.

PR #2413, opened 2018-11-08
rsa netwitness 11.1 bug fix

noaco:
When trying to get a token - if an error is returned from the server, parsing of the error response is failing. The function parsing the error response should be declared before global variables declaration.


anara123:
and the test fails

PR #2415, opened 2018-11-09
Enforcing different ids for scripts/integration/playbooks

ronykoz:
Added validation that an ID from a new scripts/integration/playbooks does not exist to the validate_files-structure script.

We have to take into account that it takes ~10 seconds to commit now, thought about creating a json with all existing ids and update it with new files, what do you think?

### DIFF
--- a/Tests/scripts/validate_files_structure.py
+++ b/Tests/scripts/validate_files_structure.py
@@ -20,6 +20,7 @@ try:
 except ImportError:
     print "Please install pykwalify, you can do it by running: `pip install -I pykwalify`"
     sys.exit(1)
+import glob
 import json
 import re
 import os
@@ -178,10 +179,27 @@ def validate_schema(file_path, matching_regex=None):
 def changed_id(file_path):
     change_string = run_git_command("git diff HEAD {0}".format(file_path))
     if re.search("\+id: .*", change_string) or re.search("\-id: .*", change_string):
-        print_error("You've changed the ID of the playbook {0} please undo.".format(file_path))
+        print_error("You've changed the ID of the file {0} please undo.".format(file_path))
         return True
 
     return False
+
+
+def changed_integration_id(file_path):
+    change_string = run_git_command("git diff HEAD {0}".format(file_path))
+    if re.search("\+  id: .*", change_string) or re.search("\-  id: .*", change_string):
+        print_error("You've changed the ID of the file {0} please undo.".format(file_path))
+        return True
+
+    return False
+
+
+def get_script_or_integration_id(file_path):
+    data_dictionary = get_json(file_path)
+
+    if data_dictionary:
+        commonfields = data_dictionary.get('commonfields', {})
+        return commonfields.get('id', ['-', ])
 
 
 def get_json(file_path):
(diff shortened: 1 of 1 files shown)


ronykoz:
Before merging I'll need to update the id_set.json so it will contain all the ids in content

Tests/scripts/validate_files_structure.py
```diff
@@ -268,21 +312,32 @@ def validate_all_files():
...
                      found_wrong_name = True
-                if not validate_schema(os.path.join(root, file_name), regex):
-                    print_error("file " + os.path.join(root, file_name) + " schema is wrong.")
+                if not validate_schema(file_path, regex):
+                    print_error("file " + file_path + " schema is wrong.")
                     wrong_schema = True
-
-    if wrong_schema or found_wrong_name:
+                if re.match(SCRIPT_REGEX, file_path, re.IGNORECASE) or re.match(INTEGRATION_REGEX, file_path, re.IGNORECASE):
+                    id = get_script_or_integration_id(file_path)
```
yaakovi: don't override builtin keywords, use `_id`

Tests/scripts/validate_files_structure.py
```diff
@@ -178,12 +177,29 @@ def validate_schema(file_path, matching_regex=None):
...
     change_string = run_git_command("git diff HEAD {0}".format(file_path))
     if re.search("\+id: .*", change_string) or re.search("\-id: .*", change_string):
-        print_error("You've changed the ID of the playbook {0} please undo.".format(file_path))
+        print_error("You've changed the ID of the file {0} please undo.".format(file_path))
         return True
 
     return False
 
 
+def changed_integration_id(file_path):
```
yaakovi: why do you need another command? just set the space as optional in the regex:
`'[+-](  )?id: .*'`

Tests/scripts/validate_files_structure.py
```diff
@@ -234,6 +264,9 @@ def validate_committed_files(branch_name):
         if re.match(PLAYBOOK_REGEX, file_path, re.IGNORECASE) or re.match(SCRIPT_REGEX, file_path, re.IGNORECASE) or re.match(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE):
```
yaakovi: since in the script the id is not at the root of the json, this should be with the integration and not the playbooks.
but in my opinion this should be the same function anyway (changed_id)

Tests/scripts/validate_files_structure.py
```diff
@@ -252,12 +286,22 @@ def validate_committed_files(branch_name):
                 missing_test = True
                 print_error("You've failed to add the {0} to conf.json".format(file_path))
 
+        if re.match(SCRIPT_REGEX, file_path, re.IGNORECASE) or re.match(INTEGRATION_REGEX, file_path, re.IGNORECASE):
+            id_to_file[get_script_or_integration_id(file_path)] = file_path
+        elif re.match(PLAYBOOK_REGEX, file_path, re.IGNORECASE) or re.match(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE):
+            id_to_file[collect_ids(file_path)] = file_path
+
+    if has_duplicated_ids(id_to_file):
```
yaakovi: isn't this already tested in the `validate_all_files`?

ronykoz: Yes, but i want to make sure that we do full check on master, just to cover all our bases

Tests/scripts/validate_files_structure.py
```diff
@@ -177,13 +176,30 @@ def validate_schema(file_path, matching_regex=None):
 
 def changed_id(file_path):
     change_string = run_git_command("git diff HEAD {0}".format(file_path))
-    if re.search("\+id: .*", change_string) or re.search("\-id: .*", change_string):
-        print_error("You've changed the ID of the playbook {0} please undo.".format(file_path))
+    if re.search("[+-]( )?id: .*", change_string):
```
yaakovi: in the original regex there was double space in between the +/- and the word id.
git removed the duplication in my comment (sorry about that).

PR #2416, opened 2018-11-10
Cylance Protect v2 Enhancement

yardensade:
Added cylance protect v2 commands, added renewed test, removed test from skip


yardensade:
https://github.com/demisto/etc/issues/13976

PR #2429, opened 2018-11-13
removed doubled instance of detonate file - generic test

bakatzir:


ronykoz:
whats the difference between those two tests? I've noticed that the other one has more integrations configured for it, are we sure there is no reason for that?

bakatzir:
which is why we don't need the first. saw in blame that itay added it so that's why i put him as a reviewer.

yaakovi:
this run will check to see that the playbook will finish successfully while there are no integrations configured in demisto.

PR #2432, opened 2018-11-13
Sandblast Context Fixes

amshamah419:
Addresses demisto/etc#14001


Integrations/integration-Sandblast.yml
```diff
@@ -242,7 +242,27 @@ script:
                     maliciousData[uniqueKeys[key].toUpperCase()] = raw.Contents.response[uniqueKeys[key]];
                 }
             }
+            setContext('DBotScore', {'Indicator': raw.Contents.response.md5, 'Type': 'hash', 'Vendor': 'SandBlast', 'Score': 3, 'SubType': 'av'});
```
yuvalbenshalom: why do you use setContext? I don't think we should use it any more, but rather return the context values in the entry

Integrations/integration-Sandblast.yml
```diff
@@ -242,7 +242,27 @@ script:
...
+                    Severity: raw.Contents.response.te.severity
+                }
+            };
+            for (var key in uniqueKeys) {
+                if (raw.Contents.response[uniqueKeys[key]]) {
+                    maliciousData[uniqueKeys[key].toUpperCase()] = raw.Contents.response[uniqueKeys[key]];
+                }
+            }
+
+            setContext('DBotScore', {'Indicator': raw.Contents.response.md5, 'Type': 'hash', 'Vendor': 'SandBlast', 'Score': 3, 'Subtype':'te'});
```
yuvalbenshalom: same here with setContext

Integrations/integration-Sandblast.yml
```diff
@@ -251,7 +271,10 @@ script:
                     ec['File(val.' + uniqueKeys[key].toUpperCase() + ' == obj.' + uniqueKeys[key].toUpperCase() + ')'] = val;
                 }
             }
+            setContext('DBotScore', {'Indicator': raw.Contents.response.md5, 'Type': 'hash', 'Vendor': 'SandBlast', 'Score': 3, 'SubType':'--'});
```
yuvalbenshalom: same here with setContext

Integrations/integration-Sandblast.yml
```diff
@@ -222,63 +222,96 @@ script:
     if (command === 'sandblast-upload' || command === 'sandblast-query' || command === 'sb-upload' || command === 'sb-query') {
         var uniqueKeys = ['md5', 'sha1', 'sha256'];
```
yuvalbenshalom: change this to hashTypes

Integrations/integration-Sandblast.yml
```diff
@@ -222,63 +222,96 @@ script:
...
                 }
             };
             for (var key in uniqueKeys) {
-                if (raw.Contents.response[uniqueKeys[key]]) {
-                    maliciousData[uniqueKeys[key].toUpperCase()] = raw.Contents.response[uniqueKeys[key]];
+                if (contents.response[uniqueKeys[key]]) {
+                    maliciousData[uniqueKeys[key].toUpperCase()] = contents.response[uniqueKeys[key]];
                 }
             }
+            var dbot_con = {'Indicator': raw.Contents.response.md5, 'Type': 'hash', 'Vendor': 'SandBlast', 'Score': 3, 'SubType': 'av'};
```
yuvalbenshalom: @mayagoldb question: are we adding file dot score only with md5? should we add it with other hashes if we have them?

mayagoldb: @yuvalbenshalom @amshamah419 If our indicator is sha1 or sha256 we should also add dbot score

yuvalbenshalom: @amshamah419 lets take a look at it together and see if there is a need to add sha1/256

PR #1987, opened 2018-08-20
Using ExtractURL instead of pdfx library

secdevopsai:
pdfx library also categorizes domains as URLs.  Switching to ExtractURL to accurately display URLs presented in PDF

related to:
https://github.com/demisto/etc/issues/12962

PR #2434, opened 2018-11-13
ReadPDF: Parse url as they should be

jochman:
working on: https://github.com/demisto/content/pull/1987
us
fixes: https://github.com/demisto/etc/issues/9432


jochman:
Still need to check why it isn't extracting URLs from hyperlinks. Please do not merge

jochman:
using of the extracturl command is ignoring hypertext links. The old one extracting malformed links, but still works better. 
I've tried to combine those lists unsuccessfully.
I suggest that we drop that, the old one works good enough. @Itay4  what do you say?

Itay4:
Why can't we do both?

jochman:
I can do both, but then we'll get a huge list with duplicate values. 
There are lots of edge cases for merging the two lists into unique values. 

I can do it, but it'll not be efficient and the solution (that I got in mind, at least) may not work every time because of the difference between the two methods used to extract the links (return values and encoding). it'll become not efficient and slow for big PDFs. But still can be done, should I put my efforts into it?

aviadl:
Lets review this together before merging when i am back

anara123:
@jochman what is the status here? review with @aviadl ?

jochman:
> @jochman what is the status here? review with @aviadl ?

It's a server-side issue (in extract indicator). I worked on a solution and should give it an effort. Right now it's not prioritized.

PR #2435, opened 2018-11-13
Update integration-PhishAI.yml

idovandijk:
Url should be URL, as in the rest of the integrations, to maintain consistency

### DIFF
--- a/Integrations/integration-PhishAI.yml
+++ b/Integrations/integration-PhishAI.yml
@@ -107,7 +107,7 @@ script:
                     ContentsFormat: formats.json,
                     HumanReadable: md,
                     EntryContext: {
-                        'URL(obj.Data.Url && obj.Data.Url===val.Data.Url)' : {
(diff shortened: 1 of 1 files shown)


hirshbergshachar:
@Itay4 officially the output was set to `URL` so unless someone opened the code and wrote something according to it we don't break BC here. 
@idovandijk I added another comment, please fix

Integrations/integration-PhishAI.yml
```diff
@@ -107,7 +107,7 @@ script:
                     ContentsFormat: formats.json,
                     HumanReadable: md,
                     EntryContext: {
-                        'Url(obj.Data.Url && obj.Data.Url===val.Data.Url)' : {
+                        'URL(obj.Data.Url && obj.Data.Url===val.Data.Url)' : {
```
hirshbergshachar: @idovandijk `obj.Data.Url && obj.Data.Url===val.Data.Url` Please change to `URL` inside here as well

PR #2438, opened 2018-11-13
Update integration-FireEyeHX.yml

idovandijk:
"containmentn" was definitely a typo


Itay4:
@idovandijk What is the status of this?

idovandijk:
> @idovandijk What is the status of this?

There are currently issues regarding playbooks that I am working on. I don't know if I will get to fixing this integration, maybe the creators of it can.

PR #2440, opened 2018-11-13
Better splunk default search for time

meirwah:


meirwah:
@Itay4 can you review?

PR #2446, opened 2018-11-16
Skipping phishing test

ronykoz:
DO NOT MERGE, still have couple of changes to do


ronykoz:
@Itay4 I've moved phishing test to skipped_tests, I need to review those tests and once i'll do that i will return them to the build.

PR #2450, opened 2018-11-17
Enforcing integration logo's to be less than 10kB

ronykoz:
fixes https://github.com/demisto/etc/issues/14064#event-1972643996

### DIFF
--- a/Tests/scripts/validate_files_structure.py
+++ b/Tests/scripts/validate_files_structure.py
@@ -28,7 +28,6 @@ from pykwalify.core import Core
 
 # Magic Numbers
 IMAGE_MAX_SIZE = 10 * 1024  # 10kB
-IMAGE_PREFIX_LENGTH = 22
 
 # dirs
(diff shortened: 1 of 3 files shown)


ronykoz:
@dorsha you can take a look at the CircleCi run and see the message, can you Force merge this(If you don't have any comments of course)?

Tests/scripts/validate_files_structure.py
```diff
@@ -223,36 +225,53 @@ def is_test_in_conf_json(file_path):
     return False
 
 
+def oversize_image(file_path):
+    data_dictionary = get_json(file_path)
+    image = data_dictionary.get('image', '')
+    if image == '':
+        return False
+
+    image_size = ((len(image) - IMAGE_PREFIX_LENGTH) / 4.0) * 3
```
dorsha: I would use: `sys.getsizeof(image)`
You can also count the prefix

PR #2456, opened 2018-11-19
Add an option to run Content build with latest server

ronykoz:
## Status
- [ ] Ready
- [x] In code review
- [X] In Progress
- [ ] Hold
      Reason for hold:

## Description
Make Content build work with latest server Master


ronykoz:
DO NOT MERGE

yuvalbenshalom:
@ronykoz looks very good.
questions: 
1. where is the triggering script we should use to run the script? am i missing something?
2. why the mix with the integration files on the same branch?

got conflicts...

ronykoz:
@yuvalbenshalom  
1. My bad, added the script.
2. We have couple of problems with the validator script so it did not allow me to commit these files till i do these changes, added a possible fix but need to think this through

Utils/trigger_content_non_ami_build.sh
```diff
@@ -0,0 +1,34 @@
...
+_circle_token=$2
+
+trigger_build_url=https://circleci.com/api/v1/project/demisto/content/tree/${_branch}?circle-token=${_circle_token}
+
+if [ -z "$3"]
+  then
+    post_data=$(cat <<EOF
+    {
+      "build_parameters": {
+        "AMI_RUN": "true"
```
yuvalbenshalom: should AMI_RUN be true in both cases?

ronykoz: Actually it should be false, but on both cases

yuvalbenshalom:
resolve conflicts with integrations

PR #2471, opened 2018-11-20
Update interval configuration for testing

ronykoz:
## Status
- [ ] Ready
- [X] In code review
- [X] In Progress
- [ ] Hold
      Reason for hold:

## Description
Removing the sleep time between each check of the playbook status


ronykoz:
@asafshen what do you think about this change, seems legit?

asafshen:
@ronykoz 👍 
its legit, but then I would remove `interval` from test conf (just add it as a constant here)

ronykoz:
@asafshen Done, thanks!

PR #2476, opened 2018-11-20
Update content build user

Itay4:


ronykoz:
@anara123  this is the issue that needs force merge

PR #2484, opened 2018-11-22
Adding an option to set fetch interval

ronykoz:
Adding a fetch interval param for fetch incidents

### DIFF
--- a/Integrations/integration-PagerDutyV2.yml
+++ b/Integrations/integration-PagerDutyV2.yml
@@ -27,6 +27,11 @@ configuration:
   defaultvalue: ""
   type: 13
   required: false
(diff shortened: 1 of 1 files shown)


Integrations/integration-PagerDutyV2.yml
```diff
@@ -27,6 +27,11 @@ configuration:
   defaultvalue: ""
   type: 13
   required: false
+- display: Initial Fetch Interval(In minutes)
```
BenJoParadise: Please add a few words explaining that this is only used for the first fetch.
I know it says "Initial" there, but customers can miss it (like i did on first read) and think that they are setting the actual fetch interval.

PR #2488, opened 2018-11-22
modify ADGetUser to return multiple results when using costume query  

noaco:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14033

## Description
Will return multiple results from AD also in case a query argument was passed (instead of a single result)


noaco:
@anara123 reviewed with Jonathan. Please approve.

PR #2491, opened 2018-11-22
document generator script from terminal

anara123:
## Status
Ready

## Description
Being able to generate docs from terminal via python script

   - No

## Must have
- [ ] [Documentation](https://github.com/demisto/etc/wiki/Creating-an-Integration-in-Demisto-%5BDocumentation%5D) 
- [ ] Code Review


Tests/doc_generator.py
```diff
@@ -0,0 +1,203 @@
...
+# Known limitations
+doc += '\n## Known limitations:'
+
+# Troubleshooting
+doc += '\n## Troubleshooting:\n'
+
+# Possible Errors
+# if 'True' == demisto.args()['withErrors']:
+#     doc += '\n## Possible Errors (DO NOT PUBLISH ON ZENDESK):\n'
+#     doc += add_error_lines(data_map['script']['script'], data_map['script']['type'])
```
ronykoz: Remove these lines

Tests/doc_generator.py
```diff
@@ -0,0 +1,203 @@
...
+
+filename = os.path.basename(sys.argv[1]).replace('documentation-', 'integration-').replace('.yml', '.txt').replace('.yaml', '.txt')  # strip all the spaces
```
ronykoz: Lets save the documentation file in other location, perhaps outside the repo so that developers won't accidentally commit these changes to the repo.

Tests/doc_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    print("You must provide full path of integration yml file")
+    exit(1)
+
+path = sys.argv[1]
+yamlFile  = open(path)
+data_map = yaml.safe_load(yamlFile)
+yamlFile.close()
+
+
+def add_lines(line):
```
ronykoz: not used, can be removed

Tests/doc_generator.py
ronykoz: If these line will be removed than we also don't use the add_error_lines function so it also can be removed.

Tests/doc_generator.py
ronykoz: where is that option, couldn't find it

PR #2506, opened 2018-11-26
Cylance Added SHA256 to output in yml

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14201

## Description
Added SHA256 to output in yml.
   - No


idovandijk:
Thanks guys

PR #2512, opened 2018-11-26
enhancement to script-ExtractDomainFromUrlFormat.yml 

lostInSpaceSomewhere:
## Status
Ready

## Related Issues
fixes: Hyphens break domain extraction from email, email domain extraction does not pull parent domain.

## Description
Added to the domain regex the ability to include hyphens ("-") in an email domain.  Also found that when extracting domains from email, it will include subdomains and therefore could impact the analysis of the domain.   Pushing the parent domain along with the full email domain to results allows better analysis of the domain.

## Related PRs

branch | PR
------ | ------

latest

   - No

## Must have
- [ ] Tests
- [x] Code Review

## Additional changes


glicht:
@lostInSpaceSomewhere thanks for submitting this. As a start I did a rebase on the branch as we recently did a rebase on our main repo. The branch is now in sync with our main repo. I've also reverted back the changes you did to EICAR.exe. It looks to be an accidental change.

We will soon do a code review and update.

Scripts/script-ExtractDomainFromUrlFormat.yml
```diff
@@ -10,11 +10,19 @@ script: |-
 
   inp = demisto.args()['input']
 
+  results = []
+
   is_email = validate_email(inp)
   if is_email:
-      domain = re.search("@[\w.]+", inp)
+      domain = re.search("@[\w.|\-]+", inp)
       if len(domain.group()) > 0:
```
yaakovi: this will always be true in case the regex return any result.
if the regex returns without any result in will return None and this command will failed.

```suggestion
      if domain is not None:
```

Scripts/script-ExtractDomainFromUrlFormat.yml
```diff
@@ -10,11 +10,19 @@ script: |-
...
-          demisto.results(domain.group()[1:])
+          extractedDomain = domain.group()[1:]
+          results.append(extractedDomain)
+          _list = extractedDomain.split(".")
+          if len(_list) > 2:
+              parentDomain = _list[len(_list)-2] + "." + _list[len(_list)-1]
```
yaakovi: why specifically the last two?

```suggestion
              parentDomain = '.'.join(_list[-2:])
```

yaakovi: seems reasonable, thanks for the clarification.

PR #2524, opened 2018-11-27
Lastline - handle quota error gracefully

bakatzir:
# re-run early, since has a daily quota

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14219

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Lastline.yml
+++ b/Integrations/integration-Lastline.yml
@@ -119,7 +119,7 @@ script:
             params.proxy
         );
         if (result.StatusCode < 200 && result.StatusCode > 299) {
-            throw 'Failed to perform request ' + path + ', request status code: ' + result.StatusCode;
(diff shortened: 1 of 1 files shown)


bakatzir:
@anara123  basically you can look at the error in circleci, it is handled now.
so it's even good that the test failed now on quota.

Integrations/integration-Lastline.yml
```diff
@@ -147,10 +144,16 @@ script:
                 case 106:
                     // analysis report not ready
                     break;
+                case 115:
+                    // reached api quota
+                    throw "You've reached your API call quota. Contact your Lastline representative.";
```
anara123: instead of throwing please change it to return error entry, or catch the error at the higher level and return error entry. When we are throwing it shows ugly stack trace

bakatzir: true. done.

PR #2525, opened 2018-11-27
dnstwist new integration

saharsabban:
dnstwist new integration

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/7336

## Description
Adding new integration of dnstwist 

x.x.x

   - No

## Must have
- [ ] Tests  - dnstwistTest playbook
- [ ] Documentation - https://github.com/demisto/etc/issues/14228
- [ ] Code Review

### DIFF
--- a/Integrations/integration-dnstwist.yml
+++ b/Integrations/integration-dnstwist.yml
@@ -1,0 +1,76 @@
+commonfields:
+  id: dnstwist
+  version: -1
+name: dnstwist
+display: dnstwist
+category: Data Enrichment & Threat Intelligence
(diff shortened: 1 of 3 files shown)


Integrations/integration-dnstwist.yml
```diff
@@ -0,0 +1,76 @@
...
+configuration: []
+script:
+  script: |+
+    import subprocess, json
+
+    if demisto.command() == 'domain':
+
+        DOMAIN = demisto.args()['domain']
+        KEY = "domain-name"
+        KEYS_TO_COMPARE = ["whois-updated", "whois-created", "dns-a", "dns-mx"]
```
erezh31: make this an argument to the command with default (comma-separated)

erezh31: and rename it - "extractKeys" - there is no comparison here

Integrations/integration-dnstwist.yml
```diff
@@ -0,0 +1,76 @@
...
+                if temp:
+                    temp["domain-name"] = x["domain-name"]
+                    results.append(temp)
+            return results
+
+        dnstwist_result = get_dnstwist_result(DOMAIN) #contents
+        new_result = get_domain_to_info_map(dnstwist_result)
+        md = tableToMarkdown('dnstwist for domain - '  + DOMAIN, new_result)
+
+        DomainContext = new_result[0]
```
erezh31: why do you take only the results in index 0?
what about the rest? I think the point of this whole integration is the similar dns results and not the exact same one (otherwise it's better to use whois or domain tools...). SO, I think we will to store in the context all the similar incidents, under Domain.dnstwist context key. ask @michalgold

erezh31:
Also fix conflicts + CI

saharsabban:
Closing this PR - will open a new one with the redesign integration.

PR #2535, opened 2018-11-28
Clear release notes - 18.11.2

Itay4:


Tests/scripts/validate_files_structure.py
```diff
@@ -409,4 +409,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
asafshen: not sure what is this change

yaakovi: we started to add something and erased it...

asafshen: cool

PR #2536, opened 2018-11-28
removed related urls from context

amshamah419:
## Status
Ready

## Description
Related URLs was added to context and not human readable. Removing from context.

   - No

### DIFF
--- a/Integrations/integration-UrlScan.yml
+++ b/Integrations/integration-UrlScan.yml
@@ -739,5 +739,6 @@ script:
     description: Submit a url to scan
   runonce: false
 fromversion: 3.5.0
+releaseNotes: "Formatted context"
 tests:
   - urlscan_malicious_Test
--- a/Integrations/integration-UrlScan.yml
+++ b/Integrations/integration-UrlScan.yml
@@ -534,7 +534,7 @@ script:
       defaultValue: "20"
     outputs:
     - contextPath: URLScan.URL
-      description: Bad URLs found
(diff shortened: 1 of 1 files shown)


Integrations/integration-UrlScan.yml
```diff
@@ -205,7 +205,6 @@ script:
             cont['Certificates'] = cert_ec
         url_cont['Data'] = url_query
         if 'urls' in scan_lists:
-            cont['RelatedURLs'] = scan_lists['urls']
```
ronykoz: why are we removing this?

amshamah419: Should be fixed now. With regards to the context outputs, they were vague so I changed the description to match what they actually do as well as removed the outputs that no longer exist.

Integrations/integration-UrlScan.yml
```diff
@@ -586,17 +585,14 @@ script:
...
-      description: Bad URLs found
+      description: URL submitted
       type: string
     - contextPath: URL.Malicious.Vendor
       description: For malicious URLs, the vendor that made the decision
       type: string
     - contextPath: URL.Malicious.Description
       description: For malicious URLs, the reason for the vendor to make the decision
       type: string
-    - contextPath: URLScan.URLs
```
ronykoz: shouldn't we have this? there is a setting for cont['URL'], which prob mean we should get URLScan.URL, no?

PR #2538, opened 2018-11-28
Added Machine Isolation playbook

idovandijk:
Added a playbook for Machine Isolation, a generic playbook, and a test playbook


idovandijk:
@JonathanMeler I pushed the required fixes for correctly validating input

PR #2539, opened 2018-11-28
Updating the exception catching form in PagerDuty integration

mcpeak:
This commit changes how exceptions are caught from 'Exception, e'
to 'Exception as e'. Among other things this makes the code Python
3 compatible.

## Status
Ready

## Description
Update PagerDuty integration to catch Exceptions in a Python3 compatible way.

   - No

### DIFF
--- a/Integrations/integration-PagerDutyV2.yml
+++ b/Integrations/integration-PagerDutyV2.yml
@@ -118,7 +118,7 @@ script:
 
             return res.json()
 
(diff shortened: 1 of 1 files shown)


ronykoz:
@mcpeak Sorry about that, but I wasn't aware of the correct work process for PRs from customers so i've done those minor changes by myself and added that to the the PagerDuty integration, closing this issue.

PR #2553, opened 2018-12-02
Base64 list to file - support zip file

erezh31:


Scripts/script-Base64ListToFile.yml
```diff
@@ -13,12 +15,17 @@ script: |-
 
   # get the list
   res = demisto.executeCommand('getList', {'listName': demisto.args()['listname']})
+  if res['Type'] == entryTypes['error']:
+    raise Exception("error reading list %s from demisto" % demisto.args()['listname'])
```
Itay4: Please use `return_error` function and remove `sys.exit(0)`

Scripts/script-Base64ListToFile.yml
```diff
@@ -13,12 +15,17 @@ script: |-
...
+    sys.exit(0)
 
   # convert base64 file to binary file
   bin_file = base64.decodestring(res[0]['Contents'])
+  if demisto.args()['isZipFile']:
```
Itay4: Did you mean to check if `yes` then get in the condition? The code in the condition will always be executed while any string is given to the argument.

erezh31: sure:)

Scripts/script-Base64ListToFile.yml
erezh31: V

erezh31:
@Itay4 fixed, plz review changes

Scripts/script-Base64ListToFile.yml
```diff
@@ -13,12 +15,17 @@ script: |-
 
   # get the list
   res = demisto.executeCommand('getList', {'listName': demisto.args()['listname']})
+  if res['Type'] == entryTypes['error']:
+    return_error("error reading list %s from demisto" % demisto.args()['listname'])
+    sys.exit(0)
```
Itay4: Please remove `sys.exit(0)` as `return_error` already execute that

Scripts/script-Base64ListToFile.yml
```diff
@@ -13,12 +15,17 @@ script: |-
...
 
   # convert base64 file to binary file
   bin_file = base64.decodestring(res[0]['Contents'])
+  if demisto.args()['isZipFile'] == 'yes':
```
Itay4: This might fail if the argument is an empty string (I know that there is default value). I think it's better to use `demisto.args().get('isZipFile', 'no')`

erezh31: can't failed - there is a default

Scripts/script-Base64ListToFile.yml
erezh31: ok

PR #2554, opened 2018-12-02
WordTokenize - fix - support single result

erezh31:


meirwah:
@erezh31 can you add test for this case?

PR #2559, opened 2018-12-03
Removed Icebrg from testing - requires BD

amshamah419:
## Status
Ready

## Description
Login credentials are no longer valid, will require BD

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -791,7 +791,6 @@
       }
     ],
     "skipped_tests": {
-        "Icebrg Test": "Requires BD",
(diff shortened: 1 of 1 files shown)


Tests/conf.json
```diff
@@ -610,11 +610,6 @@
             "integrations": "Cisco Umbrella Investigate",
             "playbookID": "Cisco Umbrella Test"
         },
-        {
```
ronykoz: Dont remove it, just add to skip

amshamah419: Ah okay, thought it had to be removed to skip. Will fix now

PR #2567, opened 2018-12-03
dnstwist new redesigned integration

saharsabban:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/7336

## Description
New dnstwist integration for detecting domain variations

any

   - No

## Must have
- [ ] Tests - dnstwistTest playbook 
- [ ] Documentation https://github.com/demisto/etc/issues/14291
- [ ] Code Review

### DIFF
--- a/Integrations/integration-dnstwist.yml
+++ b/Integrations/integration-dnstwist.yml
@@ -1,0 +1,101 @@
+commonfields:
+  id: dnstwist
+  version: -1
+name: dnstwist
+display: dnstwist
+category: Data Enrichment & Threat Intelligence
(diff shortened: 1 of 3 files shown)


Integrations/integration-dnstwist.yml
```diff
@@ -0,0 +1,101 @@
...
+            "Domains" : domains
+        }
+
+        entry_result = {
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['json'],
+            'Contents': dnstwist_result,
+            'HumanReadable':md,
+            'ReadableContentsFormat': formats['markdown'],
+            'EntryContext': {'dnstwist.Domain' : ec}
```
erezh31: make it `dnstwist.Domain(val.Name == obj.Name)` for non dups

Integrations/integration-dnstwist.yml
```diff
@@ -0,0 +1,101 @@
...
+        for item in domainsContextList:
+            temp = {}
+            temp["Name"] = item["domain-name"]
+            if "IP Address" in item:
+                temp["IP"] = item["IP Address"]
+            domains.append(temp)
+
+        ec = {
+            "Name" : domainContext["domain-name"],
+            "IP" : domainContext["IP Address"],
```
erezh31: make it domainContext.get("IP Address"), safe if the key does not exist

Integrations/integration-dnstwist.yml
erezh31: I think you can also update the "domain" context key with the IP addreeses

PR #2569, opened 2018-12-03
Make master build to run relevant tests

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14305

## Description
Make Master build to run only relevant tests for last changes


ronykoz:
DONT MERGE, i think it would be better to do so on Wednesday once i'm in the office

PR #2570, opened 2018-12-03
fixing fetchFromInstance

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14225

## Description
Handling `invalid character \'[a-zA-Z]\' looking for beginning of value` exception

### DIFF
--- a/Scripts/script-FetchFromInstance.yml
+++ b/Scripts/script-FetchFromInstance.yml
@@ -3,6 +3,7 @@ commonfields:
   version: -1
 rolename: []
 name: FetchFromInstance
+fromversion: 4.0
 script: |
   import re
 
@@ -37,5 +38,6 @@ args:
   description: The name of the instance you want to fetch from
 scripttarget: 0
 runonce: false
+deprecated: true
 tests:
   - No test - cannot create a dummy integration inorder to test that
--- a/Scripts/script-FetchFromInstance.yml
+++ b/Scripts/script-FetchFromInstance.yml
@@ -1,7 +1,6 @@
 commonfields:
   id: FetchFromInstance
   version: -1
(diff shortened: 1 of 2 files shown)


Scripts/script-FetchFromInstance.yml
```diff
@@ -17,6 +19,8 @@ script: |
           contents = inc.get('Contents', '')
           if 'Error' in contents:
               raise Exception("Error occurred while fetching incidents from {0}: {1}".format(instance_name, contents))
+          if re.match("invalid character \'[a-zA-Z]\' looking for beginning of value", contents, re.IGNORECASE):
```
Itay4: Do we get this error only with letter characters ?

ronykoz: As far as I can recall, we can always add a fix for that

ronykoz: As far as I can recall, we can always add a fix for that

Scripts/script-FetchFromInstance.yml
```diff
@@ -17,6 +19,8 @@ script: |
...
+              raise Exception("Error occurred while fetching incidents from {0}: {1}".format(instance_name, contents))
```
Itay4: please use `return_error`

Scripts/script-FetchFromInstance.yml
```diff
@@ -1,9 +1,11 @@
 commonfields:
   id: FetchFromInstance
   version: -1
-rolename: []
 name: FetchFromInstance
+fromversion: 4.0
```
Itay4: please change to `4.0.0`

PR #2612, opened 2018-12-10
Qradar change contents paths

DeanArbel:
## Status
Ready

## Description
Reversion to the original output paths for the QRadar commands prior to content version 18.11.0.

   - Yes
       - Further details: Customers who will be affected are customers with version 18.11.0 and above, who created a custom script that uses QRadar OR updated the script after it stopped working for them following that content update.

## Must have
- [X] Tests
- [X] Documentation (with link to it)
- [x] Code Review

## Additional changes
Updated the in-house scripts to match the changes to the paths.


Integrations/integration-QRadar.yml
```diff
@@ -1139,6 +1146,10 @@ script:
       - HIDDEN
       - CLOSED
       description: The new status for the offense
+    - name: closing_reason_id
+      description: 'The id of a closing reason. You must provide a valid closing_reason_name
```
anara123: Then why we don't add those options to be PREDEFINED ?

PR #2614, opened 2018-12-10
Permit having the same id by version diff

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14465

## Description
@DeanArbel is creating a new playbook which will work from 4.0.0 only and it replaces another playbook so needs to have the same id

### DIFF
--- a/Tests/scripts/validate_files_structure.py
+++ b/Tests/scripts/validate_files_structure.py
@@ -271,7 +271,7 @@ def non_valid_versioning(id_to_file, my_id, id_list):
         versioning = id_obj[id]
         if my_id == id:
             if from_version > versioning['toversion']:
-                print_error("The ID {0} already exists, please update the file {1}".format(id, id_to_file[id]))
+                print_error("The ID {0} already exists, please update the file {1} or update the id_set.json file with the script update_id_set.py with the function update_id_set".format(id, id_to_file[id]))
                 has_duplicate = True
 
     return has_duplicate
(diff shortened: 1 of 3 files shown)


Tests/scripts/validate_files_structure.py
```diff
@@ -263,13 +263,31 @@ def oversize_image(file_path):
...
 
 
+def non_valid_versioning(id_to_file, my_id, id_list):
+    has_duplicate = False
+    from_version = get_from_version(id_to_file[my_id])
+    for id_obj in id_list:
+        id = id_obj.keys()[0]
+        versioning = id_obj[id]
+        if my_id == id:
+            if from_version > versioning['toversion']:
```
DeanArbel: Have you considered how this will work if `from_version` or `toversion` are None?

ronykoz: First of all they can't be None, because of the way we write the id_set.json file.
And we are going to use this function only if the condition id in ids_list is satisfied down below in line 290.
But maybe we should add a check that if the toversion is changed/added you need to update the id_set, what do you think?

yaakovi: instead of `beginning` you can use version `0`.
instead of `current` you can use version `99.99.99` (it will take some time to pass it).

consider comparing version like that: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python
since string compare will fail if we get the version 4.10

PR #2623, opened 2018-12-11
Init commit of Markdown to HTML tool

amshamah419:
## Status
Ready - Just need to publish docker image

## Description
Allows a user to upload a markdown file and get a converted HTML file in return.

   - No

### DIFF
--- a/Scripts/script-MD2HTML.yml
+++ b/Scripts/script-MD2HTML.yml
@@ -1,0 +1,27 @@
+commonfields:
+  id: MD2HTML
+  version: -1
+name: MD2HTML
+script: |-
+  from markdown2 import Markdown
+
+  markdowner = Markdown(extras=["tables"])
+  file_entry = demisto.args().get('file_id', '')
+  file_path = demisto.getFilePath(file_entry)['path']
+  demisto.results(file_path)
+  fopen = open(str(file_path), "r")
+  data = fopen.read()
+  res = markdowner.convert(data)
+  demisto.results(fileResult("md2htmlresult.html", res))
+
+type: python
+tags: []
+comment: Converts MarkDown files to HTML files
+enabled: true
+args:
+- name: file_id
+  required: true
+scripttarget: 0
+runonce: false
+dockerimage: demisto/markdown2
+runas: DBotWeakRole
--- a/DockerImageFiles/markdown2/Dockerfile
+++ b/DockerImageFiles/markdown2/Dockerfile
@@ -1,0 +1,5 @@
+FROM python:2.7
+
+COPY requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
--- a/DockerImageFiles/markdown2/create_docker_image.sh
+++ b/DockerImageFiles/markdown2/create_docker_image.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+_image_name=$1
+
+docker build -t $_image_name .
+
--- a/DockerImageFiles/markdown2/requirements.txt
+++ b/DockerImageFiles/markdown2/requirements.txt
@@ -1,0 +1,1 @@
+markdown2


Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
...
+  id: MD2HTML
+  version: -1
+name: MD2HTML
+script: |-
+  from markdown2 import Markdown
+
+  markdowner = Markdown(extras=["tables"])
+  file_entry = demisto.args().get('file_id', '')
+  file_path = demisto.getFilePath(file_entry)['path']
+  demisto.results(file_path)
```
yardensade: I believe it should be removed.

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
...
+  fopen = open(str(file_path), "r")
```
yardensade: Better handle IO exception in case file does not actually open

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
...
+  data = fopen.read()
+  res = markdowner.convert(data)
+  demisto.results(fileResult("md2htmlresult.html", res))
+
+type: python
+tags: []
+comment: Converts MarkDown files to HTML files
+enabled: true
+args:
+- name: file_id
```
yardensade: Demisto convention dictates "fileID", we do not use underscore in arg names.

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
...
+  from markdown2 import Markdown
+
+  markdowner = Markdown(extras=["tables"])
+  file_entry = demisto.args().get('file_id', '')
+  file_path = demisto.getFilePath(file_entry)['path']
+  demisto.results(file_path)
+  fopen = open(str(file_path), "r")
+  data = fopen.read()
+  res = markdowner.convert(data)
+  demisto.results(fileResult("md2htmlresult.html", res))
```
yardensade: Might want to rethink if you want the file name to be static like that, you should have access to the original file name, not sure.

Scripts/script-MD2HTML.yml
ronykoz: You are also not closing the file that you are reading, try using: with open(str(filepath), “r”) as file: ...

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
+commonfields:
+  id: MD2HTML
+  version: -1
+name: MD2HTML
+script: |-
+  from markdown2 import Markdown
+
+  markdowner = Markdown(extras=["tables"])
+  file_entry = demisto.args().get('file_id', '')
```
yardensade: changing the default to '' is redundant in that case since the argument is required.

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,29 @@
...
+  file_path = demisto.getFilePath(file_entry)['path']
```
yardensade: Normally better to verify the answer is actual dict then .get the 'path' so it wont raise exception in case things didn't go as expected.

ronykoz:
Agreed, this should have a test

Scripts/script-MD2HTML.yml
```diff
@@ -0,0 +1,37 @@
...
+  from markdown2 import Markdown
+
+  markdowner = Markdown(extras=["tables"])
+  file_entry = demisto.args().get('fileID')
+  file = demisto.getFilePath(file_entry)
+  if isinstance(file,dict):
+    file_path =file['path']
+    file_name = file['name']
+    raw_name = file_name.split(".")
+    formatted_name = raw_name+".html"
```
ronykoz: String equals list + string?

amshamah419: Forgot to add an index to the list

amshamah419:
Need docker image to be ready for this to pass tests

PR #2628, opened 2018-12-11
Fix commonServerPython date_to_timestamp

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14442

## Description
Fixes an issue with date_to_timestamp function commonServerPython, and added test to that function.

## Must have
- [X] Tests
- [ ] Code Review


TestPlaybooks/script-TestPyCommonServer.yml
```diff
@@ -337,6 +337,10 @@ script: |-
       assert camelize([{'chookity_bop': 'asdasd'}, {'ab_c': 'd e', 'fgh_ijk': 'lm', 'nop': 'qr_st'}], '_').__str__ == "[{u'ChookityBop': 'asdasd'}, {u'AbC': 'd e', u'Nop': 'qr_st', u'FghIjk': 'lm'}]"
       assert camelize({'ab_c': 'd e', 'fgh_ijk': 'lm', 'nop': 'qr_st'}, '_').__str__ == "{u'AbC': 'd e', u'Nop': 'qr_st', u'FghIjk': 'lm'}"
 
+  def test_date_to_timestamp():
```
ronykoz: I think that it should be mentioned in the TESTS dict in order to be tested, Also saw that CAMELIZE is not there so please add both of them

Tests/conf.json
```diff
@@ -2,6 +2,10 @@
     "testTimeout": 160,
     "testInterval": 20,
     "tests": [
+        {
+            "integrations": [],
```
ronykoz: Could you remove this line, not needed

TestPlaybooks/script-TestPyCommonServer.yml
DeanArbel: Done

Tests/conf.json
DeanArbel: Done

PR #2633, opened 2018-12-12
Fetch from instance - added context option

yardensade:
Added option to output the incidents context for testing purposes.

### DIFF
--- a/Scripts/script-FetchFromInstance.yml
+++ b/Scripts/script-FetchFromInstance.yml
@@ -86,4 +86,4 @@ runonce: false
 deprecated: true
 tests:
   - No test - cannot create a dummy integration inorder to test that
-releaseNotes: "-"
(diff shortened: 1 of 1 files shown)


Scripts/script-FetchFromInstance.yml
```diff
@@ -5,7 +5,11 @@ name: FetchFromInstance
 fromversion: 4.0.0
 script: |-
   import re
+  from distutils.util import strtobool
 
+  incidents_context = []
+  context = {}
+  add_to_context = bool(strtobool(demisto.args().get('add_to_context')))
```
ronykoz: Why do you need both strtobool and bool?

yardensade: Because it returns int and I prefer it to be more definitive.

PR #2634, opened 2018-12-12
findSimilarIncindets fix error handle

erezh31:


erezh31:
@meirwah plz review again, bug fixed

PR #2637, opened 2018-12-12
supporting multiple instances for integrations

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14514

## Description
Support multiple instances for each integration

### DIFF
--- a/Tests/test_content.py
+++ b/Tests/test_content.py
@@ -159,11 +159,22 @@ def create_result_files(failed_playbooks, skipped_integration, skipped_tests):
 
 def set_integration_params(demisto_api_key, integrations, secret_params):
     for integration in integrations:
-        integration_params = [item for item in secret_params if
-                              item["name"] == integration['name']]
+        is_integration_found = False
+        for item in secret_params:
+            if item["name"] == integration['name']:
+                if not integration['instance_name']:
+                    integration_params = item
+                    is_integration_found = True
+                    break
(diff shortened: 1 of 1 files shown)


Tests/test_content.py
```diff
@@ -159,34 +159,53 @@ def create_result_files(failed_playbooks, skipped_integration, skipped_tests):
...
             has_skipped_integration = True
 
-        if integration in nightly_integrations:
+        if integration_id in nightly_integrations:
             is_nightly_integration = True
 
         # string description
         integrations.append({
             'name': integration,
+            'instance_name': integration_instance,
```
yaakovi: this should be `integration_id`
```suggestion
            'instance_name': integration_id,
```

Tests/test_content.py
```diff
@@ -314,7 +331,12 @@ def main():
         if has_skipped_integration:
             continue
 
-        set_integration_params(demisto_api_key, integrations, secret_params)
+        are_params_set = set_integration_params(demisto_api_key, integrations, secret_params, instance_names_conf)
+        if not are_params_set:
+            print_error("{} Failed to run, Couldn't match instance name to the integration provided")
```
yaakovi: missing format

anara123:
@ronykoz the message should be more clear with instruction what to do, something like:
`There are 2 instances of VirusTotal. Please mention in test playbook TestFoo which instance to use "instance_v1" op "instance_v2"`

PR #2646, opened 2018-12-13
Base64Decode Script

compsecmonkey:
## Status
Ready

## Description
Adds a Base64Decode script to Demisto

Any

   - No

## Additional changes
None

### DIFF
--- a/Scripts/script-Base64Decode.yml
+++ b/Scripts/script-Base64Decode.yml
@@ -7,11 +7,14 @@ script: |
   input_string = demisto.args()["input"]
 
   try:
+      decoded = b64decode(input_string)
       demisto.results({
       "ContentsFormat": formats["text"],
       "Type": entryTypes["note"],
-      "Contents": b64decode(input_string)
-  })
+      "Contents": decoded,
+      "EntryContext": {"Base64.decoded": decoded}
+      })
+
   except TypeError:
       demisto.results({
           "ContentsFormat": formats["text"],
@@ -19,7 +22,8 @@ script: |
           "Contents": "String '{}' is an invalid base64 encoded string".format(input_string)
       })
 type: python
-tags: []
+tags:
+- Utility
 comment: Decode a base64 encoded string
 enabled: true
 args:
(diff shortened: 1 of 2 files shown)


Scripts/script-Base64Decode.yml
```diff
@@ -0,0 +1,31 @@
...
+          "Contents": "String '{}' is an invalid base64 encoded string".format(input_string)
+      })
+type: python
+tags: []
+comment: Decode a base64 encoded string
+enabled: true
+args:
+- name: input
+  required: true
+  default: true
```
Itay4: Not sure why you set the default value for `true`, but anyway I think there shouldn't be any default for this argument.

compsecmonkey: That is not setting the default value to "true" but marking the field as a default field. This was modeled after the already exsiting Base64Encode script.

compsecmonkey:
Good call on the EntryContext. That was overlooked and is now fixed.

Scripts/script-Base64Decode.yml
```diff
@@ -0,0 +1,38 @@
...
+  try:
+      decoded = b64decode(input_string)
+      demisto.results({
+      "ContentsFormat": formats["text"],
+      "Type": entryTypes["note"],
+      "Contents": decoded,
+      "EntryContext": {"Base64.decoded": decoded}
+      })
+
+  except TypeError:
```
Itay4: Please handle `UnicodeEncodeError` exceptions as well.

Scripts/script-CommonServerPython.yml
```diff
@@ -531,12 +531,8 @@ script: |-
       elif isinstance(data, list):
           string_list = []
           for d in data:
-              try:
-                  string_list.append(str(d))
-              except UnicodeEncodeError:
-                  string_list.append(d.encode('utf-8'))
-
-          return ',\n'.join(string_list)
+              string_list.append(str(d) if not isinstance(d, STRING_TYPES) else d)
```
Itay4: Not sure why you changed that, there is no call to `flattenCell` in the script

anara123: @compsecmonkey any update here?

PR #2647, opened 2018-12-14
Microsoft Graph Mail Integration - CMU Project

Pekdz:
## Status
Ready

## Description
Integration of Microsoft Graph Mail service as a part of CMU 18Fall practicum project. Four parts:
1. Message
2. Attachment
3. Folder
4. Rule

## Known bug
All mail rule commands cannot work now, because the integration requires additional MailboxSettings.Read/MailboxSettings.ReadWrite permissions when doing OAuth authentication. 
Need to add these two permissions in backend OAuth service which working for graph-security integration.

   - No

## Must have
- [Tests](https://github.com/Pekdz/content/blob/microsoft-graph-mail/TestPlaybooks/playbook-MSGraph-Mail-Test.yml)
- [Documentation](https://github.com/Pekdz/content/blob/joe_doc/Documentation/docs/Microsoft_Graph_Mail-documantation.md)


Itay4:
will be implemented as a fix for https://github.com/demisto/etc/issues/14046

PR #2654, opened 2018-12-17
Mimecast test - changed tasks order

yardensade:
in such a way refresh token endpoint won't be used too many times in the same moment.

## Status
Ready

### DIFF
--- a/TestPlaybooks/playbook-Mimecast_Test.yml
+++ b/TestPlaybooks/playbook-Mimecast_Test.yml
@@ -45,7 +45,7 @@ tasks:
       {
         "position": {
           "x": 480,
-          "y": 1420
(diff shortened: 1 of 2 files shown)


TestPlaybooks/playbook-Mimecast_Test.yml
```diff
@@ -1227,15 +1227,15 @@ tasks:
       add_to_context:
         simple: "true"
       brand_name:
-        simple: MimecastV2
+        simple: MimecastV2_copy2
```
ronykoz: THIS IS NO Copy

yardensade: YOU ARE CORRECT SIR

PR #2656, opened 2018-12-17
MaxMind Integration

amshamah419:
## Status
In Progress

## Related Issues
fixes: demisto/etc#12786

## Description
Added integration for MaxMind. Uses the !ip command.

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/TestPlaybooks/playbook-MaxMind-test.yml
+++ b/TestPlaybooks/playbook-MaxMind-test.yml
@@ -1,6 +1,6 @@
 id: MaxMind Test
 version: -1
-name: MaxMind Test Playbook
+name: MaxMind Test
 starttaskid: "0"
 tasks:
   "0":
--- a/Tests/id_set.json
+++ b/Tests/id_set.json
@@ -15155,7 +15155,7 @@
         }, 
         {
             "MaxMind Test": {
-                "name": "MaxMind Test Playbook", 
+                "name": "MaxMind Test", 
                 "implementing_scripts": [
                     "DeleteContext"
                 ], 


Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,285 @@
...
+            headers=headers,
+            verify=INSECURE,
+            auth=HTTPBasicAuth(ACCOUNT_ID, APIKEY)
+        )
+        if r.status_code is not 200:
+            return_error('Error in API call to MaxMind [%d] - %s' % (r.status_code, r.reason))
+        return r
+
+
+    # Allows nested keys to be accesible
```
ronykoz: You can just add that as docstring

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,285 @@
...
+            if 'isp' in traits:
+                hr['ISP'] = traits['isp']
+                ip_ec['ISP'] = traits['isp']
+        return hr, ip_ec, res_json
+
+
+
+
+    ''' FUNCTIONS '''
+    def geo_ip(query):
```
ronykoz: Change to get_geo_ip

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,285 @@
...
+            return_error('Error in API call to MaxMind [%d] - %s' % (r.status_code, r.reason))
+        return r
+
+
+    # Allows nested keys to be accesible
+    def makehash():
+        return collections.defaultdict(makehash)
+
+
+    def format_results(raw):
```
ronykoz: All the fields in the response are optional??

amshamah419: Yeah, there are three different service levels which return different amounts of data depending on the service. Of the data returned, it may or may not contain everything that's available.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,285 @@
...
+    def http_request(method, query, MODE):
+        if method is 'GET':
+            headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'application/json'
+            }
+        r = requests.request(
+            method,
+            BASE_URL + MODE +'/'+ query,
+            headers=headers,
```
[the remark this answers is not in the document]
amshamah419: For this integration, it's always get. I'll set it as always get.

Tests/id_set.json
```diff
@@ -11420,6 +11420,42 @@
...
+                "fromversion": "3.5.0", 
+                "commands": [
+                    "ip"
+                ], 
+                "tests": [
+                    "MaxMind Test"
+                ]
+            }
+        }, 
+        {
```
ronykoz: Please remove these redundancies and merge from master, it should not create duplicates

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,289 @@
...
+                ip_ec['UserType'] = traits['user_type']
+            if 'domain' in traits:
+                hr['Domain'] = traits['domain']
+                ip_ec['Domain'] = traits['domain']
+            if 'is_anonymous' in traits:
+                hr['Is Anonymous'] = traits['is_anonymous']
+                ip_ec['Anonymous'] = traits['is_anonymous']
+            if 'is_hosting_provider' in traits:
+                hr['Is Hosting Provider'] = traits['is_hosting_provider']
+                ip_ec['Host'] = traits['is_hosting_provider']
```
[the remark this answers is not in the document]
amshamah419: Not sure I follow. hr is the name of the dict holding the human readable output. "Traits" is a reference to a location in the response JSON called traits.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,289 @@
...
+        'Is TOR Exit Node',
+        'Is Hosting Provider',
+        'Is Anonymous']
+
+
+
+
+    '''HELPER FUNCTIONS'''
+    def http_request(query, MODE):
+        headers = {
```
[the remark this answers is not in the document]
amshamah419: It's an argument being passed through to the helper function, not a constant.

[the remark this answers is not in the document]
ronykoz: you just pass it to the requests.requests, and don't make any changes to it at any step of the code, lets make it a constant

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,291 @@
...
+        return raw
+
+
+    def geo_ip_command():
+        ip_query = demisto.args().get('ip')
+        raw = get_geo_ip(ip_query)
+        hr, ip_ec, res_json, maxmind_ec = format_results(raw)
+        ec = ({
+          'IP(val.Address && val.Address == obj.Address)': ip_ec,
+          'MaxMind': maxmind_ec
```
ronykoz: They don't have a single param on which we can make base a DT query?

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,289 @@
...
+
+    ''' Allows nested keys to be accesible '''
+    def makehash():
+        return collections.defaultdict(makehash)
+
+
+    def format_results(raw):
+        res_json = raw.json()
+        hr = makehash()
+        ec = makehash()
```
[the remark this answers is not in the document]
amshamah419: I removed the helper function and instead am using the defaultdict() function.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+        LOG.print_log(False)
+        return_error(e.message)
+  type: python
+  commands:
+  - name: ip
+    arguments:
+    - name: ip
+      required: true
+      description: IP Address to be queried
+    outputs:
```
ronykoz: Not sure if the order matters, but lets keep that ordered, IP related than MaxMind related

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+        res_json = get_geo_ip(ip_query)
+        hr, ip_ec, maxmind_ec = format_results(res_json)
+        ec = ({
+          'IP(val.Address && val.Address == obj.Address)': ip_ec,
+          'MaxMind(val.Address && val.Address == obj.Address)': maxmind_ec
+        })
+        demisto.results({
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['markdown'],
+            'Contents': res_json,
```
ronykoz: I remember you wasn't sure about this one, so try to use the `raw-respone=true` to check it out

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+            auth=HTTPBasicAuth(ACCOUNT_ID, APIKEY)
+        )
+        if r.status_code is not 200:
+            return_error('Error in API call to MaxMind, got status code - {} and a reason: {}'.format(r.status_code, r.reason))
+        return r
+
+
+    def format_results(res_json):
+        hr = defaultdict()
+        ec = defaultdict()
```
ronykoz: why do we need this param? It's not being used anywhere

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+        maxmind_ec = defaultdict(lambda: defaultdict(int))
```
ronykoz: what are you trying to accomplish with the lambda?

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Trust any certificate (unsecure)
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Service Version
+  name: mode
```
ronykoz: A thought about this param, shouldn't we make it as an optional argument while calling the command?
Also not sure if it's ok to expose such an option for an IP command, but if you agree lets check that out

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: res_json? It's used in every one of the if statements

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: It nests defaultdict within a dict. Basically creates a dictionary within a dictionary so when you call on a nested key, it creates the key value pair as opposed to throwing a key error.

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: It was originally that way as that was my thought too, but speaking with @michalgold , we decided that setting the value as a param was more effective since we didn't want to create additional arguments for the !ip command.

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: Maxmind, this works as expected. I think you're thinking about the alexa integration with the xml output.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+            return_error('Error in API call to MaxMind, got status code - {} and a reason: {}'.format(r.status_code, r.reason))
+        return r
+
+
+    def format_results(res_json):
+        hr = defaultdict()
+        ec = defaultdict()
+        maxmind_ec = defaultdict(lambda: defaultdict(int))
+        ip_ec = defaultdict(lambda: defaultdict(int))
+        if 'continent' in res_json:
```
yuvalbenshalom: this is not about this lin specifically, but about the following context creation.
Is it really the case that you are not sure if you get continent/city/location/registered_country[..] and so on?
are those nested ifs necessary?
because if they will always exist as a group or won't exist at all, we can lose most of the ifs and make it much more readable context creation.
and create the sub dictionaries statically

amshamah419: Unfortunately yes. In many cases, they will report only partial information and when that happens, it throws a key error.

The response that they return, it is not standardised in the sense that a call for a specific mode will always return the country, city, or location results. In my testing, the API returns what ever information they have and exclude certain values depending on the mode.

For example, if you submit an IP using the country mode, you'll receive the country (if it exists) and not city or registered country since they are only delivered for different modes.

Integrations/integration-MaxMind_GeoIP2.yml
ronykoz: i meant the `ec`

Integrations/integration-MaxMind_GeoIP2.yml
ronykoz: oh yeah, my bad

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: Ah gotcha. I removed it.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,289 @@
...
+        if 'country' in res_json:
+            country = res_json['country']
+            hr['Country'] = country['names']['en']
+            maxmind_ec['Geo']['Country'] = country['names']['en']
+            ip_ec['Geo']['Country'] = country['names']['en']
+        if 'location' in res_json:
+            location = res_json['location']
+            ip_ec['Geo']['Location'] = str(location['latitude'])+', '+str(location['longitude'])
+            maxmind_ec['Geo']['Location'] = str(location['latitude'])+', '+str(location['longitude'])
+            demisto.results({
```
ronykoz: lets extract this to another function, it took me a sec to realise what are you doing here so better to call the function `create_map_entry` or something of that sort.

amshamah419:
> @amshamah419 noticed we don't set the hostname param to the IP, and it's in out context guidelines so need to see if something here could use for that

Host is a value that just tells if the IP address is associated with a hosting service. So if you hit a GoDaddy IP address, it would return true. The API doesn't return a hostname.

Integrations/integration-MaxMind_GeoIP2.yml
```diff
@@ -0,0 +1,293 @@
...
+  type: 8
+  required: false
+- display: Trust any certificate (unsecure)
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Service Version
+  name: mode
+  defaultvalue: city
```
ronykoz: Forgot to say that we should probably state the options available

TestPlaybooks/playbook-MaxMind-test.yml
```diff
@@ -1,6 +1,6 @@
 id: MaxMind Test
```
ronykoz: id should be the same as the name

Integrations/integration-MaxMind_GeoIP2.yml
amshamah419: In documentation yes?

PR #2661, opened 2018-12-17
Skipped mimecast due to too many random errors from their service 

yardensade:
on different endpoints/commands.
Including error 500/error 405 and several others which started only few days ago.


yardensade:
https://github.com/demisto/etc/issues/14593

PR #2667, opened 2018-12-18
Add CIRCL integration

itayronen:
## Status
Ready

## Related Issues
Fixes https://github.com/demisto/etc/issues/13723

## Description
Added an integration for CIRCL (https://www.circl.lu)

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Itay4:
@kirbles19 please review descriptions

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+    USE_SSL = True if demisto.params().get('insecure') else False
+    IS_USING_PROXY = demisto.params()['isUsingProxy']
+
+    if not IS_USING_PROXY:
+        del os.environ['HTTP_PROXY']
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    lastTimeKey = 'time_last'
```
Itay4: please change to Python convention: `last_time_key`

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+            'Usage': extension['extendedKeyUsage'],
+            'Distribution': extension['crlDistributionPoints'],
+            'Issuer': info['issuer'],
+            'Time': info['not_before'],
+            'Subject': info['subject'],
+            'Key': info['key'],
+            'Pem': data['pem'],
+            'Seen': data.get('icsi', {}).get('times_seen'),
+        }
+
```
Itay4: Add command name to LOG

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+    try:
+        args = demisto.args()
+
+        if demisto.command() == 'test-module':
+            responce = dnsGet('test.com')
+            demisto.results('ok')
+
+        if demisto.command() == 'circl-dns-get':
+            demisto.results(dnsGet(args.get('queryValue')))
```
Itay4: Please use `demisto.results()` in the functions, and here only call the functions

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+    - contextPath: CIRCLdns.Query.Record.LastTime
+      description: DNS record last recorded time
+      type: date
+    description: Get DNS records for your query value from CIRCL's Passive DNS.
+  - name: circl-ssl-list-certificates
+    arguments:
+    - name: queryValue
+      required: true
+      description: IP address or CIDR block
+    outputs:
```
Itay4: @mayagoldb Shouldn't we output to context standard as well?

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+      type: number
+    - contextPath: CIRCLssl.Certificate.IP
+      description: IP address associated to the certificate
+      type: string
+    description: Query a certificate value to get all associated addresses
+  - name: circl-ssl-get-certificate
+    arguments:
+    - name: certificate
+      required: true
+      description: SHA1 fingerprint of a certificate
```
Itay4: Please change to `SHA-1`

TestPlaybooks/playbook-CIRCL.yml
```diff
@@ -0,0 +1,788 @@
+id: CirclIntegrationTest
```
Itay4: Please add a screenshot of the test

Itay4:
General comments:
1. Please use Python conventions, that is non-capital letters and underscores, e.g. `lastTimeKey` -> `last_time_key`
2. You should split each function into 2: command (which implements the logic) and request (which does only the HTTP request) 
3. Please use our integration template, can be found in almost every Python integration as an example.
4. Please create an `http_request()` function which will do the requests and don't use `requests.get` in every function

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,324 @@
...
+  defaultvalue: https://www.circl.lu
+  type: 0
+  required: true
+- display: Username
+  name: credentials
+  defaultvalue: ""
+  type: 9
+  required: false
+- display: Use system proxy settings
+  name: isUsingProxy
```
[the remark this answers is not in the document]
itayronen: This name is only for the integration code, therefore I don't see a reason to be consistent with other integrations.
This name is a better name "clean code"-wise.
If you still think I should change it, I will.
Please comment.

Integrations/integration-CIRCL.yml
itayronen: Thats what I saw from other integrations.
I think its nice to keep the functions pure from demisto commands (when possible).
Any way I'll change it a little and tell me what you think.

itayronen:
Fixed the resolved comments. Now I will take a look and fix the general comments.

Integrations/integration-CIRCL.yml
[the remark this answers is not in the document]
Itay4: I understand what you're saying, but I see it a coding convention, so I prefer to have same proxy variable name in all integrations.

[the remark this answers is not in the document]
itayronen: Ok I will change it

Integrations/integration-CIRCL.yml
itayronen: I think its better now. Im resolving this one because its outdated. Please take a look at the new code.

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,343 @@
...
+    USE_SSL = True if demisto.params().get('insecure') else False
+    IS_USING_PROXY = True if demisto.params().get('proxy') else False
+
+    if not IS_USING_PROXY:
+        del os.environ['HTTP_PROXY']
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    last_time_key = 'time_last'
```
Itay4: It's a constant, please change to `LAST_TIME_KEY`

meirwah:
@itayronen your test seem to fail:
image

plz fix

itayronen:
It looks like the API was down (the 502 proxy error).
The other tests check for errors by design and should not fail the playbook.
It works locally now (when the API is up).
I will rerun.

itayronen:
Rerun worked now when the API is up. I hope the API will be stable otherwise it will fail other builds...
Alex is working on docs. https://github.com/demisto/etc/issues/14739
@Itay4 Is there something for me to fix/complete?

meirwah:
@itayronen going forward if you see this as unstable - make this test nightly

meirwah:
@Itay4 can you review fixes?

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,371 @@
...
+    IS_USING_PROXY = True if demisto.params().get('proxy') else False
+    LAST_TIME_KEY = 'time_last'
+
+    if not IS_USING_PROXY:
+        del os.environ['HTTP_PROXY']
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    def http_request(method, url, bad_request_msg = None):
```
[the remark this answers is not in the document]
itayronen: Actually it isn't used anymore. I will remove it.

Integrations/integration-CIRCL.yml
```diff
@@ -0,0 +1,371 @@
...
+    def dns_get(url):
+        responce = http_dns_get(url)
+
+        results = list(map(lambda line: json.loads(line), responce.text.splitlines()))
+        results = merge_by_rdata(results)
+
+        records = []
+
+        for result in results:
+            records.append(create_dns_record_context(result))
```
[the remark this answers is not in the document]
itayronen: When I built it, I talked to Asaf and he recommended to leave the time values as they are, because we don't have consistent convention for times, and they might be used for other requests for the same API (although in this API there is currently no filter by time).

I can change the context, or only the human readable. What do you prefer?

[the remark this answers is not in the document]
Itay4: So we do have a standard for times, and it is not keeping them as timestamps.
I'd prefer to change them to human readable time string please

PR #2683, opened 2018-12-20
Fix for ProofpointDecodeURL key error

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#14618

## Description
Gave a better description to the error when a key is not found

   - No

### DIFF
--- a/Scripts/script-ProofpointDecodeURL.yml
+++ b/Scripts/script-ProofpointDecodeURL.yml
@@ -61,3 +61,4 @@ outputs:
   description: Decoded URLs
 scripttarget: 0
 dockerimage: demisto/python3
+releaseNotes: "Gave more descriptive error for when url is not found in the query"
--- a/Scripts/script-ProofpointDecodeURL.yml
+++ b/Scripts/script-ProofpointDecodeURL.yml
@@ -13,40 +13,39 @@ script: |
   valid_versions = ['v1', 'v2']
 
   def find_version(url):
-       match = re.search(r'https://urldefense.proofpoint.com/(?P<version>v[0-9])/', url)
-       return match.group('version')
+      match = re.search(r'https://urldefense.proofpoint.com/(?P<version>v[0-9])/', url)
+      return match.group('version')
 
   def ppdecode(url, return_all=False):
       try:
           version = find_version(url)
       except AttributeError:
-          raise RuntimeError('This does not appear to be a valid proofpoint url: {}'.format(url))
+          return_error('This does not appear to be a valid proofpoint url: {}'.format(url))
       else:
           if version not in valid_versions:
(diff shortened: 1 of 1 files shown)


Itay4:
@amshamah419 Please post screenshot of an error message after this change.

amshamah419:
screen shot 2018-12-23 at 13 47 26

Scripts/script-ProofpointDecodeURL.yml
```diff
@@ -13,23 +13,26 @@ script: |
...
-      if sys.version_info[0] < 3:
-          translated_url = query_components['u'][0].translate(string.maketrans('-_', '%/'))
+      if 'u' in query_components:
+          if sys.version_info[0] < 3:
+              translated_url = query_components['u'][0].translate(string.maketrans('-_', '%/'))
+          else:
+              translated_url = query_components['u'][0].translate(str.maketrans('-_', '%/'))
       else:
-          translated_url = query_components['u'][0].translate(str.maketrans('-_', '%/'))
+          return_error('A URL was not detected in the query.')
```
gshriki: This means that if previously we returned something - now we'll fail the script, correct? Is this intentional?

anara123: I think once we failed on unexpected error, and now it will fail with nicer message - this is the whole issue is about

amshamah419: It's intentional. The original issue was that when it failed, it didn't give the user a good indication as to why.

The decode url function looks for the query string in the url and then uri-decodes it. When there is no query string, it fails. This was a way to make it more of a "The query string you provided has no url in it" versus "The script just failed on a key error"

amshamah419:
@anara123 There is a playbook for it called "ProofpointDecodeURL-Test". The test itself does not check for urls missing the query string so I think thats where we didn't catch it. I'm not too sure how we would have the playbook test for a failure. We could have the script return a value of url's captured. Then test for a string we know has no query string and see if it catches it in the playbook. Could look kind of messy though.

Scripts/script-ProofpointDecodeURL.yml
anara123: try to use the script with bad url (without u) and use extend-context to set the error in the context, this way you will be able to check the error.

amshamah419: Just checked this. The script itself is using a regex to grab the url (specifically looking for the "u") and then decodes the query string. Since this is all done on our side and doesn't hit proofpoint at all, there's no context to extend.

anara123:
@amshamah419 you can check `continue on error` at that task and extend context so that the error will be set out to context and in the next task you can check if the output is the same you expected. Not that nasty

PR #2685, opened 2018-12-23
EWS v2: Fixed predefined command argument

BenJoParadise:
## Status
Ready

## Related Issues
fixes: link to the issue

## Description
ews-delete-items has a `delete-type` argument which only accepts one of 3 predefined values.
If no one of them is given, it throws an exception.
Changed that argument type to predefined, and put the values in the predefined list.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Additional changes


BenJoParadise:
This will break BC in playbooks with `ews-delete-items` tasks

PR #2688, opened 2018-12-23
Updating nightly trigger

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14704

## Description
Update Content config.yml file to work with workflows and use it to trigger nightly builds


ronykoz:
@Bars92 it seems to be running now, i guess that the triggering part won't work until we merge this thing

PR #2694, opened 2018-12-25
ServiceNow - Handle empty responses and missing fields

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/14721

## Description
Fix a bug where `closed_by` field is not present and the integration throws a key error

   - No

- [X] Tests
- [ ] Code Review


liorblob:
@Itay4 handled more cases of missing fields and empty responses, have a look when you can

Itay4:
Looks good 👍

PR #2703, opened 2018-12-26
Add Active Directory as BYOI

noaco:
## Status
In Progress - 
todo:

- [x]  add test playbook
- [x] add logo
- [ ] create documentation for BYOI
- [x] add docker image

## Related Issues
fixes: https://github.com/demisto/etc/issues/12742

## Description
Active Directory as BYOI, using python ldap3 library.
Guidelines were to implement existing ad commands and scripts.


Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1205 @@
...
+            'Type': entryTypes['error'],
+            'Contents': message,
+            'ContentsFormat': formats['text']
+        }
+
+    def initialize_server():
+        """
+        uses the instance configuration to initialize the LDAP server
+        """
+        if SECURE_CONNECTION == "SSL":
```
glicht: Note that in current implementation ssl and tls don't refer to protocols. The naming that was chosen is not very clear. ssl is referring to an ldap server that listens on an ssl port (usually 636). tls refers to an ldap server listening on a plain tcp socket but supporting starttls (see:  https://ldap3.readthedocs.io/bind.html?highlight=starttls#starttls). Additionally, there should be the "insecure" option.

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1205 @@
...
+            'Contents': "Deleted object with dn {}".format(demisto.args().get('user-dn'))
+        }
+        demisto.results(demisto_entry)
+
+
+    ''' EXECUTION '''
+
+    # intialize connection to LDAP server with NTLM authentication
+    server = initialize_server()
+    conn = Connection(server, user=SERVER_IP + '\\' + USERNAME, password=PASSWORD, authentication=NTLM)
```
glicht: Let's make the authentication method a configuration parameter with a default of NTLM

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1205 @@
...
+
+    try:
+        # open socket and bind to server
+        if not conn.bind():
+            demisto.results(error_entry('Failed to bind to server'))
```
glicht: Try to get more details why the bind failed. Maybe `conn.result` can provide more insight.

Integrations/integration-Active_Directory_Query.yml
noaco: We had a discussion about this issue, since in the previous integration 'port', 'connection type' and 'certificate validation' are all configurable, but have dependencies between them.

[here](https://ldap3.readthedocs.io/ssltls.html) initializing the server object with a TLS object is used for certificate validation (the TLS object itself needs to be initialized with correct parameters. needs more thinking..) while start_tls() is used to start a secure connection over clear-text (usually port 389).

Integrations/integration-Active_Directory_Query.yml
noaco: ok
basic authentication will require the user DN an password.

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1319 @@
...
+        authenticate user credentials while initializing connection wiith AD server
+        verify base DN is configured correctly
+    '''
+
+    server = initialize_server()
+
+    conn = None
+    if NTLM_AUTH:
+        # intialize connection to LDAP server with NTLM authentication
+        conn = Connection(server, user=SERVER_IP + '\\' + USERNAME, password=PASSWORD, authentication=NTLM)
```
glicht: Using: `SERVER_IP + \ + USERNAME` can be problematic. You should perform this only if the username doesn't contain a `\`. Also I think it is possible and common to specify the username as `user@domain` (need to verify).

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1321 @@
...
+
+configuration:
+- display: Server IP (e.g. 192.168.0.1)
+  name: server_ip
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Server port. If this is not specified, the port will be configured by the
+    connection type default port.
+  name: port
```
anara123: remove the port, make it one parameter - server_url

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1321 @@
...
+- display: Secure Connection
+  name: secure_connection
+  defaultvalue: SSL
+  type: 15
+  required: true
+  options:
+  - None
+  - SSL
+- display: verify server certificate (with SSL connection)
+  name: tls
```
anara123: remove the unnecessary params , because either way it will be a new integration

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1291 @@
...
+    try:
+        # open socket and bind to server
+        if not conn.bind():
+            message = "Failed to bind to server. Please validate the credentials configured correctly.\n{}".format(conn.result)
+            return_error(message)
+    except LDAPSocketOpenError as e:
+    # except Exception as e:
+        demisto.info(e.message)
+        message = "Failed to access LDAP server."
+        if TLS_VERIFY_CERTIFICATE:
```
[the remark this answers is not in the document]
noaco: yes its a bug, I forgot to remove this after changing the configuration.

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1291 @@
...
+description: Active Direcory Query integration enables accessing and managing Active
+  Directory objects (users,contacts and computers).
+detaileddescription: |+
+  Searching Active directory is done with paging, when each page size pulled is set by the 'page size'
+  parameter configured in the instance settings.
+
+configuration:
+- display: Server IP (e.g. 192.168.0.1)
```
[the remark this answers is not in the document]
noaco: A port can be defined using the 'Port' parameter when initiating the Server. If this parameter is not defined, the default port number will be configured depending on the security level.

I was asked to leave only server url, and remove the port from the configuration. I can re-add this. The port cannot be specified as part of the url.

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1320 @@
...
+            attributes['telephoneNumber'] = args['telephone-number']
+        if args.get('title'):
+            attributes['title'] = args['title']
+
+        # set user custome attributes
+        if custome_attributes:
+            try:
+                custome_attributes = json.loads(custome_attributes)
+            except Exception as e:
+                demisto.info(e.message)
```
glicht:
```suggestion
                demisto.info(str(e))
```
`message` of exception has been removed in python 3.0. See: https://www.python.org/dev/peps/pep-0352/#retracted-ideas

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1320 @@
...
+        object_classes = ["top", "person", "organizationalPerson", "contact"]
+        contact_dn = args.get('contact-dn')
+
+        # set contact attributes
+        attributes = {}
+        if args.get('custom-attributes'):
+            try:
+                attributes = json.loads(args['custom-attributes'])
+            except Exception as e:
+                demisto.info(e.message)
```
glicht:
```suggestion
                demisto.info(str(e))
```
`message` has been removed in python 3: https://www.python.org/dev/peps/pep-0352/#retracted-ideas

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1320 @@
...
+        modification = {}
+        modification[attribute_name] = [('MODIFY_REPLACE', attribute_value)]
+
+        # modify user
+        modify_object(dn, modification)
+
+        demisto_entry = {
+            'ContentsFormat': formats['text'],
+            'Type': entryTypes['note'],
+            'Contents': "Updated user's {} to ".format(attribute_name, attribute_value)
```
glicht:
```suggestion
            'Contents': "Updated user's {} to {} ".format(attribute_name, attribute_value)
```
missing format

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1320 @@
...
+        modification = {}
+        modification[args.get('attribute-name')] = [('MODIFY_REPLACE', args.get('attribute-value'))]
+
+        # modify
+        modify_object(contact_dn, modification)
+
+        demisto_entry = {
+            'ContentsFormat': formats['text'],
+            'Type': entryTypes['note'],
+            'Contents': "Updated contact's {} to ".format(args.get('attribute-name'), args.get('attribute-value'))
```
glicht:
```suggestion
            'Contents': "Updated contact's {} to: {} ".format(args.get('attribute-name'), args.get('attribute-value'))
```
missing format

Integrations/integration-Active_Directory_Query.yml
```diff
@@ -0,0 +1,1320 @@
...
+    ]
+
+    ''' HELPER FUNCTIONS '''
+
+    def initialize_server():
+        """
+        uses the instance configuration to initialize the LDAP server
+        """
+        if SECURE_CONNECTION == "None" and not UNSECURE:
+            raise Exception("When using unsecure connection you must check 'Trust any certificate' option")
```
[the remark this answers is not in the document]
noaco: validating server certificate requires initialisation with ssl

noaco:
@glicht please review

glicht:
@ronykoz please look at package_creator.py . I've put in support for utf-8 python files and also added verification step that merge worked correctly.

ronykoz:
@glicht looks good.

anara123:
fix https://github.com/demisto/etc/issues/14078

PR #2714, opened 2018-12-30
test playbook for duo

glicht:
## Status
Ready

## Related Issues
fixes: demisto/etc#14746

## Description
Test playbook for DUO integration. Tests the pre-auth stage for various user types.

Any version supporting DUO 

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/TestPlaybooks/playbook-DUO-Test.yml
+++ b/TestPlaybooks/playbook-DUO-Test.yml
@@ -1,4 +1,4 @@
-id: duo_test
+id: DUO Test Playbook
 version: -1
 name: DUO Test Playbook
 description: Tests that DUO pre-auth returns expected results for various user types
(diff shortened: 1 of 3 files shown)


TestPlaybooks/playbook-DUO-Test.yml
```diff
@@ -0,0 +1,342 @@
+id: duo_test
```
ronykoz: Just a convention we are trying to make, btw it should be automatically generated to be identical as the name, so you are probably not pulled in the server/web-client:
```suggestion
id: DUO Test Playbook
```

ronykoz:
You should pull from master, Dorsha added the change for getting the logs from the Demisto instance.
You can also just force merge if you like, this changes are not affected by the Destroy instance so it should be fine

PR #2721, opened 2018-12-31
fix cveReputation script

meirwah:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14783

## Description
compare is wrong


meirwah:
I take it back, the integration itself is very unstable, sometimes returns result , sometimes not

PR #2725, opened 2019-01-01
make SsdeepRep script compatible as indicator reputation script

meirwah:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14794


meirwah:
@erezh31 plz review

PR #2728, opened 2019-01-01
Missing headers fix & error check for command result

cantordemisto:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14743

## Description
Fixed a bug where the client got "Missing headers param" error & added an error check for when executing the 'splunk-search' command.

   - No

## Must have
- [x] Tests
- [x] Code Review


Itay4:
👍

PR #2729, opened 2019-01-01
Update nightly build trigger time

ronykoz:
Update the time in which we trigger the nightly build.
Now It will be 4 AM, which means the build will finish around 7AM


ronykoz:
I need a force merge on this one please

PR #2735, opened 2019-01-02
add dbotscore outputs path

jochman:
added outputPaths.dbotscore to commonserver

### DIFF
--- a/Scripts/script-CommonServer.yml
+++ b/Scripts/script-CommonServer.yml
@@ -1411,7 +1411,7 @@ script: |-
       domain: 'Domain(val.Name && val.Name === obj.Name)',
       cve: 'CVE(val.ID && val.ID === obj.ID)',
       email: 'Account.Email(val.Address && val.Address === obj.Address)',
-      dbotscore: 'DBotScore(val.Indicator === obj.Indicator && val.Vendor === obj.Vendor)',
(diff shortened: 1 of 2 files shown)


Scripts/script-CommonServer.yml
```diff
@@ -1410,7 +1410,8 @@ script: |-
       url: 'URL(val.Data && val.Data === obj.Data)',
       domain: 'Domain(val.Name && val.Name === obj.Name)',
       cve: 'CVE(val.ID && val.ID === obj.ID)',
-      email: 'Account.Email(val.Address && val.Address === obj.Address)'
+      email: 'Account.Email(val.Address && val.Address === obj.Address)',
+      dbotscore: 'DBotScore(val.Indicator === obj.Indicator && val.Vendor === obj.Vendor)',
```
Itay4: Please verify it exists first by `val.Indicator && val.Indicator ===...`

aviadl: DBotScore path should be:
`DBotScore` and that is it

ronykoz:
The id_set.json change is because of a problematic way i write the id_set, should be ok once Anar will merge my PR

Scripts/script-CommonServer.yml
jochman: It's already exists in CommonServerPython, I just added it to CommonServer. shouldn't we use it?

PR #2756, opened 2019-01-04
Add flake8 checks

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14832

## Description
Add flake8 checks


.circleci/config.yml
```diff
@@ -16,12 +16,14 @@ jobs:
               sudo mkdir -p -m 777 $CIRCLE_ARTIFACTS
               chmod +x ./Tests/scripts/*
               chmod +x ./Tests/lastest_server_build_scripts/*
+              sudo pip install flake8
```
glicht: We should install this as part of the docker image we use. We want a reproducible build. Also possible to use pipenv.

ronykoz: Yes you are right, but until the docker image won't be ready lets add it this way, i'll try to make the docker image tomorrow

PR #2759, opened 2019-01-06
fixes for multi level dependency

ronykoz:


ronykoz:
fixes https://github.com/demisto/etc/issues/14841

PR #2771, opened 2019-01-06
update nightly trigger

ronykoz:
updating nightly trigger


Integrations/integration-AbuseDB.yml
```diff
@@ -462,6 +462,6 @@ script:
       type: string
     description: Returns a list of report categories from AbuseIPDB
   runonce: false
+releaseNotes: "fixed context issues. added AbuseIPDB-PopulateIndicators script"
```
meirwah: why is this related?

PR #2774, opened 2019-01-07
Default playbook fix

avidan-demisto:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14732

## Description
Replace deprecated CloseInvestigation script task at the end of the Default playbook with the builtin closeInvestigation script.

4.1.0

   - No

## Must have
- [ ] Tests
- [x] Code Review

### DIFF
--- a/Playbooks/playbook-default.yml
+++ b/Playbooks/playbook-default.yml
@@ -1,7 +1,7 @@
 id: playbook0
 version: -1
 name: Default
-fromversion: 4.1.0
+fromversion: 3.5.0
 description: Default playbook, Enrich indicators in incident using one or more integrations
 starttaskid: "0"
 tasks:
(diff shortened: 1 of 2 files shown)


Playbooks/playbook-default.yml
```diff
@@ -1,7 +1,7 @@
 id: playbook0
 version: -1
 name: Default
-fromversion: 3.5.0
+fromversion: 4.1.0
```
Itay4: Please change back to `3.5.0`

Playbooks/playbook-default.yml
```diff
@@ -51,7 +51,7 @@ tasks:
     conditions:
     - label: "yes"
       condition:
-      - - operator: string.isEqual
+      - - operator: isEqualString
```
Itay4: Please change back to `string.isEqual`

PR #2800, opened 2019-01-09
Parsecsv nullbyte

avidan-demisto:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14873

## Description
Updated ParseCSV script to replace nullbyte chars with empty strings.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Additional changes
Created an additional test playbook - ParseCSVnullbytesTest - to test that the ParseCSV script is correctly replacing nullbyte chars with empty strings.

### DIFF
--- a/Scripts/script-ParseCSV.yml
+++ b/Scripts/script-ParseCSV.yml
@@ -218,4 +218,4 @@ runas: DBotWeakRole
 releaseNotes: "Updated to replace null byte character with the empty string"
 tests:
   - TestParseCSV
-  - ParseCSVnullbytesTest - tests that script replaces nullbyte chars with empty strings
+  - ParseCSVnullbytesTest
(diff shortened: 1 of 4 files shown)


Scripts/script-ParseCSV.yml
```diff
@@ -7,10 +7,11 @@ script: |
   import socket
 
   def unicode_dict_reader(utf8_data, **kwargs):
-      csv_reader = csv.DictReader(utf8_data, **kwargs)
+      csv_reader = csv.DictReader((line.replace('\0', '') for line in utf8_data), **kwargs)
```
Itay4: Please add a short comment here to describe why this string replacement is required, something like `removing null bytes..`

PR #2801, opened 2019-01-09
predict phishing typo fix

erezh31:


erezh31:
No,  the automation is not in the playbook

PR #2808, opened 2019-01-12
Fix threshold sla

Bars92:
## Status
Ready

## Description
Threshold should be taken from conf rather then using static 72 hours.

4.1.0

   - No


idovandijk:
How will it look when having threshold set to 0 if there is no server configuration for that (if it's even possible that that will happen)?

Bars92:
@idovandijk yes possible. it will be default value 72 (hours)

PR #2809, opened 2019-01-13
Fixing typo

Shellyber:
## Status
Ready

   - No


Shellyber:
Added a test for the integration 
image

ronykoz:
@Shellyber why do you test this integration twice? please remove the duplicate in the conf.json file

Shellyber:
It exists only once in conf.json, but there where two instances in the build, I removed one of them

PR #2811, opened 2019-01-13
ResolveShortenedURL - Removed regex check before the request.

EdenOhana:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14986

## Description
Remove regex as it is not relevant and will not be updated with new services of url shorten sites every time there is new one

   - No

## Must have
- [ ] Tests - tested locally, works on new services.
- [ ] Code Review

### DIFF
--- a/Scripts/script-ResolveShortenedURL.yml
+++ b/Scripts/script-ResolveShortenedURL.yml
@@ -56,3 +56,4 @@ outputs:
   description: Shortened URL
 scripttarget: 0
 releaseNotes: "-"
+tests: "No test"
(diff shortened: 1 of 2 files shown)


gshriki:
@EdenOhana - do we have test playbook for this?

EdenOhana:
No that I know of, I didn't use one as this a small change (just removing regex check and using same service to unshorten), checked it locally that the script works on new services and tested random strings and numbers, it returns "Provided URL could not be un-shortened"

PR #2815, opened 2019-01-14
Added support for Email and Domain indicators

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14509

## Description
Added support for Email and Domain indicators for `IsMaliciousIndicatorFound` script.

   - No

## Must have
- [X] Tests
- [ ] Code Review

### DIFF
--- a/Scripts/script-IsMaliciousIndicatorFound.yml
+++ b/Scripts/script-IsMaliciousIndicatorFound.yml
@@ -6,6 +6,8 @@ script: |-
   var files = dq(invContext,'File(val.Malicious)');
   var urls = dq(invContext,'URL.Malicious');
   var ips = dq(invContext,'IP(val.Malicious)');
+  var domains = dq(invContext,'Domain(val.Malicious)');
+  var emails = dq(invContext,'Account.Email(val.Malicious)');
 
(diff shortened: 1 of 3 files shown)


Scripts/script-IsMaliciousIndicatorFound.yml
```diff
@@ -17,6 +19,12 @@ script: |-
       if (!urls || urls.length === 0) {
           urls = dq(invContext,"DBotScore(val.Type.toLowerCase()=='url' && val.Score>1)");
       }
+      if (!domains || domains.length === 0) {
+          domains = dq(invContext,"DBotScore(val.Type.toLowerCase()=='domain' && val.Score>1)");
```
amshamah419: Do they want DT in the DBot score? I thought we tried to avoid it?

DeanArbel:
> it seems you have deleted test cases, but haven't added any... how does that make sense?

I removed the use of `verifyContext` as it was deprecated and isn't always behaving as expected... But you are right, as I should add in another phase to test out the new feature

DeanArbel:
As per @meirwah I expanded the test to test out the new `domains` feature

PR #2828, opened 2019-01-15
Amazon Alexa

amshamah419:
## Status
Ready

## Related Issues
Addresses demisto/etc#13824

## Description
Uses Amazon's Alexa rankings as an indicator of whether or not a domain is trustworthy.
Threshold is user configurable.

## Related PRs
#2387 

branch | PR
------ | ------
alexa | #2387

   - No

## Must have
- [ x ] Tests
- [ ] Code Review


Tests/id_set.json
```diff
@@ -15302,6 +15314,17 @@
...
             }
+        }, 
+        {
+            "Alexa Test Playbook": {
+                "name": "Alexa Test Playbook", 
+                "implementing_scripts": [
+                    "DeleteContext"
+                ], 
+                "implementing_commands": [
+                    "awis-info"
```
ronykoz: I have a feeling that you forgot to update the test to use the domain command

amshamah419: Forgot to re-run update_id_set

ronykoz: make sure you run it in the pre-commit

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+description: Alexa provides website ranking information that can be useful in determining
+  if the domain in question has a strong web presence.
+detaileddescription: To use this integration, simply call ```!awis-info```. !awis
```
ronykoz: update this description

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+  accepts "url" as a parameter to query.
+configuration:
+- display: Sensitivity threshold for configuring which domains are suspicious versus
```
ronykoz: Maybe a link to an explanation about the threshold?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+                    'ContentsFormat': formats['markdown'],
+                    'Contents': 'The Alexa rank of {} is {}'.format(domain, rank),
+                    'HumanReadable': 'The Alexa rank of {} is {}'.format(domain, rank),
+                    'EntryContext': ec
+                    })
+
+
+    """EXECUTION BLOCK"""
+    try:
+        if demisto.command() == 'test-module':
```
ronykoz: lets extract the logic of the test to a function

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+            domain = 'google.com'
+            root = awis(domain)
+            domain_res = root.find("SD").attrib['HOST']
+            rank = root.find("SD[0]/POPULARITY").attrib['TEXT']
+            if rank == '1':
+                demisto.results('ok')
+            else:
+                demisto.results('Alexa failed to run')
+            sys.exit(0)
```
ronykoz: since when are we sys.exit(0) after a command?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+            'Domain(val.Name && val.Name == obj.Name)': dom_ec,
+            'DBotScore': dbot_ec
+        }
+        demisto.results({
+                    'Type': entryTypes['note'],
+                    'ContentsFormat': formats['markdown'],
+                    'Contents': 'The Alexa rank of {} is {}'.format(domain, rank),
+                    'HumanReadable': 'The Alexa rank of {} is {}'.format(domain, rank),
+                    'EntryContext': ec
+                    })
```
ronykoz: bad indent

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+
+    """MAIN FUNCTIONS"""
+    def awis(domain):
+        xml = urllib.urlopen('http://data.alexa.com/data?cli=10&dat=s&url=%s'%domain).read()
+        root = ET.fromstring(xml)
+        return root
+
+
+    """COMMAND FUNCTIONS"""
+    def awis_command():
```
ronykoz: we changed the name of the command to domain, so lets change the function names as well, awis function as well

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+    from collections import defaultdict
+
+
+    """GLOBAL VARIABLES/CONSTANTS"""
+    THRESHOLD = int(demisto.params().get('threshold'))
+
+
+    """MAIN FUNCTIONS"""
+    def awis(domain):
+        xml = urllib.urlopen('http://data.alexa.com/data?cli=10&dat=s&url=%s'%domain).read()
```
ronykoz:
```suggestion
        xml = urllib.urlopen('http://data.alexa.com/data?cli=10&dat=s&url={}'.format(domain)).read()
```

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+        return root
+
+
+    """COMMAND FUNCTIONS"""
+    def awis_command():
+        domain = demisto.args().get('domain')
+        root = awis(domain)
+        dbot_ec = defaultdict()
+        dom_ec = defaultdict()
+        # Wrapping in Try Except since when no results are found, endpoint returns 404
```
ronykoz: Can't we handle that from the response of the API?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            domain_res = root.find("SD").attrib['HOST']
+            rank = root.find("SD[0]/POPULARITY").attrib['TEXT']
+            if int(rank) > THRESHOLD:
+                dbot_score = 2
+            else:
```
ronykoz: So either unknown or suspicious? no other options for us?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+                dbot_score = 0
+        except:
+            rank = 'Unknown'
+            dbot_score = 2
```
ronykoz: why not 0?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+        dom_ec['Name'] = domain
+        dbot_ec['Score'] = dbot_score
```
ronykoz: just do something like:
```
dbot_ec = {
    'Score': dbot_score,
    'Vendor': 'Alexa',
.....
}
```

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+        dbot_ec['Rank'] = rank
+        dbot_ec['Domain'] = domain
+        ec = {
+            'Domain(val.Name && val.Name == obj.Name)': dom_ec,
+            'DBotScore': dbot_ec
+        }
+        demisto.results({
+                    'Type': entryTypes['note'],
+                    'ContentsFormat': formats['markdown'],
+                    'Contents': 'The Alexa rank of {} is {}'.format(domain, rank),
```
ronykoz: shouldn't we return the response here?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,119 @@
...
+                    'HumanReadable': 'The Alexa rank of {} is {}'.format(domain, rank),
```
ronykoz: I would add `marked as suspicous/good/bad domain` or just add `while the threshold is X`

Integrations/integration-Alexa.yml
mayagoldb: Alexa provides information about website traffic statistics, so can't really establish if a sight is malicious/benign based on Alexa (the rank doesn't establish maliciousness, only "popularity"). We will output the unknown scoring and the suspicious scoring with very high and configurable threshold.

Integrations/integration-Alexa.yml
amshamah419: When a domain is not found, the service returns a 404 error which is why we wrap in the try except. After, we catch it, we give the scoring for it.

If a site is not found, it means that the ranking is so low, that it doesn't make the alexa list/they haven't crawled the site. We treat it as suspicious because sites fall into two categories here.

1. They are brand new - If the site is new, alexa has not collected data on it.
2. They are not generating web traffic - means the site is obscure and possibly stale, or the owner is trying to hide the site. (It's fairly easy to get a site ranked on Alexa).

Integrations/integration-Alexa.yml
amshamah419: The issue is that it really isn't an API as much as it is an xml page. When the domain is searched for and not found, it returns a page with nothing on it. So when we attempt to parse the xml, it crashes out.

Integrations/integration-Alexa.yml
amshamah419: This will be addressed in documentation.

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,137 @@
...
+        LOG.print_log(False)
+        return_error(e.message)
+  type: python
+  commands:
+  - name: domain
+    arguments:
+    - name: domain
+      required: true
+      description: domain to search
+    outputs:
```
yuvalbenshalom: i think we should have a rank output other than dot score.

yuvalbenshalom:
@ronykoz please approve so we can marge. I @amshamah419 answers to your comments are satisfactory for me.

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,145 @@
...
+        xml = get_xml(domain)
+        root = ET.fromstring(xml)
+        return root
+
+
+    """COMMAND FUNCTIONS"""
+    def alexa_domain_command():
+        domain = demisto.args().get('domain')
+        dbot_score_text = 'unknown'
+        xml = get_xml(domain)
```
ronykoz: lets update the command function to that it will return the xml.
I'm just concerned that we may call the API twice and get different result on each call

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,145 @@
...
+    def alexa_domain(domain):
+        xml = get_xml(domain)
+        root = ET.fromstring(xml)
+        return root
+
+
+    """COMMAND FUNCTIONS"""
+    def alexa_domain_command():
+        domain = demisto.args().get('domain')
+        dbot_score_text = 'unknown'
```
ronykoz: lets also move this to the "else" statement, feels off here

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,145 @@
...
+        dom_ec = {'Name': domain}
+        dbot_ec = {
+            'Score': dbot_score,
+            'Vendor': 'Alexa',
+            'Domain': domain,
+        }
+        ec = {
+                'Domain(val.Name && val.Name == obj.Name)': dom_ec,
+                'DBotScore': dbot_ec,
+                'Alexa': {
```
ronykoz: what about dt query here?

Integrations/integration-Alexa.yml
```diff
@@ -0,0 +1,145 @@
...
+                    'Domain': {
+                        'Data': domain,
```
ronykoz: is data really a suitable name? I think Name would be better

Integrations/integration-Alexa.yml
amshamah419: I originally had it doing that, but for some reason when I return it as part of the ```alexa_domain``` function (using a tuple return) it crashes circleci.

PR #2833, opened 2019-01-16
CrowdStrike integration reimplementation 

DeanArbel:
## Status
Done

## Related Issues
Implements: https://github.com/demisto/etc/issues/11887

## Description
The Crowstrike Streaming v2 is intended to be a python based implementation of the original integration, which was written in go.

   - No

## Screenshots:
image

## Must have
- [x] Tests
- [x] Code Review


Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+            'Index': 2
+        },
+    ]
+
+    ''' HELPER FUNCTIONS '''
+
+
+    def http_request(method, url_suffix, params=None, data=None, headers=HEADERS, safe=False):
+        """
+            A wrapper for requests lib to send our requests and handle requests and responses better.
```
yuvalbenshalom: @DeanArbel did we start to use sphinx doc strings or was it your initiative?

yuvalbenshalom:
@DeanArbel we will need to divide it to separate files as all integrations.

Integrations/integration-CrowdStrikeFalcon.yml
DeanArbel: This method specifically is part of the default integration provided by the plugin, but it was my initiative to add docstring (reST format, as this is what is used in commonServerPython and what is generated in the IDE)

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+            :return: Access token
+        """
+        body = {
+            'client_id': USERNAME,
+            'client_secret': PASSWORD
+        }
+        headers = {
+            'Authorization': HEADERS['Authorization']
+        }
+        token_res = http_request('POST', '/oauth2/token', data=body, headers=headers, safe=True)
```
yuvalbenshalom: we need to check if there is any meaning for insecure mode in this integration. it is clearly disabled in the get token part. and maybe it should never be enabled.

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+        if not token_res:
+            err_msg = 'Authorization Error: User has no authorization to create a token. Please make sure you entered the' \
+                      ' credentials correctly.'
+            if demisto.command() != 'fetch-incidents':
```
yuvalbenshalom: avoid calling demisto.command() in any place other than the commands switch case. this creates a very confusing code.
error handling that is specific to a command should be handled in the command business process function.
I think the correct solution is to create a general http request function that will be used by all commands but not the request token itself, and it will call the get_token func in the beginning.
comment-1

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+        raw_res = lift_host_containment(ids)
+        hr = "Containment has been lift off host {}".format(str(ids)[1:-1])
+        return create_entry_object(contents=raw_res, hr=hr)
+
+
+    ''' COMMANDS MANAGER / SWITCH PANEL '''
+
+    LOG('Command being called is {}'.format(demisto.command()))
+
+    TOKEN = get_token()
```
yuvalbenshalom: better to move to a general HTTP request function. see `comment-1`

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+
+
+    def enrich_ioc_dict_with_ids(ioc_dict):
+        """
+            Enriches the provided ioc_dict with IoC ID
+            :param ioc_dict: IoC dict transformed using the SEARCH_IOC_KEY_MAP
+            :return: ioc_dict with its ID key:value updated
+        """
+        for ioc in ioc_dict:
+            ioc['ID'] = '{0}:{1}'.format(ioc.get('Type'), ioc.get('Value'))
```
yuvalbenshalom: please use named place holders when formatting more than one variable in a string.

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+            'site_name': str(args.get('site_name', '')).split(',')
+        }
+        url_filter = ''.format(str(args.get('filter', '')))
+        for k, arg in input_arg_dict.items():
+            if arg:
+                if type(arg) is list:
+                    arg_filter = ''
+                    for arg_elem in arg:
+                        if arg_elem:
+                            arg_filter = "{0}:'{1}'".format('{0},{1}'.format(arg_filter, k) if arg_filter else k, arg_elem)
```
yuvalbenshalom: please use named place holders when formatting more than one variable in a string.

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+                    if arg_filter:
+                        url_filter = "{0}{1}".format(url_filter + '+' if url_filter else '', arg_filter)
+                else:
+                    # All args should be a list. this is a fallback
+                    url_filter = "{0}+{1}:'{2}'".format(url_filter, k, arg)
```
yuvalbenshalom: please use named place holders when formatting more than one variable in a string.

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+        # Get the last fetch time, if exists
+        last_fetch = last_run.get('first_behavior_time')
+
+        # Handle first time fetch, fetch incidents retroactively
+        if last_fetch is None:
+            last_fetch, _ = parse_date_range(FETCH_TIME, date_format='%Y-%m-%dT%H:%M:%SZ')
+        last_fetch_timestamp = date_to_timestamp(last_fetch, date_format='%Y-%m-%dT%H:%M:%SZ')
+        fetch_query = demisto.params().get('fetch_query')
+        if fetch_query:
+            fetch_query = "first_behavior:>'{0}'+{1}".format(last_fetch, fetch_query)
```
yuvalbenshalom: please use named place holders when formatting more than one variable in a string.

Integrations/integration-CrowdStrikeFalcon.yml
yuvalbenshalom: OK

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+  display: Trust any certificate (not secure)
+  name: insecure
+  required: false
+  type: 8
+- display: Use system proxy
+  name: proxy
+  required: false
+  type: 8
+description: The Falcon Query API (formerly known as the Falcon Firehose API)
+detaileddescription: 'To get an API key please contact the crowdstrike support: support@crowdstrike.com'
```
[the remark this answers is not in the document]
DeanArbel: 1. How about The CrowdStrike Falcon OAuth 2 API (formerly known as the Falcon Firehose API). Enables fetching and resolving detections, searching devices, getting behaviors by ID, containing hosts, and lift host containment.
2. You are correct, this description is for the OG Query API. The right description should be:
```To get an The API client ID and secret please contact the crowdstrike support: support@crowdstrike.com```

File: Integrations/integration-CrowdStrikeFalcon.yml [hunk 10]
```diff
@@ -0,0 +1,1095 @@
...
+        demisto.params()['url']
+    # Should we use SSL
+    USE_SSL = not demisto.params().get('insecure', False)
+    # How many time before the first fetch to retrieve incidents
+    FETCH_TIME = demisto.params().get('fetch_time', '3 days')
+    # Headers to be sent in requests
+    HEADERS = {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+        'Authorization': 'Basic {}'.format(base64.b64encode('{0}:{1}'.format(USERNAME, PASSWORD).encode('utf-8')).decode())
```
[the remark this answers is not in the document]
DeanArbel: 1. Will change it in all parts of code this comment is relevant (great example!)
2. This was originally to encode it into bytestring for base64 encoder, I'm not sure this behavior was resolved in python 3 - let me check it out.

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+                data=data,
+                headers=headers,
+            )
+        except requests.exceptions.RequestException as e:
+            return_error('Error in connection to the server. Please make sure you entered the URL correctly.')
+        # Handle error responses gracefully
+        if res.status_code not in {200, 201}:
+            if safe:
+                return None
+            return_error('Error in API call [{0}] - {1}'.format(res.status_code, res.reason))
```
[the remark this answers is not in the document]
DeanArbel: Will do, by the way this is part of the default integration, so might be a good thing to raise it up in the next summary

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+        """
+            Converts a severity string to DBot score representation
+
+            :type severity: ``str``
+            :param severity: String representation of a severity
+
+            :return: DBot score representation of the severity
+            :rtype ``int``
+        """
+        if severity in ('Critical', 'High'):
```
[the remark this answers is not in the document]
DeanArbel: yes

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+
+            :rtype: ``str``
+            :return: Token
+        """
+        now = datetime.now()
+        ctx = demisto.getIntegrationContext()
+        if ctx:
+            time_delta = now - datetime.fromtimestamp(ctx.get('time'))
+            passed_mins = time_delta.seconds / 60
+            auth_token = get_token_request() if passed_mins >= TOKEN_LIFE_TIME else ctx.get('auth_token')
```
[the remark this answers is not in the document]
DeanArbel: 1. Seems you are right, weird this has worked so far and didn't cause any issues...
2. Done. added an helper function, and added some comments to make things clearer.

Integrations/integration-CrowdStrikeFalcon.yml
DeanArbel: this is actually an argument I added, which means
 `If set to true will return None in case of http error`.

I can change the name to `err_flag` or something of that like

Integrations/integration-CrowdStrikeFalcon.yml
```diff
@@ -0,0 +1,1095 @@
...
+        hr = "Containment has been lift off host {}".format(str(ids)[1:-1])
+        return create_entry_object(contents=raw_res, hr=hr)
+
+
+    ''' COMMANDS MANAGER / SWITCH PANEL '''
+
+    LOG('Command being called is {}'.format(demisto.command()))
+
+    TOKEN = get_token()
+
```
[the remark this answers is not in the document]
DeanArbel: Done, and added it as part of the solution for comment-1

PR #2835, opened 2019-01-16
Added a new script to send email on SLA breach

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14785

## Description
- Added a new script to send email on SLA breach
- Added the new `update_id_set.py` file that Rony fixed with me on another branch I worked on, that should now accomodate for line breaks in the regex, and deal with CRLF / LF and more.

## Screenshots
image

4.1.0+

   - No

### DIFF
--- a/Scripts/script-SendEmailOnSLABreach.yml
+++ b/Scripts/script-SendEmailOnSLABreach.yml
@@ -1,0 +1,68 @@
+commonfields:
+  id: SendEmailOnSLABreach
+  version: -1
+name: SendEmailOnSLABreach
+script: |-
+  """
+  This script is used to send an email about a breached SLA. The script, by default, sends the email to the assignee of the
+  incident. The email is sent using the send-email command, using all enabled integrations that support the command.
+  """
+
+  def get_owner_email():
+      owner_username = demisto.incidents()[0].get("owner")
+      if owner_username:
+          try:
+              owner_info = demisto.executeCommand('getUserByUsername', {"username":owner_username})[0]
+              owner_email = owner_info.get("EntryContext").get("UserByUsername").get("email")
+              return owner_email
+          except Exception, ex:
+              demisto.results
+              ({
+                  "Type" : entryTypes["error"],
+                  "ContentsFormat" : formats["text"],
+                  "Contents" : "Could not retrieve user email. Maybe the user has no associated email to it.\
+              Error: {}".format(ex)
+
+              })
+              return
+
+  def get_subject():
+      incident_name = demisto.incidents()[0].get("name")
+      incident_id = demisto.incidents()[0].get("id")
+      subject = "SLA Breached in incident \"{}\" #{}".format(incident_name, incident_id)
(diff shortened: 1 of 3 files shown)


Scripts/script-SendEmailOnSLABreach.yml
```diff
@@ -0,0 +1,68 @@
...
+  version: -1
+name: SendEmailOnSLABreach
+script: |-
+  """
+  This script is used to send an email about a breached SLA. The script, by default, sends the email to the assignee of the
+  incident. The email is sent using the send-email command, using all enabled integrations that support the command.
+  """
+
+  def get_owner_email():
+      owner_username = demisto.incidents()[0].get("owner")
```
JonathanMeler: What will happened if `owner` is not defined?

Scripts/script-SendEmailOnSLABreach.yml
```diff
@@ -0,0 +1,68 @@
...
+      if owner_username:
+          try:
+              owner_info = demisto.executeCommand('getUserByUsername', {"username":owner_username})[0]
+              owner_email = owner_info.get("EntryContext").get("UserByUsername").get("email")
```
JonathanMeler: I think you can use `demisto.get("EC.UBU.Email")`

JonathanMeler: Also, can this be a list?

Scripts/script-SendEmailOnSLABreach.yml
idovandijk: The script would fail because there would be no one to send the email to. But the script is meant to send the email to the assignee, so what do we want to do if there is no one assigned? I can make it do nothing but then they wouldn't know that there is no one to inform of the SLA breach

Scripts/script-SendEmailOnSLABreach.yml
idovandijk: What is that command you mentioned? 
Regarding email being a list - I really don't know. Isn't there one email configured per user?

Scripts/script-SendEmailOnSLABreach.yml
idovandijk: I should add indicative error for this case

idovandijk:
@kirbles19 Please tell me when you're available to go over documentation (basically just script description and the email that is being sent)

Scripts/script-SendEmailOnSLABreach.yml
idovandijk: Using a single "get" won't work as there are multiple nested dictionaries and you have to use "get" for each one

Scripts/script-SendEmailOnSLABreach.yml
idovandijk: I added an indicative error message but keep in mind that because of the design of error handling of scripts that run on SLA breaches, the error won't be shown

PR #2836, opened 2019-01-16
Panorama/FW no port error

bakatzir:
@shaniacht1 agreed. so i'll add a check to see if the port param is empty, and if so will return an error to the user. i'll just add that i can't delete this field(legacy we had 2 separate fields beforehand so adding it as a required is also not a good practice) and will add an example.

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14912

screen shot 2019-01-16 at 15 59 28

### DIFF
--- a/Integrations/integration-Panorama.yml
+++ b/Integrations/integration-Panorama.yml
@@ -2730,7 +2730,7 @@ script:
       type: string
     description: Get push status for a configuration
   runonce: false
-releaseNotes: "Handle port configuration for the instance errors gracefully."
+releaseNotes: "Handle port configuration for the instance errors gracefully"
 tests:
 - palo_alto_firewall_test_pb
 - palo_alto_panorama_test_pb
--- a/Integrations/integration-Panorama.yml
+++ b/Integrations/integration-Panorama.yml
@@ -2723,7 +2723,7 @@ script:
       type: string
     description: Get push status for a configuration
   runonce: false
-  releaseNotes: "Handle port configuration for the instance erros gracefully."
+releaseNotes: "Handle port configuration for the instance errors gracefully."
 tests:
 - palo_alto_firewall_test_pb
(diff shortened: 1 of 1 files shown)


shaniacht1:
@bakatzir why not add something like that?
```if !devicegroup and !vsys:
            VSYS = vsys1
```
?

bakatzir:
@shaniacht1 I prefer not to, since if the user want to have a panorama instance, but forgot to configure a device-group, we'll automatically configure it as a FW with `vsys1`, which he doesn't necessarily have

Integrations/integration-Panorama.yml
```diff
@@ -91,17 +93,11 @@ script:
...
-    if not DEVICE_GROUP and not VSYS:
-        return_error('Set a vsys for Firewall or Device Group for Panorama')
-    # setting security xpath relevant to FW or panorama managment
-    XPATH_SECURITY_RULES = None
-    if DEVICE_GROUP:
-        XPATH_SECURITY_RULES = "/config/devices/entry/device-group/entry[@name=\'" + DEVICE_GROUP + "\']/"
-    else:
-        XPATH_SECURITY_RULES = "/config/devices/entry/vsys/entry[@name=\'" + VSYS + "\']/rulebase/security/rules/entry"
-
+            return_error('Can not configure both vsys and Device Group. Set vsys for Firewall, set Device Group for Panorama')
```
shaniacht1: Firewall => firewall
Device Group => Device group/device group
Can not => Cannot

Integrations/integration-Panorama.yml
```diff
@@ -91,17 +93,11 @@ script:
...
+        if not DEVICE_GROUP and not VSYS:
+            return_error('Set a vsys for Firewall or Device Group for Panorama')
```
shaniacht1: same as above

PR #2842, opened 2019-01-17
don't delete the incidents after test'

anara123:
## Status
Ready

## Description
We should keep the incidents even after test done.
For debugging purposes it is better to keep those incidents in Demisto
No

## Must have
- [ ] Code Review

### DIFF
--- a/Tests/test_integration.py
+++ b/Tests/test_integration.py
@@ -287,7 +287,7 @@ def test_integration(client, integrations, playbook_id, options={}):
     if test_pass:
         # close incident
         __close_incident(client, incident)
-        
+
         # delete integration instance
         __delete_integrations_instances(client, instance_ids)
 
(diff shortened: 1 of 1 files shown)


anara123:
@ronykoz There should be. I think it just updating the status or something

ronykoz:
@anara123 saw that you've failed on some PEP8 errors, i'll add the flake8 checks to the pre-commit in the support packages PR, and will try to look for auto formatting for simple problems like the one you had

ronykoz:
Looks good to me, just not sure what are all the `Tag1` or `Tag2` incidents there

ronykoz:
@anara123 do we want this?

anara123:
yes

yardensade:
What are we waiting for here?

anara123:
it is not working properly. So I am closing for now

PR #2870, opened 2019-01-22
Jask fix

IkaDemisto:
## Status
Ready

## Description
Added type check to *signal_to_readable* function and edited the *Jask_Test* playbook. 
In *Jask_Test* playbook changed filter dates in tasks and added Test Fetch Incident task.
Removed *Jask_Test* from skipped test in conf.json.

   - No

## Must have
- [x] Tests
- [x] Code Review

## Dependencies
Integrations:
- [x] Jask

Playbooks:
- [x] Jask_Test

### DIFF
--- a/TestPlaybooks/playbook-Jask_Test.yml
+++ b/TestPlaybooks/playbook-Jask_Test.yml
@@ -25,6 +25,7 @@ tasks:
         }
       }
     note: false
+    timertriggers: []
   "1":
     id: "1"
     taskid: ca722df1-ae4c-4833-82b4-612a31a83782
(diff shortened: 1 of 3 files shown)


TestPlaybooks/playbook-Jask_Test.yml
```diff
@@ -520,13 +536,42 @@ tasks:
...
+    task:
+      id: 106d1f90-6828-4d80-8755-44f1a2a251d0
+      version: -1
+      name: Test Fetch Incidents
+      scriptName: FetchFromInstance
+      type: regular
+      iscommand: false
+      brand: ""
+    scriptarguments:
+      add_to_context: {}
```
yardensade: Would be nice if we verify the context at least on some level to make sure we actually pull some data

PR #2873, opened 2019-01-23
add Cut transformer

omercnet:
## Status
Ready

## Description
A Cut filter that (unlike Split) can return multiple fields, similar to the unix cut command

## Screenshots
image

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
None

## Additional changes

fixes https://github.com/demisto/etc/issues/15192


Scripts/script-Cut.yml
```diff
@@ -0,0 +1,36 @@
+commonfields:
+  id: cae96081-c034-46c1-81f6-b4d0a52dc51c
```
asafshen: change ID to be same as name 
also set version to `-1`

Scripts/script-Cut.yml
```diff
@@ -0,0 +1,36 @@
...
+
+  data = value.split(delim)
+  fields = [int(_) for _ in fields.split(',')]
+
+  demisto.results(delim.join([str(data[i-1]) for i in fields]))
+type: python
+tags:
+- transformer
+- string
+comment: Cut a string by delimiter and return specific fields
```
asafshen: lets consult with Alex about this comment.

I would give an example (e.g `Cut value=<val> delimiter=<del> ... => <result-example>`)

Scripts/script-Cut.yml
```diff
@@ -0,0 +1,36 @@
+commonfields:
+  id: cae96081-c034-46c1-81f6-b4d0a52dc51c
+  version: 38
+name: Cut
+script: |-
+  args = demisto.args()
+
+  value = args.get('value')
+  fields = args.get('fields')
+  delim = args.get('delimiter')
```
asafshen: I would verify input correctness:
- types (`value` and `delimiter` should be string, `fields` should be csv numbers)
- indexes in range

Scripts/script-Cut.yml
```diff
@@ -0,0 +1,36 @@
...
+args:
+- name: value
+  required: true
+  description: Value to split
+- name: delimiter
+  required: true
+  description: Delimiter to cut the string by
+- name: fields
+  required: true
+  description: 'Field numbers (e.g.: 1,5,7)'
```
asafshen: `Comma separated indexes to include in the result`?

Scripts/script-Cut.yml
```diff
@@ -0,0 +1,36 @@
...
+comment: Cut a string by delimiter and return specific fields
+enabled: true
+args:
+- name: value
+  required: true
+  description: Value to split
+- name: delimiter
+  required: true
+  description: Delimiter to cut the string by
+- name: fields
```
asafshen: I would name `indices` and not `fields`

omercnet: I'm borrowing the `cut` unix command syntax and parameters that people might be familiar with

Scripts/script-Cut.yml
omercnet: fixed

anara123:
@omercnet can you also add a test?
https://github.com/demisto/internal-content/tree/master/documentation/tests

anara123:
@omercnet on the second thought, forget about the test. We will take care of it

PR #2875, opened 2019-01-23
Deprecate StringContains

liorblob:
## Status
Ready


anara123:
Wait. I think changing the name can break BC. I think in playbooks we use names of scripts and not ids. Please make sure we don't break BC.

anara123:
@liorblob what is the status here? Have you checked that changing the name don't brake BC?

liorblob:
@anara123 @Itay4 Couldn't test, any special character like `(Deprecated)` or `- Deprecated` does not work, should I leave it as is?

PR #2877, opened 2019-01-23
making the content creator leave the yml in the correct order

ronykoz:
minor fix to the content-creator


glicht:
@ronykoz do you think we can refactor this a bit so I can run easily `merge-script-package` for a single package from the command line and see what I get. I think this will help developer better understand what is the final result and also test it.

ronykoz:
@glicht Sure i can split that into a different script so that the content-creator will call it, but no real reason for this, because the plugin allows you to export to Demisto, so i'm not 100% sure why would we want to do it

PR #2890, opened 2019-01-27
misp-add-sighting command

jochman:
added add_sighting command

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/13928
fixes: https://github.com/demisto/etc/issues/15273

## Description
added misp-add-sighting command and a better test connection in test-module

## Screenshots
output:
image

   - No

## Must have
- [x] Tests
- [ ] Code Review


Integrations/integration-MISP_V2.yml
```diff
@@ -1642,7 +1678,23 @@ script:
...
+    - name: type
+      required: true
+      auto: PREDEFINED
+      predefined:
+      - sighting
+      - false-positive
+      - expiration
+      description: Type of sighting to add
+    - name: id
+      description: ID of attribute to add sighting to
```
Itay4: Please add to the description where this `ID` can be retrieved from, like from which command.
And not sure I understand why this or `UUID` are not mandatory

Integrations/integration-MISP_V2.yml
```diff
@@ -1642,7 +1678,23 @@ script:
...
+    - name: uuid
+      description: UUID of attribute to add sighting to
```
Itay4: Please add to the description where this `UUID` can be retrieved from, like from which command.

Integrations/integration-MISP_V2.yml
```diff
@@ -710,6 +720,30 @@ script:
...
+            'expiration': 2
+        }
+        kargs = {
+            'id': demisto.args().get('id'),
+            'uuid': demisto.args().get('uuid'),
+            'type': sighting.get(demisto.args().get('type'))
+        }
+        att_id = demisto.args().get('id', demisto.args().get('uuid'))
+        MISP.set_sightings(kargs)
+        demisto.results(f'Sighting \'{demisto.args().get("type")}\' added to attribute {att_id}')
```
Itay4: Please add `successfully` somewhere in the message

Integrations/integration-MISP_V2.yml
```diff
@@ -710,6 +720,30 @@ script:
...
+
+
+    def test():
+        if not MISP.test_connection():
+            demisto.error('MISP has not connected.')
+        else:
```
Itay4: This is unnecessary as no reason to return anything, just move the `demisto.results('ok')` here.

Integrations/integration-MISP_V2.yml
```diff
@@ -710,6 +720,30 @@ script:
...
+            'type': sighting.get(demisto.args().get('type'))
+        }
+        att_id = demisto.args().get('id', demisto.args().get('uuid'))
+        MISP.set_sightings(kargs)
+        demisto.results(f'Sighting \'{demisto.args().get("type")}\' added to attribute {att_id}')
+
+
+    def test():
+        if not MISP.test_connection():
+            demisto.error('MISP has not connected.')
```
Itay4: Why not use `return_error()` ? And please improve the error message to be more indicative

Integrations/integration-MISP_V2.yml
```diff
@@ -456,6 +456,11 @@ script:
 
         event = MISP.new_event(**event_dic)
         event_id = event.get('id')
+        try:
```
Itay4: Don't really like this way of handling, why not just `if event_id.isdigit():` ?

Integrations/integration-MISP_V2.yml
```diff
@@ -1642,7 +1678,23 @@ script:
       description: All tag names in the event
       type: string
     description: Adds a tag to the given UUID event or attribute.
+  - name: misp-add-sighting
```
Itay4: Why are there no outputs for the command?
I think there should be the same as there are for other commands that have attributes outputs

Integrations/integration-MISP_V2.yml
```diff
@@ -1642,7 +1678,23 @@ script:
...
+    arguments:
+    - name: type
+      required: true
+      auto: PREDEFINED
+      predefined:
+      - sighting
+      - false-positive
```
Itay4: please change to `false_positive`

Integrations/integration-MISP_V2.yml
```diff
@@ -710,6 +720,30 @@ script:
             'EntryContext': ec
         })
 
+
+    def add_sighting():
```
Itay4: please add docstring

Integrations/integration-MISP_V2.yml
jochman: because it's more pythonic `The Python community uses an EAFP (easier to ask for forgiveness than permission) coding style. This coding style assumes that needed variables, files, etc. exist. Any problems are caught as exceptions. This results in a generally clean and concise style containing a lot of try and except statements` while using ifs is more jsic.
But I can still change it tho

Integrations/integration-MISP_V2.yml
jochman: one of them is mandatory, not both. If none will be filled exception will be thrown. will add to description

Itay4: What error will be thrown? from the SDK? can you please post the error message?

Integrations/integration-MISP_V2.yml
Itay4: yes so please change

Integrations/integration-MISP_V2.yml
jochman: add sighting is returning nothing (as well as search, it's returning nothing related to sightings).
There's a problem with MISP.search_sighting, might be related to our instance.

They fixed their documentation after opening an issue, but still there's no response from out instance (tried with postman also). 
https://github.com/MISP/PyMISP/issues/336

Itay4: You don't need anything from the response, even though it would be nice to get something.
Build the context object from the arguments you get.

jochman: sighting is an increased number, so adding something to the output will be misleading and will include only 1 sighting.

Integrations/integration-MISP_V2.yml
jochman: it'll never get there anyway, PyMISP will throw an indicative error before.
If it does get there, it'll be odd and without any further information.

There's a debug tool that I'll try to add to this PR in MISP

Integrations/integration-MISP_V2.yml
jochman: added better error than one raised from the sdk

PR #2893, opened 2019-01-27
Removed threatgrid url detonation duplicate

yardensade:
due to test already being done in generic detonation test playbook

## Status
Ready


yardensade:
Could have been easier to just skip it, but it should be on skip since the test already exist under generic url detonation.

PR #2901, opened 2019-01-28
Added a secret detection in file validation

yardensade:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/14863

## Description
- detect secrets in committed files

   - No

## Must have
- [x] Tests
- [x] Documentation -> https://github.com/demisto/internal-content/tree/master/documentation/secrets
- [x] Code Review


yardensade:
Possible to add:
- Add ability to comment out multi line secrets
- ability for regular commits to run only over the diff parts of a file (and test all repo at night build)
- sanity check for whitelist file in case of a lot of false positives on a single file

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+        text_files_list = get_diff_text_files(branch_changed_files_string)
+
+    else:
+        local_changed_files_string = run_git_command("git diff --name-status --no-merges HEAD")
+        text_files_list = get_diff_text_files(local_changed_files_string)
+
+    return text_files_list
+
+
+def validate_committed_files(branch_name, is_circle):
```
ronykoz: I think you've created a duplicate in the merge, take a look at line 473

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+
+    # secrets validation starts
+    # Get all files that have been modified & added to repo
```
ronykoz: create a function for line 743-760

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+
+    has_schema_problem = validate_added_files(added_files, integration_set, playbook_set,
+                                              script_set, test_playbook_set, is_circle) or has_schema_problem
+
+    if has_schema_problem:
+        sys.exit(1)
+
+
+def search_potential_secrets(secrets_file_paths):
+    """
```
ronykoz: add method docstring(An explanation about what the method does)

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+
+    :param secrets_file_paths: paths of files that are being commited to git repo
+    :return: dictionary of strings sorted by file name for secrets found in files
```
ronykoz: dictionary(filename: (LIST/TUPLE...)secrets) then the explanation

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+    """
+    # Get generic white list set
+    with io.open('./Tests/secrets_white_list.json', mode="r", encoding="utf-8") as secrets_white_list_file:
```
ronykoz: why do you need the io and not regular open?

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+                                                 playbook_set, script_set, test_playbook_set, is_circle)
+
+    has_schema_problem = validate_added_files(added_files, integration_set, playbook_set,
+                                              script_set, test_playbook_set, is_circle) or has_schema_problem
+
+    if has_schema_problem:
+        sys.exit(1)
+
+
+def search_potential_secrets(secrets_file_paths):
```
ronykoz: is it possible to split the content of this function to several functions?
It's a bit long and complicated for a person to dive in

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+    yml_file = file_path_temp + '.yml'
+    if os.path.exists(yml_file):
+        with io.open('./' + yml_file, mode="r", encoding="utf-8") as matching_yml_file:
+            matching_yml_file_contents = matching_yml_file.read()
+    return matching_yml_file_contents
+
+
+def regex_for_secrets(file_contents):
+    """
+    Scans for IOCs with potentially low entropy score
```
ronykoz: this should be in line 867

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+    :param file_contents: file to test as string representation (string)
+    :return  potential_secrets (list) IOCs found via regex, false_positives (list) Non secrets with high entropy
+    """
+    potential_secrets = []
+    false_positives = []
+
+    # URL REGEX
+    urls = re.findall(r'http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\), ]|(?:%[0-9a-fA-F][0-9a-fA-F]))+', file_contents)
```
ronykoz: Maybe create a CONST for each regex?

Tests/scripts/validate_files_structure.py
```diff
@@ -97,6 +100,10 @@
 DIRS = [INTEGRATIONS_DIR, SCRIPTS_DIR, PLAYBOOKS_DIR, REPORTS_DIR, DASHBOARDS_DIR, WIDGETS_DIR, INCIDENT_FIELDS_DIR,
         LAYOUTS_DIR, CLASSIFIERS_DIR, MISC_DIR]
 
+# secrets white list
+ENTROPY_THRESHOLD = 3.4
```
ronykoz: Add an explanation please

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,222 @@ def str2bool(v):
...
+        false_positives += [date[0] for date in dates]
+
+    if potential_secrets and potential_secrets[0] == '::':
+        raise Exception(file_contents, potential_secrets)
+    return potential_secrets, false_positives
+
+
+def calculate_shannon_entropy(data):
+    """
+    Algorithm to determine the randomness of a given data.
```
ronykoz: in line 904, not 905

ronykoz: Add a general explanation about the scoring system, you can take what you wrote in the return section(Shouldn't be there anyway)

Tests/scripts/validate_files_structure.py
yardensade: yeah already fixed, my bad, flake8 caught it

Tests/scripts/validate_files_structure.py
yardensade: was deleted, I used the main one

Tests/scripts/validate_files_structure.py
yardensade: for utf-8 encoding, or else it just converted part of the strings to unicode and it messed up the detection, so either I actively converted everything to utf-8 or just read the file as utf-8

Tests/scripts/validate_files_structure.py
yardensade: I already split it into 5 functions, I can keep going with splitting but the majority is already out.

Tests/scripts/validate_files_structure.py
yardensade: I did this for readability mostly, if you think const makes for better convention I will switch

ronykoz:
> > Didn't we say we want it to be in a different module?
> 
> Well in the end it seemed like there isn't enough volume in here to split it, yet.

it's too much logic in the same file, please make your additions in a separate file, than import the needed functions to the validate hook and just call them there

yardensade:
> it's too much logic in the same file(Plus the file is way too long already), please make your additions in a separate file, than import the needed functions to the validate hook and just call them there

Okie

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
...
+        for ipv6 in ipv6_list:
+            if ipv6 != '::':
+                potential_secrets.append(ipv6)
+    # IPV4 REGEX
+    ipv4_list = re.findall(r'(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)',
+                           file_contents)
+    if ipv4_list:
+        potential_secrets += ipv4_list
+    # Dates REGEX for false positive preventing since they have high entropy
+    dates = re.findall(r'((\d{4}[/.-]\d{2}[/.-]\d{2})[T\s](\d{2}:?\d{2}:?\d{2}:?(\.\d{5,6})?([+-]\d{2}:?\d{2})?Z?)?)',
```
Itay4: Not sure whitelisting dates is required, perhaps just skip them.
We can have way too many dates and just pump the whitelist for no reason.

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
...
+                potential_secrets, false_positives = regex_for_secrets(line)
+                for potential_secret in potential_secrets:
+                    if not any(white_list_string in potential_secret for white_list_string in secrets_white_list):
+                        regex_secrets.append(potential_secret)
+                # added false positives into white list array before testing the strings in line
+                secrets_white_list = secrets_white_list.union(false_positives)
+
+                # calculate entropy for each string in the file
+                for string_ in line.split():
+                    string_ = string_.strip("\"()[],'><:;\\")
```
Itay4: why did you choose to strip by those characters ?

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
...
+    # Dates REGEX for false positive preventing since they have high entropy
+    dates = re.findall(r'((\d{4}[/.-]\d{2}[/.-]\d{2})[T\s](\d{2}:?\d{2}:?\d{2}:?(\.\d{5,6})?([+-]\d{2}:?\d{2})?Z?)?)',
+                       file_contents)
+    if dates:
+        false_positives += [date[0] for date in dates]
+
+    return potential_secrets, false_positives
+
+
+def calculate_shannon_entropy(data):
```
Itay4: did you consider using numpy? should have best run time for this calculation

Tests/scripts/validate_files_structure.py
Itay4: perhaps we should this to be configurable with `3.4` being the default?

Tests/scripts/validate_files_structure.py
```diff
@@ -464,6 +471,20 @@ def integration_valid_in_id_set(file_path, integration_set):
...
+    # Get all files that have been modified & added to repo
+    secrets_file_paths = get_all_diff_text_files(branch_name, is_circle)
+    secrets_found = search_potential_secrets(secrets_file_paths)
+
+    if secrets_found:
+        print_error('Secrets were found in the following files:\n')
+        for file_name in secrets_found:
+            print_error('\nFile Name: ' + file_name)
+            print_error(json.dumps(secrets_found[file_name], indent=4))
+        print_error('Remove or whitelist secrets in order to proceed, then re-commit\n')
```
Itay4: need to address the case this was found only in circle build and not in the hooks (somehow..)
so let's add suggest gently to remove the file asap..

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
         raise argparse.ArgumentTypeError('Boolean value expected.')
 
 
+def is_text_file(file_path):
+    file_extension = os.path.splitext(file_path)[1]
+    text_file_types = set(['.yml', '.py', '.json', '.md', '.txt', '.sh', '.ini', '.eml', '', '.csv'])
```
Itay4: why are you converting a list to a set and not just creating a set in the first place?
how did you choose these file types?

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
...
+    if file_extension in text_file_types:
+            return True
+    return False
+
+
+def get_diff_text_files(files_string):
```
Itay4: please add docstring and somme comments

Tests/scripts/validate_files_structure.py
```diff
@@ -682,6 +703,203 @@ def str2bool(v):
...
+    accepted_file_statuses = ['M', 'A']
+    all_files = files_string.split('\n')
+    text_files_list = set([])
```
Itay4: why are you converting a list to a set and not just creating a set in the first place?

Tests/secrets_white_list.json
```diff
@@ -0,0 +1,193 @@
...
+  "translate",
+  "user",
+  "query",
+  "comment",
+  "list",
+  "date",
+  "unix",
+  "branch",
+  "files"
+
```
Itay4: please remove white spaces

Tests/scripts/validate_files_structure.py
yardensade: Those chars are the common chars I've identified in my work and tests as problematic in messing with the entropy score, after I removed all of these the entropy balanced significantly.

Tests/scripts/validate_files_structure.py
yardensade: I've tested 6 kind of implementations, including numpy. This one showed the best results in run time calculations and is very reliable on big data sets in terms of efficiency.

Tests/scripts/validate_files_structure.py
yardensade: Could be, configureable from where tho. Right now editing this var will immediately take affect.

Tests/scripts/validate_files_structure.py
yardensade: No particular reason, used same conventions that already exist in the code.

Tests/scripts/validate_files_structure.py
yardensade: No particular reason, used same conventions that already exist in the code.

Tests/scripts/validate_files_structure.py
Itay4: ok so let's remove the list there and try to understand where these files types came from

Tests/scripts/validate_files_structure.py
Itay4: ok so let's remove the list

Tests/scripts/validate_files_structure.py
yardensade: I ment no particular reason for the first question, for the second question I forgot to anwer. I chose these file types to filter only text files that we are normally using in the repo. Since I don't want to try and parse files like pictures etc.

yardensade:
> > > Didn't we say we want it to be in a different module?
> > 
> > 
> > Well in the end it seemed like there isn't enough volume in here to split it, yet.
> 
> it's too much logic in the same file(Plus the file is way too long already), please make your additions in a separate file, than import the needed functions to the validate hook and just call them there

Done.

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,223 @@
...
+        file_path_temp, file_extension = os.path.splitext(file_path)
+        if file_extension == '.py':
+            yml_file_contents = retrieve_related_yml(file_path_temp)
+
+        # Open each file, read its contents in UTF-8 encoding to avoid unicode characters
+        with io.open('./' + file_path, mode="r", encoding="utf-8") as file:
+            file_contents = file.read()
+
+            # Add all context output paths keywords to whitelist temporary
+            temp_white_list = create_temp_white_list(yml_file_contents if yml_file_contents else file_contents)
```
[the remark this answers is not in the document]
yardensade: we either parse the yml or just the yml part from a package, there is not both since if you have yml_file_contents the main file you are parsing is not yml if you it is yml you won't have yml_file_contents

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,223 @@
...
+
+# secrets settings
+# Entropy score is determined by shanon's entropy algorithm, most English words will score between 1.5 and 3.5
+ENTROPY_THRESHOLD = 3.8
+SECRETS_WHITE_LIST_FILE = 'secrets_white_list'
+
+
+def is_text_file(file_path):
+    file_extension = os.path.splitext(file_path)[1]
+    text_file_types = {'.yml', '.py', '.json', '.md', '.txt', '.sh', '.ini', '.eml', '', '.csv'}
```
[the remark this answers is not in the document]
yardensade: will have to check about pdf, I assumed not but won't hurt to verify

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,223 @@
...
+    return False
+
+
+def get_diff_text_files(files_string):
+    """Filter out only added/modified text files from git diff
+    :param files_string: string representing the git diff files
+    :return: text_files_list: string of full path to text files
+    """
+    # file statuses to filter from the diff, no need to test deleted files.
+    accepted_file_statuses = ['M', 'A']
```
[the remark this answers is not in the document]
yardensade: Did not want to confuse it with your similar const earlier

Tests/scripts/validate_files_structure.py
```diff
@@ -464,6 +465,23 @@ def integration_valid_in_id_set(file_path, integration_set):
 
 
 def validate_committed_files(branch_name, is_circle):
+
+    # Get all files that have been modified & added to repo
+    secrets_file_paths = secrets.get_all_diff_text_files(branch_name, is_circle)
+    secrets_found = secrets.search_potential_secrets(secrets_file_paths)
```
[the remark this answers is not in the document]
yardensade: Why would we want back the string tho? shouldn't I just print it in the wrapper I'm creating?

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,223 @@
...
+    context_paths = re.findall(r'contextPath: (\S+\.+\S+)', file_contents)
+    for context_path in context_paths:
+        context_path = context_path.split('.')
+        context_path = [white_item.lower() for white_item in context_path]
+        temp_white_list = temp_white_list.union(context_path)
+
+    return temp_white_list
+
+
+def retrieve_related_yml(file_path_temp):
```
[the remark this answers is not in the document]
ronykoz: https://github.com/demisto/content/pull/2911/files
Scripts/ParseEmailFiles/parse_email_files_test.py

Tests/scripts/secrets.py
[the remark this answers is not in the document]
yardensade: added pdf

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,287 @@
...
+import os
+import re
+import math
+import json
+import string
+import PyPDF2
+import validate_files_structure
+
+# secrets settings
+# Entropy score is determined by shanon's entropy algorithm, most English words will score between 1.5 and 3.5
```
ronykoz: Minor note here, i like to keep logical separation in CONSTs so that when reading/searching for something you will see it straight away, try to separate(One blank line) the ENTROPY_THRESHOLD is pseudo configurable, the regexs are definitely unique for the rest of the consts, and the lists are too

Tests/scripts/validate_files_structure.py
```diff
@@ -28,6 +28,7 @@
 import argparse
 from subprocess import Popen, PIPE
 from distutils.version import LooseVersion
+import secrets
```
ronykoz: change that to `from secrets import get_secrets` and move it to line 32(PEP8 requires separation between third party imports and regular ones(Will cause more trouble when adding pylint in the future)

Tests/scripts/secrets.py
```diff
@@ -0,0 +1,287 @@
+import io
+import os
+import re
+import math
+import json
+import string
+import PyPDF2
+import validate_files_structure
```
ronykoz: from validate_files_structure import run_git_command
it's better practice to import just the things you need 
and separate that from the other imports

relevant for lines 69,73

Tests/scripts/secrets.py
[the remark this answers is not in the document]
ronykoz: you haven't fixed this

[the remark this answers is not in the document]
yardensade: I actually see both ParseEmailFiles.py and ParseEmailFiles.yml,
do we have a better way to match them?
If not it is not critical since the worst thing that will happen is that i won't retrieve a temp white list from there, the script will keep working fine with possibly more false positives, that is all.

ronykoz:
@yardensade fix the conflict in the config.yml and then ask Anar to merge this thing, my suggestion is that to make it on Sunday because if you have a bug and someone will work at the weekend he won't be too happy about it

yardensade:
> @yardensade fix the conflict in the config.yml and then ask Anar to merge this thing, my suggestion is that to make it on Sunday because if you have a bug and someone will work at the weekend he won't be too happy about it

I'm going manually over the entire repo first, so that won't be an issue :)

yardensade:
ENTIRE repo scanned 6 times :) We are good to merge @anara123

PR #2911, opened 2019-01-29
Parse email files

anara123:
- fix - extracting msg attachments
- added HeadersMap to outputs
- added AttachedEmails to outputs
- code cleaning done

TODO
- add more tests
- define some OOTB HeadersMap outputs
- add parse_email_from_level=X functionality

## Status
In Progress

## Related Issues
https://github.com/demisto/etc/issues/14440
https://github.com/demisto/etc/issues/13217
https://github.com/demisto/etc/issues/14194
https://github.com/demisto/etc/issues/15277

## Description
Update the code, fix parsing MSG attachment files, clean the code

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Dependencies

## Additional changes


Tests/scripts/update_id_set.py
```diff
@@ -242,7 +242,7 @@ def get_script_data(file_path, script_code=None):
     deprecated = data_dictionary.get('deprecated')
     fromversion = data_dictionary.get('fromversion')
     depends_on, command_to_integration = get_depends_on(data_dictionary)
-    script_executions = sorted(list(set(re.findall("demisto.executeCommand\(['\"](\w+)['\"].*", script_code))))
+    script_executions = sorted(list(set(re.findall("demisto.executeCommand\\(['\"](\w+)['\"].*", script_code))))
```
ronykoz: Not sure why you added this `\` we don't want to identify this char, but strangely enough it catches the same thing on my env

Tests/scripts/validate_files_structure.py
```diff
@@ -563,7 +564,7 @@ def validate_added_files(added_files, integration_set, playbook_set, script_set,
         elif re.match(SCRIPT_YML_REGEX, file_path, re.IGNORECASE) or \
                 re.match(SCRIPT_PY_REGEX, file_path, re.IGNORECASE) or \
                 re.match(SCRIPT_JS_REGEX, file_path, re.IGNORECASE):
-            yml_path, code = get_script_package_data(os.path.dirname(file_path))
+            yml_path, code = get_script_package_data(os.path.dirname(file_path) + '/')
```
ronykoz: lets change this to: so it will be windows compatible for mister Van Dijk
```suggestion
            yml_path, code = get_script_package_data(os.path.join(os.path.dirname(file_path) , ''))
```

package_creator.py
```diff
@@ -75,10 +75,25 @@ def insert_image_to_yml(dir_name, package_path, yml_data, yml_text):
     return yml_text, found_img_path
 
 
-def insert_script_to_yml(package_path, script_type, yml_text, dir_name, yml_data):
+def get_code_file(package_path, script_type):
```
ronykoz: i think we can import this function from vaildate_files_structure

Tests/scripts/update_id_set.py
glicht: yes it is the same. only clearer (at least for me) if not using a raw string. we are basically passing the \( to the regex. we can leave as is.

Tests/scripts/validate_files_structure.py
glicht: let's do these changes in a separate  branch

package_creator.py
glicht: You are correct. had some problems with the import and didn't want to waste on this time. let's do these changes in a separate  branch

PR #2922, opened 2019-02-03
Added DT and created Directory for main docs page

amshamah419:
## Status
Ready

## Description
Added a description of how DT is used as well as a few examples. Also added the public directory table (found in internal-docs) to the docs readme.


docs/context_standards/README.MD
```diff
@@ -253,4 +257,28 @@ The following is the format for a process.
     "Sibling": "LIST, Sibling process objects",
     "Child": "LIST, of Child process objects"
 }
+```
+
+## DBot Score
```
anara123: you should switch between InfoFIle and DBotScore.

amshamah419: You want the order reversed? Infofile on top then Dbot score?

anara123: no, it seems that the title of DBot Score explains InfoFile and vice versa

PR #2925, opened 2019-02-03
Added version validation to added files

yardensade:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15253

### DIFF
--- a/Tests/scripts/validate_files_structure.py
+++ b/Tests/scripts/validate_files_structure.py
@@ -382,14 +382,13 @@ def changed_docker_image(file_path):
 
 
 def validate_version(file_path):
-    change_string = run_git_command("git diff HEAD {0}".format(file_path))
-    is_incorrect_version = re.search("\+([ ]+)?version: (!-1)", change_string)
-    is_incorrect_version_secondary = re.search("\+([ ]+)?\"version\": (!-1)", change_string)
-
-    if is_incorrect_version or is_incorrect_version_secondary:
-        print_error("The version for our files should always be -1, please update the file {}.".format(file_path))
-        return True
-
+    file_extension = os.path.splitext(file_path)[1]
+    if file_extension == '.yml':
+        yaml_dict = get_json(file_path)
+        version_number = yaml_dict.get('commonfields', {}).get('version')
+        if version_number != -1:
+            print_error("The version for our files should always be -1, please update the file {}.".format(file_path))
+            return True
     return False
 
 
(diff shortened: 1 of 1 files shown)


ronykoz:
Please add validation for json files as well

yardensade:
Done, but seems like not every json file has version, so I only test it's file version if it has version key.

Tests/scripts/validate_files_structure.py
```diff
@@ -382,14 +382,21 @@ def changed_docker_image(file_path):
...
-    if is_incorrect_version or is_incorrect_version_secondary:
+    file_extension = os.path.splitext(file_path)[1]
+    version_number = -1
+    if file_extension == '.yml':
+        yaml_dict = get_json(file_path)
+        version_number = yaml_dict.get('commonfields', {}).get('version')
+    elif file_extension == '.json':
+        with open("./" + file_path) as json_file:
+            json_dict = json.load(json_file)
+            if isinstance(json_dict, 'dict'):
```
Itay4: what if it isn't `dict` ?

Tests/scripts/validate_files_structure.py
```diff
@@ -382,14 +382,21 @@ def changed_docker_image(file_path):
...
+                if 'version' in json_dict:
```
Itay4: how come `version` will not exist ?

Tests/scripts/validate_files_structure.py
```diff
@@ -514,7 +521,7 @@ def validate_added_files(added_files, integration_set, playbook_set, script_set,
     has_schema_problem = False
     for file_path in added_files:
         print "Validating {}".format(file_path)
-        if not validate_schema(file_path):
+        if not validate_schema(file_path) or validate_version(file_path):
```
Itay4: shouldn't it be `not validate_version(file_path)` ?

yardensade: nope it returns true if it has validation issue

Tests/scripts/validate_files_structure.py
yardensade: Not every json file has version from what i've been looking at the repo

Tests/scripts/validate_files_structure.py
yardensade: If it isn't a dict it probably does not have versioning that needs checkup, from what I saw in the repo.

PR #2928, opened 2019-02-03
Phishme fix

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15278

## Description
Reimplemented the way dbot score is calculated.

   - Yes
       - Further details: This changes the way dbotScore is calculated - this might result in different scoring for certain indicators.

## Must have
- [X] Tests
- [X] Documentation (https://github.com/demisto/etc/issues/15182)
- [ ] Code Review

## Additional changes
Introduces a fix encountered at circelci:
image


Integrations/integration-PhishMe.yml
```diff
@@ -34,6 +34,30 @@ configuration:
   defaultvalue: "false"
   type: 8
   required: false
+- display: URL Threshold (None, Minor, Moderate or Major). Minimum severity to consider
+    the URL malicious
+  name: urlThreshold
+  defaultvalue: Major
+  type: 0
```
yuvalbenshalom: change type to multi select with the relevant values (None,Minor,Moderate,Major)

Integrations/integration-PhishMe.yml
```diff
@@ -70,34 +94,61 @@ script:
...
+        var severityLevel = 0;
+        var md = '';
+        var threshold_score = calcScore(threshold);
+        res = {}
+        for (var i = 0; i < threat.blockSet.length; i++){
+            if (threat.blockSet[i]['impact']) {
+                var threat_score = calcScore(threat.blockSet[i]['impact']);
+                var adjusted_score = threshold_score <= threat_score ? 3 : threat_score;
+                // if the queried indicator has a severity level, we'll take it
+                if (threat.blockSet[i]['data'] === indicator) {
```
yuvalbenshalom: is it possible that we will have more than one block set with the indicator?
I think in that case we will just take the first severity level

Integrations/integration-PhishMe.yml
```diff
@@ -107,25 +158,33 @@ script:
...
+                } else if (!indicatorFnd) {
+                    dbotScore = Math.max(dbotScore, threatAnalysis['severityScore']);
+                }
             }
             ec.URL['PhishMe'] = {};
             ec.URL.PhishMe['ThreatIDs'] = threatArray;
+            ec.PhishMe = {
+                URL: {
+                    Data: url,
+                    Malicious: ec.URL.Malicious,
```
yuvalbenshalom: not sure if Malicious key should be added anyway.

Integrations/integration-PhishMe.yml
```diff
@@ -182,31 +241,39 @@ script:
...
+        var threshold = params.emailThreshold;
+        var indicatorFnd = false;
+        if (threats.length && threats.length !== 0) {
             ec.Email = {
                 'Data': sender_name,
                 'Malicious': {'Vendor': 'PhishMe', 'Description': 'Match found in PhishMe database'}
             };
             ec.Account = {
-                'Account.Email(val.Address && val.Address === obj.Address)': {
+                'Email': {
```
yuvalbenshalom: why the dt was removed?

Integrations/integration-PhishMe.yml
```diff
@@ -182,31 +241,39 @@ script:
...
+                } else if (!indicatorFnd) {
+                    dbotScore = Math.max(dbotScore, threatAnalysis['severityScore']);
+                }
             }
             ec.Email['PhishMe'] = {};
             ec.Email.PhishMe['ThreatIDs'] = threatArray;
+            ec.PhishMe = {
+                Email: {
+                    Data: sender_name,
+                    Malicious: ec.Email.Malicious,
```
yuvalbenshalom: same here with Malicious

Integrations/integration-PhishMe.yml
```diff
@@ -225,25 +292,47 @@ script:
...
-                    if (threats[0].executableSet[i]['md5Hex'] === str) {
-                        var severityLevel = threats[0].executableSet[i]['severityLevel'];
-                        break;
+                var severityLevel = 0;
+                for (var i = 0; i < threats[k].blockSet.length; i++){
+                    if (threats[k].blockSet[i]['impact']) {
+                        var threat_score = calcScore(threats[k].blockSet[i]['impact']);
+                        var adjusted_score = threshold_score <= threat_score ? 3 : threat_score;
+                        // if the queried indicator has a severity level, we'll take it
+                        if (threats[k].executableSet[i] && threats[k].executableSet[i]['md5Hex'] === str) {
```
yuvalbenshalom: wh have here the same question about block lists with the indicator

Integrations/integration-PhishMe.yml
```diff
@@ -262,25 +351,54 @@ script:
...
+                        ec.IP.GEO = {
+                            "Location": (threats[k].blockSet[i].ipDetail.latitude + ', ' + threats[k].blockSet[i].ipDetail.longitude),
+                            "Country": threats[k].blockSet[i].ipDetail.countryIsoCode
+                        };
+                    }
+                    if (threats[k].blockSet[i]['impact']) {
+                        var threat_score = calcScore(threats[k].blockSet[i]['impact']);
+                        var adjusted_score = threshold_score <= threat_score ? 3 : threat_score;
+                        // if the queried indicator has a severity level, we'll take it
+                        if (threats[k].blockSet[i]['ipDetail'] && threats[k].blockSet[i]['ipDetail']['ip']===ip) {
```
yuvalbenshalom: and same hire with multi block list with the indicator

Integrations/integration-PhishMe.yml
```diff
@@ -262,25 +351,54 @@ script:
...
+                if (!indicatorFnd) {
+                    dbotScore = Math.max(dbotScore, severityLevel);
+                }
             }
             ec.IP['PhishMe'] = {};
             ec.IP.PhishMe['ThreatIDs'] = threatArray;
+            ec.PhishMe = {
+                IP: {
+                    Data: ip,
+                    Malicious: ec.IP.Malicious,
```
yuvalbenshalom: the same question about Malicious

Integrations/integration-PhishMe.yml
DeanArbel: I haven't encountered such a scenario (but have thought of it). Taking that into account, this is exactly what will happen in such a situation - once a block set with the indicator is encountered the `indicatorFound` flag will be turned on, and the functions who call `checkThreat` will turn their flag on as well. The result is that subsequent calls to `checkThreat` will not update the dbot score (i.e. take the first value).
You're probably asking, why call `checkThreat` after encountering the indicator - the answer is to update the markdown, so that it will show the full analysis.

Integrations/integration-PhishMe.yml
DeanArbel: You're right, I'll add in the logic

Integrations/integration-PhishMe.yml
DeanArbel: Had issues with it - for some reason the dt statement was set into the context as is, rather than executing the dt. I'll try to see if I can fix that and return the dt into the key.

Integrations/integration-PhishMe.yml
DeanArbel: You mean the one about the indicator appearing in more than one block set?
The logic here has the same result - line 312 and 323 make sure that only the first indicator will affect the dbot score.

Integrations/integration-PhishMe.yml
DeanArbel: You mean the one about the indicator appearing in more than one block set?
The logic here has the same result - line 378 and 389 make sure that only the first indicator will affect the dbot score.

PR #2931, opened 2019-02-03
Context path for reputation values - new context paths, cleanup, test script + test playbook

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14953

## Description
This change should fix problems resulting from indicator types not being put into context.
The following indicator types should now be put into context as part of the **auto-extract** process, **regardless of available integrations / calculated reputations.**

- Added context path for reputation values for:
MD5 hash
SHA1 hash
SHA256 hash (from referenced PR, and now removed a deprecated script from it too)
IP
Escaped IP
Domain

- Removed deprecated enhancementScriptNames from all of the above

- Replaced old `reputationScriptName` with the new `reputationCommand` (e.g., "IPReputation" will not be used. "!ip" will).

- Added test script + test playbook for:
Email addresses
MD5 hash
SHA1 hash
SHA256 hash
Domains from URL
Domain from email address
URLs
Escaped URLs
IPs
Escaped IPs

## Related PRs
Related to: https://github.com/demisto/content/pull/2886 (completes the process)

   - No

## Additional changes
Did not make changes to pre-3.0.1 values.


Misc/reputations.json
```diff
@@ -44,9 +45,10 @@
       "regex": "\\b(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\[\\.\\]){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\b",
       "reputationCommand": "ip",
       "details": "IP",
+      "contextValue": "Address",
+      "contextPath": "IP(val.Address && val.Address === obj.Address)",
       "enhancementScriptNames": [
         "IPReputation",
-        "SplunkSearch",
         "SplunkPySearch"
```
Itay4: this script is deprecated, should be removed from here

Misc/reputations.json
```diff
@@ -98,26 +100,13 @@
       "regex": "\\b[a-fA-F\\d]{40}\\b",
       "reputationCommand": "file",
       "details": "File SHA1",
```
Itay4: please change to `SHA-1`

TestPlaybooks/script-autoextract_test.yml
```diff
@@ -0,0 +1,38 @@
...
+  sha1_context_output + sha256_context_output + email_context_output
+
+
+  ###  Printing humanreadable to the war room, for auto-extract to work on ###
+  demisto.results( {
+      'Type': entryTypes['note'],
+      'ContentsFormat': formats['text'],
+      'Contents': indicator_data,
+      'HumanReadable': indicator_data
+                   })
```
Itay4: please lynt properly

idovandijk:
> 1. Make sure to add context paths to all indicators, like CVE for example.
> 2. Make sure to add context paths to all relavents objects which depend on versions, like URL for example.
> 3. Please do performance tests.
> 4. What about adding indicators that don't exist there, like `SHA-512`? We have some integrations that output these

@anara123 your thoughts on SHA-512?

Misc/reputations.json
idovandijk: Leaving for backward compatibility

Misc/reputations.json
idovandijk: Are you sure? It seems the convention has been without the hyphen so I don't think we should change them (and if we should, what about the SHA1/256 in the `toVersion: "3.0.1"` parts?
image

idovandijk:
Performance data:
* 100 incidents per minute * 5 minutes = 500 incidents
500 incidents *without* my content: 00:04:14
500 *with* my content: 00:04:15

* 200 incidents per minute * 5 minutes = 1000 incidents
1000 *without* my content: 00:07:19
1000 *with* my content: 00:07:21

anara123:
We can add SHA-512 but not in this PR. Lets complete this PR

idovandijk:
Waiting for review if someone is available

Itay4:
@meirwah performance tests results look good to you?

meirwah:
yes, fine

Misc/reputations.json
Itay4: We are changing this convention according to the standard set by @kirbles19 
All SHA indicators should be `SHA-...`

anara123: Not in the context, in the details (only UI)

Itay4: I meant in `details`

anara123:
@Itay4 @idovandijk is it ready to be merged?

idovandijk:
@anara123 Fine with me

PR #2944, opened 2019-02-04
Parse email files "smtp mail" type

glicht:
## Status
Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/15277

## Description
A fix to work with eml files which have an "smtp mail" type

   - No

## Must have
- [x] Tests: unit test run locally
- [x] Code Review

## Additional changes
Also added package_extractor.py script to extract python code from yml file.

### DIFF
--- a/Scripts/ParseEmailFiles/parse_email_files_test.py
+++ b/Scripts/ParseEmailFiles/parse_email_files_test.py
@@ -1,4 +1,6 @@
-from ParseEmailFiles import MsOxMessage
+from ParseEmailFiles import MsOxMessage, main
+from CommonServerPython import entryTypes
+import demistomock as demisto
 
 
 def test_msg_html_with_attachments():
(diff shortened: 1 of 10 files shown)


package_extractor.py
```diff
@@ -0,0 +1,55 @@
...
+
+import yaml
+import argparse
+from io import open
+
+INTEGRATION = 'integration'
+SCRIPT = 'script'
+
+
+def extract_code(yml_path, output_path, demisto_mock, commonserver=None, yml_type=None):
```
ronykoz: I would have splitted this into couple of functions or just added some blank lines for some visual/logical separation, you have a setup stage and then extraction, and creation stages here

package_extractor.py
```diff
@@ -0,0 +1,55 @@
...
+        yml_data = yaml.safe_load(yml_file)
+        script = yml_data['script']
+        if yml_type == INTEGRATION:  # in integration the script is stored at a second level
+            script = script['script']
+    with open(output_path, 'w', encoding='utf-8') as code_file:
+        if demisto_mock:
+            code_file.write(u"import demistomock as demisto\n")
+        if commonserver:
+            code_file.write(u"from CommonServerPython import *\n")
+        if type(script) == str:  # pyyaml may return either a unicode or str depending upon the content of script
```
ronykoz: lets move this to the first `with`

Scripts/ParseEmailFiles/parse_email_files_test.py
```diff
@@ -22,3 +24,29 @@ def test_msg_utf_encoded_subject():
...
+                    {'info': 'SMTP mail, UTF-8 Unicode text, with CRLF terminators'}}]
+        else:
+            raise ValueError('Unimplemented command called: {}'.format(name))
+
+    mocker.patch.object(demisto, 'args', return_value={'entryid': 'test'})
+    mocker.patch.object(demisto, 'executeCommand', side_effect=executeCommand)
+    mocker.patch.object(demisto, 'results')
+    # validate our mocks are good
+    assert demisto.args()['entryid'] == 'test'
+    # assert demisto.executeCommand('getFilePath', {})[0]['Type'] == entryTypes['note']
```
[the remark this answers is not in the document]
glicht: will do this on separate PR where will be trying to run the unit tests as part of the build

glicht:
The package_extractor.py is a small script I need for getting the CommonServerPython for moving to a mode where we run unit tests as part of the build. I was also using this locally. Let's just leave this for now until we progress a bit. I am planning to do a separate PR which includes code for running the unit tests.

PR #2945, opened 2019-02-05
Anomali DBotScore fixes

amshamah419:
##
Ready

## Related Issues
fixes: [demisto/etc#15296](https://github.com/demisto/etc/issues/15296)

## Description
Customer states that reported false positives are being treated as positives. Changes were made to check the status flag before allocating a positive score.


amshamah419:
fixes: demisto/etc#15400

PR #2947, opened 2019-02-05
Fix disabled proxy setting in RSA NetWitness

omercnet:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15413

## Description
Fixing no proxy setting to disable proxy

## Screenshots
image
image

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes

### DIFF
--- a/Integrations/integration-RSANetWitness_v11.1.yml
+++ b/Integrations/integration-RSANetWitness_v11.1.yml
@@ -74,16 +74,9 @@ script:
     """
 
     def set_proxies():
-
-        if demisto.params()['proxy']:
-            http = os.environ['http_proxy'] or os.environ['HTTP_PROXY']
-            https = os.environ['https_proxy'] or os.environ['HTTPS_PROXY']
-            proxies = {
-                'http': http,
-                'https': https
-            }
-            return proxies
-        return {"http": None, "https": None}
+        if not demisto.params()['proxy']:
+            for _ in ['HTTP_PROXY', 'HTTPS_PROXY', 'http_proxy', 'https_proxy']:
+                del os.environ[_]
 
 
     """


Integrations/integration-RSANetWitness_v11.1.yml
```diff
@@ -83,7 +83,7 @@ script:
                 'https': https
             }
             return proxies
-        return None
+        return {"http": None, "https": None}
```
Itay4: we should also delete proxy environment vars in this case

omercnet: if you really want you can use this http://docs.python-requests.org/en/master/api/#requests.Session.trust_env

but it's an overkill.. this explicitly tells requests not to use a proxy

Itay4: I meant:
```
    del os.environ['HTTP_PROXY']
    del os.environ['HTTPS_PROXY']
    del os.environ['http_proxy']
    del os.environ['https_proxy']
```

omercnet: but why?

anara123:
Because this is the way we handle proxy in all the integrations

Sent from my iPhone

On 5 Feb 2019, at 18:42, Omer Cohen <notifications@github.com<mailto:notifications@github.com>> wrote:


@omercnet commented on this pull request.

In Integrations/integration-RSANetWitness_v11.1.yml<https://github.com/demisto/content/pull/2947#discussion_r253947033>:

> @@ -83,7 +83,7 @@ script:
                 'https': https
             }
             return proxies
-        return None
+        return {"http": None, "https": None}


but why?

omercnet:
> Because this is the way we handle proxy in all the integrations

so what you're saying is that you want to keep doing things wrong? gotcha :)

anara123:
@omercnet what is the exact problem with
```
if not demisto.params()['proxy']:
        del os.environ['HTTP_PROXY']
        del os.environ['HTTPS_PROXY']
        del os.environ['http_proxy']
        del os.environ['https_proxy']
```

PR #2952, opened 2019-02-06
Changed version validate to properly handle the scheme of playbooks

yardensade:
Added version to reputations.json

## Status
Ready


Tests/scripts/validate_files_structure.py
```diff
@@ -393,18 +393,37 @@ def validate_version(file_path):
...
 
 
+def validate_reputations(json_dict):
+    is_valid = True
+    reputations = json_dict.get('reputations')
+    for reputation in reputations:
+        internal_version = reputation.get('version')
+        if internal_version != -1:
+            object_id = reputation.get('id')
+            print_error("object with id {} must have version -1".format(object_id))
```
Itay4: Please change the message to be something like `Reputation object with...`

PR #2678, opened 2018-12-19
fixed dedup conditional task

Clairvoyantx:
## Status
Ready

## Related Issues
N/A

## Description
Conditional task in the playbook requires boolean comparison for isSimilarIncidentFound

## Screenshots
Current
screen shot 2018-12-19 at 3 04 44 pm

Required
screen shot 2018-12-19 at 3 04 26 pm

## Required version of Demisto
3.5.0

## Does it break backward compatibility?
   - No

PR #2956, opened 2019-02-06
fixed dedup conditional task (#2678)

anara123:
Conditional task in the playbook requires boolean comparison for isSimilarIncidentFound

## Status
Ready

## Description
Conditional task in the playbook requires boolean comparison for isSimilarIncidentFound

## Related PRs
https://github.com/demisto/content/pull/2678

   - No

## Must have
- [x] Tests
- [ ] Code Review


Tests/scripts/secrets.py
```diff
@@ -19,7 +19,7 @@
 ENTROPY_THRESHOLD = 3.8
 
 SKIPPED_FILES = {'secrets_white_list', 'id_set.json', 'conf.json'}
-ACCEPTED_FILE_STATUSES = ['M', 'A']
+ACCEPTED_FILE_STATUSES = ['M', 'A', "R"]
```
yardensade: I'm not sure I like searching every R file for secrets is the right thing to do.
We have tried getting official docs for the many R types there are without success, I'd rather have an array of R's per case we encounter and verify to fit our flow over treating all R types the same way in advance.

Tests/scripts/validate_files_structure.py
```diff
@@ -631,6 +640,8 @@ def validate_modified_files(integration_set, modified_files, playbook_set, scrip
     has_schema_problem = False
     for file_path in modified_files:
         print "Validating {}".format(file_path)
+        if "conf.json" in file_path:
```
yardensade: perhaps like in secrets we should have an array of skipped files, for readability purposes.

yardensade:
Please fix the flake8 error so we can confirm the test is working.

PR #2965, opened 2019-02-10
Renamed TestIsEmailAddressInternal.yml with 'playbook-' prefix

yardensade:
Found a playbook that does not have playbook prefix in its file name, added prefix to name.

- Verified the old file name is not used anywhere in content.

## Status
Ready


ronykoz:
before you merge, this test is not running and not configured in conf.json please check who wrote it and what should it test exactly

yardensade:
@bakatzir please enlighten us :)

PR #2968, opened 2019-02-10
Moved RedLockTest and Tanium to skipped and increased Phishing test - attachment timeout 

ShahafBenYakir:


ronykoz:
Pretty sure Phishing playbook shouldn’t take 10 mins to run, @idovandijk could you check that out please? we need to make sure it doesn't go to the non-malicious section

idovandijk:
> Pretty sure Phishing playbook shouldn’t take 10 mins to run, @idovandijk could you check that out please? we need to make sure it doesn't go to the non-malicious section

It is not a timeout issue. I already released the fix that makes it go in the malicious path by using another indicator and using another integration but the reputation (Minemeld instead of VirusTotal), but IIRC @yardensade discovered that they decided to cancel Minemeld and so the test keeps on failing. If we keep disabling reputation integrations then the phishing test will become useless :/

yardensade:
Setting it to 30 minutes even won't help, it gets stuck on manual for a reason.

ronykoz:
Minemeld is not on the skip section..

idovandijk:
Any idea why?
image

ShahafBenYakir:
@ronykoz the Phishing test failed due to timeout, however, its incident was closed in Demisto instance.
circleci 2019-02-10 21-48-37

@idovandijk the error you are getting is exactly because what you have said, which is the lack of Minemeld to run the playbook on. What is the instance you tried to run the playbook on? local env? EWS machine? Try to add an instance for Minemeld integration and run again. 
Of course, this is all my understanding so you are invited to correct me if I'm wrong. @yardensade @ronykoz

idovandijk:
> @ronykoz the Phishing test failed due to timeout, however, its incident was closed in Demisto instance.
> circleci 2019-02-10 21-48-37
> 
> @idovandijk the error you are getting is exactly because what you have said, which is the lack of Minemeld to run the playbook on. What is the instance you tried to run the playbook on? local env? EWS machine? Try to add an instance for Minemeld integration and run again.
> Of course, this is all my understanding so you are invited to correct me if I'm wrong. @yardensade @ronykoz

It is one of the instances set up in our build process (from circle CI). If Minemeld is enabled I think it should be enabled on any of them(?)..

ShahafBenYakir:
> It is one of the instances set up in our build process (from circle CI). If Minemeld is enabled I think it should be enabled on any of them(?)..

Nope... it depends if the `Phishing test` was supposed to run. If it is, then the Minemeld integration would be configured. Anyway, send me the build number I would check (got nothing better to do) @idovandijk .

PR #2973, opened 2019-02-11
Filter by list script fix

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14858


Scripts/script-FilterByWhitelist.yml
```diff
@@ -19,15 +19,23 @@ script: |
...
 
+  # fill whitlisted array with all the the values that match the regex items in listname argument
   for item in items:
-      if item in lst:
-          hr = hr + item + " is in the list\n"
-          whitlisted.append(item)
-      else:
+      found = ''
+      for list_item in lst:
+          if re.search(list_item, item):
```
anara123: It seems that you handle all the items in the whitelist as they were regexes, I think regex in whitelist has different sign. 
I think we should handle regular "strings" as previously and regex we should handle the way you did here

adi88d: If the items in the list are regular strings, the script handle it like before, check FilterByList - Test playbook, in the list there are regular string url and regex.
In my opinion its good idea to have FilterByList script for strings and other script for regex, or maybe add boolean argument to the scrip - useRegex to avoid confusing

anara123: it should be the same script, because user wants to check if indicator X is whitelisted, he doesn't care if you have regex or regular string in the whitelist.
I believe the regular string still works because of the match.
Add the following use case to the tests:
*Whitelist*
foo.demisto.com

*Indicator to check*
foo.demisto.com - should be whitelisted
demisto.com - should not be whitelisted

adi88d: Check FilterByList - Test playbook - it checks regular strings and regex with this values:
^demisto.com$ and google.com

PR #2981, opened 2019-02-11
EclecticIQ Platform integration

Itay4:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/13350

Any

   - No

## Must have
- [x] Tests
- [ ] Documentation - waiting for @kirbles19 to review the integration and then documentation will be generated
- [ ] Code Review

## Dependencies
None


Integrations/EclecticIQ/EclecticIQ.yml
```diff
@@ -0,0 +1,354 @@
...
+  name: insecure
+  defaultvalue: "true"
+  type: 8
+  required: false
+- display: Use system proxy
+  name: proxy
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: 'IP threshold. Minimum maliciousness confidence level to consider the IP
```
yaakovi: use the single select feature instead of free text.
same for all

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+PASSWORD = demisto.params().get('credentials').get('password')
+URL = demisto.params()['url']
+SERVER = URL[:-1] if (URL and URL.endswith('/')) else URL
+USE_SSL = not demisto.params().get('insecure', False)
+HEADERS = {}
+IP_THRESHOLD = demisto.params().get('ip_threshold').lower()
+URL_THRESHOLD = demisto.params().get('url_threshold').lower()
+FILE_THRESHOLD = demisto.params().get('file_threshold').lower()
+EMAIL_THRESHOLD = demisto.params().get('email_threshold').lower()
+DOMAIN_THRESHOLD = demisto.params().get('domain_threshold').lower()
```
yaakovi: loved how the name is ordered by length 😉

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+    res = requests.request(
+        method,
+        SERVER + url_suffix,
+        headers=headers,
+        json=cmd_json,
+        proxies=PROXIES,
+        verify=USE_SSL
+    )
+    if res.status_code not in {200}:
+        if res.status_code == 405:
```
yaakovi: shouldn't this be 404?
405 seems to be more relevant to bug in the code and not the parameter.

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+            return_error(
+                'Error in API call to EclecticIQ Integration: [405] - Not Allowed - Might occur cause of an invalid '
+                'URL.')
+        return_error(res.content)
```
yaakovi: is that leftovers from debug? if not, everything pass this is unreachable

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+        Translated DBot Score
+
+    """
+
+    ip = demisto.args()['ip']
+
+    response = get_observable(ip)
+
+    if 'total_count' in response and response['total_count'] == 0:
+        human_readable = 'No results found'
```
yaakovi: why not leave it the tableToMarkdown?

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    context = {
+        'DBotScore': dbot_output
+    }
+
+    if observables:
+        human_readable_title = 'EclecticIQ IP reputation - {}'.format(ip)
+        human_readable = tableToMarkdown(human_readable_title, integration_outputs)
+        context['EclecticIQ.IP(val.ID===obj.ID)'] = integration_outputs
```
yaakovi: directly add this to context and use CreateContent with removeNulls=True to avoid empty fields.
the human_readable can also be used directly in the entry (it will write "no result")

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+        'Contents': response,
+        'ContentsFormat': formats['json'],
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': human_readable,
+        'EntryContext': context
+    })
+
+
+def get_observable(ioc):
+    cmd_url = '/api/observables?filter[value]={}'.format(ioc)
```
yaakovi: "filter[value]" should be separated from the url and passed to request as params.
this way requests module will handle the URL escaping of special chars

Integrations/EclecticIQ/EclecticIQ.py
```diff
@@ -0,0 +1,640 @@
...
+
+def original_extract(observable_id):
+    cmd_url = '/private/entities/original-extract/{}'.format(observable_id)
+    response = http_request('GET', cmd_url)
+    return response
+
+
+''' COMMANDS MANAGER / SWITCH PANEL '''
+
+login()
```
yaakovi: shouldn't be a logout?

Integrations/EclecticIQ/EclecticIQ.yml
Itay4: that would limit the integration to be from version 4.1
not sure we are there yet

Integrations/EclecticIQ/EclecticIQ.py
Itay4: worked hard on that one

Integrations/EclecticIQ/EclecticIQ.py
Itay4: well, i didn't choose the status codes.. but yea its 405

Integrations/EclecticIQ/EclecticIQ.py
Itay4: i dont like the way tabletomarkdown handles it, because it shows a big title and then in small font size that there were no entries found

Integrations/EclecticIQ/EclecticIQ.py
Itay4: i guess there should be, but their api does not support it

Integrations/EclecticIQ/EclecticIQ.py
Itay4: what do you mean by "directly" ?

yaakovi: relates to previous comment about the markdown

Integrations/EclecticIQ/EclecticIQ.py
yaakovi: just wanted to make sure it's not a typo

Integrations/EclecticIQ/EclecticIQ.yml
yaakovi: ok, then we should consider a way to track those upgrades and revisit this when the version is more popular.
it's seems pretty wasteful we have new features that can prevent user problems and not use them.

Integrations/EclecticIQ/EclecticIQ.py
Itay4: tried it, their api keep returning 401 Unauthorized for any request with `params`
will check it with them

PR #2984, opened 2019-02-11
added function return_outputs() which replaces demisto.results()

anara123:
## Status
Ready

## Description
return_outputs() which replaces demisto.results() - more return outputs in demisto should be more intuitive 

   - No

## Must have
- [ ] Tests - no tests - because it does demisto.results() and not returns anything
- [x] Documentation - will documented in Script Helper automatically
- [ ] Code Review


Scripts/script-CommonServerPython.yml
```diff
@@ -1203,6 +1203,46 @@ script: |-
       return True
 
 
+  def return_outputs(readable_output, outputs, raw_response=None):
+      """
+      This function replaces demisto.results(), makes the use more easy and intuitive.
```
ronykoz: This function wraps the demisto.results(), makes the usage of returning results to the user more intuitively.

Scripts/script-CommonServerPython.yml
```diff
@@ -1203,6 +1203,46 @@ script: |-
...
+      :return: None
+      :rtype: ``None``
+      """
+      return_entry = {
+          "Type": entryTypes["note"],
+          "HumanReadable": "",
+          "ContentsFormat": formats["text"],
+          "Contents": None
+      }
+      if readable_output:
```
ronykoz: No need, it's mandatory as is, just put it directly into the dict

Scripts/script-CommonServerPython.yml
```diff
@@ -1203,6 +1203,46 @@ script: |-
...
+          return_entry["HumanReadable"] = readable_output
+
+      if raw_response is not None:
+          return_entry["ContentsFormat"] = formats["json"]
+          return_entry["Contents"] = raw_response
+
+      if outputs is not None:
```
ronykoz: same here

Scripts/script-CommonServerPython.yml
```diff
@@ -1203,6 +1203,46 @@ script: |-
...
+      :param raw_response: must be dictionary, if not provided then will be equal to outputs. usually must be the original
+      raw response from the 3rd party service (originally Contents)
+
+      :return: None
+      :rtype: ``None``
+      """
+      return_entry = {
+          "Type": entryTypes["note"],
+          "HumanReadable": "",
+          "ContentsFormat": formats["text"],
```
ronykoz: i think it should always be json, you are changing it to json anyway..

anara123:
@ronykoz fixed

PR #2991, opened 2019-02-12
adding test existance validation

ronykoz:
Checking that caught tests appear in conf.json otherwise fail the build

### DIFF
--- a/Tests/scripts/configure_tests.py
+++ b/Tests/scripts/configure_tests.py
@@ -208,6 +208,7 @@ def collect_tests(script_ids, playbook_ids, integration_ids, catched_scripts, ca
 
     :return: (test_names, missing_ids) - All the names of possible tests, the ids we didn't match a test for.
     """
+    caught_missing_test = False
     catched_intergrations = set([])
 
     test_names = get_test_names()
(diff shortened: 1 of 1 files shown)


Tests/scripts/configure_tests.py
```diff
@@ -240,13 +244,19 @@ def collect_tests(script_ids, playbook_ids, integration_ids, catched_scripts, ca
                         if not command_to_integration.get(command) or \
                                 command_to_integration.get(command) == integration_id:
 
+                            detected_usage = True
                             tests_set.add(test_playbook_name)
                             catched_intergrations.add(integration_id)
 
+        if detected_usage and test_playbook_name not in test_names:
+            caught_missing_test = True
+            print_error("The playbook {} does not appear in the conf.json file, which means no test with it will run."
```
anara123: I think we should move to raise an Exceptions instead of passing bool flags

ronykoz: The problem with exceptions is that the user will see one error at a time instead of seeing everything at one run

anara123: we can create our exception - like DemistoException or ValidationException and catch the exceptions that are ours and print them in more nice way.

yardensade: I think we should refactor the code to OO.
Would be much easier managing errors, and will have a cleaner flow instead of passing extra vars around.

ronykoz: @anara123 creating our own exception will have 0 value, we will achieve the same thing we already have here
@yardensade not sure that OO is the right way to go here, as it is not so much object oriented code, but even if we decide to go that way we need to check if we have the capacity for it

anara123: @yardensade agree with @ronykoz on the OO, don't know if it will give us much.
@ronykoz why, using Exceptions the code will not use flags all over. If failed then failed, otherwise continue

ronykoz: @anara123 because we want to catch **all** failures at one run and not fail each time on different mistake, the point is to show everything wrong with your code so you will be able to address everything at once

yardensade: I agree with @ronykoz on this, I think knowing all of your errors in advance is a lot better over having to find out each time about different error, which will increase frustration levels by a lot when you don't actually know when your commit is going to pass.

anara123: It depends. In this specific case you use `caught_missing_test` as a flag and do not aggregate anything. outside the function you do sys.exit(1) which is bad practice.
I agree that aggregation of errors is better UX - in that case I would aggregate the errors to that specific functions and then raise an Exception for all those errors together
or print_error for each and raise Exception if flag_error=True

PR #2992, opened 2019-02-12
Protect fetch time parameter

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15432

## Description
validate fetch time param and set default value when is empty


Tests/id_set.json
```diff
@@ -8892,6 +8892,30 @@
...
+                    "fd-list-agents", 
+                    "fd-create-ticket-note", 
+                    "fd-delete-ticket", 
+                    "fd-search-tickets"
+                ], 
+                "tests": [
+                    "Freshdesk-Test"
+                ]
+            }
+        },
```
liorblob: Why is this here?...

Tests/secrets_white_list.json
```diff
@@ -646,7 +650,11 @@
       "sn1pr11mb0638d849680e1325bc4f7951cc970@sn1pr11mb0638.namprd11.prod.outlook.com",
       "idov@demisto.com",
       "bar.hochman@demisto.com",
-      "powershellteam@hotmail.com"
+      "powershellteam@hotmail.com",
+      "example3@gmail.com",
+      "example1@gmail.com",
+      "jeffrey.collings@gmail.com",
+      "example2@gmail.com"
```
liorblob: ???

yardensade: Do not add irrelevant emails that are not supposed to be exposed on public repo please.

adi88d: done

liorblob:
@anara123 I don't think the error should come from there, but it's always good to check the function inputs

anara123:
function should protect herself, if there are extra space, then we can strip it, if None passed then invalid input = exception.
And in each integration in the beginning of the code, we should run validate_date_range(date_range)

yardensade:
If we are going to strip anyway, we should strip from the function itself, no point validating it over and over again in each integration.
Also might be good to be able and pass into the function a default argument in case you don't want the function to fail on lack of input instead of creating another default var in the code.
So if you don't provide it AND there is no actual input only then it the function will throw exception.

anara123:
there are cases that this kind of protective code could be a correct way, but in this case I think it is not. by allowing default argument it means that user can provide invalid date range like "gkjfdnkhjdfnh" and we will fetch incidents 3 days back instead of returning an error and asking him to fix it

liorblob:
It will only do it in cases where the parameter is an empty space, if it has a concrete value then it will be sent to the function

PR #2996, opened 2019-02-13
Virus Total Private API enhancements

hirshbergshachar:
Updated maliciousness logic for url-report, changed test function.
Fixes: https://github.com/demisto/etc/issues/14620
Fixes: https://github.com/demisto/etc/issues/15550

### DIFF
--- a/Integrations/integration-VirusTotal-Private_API.yml
+++ b/Integrations/integration-VirusTotal-Private_API.yml
@@ -789,7 +789,7 @@ script:
                 }
             })
             dbotScore = 3
-        elif (bad_downloads_amount >= 0 or detected_urls >= 0):
+        elif (bad_downloads_amount >= 0 or len(detected_urls) >= 0):
             dbotScore = 2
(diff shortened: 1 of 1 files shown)


Integrations/integration-VirusTotal-Private_API.yml
```diff
@@ -651,14 +663,21 @@ script:
         return http_request('GET', api_endpoint, params, DEFAULT_HEADERS)
 
 
+    def check_detected_urls_threshold(detected_urls, threshold):
+        for url in detected_urls:
+            if url.get("positives") > threshold:
```
Itay4: By the description of the threshold param, I think it makes sense for the operator to be `>=`

hirshbergshachar: Fixed

hirshbergshachar:
@Itay4 for the other relevant commands VT returns a number of positive detections, so we use different logic.

PR #3007, opened 2019-02-13
sort id_set.json in order to avoid conflicts

anara123:
## Status
Ready

## Description
- sorted the id_set.json in order to avoid conflicts that will happen when there are changes in master
- sort the id_set.json every time update_id_set.py running - in order to make sure that the id_set.json stays sorted, even if someone changed id_set.json manually

   - No


Tests/scripts/update_id_set.py
```diff
@@ -53,6 +54,18 @@ def run_git_command(command):
     return output
 
 
+def insert(seq, keys, item, keyfunc=lambda v: v):
```
glicht: seems that this function is not used. I think we can remove it.

Tests/scripts/update_id_set.py
```diff
@@ -503,10 +521,12 @@ def update_id_set():
 
     if added_files or modified_files:
         new_ids_dict = OrderedDict()
-        new_ids_dict['scripts'] = script_set
-        new_ids_dict['playbooks'] = playbook_set
-        new_ids_dict['integrations'] = integration_set
-        new_ids_dict['TestPlaybooks'] = test_playbook_set
+        # we sort each time the whole set in case someone manually changed something
+        # it shouldn't take too much time
+        new_ids_dict['scripts'] = sort(script_set)
```
ronykoz: These sorts should take a lot, no? you are sorting a file of 15K lines..

anara123: it actually works very fast because we have it all in the memory and we sort in total ~1000 items in array

anara123: milliseconds

PR #3016, opened 2019-02-16
Added Joe Security to skipped integrations issue 15614

ShahafBenYakir:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/15614


ShahafBenYakir:
Cause it fails several other playbooks (like phishing and file detonato)

PR #3020, opened 2019-02-17
Azure compute rework

avidan-H:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/14043

## Description
Reworked the Azure Compute integration/commands to adhere to demisto code conventions.

   - No

## Must have
- [ ] Tests - No test playbook because the `vm-create-instance` command has side-effect of creating disk resources and the `vm-delete-instance` command does not delete those disk resources that are created.
- [ ] Code Review

## Dependencies
- NA


Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+
+        returns:
+            Resource-Group Objects
+        """
+        response = list_resource_groups()
+        # Retrieve relevant properties to return to context
+        value = response.get('value')
+        resource_groups = []
+        for resource_group in value:
+            rg = {
```
Itay4: please improve dict name

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+                'Name': resource_group.get('name'),
+                'ID': resource_group.get('id'),
+                'Location': resource_group.get('location'),
+                'ProvisioningState': resource_group.get('properties').get('provisioningState')
+            }
+            resource_groups.append(rg)
+
+        title = "List of Resource Groups"
```
Itay4: `"` -> `'`

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        human_readable = tableToMarkdown(title, resource_groups)
```
Itay4: add `removeNull=True`

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        demisto.results({
+            'Type': entryTypes['note'],
+            'Contents': response,
+            'ContentsFormat': formats['text'],
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': human_readable,
+            'EntryContext': {'Azure.ResourceGroup(val.Name && val.Name === obj.Name)': resource_groups}
```
Itay4: initialize the context object before the entry

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        })
+
+
+    #<-------- Virtual Machines -------->#
+
+    def list_vms(resource_group):
+        # Construct endpoint URI suffix
+        url_endpoint = resource_group + '/providers/Microsoft.Compute/virtualMachines?api-version=' + API_VERSION
```
Itay4: can't we append the api version in the http request?

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        # Call API
+        update_access_token()
+        response = http_request('GET', url_endpoint, codes={200})
```
Itay4: lets make codes default to be 200 in the http request and not pass it everytime

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        returns:
+            Virtual Machine Objects
+        """
+        resource_group = demisto.args().get('resource_group')
+        response = list_vms(resource_group)
+
+        num_of_vms = len(response.get('value'))
+
+        vms = []
+        for i in range(0, num_of_vms):
```
Itay4: lets improve the loop here and not use `range`

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+            vm_id = properties.get('vmId')
+            os_type = os_disk.get('osType')
+            vm = {
+                'Name': vm_name,
+                'ID': vm_id,
+                'Size': datadisk,
+                'OS': os_type,
+                'Location': location,
+                'ProvisioningState': provisioning_state,
+                'ResourceGroup': resource_group,
```
Itay4: remove unnecessary comma

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+        num_of_vms = len(response.get('value'))
+
+        vms = []
+        for i in range(0, num_of_vms):
+            value = response.get('value')[i]
+            vm_name = value.get('name')
+            location = value.get('location')
+            properties = value.get('properties')
+            provisioning_state = properties.get('provisioningState')
+            os_disk = properties.get('storageProfile').get('osDisk')
```
Itay4: need to add `{}` to be the default in the first `.get()`

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+            datadisk = os_disk.get('diskSizeGB', 'NA')
+            vm_id = properties.get('vmId')
+            os_type = os_disk.get('osType')
+            vm = {
```
Itay4: not sure i understand why you're not inserting the values directly to the dict

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+
+
+    def get_vm(args):
+        # Retrieve relevant command arguments
+        resource_group = args.get('resource_group')
+        vm_name = args.get('virtual_machine_name')
+
+        # Construct endpoint URI suffix
+        url_endpoint = resource_group + '/providers/Microsoft.Compute/virtualMachines/'
+        url_endpoint += vm_name + '?$expand=instanceView&api-version=' + API_VERSION
```
Itay4: why are we not using Python requests `params` here ?

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+
+    try:
+        if demisto.command() == 'test-module':
+            update_access_token()
+            demisto.results('ok')
+        elif demisto.command() in commands.keys():
+            commands[demisto.command()]()
+
+    except Exception as e:
+        LOG(e)
```
Itay4: no need for this

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+
+    PARAMS = demisto.params()
+    USE_SSL = not demisto.params().get('unsecure')
+    CLIENT_ID = PARAMS.get('client_id')
+    CLIENT_SECRET = PARAMS.get('client_secret')
+    SUBSCRIPTION_ID = PARAMS.get('subscription_id')
+    TENANT_ID = PARAMS.get('tenant_id')
+    SERVER = 'https://management.azure.com'
+    BASE_URL = SERVER + "/subscriptions/" + SUBSCRIPTION_ID + "/resourceGroups/"
+    API_VERSION = "2018-06-01"
```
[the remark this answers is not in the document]
avidan-H: I now changed it to be a parameter and not hard-coded, with the default of "2018-06-01" if the user doesn't specify otherwise in the configuration. I chose it because I believe it's the latest version of the API available.

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+    HEADERS = {}
+
+    # Image options to be used in the create_vm_command
+    IMAGES = {
```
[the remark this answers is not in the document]
avidan-H: No. There are a boatload more. Should I add them all?

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+            if r.status_code not in green_codes:
+                err_msg = 'Error in API call to Azure Compute Integration [{}] - {}'.format(r.status_code, r.reason)
+                err = r.json().get('error')
+                if err:
+                    err_msg1 = "\nError code: {}\nError message: {}".format(err.get('code'), err.get('message'))
+                    err_msg += err_msg1
+                raise Exception(err_msg)
+            response = json.loads(r.content)
+        except ValueError as e:
+            response = r.content
```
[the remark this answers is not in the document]
avidan-H: You're asking why'd I bother with the `as e` bit?

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+      description: Network Interface ID to link the VM with. This has to be created
+        from the Azure Portal. Note that the vm's location property must match that
+        of the Network Interface you choose to link it to.
+      isArray: false
+      name: nic_id
+      required: true
+      secret: false
+    - auto: PREDEFINED
+      default: false
+      defaultValue: Standard_D1_v2
```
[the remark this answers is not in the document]
avidan-H: No reason in particular. No default value in that case?

Integrations/integration-AzureCompute.yml
avidan-H: You mean I shouldn't include a LOG statement before `return_error`? Or the fact that I'm LOGging an Exception object, and not a string?

Integrations/integration-AzureCompute.yml
avidan-H: I thought it improved code readability but I wasn't entirely consistent across all the commands. If there was only 2 or 3 properties to fetch that were going to be returned to the context then I directly inserted the values into the dict - if it was a lot, I separated it.

Integrations/integration-AzureCompute.yml
avidan-H: I had it so the codes defaulted to `{200, 201, 202, 204}`. And I only passed codes when I wanted to limit the acceptable codes. You prefer I have it the other way? Default to 200 and pass codes only when I wish to expand the scope of acceptable codes?

Integrations/integration-AzureCompute.yml
avidan-H: I made it part of the passed parameters as you noted elsewhere in your code review

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+                err_msg = 'Error in API call to Azure Compute Integration [{}] - {}'.format(r.status_code, r.reason)
+                err = r.json().get('error')
+                if err:
+                    err_msg1 = "\nError code: {}\nError message: {}".format(err.get('code'), err.get('message'))
+                    err_msg += err_msg1
+                raise Exception(err_msg)
+            response = json.loads(r.content)
+        except ValueError as e:
+            response = r.content
+        except Exception as e:
```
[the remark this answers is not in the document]
avidan-H: You're right, it's superfluous.

Integrations/integration-AzureCompute.yml
```diff
@@ -0,0 +1,1085 @@
...
+      isArray: false
+      name: virtual_machine_name
+      required: true
+      secret: false
+    - auto: PREDEFINED
+      default: false
+      description: Location to create the VM
+      isArray: false
+      name: location
+      predefined:
```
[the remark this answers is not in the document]
avidan-H: I believe so, yes.

Integrations/integration-AzureCompute.yml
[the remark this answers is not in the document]
Itay4: If there is no reasonable logic to put one, then no.

Integrations/integration-AzureCompute.yml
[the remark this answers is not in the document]
Itay4: How many are there? what will happen if user wants to user an image which is not here?

Integrations/integration-AzureCompute.yml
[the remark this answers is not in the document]
Itay4: Yes, this will be caught anyway by flake8 once you seperate the files

Integrations/integration-AzureCompute.yml
Itay4: `return_error()` already calls `LOG`, so no need to do it in integration code

Integrations/AzureCompute/AzureCompute.py
```diff
@@ -0,0 +1,906 @@
...
+            return_error('Error: {}\nDescription:{}'.format(response.get('title'), response.get('detail')))
+        sub_id = response.get('subscription_id')
+        demisto.setIntegrationContext({
+            'token': response.get('token'),
+            'stored': epoch_seconds(),
+            'subscription_id': sub_id
+        })
+        return sub_id
+    except ValueError as e:
+        if e.message == "No JSON object could be decoded":
```
Itay4: what happens in case a `ValueError` raised, which is not this one ?

Integrations/AzureCompute/AzureCompute.py
```diff
@@ -0,0 +1,906 @@
...
+            return
+    headers = {
+        'Authorization': TOKEN,
+        'Accept': 'application/json'
+    }
+    token_retrieval_url = 'https://demistobot.demisto.com/azurecompute-token'  # disable-secrets-detection
+    parameters = {'tenant': TENANT_ID, 'product': 'AzureCompute'}
+    r = requests.get(token_retrieval_url, headers=headers, params=parameters, verify=USE_SSL)
+    if r.status_code not in {200, 201}:
+        return_error('Error in authentication with the application. Please check the credentials.')
```
Itay4: remove `Please` please

Integrations/AzureCompute/AzureCompute.py
```diff
@@ -0,0 +1,906 @@
...
+    try:
+        response = r.json()
+    except ValueError:
+        return_error('The response from Demistbot was not a json serializable object.')
```
Itay4: that's not a good error message..

PR #3026, opened 2019-02-17
Committing New Integration CanaryTools

mayagoldb:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/14319

## Description
Adding a new integration to Canary Tools Deception product.
By presenting itself as an apparently benign and legitimate service(s), the canary draws the attention of unwanted activity. When someone trips one of the Canary’s triggers, an alert is sent to notify the responsible parties so that action can be taken before valuable systems in your network are compromised.

   - No

## Must have
- [x] Tests
- [x] Code Review
- [ ] https://github.com/demisto/content-test-conf/pull/329 - Added credentials to conf.json


Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+description: By presenting itself as an apparently benign and legitiamte service(s), the canary draws the attention of
+  unwanted activity. When someone trips one of the Canary's triggers, an alert is sent to notify the responsible
+  parties so that action can be taken before valubale systems in your network are compromised.
+configuration:
+- display: Fetch incidents
```
Itay4: fetch incident params should be below insecure param

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+  name: isFetch
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Incident type
+  name: incidentType
+  defaultvalue: ""
+  type: 13
+  required: false
+- display: Canary Server URL
```
Itay4: please add example

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    ''' GLOBALS'''
+
+    SERVER = demisto.params().get('server').rstrip('/') + '/api/v1/'
+    AUTH_TOKEN = demisto.params().get('auth_token')
+    UNSECURED = demisto.params().get('insecure')
+    FETCH_DELTA = 24
```
Itay4: this should be configurable, please add parameter for it

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+    RELEVANT_TOKEN_ENTRIES = {
+        'canarytoken': 'CanaryToken',
+        'created_printable': 'CreatedTime',
+        'enabled': 'Status',
+        'kind': 'Kind',
+        'triggered_count': 'Triggered',
+        'doc_name': 'DocName',
+        'url': 'TokenURL'
+    }
+    session = requests.session()
```
Itay4: we don't use `session` by standard, unless there is good reason for it.
please `http_request()` function and use Python requests without session in it

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+  name: incidentType
+  defaultvalue: ""
+  type: 13
+  required: false
+- display: Canary Server URL
+  name: server
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: API auth token
```
Itay4: please change to `API Authentication Token` add to `detailed instructions` how to get this token

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+        del os.environ['HTTP_PROXY']
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    ''' GLOBALS'''
+
+    SERVER = demisto.params().get('server').rstrip('/') + '/api/v1/'
+    AUTH_TOKEN = demisto.params().get('auth_token')
+    UNSECURED = demisto.params().get('insecure')
```
Itay4: handle this param here completely, by adding `not` instead of adding it in every request

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+
+    def get_alerts(last_fetch=None):
+        # Retrieve all unacknowledged alerts from Canary Tools
+        params = {
+            'auth_token': AUTH_TOKEN,
+        }
+
+        if last_fetch:
+            params['newer_than'] = last_fetch
+        res = session.request('GET', SERVER + 'incidents/unacknowledged', params=params, verify=not UNSECURED).json()
```
Itay4: this `.json()` is risky, need to handle it more carefully
when you add the `http_request` function, you can take an example from other integrations for that

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+        res = session.request('GET', SERVER + 'ping', params=params, verify=not UNSECURED).json()
+        if res.get('result') == 'success':
+            demisto.results('ok')
+        else:
+            return_error("Test Failed")
+
+
+    def list_canaries():
+        # Retrieve all Canaries available in Canary Tools
+        params = {
```
Itay4: since this is used in all requested, make this a constant at top of code

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+            'auth_token': AUTH_TOKEN
+        }
+
+        res = session.request('GET', SERVER + 'devices/all', params=params, verify=not UNSECURED)
+        res_json = yaml.safe_load(res.text)
+        new_devices = [
+            {new_key: device[old_key] if old_key in device else None for old_key, new_key in RELEVANT_DEVICE_ENTRIES.items()} for
```
Itay4: lets split this to a few logic lines in order to get a better readable code

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+
+
+    def list_canaries_command():
+        res_json, new_devices = list_canaries()
+        demisto.results({
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['json'],
+            'Contents': res_json,
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown('Canary Devices', new_devices,
```
Itay4: set the headers in another line before the entry, this line is way too long

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+                                             headers=['ID', 'Name', 'Description', 'Address', 'Status', 'Location', 'Version',
+                                                      'LastSeen', 'LastUpdated']),
+            'EntryContext': {'CanaryTools.Device(val.ID && val.ID == obj.ID)': new_devices}
```
Itay4: dont initialize context in the entry, but before
use `createContext` function

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+            'ContentsFormat': formats['json'],
+            'Contents': res,
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': 'File Fetched Successfully!',
+            'EntryContext': {
+                'CanaryTools.Tokens(val.CanaryToken && val.CanaryToken == obj.CanaryToken)': res.get('token').get('canarytoken')}
+        }
+
+        if res.get('token').get('doc'):
+            name = res.get('token').get('doc_name')
```
Itay4: you should `{}` to be the default of the first `.get()`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+
+    def check_whitelist_command():
+        ip = demisto.args().get('ip')
+        port = demisto.args().get('port')
+        res = check_whitelist(ip, port)
+
+        if not port:
+            port = 'Any'
+        context = {
+            'Address': '{}'.format(ip),
```
Itay4: why is `format` needed here? just put in the string

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+            params['newer_than'] = last_fetch
+        res = session.request('GET', SERVER + 'incidents/unacknowledged', params=params, verify=not UNSECURED).json()
+        alerts = res.get('incidents')
+        return alerts
+
+
+    def create_incident(alert):
+        # Turns an alert from Canary Tools to the incident structure in Demisto
+        incident = {
+            'name': alert.get('description').get('description'),
```
Itay4: can you please provide an example of an incident name ?

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+
+    def list_tokens_command():
+        res_json, new_tokens = list_tokens()
+        demisto.results({
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['json'],
+            'Contents': res_json,
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown('Canary Tools Tokens', new_tokens, headers=sorted(new_tokens[0].keys())),
+            'EntryContext': {'CanaryTools.Tokens(val.CanaryToken && val.CanaryToken == obj.CanaryToken)': new_tokens}
```
Itay4: context keys should be singular, that is `Tokens` -> `Token`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+    - contextPath: CanaryTools.IP.Port
+      description: Destination port for the IP whitelist
+      type: string
+    - contextPath: CanaryTools.IP.Whitelisted
+      description: Is the IP address whitelisted (true/false)
+      type: boolean
+    description: Whitelist an IP address in Canary
+  - name: canarytools-edit-alert-status
+    arguments:
+    - name: alertID
```
Itay4: change to `alert_id`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+      required: true
+      description: 'Alert ID (i.e : incident:canarytoken:d6fe0ae4dfd36cc3cc6d9d4f::1548593719)' # disable-secrets-detection
```
Itay4: please add to description where this alert id can be retreived from

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+      type: string
+    - contextPath: CanaryTools.Alert.Status
+      description: Alert Status
+      type: string
+    description: Edit the status for an Alert in Canary Tools
+  - name: canarytools-get-token
+    arguments:
+    - name: token
+      required: true
+      description: Canary Token
```
Itay4: how can this token be retrieved ?

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+    outputs:
+    - contextPath: CanaryTools.Tokens.CanaryToken
+      description: Canary Token
+      type: string
+    - contextPath: File.Size
+      description: File Size
+      type: string
+    - contextPath: File.SHA1
+      description: File SHA1
```
Itay4: `SHA1` -> `SHA-1`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+      type: string
+    - contextPath: File.SHA256
+      description: File SHA256
```
Itay4: `SHA256` -> `SHA-256`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+    - contextPath: File.MD5
+      description: File MD5
+      type: string
+    - contextPath: File.Extension
+      description: File Extension
+      type: string
+    description: Fetch Canary Token file from the Canary Tools server
+  isfetch: true
+  runonce: false
+releaseNotes: "Added new integration."
```
Itay4: no need for rn in new file

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+                demisto.results({
+                    'Type': entryTypes['note'],
+                    'ContentsFormat': formats['json'],
+                    'Contents': res,
+                    'ReadableContentsFormat': formats['markdown'],
+                    'HumanReadable': 'The Alert {} was '.format(alert) + res.get('action'),
+                    'EntryContext': {'CanaryTools.Alert(val.ID && val.ID == obj.ID)': context}
+                })
+            else:
+                demisto.results('Unsopported command')
```
Itay4: please return error here

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+            'Port': '{}'.format(port),
+            'Whitelisted': '{}'.format(res.get('is_ip_whitelisted'))
+        }
+        if res.get('is_ip_whitelisted'):
+            demisto.results({
+                'Type': entryTypes['note'],
+                'ContentsFormat': formats['json'],
+                'Contents': res,
+                'ReadableContentsFormat': formats['markdown'],
+                'HumanReadable': 'The IP address ' + '{}'.format(ip) + ':' + '{}'.format(port) + ' is Whitelisted!',
```
Itay4: please remove `!`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+
+
+    def whitelist_ip_command():
+        ip = demisto.args().get('ip')
+        port = demisto.args().get('port')
+        res = whitelist_ip(ip, port)
+
+        if not port:
+            port = 'Any'
+        if res.get('result') == 'success':
```
Itay4: first initialize `res.get('result')` in an argument and then use it in the conditions

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+            })
+        elif res.get('result') == 'failure':
+            demisto.results({
+                'Type': entryTypes['note'],
+                'ContentsFormat': formats['json'],
+                'Contents': res,
+                'ReadableContentsFormat': formats['markdown'],
+                'HumanReadable': res.get('message')
+            })
+        elif res.get('result') == 'error':
```
Itay4: shouldnt we return error here?

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+                'incident': alert
+            }
+            res = session.request('POST', SERVER + 'incident/unacknowledge', params=params, verify=not UNSECURED).json()
+            if res.get('action') == 'unacknowledged':
+                demisto.results({
+                    'Type': entryTypes['note'],
+                    'ContentsFormat': formats['json'],
+                    'Contents': res,
+                    'ReadableContentsFormat': formats['markdown'],
+                    'HumanReadable': 'The Alert {} was '.format(alert) + res.get('action'),
```
Itay4: use `format` instead of `+`

Integrations/integration-Thinkst_Canary.yml
```diff
@@ -0,0 +1,558 @@
...
+        if last_fetch is None:
+            last_fetch = parse_date_range('{} hours'.format(str(FETCH_DELTA)), '%Y-%m-%d-%H:%M:%S')[0]
+        alerts = get_alerts(last_fetch)
+        incidents = []
+        next_fetch = last_fetch
+
+        for alert in alerts:
+            current_fetch = 1000 * (int(alert.get('description').get('created')))
+            current_fetch = timestamp_to_datestring(current_fetch, '%Y-%m-%d-%H:%M:%S')
+            if current_fetch > last_fetch:
```
Itay4: why is this check needed? we query only for alerts with the relevant timestamp

mayagoldb:
image

Integrations/integration-Thinkst_Canary.yml
mayagoldb: This is by design, this type of product will not normally have more than 10 max incidents per day, so no reason to make it configurable. We want to make the instance configuration as user-friendly as possible.

Integrations/integration-Thinkst_Canary.yml
mayagoldb: what do you mean

Integrations/integration-Thinkst_Canary.yml
mayagoldb: if you're talking about `if res.get('token').get('doc'):` then the default return would be `None` in case there's no result, so it will not go inside the condition, as it's supposed to.

Integrations/integration-Thinkst_Canary.yml
mayagoldb: Added an example to docstrings

Integrations/integration-Thinkst_Canary.yml
Itay4: not sure what's the point here..
if the user wants to fetch alerts from the last day / week / month, we should let him/her
the default can be 24 hours, and then it does not even impact the ux
this became our convention after understanding that we thought to be the "normal" is not exactly that
please do make it configurable

Integrations/integration-Thinkst_Canary.yml
Itay4: i meant that it if `res.get('token')` return `None` then an AttributeError would raise since NoneType object got no attribute `get`

Integrations/integration-Thinkst_Canary.yml
Itay4: this should like `'EntryContent' : context` and `context` should be initialized with that dict you have there before the entry
also make sure to use `===` operator and not `==`

mayagoldb:
1. Documentation issue: https://github.com/demisto/etc/issues/15742
2. Added docstrings
3. Handled instance configuration errors

mayagoldb:
image

Integrations/ThinkstCanary/ThinkstCanary.yml
```diff
@@ -0,0 +1,209 @@
...
+  type: 8
+  required: false
+- display: Fetch incidents
+  name: isFetch
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Hours before first fetch to retrieve incidents
+  name: fetchDelta
+  defaultvalue: "24"
```
Itay4: this is not according to our convention, we should allow some thing like `24 hours` as we do in other integrations
have a look in recorded future as an example

Integrations/ThinkstCanary/ThinkstCanary.yml
```diff
@@ -0,0 +1,209 @@
...
+  type: 0
+  required: false
+- display: Use system proxy settings
```
Itay4: this param should be above the fetch params

Integrations/ThinkstCanary/ThinkstCanary.yml
```diff
@@ -0,0 +1,209 @@
...
+      description: Device last seen time
+      type: date
+    - contextPath: CanaryTools.Device.LastUpdated
+      description: Device last updated time
+      type: date
+    description: Lists all registered Canaries.
+  - name: canarytools-list-tokens
+    arguments: []
+    outputs:
+    - contextPath: CanaryTools.token.CanaryToken
```
Itay4: `token` -> `Token`

Integrations/ThinkstCanary/ThinkstCanary.yml
```diff
@@ -0,0 +1,209 @@
...
+display: Thinkst Canary
+category: Deception
+description: By presenting itself as an apparently benign and legitimate service(s), the Canary draws the attention of
+  unwanted activity. When someone trips one of the Canary's triggers, an alert is sent to notify the responsible
+  parties so that action can be taken before valubale systems in your network are compromised.
+detaileddescription: |-
+  Generate your API Authentication Token from Canary Tools:
+  https://help.canary.tools/knowledge_base/topics/forget-logging-in-what-apis-does-canary-support
+configuration:
+- display: Incident type
```
Itay4: this should be at the bottom

Integrations/ThinkstCanary/ThinkstCanary.py
```diff
@@ -0,0 +1,429 @@
...
+
+def fetch_incidents_command():
+    """
+    Fetch alerts from Canary Tools as incidents in Demisto
+    last_fetch: The latest fetched alert creation time
+    """
+    last_fetch = demisto.getLastRun().get('time')
+
+    if last_fetch is None:
+        last_fetch = parse_date_range('{} hours'.format(str(FETCH_DELTA)), '%Y-%m-%d-%H:%M:%S')[0]
```
Itay4: `hours` shouldn't be hard coded

Integrations/ThinkstCanary/ThinkstCanary.py
```diff
@@ -0,0 +1,429 @@
...
+
+def get_alerts(last_fetch=None):
+    """
+    Retrieve all unacknowledged alerts from Canary Tools
+    :param last_fetch: Last fetch incidents time
+    """
+
+    if last_fetch:
+        params = {
+            'auth_token': AUTH_TOKEN,
```
Itay4: why are we passing the `auth_token` in every request function as part of params? just add it in the http_request

Integrations/ThinkstCanary/ThinkstCanary.py
```diff
@@ -0,0 +1,429 @@
...
+        'LastSeen',
+        'LastUpdated'
+    ]
+    demisto.results({
+        'Type': entryTypes['note'],
+        'ContentsFormat': formats['json'],
+        'Contents': res_json,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown('Canary Devices', new_devices, headers=headers),
+        'EntryContext': {'CanaryTools.Device(val.ID && val.ID === obj.ID)': context}
```
Itay4: please dont initialize the entry context object in the entry, but before

Integrations/ThinkstCanary/ThinkstCanary.yml
```diff
@@ -0,0 +1,209 @@
...
+  name: proxy
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Fetch incidents
+  name: isFetch
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Hours before first fetch to retrieve incidents e.g., 12 hours, 7 days, 3 months, 1 year
```
Itay4: please change display to: `First fetch time (<number> <time unit>, e.g., 12 hours, 7 days, 3 months,
    1 year)`

PR #3032, opened 2019-02-18
Fixed typos (Alex)

ShahafBenYakir:
Fixed typos in description and in output (Alex)


yuvalbenshalom:
@ShahafBenYakir @kirbles19 next time we will need to make sure this is ready before the merge of the integration, because now we will need to break the code freez.

PR #3034, opened 2019-02-18
updating id_set

ronykoz:


Tests/id_set.json
```diff
@@ -194,6 +200,9 @@
                     "abuseipdb-get-blacklist", 
                     "createNewIndicator"
                 ], 
+                "script_executions": [
+                    "createNewIndicator"
```
anara123: why there are `createNewIndicator` two times?

ronykoz: Because it matches the pattern of a script, should be much of a difference if we remove it, i think it is insignificant

PR #3048, opened 2019-02-20
small ArcSight ESM fix

orenzohar:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15685#event-2152513591

## Description
small yml fix
save auth token in context for further use.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Integrations/ArcSightESM/ArcSightESM.py
```diff
@@ -723,7 +709,29 @@ def get_all_query_viewers_command():
         demisto.results('No Query Viewers were found')
 
 
-AUTH_TOKEN = login()
+def validate_auth_token():
```
Itay4: you are not calling this function..

PR #3050, opened 2019-02-21
Change content version to 19.2.2

Itay4:


anara123:
@Itay4 don't you need to change git sha-1?

PR #3052, opened 2019-02-21
PhishMe test fix

DeanArbel:
## Status
Ready

## Description
Switched tested malicious IP to another malicious IP.


BenJoParadise:
Issue was present in current content package, fixed in master.

PR #3057, opened 2019-02-22
dupes fix

orenzohar:
## Status
done

## Related Issues
fixes: https://github.com/demisto/etc/issues/15708
fixes: https://github.com/demisto/etc/issues/15822
## Description
fixes Pulling duplicate cases 

   - No

## Must have
- [x] Tests
- [x] Documentation https://github.com/demisto/etc/issues/15518
- [x] Code Review


Integrations/ArcSightESM/ArcSightESM.py
```diff
@@ -296,7 +295,7 @@ def fetch():
 
     last_run = demisto.getLastRun()
     last_create_time = last_run.get('last_create_time', 0)
-    already_fetched = deque(last_run.get('already_fetched', []), maxlen=1000)
+    already_fetched = last_run.get('already_fetched', [])
```
anara123: last run will break backward compatibility
the old structure was:
```
{
    "value": {
           "caseLastEventTime": maxEventTime (in  millis),
           "lastEventTime": maxEventTime(in millis)
     }
}
```

Integrations/ArcSightESM/ArcSightESM.py
```diff
@@ -639,11 +678,12 @@ def get_entries_command():
             'Type': entryTypes['note'],
             'ContentsFormat': formats['json'],
             'Contents': entry_context,
-            'HumanReadable': tableToMarkdown(name='Active List entries', t=filtered, removeNull=True),
+            'HumanReadable': tableToMarkdown(name='Active List entries: {}'.format(resource_id), t=filtered,
```
anara123: replace all demisto.results with `return_outputs`

Integrations/ArcSightESM/ArcSightESM.py
```diff
@@ -730,9 +771,9 @@ def get_all_query_viewers_command():
         demisto.results('ok')
 
     elif demisto.command() == 'as-fetch-incidents' or demisto.command() == 'fetch-incidents':
-        if demisto.args().get('lastRun'):
-            last_run = json.loads(demisto.args().get('lastRun'))
-            demisto.setLastRun(last_run)
+        # if demisto.args().get('lastRun'):
```
anara123: remove comment

Integrations/ArcSightESM/ArcSightESM.py
orenzohar: I had a warper function for `demisto.results` similar to `return_outputs`, and you told me to remove it as it's easier to debug when you see all outputs in a straight forward way like here.

Integrations/ArcSightESM/ArcSightESM.yml
```diff
@@ -319,6 +319,13 @@ script:
     - name: id
       required: true
       description: Resource ID of the query viewer
+    - name: onlyColumns
```
anara123: remove this arg. this arg was added in as-get-matrixData because of backward compatibility reasons.
We should always return the table with columns in single entry

orenzohar: yeah, I added this following a client's request. I can set the default to always return the table with columns in a single entry (onlyColumns=False). would that be ok?

PR #3065, opened 2019-02-24
DemistoUploadFileToIncident, DemistoUploadFile - add body argument

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15704

   - No

## Must have
- [x] Tests
- [x] Documentation - scripts - no docs
- [ ] Code Review


anara123:
@bakatzir what is the difference between those two scripts? and why they are not in the Demisto integration?

bakatzir:
I believe they can both be deprecated and can added as a command to the integration.

anara123:
@bakatzir what is the difference between those two scripts?

bakatzir:
@anara123 don't see any difference.

anara123:
there is some difference. The url is different , please check what is the exact difference, so we could update the description of the script

'entry/upload/' + args.incidentID

"incident/upload/{}".format(incident_id)

bakatzir:
u5
u6



u7
u9

anara123:
@dorsha what is the difference between those two routes?
`entry/upload/:id` and `incident/upload/:id`

@Itay4 I see that you created the `DemistoUploadFileToIncident` what was the reason? why not to use DemistoUploadFile

dorsha:
Will answer in a private message.

Itay4:
@anara123 don't really remember

yuvalbenshalom:
@anara123 how do you wish we will go forward with this?

bakatzir:
Added meaningful description consulting with @kirbles, and test playbooks.

yardensade:
can we close this branch?

PR #3069, opened 2019-02-25
Vertica

yardensade:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15727

## Description
New vertica integration to query vertica DBs

## Screenshots
image

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Beta_Integrations/integration-Vertica.yml
+++ b/Beta_Integrations/integration-Vertica.yml
@@ -94,7 +94,7 @@ script:
         # Get arguments from user
         query = demisto.args().get('query')
         limit = int(demisto.args().get('limit', 50))
-        # Query and get raw response (list of tuples)
+        # Query and get raw response (list of ordered dicts)
         rows = query_request(query)
         # Parse response into context & content entries
         if rows:
@@ -124,7 +124,7 @@ script:
     def query_request(query):
         try:
             CURSOR.execute(query)
-        except vertica_python.errors.MissingRelation as err:
+        except vertica_python.errors.MissingRelation:
             return_error('Error while executing query.')
         rows = CURSOR.fetchall()
         if CURSOR.rowcount in {0, -1}:
--- a/Beta_Integrations/integration-Vertica.yml
+++ b/Beta_Integrations/integration-Vertica.yml
@@ -90,7 +90,6 @@ script:
         # Init main vars
         contents = []
         context = {}
-        context_entries = []
         title = 'Vertica Query'
         # Get arguments from user
         query = demisto.args().get('query')
--- a/Beta_Integrations/integration-Vertica.yml
+++ b/Beta_Integrations/integration-Vertica.yml
@@ -168,7 +168,7 @@ script:
       description: Rows content
       type: string
     description: Query the database
-  dockerimage: devdemisto/vertica:1.0.0.149
+  dockerimage: demisto/vertica:1.0.0.150
   runonce: false
 releaseNotes: "New integration with query ability on Vertica DB"
 tests:
--- a/Beta_Integrations/integration-Vertica.yml
+++ b/Beta_Integrations/integration-Vertica.yml
@@ -1,0 +1,173 @@
+commonfields:
+  id: Vertica
+  version: -1
+name: Vertica
+display: Vertica
+category: Database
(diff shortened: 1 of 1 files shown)


Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+      defaultValue: "50"
+    outputs:
+    - contextPath: Vertica.Query.Query
+      description: Original query
+      type: string
+    - contextPath: Vertica.Query.Rows
+      description: Rows content
+      type: string
+    description: Query the database
+  dockerimage: devdemisto/vertica:1.0.0.149
```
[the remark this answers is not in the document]
yardensade: already did in last commit

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+            'HumanReadable': tableToMarkdown(title, contents, removeNull=True),
+            'EntryContext': context
+        })
+
+
+    def query_request(query):
+        try:
+            CURSOR.execute(query)
+        except vertica_python.errors.MissingRelation:
+            return_error('Error while executing query.')
```
ronykoz: Maybe add the real error as well?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['json'],
+            'Contents': contents,
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown(title, contents, removeNull=True),
+            'EntryContext': context
+        })
+
+
+    def query_request(query):
```
ronykoz: not a good name, maybe `get_query_execution`

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+        # Get arguments from user
+        query = demisto.args().get('query')
+        limit = int(demisto.args().get('limit', 50))
+        # Query and get raw response (list of ordered dicts)
+        rows = query_request(query)
+        # Parse response into context & content entries
+        if rows:
+            if limit:
+                rows = rows[:limit]
+            title = 'Vertica Query Results:'
```
ronykoz: you override the old title?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+        """
+        # Init main vars
+        contents = []
+        context = {}
+        title = 'Vertica Query'
+        # Get arguments from user
+        query = demisto.args().get('query')
+        limit = int(demisto.args().get('limit', 50))
+        # Query and get raw response (list of ordered dicts)
+        rows = query_request(query)
```
ronykoz: do you mean db_row?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+            'database': DATABASE,
+            'connection_timeout': 5
+        }
+        try:
+            connection = vertica_python.connect(**db_params)
+        except vertica_python.errors.ConnectionError:
+            return_error('Could not connect to DB, please re-check DB params.')
+        return connection
+
+    CONNECTION = connect_db()
```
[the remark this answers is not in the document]
yardensade: I need to save the connection in order to later close it.

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+
+
+    def query_command():
+        """
+        Execute a query against the DB
+        """
+        # Init main vars
+        contents = []
+        context = {}
+        title = 'Vertica Query'
```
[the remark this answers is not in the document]
yardensade: this will be changed according to if we have results or not tho, so it gets changed later on

Beta_Integrations/integration-Vertica.yml
yardensade: tried, it has issues accessing the message itself and throws exception in exception, so I kept it like that for now

Beta_Integrations/integration-Vertica.yml
yardensade: indeed, depends on if we have results or not

Beta_Integrations/integration-Vertica.yml
yardensade: i mean the rows returned from the db

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,175 @@
...
+        """
+        Parses datetime object into string
+        """
+        if isinstance(v, datetime.datetime):
+            return v.strftime("%Y-%m-%dT%H:%M:%S")
+        return v
+
+
+    def connect_db():
+        db_params = {
```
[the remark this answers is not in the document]
yardensade: I'm not sure it is needed since it is used in a single place anyway, a bit redundent

Beta_Integrations/integration-Vertica.yml
yardensade: Simply keeping our current integration conventions do avoid confusing people who might use it

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+description: Analytic database management software
+configuration:
+- display: Host (Vertica Server URL)
```
Itay4: please add example

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+  name: url
+  defaultvalue: myhost.example.com
```
Itay4: we should add default value only if it makes sense, please remove

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+  type: 0
+  required: true
+- display: Database
+  name: database
+  defaultvalue: mydb
```
Itay4: we should add default value only if it makes sense, please remove

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+description: Analytic database management software
+configuration:
+- display: Host (Vertica Server URL)
+  name: url
+  defaultvalue: myhost.example.com
+  type: 0
+  required: true
+- display: Database
```
Itay4: i would change to `Database name`, wdyt ?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        'port': PORT,
+        'user': USERNAME,
+        'password': PASSWORD,
+        'database': DATABASE,
+        'connection_timeout': 5
+    }
+
+    ''' HELPER FUNCTIONS '''
+
+    def convert_datetime_to_string(v):
```
Itay4: don't we have something like that im common server?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        """
+        Parses datetime object into string
+        """
+        if isinstance(v, datetime.datetime):
+            return v.strftime("%Y-%m-%dT%H:%M:%S")
```
Itay4: please use `'`

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        return v
+
+
+    def connect_db():
+        try:
+            connection = vertica_python.connect(**DB_PARAMS)
+        except vertica_python.errors.ConnectionError:
+            return_error('Could not connect to DB, please re-check DB params.')
```
Itay4: please remove `please`

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        return connection
+
+
+    ''' COMMANDS + QUERY FUNCTIONS '''
+
+    def test_module():
+        """
+        Performs basic query on default system tables
+        """
+        cursor.execute("SELECT * FROM system_tables ORDER BY table_schema, table_name LIMIT 2;")
```
Itay4: please use `'`

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        rows = cursor.fetchall()
+        if cursor.rowcount == 0:
+            return_error('No results were returned from the DB.')
```
Itay4: why is this an error ?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        Execute a query against the DB
+        """
+        # Init main vars
+        contents = []
+        context = {}
+        title = 'Vertica Query'
+
+        # Get arguments from user
+        query = demisto.args().get('query')
+        limit = int(demisto.args().get('limit', 50))
```
Itay4: not sure why we need this argument
if they would like to limit the results they can just add `LIMIT` to the query
wdyt ?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        # Query and get raw response (list of ordered dicts)
+        rows = query_request(query)
+
+        # Parse response into context & content entries
+        if rows:
+            if limit:
+                rows = rows[:limit]
+            title = 'Vertica Query Results:'
```
Itay4: please remove `:`

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+
+            for i, row in enumerate(rows):
+                rows[i] = { underscoreToCamelCase(k): convert_datetime_to_string(v) for k, v in row.items() }
+
+            contents = rows
+            context['Vertica.Query(val.Query && val.Query === obj.Query)'] = {
+                'Query': query,
+                'Rows': rows
```
Itay4: please use `Row` instead of `Rows` as our context standard convention is in singular

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+        demisto.results({
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['json'],
+            'Contents': contents,
+            'ReadableContentsFormat': formats['markdown'],
+            'HumanReadable': tableToMarkdown(title, contents, removeNull=True),
```
Itay4: i dont like the way tabletomarkdown handle no results, as it looks weird
please just return `No results found` in such case

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+            'EntryContext': context
+        })
+
+
+    def query_request(query):
+        try:
+            cursor.execute(query)
+        except vertica_python.errors.MissingRelation:
+            return_error('Error while executing query.')
```
Itay4: saw your reply to rony, but are there any details we can add here?
not sure what to do when such an error occurs ?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+        rows = cursor.fetchall()
+        if cursor.rowcount in {0, -1}:
```
Itay4: please add some comment here

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+            test_module()
+        elif demisto.command() == 'vertica-query':
+            query_command()
+    # Log exceptions
+    except Exception as e:
+        LOG(e)
+        LOG.print_log()
+        raise
+    finally:
+        if 'connection' in globals():
```
Itay4: in which case this condition will be `False` ?

Beta_Integrations/integration-Vertica.yml
```diff
@@ -0,0 +1,178 @@
...
+  - name: vertica-query
+    arguments:
+    - name: query
+      required: true
+      description: SQL query to perform on DB
+    - name: limit
+      description: Limit number of results to be returned from the query
+      defaultValue: "50"
+    outputs:
+    - contextPath: Vertica.Query.Query
```
Itay4: not sure i understand why we need two levels here
I think `Vertica.Query` and `Vertica.Row` should do
wdyt ?

Beta_Integrations/integration-Vertica.yml
yardensade: In the DB/vertica lingo the database is db name, so they are essentially the same, personally I don't see any difference, so any way goes.

Beta_Integrations/integration-Vertica.yml
yardensade: There is some issue with the SDK itself when trying to extract that info, it claims the data is not accessible, I have trying to diagnose that using dir()/getattr() on the MissingRelation object, but it simply didn't work.

Beta_Integrations/integration-Vertica.yml
yardensade: In case connection var failed to instantiate properly, aka no connection was formed and Exception was thrown as result.

Beta_Integrations/integration-Vertica.yml
yardensade: Mainly to differentiate in case we add more context paths to it in the future, I thought better safe then sorry.

yardensade: But since it's main use is query, we could remove the second level without worrying too much.

Itay4: i think we can remove, its db, so there isn't much more than queries here

Itay4:
@yardensade what are the error messages raised in case of wrong params (url, db name) inserted ?

Beta_Integrations/integration-Vertica.yml
yardensade: we have ts to ds & date to ts, I needed dt to ds.

Beta_Integrations/integration-Vertica.yml
yardensade: > @yardensade what are the error messages raised in case of wrong params (url, db name) inserted ?

`Failed to establish a connection to the primary server or any backup address.`

yardensade:
> @yardensade what are the error messages raised in case of wrong params (url, db name) inserted ?

`Failed to establish a connection to the primary server or any backup address.`

Beta_Integrations/integration-Vertica.yml
yardensade: It currently just says "Vertica Query
No entries."
Maybe it's time I had custom error message to TTMD function? would be rather useful :)

Beta_Integrations/integration-Vertica.yml
yardensade: wanted to add that as example under display actually, my bad

Beta_Integrations/integration-Vertica.yml
yardensade: These are default system tables that essentially must be there, if they aren't this is a red flag that something is wrong.

Beta_Integrations/integration-Vertica.yml
yardensade: Sort of fail safe in case someone is going to query something he is about to regret without considering it in advance. Which from my experience tends to happen.

yardensade:
Tested with our instance after changes.

PR #3077, opened 2019-02-26
Disable defaultEnabled integrations

bakatzir:
## Status

## Related Issues
fixes: https://github.com/demisto/etc/issues/15248

   - No

## Must have
- [x] Tests
- [x] Code Review


Integrations/integration-UrlScan.yml
```diff
@@ -6,6 +6,7 @@ display: urlscan.io
...
 description: Urlscan.io reputation
+defaultEnabled: false
```
glicht: Kind of weird that we are disabling this and it didn't have "defaultEnabled:true" to start with. Please check with Server what logic was enabling this and if this is ok.

bakatzir: @glicht 
there was no server logic that was enabling it, it's not enabled by default, so basically this change is redundant

bakatzir:
@Itay4 when making the release, please let me know so i'll re-check we don't break existing instances (see shriki's comment from the fixed issue)

PR #3078, opened 2019-02-26
ServiceNow  - Handle customfields + fetch errors

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/15638


liorblob:
@Itay4
1. All the relevant functions call the `get_body` function so it only needs to be addressed there
2. Didn't quite understand, but there are cases where there is a `return_error` inside a command. However, from `fetch_incidents` there should be no paths with `return_error`.

PR #3079, opened 2019-02-26
Archer - added new command to reset integration cache

hirshbergshachar:


Integrations/integration-ArcherRSA.yml
```diff
@@ -3682,9 +3688,13 @@ script:
     - name: fieldID
       required: true
       description: Field Id
-    description: Returns a list of values for a specified field, e.g., fieldID=16114. This
-      command only works for value list fields (type 4).
+    description: Returns a list of values for a specified field, e.g., fieldID=16114.
+      This command only works for value list fields (type 4).
+  - name: archer-reset-cache
+    arguments: []
+    description: Reset Archer's integration cache
```
Itay4: maybe add a description when this should be done ?

hirshbergshachar:
> can we add to error messages that raise up, to use this function, when it's relevant ?

Not really.

Integrations/integration-ArcherRSA.yml
hirshbergshachar: Updated

PR #3096, opened 2019-03-03
update instance test

ronykoz:


anara123:
@ronykoz is the instance is alive if there are failed tests? if yes, then the integrations instances need to be enabled and configured

ronykoz:
@anara123 yes the instance will be configured and enabled if the test failed

PR #3100, opened 2019-03-03
Fixed cert verify error

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#15859

## Description
Enabled module to accept Verify Cert

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/integration-RTIR.yml
+++ b/Integrations/integration-RTIR.yml
@@ -587,7 +587,7 @@ script:
 
     ''' EXECUTION CODE '''
 
-    tracker = rt.Rt(BASE_URL, USERNAME, PASSWORD)
+    tracker = rt.Rt(BASE_URL, USERNAME, PASSWORD, verify_cert=USE_SSL)
     login = tracker.login()
     if login is False:
         return_error('Failed to login')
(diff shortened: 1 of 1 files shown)


yaakovi:
this integration needs to be rewritten without the use of the RT package

amshamah419:
@yaakovi  So for this release, should we release the fix and then prioritise the new version? I think converting to not use the rt package is going to be pretty involved.

Integrations/integration-RTIR.yml
```diff
@@ -587,7 +587,7 @@ script:
 
     ''' EXECUTION CODE '''
 
-    tracker = rt.Rt(BASE_URL, USERNAME, PASSWORD)
+    tracker = rt.Rt(BASE_URL, USERNAME, PASSWORD, verify_cert=USE_SSL)
```
glicht: Seems that you are doing the oposite of what you want. USE_SSL == true when unsecure is checked....

amshamah419: Just checked this, It's using `USE_SSL = not demisto.params().get('unsecure', False)` To invert the user's selection. So if a user selects "Trust any certificate (Unsecure)" the parameter is `True`, but then is inverted by the `not` statement which means the verify_cert variable would equal `False`.

If the user selects `False` for "Trust any cert", it would be inverted to `True` which would mean the verify_cert would equal `True`.

It's a really strange implementation for this type of logic...

PR #3101, opened 2019-03-03
make whois lookup option controlled by a parameter

glicht:
## Status
Ready
## Related Issues
fixes: https://github.com/demisto/etc/issues/15828

## Description
provide whois argument to dnstwist command

   - No: Note whois was being done automatically but it seems that it doesn't work for most domains. So changing this to default false should not impact users.

## Must have
- [x] Tests
- [x] Documentation: release notes
- [ ] Code Review

## Additional changes
Added a verification at test that we can run dnstwist.


glicht:
yes we will do default "no" as it seems that whois almost never works.

PR #3136, opened 2019-03-07
fix reg path reputation output + matching improvement

glicht:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15988
fixes: https://github.com/demisto/etc/issues/15992

## Description
See the issue.

   - No

## Must have
- [x] Tests
- [ ] Code Review


aviadl:
@glicht dont you want to check both paths ? shortened and full ?

glicht:
@aviadl Yes we cover both. We first normalize to full path and then check against the full.

PR #3142, opened 2019-03-07
Removed demisto.log from cut script

yardensade:
Due to server bug it was being parsed as entry causing issues with context

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16000

## Description
Cut transformer was resulting in 2 entries instead of 1, giving false values as a result.

## Screenshots
image

   - Yes
       - Further details: The cut transformer would return 2 results, 1 of them is being a demisto.log in the script, which is a bug. But still some users right now could use that with another transformer to get 1 of the results.

## Must have
- [x] Tests
- [x] Code Review


Itay4:
Agree, if it wasn’t raised by a customer we don’t need RN here

yardensade:
@omercnet as you are the one probably using that transformer in the first place, could you give us an input regarding the backward compatibility before we merge this?

omercnet:
the transformer I wrote and provided the customer returned only 1 value, so I'm all good 🙄

PR #3148, opened 2019-03-08
Docker image check script

glicht:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/13998

## Description
A script to check if a specified docker image is available for "pull"

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Scripts/CheckDockerImageAvailable/CheckDockerImageAvailable.py
+++ b/Scripts/CheckDockerImageAvailable/CheckDockerImageAvailable.py
@@ -1,6 +1,9 @@
 import demistomock as demisto
 from CommonServerPython import *
 import requests
+import urllib3
+
+urllib3.disable_warnings()
 
 ACCEPT_HEADER = {
     'Accept': "application/json, "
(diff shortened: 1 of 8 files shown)


Scripts/CheckDockerImageAvailable/CheckDockerImageAvailable.py
```diff
@@ -0,0 +1,119 @@
...
+        tag = 'latest'
+        if len(split) > 1:
+            tag = split[1]
+        if tag is None:
+            tag = 'latest'
+        auth_token = docker_auth(image_name, verify_ssl, registry)
+        headers = ACCEPT_HEADER.copy()
+        if auth_token:
+            headers['Authorization'] = "Bearer {}".format(auth_token)
+        res = requests.get("https://{}/v2/{}/manifests/{}".format(registry, image_name, tag),
```
asafshen: I haven't tried it, but will this work against localhost registry?
e.i if registry is `localhost:5000`, is it support this api?

glicht: There is need for the registry to support https. I haven't tried it against a local registry. Did test it against google gcr.io and it works fine. I can do some tests with a local registry after we complete the feature.

PR #3153, opened 2019-03-10
fixing issue - ignoring check for enabled integration

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16012

   - No


liorblob:
@Shellyber What about the test?

PR #3156, opened 2019-03-10
ArcSight enhancements

orenzohar:
## Status
In Progress

## Description
few enhancements for Comcast and CyberView:
- new param for the maximum amount of unique case IDs to fetch
- decoding more enigmatic ArcSight fields into new fields
- new flag for `as-get-case-event-ids` to get correlated events
 
## Related PRs
fixes: https://github.com/demisto/etc/issues/16071

   - No

## Must have
- [x] Tests
- [x] Documentation https://github.com/demisto/etc/issues/15518
- [ ] Code Review

### DIFF
--- a/Integrations/ArcSightESMv2/ArcSightESMv2.py
+++ b/Integrations/ArcSightESMv2/ArcSightESMv2.py
@@ -688,7 +688,14 @@ def get_entries_command():
                 filtered = [entry for entry in filtered if ((entry.get(k) == v) if k else (v in entry.values()))]
 
         contents = beautifully_json(filtered)
-        outputs = {'ArcSightESM.ActiveList.{id}'.format(id=resource_id): contents}
+        ActiveListContext = {
+            'ResourceID': resource_id,
+            'Entries': contents,
+        }
+        outputs = {
+            'ArcSightESM.ActiveList.{id}'.format(id=resource_id): contents,
+            'ArcSightESM.ActiveList(val.ResourceID===obj.ResourceID)': ActiveListContext
+        }
         human_readable = tableToMarkdown(name='Active List entries: {}'.format(resource_id), t=filtered,
                                          removeNull=True)
         return_outputs(readable_output=human_readable, outputs=outputs, raw_response=contents)
--- a/Integrations/ArcSightESMv2/ArcSightESMv2.py
+++ b/Integrations/ArcSightESMv2/ArcSightESMv2.py
@@ -577,8 +577,9 @@ def update_case(case_id, stage, severity):
 @logger
 def get_correlated_events_ids(event_ids):
     related_ids = set(event_ids)
-    for raw_event in get_security_events(event_ids):
-        related_ids.add(raw_event.get('baseEventIds'))
+    for raw_event in beautifully_json(get_security_events(event_ids)):
+        if raw_event.get('baseEventIds'):
+            related_ids.add(raw_event.get('baseEventIds'))
 
     return list(related_ids)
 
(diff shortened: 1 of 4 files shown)


Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -82,9 +102,14 @@ def beautifully_json(d, depth=0):
...
-                    d.pop(key, None)
-                elif key in IP_FIELDS and isinstance(value, int):
-                    d[key] = int_to_ip(value)
+                    if remove_nones:
+                        d.pop(key, None)
+                    else:
+                        d[key] = 'None'
+
+                elif key in IP_FIELDS:
+                    key = 'decodedAddress' if key == 'addressAsBytes' else key
```
yuvalbenshalom: why do we change the key here?

Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -492,7 +518,9 @@ def get_security_events(event_ids, last_date_range=None):
         events = res_json.get('sev.getSecurityEventsResponse').get('sev.return')
         return events if isinstance(events, list) else [events]
 
-    return_error('Events are empty for some reason. Response Body: {}'.format(res.text))
+    demisto.debug(res.text)
```
yuvalbenshalom: I think its better to add string that says what is this and then connect the `res.text` content

Integrations/ArcSightESMv2/ArcSightESMv2.py
orenzohar: we are adding the decoded output into a new key to keep backward compatibility

Integrations/ArcSightESMv2/ArcSightESMv2.py
orenzohar: res.text is a huge xml traceback who don't have much meaning to customers, hence it's only logged. also, this is not an error, we will get to this line if there are no base events related to the case.

Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -549,10 +582,29 @@ def update_case(case_id, stage, severity):
...
+@logger
+def get_correlated_events_ids(event_ids):
+    related_ids = set(event_ids)
+    correlated_events = beautifully_json(get_security_events(event_ids, ignore_empty=True))
+
+    for raw_event in correlated_events:
+        if raw_event.get('baseEventIds'):
+            try:
+                related_ids.add(raw_event.get('baseEventIds'))
+            except Exception as ex:
```
yuvalbenshalom: @orenzohar lets have a short talk about this one.
I don't understand what exception do you expect here, and in any case this seems too general and/or can be replaced with an if statement.

orenzohar: if the object we want to add to the set through `add()` is not hashable (e.g., list) it will throw an Exception, and if it is not iterable (e.g., int) `update()` will throw an Exception.

orenzohar: a way to simplify this would be to use only a list all along, and return `list(set(the_list))`

orenzohar: or something like this maybe?
image

yuvalbenshalom: yes this last example from pycharm is much better in 2 ways:
1. it dose not create the risk of silencing an exception we did not expect.
2. the expected cases are clear

yuvalbenshalom:
@orenzohar are the tests of single and multi base_event_id in?

orenzohar:
> @orenzohar are the tests of single and multi base_event_id in?

yes

PR #2907, opened 2019-01-29
Changes to phishing test to prevent VT quota errors

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15281

## Description
- Changed test data for phishing - now has a malicious URL instead of SHA256
- Changed phishing test - inline and phishing test - attachment, both to use the new testdata for phishing
- Disabled VirusTotal Private API for phishing tests. Quota makes tons of tests fail

## Does it break backward compatibility?
   - No

## Must have
- [ ] Code Review

PR #3157, opened 2019-03-11
Increasing timeout for inline and attachment phishing test playbooks

Shellyber:
## Status
Ready

   - No

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -1034,7 +1034,7 @@
         },
         {
             "playbookID": "Phishing test - attachment",
-            "timeout": 600,
+            "timeout": 700,
             "nightly": true,
             "integrations": [
                 "EWS Mail Sender",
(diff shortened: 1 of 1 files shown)


ronykoz:
Don't merge, the test is on skip, why?

Shellyber:
This two tests are not in skipped. And I need to merge a fix for playbook and I can't because this two fails on timeout

idovandijk:
Wait, @ShahafBenYakir already increased the timeout one time.
The reason it's failing on timeout is because it's stuck in a manual task. 
This PR should've fixed it https://github.com/demisto/content/pull/2907
We need to understand why the test is going in the "non-malicious" path again. The malicious URL "http://tpb.crushus.com/www.paperstreetcash.com" should be in the "Malicious" list of Palo Alto Minemeld

ronykoz:
@idovandijk please try to create a build which runs these tests and monitor it while he is running the test, try to figure out what's happening here

PR #3162, opened 2019-03-12
Fixing readfile

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16043

## Description
Fixing a bug where parsing unicode fails

### DIFF
--- a/Scripts/script-ReadFile.yml
+++ b/Scripts/script-ReadFile.yml
@@ -10,7 +10,14 @@ script: |-
       maxFileSize = 1024**2
   res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
   with open(res[0]['Contents']['path'], 'r') as f:
-      data = f.read(maxFileSize).decode('unicode_escape').encode('utf-8')
+      data = f.read(maxFileSize)
+      try:
+          data = data.decode('unicode_escape')
+      except UnicodeDecodeError:
+          pass  # Passing in a case we have \U or \N in the file
+
+      data = data.encode('utf-8')
+
   if data:
       message = 'Read %d bytes from file.' % (len(data))
       demisto.results({"Type" : entryTypes["note"],
@@ -39,4 +46,5 @@ args:
 outputs:
 - contextPath: FileData
   description: Data read from war room file
+releaseNotes: "Fixing unicode parsing error"
 scripttarget: 0
--- a/Scripts/script-ReadFile.yml
+++ b/Scripts/script-ReadFile.yml
@@ -11,12 +11,8 @@ script: |-
   res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
   with open(res[0]['Contents']['path'], 'r') as f:
       data = f.read(maxFileSize)
-      try:
-          data = data.decode('unicode_escape')
-      except UnicodeDecodeError:
-          pass  # Passing in a case we have \U or \N in the file
-
-      data = data.encode(demisto.args().get("encoding"))
+      data = data.decode(demisto.args().get("decode_type"))
+      data = data.encode('utf-8')
 
   if data:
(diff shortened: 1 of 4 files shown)


Itay4:
we should let the user decide the codec with an argument, which its default will be `utf-8'`
also need to add to the test special characters

yardensade:
Should create a test for it with utf-8 chars and non utf-8 chars.

Scripts/script-ReadFile.yml
```diff
@@ -10,7 +10,14 @@ script: |-
       maxFileSize = 1024**2
   res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
   with open(res[0]['Contents']['path'], 'r') as f:
-      data = f.read(maxFileSize).decode('unicode_escape').encode('utf-8')
+      data = f.read(maxFileSize)
+      try:
+          data = data.decode('unicode_escape')
```
glicht: I am not clear about why we are doing decode with 'unicode_escape'. Most files we will want to read will be encoded in utf-8 and not unicode_escape. 
```suggestion
          data = data.decode(demisto.args().get("encoding"), 'replace')
```

Scripts/script-ReadFile.yml
```diff
@@ -36,7 +43,11 @@ args:
   required: false
   default: false
   description: Maximal file size to load, in bytes. Default is 1MB.
+- name: encoding
+  description: The encoding type you want for your file
```
glicht:
```suggestion
  description: The encoding type of the file
```

Scripts/script-ReadFile.yml
```diff
@@ -10,7 +10,14 @@ script: |-
...
   res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
   with open(res[0]['Contents']['path'], 'r') as f:
-      data = f.read(maxFileSize).decode('unicode_escape').encode('utf-8')
+      data = f.read(maxFileSize)
+      try:
+          data = data.decode('unicode_escape')
+      except UnicodeDecodeError:
+          pass  # Passing in a case we have \U or \N in the file
+
+      data = data.encode(demisto.args().get("encoding"))
```
glicht: data returned to the context should be utf-8 encoded always
```suggestion
      data = data.encode('utf-8')
```

ronykoz:
@glicht is it good to go?

Scripts/script-ReadFile.yml
```diff
@@ -36,7 +39,11 @@ args:
   required: false
   default: false
   description: Maximal file size to load, in bytes. Default is 1MB.
+- name: decode_type
```
glicht: Let's change the name to: encoding
```suggestion
- name: encoding
```

Scripts/script-ReadFile.yml
```diff
@@ -36,7 +39,11 @@ args:
...
+  description: The decoding type of the file
```
glicht:
```suggestion
  description: The character encoding of the file
```

Scripts/script-ReadFile.yml
```diff
@@ -10,7 +10,10 @@ script: |-
       maxFileSize = 1024**2
   res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
   with open(res[0]['Contents']['path'], 'r') as f:
-      data = f.read(maxFileSize).decode('unicode_escape').encode('utf-8')
```
glicht: Use: `errors='replace'` for the decode function

ronykoz: What do you mean by `errors='replace'`?

glicht: `decode` method takes as second parameter an `errors` argument. It is best to use `replace` for the `errors` arg.

PR #3174, opened 2019-03-13
adding utf-8 encoding

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16015

## Description
Added encoding to utf-8

### DIFF
--- a/Scripts/script-ExtractIndicatorsFromWordFile.yml
+++ b/Scripts/script-ExtractIndicatorsFromWordFile.yml
@@ -115,7 +115,7 @@ script: |-
                    })
 
   # Returning all parsed data:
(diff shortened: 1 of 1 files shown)


idovandijk:
Nice. But don't the other demisto.results need it as well? Or will it work fine here?

>   demisto.results( {
      'Type': entryTypes['note'],
      'Contents': indicators_hr,
      'HumanReadable': indicators_hr
                   })

ronykoz:
@idovandijk it should work fine as extractIndicators returns a valid output, just not sure about the data i'm sending to it, @yaakovi what do you think(line 109)?

yaakovi:
think it will, but need to be tested

PR #3177, opened 2019-03-13
adding appropriate error msg for encrypted pdf file

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16026

## Description
Adding appropriate error msg for encrypted pdf file


Scripts/script-ReadPDFFile.yml
```diff
@@ -56,6 +57,9 @@ script: |-
           except PSEOF:
               mark_suspicious()
               sys.exit(0)
+          except PDFEncryptionError as err:
```
glicht: Let's handle this similar to how we handle PSEOF and not return an error. Human readable content should contain the error message as a regular entry.

ronykoz: Not sure if that's the right way to go here, when we encounter PSEOF we mark the file as suspicious and return an error entry, are you sure it's the right way to go here?

glicht:
@ronykoz Looks good. Can you just add a test with an encrypted file.

glicht:
@ronykoz All looks good. Build failure seems unrelated. Do you want me to force merge?

ronykoz:
@glicht yes please

PR #3179, opened 2019-03-13
Fix find similar incidents - ignore future incidents

erezh31:


erezh31:
fixes https://github.com/demisto/etc/issues/16109

meirwah:
@erezh31 can you fix test?
image

Scripts/script-FindSimilarIncidents.yml
```diff
@@ -142,7 +142,7 @@ script: |-
       condition_string = ' %s ' % applied_condition.lower()
       similar_keys_query = condition_string.join(map(lambda t: '%s:"%s"' % (t[0], escape_special_query_chars(t[1])), similar_incident_keys.items()))
       incident_time = parse_datetime(incident_time)
-      max_date = incident_time + timedelta(hours=hours_back)
+      max_date = incident_time
```
meirwah: why not make this an argument?

erezh31: I can, not sure it's correct logic - we want to prevent "loop" option they talked about it in the issue

meirwah: ok, then nevermind.. let's fix tests and merge

erezh31: will fix

PR #3181, opened 2019-03-14
reorder tests by mock usage

yaakovi:
## Status
Ready

## Related Issues
fixes: link to the issue

## Description
run mock tests first so the container will not contain side effect from the other tests.

   - No

## Must have
- [ ] Tests
- [X] Code Review


yaakovi:
@ronykoz when using the mocks there is no change to the environment variables.
when disabling proxy (mock-less tests), we delete the environment variables. this could cause a problem if the container is recycled for the next test.

that is why the mock tests should come first.

PR #3197, opened 2019-03-17
ASCII support in formattedArgs in commonServerPython

orenzohar:
## Status
Ready

## Description
better support for ASCII in commonServerPython `formattedArgs`


yaakovi:
@orenzohar please pull from master, this should resolve the problem in the tests

PR #3198, opened 2019-03-17
Test playbook retry and debugging for Whois

amshamah419:
## Status
Ready

## Description
Added a retry to the test playbook as well as added debugging to the whois request


yuvalbenshalom:
@amshamah419 it fixes on issue?

amshamah419:
Not yet, it needs to be excluded from tests, but I wanted to verify that it would pass testing.

PR #3206, opened 2019-03-18
Occurred report format fix

Bars92:
## Status
Ready

## Related Issues
fixes: demisto/etc#16096

## Description
change the report occurred time format in the decoder

   - Yes
       - Change occurred format for default reports.


ronykoz:
@Bars92 why are we making this change?
If thats a server change maybe we need to add a fromversion here?

Bars92:
@ronykoz no need. it will only change the "occurred" field format display for those reports that was initially wronged.  There was a client fix for this as well.

ronykoz:
Ok, then merge this issue

PR #3207, opened 2019-03-18
Added reputation command for email

hirshbergshachar:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16088

## Description
Added missing reputation command for email indicator

   - No

## Must have
- [ ] Code Review


idovandijk:
Shachar, not sure I understood correctly. Will this fix make sure that reputation is properly reflected? Even if there is no reputation command, shouldn't the general reputation reflect the most severe reputation given by source integrations?

hirshbergshachar:
@idovandijk this fix will ensure reputation is calculated for emails.
If we don't have reputation command or script we won't calculate an overall score for the indicator (based on external integrations overall score)

glicht:
Can we add a small test for this fix?

hirshbergshachar:
@glicht how would you suggest testing it from the content side? Running a playbook won't suffice as we need access to the indicator's final score rather than context DBot values

glicht:
You can use: `!getIndicator value="test@email.com"`

I think from a playbook it is probably sufficient to use the wrapper script: GetIndicatorDBotScore and check the returned score.

hirshbergshachar:
@glicht that won't check the original bug though since the indicator score is a different entity from DBot score. However, it's ok for just testing that it's a reputation command, I'll add it

glicht:
Added a test playbook. Build where test fails before merge with the fix: https://circleci.com/gh/demisto/content/20581

PR #3218, opened 2019-03-20
Adding util for triggering build with no backward compatibility check

ronykoz:
## Status
Ready

Adding util for triggering build with no backward compatibility check

### DIFF
--- a/Tests/scripts/validate_files.py
+++ b/Tests/scripts/validate_files.py
@@ -295,14 +295,16 @@ def main():
 
     parser = argparse.ArgumentParser(description='Utility CircleCI usage')
     parser.add_argument('-c', '--circle', type=str2bool, default=False, help='Is CircleCi or not')
+    parser.add_argument('-b', '--backwardComp', type=str2bool, default=True, help='To check backward compatibility.')
     options = parser.parse_args()
     is_circle = options.circle
+    is_backward_check = options.backwardComp
 
     logging.basicConfig(level=logging.CRITICAL)
 
     print_color("Starting validating files structure", LOG_COLORS.GREEN)
     files_validator = FilesValidator(is_circle)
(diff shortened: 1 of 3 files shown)


Utils/trigger_content_build_with_no_backward_check.sh
```diff
@@ -0,0 +1,20 @@
...
+
+_branch=$1
+_circle_token=$2
+
+trigger_build_url=https://circleci.com/api/v1/project/demisto/content/tree/${_branch}?circle-token=${_circle_token}
+
+post_data=$(cat <<EOF
+{
+  "build_parameters": {
+    "BACKWARD_COMPATIBILITY": "true"
```
yaakovi: shouldn't this be false?
we want to trigger the build without the BC checks

ronykoz:
@yuvalbenshalom @yaakovi review me

yaakovi:
on second look, i think the parameter should be passed to the actual validation of the BC and skip/ignore the result.
in the current way, this will ignore other problems that not related to BC (secrets, structure, image size, etc.)

Tests/scripts/validate_files.py
```diff
@@ -140,7 +140,7 @@ def validate_modified_files(self, modified_files):
                 old_file_path, file_path = file_path
 
             print("Validating {}".format(file_path))
-            structure_validator = StructureValidator(file_path, is_added_file=False,
+            structure_validator = StructureValidator(file_path, is_added_file=not(False and is_backward_check),
```
yaakovi:
```suggestion
            structure_validator = StructureValidator(file_path, is_added_file=not(False or is_backward_check),
```

yaakovi:
with BC check:
image


without BC check:
image

PR #3221, opened 2019-03-20
Add integration with Minerva Labs Anti-Evasion Platform

danikdanik:


Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
...cHFXov733bQIh/BC1GThGlGImpQHDhwAgNlwNTIDwaHs2iEryJhDEThZ7hJYcTkH6zq0+1qs1zWV2Cau1DJZaC3nUpTV2cPhHAN7+I39ts06rwDcSwtxF3XAatGnVgiWTCWkJdtHTa6gEUtGBGJpiOZanDRGe0VgjkmPBAIJJy+P9rz6bXmMLt0gM5JtsOqptLDz/nAdr+I39rZ7FM8A5gY7Cbd6HpNXK4i5GL5hy0UQdXTibjCAkjCTHHbEMVxT0ddR4hvTrorMZAc1/aUgEEcE3iFswS0vzC/+tt4tY5JnF4AWl4tJOAB2k1BqocDpNW+sbPwbgyNpM1B5YSRmRONGm1hI9ic+TmjJqPXxavny54bR5yHC+Mvric+rV4ZOEc1CJJc3YgKn5Nu2A93bgBitpGRpCMiEKTD6q++BHP8ArUb7Bu+YzwWRFL1sFXox2VpNzTE9pfzjV6NjoEJGItGSzy3CceCBYsHeTuaVGnuX/TcCAJVgKNUIO0A+trM9nGSC5KeqTGKFMwZ/GFTQ1D74E2tijmL8Dv2INdImXpHfhXUgAxcg8yCcylKgCrELLTAimoXlZMCCcBHDUdtpyl0P2/ApQrqSjwEJ2gwYm+G5EZOVS4yIlx6UBmyTAwdBrSB7jc4APIQc5h7TfdsQXrKfBwSCygYq2Cr6jeDkH1SDa02EkSsodnWjFfJUUEr0A2gGKcdEPvZKLQ7bcB5DqU1knbkkm5MRWFwZk5xFgOzRaVzMJsTrgfOZQxr5mm/gvu6g/VKCAFGY7zJNiu88AVhXRPhT5txUbBjcDE1Y0vU2DSddr5ansPmUswDEYu/zassiOgj9bMA6fe5fWgWY2J6yd9wdJbQIZhTEnJoTSXJjLJk/zJU3ZuMXkBHqeKGSWdQA3jovZSKgOUuaEUfzYvZ8imDVAITqbfeDIpvT9cL5YEYhIjxRbVs+CCPdKCDEh+uL7KwyMJewXzX8f9VgHeb6r0Aj6rLb0nwUAEFjayeG8gZahLVHAEA5wuQ3YbT102lwlOxIUiTGI6LTlH1FRRSsYFSS7nKgT1x0b0YfejTfJQBEkBEkeDhwjUQQo6Lc7U2VuaeDQ0nZcVyWKL2ebACA8fgNbLr5GKMEYGN50mZBMJgXlS0lW+OXL/iSpHWfAih1okntE84PqrLjXqVcHVBSh16NpAz7Yr8RRMXx/wDhP//Z
```
liorblob: The image has to be a PNG file and up to 3 MB

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+description: Minerva eliminates the endpoint security gap while empowering companies
+  to embrace technology fearlessly.
+detaileddescription: |-
+  Minerva Labs integration with Demisto.
+  The integration queries the Minerva server for new events, endpoint status, groups and able to create active response such as adding vaccinations and exclusions.
+configuration:
+- display: url
```
liorblob: Please give more descriptive display names, e.g. Server URL

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+  name: url
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: user
+  name: user
```
liorblob: Please use the `Authentication` parameter type for credentials and make it mandatory.

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+  type: 8
+  required: false
+- display: Incident type
+  name: incidentType
+  defaultvalue: ""
+  type: 13
+  required: false
+script:
+  script: |
+    # ''' IMPORTS '''
```
liorblob: No need for `#` before `'''`

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+    import json
+
+    import requests
+
+    # disable insecure warnings
+    requests.packages.urllib3.disable_warnings()
+
+    # ''' GLOBALS '''
+    USERNAME = demisto.params().get('user')
```
liorblob: After changing the parameter type to credentials, you would need to use `demisto.params()['credentials'].identifier/password`

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+
+
+    def logout():
+        session.post(BASE_URL + '/api/login/logout')
+
+
+    def fetch_incidents():
+        try:
+            r_events = session.post(BASE_URL + '/api/events',
+                                    json={"archive": False})
```
liorblob: Do we want the `archive` to be a configurable parameter?

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+            if r_events.status_code is not 200:
+                return_error('Error in API call. More information: {}, {}'.format(r_exclusion.status_code, r_exclusion.reason))
```
liorblob: `return_error` should not be used in `fetch_incidents` as it uses `demisto.results` which is not relevant in this case. In this scenario, you can use `raise`.

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+                                  'password': PASSWORD})
+
+
+    def logout():
+        session.post(BASE_URL + '/api/login/logout')
+
+
+    def fetch_incidents():
+        try:
+            r_events = session.post(BASE_URL + '/api/events',
```
liorblob: We don't want to filter by event time or something?

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+    def add_exclusion(exclusion_type, exclusion_data, exclusion_description, groups_id):
+        json_params = {'type': exclusion_type,
+                       'description': exclusion_description,
+                       'data': exclusion_data,
+                       'appliedGroupsIds': groups_id}
+        response = session.post(BASE_URL + '/api/exclusions',
+                                json=json_params)
+        if response.status_code is not 200:
+            return_error('Error while adding exclusion. More information: {}, {}'.format(response.status_code, response.reason))
+        demisto.results(response.json())
```
liorblob: The output needs to be an entry. See:
https://github.com/demisto/content/blob/master/docs/context_and_ouputs/README.MD
https://github.com/demisto/content/blob/master/docs/code_conventions/README.MD (`demisto.results()`)

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+    def add_vaccine(vaccine_name, vaccine_desc, monitor_only):
+        json_params = {'name': vaccine_name,
+                       'description': vaccine_desc,
+                       'isMonitorOnly': monitor_only,
+                       'type': 'mutex'}
+        response = session.post(BASE_URL + '/api/vaccination',
+                                json=json_params)
+        if response.status_code is not 200:
+            return_error('Error while adding a vaccine. More information: {}, {}'.format(response.status_code, response.reason))
+        demisto.results(response.json())
```
liorblob: Needs to be an entry, as described above.

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+    @logger
+    def search(search_url, search_param, search_condition, search_value):
+        json_params = {'filters': [{'param': search_param,
+                                    'condition': search_condition,
+                                    'value': search_value}]}
+        response = session.post(search_url,
+                                json=json_params)
+        if response.status_code is not 200:
+            return_error('Error while perfroming search for {}. More information: {}, {}'.format(search_url.rsplit('/')[1], response.status_code, response.reason))
+        demisto.results(response.json())
```
liorblob: Needs to be an entry, as described above.

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,293 @@
...
+        if demisto.command() == 'minerva-search-endpoint':
+            search_endpoint_command()
+    except Exception, e:
+        LOG('Cannot perform the command: {}. Error: {}'.format(demisto.command(), e.message))
+        LOG.print_log()
+        return_error(e.message)
+    finally:
+        logout()
+  type: python
+  commands:
```
liorblob: Please add outputs - https://github.com/demisto/content/blob/master/docs/context_and_ouputs/README.MD

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: Couldn't find relevant documentation / demo regarding it, but by some guess work it seems to work.
Hope it's ok now.

liorblob: @danikdanik you've raised a good point though, so we've created the following:
https://github.com/demisto/content/blob/master/docs/parameter_types/README.md

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: 👍

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: @liorblob, I'm not sure context and outputs are relevant for this kind of integration

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: There's no need, as the events being fetched and marked as _Archived_ in our Management Console (which is the way to say "an event was processed")

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: No, as this is how we want to fetch the events that were not yet processed

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,424 @@
...
+
+        results = {
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['markdown'],
+            'Contents': response.json(),
+            'HumanReadable': tableToMarkdown('Search results for \'{}\''.format(search_url.split('/')[-1]), response.json(), headerTransform=pascalToSpace, removeNull=True)
+        }
+
+        if response.json():
+            results['EntryContext'] = create_entry_context(search_url, response.json()[0])
```
liorblob: Why only the first element?

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,424 @@
...
+    def get_groups():
+        response = session.get(BASE_URL + '/api/groups', params={'_limit': '1000'})
+        if response.status_code is not 200:
+            return_error('Error while fetching groups. More information: {}, {}'.format(response.status_code, response.reason))
+
+        demisto.results({
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['markdown'],
+            'Contents': response.json(),
+            'HumanReadable': tableToMarkdown('Available groups', t=response.json(), headerTransform=pascalToSpace, removeNull=True)
```
liorblob: We need entry context if there are outputs

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
liorblob: Also need to add `Endpoint` or `Process` to the context object, e.g., `"Minerva.Process": create_entry_context`

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,424 @@
...
...QSkUQumPulX+vwccRADnQkBRWJKhwZIMDR6sqYOHYXB6dMT/3yiaTCZ0Wy0wTRHFfJkMNxaX4JsLFqX6JcwI7wQQAD5fWY2nz5+d8340RaE+LR0bcn3zG9Zm58Q1gmQj54yjGBobw31VNYFj3VYrbt+/Dy6vF+eMo9i97cq4P+/2ykqs8NtheTyeaUOY7G43nP5jIokE2Wr1tJo4uVAYMFOYEDcRTUMmEEImEEBM04H7qITCOVsCCdEhoCgs02Ri2RR3H4vLBbM/CakSiZAhFqf6UueEl9/2R+sbUKVS4Y2+XrRbzGAAiCkaGokE5UolKpQqLNVosFidkbQhOWzhL/7BSFcWXFqWvDdpZKY9QUXIW0vLUv3SCQlGJRJBlUBD3UTA22//9UXFuJ6FdV6pxOHx4LXeHqzKzEJN2iXrq8mjG28rK0v1ZRIISYO3AkiYzu/aWmFyOnHPlBGVNxSXQO9wQCUS4a6Kuc0UCAS+QDHMPKoanseYXC5ctuMt5Mvk2L1tO9kfIxDAUzMEwnS+e+oERp1OfHfxEiJ+BIIfIoDzgB3afvytpxu3l5VjS65vNCUDYJdOy2q3XgIh0RAB5DnnTUY8dOQQKlUq/GDpssDxXzRdwGc/3o9vnjie6kskEFIGEUAe02W14q6PPoSIovDyug1QTir52e+fUJbP4jYlAiHRkCwwTzlnNOL2/fsw7vXg1Q2bUKEMHpT0cJ3PP/CR+oZUXyqBkDJIFpiHvNrdhW+cPI4MsQQvr9uAhWp1qi+JQGAlJAJMMV6GQY/NhhKFImbboB6bFd89dRK7dFqszc7Bc6vXIFtK+lwJhJkgESAL8DIMjhoMGHDYsTUvP+J2onaLGc+3teIv3V2QCgT4zqIl+GxFJUixC4EwO0QAWYTD48E/+3rRY7OiLi0dyzSZKAnhSGNxudBkNuHg0CB2avtxcmQE6SIR7q2uwb3VtUifIqA2txuf/HAvHqypI+2BBMIkyBKYRcgEAtxRVg4GwJHhITzX1oKjhmEMj48jQySG3eOGxeXG8LjP6itXKsNl2dn4Um09rsgvmNEsVT/mwNDYOMqUylS/RAKBVZAIkAOYXC40m4zQOhxgGAb5MjlKFApWO+0SCFyACCDP+PLRw1igVuP+6tpUXwqBwHqIABIIhHkL6QQhEAjzlv8P271NO5Q5NtcAAAAASUVORK5CYII=
```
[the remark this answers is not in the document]
danikdanik: 👍

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: I treated the first element only, as the output retrieved from our Management Console server appears to be a list (with a single dictionary in it)

danikdanik:
Hi @liorblob, can you please take another look here? Thanks.

liorblob:
@danikdanik Still some confusion around the context - We need to return an object under the tree of `Minerva.Process` (or endpoint/group) so what we do is create an object that has, for example:
```
context = {
    'IP': ip,
    'OS': os...
}
```
and when we assign the entry context we do the following:
```
entry['EntryContext'] = 'Minerva.Endpoint(val.IP === obj.IP)': context
```
The `(val.IP === obj.IP)` code is for overridng duplicates, but it's not a must currently.

liorblob:
@danikdanik looks good! when can we try testing this?

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,721 @@
...
+  name: url
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: Minerva Management Console user and password
+  name: credentials
+  defaultvalue: ""
+  type: 9
+  required: true
+- display: Checking this indicates that the Minerva Management Console uses insecure
```
liorblob: The convention for the display is `Trust any certificate (not secure)`

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,721 @@
...
+                context = {
+                    'Id': data['id'],
+                    'Type': data['type'],
+                    'Exclusion data': data['data'],
+                    'Description': data['description'],
+                    'Last modified by': data['lastModifiedBy'],
+                    'Last modified on': data['lastModifiedOn'],
+                    'Applied groups': data['appliedGroupsIds']
+                }
+                ec['Minerva.Exclusion(val.Id === obj.Id)'] = context
```
liorblob: Do we need the val.Id == obj.Id?

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
```diff
@@ -0,0 +1,721 @@
...
+            return_error('Error while fetching {}. More information: {}, {}'.format(url_text, response.status_code, response.reason))
+
+        results = {
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['markdown'],
+            'Contents': response.json(),
+            'HumanReadable': tableToMarkdown(pascalToSpace(url_text), response.json(), headerTransform=pascalToSpace, removeNull=True)
+        }
+
+        if response.json():
```
liorblob: I think that should be checked before `results` is created, and then return the results accordingly

Integrations/integration-MinervaLabsAntiEvasionPlatform.yml
danikdanik: Sure, as for the case the data retrieved is not a list, we'd like to remove duplicates (as you wrote me in the email)

liorblob: That applies also to lists, it should be in all context entries. Does it not work when you add it to the other entries?

danikdanik: my bad, omitted the context ids. pushed the fix

PR #3223, opened 2019-03-20
Fixed Namings in Duo Admin API

Winitzky:
## Status
Ready

## Description
After review from Alex, changed argument name

### DIFF
--- a/Integrations/integration-DuoAdminApi.yml
+++ b/Integrations/integration-DuoAdminApi.yml
@@ -70,6 +70,8 @@ script:
 
     OPTIONS_TO_TIME = {
         '10_seconds_ago' : datetime.datetime.now() - datetime.timedelta(seconds = 10),
+        #left here for backwards compatability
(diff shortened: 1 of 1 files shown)


Integrations/integration-DuoAdminApi.yml
```diff
@@ -70,7 +70,7 @@ script:
 
     OPTIONS_TO_TIME = {
         '10_seconds_ago' : datetime.datetime.now() - datetime.timedelta(seconds = 10),
-        '1_minutes_ago' : datetime.datetime.now() - datetime.timedelta(minutes = 1),
```
glicht: Since we've already released leave this option too for backwards compatibility. So if there is a playbook using this already we will still parse this value fine. in the predefined it is ok to remove as you did below.

Winitzky: Done

glicht:
Looks good. Merge when you are ready.

PR #3225, opened 2019-03-21
push incidents in chunks (ArcSight v2)

orenzohar:
## Description
- limits the number of incidents that are fetched each time. (added parameter).
- better documentation for `as-update-case` permissions.

## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16225#issuecomment-475135377
fixes: https://github.com/demisto/etc/issues/16233#event-2219110296

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/ArcSightESMv2/ArcSightESMv2.py
+++ b/Integrations/ArcSightESMv2/ArcSightESMv2.py
@@ -579,8 +579,9 @@ def update_case(case_id, stage, severity):
 
     if not res.ok:
         demisto.debug(res.text)
-        return_error('Failed to get security update case {}\nFull URL: {}\nStatus Code: {}\nResponse Body: {}'.format(
-            case_id, BASE_URL + query_path, res.status_code, res.text))
+        return_error('Failed to get security update case {}. \nPlease make sure user have edit permissions,'
+                     ' or case is unlocked. \nStatus Code: {}\nResponse Body: {}'.format(case_id, res.status_code,
+                                                                                         res.text))
 
     res_json = res.json()
     if 'cas.updateResponse' in res_json and 'cas.return' in res_json.get('cas.updateResponse'):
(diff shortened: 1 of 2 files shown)


Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -345,6 +346,9 @@ def fetch():
     already_fetched = last_run.get('already_fetched', [])
 
     fields, query_results = get_query_viewer_results(events_query_viewer_id or cases_query_viewer_id)
+    # sort query_results by creation time
+    query_results.sort(key=lambda k: int(k.get('Start Time') or k.get('Create Time')))
```
omercnet: this could cause unexpected behavior
they should get ingested in the same order they appear in ArcSight

orenzohar: @omercnet @anara123 the order is not always ascending (the opposite actually, most the times). since we now fetching in batches, we might lose the oldest cases/events. e.g., a query with for all the cases in the last 2 hours (and let's say there are 200)- if we'll use ArcSight order we will fetch the most recent first, wait for 15 minutes to the next fetch (default) and then get the next batch. this will take us at total an hour. In the meantime, the query viewer could update and discard the cases/events from 3 hours ago.

omercnet: not sure what you refer to but we wait 1 minute between fetches, not 15

so maybe consider reversing the order, not sorting?

orenzohar: to add to what I said, the data from ArcSight is not always ordered, and even if it is, not always in ascending order.

omercnet: so let's please keep the same (dis)order so it's ingested in the same order as in the UI

PR #3226, opened 2019-03-21
Common server python 3 fixes

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16224

## Description
Added python 3 support for compareDates and date_to_timestamp.

   - No

## Must have
- [X] Tests - no tests for python 3, but there are tests for python 2
- [ ] Code Review


Scripts/script-CommonServerPython.yml
```diff
@@ -182,13 +182,13 @@ script: |-
       first = None
       if date1 == 'now':
           first = datetime.now(tzlocal())
-      elif isinstance(date1, basestring):
+      elif is_str_or_unicode(date1):
```
yaakovi: why not just using:
`isinstance(s, STRING_TYPES)`

DeanArbel: Because I didn't know of it. Awesome feedback, will switch implementation

Scripts/script-CommonServerPython.yml
```diff
@@ -14,6 +14,7 @@ script: |-
   import sys
   import os
   import re
+  import six
```
yaakovi: remove the import if it's not being used

yaakovi:
`compareDates` is missing a test, please add it as well

PR #3227, opened 2019-03-21
Zscaler - fixed short interval bug

guykeller:
## Status

## Related Issues
fixes: https://github.com/demisto/etc/issues/15722

## Description
Fixed bug in Zscaler which caused an exception when sending several requests in a short interval.
Now the requests are retried several times before any exception is thrown.

   - No

## Must have
- [x] Tests
- [x] Code Review


Itay4:
why did you remove the proxy ?

guykeller:
I have no idea how it happened, not intentional but not sure what caused it. I'll re-add it

Integrations/integration-Zscaler.yml
```diff
@@ -84,7 +86,12 @@ script:
                 headers=headers
             )
             if res.status_code not in (200, 204):
-                raise Exception('Your request failed with the following error: ' + ERROR_CODES_DICT[res.status_code])
+                if res.status_code == EXCEEDED_RATE_LIMIT_STATUS_CODE and num_of_seconds_to_wait <= MAX_SECONDS_TO_WAIT:
+                    time.sleep(num_of_seconds_to_wait)
+                    #time.sleep(num_of_seconds_to_wait)
```
ronykoz: remove this line

ronykoz:
@guykeller On second thought i just came across this https://help.zscaler.com/zia/api-rate-limit-summary
So we might be able to want to have this functionality as optional so it will not finish the hourly quota for the user, @mayagoldb do you think the numbers there are reasonable to achieve?

mayagoldb:
@ronykoz Well actually I don't think it should be optional because we are thinking about a case in which the customer has 10 incidents at the same time (and then 10 playbooks run at the same time), and the commands must run. But, we can base our number of retries on their quota limitation and document it under 'known limitations'.

ronykoz:
OK sounds good to me, @guykeller please update the docs as well

PR #3237, opened 2019-03-21
File Enrichment - Generic v2 + Test Playbook 

idovandijk:
## Status
Ready

## Related Issues

## Description
Made a v2 playbook of `File Enrichment - Generic`, named `File Enrichment - Generic v2`. The playbook includes (but is not limited to) the following changes:
- **Uniq** transformer applied in multiple locations to prevent enrichment duplication
- Auto-extract set to correct types in playbook tasks, to improve performance and indicator extraction, enrichment and reputation
- Configured file enrichment test to run for file enrichment playbook
- Improved file enrichment test to use filters instead of deprecated `VerifyContext` script
- Improved file enrichment test to test a SHA256 that is present in Cylance Protect, thus making the test more stable because it doesn't rely on VirusTotal enrichment exclusively
- Removed FileReputation subplaybook which uses deprecated scripts, preventing excessive enrichment (letting auto-extract work instead)
- Removed Palo Alto Networks AppFramework from the playbook
- The playbook will not provide reputation for indicators

## Screenshots
image

4.1.0

   - No. A v2 playbook is being released and it should not affect the previous version of the playbook.

## Must have
- [x] Tests
- [ ] Code Review


idovandijk:
Blocked by this issue: https://github.com/demisto/etc/issues/16336

PR #3240, opened 2019-03-21
Minemeld - Added validation of deleting indictors from miners of type localDB

bakatzir:
…emove unnecessary line in params.

## Status
Ready

## Description:
we checked that all indicators are on the miner only for non localdb type miners.
added a check for this type as well.
please note that removing an indicator from localdb vs other types is different.

## Related Issues
fixes: https://github.com/demisto/etc/issues/16141

## Screenshots
in issue

   - No

## Must have
- [x] Tests
- [x] Documentation - not changed
- [x] Code Review


bakatzir:
@yuvalbenshalom it's error handling, so no test.
@yardensade  please see last commit for lowercasing outputs.

PR #3242, opened 2019-03-21
Fixed a bug where the playbook would confuse MSG / EML files with doc/x.

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16137

## Description
- Fixed a bug where the playbook would confuse MSG / EML files with doc/x.
- Added downloading an EML file to the test. If the EML file is parsed then the test will fail. If the EML file was parsed and did not fail the test, then we also look for a URL from the file, and if we find it - we will intentionally fail the test.

4.1.0 and up.

   - No

## Must have
- [x] Tests
- [ ] Code Review


glicht:
@idovandijk  build is failing

idovandijk:
> @idovandijk build is failing

OK, Re-ran it to see why it failed. Might be unrelated to the fix.

idovandijk:
Can't access the server. It dies instantly. What's going on?

idovandijk:
Contacted @yaakovi to take a look at why the instance keeps dying so I can see why the test failed.

idovandijk:
Was blocked by this: https://github.com/demisto/etc/issues/16336.
Found the problem and resolved here https://github.com/demisto/content/pull/3292.
Now getting this error but should be resolved within ~30min (devops)
image

Should then pass, at least on latest GA.

glicht:
@idovandijk can you add to the test playbook (or a new test playbook) a verification that the bug is fixed. Meaning add an msg/eml file and make sure it is not being extracted as a word doc.

idovandijk:
> @idovandijk can you add to the test playbook (or a new test playbook) a verification that the bug is fixed. Meaning add an msg/eml file and make sure it is not being extracted as a word doc.

Sure

idovandijk:
Green build :)

PR #3251, opened 2019-03-24
fix Extract Indicators From File - Generic playbook when the pdf file is invalid

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16208

## Description
Extract Indicators From File - Generic breaks when a malformed image file with a pdf extension is passed to it.

## Related PRs
https://github.com/demisto/content/pull/3239


yaakovi:
@adi88d test is failing

JonathanMeler:
@yaakovi it's hard to follow a YMAL for changes.
Please let me know what exactly you did to fix the issue

yaakovi:
only the condition task that determine if a file is a pdf or not:
File Info and File.Type should **contain** pdf (used to be equal "application/pdf")
there used to be an "or" between the conditions and not "and"

JonathanMeler:
Sounds good

PR #3253, opened 2019-03-24
ServiceNow - secure fetch incidents

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/16264

Backward compatibility break:
Removing `template` argument from the `create-record` command, as it has no effect and not really working.


anara123:
@liorblob you break backward compatibility - add description what you break and why

liorblob:
@anara123 updated

PR #3255, opened 2019-03-24
urlscan.io - add redirected urls data and threshold parameter

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16241
fixes: https://github.com/demisto/etc/issues/16286 FYI @hirshbergshachar 

## Description
add redirected urls data (to !url, !urlscan-submit) and threshold parameter

## Screenshots
in issue

   - No

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/16282
- [x] Code Review

### DIFF
--- a/Integrations/integration-UrlScan.yml
+++ b/Integrations/integration-UrlScan.yml
@@ -47,7 +47,7 @@ script:
     '''GLOBAL VARS'''
     BASE_URL = 'https://urlscan.io/api/v1/'
     APIKEY = demisto.params().get('apikey')
-    THRESHOLD = int(demisto.params().get('url_threshold', None)) or 1;
+    THRESHOLD = int(demisto.params().get('url_threshold', '1'))
     INSECURE = demisto.params().get('insecure')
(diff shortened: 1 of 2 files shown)


Integrations/integration-UrlScan.yml
```diff
@@ -38,10 +44,10 @@ script:
     requests.packages.urllib3.disable_warnings()
 
 
-
     '''GLOBAL VARS'''
     BASE_URL = 'https://urlscan.io/api/v1/'
     APIKEY = demisto.params().get('apikey')
+    THRESHOLD = int(demisto.params().get('url_threshold', None)) or 1;
```
yuvalbenshalom: why not `int(demisto.params().get('url_threshold', '1'))`?
this way `0` is not  a possible threshold because you will get `1`

bakatzir:
@yuvalbenshalom 
you're right.
There are no context checks in the test at all, i'll pick a few and add them

PR #3262, opened 2019-03-25
Refactor proxy logic in CommonServerPython

BenJoParadise:
## Status
Ready

## Description
Modified proxy logic functions in CommonServerPython:
enable_proxy is not needed (the logic is done in python loop)
disable_proxy changed to handle_proxy:
- checkbox logic implemented
- customizable proxy param name and checkbox default value
- returns proxies dictionary for `requests` methods "proxies" parameter.

Should check option for bypass mechanism that always disables proxy.

## Screenshots

## Related PRs

branch | PR
------ | ------

x.x.x

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes


Scripts/script-CommonServerPython.yml
```diff
@@ -28,31 +28,31 @@ script: |-
...
   dbotscores = {'Critical': 4, 'High': 3, 'Medium': 2,'Low': 1, 'Unknown': 0, 'Informational': 0.5}
 
-  SAVE_PROXIES = {}
-
-  def disable_proxy():
+  def handle_proxy(proxy_param_name='proxy', checkbox_default_value=False):
       """
-          Disable routing traffic through the system proxy.
+          Handle logic for routing traffic through the system proxy.
           Should usually be called at the beginning of the integration, depeding on proxy checkbox state.
```
yaakovi:
```suggestion
           Should usually be called at the beginning of the integration, depending on proxy checkbox state.
```

Scripts/script-CommonServerPython.yml
```diff
@@ -28,31 +28,31 @@ script: |-
...
+      proxies = {}
+      if demisto.params().get(proxy_param_name, checkbox_default_value):
+          proxies = {
+              'http': os.environ.get('HTTP_PROXY') or os.environ.get('http_proxy', ''),
+              'https': os.environ.get('HTTPS_PROXY') or os.environ.get('https_proxy', '')
+          }
+      else:
+          for k in ('HTTP_PROXY', 'HTTPS_PROXY', 'http_proxy', 'https_proxy'):
+              if k in os.environ:
+                  del os.environ[k]
```
yaakovi: when will this be restored?

Integrations/integration-Centreon.yml
```diff
@@ -36,8 +36,7 @@ script:
     # disable insecure warnings
     requests.packages.urllib3.disable_warnings()
 
-    if not demisto.params().get('proxy', False):
-        disable_proxy()
+    handle_proxy()
```
yaakovi: shouldn't we move this to the `EXECUTION CODE` section? 
i know until now we put it at the top, but is that really the best place?

also, this needs to be updated in the plugin template (after merging to master)

Scripts/script-CommonServerPython.yml
glicht: @yaakovi the restore is done in the python loop. No need to perform this logic in common server.

Integrations/integration-Centreon.yml
BenJoParadise: Theoretically yes.

However, if we do that, any request that will be sent before the `EXECUTION CODE` section will try to pass through proxy. I know it shouldn't happen, but i've already seen integrations with active code before that block.

BenJoParadise: In the case of this integration, yes, i'll do it now.

PR #3264, opened 2019-03-25
improved package_creator.py detailed description parsing

orenzohar:
## Status
Ready

## Description
package_creator.py wraps `detaileddescription` in quotation marks, so any other quotation marks in the description will break the yml.
fix now only add `|` at the beginning, and shift everything to the right (for multilined `detaileddescription`)

## Before
image

## After
image


yardensade:
Can we delete this branch?

PR #3214, opened 2019-03-19
Active Directory Query v2 - Fixed error when specifying a port

vector-sec:
## Status
Ready

## Related Issues
No open issues

## Description
Because the port is given to the ldap3 library as a string, a "port must be an integer" error is thrown when you specify a port in the instance configuration. Casting this parameter to an integer resolves this issue.

## Screenshots
Paste here any images that will help the reviewer

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------

## Required version of Demisto
x.x.x

## Does it break backward compatibility?
   - No

## Must have
- [ ] Tests
- [ ] Documentation (with link to it)
- [ ] Code Review

## Dependencies
Mention the dependencies of the entity you changed as given from the precommit hooks in checkboxes, and tick after tested them.
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes
Describe additional changes done, for example adding a function to common server.

PR #3265, opened 2019-03-25
Fixed error when specifying a port (#3214)

noaco:


glicht:
fixes: https://github.com/demisto/etc/issues/16283

noaco:
> What will happen if you dont provide port, then PORT will be None. `initialize_server` is ok with that?

yes

PR #3271, opened 2019-03-25
Cisco AMP - configuration test fix

noaco:
## Status
In Hold - missing test, missing credentials

## Related Issues
fixes: https://github.com/demisto/etc/issues/13767


Integrations/integration-AMP.yml
```diff
@@ -174,30 +174,32 @@ script:
         return undefined;
     }
 
+    function executeCommand(command, cmdArgs) {
```
anara123: executeCommand is saved name in demisto. We have builtin command called executeCommand.
so change this function name to something else

Integrations/integration-AMP.yml
```diff
@@ -174,30 +174,32 @@ script:
...
+            body
+        );
+        return res;
+    }
+
     switch (command) {
         case 'test-module':
-            command = 'amp_get_version';
+            executeCommand('amp_get_version', args);
+            // successful reponse
```
anara123: reponse => response

PR #3275, opened 2019-03-25
Unit tests

glicht:
## Status
Ready

## Description
Support for running unit tests and linters that need full env such as pylint. Will run the unit tests and lint in the appropriate docker. 

Currently build doesn't fail when lint or unit testing fail.   

To test locally run:
```
./Tests/scripts/pkg_dev_tasks_in_docker.py -d Integrations/Active_Directory_Query 
```

## Additional changes
Lint fixes to ParseEmailFiles


yaakovi:
should the example take long?
it's running for a few minutes now
```
shais-MacBook-Pro:content shaiyaakovi$ python ./Tests/scripts/pkg_dev_tasks_in_docker.py -d Integrations/Active_Directory_Query -v
Using yaml file: /Users/shaiyaakovi/dev/demisto/content/Integrations/Active_Directory_Query/Active_Directory_Query.yml
Using docker image: demisto/ldap:1.0.0.75


```

glicht:
Shouldn't take long. You can run with `-v` to get more output. In the build it takes under a minute for a single integration.

Tests/scripts/pkg_dev_container_setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+# Setup a container for dev testing. 
+#
+# In alpine will install necessary python dev tools and then use pip to install
+
+cp /dev/stdin /pkg-dev-test-requirements.txt
```
ronykoz: I think we should add a comment that states from where are we getting the stdin, would help future maintenance of this code

Tests/scripts/pkg_dev_tasks_in_docker.py
```diff
@@ -0,0 +1,172 @@
...
+    subprocess.check_call(['python2', CONTENT_DIR + '/package_extractor.py', '-i',
+                           'Scripts/script-CommonServerPython.yml', '-o',
+                           project_dir + '/CommonServerPython.py'], cwd=CONTENT_DIR)
+    run_params = ['docker', 'create', '-v', workdir, '-w', workdir,
+                  '-e', 'PYLINT_FILES={}'.format(pylint_files)]
+    if no_test:
+        run_params.extend(['-e', 'PYTEST_SKIP=1'])
+    if no_lint:
+        run_params.extend(['-e', 'PYLINT_SKIP=1'])
+    run_params.extend([docker_image, 'sh', './{}'.format(RUN_SH_FILE_NAME)])
```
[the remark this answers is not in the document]
glicht: It needs to be at the end

Tests/scripts/pkg_dev_container_setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+pip install -r /pkg-dev-test-requirements.txt
+
+# for alpine we install dev dependencies to support dependencies that need native code (seen with typed_ast)
+if [ -f "/etc/alpine-release" ]; then
+    apk --update add --no-cache --virtual .build-dependencies python-dev build-base wget
+fi
+
+pip install -r /pkg-dev-test-requirements.txt
```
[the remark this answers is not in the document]
glicht: No need twice. First one is not necessary. Good catch

glicht:
@ronykoz removed the output of the docker create. So it is a clearer output.

Tests/scripts/pkg_dev_container_setup.sh
glicht: added in new commit

Tests/scripts/pkg_dev_tasks_in_docker.py
```diff
@@ -0,0 +1,176 @@
...
+import hashlib
+import sys
+import shutil
+
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+CONTENT_DIR = os.path.abspath(SCRIPT_DIR + '/../..')
+sys.path.append(CONTENT_DIR)
+from package_creator import get_code_file  # noqa: E402
+
+DEF_DOCKER = 'demisto/python:1.3-alpine'
```
Itay4: shouldn't this be 1.4 ?

glicht: No. Server is still using 1.3 as default

glicht:
i think all comments have been addressed. Let me know if you see any additional issues.

PR #3278, opened 2019-03-26
Fix tanium test error

shaniacht1:
## Status
ready

## Related Issues
fixes https://github.com/demisto/etc/issues/16202

## Description
Added log messages to all commands accessing Pytan
Removed "get all packages" test from the integration test. 

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Tanium.yml
+++ b/Integrations/integration-Tanium.yml
@@ -28,6 +28,11 @@ configuration:
   defaultvalue: "false"
   type: 8
   required: false
+- display: Test Question
+  name: test_question
+  defaultvalue: ""
+  type: 0
+  required: false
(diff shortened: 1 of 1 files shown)


Itay4:
Can we add to the test something instead, which does not require non-admin permissions?

shaniacht1:
@Itay4 let me know what you think (I made it user configurable so we don't run into other unexpected permission issues

Itay4:
@shaniacht1 dont really think we need a param just for the test, can't we just use the default one you added there?

Itay4:
@shaniacht1 what do you say?

shaniacht1:
@Itay4 done

PR #3282, opened 2019-03-26
URL Enrichment - Generic v2 + Test Playbook

idovandijk:
## Status
Ready

## Related Issues

## Description
- Reworked the test playbook completely. No more deprecated scripts, and now the test thoroughly tests the new `URL Enrichment - Generic` playbook. The old test was incomplete, has been in our content, but hasn't been used anywhere until now
- Completely removed the GetReputation flag from the playbook inputs and from the utilization of that flag throughout the playbook, in accordance with the decision to remove reputation from the enrichment playbook
- Changed playbook to meet playbook conventions (wording of section headers, descriptions of tasks and more)
- Added `Uniq` transformer to any task that deals with indicators, to eliminate duplication of indicators
- Removed `URLReputation` script from the playbook. Enrichment playbooks will not provide reputation anymore
- Multiple consistency fixes: Url, url, URL, URLs
- Reorganized and simplified overall visualization of playbook
- Fixed VirusTotal Private API task to not be bound to specific instance. Also fixed names and descriptions.
- Changed old DT selectors to new friendly selectors (selecting URL properly with transformers instead of using `${URL}` as input for tasks)
- Set "Autoextract" correctly for all playbook tasks + test tasks. New playbooks should not use `Use System Default`, but rather `None`, `Out of Band` or `Inline`

## Screenshots
New playbook:
image

Test passing successfully on 4.5:
image

4.1.0

   - Yes - some naming changes + the removal of reputation from the playbook.

## Must have
- [x] Tests
- [ ] Code Review


idovandijk:
Blocked by this issue: https://github.com/demisto/etc/issues/16336

PR #3283, opened 2019-03-26
Add String to Boolean function in CommonServerPython

amshamah419:
## Status
Ready

## Description
Converts argument string of `'True'`/`'true'` and `'False'`/`'false'` to their respective boolean values.

Example: (`"True"` becomes `True`, `"False"` becomes `False`)

   - No

## Must have
- [ ] Tests
- [ ] Code Review


amshamah419:
Use bool(strtobool()) Instead

PR #3286, opened 2019-03-26
Disk size with partition

Spivaka:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/11714

## Description
A new widget to support any partition in the disk and not just root

## Related PRs

branch | PR
------ | ------
server | https://github.com/demisto/server/pull/12509

will start to take place from version 5.0 and above

   - No


Itay4:
I don’t think toversion and fromversion can be the same

Spivaka:
they are not... for the old widget the from is 3.5.1, and to is 5.0
for the new one its just from version 5.0
or I am missing something in your comment

Itay4:
It happened to us in the past with playbooks that the new one was not updated in upgrade to new server version, because the toversion of the old one was the new server version

Get Outlook for iOS<https://aka.ms/o0ukef>

Spivaka:
ok, fixed

PR #3288, opened 2019-03-26
Export in stix format

idobloch:
## Status
In Hold(wait for the server CircleCI to pass and PR to be approved)
https://github.com/demisto/server/pull/12454

## Related Issues
fixes: https://github.com/demisto/etc/issues/5793

## Description
Create automation which will be used when exporting indicators in STIX format 

## Related PRs
https://github.com/demisto/server/pull/12454
https://github.com/demisto/web-client/pull/4734

branch | PR
------ | ------

5.0.0

   - No

## Must have
- [X] Tests
- [X] Documentation (with link to it) - https://github.com/demisto/etc/issues/16347
- [ ] Code Review


Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+
+  from stix2 import Indicator, Bundle
+
+  indicators = []
+
+  userArgs = demisto.args().get('indicators','Unknown')
+  if isinstance(userArgs, dict) == True:
+      all_args = json.loads(json.dumps(userArgs))
+  else:
+      all_args = json.loads(demisto.args().get('indicators','Unknown'))
```
anara123: json.loads will raise an Exception if string passed to it. So either replace 'Unknown' with dict or change the whole error logic

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+  ''' IMPORTS '''
+  import json
+
+  from stix2 import Indicator, Bundle
+
+  indicators = []
+
+  userArgs = demisto.args().get('indicators','Unknown')
+  if isinstance(userArgs, dict) == True:
+      all_args = json.loads(json.dumps(userArgs))
```
anara123: why you are doing dumps and then loads?

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+      elif demisto_indicator_type.lower() == "File SHA256".lower():
+          stix_type_and_value = "[file:hashes.sha256 = '" + value + "']"
+      elif demisto_indicator_type.lower() == "File SHA-256".lower():
+          stix_type_and_value = "[file:hashes.sha256 = '" + value + "']"
+      elif demisto_indicator_type.lower() == "IP".lower():
+          stix_type_and_value = "[ipv4-addr:value = '" + value + "']"
+      elif demisto_indicator_type.lower() == "URL".lower():
+          stix_type_and_value = "[url:value = '" + value + "']"
+      else:
+          stix_type_and_value = "[{}:value = '{}']".format(demisto_indicator_type.lower(),value)
```
anara123: is there predefined key for account? foo@demisto.com?

anara123: or domain?

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+  indicators = []
+
+  userArgs = demisto.args().get('indicators','Unknown')
+  if isinstance(userArgs, dict) == True:
+      all_args = json.loads(json.dumps(userArgs))
+  else:
+      all_args = json.loads(demisto.args().get('indicators','Unknown'))
+
+  counter = 0
+  for indicator_fields in all_args:
```
anara123: if all_args is list then why you check that it is dict in line 14?

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+      demisto_indicator_type = all_args.get(str(counter)).get('indicator_type','Unknown')
```
anara123: if you want to get an index in list then do all_args[counter] instead of all_args.get(str(counter))

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+                       'Contents': str(bundle),
+                       'EntryContext': context,
+                       'ContentsFormat': formats['json']})
+  else:
+
+      context = {
+          'StixExportedIndicators(val.pattern && val.pattern == obj.pattern)': json.loads(str(indicators[0]))
+      }
+
+      demisto.results({'Type': entryTypes['note'],
```
anara123: use return_outputs instead demisto.results

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+                            score=demisto_score,
+                            allow_custom=True)
+
+      indicators.append(indicator)
+  if counter > 1:
+      bundle = Bundle(indicators)
+      context = {
+          'StixExportedIndicators(val.pattern && val.pattern == obj.pattern)': json.loads(str(bundle))
+      }
+      demisto.results({'Type': entryTypes['note'],
```
anara123: use return_outputs instead of demisto.results

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+  description: Indicator creation date
+  type: string
+- contextPath: StixExportedIndicators.firstSeen
+  description: Indicator first appearance
+  type: string
+- contextPath: StixExportedIndicators.source
+  description: Source system for this indicator
+  type: string
+- contextPath: StixExportedIndicators.type
+  description: Indicator type
```
anara123: describe which possible type options exists

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+enabled: true
+args:
+- name: indicators
+  description: json object of all indicators  fields index mapped to indicators fields
+outputs:
+- contextPath: StixExportedIndicators.created
+  description: Indicator creation date
+  type: string
+- contextPath: StixExportedIndicators.firstSeen
+  description: Indicator first appearance
```
anara123: add the firstSeen date format, same for created date format

Scripts/script-StixCreator.yml
```diff
@@ -0,0 +1,118 @@
...
+  description: Source system for this indicator
+  type: string
+- contextPath: StixExportedIndicators.type
+  description: Indicator type
+  type: string
+- contextPath: StixExportedIndicators.pattern
+  description: STIX pattern
+  type: string
+- contextPath: StixExportedIndicators.score
+  description: STIX impact score
```
anara123: describe score options, is it number of enum ?

Scripts/script-StixCreator.yml
idobloch: yes, this specific email already been defined in the white list before I worked on it

Scripts/script-StixCreator.yml
idobloch: I'm running this script from go and it passes a dictionary, when I'm using json.loads only it fails it has to convert the go dictionary to string first and only then it works.

Scripts/script-StixCreator.yml
idobloch: because currently there are 2 ways to use this script one is from the server (which send golang dictionary) and one is from the user interface (! - the bang), so when the indicators are being sent from the user it doesn't get as dict, but as string

Scripts/script-StixCreator.yml
idobloch: fixed it (added description), it is string which decided that can be in our case High, Medium or empty

Scripts/script-StixCreator.yml
idobloch: fixed

Scripts/script-StixCreator.yml
idobloch: actually I'm not sure what are the possible types are, it is the Source field from Indicators page, the only type that i saw was DBot

Scripts/script-StixCreator.yml
idobloch: fixed

Scripts/script-StixCreator.yml
idobloch: fixed

Scripts/script-StixCreator.yml
idobloch: fixed

Scripts/script-StixCreator.yml
idobloch: not sure exactly what you meant here I would like to talk with you about it tomorrow

idobloch:
I've seen there is a new branch of 19.4.0, should it merge to master or to 19.4.0, if we want it to be delivered with Demisto 5.0.0 (I think)

Scripts/script-StixCreator.yml
anara123: I would imagine that indicator types are url, IPv4, domain, email, etc

anara123:
If we want to deliver it with 5.0.0 then it doesn't matter. You should add fromversion: 5.0.0 to the script

Scripts/script-StixCreator.yml
idobloch: Actually no, because STIX itself have types and we use only the indicator. 
but in the pattern field there are the indicator types, so i added the types as you said to the pattern field

here just example of STIX types:
image

meirwah:
@anara123 what is blocking this PR? can we merge this?

meirwah:
@anara123 @yuvalbenshalom what is the status here? can we review and merge?

PR #3289, opened 2019-03-27
Snowflake rework

avidan-H:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15673

## Description
Reworks the pre-existing content and adds incident fetching

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Dependencies
- N/A

## Additional changes
Added Test Playbook.


Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+
+class DemistoError(Exception):
+    """Basic exception for errors raised by Demisto"""
+    def __init__(self, msg=None):
+        if not msg:
+            msg = 'A Demisto integration error occurred'
+        super(DemistoError, self).__init__(msg)
+
+
+class FetchError(DemistoError):
```
Itay4: lets discuss this

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    def __init__(self, msg=None):
+        if not msg:
+            msg = 'An error occurred whilst trying to fetch incidents'
+        super(FetchError, self).__init__(msg)
+
+
+'''HELPER FUNCTIONS'''
+
+
+def convert_datetime_to_string(v):
```
Itay4: can we generalize it to a function in common server ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    if isinstance(v, datetime):
+        return v.strftime('%Y-%m-%d %H:%M:%S.%f %z')
+    elif isinstance(v, date):
+        return v.strftime('%Y-%m-%d')
+    elif isinstance(v, dttime) or isinstance(v, timedelta):
+        return v.strftime('%H:%M:%S.%f')
+    return v
+
+
+def set_provided(params, key, val1, val2=None):
```
Itay4: in case both values are `None`, the value will be `None`, can that happen?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+    if val1:
+        params[key] = val1
+    elif val2:
+        params[key] = val2
+
+
+def process_table_row(row, checks):
+    """
+    Check row data and reformat if necessary
```
Itay4: please add example

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+        Dictionary where the key is a string indicative of the type (or bucket of types) that needs
+        reformatting and the values are a list of column names whose data is of that type
+
+    returns:
+        Reformatted Row
+    """
+    for column_name, val in row.items():
+        if column_name in checks.get('isDecimal', []):
+            # Then check the value and reformat it if necessary
+            if type(val) == Decimal:
```
Itay4: use `isinstance`

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+                row[column_name] = str(val)
```
Itay4: is this safe casting ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    name = 0
+    type_code = 1
+
+    checks = {}
+    # Screen by type_code
+    for col in column_descriptions:
+        if col[type_code] == 0:
+            # Then need to check that column's data to see if its data type is Decimal
+            checks.setdefault('isDecimal', []).append(col[name])
+        elif col[type_code] in {3, 4, 6, 7, 8, 12}:
```
Itay4: lets add a dictionary which translates this numbers to get a more readable code

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+            # Then need to check that column's data to see if its data type is date, time, timedelta or datetime
+            checks.setdefault('isDT', []).append(col[name])
+
+    # if 'results' is a list then it is a data table (list of rows) and need to process each row
+    # in the table, otherwise if 'results' is a dict then it a single table row
+    # Check candidates and reformat if necessary
+    if isinstance(results, dict):
+        results = process_table_row(results, checks)
+    else:
+        # if 'results' isn't a dict, assume it's a list
```
Itay4: why can we assume that?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    set_provided(params, 'warehouse', args.get('warehouse'), WAREHOUSE)
+    set_provided(params, 'database', args.get('database'), DATABASE)
+    set_provided(params, 'schema', args.get('schema'), SCHEMA)
+    set_provided(params, 'role', args.get('role'), ROLE)
+    if CERTIFICATE:
+        p_key = serialization.load_pem_private_key(CERTIFICATE, backend=default_backend())
+        pkb = p_key.private_bytes(
+            encoding=serialization.Encoding.DER,
+            format=serialization.PrivateFormat.PKCS8,
+            encryption_algorithm=serialization.NoEncryption())
```
Itay4: move closing parentheses to next line

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+        to which it belongs
+
+    returns:
+        Incident Object
+    """
+    incident = {}
+    occurred = row.get(DATETIME_COLUMN)
+    timestamp = None
+    if occurred:
+        if isinstance(occurred, dttime) or isinstance(occurred, timedelta):
```
Itay4: change to `isinstance(v occurred r, (dttime, timedelta))`

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+        err_msg = 'Nothing found when trying to fetch the datetime field specified in'
+        err_msg += ' the integration parameters. Please check that the name was correct.'
+        err_msg += ' If the field name was correct, verify that the returned value for'
+        err_msg += ' the specified field is not NULL for ALL of the rows to be fetched.'
+        raise FetchError(err_msg)
+    # Incident Title
+    if INCIDENT_NAME_COLUMN:
+        name = row.get(INCIDENT_NAME_COLUMN)
+    else:
+        name = 'Snowflake Incident -- ' + convert_datetime_to_string(occurred) + '-' + datetime.now().timestamp()
```
Itay4: why do we need two timestamps here? isnt there some unique key to use instead of the timestamps ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    incident['name'] = name
+    # Incident occurrence time as timestamp - the datetime field specified in the integration parameters
+    incident['timestamp'] = timestamp
```
Itay4: why not `occurred` ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+
+    column_descriptions, data = snowflake_query(args)
+    # convert the data/events to demisto incidents
+    incidents = []
+    for row in data:
+        incident = row_to_incident(column_descriptions, row)
+        incident_timestamp = incident.get('timestamp')
+
+        # Update last run and add incident if the incident is newer than last fetch
+        if incident_timestamp and incident_timestamp >= last_fetch:
```
Itay4: shouldn't the operator be `>` ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+            if first_fetch:
```
Itay4: can't we query by the timestamp and avoid all that ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                return_error('No data found matching the query')
+        except snowflake.connector.errors.ProgrammingError as e:
+            return_error(str(e))
+        finally:
+            if cur:
+                cur.close()
+            connection.close()
+    else:
+        return_error('Unable to connect')
```
Itay4: can we improve the error message?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+def snowflake_query_command():
+    args = demisto.args()
+    query = args.get('query')
+    db = args.get('database') if args.get('database') else DATABASE
```
Itay4: what happens in case neither are given ?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+            rows = args.get('rows')
+            if rows:
+                rows = int(rows)
+            if rows and rows > MAX_ROWS:
+                rows = MAX_ROWS
+            results = cur.fetchmany(rows)
+            if results:
+                return cur.description, results
+            else:
+                return_error('No data found matching the query')
```
Itay4: why is this an error?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+        except snowflake.connector.errors.ProgrammingError as e:
```
Itay4: lets discuss this

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+            return_error(str(e))
+        finally:
```
Itay4: can we add it to the `finally` in the EXECUTION block? check if connection, then..

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+        'Query': query,
+        'Result': results
+    }
+    columns = argToList(args.get('columns'))
+    human_readable = tableToMarkdown(query, results, columns)
+    demisto_transform = 'Snowflake(val.Query && val.Query === obj.Query'
+    demisto_transform += ' && val.Database && val.Database === obj.Database'
+    demisto_transform += ' && val.Schema && val.Schema === obj.Schema)'
+
+    demisto.results({
```
Itay4: use `return_outputs`

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    results = format_to_json_serializable(col_descriptions, results)
+
+    entry_context = {
+        'Database': db,
+        'Schema': schema,
+        'Query': query,
+        'Result': results
+    }
+    columns = argToList(args.get('columns'))
+    human_readable = tableToMarkdown(query, results, columns)
```
Itay4: add `removeNull=True`

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+  commands:
+  - arguments:
+    - default: true
+      description: The query to execute
+      isArray: false
+      name: query
+      required: true
+      secret: false
+    - default: false
+      description: Warehouse to use for the query. Will use default if not provided.
```
Itay4: what default ?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      isArray: false
+      name: warehouse
+      required: false
+      secret: false
+    - default: false
+      description: Database to use for the query. Will use default if not provided.
```
Itay4: what default ?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      isArray: false
+      name: database
+      required: false
+      secret: false
+    - default: false
+      description: Schema to use for the query. Will use default if not provided.
```
Itay4: what default ?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      isArray: false
+      name: schema
+      required: false
+      secret: false
+    - default: false
+      description: Role to use for the query. Will use default if not provided.
```
Itay4: what default ?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      isArray: false
+      name: role
+      required: false
+      secret: false
+    - default: false
+      defaultValue: '100'
+      description: Number of rows to retrieve
+      isArray: false
+      name: rows
```
Itay4: i think arg name should be `limit`

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      required: false
+      secret: false
+    - default: false
+      description: Optional list of columns to display in the given order
```
Itay4: need to say these need to be comma separated and provide example

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+      type: Unknown
+    - contextPath: Snowflake.Database
+      description: The name of the database object
+      type: String
+    - contextPath: Snowflake.Schema
+      description: The name of the schema object
+      type: String
+  - arguments:
+    - default: true
+      description: The command to execute
```
Itay4: can we provide the optional commands ?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
+category: Database
+commonfields:
+  id: Snowflake
+  version: -1
+configuration:
```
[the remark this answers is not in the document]
avidan-H: We typically use that param when using the requests library. Because this integration uses snowflake's python module and not a REST API, I don't see how it would be used in this context.

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,179 @@
...
+- display: Default schema to use
+  name: schema
+  required: false
+  type: 0
+- display: Default role to use
+  name: role
+  required: false
+  type: 0
+- display: ┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉‎                                                                                  Fetch
+    query to retrieve new incidents
```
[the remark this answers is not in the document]
avidan-H: The first word of the `Display name` for this parameter is actually not 'query', it's the word 'Fetch' on the previous line and its first letter is already capitalized. You want the first letter of 'query' to be capitalized as well?

Integrations/Snowflake/Snowflake.py
```diff
@@ -0,0 +1,392 @@
...
+    del os.environ['HTTP_PROXY']
+    del os.environ['HTTPS_PROXY']
+    del os.environ['http_proxy']
+    del os.environ['https_proxy']
+
+if IS_FETCH and not (FETCH_QUERY and DATETIME_COLUMN):
+    err_msg = 'When fetching is enabled there are two additional parameters that are required;'
+    err_msg += ' The fetch query that determines what data to fetch and the name of the column'
+    err_msg += ' in the fetched data that contains a datetime object or timestamp.'
+    return_error(err_msg)
```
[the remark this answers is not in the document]
avidan-H: I changed it to `raise FetchError(err_msg)` and it looks like -
Error Message 1

Integrations/Snowflake/Snowflake.py
avidan-H: I was worried about loss of precision when casting to a float...

Integrations/Snowflake/Snowflake.py
avidan-H: A query returns either a table ( a list of rows where each row is a dict ) or a row ( which is a dict ) even in the case where one cell is being returned it's still returned as a dict.

Integrations/Snowflake/Snowflake.py
avidan-H: I needed a way to make incident names unique in the case that the user configuring the instance of the integration didn't supply the name of a table column whose value would be used in assigning the incident name (Maya's advice was to not make it required)

Integrations/Snowflake/Snowflake.py
avidan-H: I want to discuss this with you cause I have questions about this bit.

Integrations/Snowflake/Snowflake.py
avidan-H: I was allowing for two different cases 
a) timestamps for different alert entries are the same - could be that they are ingesting alerts into their database from a variety of sources
b) if they specified a column whose values are of type date

Integrations/Snowflake/Snowflake.py
avidan-H: Not if we allow for identical timestamps for different alerts as I described that I was doing in the above comment

Integrations/Snowflake/Snowflake.py
avidan-H: I switched `demisto.results` for `return_outputs` for the `snowflake-query` command as per your request but for the `snowflake-update` command I'm only returning a string to the war room and to do that with the `return_outputs` wrapper it would look like 
```
return_outputs(outputs={}, readable_output='Operation executed successfully')
```
as opposed to 
```
demisto.results('Operation executed successfully')
```
which I think actually looks cleaner in this particular case.  

---
Do you still want me to switch it there as well?

Integrations/Snowflake/Snowflake.yml
avidan-H: To address this point I changed the command argument descriptions to be more specific - for this particular command argument the example being 
`Warehouse to use for the query. Will use the warehouse value specified in the integration parameters if it is not provided here.`

Integrations/Snowflake/Snowflake.yml
avidan-H: Wouldn't that be limiting for the user in terms of the SQL commands they could execute - less flexible and potentially more confusing for the user?

Integrations/Snowflake/Snowflake.py
avidan-H: I made `warehouse` and `database` required parameters as we discussed so those won't cause an issue - the others will be `None` but it shouldn't be an issue really - it depends on the command given I suppose.

Integrations/Snowflake/Snowflake.py
avidan-H: Should it be a regular war room message instead?

Integrations/Snowflake/Snowflake.py
avidan-H: can we confer in person about this?

Integrations/Snowflake/Snowflake.yml
```diff
@@ -0,0 +1,208 @@
...
+- defaultvalue: 'true'
+  display: Use system proxy settings
+  name: proxy
+  required: false
+  type: 8
+- display: Incident type
+  name: incidentType
+  required: false
+  type: 13
+- display: Trust server certificate (insecure)
```
Itay4: this should be placed above the incident type param, below the proxy

PR #3290, opened 2019-03-27
stix 1 example upload

jochman:
just an xml example (from cybox)


Itay4:
Don’t merge yet

jochman:
will use file from Cybox repo instead

PR #3302, opened 2019-03-29
utf subject bug fix

glicht:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/15344

## Description
Fixes a bug in convert_to_unicode where headers which are encoded in multiple tokens were not decoded.

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Scripts/ParseEmailFiles/ParseEmailFiles.py
+++ b/Scripts/ParseEmailFiles/ParseEmailFiles.py
@@ -3249,8 +3249,7 @@ def convert_to_unicode(s):
             else:
                 res += decoded_s
         return res.strip()
-    except Exception as ex:
-        print('exception: {}'.format(ex))
+    except Exception:
         for file_data in ENCODINGS_TYPES:
             try:
                 s = s.decode(file_data).encode('utf-8').strip()
--- a/Scripts/ParseEmailFiles/ParseEmailFiles.py
+++ b/Scripts/ParseEmailFiles/ParseEmailFiles.py
@@ -3241,13 +3241,19 @@ def get_utf_string(text, field):
 def convert_to_unicode(s):
     global ENCODINGS_TYPES
     try:
-        encoded_s, encoding = decode_header(s)[0]
-        s = encoded_s.decode(encoding).encode('utf-8').strip()
-        ENCODINGS_TYPES.add(encoding)
-    except Exception:
+        res = ''  # utf encoded result
+        for decoded_s, encoding in decode_header(s):  # return a list of pairs(decoded, charset)
+            if encoding:
+                res += decoded_s.decode(encoding).encode('utf-8')
+                ENCODINGS_TYPES.add(encoding)
+            else:
+                res += decoded_s
+        return res.strip()
+    except Exception as ex:
+        print('exception: {}'.format(ex))
         for file_data in ENCODINGS_TYPES:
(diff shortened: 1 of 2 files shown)


glicht:
unit test which fails the build before fix is committed (TDD): https://circleci.com/gh/demisto/content/20592

Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3241,13 +3241,19 @@ def get_utf_string(text, field):
...
+        res = ''  # utf encoded result
+        for decoded_s, encoding in decode_header(s):  # return a list of pairs(decoded, charset)
+            if encoding:
+                res += decoded_s.decode(encoding).encode('utf-8')
+                ENCODINGS_TYPES.add(encoding)
+            else:
+                res += decoded_s
+        return res.strip()
+    except Exception as ex:
+        print('exception: {}'.format(ex))
```
Itay4: did you mean to log the exception here ?

Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3241,13 +3241,19 @@ def get_utf_string(text, field):
...
-    except Exception:
+        res = ''  # utf encoded result
+        for decoded_s, encoding in decode_header(s):  # return a list of pairs(decoded, charset)
+            if encoding:
+                res += decoded_s.decode(encoding).encode('utf-8')
+                ENCODINGS_TYPES.add(encoding)
+            else:
+                res += decoded_s
+        return res.strip()
+    except Exception as ex:
```
Itay4: shouldn't we catch only UnicodeDecodeError exceptions here?

Scripts/ParseEmailFiles/parse_email_files_test.py
```diff
@@ -197,3 +197,13 @@ def executeCommand(name, args=None):
     assert 'test - inner attachment eml.eml' in results[0]['EntryContext']['Email']['Attachments']
     assert isinstance(results[0]['EntryContext']['Email'], dict)
     assert results[0]['EntryContext']['Email']['Depth'] == 0
+
+
+def test_utf_subject_convert():
+    subject = ('[TESTING] =?utf-8?q?=F0=9F=94=92_=E2=9C=94_Votre_colis_est_disponible_chez_votre_co?='
```
Itay4: what about adding some special chars here?

Scripts/ParseEmailFiles/ParseEmailFiles.py
glicht: No it shouldn't be there. Good catch.

Scripts/ParseEmailFiles/ParseEmailFiles.py
glicht: There are different errors that can be raised. We want to catch them all (this is also how it was before my fix) as we don't want to fail the email parsing because of a faulty header. At most we return the original header without decoding.

Scripts/ParseEmailFiles/parse_email_files_test.py
glicht: This encoding contains special chars. This is how it looks decoded:

image

PR #3308, opened 2019-03-31
fix ParseEmailFiles script by using getaddresses method instead of regex

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16168

## Description
fix ParseEmailFiles script by using `getaddresses` method instead of regex to avoid duplicates.

## Screenshots
before the fix:
image
 after:
image


adi88d:
> do the same for `To` and `CC`

done

PR #3318, opened 2019-04-02
Updated the Zscaler test

guykeller:
## Status
Ready

## Description
Updated the Zscaler test to include less simultaneous IP requests.

   - No

## Must have
- [ ] Code Review


TestPlaybooks/playbook-Zscaler_Test.yml
```diff
@@ -1,5 +1,5 @@
 id: Zscaler Test
-version: -1
+version: 13
```
ronykoz: changing to -1

PR #3324, opened 2019-04-03
Phish.AI integration logic and detonate url playbook fixes

bakatzir:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16435

## Description
fix logic of check-status, add outputs, fix logic of detonate playbook, add a meaningful test to it.

   - Yes
phish-ai-scan-url - change of context from DBotScore.ScanID to PhishAI.ScanID
since we shouldn't save the scanid under DBotScore
phish-ai-check-status - change of argument from url to scan_id
since we want to pull on a specific scan its status, and not scan a new url.

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/16455
- [x] Code Review

### DIFF
--- a/Integrations/integration-PhishAI.yml
+++ b/Integrations/integration-PhishAI.yml
@@ -42,6 +42,7 @@ script:
         var report = JSON.parse(res.Body);
 
         md = tableToMarkdown('Phish.AI Scan report ' + scan_id, report);
+
         return {
             Type: entryTypes.note,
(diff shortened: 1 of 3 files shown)


kirbles19:
> changes approved.
> all we need is for @kirbles19 to approve docs, then @anara123 will force merge.

Approved (and updated in documentation).

PR #3331, opened 2019-04-03
Added logic to ensure malicious indicator is found in phishing tests

idovandijk:
## Status
Ready

## Description
- Added tasks that add (and at the end, also remove) a malicious URL to Minemeld miner. This way, Minemeld, which we rely on in the phishing tests, will always find this indicator malicious.

## Screenshots
image

### DIFF
--- a/TestPlaybooks/playbook-Phishing_test_-_Inline.yml
+++ b/TestPlaybooks/playbook-Phishing_test_-_Inline.yml
@@ -5,10 +5,10 @@ starttaskid: "0"
 tasks:
   "0":
     id: "0"
-    taskid: fc2ec0f9-a350-46e6-82c6-66920f1418ec
+    taskid: 4fc9bbe3-e88c-4ef7-8bc5-25f1b292ca72
     type: start
     task:
-      id: fc2ec0f9-a350-46e6-82c6-66920f1418ec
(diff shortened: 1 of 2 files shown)


yardensade:
I don't think it should be removed.
Several test like this running at the same time could in theory cause conflict when the indicator is removed yet another phishing playbook is trying to find it at that time.
Better solution is simply to keep the add task and ignore errors, that way if the miner exists already, you will skip it and move on.
Another less lazy way would be to check first if the indicator exists in that miner, only if it doesn't -> add it. That should be bullet proof.
But in that case the lazy way should do.

idovandijk:
Won't re-adding it all the time just flood Minemeld with duplicates?


> I don't think it should be removed.
> Several test like this running at the same time could in theory cause conflict when the indicator is removed yet another phishing playbook is trying to find it at that time.
> Better solution is simply to keep the add task and ignore errors, that way if the miner exists already, you will skip it and move on.
> Another less lazy way would be to check first if the indicator exists in that miner, only if it doesn't -> add it. That should be bullet proof.
> But in that case the lazy way should do.

yardensade:
Nope.
image
image

idovandijk:
I took your advice. Thanks.

idovandijk:
Green build :)

PR #3338, opened 2019-04-04
AD Query: fix error message with faulty server

glicht:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16446

## Description
Better error handling for when the ldap server mis-behaves

   - No

## Must have
- [x] Tests
- [ ] Code Review


glicht:
TDD: build with test that fails with similar scenario to the issue

Integrations/Active_Directory_Query/Active_Directory_Query.py
```diff
@@ -912,9 +912,10 @@ def main():
             demisto.info(message)
             return_error(message)
             return
-    except LDAPSocketOpenError as e:
+    except Exception as e:
         exc_msg = str(e)
-        demisto.info(exc_msg)
+        demisto.info("Failed bind to: {}:{}. {}: {}".format(SERVER_IP, PORT, type(e), exc_msg
```
anara123: this should be `demisto.error`

glicht: Depends. I am ok with info (what it was also before) as it is in the flow of "test"

yardensade:
Can we close this branch?

PR #3339, opened 2019-04-05
Don't trigger sample tests run after a change in non yml related files in a package

ronykoz:
## Status
In Progress

## Description

- [x] Don't trigger sample tests run after a change in non yml related files
- [x] Run configure_tests test on Circle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,15 @@ jobs:
               - "02:df:a5:6a:53:9a:f5:5d:bd:a6:fc:b2:db:9b:c9:47" # disable-secrets-detection
               - "f5:25:6a:e5:ac:4b:84:fb:60:54:14:82:f1:e9:6c:f9" # disable-secrets-detection
       - run:
+          name: Create ID Set
+          command: |
+            python ./Tests/scripts/update_id_set.py -r
+      - run:
           name: Infrastucture testing
           command: |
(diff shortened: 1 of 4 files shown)


glicht:
@ronykoz I committed a unit test I wrote before the fix. It is passing now. Also created the dev-task: https://github.com/demisto/etc/issues/16474 to fix the other failing tests.

ronykoz:
@glicht should we spend time adding more tests here or should I merge it this way?

.circleci/config.yml
```diff
@@ -28,15 +28,16 @@ jobs:
...
       - run:
           name: Infrastucture testing
           command: |
             pytest ./Tests/scripts/hook_validations/tests/ -v
             pytest ./Tests/scripts/infrastructure_tests/ -v
-      - run:
-          name: Create ID Set
-          command: |
-            python ./Tests/scripts/update_id_set.py -r
+            python ./Tests/scripts/test_configure_tests.py
```
glicht:
```suggestion
            pytest ./Tests/scripts/test_configure_tests.py -v
```
Let's run the tests with pytest. The output is cleaner and easier to understand when there is a failure.

glicht:
Looks good. Don't think we need more tests.

ronykoz:
@glicht good to go

glicht:
feel free to merge

PR #3346, opened 2019-04-07
Email Address Enrichment - Generic v2 + Test Playbook

idovandijk:
## Status
Ready

## Related Issues

## Description
- Fixed a bug where we didn't provide any value for the "domain" input of the playbook. It now takes unique domains from `Domain.Name` as an input.
- Created a new test playbook for the `Email Address Enrichment - Generic` playbook. The old test has been in our content, but hasn't been used anywhere until now (maybe it should be deleted). The test currently tests some of the functionality but not all of it, due to 2 issues with the playbook (it seems domain squatting and determining whether email addresses are internal or external, are both bugged).
- Added `Uniq` transformer to any task that deals with indicators, to eliminate duplication of indicators
- Completely removed the GetReputation flag from the playbook inputs and from the utilization of that flag throughout the playbook, in accordance with the decision to remove reputation from the enrichment playbooks
- Changed playbook to meet playbook conventions (wording of section headers, descriptions of tasks and more)
- Added "ignore case" in condition tasks across the playbook
- Updated and improved descriptions for tasks
- Updated playbook description
- Reorganized and simplified overall visualization of playbook
- Changed old DT selectors to new friendly selectors in various locations
- Set "Autoextract" correctly for all playbook tasks + test tasks. New playbooks should not use `Use System Default`, but rather `None`, `Out of Band` or `Inline`

## Screenshots
New playbook
image

New test playbook
image

4.1.0 and higher

   - Yes - GetReputation flag used to mandatory in playbook inputs, and now doesn't exist.

## Must have
- [x] Tests
- [ ] Code Review


Itay4:
@idovandijk what is the status here?

idovandijk:
> @idovandijk what is the status here?

Pending some changes since they decided to make it all v2 instead of replacing existing playbooks. This is true for all my open PRs. It's going to go into next release.

PR #3354, opened 2019-04-10
moved virus total - private API to unmockable

DeanArbel:
## Status
Ready

## Description
moved virus total - private API to unmockable due to it failing on master.


ShahafBenYakir:
👍

PR #3358, opened 2019-04-11
vt private add fullResponseGlobal param

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16483

## Description
add fullResponseGlobal: Determines whether to return all results, which can number in the thousands. If “true”, returns all results and overrides the _fullResponse_ argument (if set to “false”) in a command. If “false”, the _fullResponse_ argument in the command determines how results are returned.

   - No

## Must have
- [x] Tests
- [x] Documentation https://github.com/demisto/etc/issues/16607
- [ ] Code Review


yuvalbenshalom:
@bakatzir build fails

PR #3361, opened 2019-04-11
Fix fetch credentials

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/16530
https://github.com/demisto/etc/issues/16440

## Description
Add identifier argument to HashiCorp
Workaround server bug


Scripts/script-CommonServerPython.yml
```diff
@@ -2,6 +2,7 @@ commonfields:
   id: CommonServerPython
   version: -1
 name: CommonServerPython
+toversion: 4.2.0
```
orius123: What is this for?

Integrations/integration-HashiCorp-Vault.yml
```diff
@@ -620,14 +621,17 @@ script:
...
-                    return_error('Version not configured for KV engine, Re-configure the engine')
+                    return_error('Version not configured for KV engine, re-configure the engine')
                 if engine['version'] == '1':
                     credentials += get_kv1_secrets(engine['path'])
                 elif engine['version'] == '2':
                     credentials += get_kv2_secrets(engine['path'])
             elif engine['type'] == 'Cubbyhole':
                 credentials += get_ch_secrets(engine['path'])
 
+        if identifier:
```
orius123: No way to query hashicorp for this specific record? 
Also need to make sure it returns only 1 value

liorblob: @orius123 There is a way, however, I have no way of knowing which engine I need when I have the identifier and if I query an engine that doesn't have that identifier I get a 404 error.
Users might configure more than 1 set of credentials, so I think an error in that case is the way to go.

Scripts/script-CommonServerPython.yml
liorblob: This is for versions before the fix, so the integrations will still work

Integrations/integration-HashiCorp-Vault.yml
orius123: but if we fetch the wrong engine we will get no results also with this lambda, no?

liorblob: We fetch everything anyway, then filter according to the identifier

liorblob:
@ronykoz Note the change in id.py - added check for the reversed scenario

Tests/scripts/hook_validations/id.py
```diff
@@ -163,7 +163,8 @@ def is_id_duplicated(self, obj_id, obj_data, obj_type):
                             is_duplicated = True
                             break
 
-                    elif LooseVersion(obj_fromversion) <= LooseVersion(instance_to_version):
+                    elif not (LooseVersion(obj_fromversion) > LooseVersion(instance_to_version)
```
ronykoz: Lets change it to <= and also change it to and condition instead

liorblob: @ronykoz an `and` condition means that either both of them are wrong or both of them are ok. The first case is impossible, and the second case means `is_duplicated` should be false. So I don't see how we can make it an `and` condition.

ronykoz: You’ve done not(a or b) which means not a and not b, it would be more readable

liorblob: ok, and why do we need the `=`?

ronykoz: Because you want to make sure the versions don’t collide, we must have a gap between them otherwise they both turn up as valid, and you won’t be able to tell which of them was loaded

liorblob: but it's a `not`, so it's the other way around

PR #3363, opened 2019-04-11
fix file validation

ronykoz:
fix file validation in master branch


ronykoz:
Tested on master branch mode, seem to work

PR #3365, opened 2019-04-11
Lastline detonation fixes

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16436

## Description
add lastline-check-status command, add outputs to detonation playbooks

   - No

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/16544
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Lastline.yml
+++ b/Integrations/integration-Lastline.yml
@@ -1009,7 +1009,4 @@ script:
       type: string
     description: Check submissions status
   runonce: false
-tests:
-- Detonate URL - Generic Test
-- Detonate File - Generic Test
 releaseNotes: Add ___lastline-check-status___ command
--- a/Integrations/integration-Lastline.yml
+++ b/Integrations/integration-Lastline.yml
@@ -1010,7 +1010,6 @@ script:
     description: Check submissions status
   runonce: false
 tests:
-- Lastline - testplaybook
 - Detonate URL - Generic Test
 - Detonate File - Generic Test
 releaseNotes: Add ___lastline-check-status___ command
--- a/Integrations/integration-Lastline.yml
+++ b/Integrations/integration-Lastline.yml
@@ -459,7 +459,6 @@ script:
                 'UUID': uuid,
                 'Status': 'Completed',
             };
-            //ec.Lastline.Submission.yara_signatures = '000000';
             var analysis = result.obj.data;
 
             if('report' in result.obj.data) {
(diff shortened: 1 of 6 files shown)


Integrations/integration-Lastline.yml
```diff
@@ -612,6 +639,18 @@ script:
                 }
                 return entries;
 
+            case 'lastline-check-status':
+                var uuids = argToList(args.uuid);
+                var entries = [];
+                for (var i=0; i < uuids.length; i++) {
+                    var reportResponse = CheckStatus(uuids[i]);
+                    if (reportResponse.error){
```
yaakovi: you don't have an error in the response, it's either a string or an entry dictionary.
maybe you need try-catch?

Integrations/integration-Lastline.yml
```diff
@@ -958,8 +997,21 @@ script:
       description: 'Tasks after in UTC. format: %Y-%m-%d %H:%M:%S. Example: 2018-07-10
         12:00:00'
     description: Gets a list of tasks
+  - name: lastline-check-status
+    arguments:
+    - name: uuid
+      required: true
+      description: Task UUID of the submitted Lastline analysis
```
yaakovi: where do you get the UUID from? specify the command

Playbooks/playbook-Detonate_File_-_Lastline.yml
```diff
@@ -568,5 +610,24 @@ outputs:
 - contextPath: Lastline.Submission.YaraSignatures.internal
   description: True if the signature is only for internal usage
   type: boolean
+- contextPath: URL
```
yaakovi: why detonate file has url in the output?

Playbooks/playbook-Detonate_URL_-_Lastline.yml
```diff
@@ -425,5 +472,25 @@ outputs:
 - contextPath: Lastline.Submission.Process.executable.yara_signature_hits
   description: Yara signature of the executable of the process
   type: string
+- contextPath: URL
+  description: URL object
+  type: unknown
+- contextPath: URL.Malicious
+  description: URL Malicious object
+  type: unknown
+- contextPath: File
```
yaakovi: does this supposed to be InfoFile?

Tests/conf.json
```diff
@@ -1213,6 +1213,7 @@
         }
     ],
     "skipped_tests": {
+        "Lastline - testplaybook": "Chckeing the integration via Generic detonation playbooks, don't want to load the quota",
```
yaakovi:
```suggestion
        "Lastline - testplaybook": "Checking the integration via Generic detonation playbooks, don't want to load the quota",
```

Playbooks/playbook-Detonate_URL_-_Lastline.yml
bakatzir: it's a mistake.
no infofile -you can't get the original sample in lastline.

yaakovi: InfoFile is for the reports.
File is for the original sample

Integrations/integration-Lastline.yml
bakatzir: i think it's not needed. since i catch the report not ready error and return it as a string. other errors should stop the execution.

PR #3366, opened 2019-04-11
fix bug in DeleteContext when index equals to zero

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16485

## Description
fix bug in DeleteContext - the scrip refer index param as undefined when it equals to zero


Scripts/script-DeleteConext.yml
```diff
@@ -53,7 +53,7 @@ script: |
           ReadableContentsFormat: formats.markdown,
           EntryContext: keysToKeepObj
       };
-  } else if (args.index) {
+  } else if (args.index != undefined)) {
```
glicht: Best to use strict compare
```suggestion
  } else if (args.index !== undefined)) {
```

glicht: Never mind. For undefined strict compare is not needed.

anara123:
@adi88d fix the conflict

Tests/conf.json
```diff
@@ -1385,7 +1472,13 @@
         "JoeSecurityTestDetonation": "command joe-download-report fails (issue 16118)",
         "af2f5a99-d70b-48c1-8c25-519732b733f2": "Added here because test existed but was not configured - need to review the test",
         "Zoom_test": "Added here because test existed but was not configured - need to review the test",
-        "Test - CrowdStrike Falcon": "Test is unmockable, and has its data deleted every few weeks"
+        "Test - CrowdStrike Falcon": "Test is unmockable, and has its data deleted every few weeks",
+        "block_endpoint_-_carbon_black_response_-_test": "Test for deprecated integration",
+        "get_file_sample_from_path_-_generic_-_test": "Test for deprecated integration",
+        "get_file_sample_from_path_-_carbon_black_enterprise_response_-_test": "Test for deprecated integration",
```
anara123: if the integration is deprecated, we can delete the test

PR #3369, opened 2019-04-13
Snowflake: small doc and naming fixes of params

glicht:
## Status
Ready/

## Description
Change order so `Account` field is at start. Rename `trust any certificate`. Update detailed description to use `Authenticator`.

## Screenshots
image

image

   - No

## Must have
- [x] Tests (existing)
- [ ] Code Review

### DIFF
--- a/Integrations/Snowflake/Snowflake.yml
+++ b/Integrations/Snowflake/Snowflake.yml
@@ -3,14 +3,14 @@ commonfields:
   id: Snowflake
   version: -1
 configuration:
-- display: Username
-  name: credentials
-  required: true
-  type: 9
 - display: Account - See Detailed Description section.
   name: account
   required: true
   type: 0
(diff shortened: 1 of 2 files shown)


Integrations/Snowflake/Snowflake.yml
```diff
@@ -40,7 +40,7 @@ configuration:
   name: proxy
   required: false
   type: 8
-- display: Trust server certificate (insecure)
+- display: Trust any certificate (not secure)
```
avidan-H: I thought it was agreed that we were switching back from `unsecure` (since it's improper english) to `insecure` as a convention?

glicht: we finalised on `not secure`. See: https://github.com/demisto/etc/issues/14980

PR #3379, opened 2019-04-15
WildFire - Add get-verdict, upload-file-url commands, deprecate upload-remote-file

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16456
fixes: https://github.com/demisto/etc/issues/16586
fixes: https://github.com/demisto/etc/issues/16588
reference: https://github.com/demisto/etc/issues/16584
reference: https://github.com/demisto/etc/issues/16583

break BC?
   - No

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/16576
- [x] Code Review

### DIFF
--- a/Integrations/integration-Wildfire.yml
+++ b/Integrations/integration-Wildfire.yml
@@ -5,7 +5,7 @@ name: WildFire
 fromversion: 4.0.0
 display: Palo Alto Networks WildFire
 category: Forensics & Malware Analysis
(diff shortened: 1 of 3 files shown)


Integrations/integration-Wildfire.yml
```diff
@@ -136,14 +146,47 @@ script:
...
+        prettyVerdict.Verdict = Number(verdictData.verdict);
+        prettyVerdict.VerdictDescription = VERDICTS_DICT[verdictData.verdict];
+
+        return prettyVerdict;
+    }
+
+
+    function createDBotScoreFromVerdict(prettyVerdict) {
+        var DbotScore = {
+            'Indicator': prettyVerdict.SHA256? prettyVerdict.SHA256 : prettyVerdict.MD5,
```
yuvalbenshalom: This function needs to log error in case hash is missing or in case VERDICTS_TO_DBOTSCORE will not have a value for the key.

Integrations/integration-Wildfire.yml
```diff
@@ -180,13 +223,14 @@ script:
...
-                EntryContext: {
-                    "WildFire.Report(val.SHA256 == obj.SHA256)": prettifyUploadBody(body)
+        return {
+            Type: entryTypes.note,
+            Contents: result,
+            ContentsFormat: formats.json,
+            HumanReadable: md,
+            ReadableContentsFormat: formats.markdown,
+            EntryContext: {
+                "WildFire.Report(val.SHA256 == obj.SHA256)": prettifyUploadBody(body)
```
yuvalbenshalom: are we sure there is SHA256 and not just MD5 in this case also?

bakatzir: this is the report entry, it's old code, and checking for md5 is indeed missing here.

Integrations/integration-Wildfire.yml
bakatzir: will add this check here here and also in `createDBotScoreFromVerdicts`
product writer will phrase the error

PR #3380, opened 2019-04-15
fix demisto.results to decode bytes to unicode

noaco:
fixes: https://github.com/demisto/etc/issues/15721

in progress - adding test


noaco:
@anara123 please approve

PR #3382, opened 2019-04-16
Fix vault for Javascript integrations

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/16440


liorblob:
@Itay4 tested on 4.1.0. I think we can force merge since only Jask is failing on some unrelated error.

PR #3383, opened 2019-04-16
Fix ExtractIndicatorsFromWordFile - get file type

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16472

## Description
get_file_details method failed when file entry is not a list


adi88d:
@yaakovi I tested it with one and more files in context

kirbles19:
RN: The automation executes as expected when the entry is a single object.

PR #3388, opened 2019-04-16
Test filter by version

yaakovi:
## Status
Ready

## Description
support test playbooks that should run only on specific versions.
this way we can test features in newer versions that are not supported in previous ones.

## Screenshots
on 4.1 and above:
image

on 4.0:
image

## Related PRs

branch | PR
------ | ------

x.x.x

   - Yes
       - Further details:
   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes


Tests/test_content.py
```diff
@@ -564,7 +586,7 @@ def main():
                 print_color("Starting tests for {}".format(ami_instance_name), LOG_COLORS.GREEN)
                 print("Starts tests with server url - https://{}".format(ami_instance_ip))
                 server = SERVER_URL.format(ami_instance_ip)
-                execute_testing(server, ami_instance_ip, server_version)
+                execute_testing(server, ami_instance_ip, server_version, server_numeric_version)
                 sleep(8)
 
     else:  # Run tests in Server build configuration
```
ronykoz: shouldn't we add the server_version to line 596 as well?

Tests/test_content.py
```diff
@@ -411,14 +413,26 @@ def run_test_scenario(t, c, proxy, default_test_timeout, skipped_tests_conf, nig
...
         return
 
     # Skip integration
     if has_skipped_integration:
         return
 
+    # Skip version mismatch test
+    test_from_version = t.get('fromversion', '0.0.0')
+    test_to_version = t.get('toversion', '99.99.99')
+    if (server_version_compare(test_from_version, server_numeric_version) > 0
```
ronykoz: mmm shouldn't it be `server_version_compare(test_from_version, server_numeric_version) <= 0`?

Tests/test_content.py
```diff
@@ -546,14 +560,22 @@ def main():
...
 
     if is_ami:  # Run tests in AMI configuration
         with open('./Tests/images_data.txt', 'r') as image_data_file:
             image_data = [line for line in image_data_file if line.startswith(server_version)]
             if len(image_data) != 1:
                 print('Did not get one image data for server version, got {}'.format(image_data))
             else:
+                server_numeric_version = re.findall('Demisto-Circle-CI-Content-[\w-]+-([\d.]+)-[\d]{5}', image_data[0])
+                if not server_numeric_version:
+                    server_numeric_version = '99.99.98'  # latest
```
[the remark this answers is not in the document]
yaakovi: since the toversion might be 99.99.99 that means it doesn't include version 99.99.99 itself.

Tests/test_content.py
yaakovi: no, negative means that the from version is lower that the server version. in this case we don't want to skip

PR #3405, opened 2019-04-21
VT - fix duplicate keys in IP.Address context key

bakatzir:
@GavrielFilippov  FYI 

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16488

   - No

## Must have
- [x] Tests
- [x] Documentation - no need
- [ ] Code Review

### DIFF
--- a/Integrations/integration-VirusTotal.yml
+++ b/Integrations/integration-VirusTotal.yml
@@ -452,7 +452,7 @@ script:
             }
         }
         ip_ec = {
-            'Address(val.ip === obj.ip)': ip,
+            'Address(val === obj)': ip,
             "VirusTotal": {
                 'DownloadedHashes': detectedDownloadedSamples.slice(0,maxLen),
                 'UnAVDetectedDownloadedHashes': undetectedDownloadedSamples.slice(0,maxLen),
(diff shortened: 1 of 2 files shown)


bakatzir:
i believe not, it's under: IP.Address, all items are as list
https://github.com/demisto/internal-content/tree/master/documentation/context_standards

yardensade:
should be added to tests imo, add something to context, add it again, make sure there is only 1 entry there and not a list or something of the sort.

bakatzir:
@yardensade would you mind have another look in the test playbook before merging?

yardensade:
it seems to me like you repeat the IP command 2 times in the entire playbook, yet you delete the context after the first time you execute this command, and therefore it does not simulate the situation where the context is being updated.

This is by looking at the yml, I did not download the actual file, so correct me if I'm wrong.

bakatzir:
@yardensade  834 tasks playbook can be counter intuitive to look at :)
Screen Shot 2019-04-21 at 20 02 57

yardensade:
True story, that picture is great, looks good to me.

bakatzir:
improve readability and checked for malicious ip as well.

PR #3407, opened 2019-04-21
Fix ADGetUser to also be compatible with AD Query V2

noaco:
fixes https://github.com/demisto/etc/issues/16366

low effort fix 
- AD v2 has a command equivalent to this script
- the script should be deprecated

### DIFF
--- a/Scripts/script-ADGetUser.yml
+++ b/Scripts/script-ADGetUser.yml
@@ -30,6 +30,8 @@ script: |-
       'PARTIAL_SECRETS_ACCOUNT': 0x04000000
   }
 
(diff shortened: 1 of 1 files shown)


Scripts/script-ADGetUser.yml
```diff
@@ -108,6 +110,7 @@ script: |-
       args['size-limit'] = limit
       if using:
           args['using'] = using
+      args['using-brand'] = BRAND # low effort fix (script should be dprecated)
```
anara123: dprecated = > deprecated

noaco:
A different approach will be taken with the script

PR #3413, opened 2019-04-22
ADGetUser - enable script execution with "Active Directory Query" instances only

noaco:
fixes https://github.com/demisto/etc/issues/16366


Scripts/script-ADGetUser.yml
```diff
@@ -189,7 +189,12 @@ script: |-
   else:
       resp = listAll(attrs, demisto.get(demisto.args(),'using'), userAccountControlOut)
 
+  expectedErrs = ['invalid attribute type dn', 'Unsupported Command']
```
anara123: expectedErrs => expected_errs
errData => err_data

TestPlaybooks/playbook-ADGetUser_fails_with_v2-Test.yml
```diff
@@ -0,0 +1,187 @@
+id: Test ADGetUser Fails with no instances 'Active Directory Query' (old version)
+version: 18
```
anara123: 18 => -1

Scripts/script-ADGetUser.yml
noaco: used the same style as appears on the script...

anara123:
```
Test ADGetUser Fails with no instances 'Active Directory Query' (old version) failed with error/s
Playbook Test ADGetUser Fails with no instances 'Active Directory Query' (old version) has failed:
	- Command: !ADGetUser extend-context="errData="
	- Body: Error from Scripts is : Command ADGetUser was not found in module supports list (88)
	- Command: Executing conditions:
Label: yes, Condition: [null Equals No instances of "Active Directory Query" are configured.
If you are trying to run "ADGetUser" with "Active Directory Query V2", please use the aquivilant command "ad-get-user" instead.]
	- Body: The output [] didn't satisfy any of the conditions of task 3 (31)
```

PR #3415, opened 2019-04-22
PagerDuty Enhancment

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16665
fixes: https://github.com/demisto/etc/issues/16666

## Description
Adding 2 use-cases for PagerDuty integration


ronykoz:
I gave our contact at SLB the yml file, waiting for his response - so we should wait for him before merging this thing.

Integrations/PagerDuty/PagerDuty.py
```diff
@@ -38,6 +38,8 @@
 '''SUFFIX ENDPOINTS'''
 GET_SCHEDULES_SUFFIX = 'schedules'
 CREATE_INCIDENT_SUFFIX = 'incidents'
+GET_INCIDENT_SUFFIX = 'incidents/'
+GET_SERVICES_SUFFIX = 'services?offset={}'
```
yaakovi: better to pass `offset` in a dictionary as `params` so the requests module will handle correct escaping (not necessary here but for future reference)

PR #3417, opened 2019-04-22
Microsoft Graph User

orenzohar:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16625
fixes: https://github.com/demisto/etc/issues/16035

## Description
Unified gateway to security insights - all from a unified Microsoft Graph User API.

x.x.x

## Must have
- [x] Tests
- [ ] Code Review


Integrations/MicrosoftGraphUser/MicrosoftGraphUser.py
```diff
@@ -0,0 +1,301 @@
...
+        if '@removed' in users_data:
+            user_readable['Status'] = 'deleted'
+        user_outputs = {k.replace(' ', ''): v for k, v in user_readable.copy().items()}
+
+        return user_readable, user_outputs
+
+
+def epoch_seconds(d=None):
+    """
+    Return the number of seconds for given date. If no date, return current.
```
yuvalbenshalom: remove the d parameter since you don't use it anyway, and update docstring

Integrations/MicrosoftGraphUser/MicrosoftGraphUser.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+    if not d:
+        d = datetime.utcnow()
+    return int((d - datetime.utcfromtimestamp(0)).total_seconds())
+
+
+def get_token():
```
yuvalbenshalom: If I understand correctly, In case token was revoked before time, we will get error until the round hour will pass.
maybe we can add `get_token` an argument to discard the old token anyway, than catch the relevant response and call it then.
also, what is the token life time? is it configurable?

Integrations/MicrosoftGraphUser/MicrosoftGraphUser.py
```diff
@@ -0,0 +1,301 @@
...
+    demisto.setIntegrationContext({'token': data.get('token'), 'stored': epoch_seconds()})
+    return data.get('token')
+
+
+def http_request(method, url_suffix, params=None, body=None):
+    """
+    Generic request to Microsoft Graph
+    """
+    token = get_token()
+    response = requests.request(
```
yuvalbenshalom: since there is no `try` here, is unreachable error is readable to the user in test and commands? dose it tell the use to check the address?

Integrations/MicrosoftGraphUser/MicrosoftGraphUser.yml
```diff
@@ -0,0 +1,244 @@
...
+    type: 0
+    required: true
+  - display: Token ID (as received from the admin consent - see detailed instructions (?))
+    name: token
+    defaultvalue: ""
+    type: 0
+    required: true
+  - display: Trust any certificate (unsecure)
+    name: unsecure
+    defaultvalue: "true"
```
yuvalbenshalom: This probably should be false

PR #3421, opened 2019-04-23
Google big query integration

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16135

## Description
A new integration for Google BigQuery.
Currently supporting the query command with several extra configuration arguments.

   - No

## Must have
- [x] Tests
- [ ] Code Review


Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+        bigquery_client = start_and_return_bigquery_client(demisto.params()['google_service_creds'])
+        query_job = bigquery_client.query(TEST_QUERY)
+        query_results = query_job.result()
+        results_rows_iterator = iter(query_results)
+        first_item = next(results_rows_iterator)
+        if str(first_item.get("name")) == "Frances":
+            raise ValueError("Data from DB not matching expected results")
+        demisto.results("ok")
+    except Exception as ex:
+        return_error(str(ex)) # not sure if should be ex.message
```
ronykoz: its ok this way, remove the comment

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
...56vDC28jmaLnbkQZ9VSoJhqxzzh08rmhguyXPqI9xxINzOotPsDkhbXavMZlWmZgeVjh93gbHYx5//kWYFZQkfLg0P2hAg3+M3xwQb1GFe0nacVaDz3XOB3m9xiU5UP64i/M9rzG5dq+6xnanqsZ2u4rc4OEX+P3eBt9ZYQRgnf/4IAwERe7IsyARnl1PBTX9IRKYteD7mNwfNb5gD+4nKG9/9fcIOHX+D3eRggQbp42Pp616wYBUVeaVtwPhM8+H/iuPzODhF/j94QB4RNaFviR1h1K6wPxYkcyZvqHUQckC2y6utN3bzWatzDKqtZGHZDcuh56oWNSH4h3dqRRSdP5frsfXUAkVn1vNTxyO+V5zfG6GGe/ohMIn1BOKFJEej5NBfKEwfHRAWE0OBL6ldh+0H4ussRpOqVXrfFQRdtJneqN+7qD+G9J8N7hbkzEalRBusynrDFuiE7lzbCbK990s6s/Ns732WGkweMMQgxs95JSB0BnzPF61CHKWVC+cY1rZWAG5fmuhs1Vlq96JprsbzLyZRmoL1gaADrZGxHHsiSwKagwwUtDLmV9rBEwKiasvb9sY0FezgIRx/BsUybCoN5Bz32G2CppZLlJTfNssrO94sEoG6MrCLIytN8KS/EB/CQBNS1WCflNIJgeOHvKfcrLlULd2YT5ibwZ5zHxB+JAicbBjgg1MV5KcjcuIk/TdFBrKnwzTsCcyk6LmIynBd04f0Lu1qctNi05CZ2dtbxkKiknyNmQaA0Il5qOTkLDmZyXrQDzMyRNLACPmgmlu+Ebt0MTG991Yhs6sUM34n6nvufYTfsMJ4G9VMjGigNR0+LRbTf84P4lW31wWdOtZlCx8qUpMPw9fZEyUxyI1rYpCA8uGQ5Kkr8I+5y39RFa4Ds8QiCXwVpZYk2q1C9FeIexw/BZm5yMmY01PbgBJ7egcF3oLxnWjl1Kc4Ty6USAPRBDSbTYGpZqacMGJf8TvnqBxEaS8yLqyKpJhnyqfzRgHPrzzzQsfIWweE0CuHw12euHMpVG5f7N+Nwiuh2Yf77hbj6mf84Rap6/YoO1x2cnL3m1Dp2e0h1RwtkuB5CSxhZGPL9zKj4UONX/NncAxEWS/E/FOcaSJ+mfHDlNhNFlrTP0l5x9weL38LF0GrWHq2m2/uLGpT4pWvX/zRDqQjZWrVoAAAAASUVORK5CYII=
```
ronykoz: remove the image from here and move it to the image file

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+description: Integration for Google BigQuery, a data warehouse for querying and analyzing
+  large databases
+detaileddescription: Use these detailed instructions in order to retrieve the API
```
ronykoz: move this to a separate file

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+    - contextPath: BigQuery.Query
+      description: The query performed.
+      type: String
+    - contextPath: BigQuery.Row
+      description: The table rows the given query returned.
+      type: Unknown
+    description: Perform a query on BigQuery
+  dockerimage: bigquery_image_local_3
+  runonce: false
+tests: Google BigQuery Test
```
ronykoz: not needed

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: Specifies the action that occurs if destination table already exists.
+    outputs:
+    - contextPath: BigQuery.Query
+      description: The query performed.
+      type: String
+    - contextPath: BigQuery.Row
+      description: The table rows the given query returned.
+      type: Unknown
+    description: Perform a query on BigQuery
+  dockerimage: bigquery_image_local_3
```
ronykoz: update to a new docker image that you will create

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+script:
+  script: ''
+  type: python
+  commands:
+  - name: bigquery-query
+    arguments:
+    - name: query
+      required: true
+      description: A query string, following the BigQuery query syntax, of the query
+        to execute.
```
ronykoz: add a link for further explanation if you can find one

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+    - name: location
+      description: The geographic location where the job should run. Required except
+        for US and EU.
+    - name: allow_large_results
```
ronykoz: set a default value of False

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      auto: PREDEFINED
+      predefined:
+      - "True"
+      - "False"
+      description: Allow large query results tables (legacy SQL, only)
+    - name: default_dataset
+      description: A stringof the fully-qualified dataset ID in standard SQL format.
```
ronykoz:
```suggestion
      description: A string of the fully-qualified dataset ID in standard SQL format.
```

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+        The value must included a project ID and dataset ID separated by dots.
```
ronykoz:
```suggestion
        The value must include a project ID and dataset ID separated by dots.
```

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+    - name: destination_table
```
ronykoz: set the None default value

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: The table where results are written. Default value is None if not
+        set.
+    - name: kms_key_name
```
ronykoz: is that the name they are giving it?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: Custom encryption configuration for the destination table.
+    - name: dry_run
```
ronykoz: set the default value to false

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: If set to true, BigQuery doesn't run the job. Instead, if the query
+        is valid, BigQuery returns statistics about the job such as how many bytes
+        would be processed. If the query is invalid, an error returns. The default
+        value is false.
+    - name: priority
+      auto: PREDEFINED
+      predefined:
+      - BATCH
+      - INTERACTIVE
+      description: Priority of the query. Possible values include INTERACTIVE and
```
ronykoz: add an explanation for each of them

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+        BATCH. The default value is INTERACTIVE.
+    - name: use_query_cache
```
ronykoz: default value

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      auto: PREDEFINED
+      predefined:
+      - "True"
+      - "False"
+      description: Look for the query result in the cache.
+    - name: use_legacy_sql
+      auto: PREDEFINED
+      predefined:
```
ronykoz: default value

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      - "True"
+      - "False"
+      description: Use legacy SQL syntax.
+    - name: job_id
```
ronykoz: is it required or we can just set the ID?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      description: The ID of the job. The ID must contain only letters (a-z, A-Z),
+        numbers (0-9), underscores (_), or dashes (-). The maximum length is 1,024
+        characters.
+    - name: write_disposition
```
ronykoz: shouldn't it be required?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
```diff
@@ -0,0 +1,96 @@
...
+      - WRITE_TRUNCATE
+      - WRITE_APPEND
+      description: Specifies the action that occurs if destination table already exists.
+    outputs:
+    - contextPath: BigQuery.Query
+      description: The query performed.
+      type: String
+    - contextPath: BigQuery.Row
+      description: The table rows the given query returned.
+      type: Unknown
```
ronykoz: String?

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+        query_results = query_job.result()
+        return query_results
+    # if dry run is activated, the results (number of bytes the query will process) are returned in the job itself
+    else:
+        return query_job
+
+
+def query_command():
+    args = demisto.args()
+    query_results = query(args['query'], args.get('project_id', None), args.get('location', None), args.get('allow_large_results', None),
```
ronykoz: If there are default values this will be readable + i prefer you split the `args.get('project_id', None)` to not be in the function call but rather for you to assign params for them beforehand and send the params instead.
And then you can do `query(query=query, project_id=project_id, location=location` or maybe just `query(query, project_id, location`

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+                          args.get('priority', None), args.get('use_query_cache', None), args.get('use_legacy_sql', None),
+                          demisto.params()['google_service_creds'], args.get('job_id', None), args.get('write_disposition', None))
+
+    context = {}
+    contents = []
+    rows_contexts = []
+    human_readable = 'No results found.'
+    dry_run = args.get('dry_run', None)
+    if dry_run and str_to_bool(dry_run):
+        human_readable = '### Dry run results: \n This query will process {0} bytes'.format(query_results.total_bytes_processed)
```
ronykoz: this is all the data for dry run?

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+            contents = query_results
+            context['BigQuery(val.Query && val.Query == obj.Query)'] = {
+                'Query': args['query'],
+                'Row': rows_contexts
+            }
+
+            title = 'BigQuery Query Results'
+            human_readable = tableToMarkdown(title, rows_contexts, removeNull=True)
+
+    demisto.results({
```
ronykoz: use return_outputs

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+    if dry_run and not represents_bool(dry_run):
+        return_error("Error: dry_run must have a boolean value.")
+    if priority and not (priority == 'INTERACTIVE' or priority == 'BATCH'):
+        return_error("Error: priority must have a value of INTERACTIVE or BATCH.")
+    if write_disposition and not (write_disposition == 'WRITE_TRUNCATE' or write_disposition == 'WRITE_APPEND' or write_disposition == 'WRITE_EMPTY'):
+        return_error("Error: write_disposition must have a value of WRITE_TRUNCATE, WRITE_APPEND or WRITE_EMPTY.")
+
+
+def build_query_job_config(allow_large_results, default_dataset_string, destination_table, dry_run, priority, use_query_cache, use_legacy_sql, kms_key_name, write_disposition):
+    validate_args_for_query_job_config(allow_large_results, priority, use_query_cache, use_legacy_sql, dry_run, destination_table, write_disposition)  # send relevant params
```
ronykoz: remove the comment

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+    query_job_config = bigquery.QueryJobConfig()
+    if allow_large_results:
+        query_job_config.allow_large_results = str_to_bool(allow_large_results)
+    if default_dataset_string:
+        query_job_config.default_dataset = default_dataset_string
+    if destination_table:
+        query_job_config.destination = destination_table # must be in the format your-project.your_dataset.your_table
```
ronykoz: add the comment to the description for the arg itself

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+    if kms_key_name:
+        query_job_config.destination_encryption_configuration = bigquery.table.EncryptionConfiguration(kms_key_name) # make sure this shouldn't be just kms_key_name
```
ronykoz: Couldn't understand the comment

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+    return query_job_config
+
+
+''' COMMANDS + REQUESTS FUNCTIONS '''
+
+def query(query_string, project_id, location, allow_large_results, default_dataset, destination, kms_key_name, dry_run, priority, use_query_cache, use_legacy_sql,
+          google_service_creds, job_id, write_disposition):
+    bigquery_client = start_and_return_bigquery_client(google_service_creds)
+    job_config = build_query_job_config(allow_large_results, default_dataset, destination, dry_run, priority, use_query_cache, use_legacy_sql, kms_key_name, write_disposition)
+    # not sure if query should be a keyword arg
```
ronykoz: why not? + remove the comment

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+    query_job = bigquery_client.query(query = query_string, job_config = job_config, location = location, job_id = job_id, project = project_id)
+    if not (dry_run and str_to_bool(dry_run)):
+        query_results = query_job.result()
+        return query_results
+    # if dry run is activated, the results (number of bytes the query will process) are returned in the job itself
```
ronykoz: move this comment to be in the else section

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+import json
+import requests
+import os
+from distutils.util import strtobool
+from google.cloud import bigquery
+from oauth2client import service_account
+
+# Disable insecure warnings
+requests.packages.urllib3.disable_warnings()
+'
```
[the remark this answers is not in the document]
guykeller: The whole disable warnings section or just the comment explaining it?

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+
+''' GLOBALS/PARAMS '''
+
+TEST_QUERY = ('SELECT name FROM `bigquery-public-data.usa_names.usa_1910_2013` '
+          'WHERE state = "TX" '
+          'LIMIT 100')
```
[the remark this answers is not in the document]
guykeller: which part? It's an SQL-ish syntax (like the answer [here](https://stackoverflow.com/questions/5243596/python-sql-query-string-formatting) ).
Could you explain what should be improved?

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
+import demistomock as demisto
+from CommonServerPython import *
+from CommonServerUserPython import *
+''' IMPORTS '''
+
+import json
+import requests
+import os
+from distutils.util import strtobool
+from google.cloud import bigquery
```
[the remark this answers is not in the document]
guykeller: Yes, that's right - further data [here](https://pypi.org/project/google-cloud-bigquery/). Hope I understood your question correctly and this answers it

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: Since when do we move this to a separate file? How should it be called and what should it contain?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: Why not? And if not, why are tests ever specified that way?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: Why? that feels counter intuitive. What if suddenly BigQuery decide that the default value is True?
I think the integration should behave as similar to the product as possible - meaning, if you don't specify some arg, you get the default value from BigQuery for that arg (Whatever that is).
What do you think?

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: Same as above with allow_large_results

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: as above

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: as above

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: as above

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: It is not required

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: No, you can certainly not specify it. On the BigQuery documentation it says the default value is WRITE_APPEND, meaning to add new data to the existing table if one exists, but from what I saw the actual default value is WRITE_EMPTY, meaning an error is thrown if the table already exists.
This is yet another reason I think not specifying a default value would be better here

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: It's an array of dictionaries, it isn't a string

Integrations/GoogleBigQuery/GoogleBigQuery.py
guykeller: I agree, but Anar prefers it to be done this way (sending the value straight away instead of naming it first).
I'd love to change the concept but I think we need consistency through integrations, whatever the consistent styling decided is.
What do you think?

Integrations/GoogleBigQuery/GoogleBigQuery.py
guykeller: Somewhere I saw another field returned, but:
A. I can't seem to find that documentation after searching for it. Only this data is found.
B. Every example, including in the Google documentation, only uses this data as this is the significant data which will allow you to estimate the cost of the query. This is what users expect to see, which is the key factor in human readables to my understanding.

Integrations/GoogleBigQuery/GoogleBigQuery.yml
guykeller: No, but it is more similar to the UI.
In the API this is called destination_encryption_configuration, but that is completely unreadable and also non descriptive.
In the UI, if you want to specify a custom encryption you need to choose one of your KMS keys to use.
So in this case I went with what's readable.
I can change it to "encryption_kms_key" or "custom_encryption_kms_key". Would that be better?

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -0,0 +1,192 @@
...
+
+def bool_arg_set_to_true(arg):
+    return arg and str_to_bool(arg)
+
+
+def start_and_return_bigquery_client(google_service_creds_json_string):
+    cur_directory_path = os.getcwd() # maybe better to take the root directory using os.path.abspath(os.sep)? currently it's just the same as not specifying a path at all
+    creds_file_name = '{0}.json'.format(demisto.uniqueFile())
+    path_to_save_creds_file = os.path.join(cur_directory_path, creds_file_name)
+    creds_file = open(path_to_save_creds_file, "w")
```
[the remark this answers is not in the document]
guykeller: Changed, and would love to hear why you think that is more readable or better

PR #3423, opened 2019-04-24
Whois Integration

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#16414

## Description
Moved package inside integration to allow more rapid deployment of regex changes.

   - No

## Must have
- [x] Tests
- [x] Documentation (with link to it) https://github.com/demisto/etc/issues/17130
- [x] Code Review


amshamah419:
I originally was going to make this a V2, but there was a decision made to keep the current integration and improve it.

Integrations/Whois/Whois.py
```diff
@@ -0,0 +1,8289 @@
...
+    },
+    "hotmail": {
+        "_type": "newgtld",
+        "adapter": "none"
+    },
+    "house": {
+        "_group": "donuts",
+        "_type": "newgtld",
+        "host": "whois.nic.house"
+
```
yuvalbenshalom: get the domain argument here

yuvalbenshalom: catch value error when using `strftime`
this is relevant also to the following lines.

yuvalbenshalom: why do you expect OSError here?

yuvalbenshalom: consider moving the code down after the `grammar` so you will not have a short code "hiding" between 2 huge dictionary initialisation

Integrations/Whois/Whois.yml
```diff
@@ -191,6 +68,84 @@ script:
     - contextPath: Whois.Raw
       description: Raw Output
       type: string
+    - contextPath: Whois.Administrator.country
```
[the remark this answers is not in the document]
amshamah419: I was originally dumping the results into the context for the Admin, Tech Admin, Registrant, and Billing keys. I changed it to use the createContext keyTransformer to capitalise these keys.

glicht:
How is this related to #3567 ? which one are we merging in the end of day?

amshamah419:
@glicht This is the pull request that will be in master at the end of the day. We had to split it into two pull requests since the original was a single page integration. After `whois-pkg` is merged, then this one will be merged so we can track the changes.

PR #3424, opened 2019-04-24
Wildfire - add wildfire-get-sample command

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16688

## Description
add wildfire-get-sample command

   - No

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/16700
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Wildfire.yml
+++ b/Integrations/integration-Wildfire.yml
@@ -1060,7 +1060,7 @@ script:
       description: MD5 hash to retrieve
       isArray: true
     - name: sha256
-      description: SHA256 hash to retreive
+      description: SHA256 hash to retrieve
       isArray: true
     outputs:
     - contextPath: File.Size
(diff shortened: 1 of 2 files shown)


Integrations/integration-Wildfire.yml
```diff
@@ -1017,7 +1054,48 @@ script:
...
+      type: string
+    - contextPath: File MD5
+      description: MD5 hash of the file
+      type: string
+    - contextPath: File.Extension
+      description: File Extension
+      type: string
+    description: Retrieve a sample from WildFire
   - name: wildfire-upload-file-remote
+    deprecated: true
```
yuvalbenshalom: should not be deprecated before the new integration is in

bakatzir: we already have the wildfire-upload-file-url, so should be deprecated I believe.

Clairvoyantx:
Screen Shot 2019-04-24 at 10 37 02 AM
The get-sample command returns an error on not finding a hash.

bakatzir:
@clairvoyantx 
I'll fix that on sunday morning.
Note that non malicious hashes are available for 14 days only.

PR #3433, opened 2019-04-28
intezer - stabilze test

bakatzir:
## Status
Ready

   - No


bakatzir:
@DeanArbel yes, i wait inline in previous task for enrichment
just ant context to be updated

Nitzperetz:
@bakatzir , best practice would be to use command sleep/retries arguments and not adding a sleep task.

bakatzir:
@Nitzperetz you are right.
But I use the extract indicators inline, and not a command.
Screen Shot 2019-04-28 at 11 28 22

PR #3437, opened 2019-04-28
Read pdf file fix

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16442

## Description
Added error for failing to read image file.

## Screenshots
image
   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Scripts/script-ReadPDFFile.yml
+++ b/Scripts/script-ReadPDFFile.yml
@@ -7,6 +7,7 @@ script: |-
   import pdfx
   from pdfminer.psparser import PSEOF
   from pdfminer.pdfdocument import PDFEncryptionError
+  import traceback
   import sys
   reload(sys)
   sys.setdefaultencoding("utf-8")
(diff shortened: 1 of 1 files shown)


Scripts/script-ReadPDFFile.yml
```diff
@@ -58,6 +58,8 @@ script: |-
           except PDFEncryptionError:
               mark_suspicious('Possibly encrypted file')
               sys.exit(0)
+          except TypeError:
```
ronykoz: Can we try to add some other identification as well? this seems a too broad exception type for this error.
And maybe we should mark it as suspicious as we are doing for the other exceptions

Scripts/script-ReadPDFFile.yml
```diff
@@ -58,6 +59,13 @@ script: |-
...
+              tb = traceback.format_exc()
+              if 'pdf = pdfx.PDFx(args.pdf)' in tb and "out_str = in_str.decode(enc['encoding'])" in tb:
+                  mark_suspicious("The script failed to read the PDF file. This might happen if your file is an image")
```
ronykoz:
```suggestion
                  mark_suspicious("The script failed to read the PDF file. This might happen if your PDF file contains onlu an image")
```

PR #3450, opened 2019-04-29
fix for when compare url is empty

glicht:
## Status
Ready

## Description
Fixes case that compare url from circle ci is empty. This happens when doing a manual re-build.

See following build: https://circleci.com/gh/demisto/content/22188


glicht:
Merged as part of: https://github.com/demisto/content/pull/3451

PR #3469, opened 2019-04-30
fix mail sender exception handling

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16750

## Description
fixed the usage of return_error

looks like this now:
image


liorblob:
@ronykoz Is this synced with https://github.com/demisto/content/pull/3403?

ronykoz:
@liorblob tested the changes on content version 19.4.1 as discussed, figuered out that the problem is due to the local return_error function so i've just replaced it everywhere in the yml(so Dean's changes are also there)

PR #3477, opened 2019-05-01
Fixed BigQuery test module

guykeller:
## Status
Ready

## Description
The test module was unstable, depended on results from a BigQuery public DB.
Should be stable now and still checks all parameters are valid.

   - No

## Must have
- [ ] Code Review


Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -173,9 +173,7 @@ def test_module():
         query_job = bigquery_client.query(TEST_QUERY)
         query_results = query_job.result()
         results_rows_iterator = iter(query_results)
-        first_item = next(results_rows_iterator)
-        if str(first_item.get("name")) != "Ruby":
-            raise ValueError("Data from DB not matching expected results")
+        next(results_rows_iterator)
```
ronykoz: try to catch an error here and post a good error msg if an error occurs

guykeller: There shouldn't be an error here, why would there be? It's a huge DB which consistently returns results

PR #3479, opened 2019-05-02
Change time to wait for Wildfire analysis before pulling report

DctrG:
## Status
Ready

## Description
Sleep time before trying to fetch Wildfire report reduced from 15 to 6 minutes as the report should be available within that time.

## Must have
- [X] Tests
- [X] Code Review


bakatzir:
@DctrG  Hi, nice to get a PR from you.
we're deprecating this script, and using the `GenericPooling` mechanism we've implemented inside a designed playbook:
https://github.com/demisto/content/blob/master/Playbooks/playbook-Detonate_File_-_WildFire.yml
this way, we don't just guess when the submission will complete its analysis, and poll it's status every 60 seconds(configurable)

PR #3482, opened 2019-05-02
Return human readable entry when there is no attachment

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16783

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/Gmail/Gmail.yml
+++ b/Integrations/Gmail/Gmail.yml
@@ -833,7 +833,7 @@ script:
   dockerimage: demisto/google-api:1.0
   isfetch: true
   runonce: false
-releaseNotes: Added human readable in case there is no attachments to email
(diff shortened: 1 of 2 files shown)


yaakovi:
@kirbles19 
Please review the releaseNotes here

Integrations/Gmail/Gmail.yml
```diff
@@ -833,7 +833,7 @@ script:
   dockerimage: demisto/google-api:1.0
   isfetch: true
   runonce: false
-releaseNotes: added gmail-list-filters and gmail-remove-filter commands
+releaseNotes: Added human readable in case there is no attachments to email
```
kirbles19: Added human readable output for cases when an email does not contain an attachment.

PR #3486, opened 2019-05-02
Cherwell new integration

ShahafBenYakir:
## Status
Ready

## Description- Fetch incidents - invalid character 'F' (probably returned error with first letter 'F') - add simulation to fetch when testing
- Integration description replace ':' with ','
- Scripts - add 'dependsOn','Tags'
- Add to test flow bad time range string
- Add to advanced query parameter instruction to go to detailed instructions of the integration.

Fixes https://github.com/demisto/etc/issues/16864
## Must have
- [x] Tests
- [x] Documentation https://github.com/demisto/etc/issues/16735
- [ ] Code Review


Scripts/CherwellCreateIncident/CherwellCreateIncident.yml
```diff
@@ -55,9 +55,14 @@ outputs:
   type: String
 script: '-'
 system: false
+tags:
+- Cherwell
```
yuvalbenshalom: I think this should be an array

ShahafBenYakir: - [x] Check

PR #3493, opened 2019-05-03
Carbon black protection v2 enhancement

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16776

## Description
Add explanation about Fetch query
in detailed description
Add to main page (Click ? for more information)
If fetch incidents is turned on - check if fetch incidents is successful
Validate User custom fetch query
Add check to fetched time.
Change order of human-readable fields for the command cbp-fileCatalog-search Change order, so that Name will be first.
Add to descriptions the commands that are used in order to get the required param.
Add specific query args for search commands (e.g. hash)
Change (int32) to int in descriptions.

   - No

## Must have
- [X] Tests
- [X] Documentation (https://github.com/demisto/etc/issues/16711)
- [ ] Code Review


Integrations/CarbonBlackProtect/CarbonBlackProtect.yml
```diff
@@ -132,8 +193,8 @@ script:
         is any valid field in the object that is being queried. Operator (: LIKE,
         ! NOT LIKE, < Less than, > Greater than, + logical AND, - logical OR, | separating
         values) is any of valid operators (see below). All operators consist of a
-        single character. Value is compared with an operator and depends on field type.
-        See more: https://developer.carbonblack.com/reference/enterprise-protection/8.0/rest-api/#searching'
+        single character. Value is compared with an operator and depends on field
+        type. See more: https://developer.carbonblack.com/reference/enterprise-protection/8.0/rest-api/#searching'
       isArray: false
       name: query
       required: false
```
yuvalbenshalom: line 203 still has "int23" in it

Integrations/CarbonBlackProtect/CarbonBlackProtect.py
```diff
@@ -506,38 +701,138 @@ def search_event_command():
...
+    """
+    severity_dict = {
+        'Critical': 2,
+        'Error': 3,
+        'Warning': 4,
+        'Notice': 5,
+        'Info': 6,
+        'Debug': 7
+    }
+    return severity_dict.get(severity, severity)
```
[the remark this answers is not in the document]
DeanArbel: Yes, this is both a workaround and a fail safe - I noticed that in some cases there were values outside the ones documented by the API (e.g. in one case the API described values of 20,40,60,80 - however in our instance I encountered a 35 value - which doesn't has a string value to describe it.

This way if the function can receive a number value, and support those cases as well (though they're supposed to be מקרי קצה)

Integrations/CarbonBlackProtect/CarbonBlackProtect.yml
DeanArbel: Weird I was sure I removed it... good catch

PR #3497, opened 2019-05-04
proxy fix AbuseIPDB

orenzohar:
## Status
In Progress

## Description
fixes proxy handling In order to enable mocking

   - No

## Must have
- [x] Tests
- [ ] Code Review


anara123:
@orenzohar the test still fails

orenzohar:
@anara123 yes it looks like it doesn't support it.

anara123:
requests package definitely support verify and proxy setting. need to figure out why the test fails. you dont use here SDK or some other http library

PR #3500, opened 2019-05-05
Fixed Typo

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16840

   - No

### DIFF
--- a/Integrations/Active_Directory_Query/Active_Directory_Query.py
+++ b/Integrations/Active_Directory_Query/Active_Directory_Query.py
@@ -90,6 +90,7 @@ def account_entry(person_object, custome_attributes):
         'Email': person_object.get('email'),
         'Username': person_object.get('samAccountName'),
         'DisplayName': person_object.get('displayName'),
+        'Managr': person_object.get('manager'),
         'Manager': person_object.get('manager'),
         'Groups': person_object.get('memberOf')
     }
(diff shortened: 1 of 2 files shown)


Integrations/Active_Directory_Query/Active_Directory_Query.py
```diff
@@ -90,7 +90,7 @@ def account_entry(person_object, custome_attributes):
         'Email': person_object.get('email'),
         'Username': person_object.get('samAccountName'),
         'DisplayName': person_object.get('displayName'),
-        'Managr': person_object.get('manager'),
+        'Manager': person_object.get('manager'),
```
ronykoz: we need to maintain the old one as well, so keep the old version and also add the new one.
Also please verify the outputs

Integrations/Active_Directory_Query/Active_Directory_Query.yml
```diff
@@ -388,5 +388,6 @@ script:
       specified group
   runonce: false
   dockerimage: demisto/ldap:1.0.0.75
+releaseNotes: " - "
```
yaakovi: remove the spaces from the release notes

Integrations/Active_Directory_Query/Active_Directory_Query.py
ronykoz: @Shellyber what about this?

PR #3504, opened 2019-05-06
Rasterize url error handeling

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16238

   - No


Integrations/rasterize/rasterize.py
```diff
@@ -0,0 +1,135 @@
...
+        error_message = e.message
+
+
+if demisto.command() == 'test-module':
+    rasterize_email('test text', 'email.png')
+    if return_code == 0:
+        demisto.results('ok')
+    else:
+        demisto.results(error_message)
+    sys.exit(0)
```
Itay4: remove this one as well

PR #3507, opened 2019-05-06
WF - RN and file arg in file command

bakatzir:
## Status
Ready

master failed due to missing RN because of last release - fixed.
file argument in the file command should not be deprecated. fixed as well.

   - No


yuvalbenshalom:
@glicht @anara123, this should fix the build, but it fail because of McAfee ATD fail that is unrelated.
So it needs force merge if we want to fix it now.

PR #3509, opened 2019-05-07
disabled duoadmin from mocking

glicht:
## Status
Ready

## Description
Until we fix: https://github.com/demisto/etc/issues/16893 . Disabling from mock.


glicht:
Force merge as failure is unrelated

PR #3515, opened 2019-05-09
move ATD to skip until we get license

bakatzir:
## Status
Ready

reference: https://github.com/demisto/etc/issues/16909


Itay4:
why under `will not be resolved` ?

PR #3518, opened 2019-05-11
Fixed - Google BigQuery date handling bug

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16925

## Description
Fixed the way Google BigQuery handles results from query in datetime format.
No they are returned properly and no error is returned.
Also added a task to the playbook that tests the command that used to return an error.

   - No

## Must have
- [x] Tests
- [ ] Code Review


Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -173,12 +190,12 @@ def test_module():
         query_job = bigquery_client.query(TEST_QUERY)
         query_results = query_job.result()
         results_rows_iterator = iter(query_results)
-        next(results_rows_iterator)
+        first_item = next(results_rows_iterator)
+        if str(first_item.get("name")) != "Ruby":
```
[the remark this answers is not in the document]
guykeller: Added by mistake from a previous version, will remove

guykeller:
As for unit tests - There really isn't anything to unit test, everything is included in the test playbook (including most of the possible arguments) and there isn't anything purely pythonic to test. What do you think?
As for the bug fixing process - will do

Integrations/GoogleBigQuery/GoogleBigQuery.py
```diff
@@ -176,9 +184,7 @@ def test_module():
         next(results_rows_iterator)
         demisto.results("ok")
     except Exception as ex:
-        return_error("There was a problem creating a BigQuery client.\n"
```
Itay4: please post screenshot of an error returned for example

Integrations/GoogleBigQuery/google_bigquery_test.py
```diff
@@ -0,0 +1,9 @@
+import datetime
+from GoogleBigQuery import convert_to_string_if_datetime
+
+
+def test_convert_to_string_if_datetime():
+    test1 = convert_to_string_if_datetime(None)
```
Itay4: name args in meaningful way
also add test for empty string

Integrations/GoogleBigQuery/GoogleBigQuery.py
guykeller: Added a more clear message

PR #3523, opened 2019-05-13
Build fix

ronykoz:
moving integration to skip + changing slack msg


ronykoz:
Don't merge yet, i want the new report to get out so that i'll be able to move more tests

ronykoz:
fixes https://github.com/demisto/etc/issues/16931

PR #3524, opened 2019-05-13
fix context output for AD query 2

noaco:
fixes https://github.com/demisto/etc/issues/16843


noaco:
@anara123 please review

PR #3533, opened 2019-05-13
Forescout integration

avidan-H:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15198

## Description
New Forescout 8.1 integration

   - Don't think so

## Must have
- [x] Tests
- [ ] Code Review

## Dependencies
- [x] ~~Dependency 1~~ N/A


Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+            verify=USE_SSL,
+            params=params,
+            data=data,
+            files=files,
+            headers=headers,
+            auth=auth
+        )
+
+        # Handle error responses gracefully
+        if res.status_code not in {200, 201, 304}:
```
Itay4: whats 201 and 304 ?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+                    resp_xml = ET_PHONE_HOME.fromstring(res.content)
+                    demisto.info(str(res.content))
+                    codes = [child.text for child in resp_xml.iter() if child.tag == 'CODE']
+                    messages = [child.text for child in resp_xml.iter() if child.tag == 'MESSAGE']
+                    err_msg += ''.join([f'\n{code}: {msg}' for code, msg in zip(codes, messages)])
+                return_error(err_msg)
+
+        resp_type = resp_type.casefold()
+        if resp_type == 'json':
+            return res.json()
```
Itay4: need to wrap this with try-except as well

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+        elif resp_type == 'text':
+            return res.text
+        elif resp_type == 'content':
+            return res.content
+        else:
+            return res
+
+    except requests.exceptions.ConnectionError:
```
Itay4: lets discuss this

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+
+
+''' COMMANDS + REQUESTS FUNCTIONS '''
+
+
+def test_module():
+    """
+    Performs basic get request that requires proper authentication
+    """
+    login()
```
Itay4: i prefer we do some kind of get query here

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    elif id_type == 'mac':
+        # API endpoint format - https://{EM.IP}/api/hosts/mac/{mac}?fields={prop},..,{prop_n}
+        url_suffix += 'mac/'
+    # if id_type == 'id' don't change url_suffix -it's already in desired format as shown below
+    # API endpoint format - https://{EM.IP}/api/hosts/{obj_ID}?fields={prop},..,{prop_n}
+
+    url_suffix += '='.join(ident)
+    params = {'fields': fields} if fields != '' else None
+    headers = create_web_api_headers()
+    response = http_request('GET', url_suffix, headers=headers, params=params, resp_type='response')
```
Itay4: why are we not asking for `json` here?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    'Content-Type': 'application/xml',
+    'Accept': 'application/xml'
+}
+DEX_AUTH = (DEX_USERNAME + '@' + DEX_ACCOUNT, DEX_PASSWORD)
+WEB_AUTH = ''
+LAST_JWT_FETCH = None
+# Default JWT validity time set in Forescout Web API
+JWT_VALIDITY_TIME = timedelta(minutes=5)
+# Host fields to be included in output of get_host_command
+HOSTFIELDS_TO_INCLUDE = {
```
Itay4: make this `defaultdict` please

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+
+    if requested_fields and set(requested_fields).issubset(HOSTFIELDS_TO_INCLUDE.keys()) or not requested_fields:
+        included_fields = {
+            HOSTFIELDS_TO_INCLUDE.get(key): val for key, val in fields.items() if key in HOSTFIELDS_TO_INCLUDE.keys()
+        }
+        included_fields_readable = {
+            HOSTFIELDS_TO_INCLUDE.get(key): dict_to_formatted_string(val)
+            for key, val in fields.items() if key in HOSTFIELDS_TO_INCLUDE.keys()
+        }
+    elif requested_fields:
```
Itay4: you already check `requested_fields` in the condition

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    params: Dict = {}
+    rule_ids = args.get('rule_ids')
+    properties = args.get('properties')
+    if rule_ids and properties:
+        url_suffix += '?matchRuleId=' + rule_ids + '&' + properties
+    elif rule_ids:
+        url_suffix += '?matchRuleId=' + rule_ids
+    elif properties:
+        url_suffix += '?' + properties
+    response = http_request('GET', url_suffix, headers=headers, params=params, resp_type='response')
```
Itay4: why are we not asking for `json` here?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    context = {'Forescout.Host(val.ID && val.ID === obj.ID)': content}
+    title = 'Active Endpoints'
+    human_readable = tableToMarkdown(title, content, removeNull=True)
+    return_outputs(readable_output=human_readable, outputs=context, raw_response=response)
+
+
+def get_hostfields():
+    url_suffix = '/api/hostfields'
+    headers = create_web_api_headers()
+    response = http_request('GET', url_suffix, headers=headers, resp_type='response')
```
Itay4: why are we not asking for `json` here?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+def get_hostfields_command():
+    args = demisto.args()
+    response = get_hostfields()
+    data = response.json()
+    filtered_data = filter_hostfields_data(args, data)
+    if not filtered_data:
+        demisto.results('No hostfields matched the specified filters.')
+    else:
+        content = [{key.title(): val for key, val in x.items()} for x in filtered_data]
+        context = {'Forescout.HostField': content}
```
Itay4: no unique field here ?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+        title = 'Index of Host Properties'
+        human_readable = tableToMarkdown(title, content, removeNull=True)
+        return_outputs(readable_output=human_readable, outputs=context, raw_response=data)
+
+
+def get_policies():
+    url_suffix = '/api/policies'
+    headers = create_web_api_headers()
+    response = http_request('GET', url_suffix, headers=headers, resp_type='response')
```
Itay4: why are we not asking for json here?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    url_suffix = '/fsapi/niCore/Lists'
+    response = http_request('POST', url_suffix, headers=DEX_HEADERS, auth=DEX_AUTH, data=data, resp_type='response')
+    return response
+
+
+def update_lists_command():
+    args = demisto.args()
+    response = update_lists(args)
+    resp_xml = ET_PHONE_HOME.fromstring(response.content)
+    code = [child.text for child in resp_xml.iter() if child.tag == 'CODE'][0]
```
Itay4: are we sure this list wont be empty?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    msg = [child.text for child in resp_xml.iter() if child.tag == 'MESSAGE'][0]
```
Itay4: are we sure this list wont be empty?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+def update_host_properties_command():
+    args = demisto.args()
+    update_type = args.get('update_type', '')
+    properties = args.get('properties', '')
+    response = update_host_properties(args)
+
+    # Because the API has an error and says it deletes multiple things when it only deletes one
+    # have to take care of it behind the curtains
+    if update_type == 'delete':
+        args_copy = deepcopy(args)
```
Itay4: why do we need a copy?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+        args_copy['properties'] = ''
+        update_host_properties(args_copy)  # Takes care of composite_property
+        args_copy['composite_property'] = ''
+        for prop in properties.split('&'):
+            args_copy['properties'] = prop
+            update_host_properties(args_copy)
+
+    resp_xml = ET_PHONE_HOME.fromstring(response.content)
+    code = [child.text for child in resp_xml.iter() if child.tag == 'CODE'][0]
```
Itay4: are we sure this list wont be empty?

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+    msg = [child.text for child in resp_xml.iter() if child.tag == 'MESSAGE'][0]
```
Itay4: are we sure this list wont be empty?

Integrations/Forescout/Forescout.yml
```diff
@@ -0,0 +1,401 @@
...
+  type: 9
+- display: Data Exchange (DEX) Username
+  name: dex_credentials
+  required: false
+  type: 9
+- display: The Data Exchange (DEX) Account
+  name: dex_account
+  required: true
+  type: 0
+- defaultvalue: 'true'
```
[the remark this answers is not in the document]
avidan-H: ? It's not, I think that is for the parameter after `dex_account`

Integrations/Forescout/Forescout.yml
```diff
@@ -0,0 +1,401 @@
...
+      isArray: true
+      name: fields
+      required: false
+      secret: false
+    deprecated: false
+    description: Retrieve specified host property values for an endpoint.
+    execution: false
+    name: forescout-get-host
+    outputs:
+    - contextPath: Forescout.Host.Field.MatchedFingerprint
```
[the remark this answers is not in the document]
avidan-H: What's the best approach then here?

Integrations/Forescout/Forescout.yml
```diff
@@ -0,0 +1,401 @@
...
+      type: Unknown
```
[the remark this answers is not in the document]
avidan-H: The ones that are left as `Unknown` are because they are either lists or dicts

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
+import demistomock as demisto
+from CommonServerPython import *
+from CommonServerUserPython import *
+''' IMPORTS '''
+
+import json
+import requests
+from typing import Dict, List, Tuple, Any, Union
+import xml.etree.ElementTree as ET_PHONE_HOME
```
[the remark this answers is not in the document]
avidan-H: I would have except that it is a conflict with an import in commonserverpython so I went for (the way more fun) `ET_PHONE_HOME`

Integrations/Forescout/Forescout.py
```diff
@@ -0,0 +1,678 @@
...
+
+PARAMS = demisto.params()
+
+CREDENTIALS = PARAMS.get('credentials', {})
+USERNAME = CREDENTIALS.get('identifier', '')
+PASSWORD = CREDENTIALS.get('password', '')
+
+DEX_CREDENTIALS = PARAMS.get('dex_credentials', {})
+DEX_USERNAME = DEX_CREDENTIALS.get('identifier')
+DEX_USERNAME = DEX_USERNAME if DEX_USERNAME != '' else USERNAME
```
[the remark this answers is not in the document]
avidan-H: In the case that the user created the same Username and Password for both the `Web API` and for `Data Exchange (DEX)` as I did, then instead of having to enter both - will just use the same credentials entered for `Web API`

Integrations/Forescout/Forescout.py
avidan-H: > If the queried data has not changed in the Forescout platform since the
last request, clients receive an HTTP 304 (Not Modified) return message.

201 is leftover from previous integration I copied the basic version of `http_request` from and I just left it. I can remove it if you'd like.

Integrations/Forescout/Forescout.py
avidan-H: yea it's problematic for another reason which is that it only is testing the `Web API` essentially where in reality should be performing some kind of request to `DEX` as well - the problem with that is all the `DEX` API calls are to update stuff, not just retrieve some data...

Integrations/Forescout/Forescout.py
avidan-H: my mistake. It was left over when I initially was using a function decorator on all of the functions that made `Web API` calls to work with the entity tags they described in the Forescout Web API documentation before I realized it was unnecessary. I'll change it for all of those calls.

Integrations/Forescout/Forescout.py
avidan-H: yes but in the previous condition, the expression only evaluates to `True` if there are `requested_fields` _and_ if `requested_fields` are a subset of `HOSTFIELDS_TO_INCLUDE` keys.

Integrations/Forescout/Forescout.py
avidan-H: there are over 900 potential fields that can be returned...

Integrations/Forescout/Forescout.py
avidan-H: otherwise when I manipulate members of the dict, e.g. when I alter the appearance to look nice in the war room readable output, it will also be changed in the context outputs.

avidan-H:
@kirbles19 please review:
https://github.com/demisto/content/blob/forescout-integration/Integrations/Forescout/Forescout.yml
https://github.com/demisto/content/blob/forescout-integration/Integrations/Forescout/Forescout_description.md

avidan-H:
@kirbles19 please take a look at the language of the error messages in the integration
https://github.com/demisto/content/blob/forescout-integration/Integrations/Forescout/Forescout.py

kirbles19:
All done with .yml and .md files.

PR #3547, opened 2019-05-15
Fixed the error shown in case of failed authentication

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16970

## Description
Running the test with bad credentials used to result in an exception.
Now it returns a clear error message. Also works in case of several errors.

## Screenshots
Screen Shot 2019-05-15 at 12 18 20

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Beta_Integrations/integration-ThreatQ.yml
+++ b/Beta_Integrations/integration-ThreatQ.yml
@@ -73,23 +73,51 @@ script:
     def get_errors_string_from_bad_request(bad_request_results):
         errors_list = bad_request_results.json().get("errors", [])
         errors_string = ""
-        for error in errors_list:
-            errors_string += "Error: {}\n".format(error)
+        error_num = 1
+        if errors_list:
+            errors_string = "Errors from server: \n"
+            for error in errors_list:
+                errors_string += "Error #{0}: {1}\n".format(error_num, error)
+                error_num += 1
         return errors_string
 
     # ThreatQ auth based on OAuth 2.0 credential grand method
     def tq_access():
         data = {'grant_type': 'password','email': EMAIL, 'password': PASSWORD, 'client_id': CLIENT_ID}
         access_token_response = requests.post(API_TOKEN_URL, data=data, verify=False, allow_redirects=False)
+
         tokens = json.loads(access_token_response.text)
         if int(access_token_response.status_code) >= 400:
             errors_string = get_errors_string_from_bad_request(access_token_response)
-            error_message = "Authentication failed, unable to retrieve an access token.\n Errors from server:\n {}".format(errors_string)
+            error_message = "Authentication failed, unable to retrieve an access token.\n {}".format(errors_string)
             return_error(error_message)
+
+        new_integration_context = {
+            "access_token": tokens['access_token'],
+            "access_token_creation_time": int(time.time()) -1, # decrementing one second to be on the safe side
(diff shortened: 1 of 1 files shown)


Beta_Integrations/integration-ThreatQ.yml
```diff
@@ -70,30 +70,72 @@ script:
...
+
     # ThreatQ auth based on OAuth 2.0 credential grand method
     def tq_access():
         data = {'grant_type': 'password','email': EMAIL, 'password': PASSWORD, 'client_id': CLIENT_ID}
         access_token_response = requests.post(API_TOKEN_URL, data=data, verify=False, allow_redirects=False)
+
         tokens = json.loads(access_token_response.text)
+        if int(access_token_response.status_code) >= 400:
+            errors_string = get_errors_string_from_bad_request(access_token_response)
+            error_message = "Authentication failed, unable to retrieve an access token.\n {}".format(errors_string)
```
ronykoz: this should be done as part of `get_errors_string_from_bad_request`

Beta_Integrations/integration-ThreatQ.yml
```diff
@@ -70,30 +70,72 @@ script:
...
         token = tokens['access_token']
         return token
 
-    ACCESS_TOKEN = tq_access()
+
+    def access_token_not_expired():
+            epoch_time_now = time.time()
+            epoch_time_when_token_granted = demisto.getIntegrationContext().get("access_token_creation_time")
+            token_time_until_expiration = demisto.getIntegrationContext().get("access_token_expires_in")
+            return int(epoch_time_now) - int(epoch_time_when_token_granted) < int(token_time_until_expiration)
```
ronykoz: Better to put default values to make sure they are not None

Beta_Integrations/integration-ThreatQ.yml
```diff
@@ -70,30 +70,72 @@ script:
...
+
+
+    def get_access_token():
+        existing_access_token = demisto.getIntegrationContext().get("access_token")
+        if existing_access_token and access_token_not_expired():
+            return existing_access_token
+        else:
+            # What if the token is expired AND there are no more available access tokens?
```
ronykoz: We should get the error from tq_access, no?

guykeller: Yes, and how should we handle it? Just return the error or wait and retry?

Beta_Integrations/integration-ThreatQ.yml
guykeller: This function is only called if access_token is not None.
If access_token is not None, the other fields are not None as well, since all the fields are only defined together

Beta_Integrations/integration-ThreatQ.yml
guykeller: I'm not sure I agree.
The "get_errors_string_from_bad_request" function is generic, and in case it will be needed to clarify any other error messages it could be used in other cases too.
This message is specific to authentication, which means it should be a part of tq_access and not the generic command.
What do you think?

ronykoz: I don't agree for 100% but ok

Beta_Integrations/integration-ThreatQ.yml
ronykoz: I think that just returning the error would be fine

PR #3512, opened 2019-05-07
AWS - S3 Improvements

amshamah419:
## Status
In Progress

## Related Issues
fixes: link to the issue

## Description
Bring up to code standards as well as adding new functionality

## Does it break backward compatibility?
   - No

## Must have
- [x] Tests
- [x] Documentation (demisto/etc#17177)
- [x] Code Review

PR #3220, opened 2019-03-20
EC2 Improvements

amshamah419:
## Status
In Progress

## Related Issues
fixes: demisto/etc#16217
fixes: demisto/etc#16487

## Description
Added configuration to allow use of AWS access keys

## Does it break backward compatibility?
   - No

## Must have
- [x] Tests
- [x] Documentation (with link to it)
- [x] Code Review

## Additional changes
Added 'deprecated' to old aws integration name

PR #3548, opened 2019-05-15
add WAS tests to skipped

adi88d:
## Status
Ready

## Description
skip AWS tests until they will be fixed

## Related PRs
https://github.com/demisto/content/pull/3220
https://github.com/demisto/content/pull/3512


ShahafBenYakir:
@adi88d Please look at this test and say if it can be unskipped

amshamah419:
@ShahafBenYakir I'll take care of these. Most are destructive tests we don't want to run, and others are expensive to run. I believe one of them will be moved from skipped, but the others I need to verify.

PR #3556, opened 2019-05-16
remove validation for email and password outputs

adi88d:
## Status
Ready

## Related Issues
opened issue to be track later https://github.com/demisto/etc/issues/17022

## Description
The test failed because the command vigilante-query-domains output missing the email and the password for the account.

## Screenshots
Screen Shot 2019-05-16 at 15 02 11

Screen Shot 2019-05-16 at 15 03 43

### DIFF
--- a/TestPlaybooks/playbook-InfoArmorVigilanteATITest.yml
+++ b/TestPlaybooks/playbook-InfoArmorVigilanteATITest.yml
@@ -1,6 +1,6 @@
-id: InfoArmorVigilanteATITest2
+id: InfoArmorVigilanteATITest
 version: -1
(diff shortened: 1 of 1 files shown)


TestPlaybooks/playbook-InfoArmorVigilanteATITest.yml
```diff
@@ -1,6 +1,6 @@
-id: InfoArmorVigilanteATITest
+id: InfoArmorVigilanteATITest2
```
anara123: change the id and name to be `InfoArmorVigilanteATITest`

PR #3557, opened 2019-05-16
FortiGate - Fixed src adresses and dst adresses handling on get-policy

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16844

## Description
Before the fix, when calling get-policy on a policy with several destination or source addresses would only the first address would be shown. Now they are all shown

   - No

## Must have
- [ ] Code Review


Integrations/integration-Fortigate.yml
```diff
@@ -92,8 +92,6 @@ script:
         )
         if res.status_code not in {200}:
             return_error('Error in API call to FortiGate [%d] - %s' % (res.status_code, res.reason))
-        if method.upper() != 'GET':
```
ronykoz: why?

Integrations/integration-Fortigate.yml
```diff
@@ -410,6 +408,16 @@ script:
...
 
+    def create_addr_string(list_of_addr_data_dicts):
+        addr_string = ""
+        for addr_index in range(0, len(list_of_addr_data_dicts)):
+            cur_addr_data = list_of_addr_data_dicts[addr_index]
+            cur_addr_name = cur_addr_data.get("name")
+            if addr_index == len(list_of_addr_data_dicts) - 1:
+                addr_string += "{}".format(cur_addr_name)
+            else:
+                addr_string += "{},\n".format(cur_addr_name)
```
ronykoz:
```suggestion
                addr_string += "{}\n".format(cur_addr_name)
```

Integrations/integration-Fortigate.yml
guykeller: It caused missing data. POST methods return data as well

guykeller:
There are no tests for this since we have no instance, and no API documentation worth mentioning (building the integration through scraping)

Integrations/integration-Fortigate.yml
ronykoz: make sure no one uses the statuscode

PR #3558, opened 2019-05-16
update playbook script to make the dev process easy

anara123:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15539
fixes https://github.com/demisto/etc/issues/16518

## Description
created update_playbook.py script. Need to run this script like:
`python update_playbook.py my/path/to/phishing.yml` and it will modify and yml and save it in `playbook-phishing.yml` 
- will remove _dev, _copy
- will add empty description to each start,end,title tasks

   - No

## Must have
- [x] Tests
- [ ] Code Review


Utils/update_playbook.py
```diff
@@ -0,0 +1,126 @@
...
+    # replace version to be -1
+    playbook = replace_version(playbook)
+
+    playbook = update_id_to_be_equal_name(playbook)
+
+    if not destination_path:
+        destination_path = ntpath.basename(source_path)
+
+    if not destination_path.startswith("playbook-"):
+        destination_path = "playbook-{}".format(destination_path)
```
ronykoz: why not create it in the same folder?

anara123: I was thinking it is better experience if I create the output yml in the folder where I am. This way easily I just copy it where I need it to be

ronykoz: I think that the developers would prefer it to be copied to the playbooks dir.

PR #3574, opened 2019-05-20
ServiceNow - fixes

liorblob:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16960

## Description
Fixed couple of issues experienced by Wipro, they tested the fixes and confirmed it's working.

   - No


liorblob:
Not sure, those bugs only occur on their environment

PR #3579, opened 2019-05-21
optional flag to throw warning instead of errors for rasterize

orenzohar:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16858

## Description
adds a parameter that makes failed command throw warning instead of an error, so playbook will not fail.

## Must have
- [x] Tests
- [x] Code Review


Integrations/rasterize/rasterize.yml
```diff
@@ -3,6 +3,11 @@ commonfields:
   id: Rasterize
   version: -1
 configuration:
+- display: 'Throw Errors '
```
yuvalbenshalom: change to 'Return Errors' and add in the detailed description that explains that it will return warning entries instead of error entry when unmarked.

Integrations/rasterize/rasterize.yml
```diff
@@ -100,5 +105,6 @@ script:
   script: '-'
   type: python
 defaultEnabled: true
+releaseNotes: "added Param to throw warning instead of error"
```
yuvalbenshalom: change to "Added a parameter that permits to return warnings instead of errors"

PR #3582, opened 2019-05-21
Clear release notes - for deletecontext

yardensade:
## Status
Ready


yardensade:
I need force merge on this, it won't pass because of needed RN.
(We had to split deletecontext from the clear release notes build in order to pass it. Collected almost all tests)

PR #3587, opened 2019-05-21
PANW Cortex - handle logging query status

Itay4:
## Status
Ready

## Description
Added handling for logging query status.
Increased the max time out of logging query to 30 seconds.
For commands, if query status is `RUNNING` or `FAILED` - an error will be raised.
For fetch, if status is `RUNNING`, the query id will be passed to next run, which will try poll the results of query job is finished.
Removed the deletion of query from the service cause it is not longer needed in current sdk version.

4.1
   - No

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review

## Dependencies
None

### DIFF
--- a/Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
+++ b/Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
@@ -661,8 +661,11 @@ def fetch_incidents():
         response_json = response.json()
         query_status = response_json.get('queryStatus', '')
         if query_status == 'JOB_FAILED':
-            raise Exception('Logging query job failed with status: JOB_FAILED')
+            raise Exception(f'Logging query job failed with status: JOB_FAILED\nResponse: {response.text}')
         elif query_status == 'RUNNING':
+            if isinstance(last_fetched_event_timestamp, datetime):
+                # In case we don't have query ID from previous run
+                last_fetched_event_timestamp = last_fetched_event_timestamp.strftime('%Y-%m-%dT%H:%M:%S.%f')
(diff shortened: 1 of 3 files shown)


Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
```diff
@@ -608,32 +625,52 @@ def process_incident_pairs(incident_pairs, max_incidents):
...
 
-    response = query_loggings(query_data)
+        response = query_loggings(query_data)
 
     try:
-        result = response.json()['result']
+        response_json = response.json()
+        query_status = response_json.get('queryStatus', '')
+        if query_status == 'JOB_FAILED':
+            raise Exception('Logging query job failed with status: JOB_FAILED')
```
glicht: Let's include the full `response_json` object in the exception. There may be there additional information which is useful.

Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
```diff
@@ -608,32 +625,52 @@ def process_incident_pairs(incident_pairs, max_incidents):
...
+        elif query_status == 'RUNNING':
+            # If query job is still running after 30 seconds (max timeout), pass it to next run
+            demisto.setLastRun({
+                'last_fetched_event_timestamp': last_fetched_event_timestamp,
```
glicht: Shouldn't you be doing: strftime('%Y-%m-%dT%H:%M:%S.%f') as is done in the non `RUNNING` case?

PR #3588, opened 2019-05-21
Trigger script improvements

glicht:
## Status
Ready

## Description
Improvements to run a nightly build with a specific server version


glicht:
@ronykoz do you have anything to add? If not, I think we can merge.

ronykoz:
@glicht yes i think we can merge

PR #3590, opened 2019-05-22
Fixed update-policy bug

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16866

## Description
The "Update Policy" command used to have no effect on the actual policies, and return false success status. Now it works as expected.

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Fortigate.yml
+++ b/Integrations/integration-Fortigate.yml
@@ -127,14 +127,16 @@ script:
         params = {}
         session.post(SERVER+url_suffix, data=params, verify=USE_SSL)
 
-
-    def policy_addr_array_from_arg(policy_addr_str):
+    # When updating a policy we have to add another key to each addr, an q_origin key
+    def policy_addr_array_from_arg(policy_addr_str, should_add_origin_q = False):
         policy_adr_str_array = policy_addr_str.split(",")
         policy_addr_dict_array = []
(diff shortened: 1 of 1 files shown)


Integrations/integration-Fortigate.yml
```diff
@@ -127,14 +127,16 @@ script:
...
-    def policy_addr_array_from_arg(policy_addr_str):
+    # When updating a policy we have to add another key to each addr, an q_origin key
+    def policy_addr_array_from_arg(policy_addr_str, should_add_origin_q = False):
         policy_adr_str_array = policy_addr_str.split(",")
         policy_addr_dict_array = []
         for src_addr_name in policy_adr_str_array:
             cur_addr_dict = {
                 "name": src_addr_name
             }
+            if should_add_origin_q:
```
ronykoz: it seems you are not using this

Integrations/integration-Fortigate.yml
```diff
@@ -127,7 +127,7 @@ script:
         params = {}
         session.post(SERVER+url_suffix, data=params, verify=USE_SSL)
 
-
+    # When updating a policy we have to add another key to each addr, an q_origin key
```
ronykoz: irrrelevant here

PR #3592, opened 2019-05-22
add Minerva to skipped integrations - INSTANCE ISSUES

adi88d:
## Status
Ready

## Description
add Minerva to skipped integrations - no instance


yardensade:
Should also note why it fails, liorblob suggested they probably shutdown the instance.
IF we are not sure, should note it is pending checks.

PR #3601, opened 2019-05-23
DocumentationAutomation Enhancement

yaakovi:
## Status
Ready

## Description
run command examples by the order they appear in the input.
easier to create command file (can include setup/teardown commands).

   - Yes: command must include the prefix of `!`. this was made for the option to include comments in the command file.
since no one will use this in a playbook, then it doesn't really matters.

## Must have
- [ ] Tests
- [x] Code Review


Scripts/DocumentationAutomation/DocumentationAutomation_test.py
```diff
@@ -0,0 +1,84 @@
...
+    # error count should not change
+    assert return_error_mock.call_count == 0
+    # call_args last call with a tuple of args list and kwargs
+    assert data['commonfields']['id'] == 'ANYRUN'
+
+    # invalid yml
+    mocker.patch.object(demisto, 'getFilePath',
+                        return_value={'path': os.path.join('..', '..', 'Integrations', 'ANYRUN', 'ANYRUN.py')})
+    get_yaml_obj('234')
+    assert return_error_mock.call_count == 1
```
ronykoz: why is it changing now?

yaakovi: because i passed a python file and not a yaml

PR #3609, opened 2019-05-26
ReadFile - enabling auto-detection of encoding

amshamah419:
## Status
In Progress

## Related Issues
fixes: demisto/etc#16953

## Description
Main issue amounted to user error, but the implementation was not as intuitive as it could have been. We are introducing the ability to detect the file's encoding type to take the guesswork out of a customers hands. It also allows this script to run automatically within a playbook without using default encoding types.

   - No

~~In its current implementation, we are planning for a V2 which will not break backwards compatibility. If we are able to detect docker images, then we can keep the original and still prevent breaking backwards compatibility.~~

Chardet is part of the default python docker image so there will be no requirement to move docker images.

## Must have
- [x] Tests
- [ ] Code Review


glicht:
docker files don't go here. also see my comment on the issue.

Scripts/ReadFile/ReadFile.py
```diff
@@ -0,0 +1,54 @@
...
+    if max_file_size:
+        max_file_size = int(max_file_size)
+    else:
+        max_file_size = 1024 ** 2
+    res = demisto.executeCommand('getFilePath', {'id': demisto.args()['entryID']})
+    file_path = res[0]['Contents']['path']
+    if not demisto.args().get('encoding'):
+        char_enc = get_file_encoding(file_path)
+    else:
+        demisto.args().get('encoding')
```
yuvalbenshalom: looks like you do not assign the argument to  char_enc

PR #3621, opened 2019-05-27
Enabled keeping the existing data in update-policy

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17186

## Description
The update-policy command used to replace the existing data with the arguments supplied.
Now there is a possibility to keep the existing data and add / remove the supplied arguments to it.

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Fortigate.yml
+++ b/Integrations/integration-Fortigate.yml
@@ -119,6 +119,26 @@ script:
             return addr_string
 
 
+    def bool_to_str(str_representing_bool):
+        return str_representing_bool and str_representing_bool.lower() == 'true'
+
+
+    def generate_src_or_dst_request_data(policy_id, policy_field, policy_field_value, keep_original_data, add_or_remove):
+        address_list_for_request = policy_field_value.split(",")
+        if bool_to_str(keep_original_data):
+            policy_data = get_policy_request(policy_id)[0] # the return value is an array with one element
+            existing_adresses_list = policy_data.get(policy_field)
+            existing_adresses_list = [address_data["name"] for address_data in existing_adresses_list]
+            if add_or_remove.lower() == "add":
+                for address in existing_adresses_list:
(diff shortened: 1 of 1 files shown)


ronykoz:
@kirbles19 this is for the next release, could you check out the new arguments description please? + the RN

Integrations/integration-Fortigate.yml
```diff
@@ -119,6 +119,26 @@ script:
...
+        if bool_to_str(keep_original_data):
+            policy_data = get_policy_request(policy_id)[0] # the return value is an array with one element
+            existing_adresses_list = policy_data.get(policy_field)
+            existing_adresses_list = [address_data["name"] for address_data in existing_adresses_list]
+            if add_or_remove.lower() == "add":
+                for address in existing_adresses_list:
+                    if not address in address_list_for_request:
+                        address_list_for_request.append(address)
+            else:
+                address_list_for_request = [address for address in existing_adresses_list if address not in address_list_for_request]
```
ronykoz: how is it different from lines 133-135

Integrations/integration-Fortigate.yml
```diff
@@ -119,6 +119,26 @@ script:
...
+        address_data_dicts_for_request = policy_addr_array_from_arg(address_list_for_request, False)
```
ronykoz: are you making it a recursive call?

Integrations/integration-Fortigate.yml
guykeller: Look again, these are two opposite actions:
On lines 133-135, I add every address from the existing addresses to the arg addresses if it isn't there yet.
The purpose of these lines is to make sure the updated source/destination includes all the existing addresses + the new ones.
On the other hand, on line 137 I'm taking the existing addresses and removing all the arg addresses.

PR #3624, opened 2019-05-27
Fixed Extract Indicators From File playbook issue with excel files

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17137

## Description
The playbook used to identify Excel sheets as Word documents. Now it does not.

## Screenshots
Ran the new playbook on an Excel sheet, it was successful:
Screen Shot 2019-05-27 at 17 06 43

   - No

## Must have
- [ ] Code Review


guykeller:
The part you specified is needed because not all files have an extension in context (tested that and seen for myself), but excel files do. This could cause problems for non-excel files.

As for the other two points, changing as requested

idovandijk:
Are you sure "extension doesn't exist" is still needed? I think the File.Type changes should suffice. 
Also File.Type doesn't exist is risky. And the 2nd tasks still must be changed accordingly.

Can we try the following:
File.Info doesn't equal xls (IGNORE CASE)
AND
FIle.Info doesn't equal xlsx (IGNORE CASE)
AND 
File.Type doesn't include excel (IGNORE CASE)

This way we avoid relying on the extension of the file

PR #3612, opened 2019-05-26
Add Domain ID Search to Q Radar

amshamah419:
## Status
In Progress

## Related Issues
fixes: demisto/etc#15857

## Description
Added two commands as well as performed some code cleanup and refactoring.

## Does it break backward compatibility?
   - Other:
      - 🤞 

## Must have
- [x] Tests
- [ ] Documentation (demisto/etc#17199)
- [ ] Code Review

PR #3625, opened 2019-05-27
Package QRadar

amshamah419:
## Status
Ready

Packaging QRadar to enable improvements in the next PR - https://github.com/demisto/content/pull/3612


Integrations/QRadar/integration-QRadar.py
```diff
@@ -0,0 +1,971 @@
...
+# Retrieves domain name using domain id
+def get_domain_name(domain_id):
+    try:
+        query_param = {
+            'query_expression': "SELECT DOMAINNAME({0}) AS 'Domain name' FROM events GROUP BY 'Domain name'".format(
+                domain_id)}
+        search_id = search(query_param)['search_id']
+        return get_search_results(search_id)['events'][0]['Domain name']
+    except Exception:
+        demisto.results({
```
yuvalbenshalom: please return also the exception str in the warning entry so it will not be hided

amshamah419: Fixed

PR #3630, opened 2019-05-28
Deprecate Cymon

amshamah419:
## Status
Ready

## Description
Packaging Cymon so we can perform bug fixes at a later PR


amshamah419:
I don't believe we have a working instance because yes, it's completely shut down

Itay4:
Why are we still supporting this if it’s being deprecated?

amshamah419:
In my opinion, we really should just remove it.

Itay4:
Let’s start with deprecating the commands and add (Deprecated) to the display name

yuvalbenshalom:
I think that the whole integration should be replaced with a code that return the same error entry that say that this service dose not exist any more.
and also to mark every command as deprecated.

amshamah419:
Test button now returns an error indicating service is discontinued.
Screen Shot 2019-06-05 at 11 24 16

Deprecated is now part of the display name.
Screen Shot 2019-06-05 at 11 24 23

For commands, it now returns an error indicating the service is discontinued.
Screen Shot 2019-06-05 at 11 24 39

Itay4:
Let’s discuss before merging

gshriki:
What happens today, before this change, for customers who have Cymon configured?

Think we should say: `Cymon service discontinued. Please disable or delete the integration instannce`.
We also need to have proper documentation in the release notes.

amshamah419:
As of right now, customers get an error message indicating that the http handler failed to connect. This is technically correct, since the endpoint has simply vanished, however most customers would see it as a failure on our end as opposed to Cymon being defunct.

PR #3635, opened 2019-05-28
Alienvault "URL" command fix

orenzohar:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/17004

## Description
if `!url` in AlienvaultOTX doesn't get the base URL, it fails. added a function that parses the base URL, and informative error when it fails to do so, so the user might be able to do it.

## Must have
- [x] Tests
- [x] Code Review


Integrations/integration-AlienVaultOTX.yml
```diff
@@ -452,3 +459,6 @@ script:
     description: Query url in AlienVault OTX
   runonce: false
 fromversion: 3.0.1
+releaseNotes: fixing !url plus adding a test
```
yuvalbenshalom: please change to "Fix !url to try extract base url, and return a readable error in case of failure"

PR #3636, opened 2019-05-28
created GIT integration

reutshal:
## Status
In Progress

## Related Issues
fixes: link to the issue

## Description
created github integration


Tests/conf.json
```diff
@@ -1213,6 +1213,10 @@
             "playbookID": "Freshdesk-Test",
             "timeout": 500,
             "nightly": true
+        },
+                {
```
Itay4: please lint properly

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+    except Exception as e:
+        raise Exception(e.message)
+    demisto.results("ok")
+
+
+''' EXECUTION CODE '''
+
+command = demisto.command()
+# LOG('command is %s' % (command,))
```
ronykoz: remove

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+def test_module():
+    """
+    Performs basic get request to get item samples
+    """
+    try:
+        http_request('GET', '/repos/reutshal/GitHubRepo/issues')
+
+    except Exception as e:
+        raise Exception(e.message)
```
ronykoz: use return_error instead + add an empty line after this code block

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+''' EXECUTION CODE '''
+
+command = demisto.command()
+# LOG('command is %s' % (command,))
+
+try:
+    if command == 'test-module':
+        test_module()
+    elif demisto.command() == 'fetch-incidents':
```
ronykoz:
```suggestion
    elif command == 'fetch-incidents':
```

Integrations/GitHub/GitHub.yml
```diff
@@ -0,0 +1,190 @@
...
+- display: apikey
+  name: apikey
+  required: true
+  type: 0
+- display: repository
+  name: repository
+  required: true
+  type: 0
+- defaultvalue: '720'
+  display: fetch_interval
```
ronykoz: you should have an explanation for this, is it minutes? hours? days?

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+        'Git.issue(val.ID && val.ID == obj.ID)': issue_list
+    }
+
+    return_outputs(human_readable, entry_context, issues)
+
+
+def create_issue(args):
+    suffix = '/repos/{}/{}/issues'.format(OWNER, REPO)
+    data = {}
+    for key in args.keys():
```
ronykoz: why do you need to copy this dict? + why the `if 'labels' in data:` and `if 'assignees' in data:` not in the loop?

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+        data[key] = args[key]
+
+    if 'labels' in data:
+        data['labels'] = data['labels'].split(',')
+
+    if 'assignees' in data:
+        data['assignees'] = data['assignees'].split(',')
+
+    data = json.dumps(data)
+    # return_error(data)
```
ronykoz: ?

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    res = http_request('POST', suffix, data, None)
```
ronykoz: you have a default value of None for the fourth element, so you don't have to send None to the function

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    demisto.incidents(incidents)
+    demisto.setLastRun({'time': last_run_time.isoformat().split('.')[0] + 'Z'})
+
+
+''' COMMANDS + REQUESTS FUNCTIONS '''
+
+
+def get_issue_context(issue):
+    label_context = []
+    for l in issue['labels']:
```
ronykoz: l ?
Please provide a better name

ronykoz: maybe try to do it in a one liner as well
label_context = [label['name'] for label in issue['labels']]

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+        label_context.append(l['name'])
+
+    assignee_context = []
+    for a in issue['assignees']:
```
ronykoz: same here

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+def create_issue_command():
+    args = demisto.args()
+    res = create_issue(args)
+    context = get_issue_context(res)
+
+    header_list = ['ID', 'title', 'state', 'locked', 'body', 'assignees', 'labels']
+    human_readable = tableToMarkdown('created issue successfully: ', context, header_list, removeNull=True)
+    entry_context = {
+        'Git.issue(val.id && val.id == obj.id)': context
```
ronykoz:
```suggestion
        'Git.Issue(val.ID && val.ID == obj.iD)': context
```

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    return_outputs(human_readable, entry_context, res)
+
+
+def close_issue(issue_id):
+    suffix = '/repos/{}/{}/issues/{}'.format(OWNER, REPO, issue_id)
```
ronykoz: I think it would be nicer to have all the suffixes as CONST and then just to fill them in the functions

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    data = {"state": "close"}
+
+    data = json.dumps(data)
+    res = http_request('PATCH', suffix, data, None, 'close')
```
ronykoz: when sending so many params better to use the argument name as well + no need for NONE
`http_request('PATCH', suffix, data, command='close')`

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+def close_issue_command():
+    args = demisto.args()
+    res = close_issue(args['issue_id'])
+
+    context = get_issue_context(res)
+
+    header_list = ['ID', 'title', 'state', 'locked', 'body', 'assignees', 'labels']
+    human_readable = tableToMarkdown('issues list: ', context, header_list)
```
ronykoz:
```suggestion
    human_readable = tableToMarkdown('Issue list', context, header_list)
```

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    entry_context = {'Git.issue(val.ID && val.ID == obj.ID)': context}
```
ronykoz: multiline

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+    return_outputs(human_readable, entry_context, res)
+
+
+def update_issue(args):
```
ronykoz: seems very similar to another function, create a function that suits them both

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+def update_issue_command():
+    args = demisto.args()
+    res = update_issue(args)
+
+    context = get_issue_context(res)
+
+    header_list = ['ID', 'title', 'state', 'locked', 'body', 'assignees', 'labels']
+    human_readable = tableToMarkdown('issues list: ', context, header_list)
+    entry_context = {'Git.issue(val.ID && val.ID == obj.ID)': context}
```
ronykoz: same here

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+    return_outputs(human_readable, entry_context, res)
+
+
+def search_issue(issue_id):
+    suffix = '/repos/{}/{}/issues/{}'.format(OWNER, REPO, issue_id)
+    res = http_request('GET', suffix, None, None)
```
ronykoz: remove Nones

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+
+def search_issue_command():
+    issue_id = demisto.args()['issue_id']
+    res = search_issue(issue_id)
+
+    context = get_issue_context(res)
+
+    header_list = ['ID', 'title', 'state', 'locked', 'body', 'assignees', 'labels']
+    human_readable = tableToMarkdown('issues list: ', context, header_list)
+    entry_context = {'Git.issue(val.ID && val.ID == obj.ID)': context}
```
ronykoz: multiline

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    res = get_download_count()
+    header_list = ['ID', 'release name', 'download_count']
+    download_counts = []
+    for release in res:
+        count = 0
+
+        for asset in release['assets']:
+            count = count + asset['download_count']
+
+        download_counts.insert(0, {
```
ronykoz: create the dict separately, seems off this way

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+            'ID': release['id'],
+            'release name': release['name'],
+            'download_count': count})
+
+    human_readable = tableToMarkdown('the download count is:', download_counts, header_list)
+    entry_context = {'Git.issue(val.ID && val.ID == obj.ID)': res}
```
ronykoz: multiline

TestPlaybooks/playbook-GitHub_test.yml
```diff
@@ -0,0 +1,146 @@
...
+  "3":
+    id: "3"
+    taskid: 1538a5db-b23d-4a6a-8bc0-6150f4006e2e
+    type: regular
+    task:
+      id: 1538a5db-b23d-4a6a-8bc0-6150f4006e2e
+      version: -1
+      name: closeInvestigation
+      description: Close the current incident
+      script: Builtin|||closeInvestigation
```
ronykoz: you need to add more checks for your playbook

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+BASE_URL = 'https://api.github.com'
+FETCH_INTERVAL = demisto.params()['fetch_interval']
+
+
+''' HELPER FUNCTIONS '''
+
+
+def http_request(method, url_suffix, json=None, params=None, command=None):
+    if method == 'GET':
+        headers = {}
```
ronykoz: no need for headers in GET???

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+    elif method in ('POST', 'PATCH'):
+        if not API_KEY:
```
ronykoz: why won't you have a key?

ronykoz: Just save the headers as CONST

Integrations/GitHub/GitHub.py
```diff
@@ -0,0 +1,312 @@
...
+        if res.status_code == 410:
+            return_error("Error - this issue was deleted")
+
+        elif res.status_code == 404:
+            return_error("Error - this issue doesn't exists ")
+
+        return res.json()
+
+    except Exception as e:
+        raise(e)
```
ronykoz: return_error("Error:\n{}".format(str(e)))

Integrations/GitHub/GitHub.py
```diff
@@ -17,33 +17,28 @@
 REPO = demisto.params().get('repository')
 BASE_URL = 'https://api.github.com'
 FETCH_INTERVAL = demisto.params()['fetch_interval']
-
+SUFFIX = {'list': '/repos/{}/{}/issues'.format(OWNER, REPO),
```
ronykoz: No real need in dictionary here, but if you already did that with a dict please format it accordingly

Integrations/GitHub/GitHub.py
```diff
@@ -17,33 +17,28 @@
...
+          'create': '/repos/{}/{}/issues'.format(OWNER, REPO),
+          'close': '/repos/{}/{}/issues/'.format(OWNER, REPO,),
+          'update': '/repos/{}/{}/issues/'.format(OWNER, REPO),
+          'search': '/repos/{}/{}/issues/'.format(OWNER, REPO),
+          'download_count': '/repos/{}/{}/releases'.format(OWNER, REPO)
+          }
+HEADERS = {'Content-Type': 'application/json',
```
ronykoz:
```
HEADERS = {
    'Content-Type': 'application/json',
    'Accept': 'application/vnd.github.v3+json',
    'Authorization': 'Bearer ' + API_KEY
}
```

Integrations/GitHub/GitHub.py
```diff
@@ -116,9 +107,9 @@ def get_issue_context(issue):
 
 
 def List_all_issues(show_all):
-    suffix = '/repos/{}/{}/issues'.format(OWNER, REPO)
+    suffix = SUFFIX['list']
     if show_all == 'false':
-        res = http_request('GET', suffix, None, None)
+        res = http_request('GET', suffix)
     else:
         res = http_request('GET', suffix, None, {'state': 'all'})  # include close issues
```
ronykoz: no need to send the json param if you are not using it, just send the state dict as `params={'state': 'all'}`

Integrations/GitHub/GitHub.py
ronykoz: You should probably do something like this:
```
if 'labels' in args.keys():
    args['labels'] = args['labels'].split(',')
if 'assignees' in args.keys():
    args['assignees'] = args['assignees'].split(',')
```
No need for the loop

Integrations/GitHub/GitHub.py
```diff
@@ -128,148 +119,127 @@ def List_all_issues(show_all):
...
+            args['labels'] = args['labels'].split(',')
+        if key == 'assignees':
+            args['assignees'] = args['assignees'].split(',')
+    args = json.dumps(args)
+    if command == 'create':
+        suffix = SUFFIX['create']
+        res = http_request('POST', suffix, json=args)
+    else:
+        suffix = SUFFIX['update'] + demisto.args()['issue_id']
+        res = http_request('PATCH', suffix, args)
```
ronykoz: Are you sure it's ok to send the issue_id arg as well?

Integrations/GitHub/GitHub.py
```diff
@@ -128,148 +119,127 @@ def List_all_issues(show_all):
...
 
-def close_issue(issue_id):
-    suffix = '/repos/{}/{}/issues/{}'.format(OWNER, REPO, issue_id)
+def close_issue():
+    suffix = SUFFIX['close'] + demisto.args()['issue_id']
     data = {"state": "close"}
-
     data = json.dumps(data)
-    res = http_request('PATCH', suffix, data, None, 'close')
+    res = http_request('PATCH', suffix, data, params='close')
```
ronykoz: shouldn't the params be a dictionary?

Integrations/GitHub/GitHub.py
reutshal: Why not? issue_id is a mandatory argument to the Update command

Integrations/GitHub/GitHub.py
reutshal: This shouldn't be here, something I tried to do in the previous version. 
This specific request only uses the data argument so it didn't make any difference in the command behaviour. Removed

IkaDemisto:
from @ronykoz request, closed

PR #3637, opened 2019-05-28
Fix local file validation

avidan-H:
## Status
Ready

## Description
Fixes the issue of `validate_files.py` trying to validate `*_unified.yml` files locally if they appear before the integration yml in a given directory.
Also on occasion the `get_modified_files` function would raise an `IndexError`. I believe that that this would occur when `glob.glob` didn't return any results because the directory (and files of the directory) of the modified file in the list of modified files in `origin/master` do not exist in the local branch being worked on. Now the loop will continue if this happens.

## Screenshots
The `validate_files.py` mistakenly trying to validate an unstaged `*_unified.yml` file.
image

x.x.x

   - Don't think so

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Pre-commit hooks

### DIFF
--- a/Tests/scripts/validate_files.py
+++ b/Tests/scripts/validate_files.py
@@ -72,7 +72,10 @@ class FilesValidator(object):
 
             if checked_type(file_path, CODE_FILES_REGEX) and file_status.lower() != 'd':
                 dir_path = os.path.dirname(file_path)
-                file_path = glob.glob(dir_path + "/*.yml")[0]
+                try:
+                    file_path = list(filter(lambda x: not x.endswith('unified.yml'), glob.glob(dir_path + "/*.yml")))[0]
+                except IndexError:
+                    continue
             elif file_path.endswith('.js') or file_path.endswith('.py'):
                 continue
 


Tests/scripts/validate_files.py
```diff
@@ -72,7 +72,10 @@ def get_modified_files(files_string):
 
             if checked_type(file_path, CODE_FILES_REGEX) and file_status.lower() != 'd':
                 dir_path = os.path.dirname(file_path)
-                file_path = glob.glob(dir_path + "/*.yml")[0]
+                try:
+                    file_path = list(filter(lambda x: not x.endswith('unified.yml'), glob.glob(dir_path + "/*.yml")))[0]
```
Itay4: `filter` return `list`, why do you need to cast it ?

avidan-H: I thought that `filter` returned a generator and not a list but tbh I could be mixing it up with something else - in which case it's completely unnecessary

Itay4: it does return an iterator in py3, but not in py2

avidan-H: rony-scripts are py2, so it makes sense as-is, no?

Itay4: no, because in py2 you get a list from filter

ronykoz: Why would we get an index error? Can you think of such scenario?

avidan-H: it happened to me

ronykoz: Describe the scenario pls

PR #3650, opened 2019-06-02
Vmray fixing

jochman:
## Status
Ready

## Description
Fixed error messages, duplicate entries and file attributes from get_iocs command.
unmocked from tests: issue #17307

## Related PRs

branch | PR
------ | ------

x.x.x

   - No


Integrations/VMRay/VMRay.py
```diff
@@ -830,20 +835,48 @@ def get_iocs_command():
...
             urls_list.append(entry)
 
+    files = data.get('files')
+    files_list = list()
+    if files:
+        for file_entry in files:
+            size = len(files)
+            iocs_size_table['File'] = size
+            iocs_size += size
+            entry = dict()
```
Itay4: no need to initialise and then fill it, do it in the same operation

Integrations/VMRay/VMRay.py
```diff
@@ -830,20 +835,48 @@ def get_iocs_command():
...
+            entry['AnalysisID'] = file_entry.get('analysis_ids')
+            entry['Filename'] = file_entry.get('filename')
+            entry['Operation'] = file_entry.get('operations')
+            entry['ID'] = file_entry.get('id')
+            hashes_list = list()
+            hashes = file_entry.get('hashes')
+            for hash_entry in hashes:
+                hashes_dict = dict()
```
Itay4: no need to initialise, fill it and then append it, do it in the same operation

PR #3652, opened 2019-06-02
Proxy process validation

ShahafBenYakir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17327

## Description
Added validation to proxy process when running tests to make it is up and running before initiating any test


Tests/mock_server.py
```diff
@@ -267,6 +269,18 @@ def start(self, playbook_id, path=None, record=False):
         if self.process.returncode is not None:
             raise Exception("Proxy process terminated unexpectedly.\nExit code: {}\noutputs:\nSTDOUT\n{}\n\nSTDERR\n{}"
                             .format(self.process.returncode, self.process.stdout.read(), self.process.stderr.read()))
+        log_file_status = 2
+        counter = 0
+        # Make sure process is up and running
+        # While log file doesn't exist and less than PROXY_PROCESS_INIT_TIMEOUT seconds has passed
+        while log_file_status > 0 and counter < PROXY_PROCESS_INIT_TIMEOUT:
+            # Check if log file exist
+            log_file_status = self.ami.call(['ls', log_file])
```
BenJoParadise: It might be nicer to have a boolean that tells us if the `log_file_exists`:
- Change `log_file_status` to `log_file_exists`
- init it to `False`
- set it to contain the (boolean) value of `self.ami.call(['ls', log_file]) == 0`

Tests/mock_server.py
```diff
@@ -267,6 +269,18 @@ def start(self, playbook_id, path=None, record=False):
         if self.process.returncode is not None:
             raise Exception("Proxy process terminated unexpectedly.\nExit code: {}\noutputs:\nSTDOUT\n{}\n\nSTDERR\n{}"
                             .format(self.process.returncode, self.process.stdout.read(), self.process.stderr.read()))
+        log_file_status = 2
+        counter = 0
+        # Make sure process is up and running
+        # While log file doesn't exist and less than PROXY_PROCESS_INIT_TIMEOUT seconds has passed
+        while log_file_status > 0 and counter < PROXY_PROCESS_INIT_TIMEOUT:
```
BenJoParadise: In it's current logic, `PROXY_PROCESS_INIT_TIMEOUT` is the number of times that `sleep` happened. It would be nicer if `PROXY_PROCESS_INIT_TIMEOUT` was the number of seconds that passed since trying to raise the process. Here's how we can change it:

- Instead of `counter`, have a variable like `seconds_since_init` - it should hold the number of seconds that we've already waited since starting the loop.
- During each iteration, the value of `seconds_since_init` should increase by `PROXY_PROCESS_INIT_INTERVAL` seconds
- Compare it to `PROXY_PROCESS_INIT_TIMEOUT` in the `while` condition
- ???
- PROFIT!! - we can now define `PROXY_PROCESS_INIT_TIMEOUT` to be how long (in seconds) to wait for the process to start, and `PROXY_PROCESS_INIT_INTERVAL` will be how long to sleep before the next iteration.

ShahafBenYakir: - [x] Done

Tests/mock_server.py
ShahafBenYakir: - [x] Done

PR #3653, opened 2019-06-03
Fixed a bug where the playbook would fail if the integration was disabled

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17399

## Description
There was a missing "else" path when running a check for whether the integration is enabled.

4.1.0

   - No


idovandijk:
Failing due to Active Directory. Says:
`Body: Failed to access LDAP server. Please validate the server host and port are configured correctly`
Integration test passes, not sure what the problem is.

idovandijk:
Failing due to Active Directory. Says:
Body: Failed to access LDAP server. Please validate the server host and port are configured correctly
Integration test passes, not sure what the problem is.
@noaco any idea?

PR #3656, opened 2019-06-03
New playbook Extract Indicators From File - Generic v2, new test playbook, improved test data

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17209

## Description
- Created the playbook "Extract Indicators From File - Generic v2"
- Created a test that also checks for the image-extracted text using Image OCR
- Updated the test PDF file to contain an image to use with Image OCR.
- Added a fix to distinguish between Excel and Word files

Will later add Phishing v2 with the new changes, to this PR, which has those changes:
- Uses Email Address Enrichment - Generic **v2.1** where it previously used v2.0
- Calls all enrichment subplaybooks directly instead of through `Entity Enrichment - Phishing`. This will reduce complexity and maintenance. Inputs of `Entity Enrichment - Phishing v2` were moved to the inputs of `Phishing Investigation - Generic v2` or left to be entered directly for each playbook subplaybook
- Visual rework of `Phishing Investigation - Generic v2`

## Screenshots
image

4.1.0

   - Yes - but only for Phishing Investigation - Generic v2, because it will use the enrichment playbooks directly

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review


idovandijk:
@kirbles19, my changes are the 3 tasks at the bottom right corner, and the description of File.Text output of the playbook, so not much to be reviewed on your side.
They're all under the file `playbook-Extract_Indicators_From_File_-_Generic_v2.yml`. You can find them by the names of the tasks.

idovandijk:
Having this problem with the build
image

Checked with Dean and adding urlregex manually to CommonServerUserPython helps fix the problem. However, re-running the build will produce this error again. For some reason urlregex is found in CommonServerPython
image

But is not present in the test server itself
image


@ronykoz I was told you may know what the problem is

PR #3657, opened 2019-06-03
Proxy process check

ShahafBenYakir:
## Status
In progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/17327

## Description
Added validation to proxy process when running tests to make it is up and running before initiating any test


Tests/mock_server.py
```diff
@@ -274,12 +274,13 @@ def start(self, playbook_id, path=None, record=False):
         # Make sure process is up and running
         while not log_file_exists and seconds_since_init < PROXY_PROCESS_INIT_TIMEOUT:
             # Check if log file exist
-            log_file_exists = self.ami.call(['ls', log_file]) == 0
+            log_file_exists = silence_output(self.ami.call, ['ls', log_file], stdout='null', stderr='null') == 0
```
yuvalbenshalom: what dose this silence_output do? @ShahafBenYakir

ShahafBenYakir: Making sure the shell commands output won’t appear in the Circle log, only messages/errors we decide to print from this python script will appear.

PR #3671, opened 2019-06-05
Added 'proxy' and 'secure' options to DuoAdmin integration

teizenman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16893

## Description
Added the user ability to choose whether he would like to use proxy, and whether to use SSL with DuoAdmin


Integrations/DuoAdminApi/DuoAdminApi.py
```diff
@@ -113,19 +113,19 @@ def test_instance():
     except Exception as e:
         if hasattr(e, 'data'):
             # error data for 40103 is not informative enough so we write our own
-            if e.data['code'] == 40103:
+            if e.__getattribute__('data')['code'] == 40103:
```
ronykoz: why not get?

Integrations/DuoAdminApi/DuoAdminApi.py
```diff
@@ -70,7 +70,7 @@ def get_entry_for_object(title, obj, contents, context, headers=None, human_read
         }
 
     if headers:
-        if isinstance(headers, basestring):
+        if isinstance(headers, (str, unicode)):
```
ronykoz: are you sure it's working?

teizenman: from the Python software foundation:

> This abstract type is the superclass for str and unicode. It cannot be called or instantiated, but it can be used to test whether an object is an instance of str or unicode. isinstance(obj, basestring) is equivalent to isinstance(obj, (str, unicode)).

ronykoz:
Just add RN(ReleaseNotes section with the explanation of what you did and we are good to go)

PR #3673, opened 2019-06-05
Add Python3 subtype

Itay4:
## Status
Ready

## Description
Added subtype `python3` to relevant scripts and integrations and to schemas

Will apply from version 5.0.0

   - No


glicht:
@yaakovi see commits I added to this PR 

I added that the release notes file is generated even if some are missing (build still fails) so tests will continue as usual. See an example here: https://circleci.com/gh/demisto/content/24304?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Also added logic to treat release branches 19.* and 20.* same way as we treat master when validating files.

PR #3678, opened 2019-06-06
ReadPDFFileV2 bug fix

roysagi:
## Related Issues
https://github.com/demisto/etc/issues/17219
## Description
Worked in a bug in ReadPDFFileV2 - the required behavior was that all context created about the PDF file should be listed in the script's outputs for further use in playbooks.

Sat with Dean on the matter to make sure all fields are accounted for.
Added description to all added fields as well as adjusted the 'type'  field for all outputs.

Also to the request of Dean, added some code to camelize the field names.

branch | PR
------ | ------

x.x.x

   - Yes
       - Further details:
   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes


roysagi:
@DeanArbel and @idovandijk your assigned so you can check up on this bug if you want to.
I'm handling it :)

PR #3684, opened 2019-06-10
Now CRLF!

idovandijk:
## Status
Ready

## Description
I was having problems with the EML file having LF line separators, instead of CRLF.
There were lots of issues with forcing the CRLF change. Trying it now here.
The reason for this is that `Process Email - Generic` has a conditional task with a DT that expects CRLF but not LF, so LF EMLs are not detected as attached emails.

## Related PRs
https://github.com/demisto/content/pull/3656/

   - No


idovandijk:
Still not CRLF..

PR #3687, opened 2019-06-10
Extract Indicators From File - Generic and v2 fix

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17351

## Description
Removed `File.Info` check for word files in the playbooks `Extract Indicators From File - Generic` after a customer encountered a malicious file that pretended to have file info of doc. The reason I removed it and not changed it to an `and` condition is because while testing the condition I found online some word files that did not have doc or docx file info, as such the field is unreliable.

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Playbooks/playbook-Extract_Indicators_From_File_-_Generic.yml
+++ b/Playbooks/playbook-Extract_Indicators_From_File_-_Generic.yml
@@ -485,22 +485,6 @@ tasks:
                     right:
                       value:
                         simple: Microsoft Word
-                  - operator: isEqualString
-                    left:
-                      value:
-                        simple: File.Info
-                      iscontext: true
-                    right:
-                      value:
-                        simple: doc
-                  - operator: isEqualString
-                    left:
-                      value:
-                        simple: File.Info
-                      iscontext: true
-                    right:
-                      value:
-                        simple: docx
                 - - operator: isNotEqualString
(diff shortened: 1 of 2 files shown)


DeanArbel:
Test fails on One Before because of known issues with CommonServerPython and scripts/playbooks that use that argument.

idovandijk:
Need to test this with a variety of DOC, DOCX, XLS, XLSX, TXT, EML and PDF files to make sure we don't break something.
As for the problem with the "Info", need to check with server how "File.Info" is populated. Maybe we can rely on "Type" which is taken from the result of the unix `file` command.

yuvalbenshalom:
@DeanArbel what is the functionality now? just checks the file extension for doc/docx?

DeanArbel:
Here is the condition. I do not check the file extension, it was not part of the original condition and I believe that it might be unreliable as well.
image

DeanArbel:
@idovandijk do you have a set of such documents? If not, let's compile one like that together and add it to the test.

idovandijk:
We should test this with multiple DOC, DOCX, XLS, XLSX, PDF, TXT, EML and MSG files. If you need some files to test this on, DM me, I should be able to gather some files.

PR #3690, opened 2019-06-10
remove logging in appropriate places

orenzohar:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17440

## Description
remove logging in appropriate places in ArcSight v2

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/ArcSightESMv2/ArcSightESMv2.py
+++ b/Integrations/ArcSightESMv2/ArcSightESMv2.py
@@ -122,7 +122,6 @@ def decode_arcsight_output(d, depth=0, remove_nones=True):
     return d
 
 
-@logger
 def login():
     query_path = 'www/core-service/rest/LoginService/login'
     headers = {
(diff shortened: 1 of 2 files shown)


Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -136,7 +135,6 @@ def login():
     }
     res = send_request(query_path, headers=headers, params=params, is_login=True)
     if not res.ok:
-        demisto.debug(res.text)
```
anara123: I think we can leave it

Integrations/ArcSightESMv2/ArcSightESMv2.py
```diff
@@ -185,8 +182,7 @@ def send_request(query_path, body=None, params=None, json=None, headers=None, me
             )
         return res
 
-    except Exception as ex:
-        demisto.debug(str(ex))
+    except Exception:
```
anara123: We can still print the exception.

orenzohar:
## Before
image

## After
image

PR #3691, opened 2019-06-11
Powershell content support

Itay4:
## Status
Ready

## Description
This PR adds support for Powershell script type on content side.
It includes common server scripts (which still require implementation - TBD in the following weeks), package support, schema support and common server docs generation.

5.0.0

   - No


Documentation/common_server_docs.py
```diff
@@ -156,18 +157,74 @@ def createPyDocumentation(path, origin, language):
...
+            'origin': origin
+        }
+        function_name = function[0]
+        function_doc['name'] = function_name
+        parameters = function[1].split('.PARAMETER')
+
+        description = parameters[0].split('.DESCRIPTION')[1].strip()
+        if not description:
+            isErrorPS = True
+            print("Missing description for in PS function {}.\n".format(function_name))
```
yaakovi:
```suggestion
            print("Missing description for PS function {}.\n".format(function_name))
```

Tests/scripts/pkg_dev_test_tasks.py
```diff
@@ -209,6 +209,8 @@ def main():
         script_obj = script_obj.get('script')
     script_type = script_obj.get('type')
     if script_type != 'python':
+        if script_type == 'powershell':
```
yaakovi: why is that different from javascript?

yardensade:
Please note that powershell file extension does not have to be .ps1, we need to be careful about that

ronykoz:
@yardensade which file extensions should we support?

Itay4:
We will support only ps1 for now

Tests/scripts/hook_validations/structure.py
```diff
@@ -43,6 +43,7 @@ class StructureValidator(object):
         REPORT_REGEX,
         SCRIPT_PY_REGEX,
         SCRIPT_JS_REGEX,
+        SCRIPT_PS_REGEX,
```
ronykoz: what about INTEGRATION_PS_REGEX

PR #3695, opened 2019-06-11
Fix find similar incidents - dead code 

erezh31:
fixes https://github.com/demisto/etc/issues/17416


erezh31:
@Itay4 is the same code, but export to python (removed dead code)
Also I added a small fix to support dict \ list comparison

PR #3696, opened 2019-06-11
fix unkown charachter issue

jochman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17131

   - No

- [ ] Code Review

### DIFF
--- a/Scripts/ParseEmailFiles/ParseEmailFiles.py
+++ b/Scripts/ParseEmailFiles/ParseEmailFiles.py
@@ -12,7 +12,6 @@ import traceback
 import tempfile
 import sys
 
-
 # -*- coding: utf-8 -*-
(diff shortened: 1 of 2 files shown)


Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -2625,6 +2624,32 @@ def normalize(input_str):
     }
 }
 
+''' HELPER FUNCTION '''
+
+
+def recursive_convert_to_utf8(replace_to_utf):
```
anara123: eventually you convert str to unicode. So the name should be `recursive_convert_to_unicode`

anara123:
fixes https://github.com/demisto/etc/issues/17278

PR #3705, opened 2019-06-13
trying to upload CRLF again

idovandijk:
## Description
I was having problems with the EML file having LF line separators, instead of CRLF.
There were lots of issues with forcing the CRLF change. Trying it now here.
The reason for this is that `Process Email - Generic` has a conditional task with a DT that expects CRLF but not LF, so LF EMLs are not detected as attached emails.

   - No


idovandijk:
Seems it's finally CRLF. I downloaded the file from this PR and ran `file` on it.
image

PR #3708, opened 2019-06-13
Whois bug

reutshal:
## Status
Ready

## Related Issues 
fixes : https://github.com/demisto/etc/issues/17280
fixes : https://github.com/demisto/etc/issues/17281

## Description
* Fixed the two bugs linked
* Updated the command outputs to match the context data
* Updated the test-playbook to validate output paths
* Tested the integration in self-service- contact me for a link

   - Yes
       - All the outputs under Whois.something updated to Domain.Whois.something and now identical 
          to the context data

## Must have
- [ ] Tests
- [x] Code Review


Integrations/Whois/Whois.py
```diff
@@ -6,10 +6,14 @@
 import sys
 from codecs import encode, decode
 
+RETURN_CODE = 0
+ERROR_MESSAGE = ''
+ENTRY_TYPE = entryTypes['error'] if demisto.params().get('with_error', False) else 11  # 11 := entryTypes['warning']
```
ronykoz: any reason for us to not add this to the dict of entryTypes, @anara123 ?

Integrations/Whois/Whois.py
```diff
@@ -8243,14 +8305,18 @@ def whois_command():
             md['Billing Admin'] = contacts['billing']
             ec['BillingAdmin'] = contacts['billing']
     if 'emails' in whois_result:
+        ec['Emails'] = whois_result.get('emails')
```
ronykoz: why the duplicate

Integrations/Whois/Whois.py
```diff
@@ -6,10 +6,14 @@
 import sys
 from codecs import encode, decode
 
+RETURN_CODE = 0
```
ronykoz: wrong usage of CONSTs in python, need to fix this

Integrations/Whois/Whois.py
```diff
@@ -8246,11 +8294,14 @@ def whois_command():
...
+    md['QueryStatus'] = 'Success'
+
     context = ({
-        'Domain': {
-            'Name': domain
+        outputPaths['domain']: {
+            'Name': domain,
+            'Whois': ec
         },
-        'Domain.Whois(val.Name && val.Name == obj.Name)': ec
```
amshamah419: We may still need this (or some form of it) to update existing entries in the context. Though I'm not entirely sure why it wasn't doing so, to begin with.

ronykoz: it's there, note that 
outputPaths['domain'] = 'Domain(val.Name && val.Name == obj.Name)'

or do you mean we should have several entries for whois under domain?

amshamah419: No, that's what I was looking for, just didn't see it. Looks good to me :)

PR #3716, opened 2019-06-16
AWS Athena Beta Integration

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#17515

## Description
Release of Beta-Integration for POC

   - No

## Must have
- [ ] Tests
- [ ] Code Review


yuvalbenshalom:
@amshamah419 why is it in python2? is it necessary?

amshamah419:
@yuvalbenshalom it uses the boto3 docker file which is currently on py2

Itay4:
lets create a new docker image for boto3 which is based on python3
otherwise, this might cause us some headache in the future when we would like to move it from beta
another note: we agreed to use different id for beta integrations, so when we move it to production, we won't have issues installing content, so please change to another id/name

PR #3725, opened 2019-06-18
Esm packagify

yaakovi:
## Status
Ready

## Description
moving to package before fixing some problems.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


yaakovi:
max session error.
will solve this in a separate PR to keep the conversion to package clean

PR #3729, opened 2019-06-18
Extract indicators - Word document check now ignores file info data

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17351

## Description
Removed `File.Info` check for word files in the playbooks `Extract Indicators From File - Generic` after a customer encountered a malicious file that pretended to have file info of doc. The reason I removed it and not changed it to an `and` condition is because while testing the condition I found online some word files that did not have doc or docx file info, as such the field is unreliable.

   - No

## Additional Information:
Manually tested using files provided by @idovandijk 

image

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review


DeanArbel:
Test fails on One Before because of known issues with CommonServerPython and scripts/playbooks/integrations that use the constant urlRegex from the CommonServerPython.

PR #3745, opened 2019-06-22
validate_files: new_added_files logic fix

glicht:
## Status
Ready

## Description
It seems that the `new_added_files` logic removes all new_files from validation (at least those that are fully committed). They are verified during circle build as this happens only during a local validation (pre-commit hooks). 

I've commented this out. Need to better understand why this was added.

### DIFF
--- a/Tests/scripts/validate_files.py
+++ b/Tests/scripts/validate_files.py
@@ -152,12 +152,12 @@ class FilesValidator(object):
             modified_files = modified_files - set(non_committed_deleted_files)
             added_files = added_files - set(non_committed_modified_files) - set(non_committed_deleted_files)
 
-            new_added_files = set([])
-            for added_file in added_files:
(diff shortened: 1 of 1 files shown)


glicht:
@anara123 what do you say?

anara123:
I think it is not working because of previous code `added_files = added_files - set(non_committed_modified_files) - set(non_committed_deleted_files)` 
we remove the `non_commited_modified_files` and then try to get the common group.
I think we added this line after my changes.
@glicht have validated the problematic scenario? And it works with this changes?

glicht:
The scenario of editing a new file works with this fix. But it may introduce other scenarios that don't work.

ronykoz:
@glicht i think we can remove these lines entirely, and then merge this PR

glicht:
I'll merge. Let's follow if people complain...

PR #3756, opened 2019-06-23
EWS Mail Sender - added support for attaching inline images

IkaDemisto:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17330
fixes: https://github.com/demisto/etc/issues/16971

## Description
Added `attachCIDs` argument to `send-mail` command in order to send embedded images (inline images) in outgoing email. 
Example of usage:

1. Upload wanted image to the war room.
2. Run:
`!send-mail subject="Subject Of The Email" to="email_address" htmlBody="<h1>Header Of The Email</h1><h3>The End</h3>" attachCIDs=enrty_id_of_uploaded_file`
3) **cid** in htmlBody must be full name of uploaded file, including file extension.

## Screenshots
Explanation from `exchangelib` library:
exchangelib

   - No

## Must have
- [x] Tests
- [x] Code Review


Integrations/EWSMailSender/EWSMailSender.py
```diff
@@ -126,6 +129,14 @@ def send_email(to, subject, body="", bcc=None, cc=None, replyTo=None, htmlBody=N
         if len(file_entries_for_attachments) != len(attachments_names):
             raise Exception("attachIDs and attachNames lists should be the same length")
 
+    if attachCIDs:
+        file_entries_for_attachments_inline = attachCIDs.split(",")
+        for att_id_inline in file_entries_for_attachments_inline:
+            att_name_inline = demisto.getFilePath(att_id_inline)['name']
+            if isinstance(att_name_inline, list):
```
yaakovi: when did you ever got a list from getFilePath?

Integrations/EWSMailSender/EWSMailSender.py
```diff
@@ -138,6 +149,18 @@ def send_email(to, subject, body="", bcc=None, cc=None, replyTo=None, htmlBody=N
         with open(file_path, 'rb') as f:
             attachments.append(FileAttachment(content=f.read(), name=attachment_name))
 
+    for i in range(0, len(file_entries_for_attachments_inline)):
```
yaakovi: you are not using `i` (if you do need it, use `enumerate`)
```suggestion
    for entry_id, attachment_name_inline in file_entries_for_attachments_inline:
```

Integrations/EWSMailSender/EWSMailSender.py
```diff
@@ -138,6 +149,18 @@ def send_email(to, subject, body="", bcc=None, cc=None, replyTo=None, htmlBody=N
...
+        entry_id = file_entries_for_attachments_inline[i]
+        attachment_name_inline = attachments_names_inline[i]
+        try:
+            res = demisto.getFilePath(entry_id)
+        except Exception as ex:
+            raise Exception("entry %s does not contain a file" % entry_id)
+        file_path = res["path"]
```
yaakovi: why do you separate this loop from the previous one (132-139)? you already run the `getFilePath`

IkaDemisto: I run getFilePath on different lists. The second list includes list with item-ids that will be attached as inline. I append to attachment list with `attachments.append(FileAttachment(content=f.read(), name=attachment_name_inline, is_inline=True, content_id=attachment_name_inline))` and in first loop I append ` attachments.append(FileAttachment(content=f.read(), name=attachment_name))`. Need to separate list with attachIDs and attachCIDs as one is attached as regular attachment and the second is will be attached inline.

Integrations/EWSMailSender/EWSMailSender.py
IkaDemisto: Don't think that it is necessary but tried to save the same convention as in previous block of code. I don't want to change the previous block of code.

Integrations/EWSMailSender/EWSMailSender.py
IkaDemisto: Same reason here. If we change to enumerator, it's better to change also in first for loop.
Basically the main reason is not to break backward.

IkaDemisto:
email
@yaakovi note that you can still attach files as regular attachment, additionally to inline image.

Integrations/EWSMailSender/EWSMailSender.py
yaakovi: breaking backward is not an issue here, since the result is the same.
this why is just more pythonic while the other is the C-way

Integrations/EWSMailSender/EWSMailSender.py
yaakovi: weird... but seems legit

Integrations/EWSMailSender/EWSMailSender.py
yaakovi: you are looking at different loop. look at the lines i specified.

IkaDemisto:
@yaakovi added support to array input also.

kirbles19:
Done editing .yml and RN. Check my commit message for the RN.

PR #3770, opened 2019-06-26
RTIR - Added referer header parameter

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17615

## Description
Added new optional paremeter - Referer: Adds a referer header to the requests sent by the integration

   - No

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review


Itay4:
dont really understand why would a user want to fill in the referer..
if it works with empty and covers the issue, lets just always pass it empty

DeanArbel:
> dont really understand why would a user want to fill in the referer..
> if it works with empty and covers the issue, lets just always pass it empty

That's the thing, it shouldn't be empty - it's used by the RTIR system to prevent CSRF - basically if this protection is turned on, then it checks the referer header to see if the site is in its whitelist  (it's a pretty lame protection but that's how they work).

In the issue RTIR returned an error for not having the header, however they can also return an error if the header is in the request, but its value is not in the whitelist...

Integrations/RTIR/RTIR.yml
```diff
@@ -569,3 +573,4 @@ script:
   type: python
 tests:
 - RTIR Test
+releaseNotes: "Added new optional paremeter - Referer: Adds a referer header to the requests sent by the integration."
```
Itay4: rn should be in a seperate file

Itay4:
@kirbles19 please review the added parameter in the yml and the detailed instructions in the md file

kirbles19:
@DeanArbel All files reviewed.

Itay4:
@kirbles19 i think we should update the display name of the parameter, as it is not clear enough to me, from the current display name (Referrer), what this parameter is used for. wdyt?

Integrations/RTIR/RTIR.yml
```diff
@@ -44,6 +44,10 @@ configuration:
   name: fetch_queue
   required: false
   type: 0
+- display: Referer
```
Itay4:
```suggestion
- display: Referer request header
```

DeanArbel: 👍

PR #3771, opened 2019-06-26
add WF detonate URL PB

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16548

4.0.0

   - No

## Must have
- [x] Tests
- [x] Documentation - after CR
- [x] Code Review


Playbooks/playbook-Detonate_URL_-_WildFire-v2.yml
```diff
@@ -0,0 +1,475 @@
+id: Detonate URL - WildFire-v2
+version: -1
+name: Detonate URL - WildFire-v2
+fromversion: 4.0.0
```
yuvalbenshalom: is that correct?

kirbles19:
Documentation updated with PBs and new param

PR #3777, opened 2019-06-27
Sentinel One V2

Shellyber:
## Status
In Progress/

## Related Issues
fixes: https://github.com/demisto/etc/issues/13292 https://github.com/demisto/etc/issues/17605 https://github.com/demisto/etc/issues/16547

   - No

## Must have
- [ ] Code Review


Tests/secrets_white_list.json
```diff
@@ -718,6 +718,7 @@
       "te.checkpoint.com",
       "api.secureworks.com",
       "www.demisto.com",
+      "https://usea1.sentinelone.net",
```
yardensade:
```suggestion
      "usea1.sentinelone.net",
```

yardensade:
@kirbles19 This is going to beta integration so won't have official docs but we still want to release it after a proper review since it will probably be released at some point.

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    Gets the group data.
+    """
+
+    context = {}
+    contents = []
+    headers = ['ID', 'Name', 'Type', 'Creator', 'Creator ID', 'Created at', 'Rank']
+
+    group_type = demisto.args().get('type')
+    group_id = demisto.args().get('id')
+    group_ids = argToList(demisto.args().get('groupIds', []))
```
yardensade:
```suggestion
    group_ids = argToList(demisto.args().get('group_ids', []))
```

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    is_default = demisto.args().get('isDefault')
```
yardensade:
```suggestion
    is_default = demisto.args().get('is_default')
```

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+
+
+def delete_group():
+    """
+    Deletes a group by ID.
+    """
+    group_id = demisto.args().get('group_id')
+
+    delete_group_request(group_id)
+    demisto.results('The group was deleted successfully')
```
yardensade: Are we sure there is no situation where you don't get error in response yet the call still fails?
If our only indication for success is 200 response then you can resolve.

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    }
+
+    context['SentinelOne.Agents(val.Date && val.Date === obj.Date)'] = context_entries
+
+    demisto.results({
+        'Type': entryTypes['note'],
+        'ContentsFormat': formats['json'],
+        'Contents': contents,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown('Sentinel One - Shutdown Agents \n' + 'Total of: ' + str(
```
yardensade: Remove `-` like in rest of commands

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+                'Pid': process.get('pid')
+            })
+        context['SentinelOne.Agents(val.Pid && val.Pid === obj.Pid)'] = processes
+
+    demisto.results({
+        'Type': entryTypes['note'],
+        'ContentsFormat': formats['json'],
+        'Contents': contents,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown('Sentinel One Agent processes', contents, headers, removeNull=True),
```
yardensade:
```suggestion
        'HumanReadable': tableToMarkdown('Sentinel One Agent Processes', contents, headers, removeNull=True),
```

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+
+def get_threats_command():
+    """
+    Gets a list of threats.
+    """
+    # Init main vars
+    contents = []
+    context = {}
+    context_entries = []
+    title = ''
```
yardensade: Seems like there are functions that use title and functions that don't, should go either way - either use title everywhere or static title everywhere.

yardensade: In this case if no result there will be no title at all

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    Apply a mitigation action to a group of threats
+    """
+    # Init main vars
+    headers = ['ID', 'Mitigation Action', 'Mitigated']
+    contents = []
+    context = {}
+    context_entries = []
+
+    # Get arguments
+    threat_ids = argToList(demisto.args().get('threat_ids'))
```
yardensade: should have a default empty list

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    Mark suspicious threats as threats
+    """
+    # Init main vars
+    headers = ['ID', 'Marked As Threat']
+    contents = []
+    context = {}
+    context_entries = []
+
+    # Get arguments
+    threat_ids = argToList(demisto.args().get('threat_ids'))
```
yardensade: should have a default empty list

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    """
+    # Init main vars
+    contents = []
+    context = {}
+    context_entries = []
+    title = ''
+
+    # Get arguments
+    content_hash = demisto.args().get('content_hash')
+    mitigation_status = argToList(demisto.args().get('mitigation_status'))
```
yardensade: should have a default empty list

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+        return response.get('data')
+    return {}
+
+
+def move_agent_to_group_command():
+    """
+    Move agents to a new group.
+    """
+    group_id = demisto.args().get('group_id')
+    agents_id = argToList(demisto.args().get('agents_ids'))
```
yardensade: should have a default empty list

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+        'Type': entryTypes['note'],
+        'ContentsFormat': formats['json'],
+        'Contents': contents,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, contents, headers, removeNull=True),
+        'EntryContext': context
+    })
+
+
+def get_static_indicators_request():
```
yardensade: seems like this request is static, what happens if there are many results? will you get all results no matter what? isn't there a way to filter the results? (or atleast limit them)

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+        "createdAt": created_at,
+        "limit": limit,
+        "siteIds": site_ids
+    }
+
+    response = http_request('GET', endpoint_url, params)
+    if response.get('errors'):
+        return_error(response.get('errors'))
+    if 'data' in response:
+        return response.get('data').get('sites'), response.get('data').get('allSites')
```
yardensade: should probably add check if data is a dict before doing `.get`, just to be on the safe side

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+    incidents = []
+    threats = get_threats_request()
+    for threat in threats:
+        # If no fetch threat rank is provided, bring everything, else only fetch above the threshold
+        if not FETCH_THREAT_RANK or (FETCH_THREAT_RANK and threat.get('rank') >= FETCH_THREAT_RANK):
+            incident = threat_to_incident(threat)
+            incident_date = date_to_timestamp(incident['occurred'], '%Y-%m-%dT%H:%M:%S.%fZ')
+            # update last run
+            if incident_date > last_fetch:
+                last_fetch = incident_date
```
yardensade: we don't compare only last_fetch anymore, but take another var which is "max" time of events being fetched, then compare if current_fetch > max_fetch and update it accordingly and pass it to setLastRun. Can see it in any new integration with fetch, it is the same in most integrations.

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -0,0 +1,1648 @@
...
+def http_request(method, url_suffix, params={}, data=None):
+    res = requests.request(
+        method,
+        BASE_URL + url_suffix,
+        verify=USE_SSL,
+        params=params,
+        data=data,
+        headers=HEADERS
+    )
+
```
yardensade: Might be nice to log here the API request being sent, like we log the commad before the flow of the code starts (see before switch case), just don't forget to not log HEADERS var as it contains sensitive info.

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.py
Shellyber: I'm removing this command since the data is irrelevant. Discussed with Maya about this.

Beta_Integrations/SentinelOne-V2/SentinelOne-V2.yml
```diff
@@ -0,0 +1,1264 @@
...
+    deprecated: false
+    description: Get all Deep Visibility events from query
+    execution: false
+    name: sentinelone-get-events
+  dockerimage: demisto/python3:3.7.3.286
+  isfetch: true
+  runonce: false
+  script: '-'
+  type: python
+releaseNotes: New integration - SentinelOne V2
```
yardensade: we don't use releaseNotes in yml anymore, but use external file under release notes folder. Should see @yaakovi 's explanation on slack for more info.

yardensade: Also we should make sure we explain to users this is beta and if any known issue we know about in advance, to prevent confusion.

yardensade:
Only standing issue we have right now is context outputs, in some of them it is ok but in some of them it is `SentinelOne.Activity.agentId` while it should be `SentinelOne.Activity.AgentID`, this happens in too much places due to refactor so I didn't note each one

Itay4:
@kirbles19 please review the yml and the md files

PR #3783, opened 2019-06-27
Qradar fetch-incidents fix

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17709 (tested)
fixes: https://github.com/demisto/etc/issues/18186 (tested)

## Description
* Fixed a bug where sometimes fetch-incidents would fail due to a non-ASCII character in the data.
* Fixed a bug where fetch-incident would ignore filter if the max amount of offenses set in the instance configuration were fetched in a single fetch.
* Improved fetch-incidents error handling

   - No

## Must have
- [x] Tests
- [x] Documentation


DeanArbel:
@kirbles19 please review the detailed instructions in the md file

kirbles19:
@DeanArbel .md updated.

DeanArbel:
PR is failing since we lost our license to QRadar and need to renew it.

DeanArbel:
Earlier today I had a session with a customer who encountered this bug. He provided me with the raw-response of the offense that is causing the issue, and thanks to it I was able to set up a mock server to test out the fix, here are the results:
image

image

DeanArbel:
I was able to test out a fix using a mock server that returns an incident that raises this error:
image
image

DeanArbel:
I was able to test out this version of the integration with a customer who experienced this bug, and confirmed that the fix works.
Since there is no instance for the test, the build fails at circle ci.

yuvalbenshalom:
@glicht or @anara123 please force marge.
we don't have an instance but it was tested on a customer env.

PR #3789, opened 2019-06-29
Pwned - Bug fix in Reason field in the context

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17673

   - Yes
current context output had not only wrong data, but the data was going to a different path, which was not mentioned. so I edited the exisiting always empty context field to the correct one.
line 274 in yml file

## Must have
- [x] Tests
- [x] Documentation - https://github.com/demisto/etc/issues/17764
- [x] Code Review


Integrations/integration-Pwned.yml
```diff
@@ -182,7 +189,7 @@ script:
     - contextPath: Account.Email.Compromised.Vendor
       description: For comporomised emails, the vendor that made the decision
       type: string
-    - contextPath: Account.Email.Reason
+    - contextPath: Account.Email.Compromised.Reason
```
yuvalbenshalom: since we are changing the path anyway, maybe we can add a more meaningful name than reason?
or is it the agreed convention for us to say "who found this?"
@kirbles19 lets check this.

yuvalbenshalom: @Arsenikr proposed to use "reporters"

bakatzir:
@yuvalbenshalom this is break of BC, so if approved will need to force merge

Integrations/integration-Pwned.yml
kirbles19: Still relevant?

Integrations/integration-Pwned.yml
```diff
@@ -19,25 +19,24 @@ configuration:
...
 
-    var sendRequest = function(url) {
+    function sendRequest(url) {
         var res = http(
             url,
             {
                 Method: 'GET',
                 Headers: {
                     'Accept': ['application/json'],
-                    'User-Agent' : ['DBOT-API'],
```
yuvalbenshalom: why did you remove?

Integrations/integration-Pwned.yml
bakatzir: @kirbles19  no ty, we're good

Integrations/integration-Pwned.yml
bakatzir: by mistake. reverting.

bakatzir:
@kirbles19 see .yml:
https://github.com/demisto/content/edit/pwned_domains_fix/Integrations/integration-Pwned.yml?pr=%2Fdemisto%2Fcontent%2Fpull%2F3789

kirbles19:
@bakatzir done reviewing .yml

PR #3799, opened 2019-07-01
packagify extractIndicators

DeanArbel:
## Status
Ready

## Description
packagify extractIndicators in preparation for another issue.


yuvalbenshalom:
@DeanArbel you should add a release note in change to package

DeanArbel:
> @DeanArbel you should add a release note in change to package

Added a blank RN. I think packagifying is an internal change and shouldn't be in RN. I will add a proper RN after this PR is merged (as I'll add in a fix).

On another note, seems there are tests failing. I'll try to fix them with the help of @idovandijk

PR #3800, opened 2019-07-01
Find similar incidents fix - support list in context

erezh31:


Scripts/FindSimilarIncidents/FindSimilarIncidents.py
```diff
@@ -85,8 +85,6 @@ def nested_dict_flatted(d, parent_key='', sep='.'):
         items = []  # type: ignore
         for k, v in d.items():
             new_key = parent_key + sep + k if parent_key else k
-            if isinstance(v, list) and len(v) > 0:
```
anara123: why do you remove this?

anara123:
is there related issue?

erezh31:
No issue, came from a customer

Scripts/FindSimilarIncidents/FindSimilarIncidents.py
erezh31: to support list comparison - now it's consider only the fist element if it's list - and it's not a good enough logic for some use-cases

PR #3803, opened 2019-07-02
Revert "Powershell content support (#3691)"

Itay4:
This reverts commit cc5058c7


yaakovi:
pull from master to fix RN problem

PR #3809, opened 2019-07-02
Bugfix: SendEmailToManager

wjsetzer:
SendEmailToManager previously did not do anything because the result of executing the AD search does not parse the attributes to JSON and leaves them as newline-separated

Changes:
Moved imports to top.
Added explicit dict contains checks.
Removed demisto.get to prevent silent errors from occurring when accessing dicts
Added early returns for errors.
Annotated errors to be error type.

## Status
Ready

## Description
Bugfix to SendEmailToManager

branch | PR
------ | ------

   - Possibly. Some Python 3 features and syntax used, including using Python 3's map, which is a generator and does not perform eager evaluation and Python 3 f-strings for string interpolation.

### DIFF
--- a/Scripts/script-SendEmailToManager.yml
+++ b/Scripts/script-SendEmailToManager.yml
@@ -2,84 +2,167 @@ commonfields:
   id: SendEmailToManager
   version: 1
 name: SendEmailToManager
-fromversion: 3.5.0
 script: |-
-  email = demisto.get(demisto.args(), 'email')
-  if not email:
-      for t in demisto.incidents()[0]['labels']:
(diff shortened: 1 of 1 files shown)


Itay4:
@wjsetzer Thanks for the contribution !
Sorry for the late reply.
As you've mentioned, you are using Python 3 features and syntax, and this script runs using Python 2 docker image, so this won't work.
Can you please rewrite the Python 3 parts in Python 2 ?

Itay4:
@wjsetzer Closing due to inactivity.
You are more than welcome to reopen if you wish.

PR #3811, opened 2019-07-03
fix bug in lr-get-alarm-events-by-id command

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17739

## Description
The command !lr-get-alarm-events-by-id crash when there was no events for the alarm


kirbles19:
Reviewed the RN.

PR #3822, opened 2019-07-07
[Bug] Rich Text Files are not inspected with 'Extract Indicators From File - Generic v2'

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17655

## Description
Fixed 
* playbook to handle a bug where `.rtf` files are not inspected with `Extract Indicators From File - Generic v2`.
* bug in `ExtractIndicatorsFromTextFile` script where certain strings had their escape characters removed, which created impossible characters (which threw an exception).
* Follow up of a previous bug that I noticed while working on this - in https://github.com/demisto/etc/issues/17351. (look for the next part in the comments).
Though there is a condition prior to this task that was fixed, I realised that in a multi-file scenario this task might still be reached, and so this task had to be changed as well

4.1.0+

   - No

#### Playbook:
image

#### Script (before + after):
image

#### Playbook:
image

#### Script (before + after):
image

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review

### DIFF
--- a/Playbooks/playbook-Extract_Indicators_From_File_-_Generic_v2.yml
+++ b/Playbooks/playbook-Extract_Indicators_From_File_-_Generic_v2.yml
@@ -13,13 +13,13 @@ starttaskid: "0"
 tasks:
   "0":
     id: "0"
-    taskid: 7c475fe8-ef6f-4203-8649-88c8ee1355a4
+    taskid: 7820586b-f2e6-4673-8dd8-08caab0d1173
     type: start
     task:
-      id: 7c475fe8-ef6f-4203-8649-88c8ee1355a4
+      id: 7820586b-f2e6-4673-8dd8-08caab0d1173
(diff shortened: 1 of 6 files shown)


Playbooks/playbook-Extract_Indicators_From_File_-_Generic_v2.yml
```diff
@@ -632,22 +648,6 @@ tasks:
               right:
                 value:
                   simple: Microsoft Word
-            - operator: isEqualString
```
DeanArbel: This part is a follow up of a previous bug that I noticed while working on this - in https://github.com/demisto/etc/issues/17351.
Though there is a condition prior to this task that was fixed, I realised that in a multi-file scenario this task might still be reached, and so this task had to be changed as well

PR #3829, opened 2019-07-07
Remove releaseNotes label in schemas

yaakovi:
## Status
Ready

## Description
remove releaseNotes label from content schemas to avoid confusion between new format and old format.
test only fails on "AttivoBotsink" conversion to package (not converting on purpose).

   - No

## Must have
- [ ] Tests
- [ ] Code Review


yaakovi:
@yuvalbenshalom 
please also review this, this is going to be forced-pushed because of the package.

yaakovi:
force-pushed because i don't want to convert `AttivoBotsink` to a package.
we don't have an instance for testing the conversion.

PR #3835, opened 2019-07-09
draft RN replace xxxxx with actual asset ID

yardensade:
## Status
Ready


Itay4:
lets do the replacement in `get_release_notes_draft` function

yaakovi:
we should also update the `add_dot` function and better catch ignored RN

PR #3843, opened 2019-07-10
deprecate site review integration

ronykoz:
deprecating the site review integration


yaakovi:
removing an integration causes BC errors.

PR #3845, opened 2019-07-10
[Bug] ExtractIndicatorsFromWordFile: docker run with non-root user

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17100

## Description
Added export for HOME.

### Error example:
Screen Shot 2019-07-10 at 13 37 27

### Extract on a file that causes the error (error on top, fix on bottom):
image

   - No

## Must have
- [x] Tests - testing is done manually, because it only fails when user is not admin
- [ ] Code Review


glicht:
Let's add a unit test

TestPlaybooks/playbook-Process_Email_-_Generic_-_Test.yml
```diff
@@ -304,7 +304,7 @@ tasks:
         simple: "yes"
       unsecure: {}
       url:
-        simple: https://raw.githubusercontent.com/demisto/content/master/TestData/docwithindicators.doc
+        simple: https://raw.githubusercontent.com/demisto/content/master/TestData/docwithindicators
```
glicht: Why did you remove the `.doc`. Seems like a bug

TestPlaybooks/playbook-Extract_Indicators_From_File_-_test_4_0.yml
```diff
@@ -275,7 +275,7 @@ tasks:
         simple: "yes"
       unsecure: {}
       url:
-        simple: https://raw.githubusercontent.com/demisto/content/master/TestData/docxwithindicators.docx
+        simple: https://raw.githubusercontent.com/demisto/content/master/TestData/docxwithindicators
```
glicht: Seem like a bug. I think it should be with `.doc` ending

TestPlaybooks/playbook-Process_Email_-_Generic_-_Test.yml
DeanArbel: this was done by mistake (refactored another item by this name, and this happened automatically). Fixing it

TestPlaybooks/playbook-Extract_Indicators_From_File_-_test_4_0.yml
DeanArbel: Done by mistake. see https://github.com/demisto/content/pull/3845#discussion_r303747258

PR #3853, opened 2019-07-11
[Bug] Nexpose nexpose-get-assets returns a date format error

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17765

## Description
The issue was caused by a known behaviour of `time.strptime` function where if given the flag `%f` (milliseconds) when milliseconds is 0 will result in an exception.
The fix is to wrap the call with a function that contains a fallback if there aren't milliseconds.
   - No

### Unit test result on integration prior to fix:
image

## Must have
- [x] Tests
- [x] Code Review

### DIFF
--- a/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
+++ b/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
@@ -225,7 +225,7 @@ def get_list_response(path, method='get', limit=None, body={}, params={}):
 def get_last_scan(asset):
     if asset['history'] is None:
         return "-"
-    sorted_dates = sorted(asset['history'], key=lambda h: time.strptime(h['date'], "%Y-%m-%dT%H:%M:%S.%fZ"),
+    sorted_dates = sorted(asset['history'], key=get_datetime_from_asset_history_item,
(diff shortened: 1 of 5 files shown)


Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -225,7 +225,7 @@ def get_list_response(path, method='get', limit=None, body={}, params={}):
 def get_last_scan(asset):
     if asset['history'] is None:
         return "-"
-    sorted_dates = sorted(asset['history'], key=lambda h: time.strptime(h['date'], "%Y-%m-%dT%H:%M:%S.%fZ"),
+    sorted_dates = sorted(asset['history'], key=get_datetime_from_asset_history_item,
```
yuvalbenshalom: will the sorting work correctly if some of the strings are with milliseconds and some aren't?
please test this.

DeanArbel: I'll add a test for it, however it shouldn't affect the sort, as datetime is sorted by its timestamp

PR #3872, opened 2019-07-16
[Bug] Jask fixed issue with time parsing

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16655

## Description
Fixes issue with time parsing fetch-incidents.

### Test failing prior to bug fix:
image

   - No

## Must have
- [x] Tests
- [x] Code Review


DeanArbel:
The change was done to `fetch-incidents` however there is no risk to break BC, as the change comes only in case the original flow fails.

PR #3880, opened 2019-07-17
Circl ssl fix

yaakovi:
## Status
Ready

## Description
SSL parameter was ignored.
removed from unmockable


Integrations/CIRCL/CIRCL.py
```diff
@@ -13,22 +12,17 @@
 USERNAME = demisto.getParam('credentials')['identifier']
 PASSWORD = demisto.getParam('credentials')['password']
 AUTH = (USERNAME, PASSWORD)
-USE_SSL = True if demisto.params().get('insecure') else False
+USE_SSL = demisto.params().get('insecure', False)
```
ronykoz:
```suggestion
USE_SSL = not demisto.params().get('insecure', False)
```

yaakovi: good catch

PR #3882, opened 2019-07-17
Adding new output and removing the attachments one

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17892

## Description
adding a new output, and removing and old one(still in the context so not really breaking BC)


ronykoz:
@anara123 not sure what you mean, it should be covered by the code I wrote, or am I missing something?

Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3505,6 +3505,7 @@ def handle_eml(file_path, b64=False, file_name=None, parse_only_headers=False, m
                 'Headers': header_list,
                 'HeadersMap': headers_map,
                 'Attachments': ','.join(attachment_names) if attachment_names else '',
+                'AttachmentNames': attachment_names if attachment_names else [],
```
anara123: here you are adding new output, but in the context outputs you are replacing Attachments via AttachmentNames. It will break backward compatibility

Scripts/ParseEmailFiles/ParseEmailFiles.yml
```diff
@@ -85,7 +85,7 @@ outputs:
 - contextPath: Email.HeadersMap.Message-ID
   description: A unique string assigned by the mail system when the message is first created. These can easily be forged. (e.g. 5c530c1b.1c69fb81.bd826.0eff@mx.google.com)
   type: String
-- contextPath: Email.Attachments
+- contextPath: Email.AttachmentNames
```
anara123: Don't replace, but add new output. You can deprecate the old one

Scripts/ParseEmailFiles/ParseEmailFiles.py
ronykoz: It will not break BC because i still populate the old path, just stop showing it in the outputs, there is no need for the changes you've sugested

ronykoz:
@kirbles19 could you please review the changelog.md file?

richardbluestone:
Hi, I've reviewed the changelog.md file.

ronykoz:
@anara123 i need force merge here, the build passed but flagged the BC break

PR #3885, opened 2019-07-17
Opswat metadefender bug fix: corrupted file upload

yuvalbenshalom:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17189

## Description
Fixes the curated file upload by *not* using muiltipart
Handling responses when `total_detected_avs` is missing.

the one dose not have tests because we don't have an instance.
It was tested in the customer environment. this can be seen in the issue.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Itay4:
@yuvalbenshalom add test

PR #3892, opened 2019-07-18
General proxy fixes

yaakovi:
## Status
Ready

## Related Issues
relevant issue: https://github.com/demisto/etc/issues/15544

## Description
fixed proxy settings of several integrations (more fixes will be done in another PR to keep them simple). in addition, fixed some bugs in the mock mechanism.

## Related PRs
https://github.com/demisto/content/pull/3877

   - Yes
       - Further details: fix bug in `insecure` parameter logics in `Cisco Umbrella`, `MaxMind GeoIP2`, `UrlScan`
the logic was reversed

## Must have
- [ ] Tests
- [ ] Code Review

## Additional changes


yuvalbenshalom:
@yaakovi I don't think that it is accurate to state that the change in umbrella is breaking backward compatibly because you are preserving the old(and bad) functionality.

Integrations/Cisco-umbrella/Cisco-umbrella.py
```diff
@@ -21,8 +21,7 @@
 
 API_TOKEN = demisto.params()['APIToken']
 BASE_URL = demisto.params()['baseURL']
-USE_SSL = demisto.params().get('insecure') if demisto.params().get('insecure') else \
-    not demisto.params().get('insecure_new')  # Backward compatibility issue, the old logic of insecure was reversed
```
yuvalbenshalom: when did decide to remove this?
if we do, we have other integrations that should be changed.

PR #3894, opened 2019-07-19
Revert fqdn indicator and skip TPB

bakatzir:
## Status
Ready

   - No


bakatzir:
@omercnet FYI
@glicht has a few comments - let's discuss in issue if needed not in PR.
I'm ooo this week so we'll push this for next release.

PR #3904, opened 2019-07-22
MailSenderNew add support for custom headers

pingunaut:
## Status
Ready

## Description
This PR adds support for custom mail headers to the MailSenderNew integration. Users can simply add additional one or multiple mail headers to an email by providing `headerName1=headerName1,headerName2=headerName2`

4.5.0

   - No

## Must have
- [ ] Tests
- [x] Documentation: provided within the command description
- [ ] Code Review

### DIFF
--- a/Integrations/MailSenderNew/MailSenderNew.py
+++ b/Integrations/MailSenderNew/MailSenderNew.py
@@ -223,6 +223,7 @@ def create_msg():
     to = argToList(demisto.getArg('to'))
     cc = argToList(demisto.getArg('cc'))
     bcc = argToList(demisto.getArg('bcc'))
+    additional_header = argToList(demisto.getArg('additionalHeader'))
     subject = demisto.getArg('subject') or ''
     body = demisto.getArg('body') or ''
     htmlBody = demisto.getArg('htmlBody') or ''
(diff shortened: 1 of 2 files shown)


pingunaut:
image

Headers are then added to the email:
image

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -286,6 +287,10 @@ def create_msg():
         msg['To'] = header(','.join(to))
     if cc:
         msg['CC'] = header(','.join(cc))
+    if additional_header:
+        for h in additional_header:
+            header_name_value = h.split('=')
+            msg.add_header(header_name_value[0], header(header_name_value[1]))
```
avidan-H: Your changes weren't working for me. I think you have a small bug here. I think it should be 
```
msg.add_header(header_name_value[0], header_name_value[1])
```

pingunaut: You are correct. I wanted to take advantage of the `header` function which is able to handle new line characters correctly. But unfortunately it returns a `Header` object which is not supposed to be added as a parameter here. 
I'll update the code accordingly.

pingunaut: @avidan-H Thanks for your hint. PR is updated to add the headers using the `msg` dict directly instead of the `add_header` function (like it's done for the existing headers also).

avidan-H: awesome 👍

PR #3911, opened 2019-07-22
Fixed Active Directory V2 default query & fixed DN bug

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17889

## Description
Fixed two things:
1. Active Directory Query V2 default query
2. Active Directory Query V2 usage of DN argument

   - No

## Must have
- [ ] Code Review


glicht:
@guykeller  build is still failing on AD test

guykeller:
@glicht Just merging a fix by @roysagi for that, then re-running the workflow. That's what I've waited for.

PR #3913, opened 2019-07-22
JSON File to CSV

yaakovi:
## Status
Ready

## Description
as a result of an external PR: https://github.com/demisto/content/pull/3895
in addition, fixed the delimiter argument behavior.
did not separated the packagify from the actual changes since most of the changes to the integration were as a result of the packagify process.

fix delimiter here: https://github.com/demisto/content/pull/3913/commits/d5ab90021211c1788cea08f1494730a6a4169a58

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes


Scripts/JSONFileToCSV/JSONFileToCSV.py
```diff
@@ -0,0 +1,54 @@
...
+    dictlist = json.loads(data)
+
+    csv_out = json_to_csv(dictlist, delimiter)
+
+    # output cvs as a file to war-room
+    demisto.results(fileResult(out_filename, csv_out.encode("utf-8")))
+
+
+if __name__ in ('__builtin__', 'builtins'):
+    args = demisto.args()
```
avidan-H: Stylistically speaking it's a little strange to me that you are calling `main()` with arguments - as it is conventionally without any arguments. I would just call `demisto.args()` from inside the main function.

Scripts/JSONFileToCSV/JSONFileToCSV.py
```diff
@@ -0,0 +1,54 @@
...
+    for d in data:
+        val_lst = []
+        for k in keys:
+            val_lst.append(d[k])
+        cw.writerow(val_lst)
+    return si.getvalue().strip("\r\n")
+
+
+def main(entry_id, out_filename, delimiter):
+    if isinstance(entry_id, list):
```
avidan-H: need to call `argToList` on entry_id, no?

Scripts/JSONFileToCSV/JSONFileToCSV.py
```diff
@@ -0,0 +1,54 @@
...
+import io
+import csv
+
+
+def json_to_csv(data, delimiter):
+    si = io.BytesIO()
+    cw = csv.writer(si, delimiter=delimiter)
+    keys = list(data[0].keys())
+    cw.writerow(keys)
+    for d in data:
```
avidan-H: couldn't you simplify these two `for` loops into one `for` loop using `for key, val in data.items():`?

Scripts/JSONFileToCSV/JSONFileToCSV.py
yaakovi: calling `demisto.args()` from the executed code section only allow writing unit test for the main function without the need for mocking it.

Scripts/JSONFileToCSV/JSONFileToCSV.py
yaakovi: actually this can be removed.
since the `is_array` flag is not set, entry_id will never be a list

Scripts/JSONFileToCSV/JSONFileToCSV.py
yaakovi: i don't think so. think of `data` as a two-dimentional matrix.
data is a list of object.
each object correspond a line in the final csv file (hence the loop over the data record).
for each line, we collect the relevant fields (by the `keys` variable) into a list to append as a csv row.

avidan-H: I see you still simplified it though ✓

PR #3930, opened 2019-07-25
Unified and streamlined indicator types version

DeanArbel:
## Status
In Progress (pending manual tests)

## Related Issues
related to: https://github.com/demisto/etc/issues/18105 (doesn't fix, the issue, but lowers the severity of it substantially).

**Note:** If the related issue is fixed, then this PR can be closed, as it's just a workaround.

## Description
This PR should improve the way we handle the new indicator types across demisto versions.


DeanArbel:
@yuvalbenshalom you're right, we can close this

PR #3931, opened 2019-07-25
Fixed test CheckPoint FW

Shellyber:
## Status
Ready

   - No


yardensade:
Lack details, just to make sure, are we basically reverting last week's changes where a command was added to the test PB because the service does not work well and basically this command is untestable at the moment?

PR #3934, opened 2019-07-25
Enable proxy and OAuth in Vectra

amshamah419:
## Status
In Hold(Verifying bugfix with customers)

## Related Issues
related to: demisto/etc#15183
related to: demisto/etc#14414

## Description
Enabled use of system proxy as well as OAuth tokens.

   - Yes
       - Further details: Basic auth has been deprecated and tokens are the preferred auth method.

## Must have
- [ ] Tests (We don't have an instance)
- [ ] Code Review

### DIFF
--- a/Integrations/integration-Vectra.yml
+++ b/Integrations/integration-Vectra.yml
@@ -12,11 +12,6 @@ configuration:
   defaultvalue: ""
   type: 0
   required: true
-- display: Credentials
-  name: credentials
-  defaultvalue: ""
-  type: 9
-  required: true
 - display: Fetch incidents
   name: isFetch
   defaultvalue: ""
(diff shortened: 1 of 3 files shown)


amshamah419:
@GavrielFilippov @AndyDemisto Since we do not have an instance, could you guys have the customer verify that the integration works for them? If we have confirmation, we can merge these changes.

Integrations/integration-Vectra.yml
```diff
@@ -56,20 +63,20 @@ script:
...
-        'vectra-health':'/api/health/',
-        'vec-get-detetctions-by-id':'/api/detections/', //deprecated
-        'vec-detections': '/api/detections/',   //deprecated
-        'vec-hosts':'/api/hosts',               //deprecated
-        'vec-settings':'/api/settings',         //deprecated
-        'vec-triage':'/api/rules',              //deprecated
-        'vec-sensors':'/api/sensors',           //deprecated
-        'vec-health':'/api/health/',            //deprecated
-        'vec-get-host-by-id':'/api/hosts/',     //deprecate
+        'vectra-detections': '/api/v2/detections/',
```
yuvalbenshalom: why are we using v2 api here? is that connected to the fix?

Itay4:
@yuvalbenshalom whats our status here?

amshamah419:
@Itay4 @yuvalbenshalom We may need to close out this PR. We tried doing the simple fixes with the customer, but there seems to be a major discrepancy between V1 and V2. Since we don't have an instance, we are stalled at the moment.

amshamah419:
Going to close out this PR since we really need an instance to get the fixes done.

PR #3940, opened 2019-07-25
fix health dashboard

Spivaka:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18218

## Description
Fix default health dashboard to support disk partitions

## Screenshots
image

5.0.0

Not that I know of, as the content is overriden in the new version

### DIFF
--- a/Dashboards/dashboard-SystemHealth-old.json
+++ b/Dashboards/dashboard-SystemHealth-old.json
@@ -1,7 +1,8 @@
 {
 	"id": "health-status-dashboard",
     "version": -1,
-	"fromVersion": "5.0.0",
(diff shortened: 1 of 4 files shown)


Bars92:
can we automaticlly do in server `disk.usedPercent` match `disk.usedPercent./` in 5.0.0? that should save the need for old yaml

Spivaka:
we can, but I think adding patches in the server is less preferred than changing the content...

idobloch:
isn't it already fixed in: https://github.com/demisto/server/pull/13214 ?

Itay4:
@Spivaka whats our status here?

PR #3945, opened 2019-07-28
Moved Fortigate to Package format

guykeller:
## Status 
Ready

## Description
Moved Fortigate to Package format

   - No


Itay4:
please rename to `FortiGate`

guykeller:
Why renaming to FortiGate?
It was "Fortigate" way before I added new content.
Moreover - this PR is dedicated strictly to moving the integration to a package, which as far as I know should be named like the integration was before

Integrations/FortiGate/FortiGate.py
```diff
@@ -0,0 +1,1025 @@
+import demistomock as demisto
+from CommonServerPython import *
+''' IMPORTS '''
+
+import json
+import requests
+from typing import *
```
yaakovi: why?

guykeller: removed

PR #3952, opened 2019-07-28
New layout and fields

yaakovi:
## Status
Ready

## Description
all of the following were created by @JonathanMeler 
* new phishing layout.
* new script
* new incident field.
* updated relevant classifiers with the new field.

5.0.0 - for new layout

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Layouts/layout-details-Phishing.json
```diff
@@ -1,347 +1,755 @@
 {
-	"typeId": "Phishing",
+	"TypeName": "Phishing",
+    "fromVersion": "5.0.0",
```
asafshen: we should add new layout files, not replacing the existing ones - since we still have users with 4.5 - we don't want to delete existing layouts

yaakovi: i'm not deleting, the old one moved to be under `Layouts/layout-details-Phishing_4_5_0.json`

asafshen: 😅

yaakovi:
@JonathanMeler 
what is the status here?

PR #3954, opened 2019-07-28
AF - handle trying to retrieve pending query

bakatzir:
## Status
waiting for product writer - see in related issue

## Related Issues
fixes: https://github.com/demisto/etc/issues/18226

   - No

## Must have
- [x] Tests 
- [x] Documentation - not needed
- [ ] Code Review


Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -185,6 +185,9 @@ def parse_response(resp, err_operation):
         if res_json.get("message").find('Requested sample not found') != -1:
             demisto.results(err_msg)
             sys.exit(0)
+        elif res_json.get("message").find("AF Cookie Not Found") != -1:
+            demisto.results(err_msg)
```
yuvalbenshalom: if there are no results, 
please return "No entries found that match the query" as @kirbles19 suggested

bakatzir: @yuvalbenshalom it's in a different part in the code, changed.

bakatzir:
example:
Screen Shot 2019-07-29 at 10 04 30

PR #3958, opened 2019-07-29
PAN-OS - add edl param in custom-block command

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18266

## Description
added the param to custom-block, logic is the same

3.0.0

   - No

## Must have
- [x] Tests
- [x] Documentation - wrote directly to @kirbles19 
- [x] Code Review


bakatzir:
@yuvalbenshalom I've reviewed them with alex.

PR #3960, opened 2019-07-29
Fix script CBSearch

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17654

   - No

## Must have
- [ ] Code Review

## Dependencies
- [ ] CBFindIP
- [ ] Dependency 2
- [ ] Dependency 3

### DIFF
--- a/Scripts/script-CBSearch.yml
+++ b/Scripts/script-CBSearch.yml
@@ -62,22 +62,20 @@ tags:
 - carbon-black
 - endpoint
 - enhancement
-comment: Search Carbon Black for process & binary information
+comment: Use the CBSearch script to search Carbon Black for process and binary information.
 enabled: true
 args:
 - name: type
-  description: '"processes" or "binary" according to the type of search you want to
-    run'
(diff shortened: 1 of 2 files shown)


Scripts/script-CBSearch.yml
```diff
@@ -49,7 +49,10 @@ script: |-
       };
       if(args.query)
           myArgs.query = args.query;
-      res = executeCommand('cb-' + searchType, myArgs);
+          if(searchType == 'binary')
```
Itay4:
```suggestion
          if (searchType == 'binary')
```

Scripts/script-CBSearch.yml
```diff
@@ -49,7 +49,10 @@ script: |-
...
+              res = executeCommand('cb-' + searchType, myArgs);
```
Itay4: if we already know the the command name, no need for the string concat here, just execute the command name

Itay4:
@kirbles19 please review yml and rn files

kirbles19:
done.

PR #3961, opened 2019-07-29
Anomali ThreatStream v2 - fix description argument

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17948

## Description
see fix in related issue

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/Anomali_ThreatStream_v2/Anomali_ThreatStream_v2.py
+++ b/Integrations/Anomali_ThreatStream_v2/Anomali_ThreatStream_v2.py
@@ -356,8 +356,8 @@ def build_model_data(name, is_public, tlp, tags, intelligence, description):
     """
         Builds data dictionary that is used in Threat Model creation/update request.
     """
-    data = {k: v for (k, v) in (('name', name), ('is_public', is_public), ('tlp', tlp), ('description', description))
-            if v}
+    data = {k: v for (k, v) in (('name', name), ('is_public', is_public), ('tlp', tlp),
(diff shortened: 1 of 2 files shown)


Integrations/Anomali_ThreatStream_v2/Anomali_ThreatStream_v2.py
```diff
@@ -356,8 +356,8 @@ def build_model_data(name, is_public, tlp, tags, intelligence, description):
     """
         Builds data dictionary that is used in Threat Model creation/update request.
     """
-    data = {k: v for (k, v) in (('name', name), ('is_public', is_public), ('tlp', tlp), ('description', description))
-            if v}
+    data = {k: v for (k, v) in (('name', name), ('is_public', is_public), ('tlp', tlp),
```
IkaDemisto: @bakatzir this change will fix the description for `Tipeport` but will break description for all other ThreatModel like Actors, Campaign etc. Need to check if the threat model is `Tipeport` and only then replace it with `body` parameter as all the other model receive description.

bakatzir: @IkaDemisto correct. fixed it, tried for vulnerability, ttp and tipreport and looks good.

IkaDemisto: @bakatzir perfect

PR #3962, opened 2019-07-29
PAN-OS - get url category fix

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18078

3.0.0

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/Panorama/Panorama.py
+++ b/Integrations/Panorama/Panorama.py
@@ -1796,13 +1796,13 @@ def populate_url_filter_category_from_context(category):
     url_filter_category = demisto.dt(demisto.context(), f'Panorama.URLFilter(val.Category === "{category}")')
     if not url_filter_category:
         return []
-    elif isinstance(url_filter_category, dict):
+    elif isinstance(url_filter_category, list):
+        return url_filter_category[0].get("URL")
+    else:  # url_filter_category is a dict
         if isinstance(url_filter_category.get("URL"), str):
             return [url_filter_category.get("URL")]
         else:
             return url_filter_category.get("URL")
-    else:  # url_filter_category is a list of only 1 item
-        return url_filter_category[0].get("URL")
 
 
 def panorama_get_url_category_command():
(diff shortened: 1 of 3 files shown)


bakatzir:
@LPettay tested a few cases, works good. you're most welcome to try as well if you want.
Screen Shot 2019-07-29 at 15 13 08

LPettay:
@bakatzir Gave it some testing. This is a huge improvement, it works great! Thanks!

Integrations/Panorama/Panorama.py
```diff
@@ -1792,45 +1792,50 @@ def panorama_get_url_category(url):
...
 
-    category = panorama_get_url_category(url)
-
-    url_category_hr = {
-        'URL': url,
-        'Category': category
-    }
-
-    url_category_output = populate_url_filter_category_from_context(url_category_hr)
+    categories_dict: Dict[str, list] = {}
```
yuvalbenshalom: did you test this wit a single url?

Integrations/Panorama/Panorama.py
```diff
@@ -1792,45 +1792,50 @@ def panorama_get_url_category(url):
     return s.split(' ')[1]
 
 
-def populate_url_filter_category_from_context(url_category_hr):
-    url_filter_category = demisto.dt(demisto.context(),
-                                     'Panorama.URLFilter(val.Category === "{0}")'.format(url_category_hr['Category']))
-
+def populate_url_filter_category_from_context(category):
```
yuvalbenshalom: did you test it with empty and with a filter?
and why are we populating from context and not from argument?

Integrations/Panorama/Panorama.py
bakatzir: yes

Integrations/Panorama/Panorama.py
bakatzir: yes. because context may have a category with 1 or more urls, and i want to union all the urls from the context and from the command arguments to be under the same category

Integrations/Panorama/Panorama.py
bakatzir: yes, it's working
Screen Shot 2019-07-31 at 12 03 17

PR #3965, opened 2019-07-29
Fixed email on playbook + replaced operations to fix playbook

guykeller:
## Status 
Ready

## Description
Fixed email on playbook + replaced operations to fix playbook

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/TestPlaybooks/playbook-GmailTest.yml
+++ b/TestPlaybooks/playbook-GmailTest.yml
@@ -166,7 +166,7 @@ tasks:
       - "21"
     scriptarguments:
       email:
-        simple: JohnS@demistodev.com
+        simple: JohnS1@demistodev.com
       family-name:
         simple: Smith
       first-name:
@@ -360,7 +360,11 @@ tasks:
                   simple: Gmail.Attachments
                 iscontext: true
           transformers:
-          - operator: GetFirstObject
+          - operator: atIndex
+            args:
+              index:
+                value:
+                  simple: "0"
           - operator: getField
             args:
               field:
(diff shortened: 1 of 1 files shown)


guykeller:
@roysagi FYI

TestPlaybooks/playbook-GmailTest.yml
```diff
@@ -443,7 +447,11 @@ tasks:
               field:
                 value:
                   simple: ID
-          - operator: GetFirstObject
+          - operator: atIndex
```
yaakovi: any reason you replaced that?
if the context you are looking for is not a list, `atIndex` will fail.
this is why `GetFirstObject` is used here: if value is a list, return the first entry, otherwise, return the original object (since there is only one)

TestPlaybooks/playbook-GmailTest.yml
```diff
@@ -166,7 +166,7 @@ tasks:
       - "21"
     scriptarguments:
       email:
-        simple: JohnS@demistodev.com
+        simple: JohnS1@demistodev.com
```
yaakovi: better to manually delete the user from the instance (deleted it myself)

TestPlaybooks/playbook-GmailTest.yml
guykeller: I replaced it because GetFirstObject failed.
Namely - when GetFirstObject was being used, an error was received (Error while parsing argument, or something of that nature). When using atIndex the playbook works.
I still need to look into why GetFirstObject failed

TestPlaybooks/playbook-GmailTest.yml
guykeller: No problem, will revert the change.
Just for future knowledge - why is that better? Meaning, what are the considerations?

guykeller:
Closing this as issue was repaired by fixes unrelated to this PR.

PR #3967, opened 2019-07-29
Request debug mode

avidan-H:
## Status
Ready

## Description
Add logging of requests DEBUG level messages when `debug-mode=True` (logs any/all DEBUG level logger messages).

## Related PRs

branch | PR
set-integration-debug-mode | https://github.com/demisto/server/pull/13189

5.0.0

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Everything


Itay4:
@avidan-H please add test

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+            self.root_logger.removeHandler(self.handler)
+            self.handler.flush()
+            self.handler.close()
+
+
+_requestsLogger = None
+try:
+    # use `hasattr(demisto, 'is_debug')` to ensure compatibility with server version <= 4.5
+    if hasattr(demisto, 'is_debug') and demisto.is_debug:
+        demistoHandler = DemistoHandler()
```
Itay4:
```suggestion
        demisto_handler = DemistoHandler()
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+        demistoFormatter = logging.Formatter(fmt='%(asctime)s - %(message)s', datefmt=None)
```
Itay4:
```suggestion
        demisto_formatter = logging.Formatter(fmt='%(asctime)s - %(message)s', datefmt=None)
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+
+    def __del__(self):
+        self.root_logger.setLevel(self.prev_log_level)
+        if self.handler:
+            self.root_logger.removeHandler(self.handler)
+            self.handler.flush()
+            self.handler.close()
+
+
+_requestsLogger = None
```
Itay4:
```suggestion
_requests_logger = None
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+        self.handler = handler if handler else None
+        if self.handler:
+            self.root_logger.addHandler(self.handler)
+
+    def __del__(self):
+        self.root_logger.setLevel(self.prev_log_level)
+        if self.handler:
+            self.root_logger.removeHandler(self.handler)
+            self.handler.flush()
+            self.handler.close()
```
Itay4: what in case it fails to close?

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+
+_requestsLogger = None
+try:
+    # use `hasattr(demisto, 'is_debug')` to ensure compatibility with server version <= 4.5
+    if hasattr(demisto, 'is_debug') and demisto.is_debug:
+        demistoHandler = DemistoHandler()
+        demistoFormatter = logging.Formatter(fmt='%(asctime)s - %(message)s', datefmt=None)
+        demistoHandler.setFormatter(demistoFormatter)
+        _requestsLogger = DebugLogger(demistoHandler)
+except NameError:
```
Itay4: is `NameError` broad enough?
if this fails, it should be done silently
dont want this to fail commands/playbooks executions

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1682,3 +1683,55 @@ def remove_nulls_from_dictionary(dict):
...
+class DebugLogger(object):
+    """
+    Wrapper to initiate logging at logging.DEBUG level.
+    Is used when `debug-mode=True`.
+    """
+    def __init__(self, handler):
+        self.root_logger = logging.getLogger()
+        self.prev_log_level = self.root_logger.getEffectiveLevel()
+        self.root_logger.setLevel(logging.DEBUG)
+        self.handler = handler if handler else None
```
Itay4: the condition is pointless since you dont set a default for `handler` in the function arguments declaration

Scripts/CommonServerPython/CommonServerPython.py
avidan-H: so I should change it to catch all Exceptions then?

avidan-H: So I should change it to catch all Exceptions then?

Scripts/CommonServerPython/CommonServerPython.py
avidan-H: if this code block doesn't execute then I believe the handler will point to an instance of the demisto object that is deleted upon completion of the command that was called from the command line (in demisto) - but we put a try/except block inside the `emit` method of the handler, so if that happens it will fail silently

glicht:
Can we add some unit tests?

Itay4:
@glicht can we merge this please?

glicht:
I want to rethink the urllib debug option.

glicht:
@yaakovi fyi. small changes to common server docs and package creator

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1814,6 +1874,75 @@ def get_demisto_version():
...
+    """
+        Handler to route logging messages to demisto.debug
+    """
+    def __init__(self):
+        logging.Handler.__init__(self)
+
+    def emit(self, record):
+        msg = self.format(record)
+        try:
+            demisto.debug(msg)
```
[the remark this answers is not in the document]
glicht: Only exception I see is if the demisto object gets cleaned up and some how a reference to this object from the logger remains. I think it is good to be here on the safe side.

glicht:
@avidan-H @Itay4 if no comments, I will merge

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -562,16 +585,23 @@ def __init__(self, ):
...
-            self.add_repalce_strs(pswrd, base64.b64encode(to_encode))
+        if (not os.getenv('COMMON_SERVER_NO_AUTO_REPLACE_STRS') and hasattr(demisto, 'getParam')):
+            # add common params
+            if isinstance(demisto.getParam('credentials'), dict) and demisto.getParam('credentials').get('password'):
+                pswrd = self.encode(demisto.getParam('credentials').get('password'))
+                self.add_replace_strs(pswrd, b64_encode(pswrd))
+            if demisto.getParam('api_key'):
+                s = demisto.getParam('api_key')
+                self.add_replace_strs(s, b64_encode(s))
+            if demisto.getParam('apikey'):
```
Itay4: we are missing many parameters, i think it will be better to iterate the parameters and see if they include some keywords
for example we won't catch `apiKey` or `bot_password`

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1815,6 +1875,75 @@ def get_demisto_version():
...
+_requests_logger = None
+try:
+    # use `hasattr(demisto, 'is_debug')` to ensure compatibility with server version <= 4.5
+    if hasattr(demisto, 'is_debug') and demisto.is_debug:
+        _requests_logger = DebugLogger()
+except Exception as ex:
+    # Should fail silently so that if there is a problem with the logger it will
+    # not affect the execution of commands and playbooks
+    demisto.info('Failed initializing DebugLogger: {}'.format(ex))
+    pass
```
Itay4: do we need that `pass` here?

Scripts/CommonServerPython/CommonServerPython.py
glicht: Good idea. I pushed an update.

glicht:
@kirbles19 please review changelog: https://github.com/demisto/content/pull/3967/files?file-filters%5B%5D=#diff-05e93964b4b3a54428c74e13d263e297R3

kirbles19:
@avidan-H Done.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -582,16 +605,21 @@ def __init__(self):
...
-            to_encode = pswrd
-            if IS_PY3:
-                to_encode = pswrd.encode('utf-8', 'ignore')
-            self.add_repalce_strs(pswrd, base64.b64encode(to_encode))
+        if (not os.getenv('COMMON_SERVER_NO_AUTO_REPLACE_STRS') and hasattr(demisto, 'getParam')):
+            # add common params
+            if isinstance(demisto.getParam('credentials'), dict) and demisto.getParam('credentials').get('password'):
+                pswrd = self.encode(demisto.getParam('credentials').get('password'))
+                self.add_replace_strs(pswrd, b64_encode(pswrd))
+            sensitive_params = ('api_key', 'apikey', 'password', 'secret')
```
Itay4: i would add the following: key, token

anara123: `access_key`, `private_key` could be also options

glicht:
Fix: https://github.com/demisto/etc/issues/19311

Scripts/CommonServerPython/CommonServerPython.py
glicht: I change to simply treat any param with key/secret/password/token/private as relevant

PR #3968, opened 2019-07-29
Moved SumoLogic and Windows Defender ATP to skipped until fixed

guykeller:
## Status 
Ready

## Description
Moved SumoLogic and Windows Defender ATP to skipped until fixed by @avidan-H and @Shellyber accordingly.

   - No


guykeller:
@roysagi - FYI

PR #3979, opened 2019-07-30
create package

jochman:
Create package out of stix parser


Releases/LatestRelease/StixParser.md
```diff
@@ -0,0 +1 @@
+- convert to package
```
yaakovi: no need to say that in release notes

jochman: buy something should be there. don't worry, itllbe fixed

PR #3983, opened 2019-07-30
Added status to outputs

ShahafBenYakir:
## Status
Ready

## Description
Added the status to outputs


Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -686,8 +687,8 @@ def top_tags_results_command():
         'ContentsFormat': formats['text'],
         'Contents': results,
         'EntryContext': {'AutoFocus.TopTagsResults(val.PublicTagName == obj.PublicTagName)': context,
-                         'AutoFocus.SessionsSearch(val.AFCookie == obj.AFCookie)': {'Status': status,
-                                                                                    'AFCookie': af_cookie}},
+                         'AutoFocus.TopTagsSearch(val.AFCookie == obj.AFCookie)': {'Status': status,
```
yuvalbenshalom: why did you change the path?

ShahafBenYakir: It was a mistake. The old path was only entered last week so no danger of breaking backward compatibility.

PR #3984, opened 2019-07-30
Reputation.json - email/url/domain support more languages

bakatzir:
@idovandijk 

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17961

## Description
I've edited the regexes only in reputations.json and added tests in a playbook that runs on this file.
how should we handle the indicators in 5.0? what's the test scheme for that?
Deletions are from removing the FQDN tests which is pushed to later release.

   - No

## Must have
- [x] Tests
- [x] Documentation - no
- [x] Code Review


bakatzir:
e.g:
Screen Shot 2019-07-30 at 18 26 15

TestPlaybooks/playbook-reoutation.json-Test.yml
```diff
@@ -16,295 +16,13 @@ tasks:
       brand: ""
```
yaakovi: please rename the typo in the file name: `reputation`

Tests/secrets_white_list.json
```diff
@@ -256,6 +256,9 @@
     ],
     "_comment": "ENTER URLS WITHOUT http/s, www and path",
     "urls": [
+      "http://ru.wikipedia.org",
```
yaakovi: urls should be listed without the protocol prefix

Tests/secrets_white_list.json
```diff
@@ -256,6 +256,9 @@
...
+      "https://golang.org",
+      "http://b",
```
yaakovi: this could cause a lot of problems since the comparison is by "contains".
in the test, use longer text before the "problematic" letter

idovandijk:
@bakatzir as per your request:

### Invalid URLs

- http://a.b--c.de/
- http://-a.b.co
- http://3628126748
- http://.www.foo.bar/
- http://10.1.1.254
- http:// shouldfail.com
- http://foo.bar/foo(bar)baz quux
- http://-error-.invalid/



### Invalid email addresses

- #@%^%#$@#$@#.com
- @example.com
- Joe Smith <email@example.com>
- email.example.com
- email@example@example.com
- .email@example.com
- email.@example.com
- email..email@example.com
- あいうえお@example.com
- email@example.com (Joe Smith)
- email@example
- email@-example.com
- email@example.web
- email@111.222.333.44444
- email@example..com
- Abc..123@example.com
- ”(),:;<>[\]@example.com
- just”not”right@example.com
- this\ is"really"not\allowed@example.com

bakatzir:
@yaakovi  it runs on master though I added `toVersion` to the tpb.

yaakovi:
@bakatzir there is no `toversion` in the test playbook.
beside, in order to skip a test in a specific version, this needs to be set in the conf json.
AFAIK, we only support 'fromversion' in the conf.json file.

bakatzir:
@yaakovi good to know :)
anyhow i've changes also the 5.0 reputation files so no need for that now.
i'll merger after the release checkout so we can test these changes thoroughly

PR #4001, opened 2019-08-01
Fix upload unknown STIX indicator

jochman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17290

## Description
Fixes a bug where a legit STIX file added with unknown indicator would throw an error and fail. Now it'll ignore unknown indicators.
added unitest to test the bug.

   - No

## Must have
- [X] Tests
- [ ] Code Review# Additional changes


idobloch:
The stix2.json looks good

Scripts/StixParser/StixParser.py
```diff
@@ -184,6 +184,21 @@ def extract_indicators(data):
...
+        Args:
+            indicator (dict): STIX-formatted entry
+        """
+        pattern = indicator.get("pattern")
+        if pattern:
+            groups = regex.findall(pattern)
+            if groups:
+                for key, value in patterns_dict.items():
+                    for term in groups:
+                        if len(term) == 2 and key in term[0]:
```
Itay4: add comment to explain the logic

Scripts/StixParser/StixParser.py
```diff
@@ -184,6 +184,21 @@ def extract_indicators(data):
         dict: containing all indicators data
 
     """
+    def get_indicators(indicator):
```
Itay4: take it out and add test for it

jochman:
after testing with @idobloch we found a bug that created duplicates indicator.
not fixed.

Scripts/StixParser/StixParser.py
```diff
@@ -128,61 +128,44 @@ def dscore(score):
     return 0
 
 
-def build_entry(stx_obj, indicators):
+def build_entry(indicators_dict, stix_indicators_dict, pkg_id):
```
Itay4: add return to docstring

Scripts/StixParser/StixParser.py
```diff
@@ -128,61 +128,44 @@ def dscore(score):
...
+                )
+
+                result = create_indicator_entry(
+                    indicator_type=key,
+                    value=indicator,
+                    pkg_id=pkg_id,
+                    ind_id=ind_id,
+                    timestamp=timestamp,
+                    source=source,
+                    score=score,
```
Itay4: remove trailing comma

Scripts/StixParser/StixParser.py
jochman: outdated (wrong line marked)

Scripts/StixParser/StixParser.py
jochman: problem with taking it out. anyway, it'll be used only in this function. you can use this function with sending input to the parent function (which only checks if we got dict or list

PR #4004, opened 2019-08-01
Anomali solved

roysagi:
## Status
Ready

## Description
Dropped the '.' in the context created - this seems to be the cause of the index out of range error in this command: !threatstream-email-reputation email=mailonline_16@filposcv.com

   - Yes
       - Further details: @yaakovi suggested it might need to see in circle
   - No


roysagi:
@yaakovi - can you say why do you think it might break backwards compatibility?

Itay4:
because users who use populate context path will no longer be able to do so
need to populate to both context paths and change to the correct one (without the trailing dot) in the integration settings

roysagi:
this context path causes error of `index out of range` on server side when running the command: `!threatstream-email-reputation email=mailonline_16@filposcv.com'

After the change context is created correctly
this is causing the command to return both valid output and error entry (of runtime error)   - causing any playbook to stop.

As you can see - context is not created *before* this change was made:
Screen Shot 2019-08-01 at 18 25 34

Screen Shot 2019-08-01 at 18 22 32

Thus this should not cause BC errors.

Releases/LatestRelease/integration-AnomaliThreatStream.md
```diff
@@ -0,0 +1 @@
+In line 251 - dropped the '.' in the end of the word 'Reputation'
```
yaakovi: @kirbles19 please fix this release notes. they are too developer oriented.

yaakovi:
since the context is not created and the original functionality causes errors, this shouldn't be considered as a BC breaking.

kirbles19:
@roysagi @yaakovi RN updated, all done.

roysagi:
It seems Anomali currently has an unrelated issue concerning credentials (https://github.com/demisto/etc/issues/18561) thus, Anomali's test has been skipped.
I've consulted with @yaakovi and because this issue was solved and tested before the credential problem - I'll still merge this fix.

yaakovi:
@roysagi pull from master and update the RN file to the new standard (*CHANGELOG.md*)

PR #4011, opened 2019-08-04
Skipped SCADAfence CNM and Cortex XDR tests

roysagi:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/18376
https://github.com/demisto/etc/issues/18377

## Description
SCADAfence CNM - instance fails sporadically with 502 bad gateway error.
Cortex XDR - instance fails with 500 internal server error.


yardensade:
What is the status here? are we going to merge this?

PR #4015, opened 2019-08-04
Email Enrichment v2.1 - filter empty domains

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18288

## Description
#17390 will fix only from 5.0. fixing the pb as well.

4.1.0

   - No

## Must have
- [x] Tests
- [ ] Code Review

### DIFF
--- a/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
+++ b/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
@@ -493,6 +493,12 @@ tasks:
       value:
         complex:
           root: inputs.Domain
+          filters:
+            - - operator: isNotEmpty
+                left:
+                  value:
+                    simple: inputs.Domain
+                  iscontext: true
           transformers:
           - operator: uniq
     separatecontext: false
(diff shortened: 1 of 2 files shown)


yuvalbenshalom:
@bakatzir please do test that re create the problem and show that it is fixed. even manual.

bakatzir:
@yuvalbenshalom  looks good.
Screen Shot 2019-08-05 at 10 26 09
Screen Shot 2019-08-05 at 10 26 19

bakatzir:
@yuvalbenshalom prior to fix it went to the else path:
Screen Shot 2019-08-05 at 11 00 58

idovandijk:
Can you please provide server for review / show me on your local instance that it solves the bug?
I have 2 concerns:
1. I am not sure whether "Is Not Empty" operator can check all elements of `inputs.Domain`, I think it checks if the object itself is not empty and not whether its children all have data.
However, if I am wrong, and it does work as intended, then the filter should be added to the `playbook inputs` **as well as the** `Is there a domain list input?` task

bakatzir:
Set down with @idovandijk  and changed accordingly, adding more filters to playbook input as well

PR #4016, opened 2019-08-04
Stix2 new indicators support

jochman:
## Status
Ready

## Description
Adds support for registry and CVE indicators. 
added unit tests
fixed bug with `ip-x-x-x-x` to `x.x.x.x` IP format

## Screenshots

   - No

## Must have
- [X] Tests
- [ ] Code Review

## Additional changes

### DIFF
--- a/Scripts/StixParser/StixParser.py
+++ b/Scripts/StixParser/StixParser.py
@@ -14,10 +14,8 @@ PATTERNS_DICT = {
     "ipv4-addr:": "IP",
     "url:": "URL",
     "domain-name:": "Domain",
-    "email:": "Email",
-    "email-message": "Email",
-    "win-registry-key:key": "Registry Path Reputation",
-    "windows-registry-key:key": "Registry Path Reputation"
+    "email": "Email",
+    "registry-key:key": "Registry Path Reputation"
 }
 
 """ HELPER FUNCTIONS"""
@@ -27,13 +25,15 @@ def ip_parser(ip):
     """IP can be in the form of `ip-x-x-x-x`.
     function will return it to `x.x.x.x` format
 
+    if it's not an ip, will return `ip` arg back.
+
     Args:
         ip (str): ip to parse
 
     Returns:
         str: parsed ip indicator
     """
-    if ip.startswith("ip-"):
+    if ip.lower().startswith("ip-"):
         return ip.lower().replace("ip-", "").replace("-", ".")
     return ip
 
@@ -206,7 +206,7 @@ def get_indicators(indicators):
             {
                 "<key>": "<stix entry>"
             }
-            )
+        )
     """
 
     def indicators_parser(stix_indicator):
--- a/Scripts/StixParser/StixParser.py
+++ b/Scripts/StixParser/StixParser.py
@@ -245,7 +245,6 @@ def get_indicators(indicators):
         "Domain": list(),
         "Email": list(),
         "URL": list(),
-        # TODO implement below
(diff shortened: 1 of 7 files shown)


Itay4:
what issue does this refer to ?

Scripts/StixParser/StixParser.py
```diff
@@ -15,12 +15,29 @@
     "url:": "URL",
     "domain-name:": "Domain",
     "email:": "Email",
-    "email-message": "Email"
+    "email-message": "Email",
+    "win-registry-key:key": "Registry Path Reputation",
```
Itay4: why do we need both ?
why isnt CVE here?

Scripts/StixParser/StixParser.py
```diff
@@ -178,29 +195,48 @@ def get_indicators(indicators):
...
             {
                 "File": [],
                 "IP": [],
                 "Domain": [],
                 "URL": []
+            },
+            {
+                "<key>": "<stix entry>"
             }
+            )
```
Itay4: lint properly

Scripts/StixParser/StixParser_test.py
```diff
@@ -94,3 +94,82 @@ def test_stix2_to_json_empty_case(self, mocker):
...
+
+    def test_create_timstamp_false(self):
+        from StixParser import create_timestamp
+        false_value = "nottimestamp"
+        result = create_timestamp(false_value)
+        assert result is None, self.err_msg.format(None, result)
+
+    def test_create_indicator_entry(self):
+        from StixParser import create_indicator_entry
+        expected = {'indicator_type': 'ip',
```
Itay4: lint properly

Scripts/StixParser/StixParser.py
```diff
@@ -178,29 +195,48 @@ def get_indicators(indicators):
...
     """
 
     def indicators_parser(stix_indicator):
```
Itay4: can we take that function out?

Scripts/StixParser/StixParser.py
jochman: because CVE entry has no `pattern` key, so it's parsed differently

Scripts/StixParser/StixParser.py
```diff
@@ -209,6 +245,9 @@ def indicators_parser(stix_indicator):
         "Domain": list(),
         "Email": list(),
         "URL": list(),
+        # TODO implement below
```
Itay4: remove comment

Scripts/StixParser/TestData/stix2_results.json
```diff
@@ -1,123 +1,295 @@
...
-    "indicator_type": "Email",
-    "value": "user@example.com",
-    "CustomFields": {
-      "indicatorId": "indicator--7ed80192-4b84-4954-8b41-6268290cd5c7",
-      "stixPackageId": "bundle--6ab84b67-f7a2-47b4-b7cd-097231cb7031"
-    },
-    "source": "indicator",
-    "score": 0,
-    "timestamp": "2019-05-26T16:11:08.000Z"
+    "http://johnson.com": {
```
Itay4: jonhson -> demisto

Scripts/StixParser/StixParser.py
jochman: problem with taking it out. anyway, it'll be used only in this function. you can use this function with sending input to the parent function (which only checks if we got dict or list

PR #4007, opened 2019-08-02
detailed documentation added

anara123:
## Status
In progress

## Related Issues
https://github.com/demisto/etc/issues/18211

## Description
- sync will update grid fields instead of markdown field
- added verbose option

## Related PRs
https://github.com/demisto/content/pull/3875

## Does it break backward compatibility?
   - Yes
       - Currently yes, XDRSyncScript will populate alerts to xdralerts instead of markdown field

## Must have
- [x] Tests
- [x] Documentation https://github.com/demisto/etc/issues/18211
- [ ] Code Review

PR #4020, opened 2019-08-04
XDR fixes

glicht:
* detailed documentation added
- sync will update grid fields instead of markdown field
- added verbose option
-

* fix lint

* fix build

* fix XDR yml

* changed detailed descriptsion so that the yaml could be parsed

* try without description

* Updated description.

## Status
Ready

## Description
Include of: #4007 into the 19.8.0 branch


anara123:
don't merge yet, I will add proper description

PR #4026, opened 2019-08-05
SymantecEndpointProtection test playbook fix

roysagi:
## Status
Ready

## Description
Test playbook for SymantecEndpointProtection searched for 192.168.1.17 while it got 192.168.1.12 in the response - failing the playbook

### DIFF
--- a/TestPlaybooks/playbook-SymantecEndpointProtection_Test.yml
+++ b/TestPlaybooks/playbook-SymantecEndpointProtection_Test.yml
@@ -718,7 +718,7 @@ tasks:
             iscontext: true
           right:
             value:
-              simple: 192.168.1.17
+              simple: 192.168.1.12
     view: |-
       {
         "position": {


Itay4:
do we know what wrong with the mocks that this change is needed?

roysagi:
> do we know what wrong with the mocks that this change is needed?

It seems the mock failed on the same error:
image

Itay4:
we need to investigate this further to understand why the mock recorded this in the first place
this is exactly what the mocks are for

roysagi:
Me and @yaakovi reviewed the several nightly runs back - we saw this happened about a week ago as well (same exact issue), while the mock was disabled then.

Because this has not happened since then until today and it did pass every time in between - we decided to wait till tomorrow to see if this issue returns before we change anything.

roysagi:
It seems it worked today using the mock:
image

Thus this correction might not be necessary - we should monitor it or consider altering the playbook to match spheroidic failure reason.

Itay4:
@yaakovi @roysagi lets discuss this

roysagi:
Update: Circle started to rerun today (with no changes in the build) and the error has returned (both in mock and actual run).

yaakovi:
closing for now, we will revisit this once the test fails again.

PR #4028, opened 2019-08-05
Vertica error name fix

roysagi:
## Status
Ready

## Description
At line 14 in the integration - it looked only for "KeyError" exception.
this at times resulted in this error - "name ‘KeyError’ is not defined"

After reproducing the problem - found that the error being thrown is "NameError".
Checked the test playbook several times after the change - it didn't occur again


roysagi:
@yaakovi @yardensade - FYI

PR #4035, opened 2019-08-06
ThinkstCanary - handle error msg with grace

bakatzir:
## Status
Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/18321

## Description
please see details in related issue.

   - No

## Must have
- [x] Tests
- [x] Code Review


bakatzir:
@jpeng19 would love if you can take this version and check with the customer, this way we can get a relevant error message and address the problem.

Integrations/ThinkstCanary/ThinkstCanary.py
```diff
@@ -62,8 +62,13 @@ def http_request(method, url, params=None):
     )
 
     if not res.ok:
-        demisto.debug(res.text)
-        return_error('Could not execute the request')
+        try:
+            res_json = res.json()
```
yuvalbenshalom: the result may not be a json and may throw ValueError.
in this case you should catch it sepcificly and return the content sting.
more over, if you get another exception, you should print the exception str.

PR #4050, opened 2019-08-07
fix bug in get alarm events when DrillDownLogs is empty

adi88d:
## Status
Ready

## Description
The command crash when DrillDownLogs was empty


Integrations/LogRhythmRest/LogRhythmRest.py
```diff
@@ -556,6 +556,8 @@ def get_alarm_events(data_args):
     events = []
 
     for block in res:
+        if not block['DrillDownLogs']:
```
yaakovi: better to be on the safe-side
```suggestion
        if not block.get('DrillDownLogs'):
```

kirbles19:
@adi88d @yaakovi Looked great, all done.

PR #4057, opened 2019-08-08
Beta_Integration - add beta tag to integration and PB, remove TPB

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18476
fixes: https://github.com/demisto/etc/issues/18498

   - No


ronykoz:
How changing the order of which we copy the files fixes anything? 
The Demisto server is down at that point which means it doesn’t matter

bakatzir:
a test playbook was missing sub-playbooks that were in the beta repo so the content was not updated in the server. changing the order we copy meant the needed playbooks were found.

Beta_Integrations/Blueliv/Blueliv.yml
```diff
@@ -48,3 +48,4 @@ script:
   runonce: false
   script: ''
   type: python
+  beta: True
```
yuvalbenshalom: replace `True` with `true`

Beta_Integrations/MailListener_-_POP3/MailListener_-_POP3.yml
```diff
@@ -43,12 +43,13 @@ configuration:
...
-display: MailListener - POP3
-name: MailListener - POP3
+display: MailListener - POP3 (Beta)
+name: MailListener - POP3 Beta
 script:
   isfetch: true
   runonce: false
   script: '-'
   type: python
+  beta: True
```
yuvalbenshalom: replace True with true

Beta_Integrations/Telegram/Telegram.yml
```diff
@@ -44,3 +44,4 @@ script:
   runonce: false
   script: '-'
   type: python
+  beta: True
```
yuvalbenshalom: replace True with true

Beta_Integrations/playbook-TrendMicro_DDA_DetonateURL.yml
```diff
@@ -368,5 +368,6 @@ outputs:
 - contextPath: File.Name
   description: File name
   type: string
+beta: True
```
yuvalbenshalom: replace True with true

PR #4060, opened 2019-08-08
PAN-OS - Handle error trying to refresh an EDL object on Panorama 

bakatzir:
…rror gracefully

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17807

   - No


bakatzir:
already reviewed the error message with alex.

PR #4066, opened 2019-08-08
Package VirusTotal Private API

amshamah419:
## Status
Ready

## Related Issues
fixes: Nothing

## Description
Packaging for future bugfix


Integrations/VirusTotal-Private_API/VirusTotal-Private_API.py
```diff
@@ -0,0 +1,1152 @@
...
+import requests
+import json
+import os
+import time
+import sys
+
+# disable insecure warnings
+requests.packages.urllib3.disable_warnings()
+
+if not demisto.params().get('useProxy', False):
```
yuvalbenshalom: use handle proxy,
notice that the parameter name is "useProxy" so pass it to the function:
`handle_proxy(proxy_param_name='useProxy')`

yuvalbenshalom:
@amshamah419 and also see guardraills fail

PR #4069, opened 2019-08-08
add Exception object to return_error

jochman:
## Status
Ready

## Description
Added the option to a raised error to return_error so we can log the Error and not depend on the user.

## Screenshots
image

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Scripts/CommonServerPython/CommonServerPython.py
@@ -1352,7 +1352,7 @@ def return_outputs(readable_output, outputs, raw_response=None):
     demisto.results(return_entry)
 
 
-def return_error(message, error='', outputs=None):
+def return_error(message, error='', outputs=None, exception_object=None):
     """
         Returns error entry with given message and exits the script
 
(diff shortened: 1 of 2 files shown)


Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -480,5 +482,16 @@ def test_is_mac_address():
...
+    assert (is_mac_address(mac_address_true))
+
+
+def test_exception_in_return_error(mocker):
+    from CommonServerPython import return_error, IntegrationLogger
+
+    expected = {'EntryContext': None, 'Type': 4, 'ContentsFormat': 'text', 'Contents': 'Message'}
+    mocker.patch.object(demisto, 'results')
+    with pytest.raises(SystemExit, match='0'):
+        return_error("Message", exception_object=ValueError("Error!"))
```
Itay4: need to test that `LOG` is called with the expected value

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1365,12 +1365,17 @@ def return_error(message, error='', outputs=None):
         :type outputs: ``dict or None``
         :param outputs: the outputs that will be returned to playbook/investigation context (optional)
 
+        :type exception_object: ``Exception or None``
```
yaakovi: why not giving the `error` argument accept Exceptions as well, that was the original purpose of it

yaakovi:
forced since it run all tests and failed. (relevant steps finished successfully)

PR #4082, opened 2019-08-11
stix fix username field

jochman:
## Status
Ready

## Description
fixed missed username regex.

   - No

## Must have
- [X] Tests


jochman:
I and @idobloch reviewed it again with import/export and it now fully works.

PR #4087, opened 2019-08-11
Tenable_sc Integration Packagifying:

dantavori:
Fixed ReleaseNotes format


Integrations/Tenable_sc/_changelog.md
```diff
@@ -0,0 +1,2 @@
+## [Unreleased]
+  - Packagified the integration.
```
Itay4: no need this to the rn, should just be empty: `-`

PR #4088, opened 2019-08-11
Fix BC checks for packagify

yaakovi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18434

## Description
fix can be seen here, docker was changed at first commit (before merge - success, after merge - failure):
https://circleci.com/gh/demisto/content/tree/LuminatePackagify

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Tests/test_utils.py
```diff
@@ -57,6 +58,70 @@ def run_command(command, is_silenced=True, exit_on_error=True):
     return output
 
 
+def filter_packagify_changes(modified_files, added_files, removed_files, tag='master'):
```
yuvalbenshalom: release_notes.py is importing functions from this file.(test_utils.py) so I think it should import this one also and avoid the copy-code

PR #4095, opened 2019-08-12
Update git sha1 + server version to current GA

yardensade:
## Status
Ready


yardensade:
The git sha1 I took is the last commit in the release branch

PR #4096, opened 2019-08-12
Revert "Update git sha1 + server version to current GA"

yardensade:
Reverts demisto/content#4095
will add changes via @yaakovi 's PR


yaakovi:
no need to revert

PR #4100, opened 2019-08-12
set default encoding to utf-8 in Intsights

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18055

## Description
the integration crashed when the extracting alerts tags with different encoding, set default encoding to utf-8 fix that


orenzohar:
it says that you're breaking backward compatibility when changing the docker image. you might want to discuss this with @yaakovi or @glicht

Itay4:
please update the pr title

adi88d:
updated the PR title and description

PR #4113, opened 2019-08-13
[Fix] Changed Indicator Fields IDs

DeanArbel:
## Status
Ready

## Description
Changed id for indicators that did not match expected id name (was: `incident_*` should've been `indicator_*`)

### Before fix: 
image

### After fix (to be uploaded after artifact is read):
Os version and description successfully installed:
image

image

5.0.0

   - No

### DIFF
--- a/IndicatorFields/incidentfield-description.json
+++ b/IndicatorFields/incidentfield-description.json
@@ -5,7 +5,7 @@
   "neverSetAsRequired": false,
   "threshold": 72,
   "defaultRows": null,
-  "id": "incident_description",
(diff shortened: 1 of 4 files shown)


DeanArbel:
This was missed because in version 5.0.0 the contest build in circleci skips problematic files (i.e. files that make the content build fail), so there was no indication of this error.

For future changes to indicator fields I highly recommend doing manual tests:
1. Try to update content of 4.5 demisto and see that the new fields don't appear.
2. Try to update content of 5.0 demisto and see that **ALL** the new fields appear.

PR #4125, opened 2019-08-14
QRadar outgoing http requests logging

DeanArbel:
## Status
Ready

## Description
Added to QRadar integration outgoing logging for http requests including headers (excl. SEC header), url params, url method and url address.

   - No

## Must have
- [ ] Code Review


Itay4:
@DeanArbel what will be logged in case credentials are used ?

DeanArbel:
> @DeanArbel what will be logged in case credentials are used ?

The same logging as would be if the user would use API token. The username/password aren't in the headers, but are added later via the authorization argument.

PR #4131, opened 2019-08-18
fixed fetch-incidents when fetch time is more than one hour

jochman:
## Status
On Hold (waiting for the customer to validate)

## Related Issues
fixes: https://github.com/demisto/etc/issues/18609

## Description
fetch-incidents now fetching for every hour from last_fetch.
we have no instance, but our tests are checking fetch-incident.
needed to mock current time to make those tests pass, but it should not break bc.

   - No

## Must have
- [X] Tests
- [ ] Code Review


Itay4:
@anara123 can you please review?

anara123:
yes

Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
```diff
@@ -18,15 +18,50 @@
...
+
+def get_fetch_times(last_fetch):
+    """ Get list of every hour since last_fetch
+    Args:
+        last_fetch (datetime or str): last_fetch time
+
+    Returns:
+        List[str]: list of str represents every hour since last_fetch
+    """
+    now = get_now() - timedelta(seconds=30)
```
anara123: why you reduce 30 seconds?

Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
```diff
@@ -18,15 +18,50 @@
...
+    times = list()
+    time_format = "%Y-%m-%dT%H:%M:%SZ"
+    if isinstance(last_fetch, str):
+        times.append(last_fetch)
+        last_fetch = datetime.strptime(last_fetch, time_format)
```
anara123: dont override function argument `last_fetch` , it is a bad practice. instead define new variable

Integrations/ProofpointTAP_v2/ProofpointTAP_v2_test.py
```diff
@@ -260,4 +269,13 @@ def test_next_fetch(requests_mock):
 
     assert len(incidents) == 4
     assert json.loads(incidents[0]['rawJSON'])["messageID"] == "1111@evil.zz"
-    assert next_run == {"last_fetch": "2010-01-30T00:01:00.000Z"}
+
+
+def test_get_fetch_times():
```
anara123: add one more test with predefined `last_fetch="2019-01-01T22:00:00"` and validate the array first 3 objects.
expected to be 
`["2019-01-01T22:10:05", "2019-01-01T23:10:05", "2019-01-02T00:10:05", ...] `

Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
jochman: Proof Point unable to pull event the from the last 30 seconds so I'm adding it to the time gap.

jochman:
validation: https://github.com/demisto/etc/issues/18320

glicht:
@jochman unit tests are failing in the build

PR #4133, opened 2019-08-18
tableToMarkdown fix - dictionary with special character

IkaDemisto:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17450

## Description
In case of passing dictionary with unicode key and list that includes non ascii character (for example: {'header_1': u'foo', 'header_2': [u'â']}) to `tableToMarkdown`, `UnicodeDecodeError` error was raised. The issue was due to joining unicode and str types.

4.5.0

   - No

## Must have
- [x] Tests
- [x] Code Review

## Dependencies
- [ ] All tests


Itay4:
what in case non-ascii in table headers?

IkaDemisto:
@Itay4 correct, fails in particular cases. Fails with `{'header_1': u'foo'}` but not with `{'header_1': [u'foo']}`. Will include fix for it also.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -830,9 +830,16 @@ def tableToMarkdown(name, t, headers=None, headerTransform=None, removeNull=Fals
...
+                try:
+                    mdResult += vals[0]
+                except UnicodeDecodeError:
+                    mdResult += vals[0].encode('utf-8')
             else:
-                mdResult += '|'.join(vals)
+                try:
+                    mdResult += '|'.join(vals)
+                except UnicodeDecodeError:
+                    vals = [str(v) for v in vals]
```
yaakovi: let's use encode here too

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -830,9 +830,16 @@ def tableToMarkdown(name, t, headers=None, headerTransform=None, removeNull=Fals
                                    True, True) for h in headers]
             mdResult += '|'
             if len(vals) == 1:
```
yaakovi: let's remove the condition, the else part works for both cases

anara123:
@IkaDemisto @yaakovi what is the status here?

IkaDemisto:
Waited to content checkout. Will be force merged tomorrow @yaakovi?

yaakovi:
changing `CommonServerPython` - will fail on tests.
unit tests and file validation passed

PR #4135, opened 2019-08-19
Add get_demisto_version function

Itay4:
## Status
Ready

## Description
Added common server function which wraps a server function that returns server version and build number

5.0.0

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Dependencies
need to check

## Additional changes
Added function to mock


Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1752,3 +1752,19 @@ def remove_nulls_from_dictionary(dict):
...
+        Returns Demisto version and build number
+
+        :return: Demisto version object
+        :rtype: ``dict``
+    """
+    if hasattr(demisto, 'demistoVersion'):
+        return demisto.demistoVersion()
+    else:
+        return {
+            'version': '0.0.0',
```
yuvalbenshalom: isn't it better to return an error msg if you cannot fetch the version, rather than returning magic number that looks like  a valid version?

Itay4: it would require wrapping it in `try-except` block if used in an integration, for example
the idea was to avoid it and get the version, and then be able to compare between versions

yuvalbenshalom: I dont see how warping in an `if` statement and comparing it to "0.0.0" is better than `try-except`.
and if you are just checking if the answer equals the version you expect (say, `if res['version'] == '5.0.0':`),
than you may never know that it is not that you didn't got the right version but you didn't get a version at all.

yaakovi:
@Itay4 go over the function description with @kirbles19 before merging since this is shown in the product under the script helper

Itay4:
@kirbles19 please review the docstring in https://github.com/demisto/content/pull/4135/files#diff-350c360238877bfc4845e7598d2f392a

richardbluestone:
Made some minor change - review done

PR #4137, opened 2019-08-19
Cybereason packaging

orlichter1:
## Status
Ready

## Related Issues
fixes: #18605
(don't want to link so the issue won't close)

## Description
Just packaging of the Integration 

   - No


Integrations/Cybereason/Cybereason.py
```diff
@@ -0,0 +1,1478 @@
...
+    elif demisto.command() == 'cybereason-query-file':
+        query_file_command()
+
+    elif demisto.command() == 'cybereason-query-domain':
+        query_domain_command()
+
+    elif demisto.command() == 'cybereason-query-user':
+        query_user_command()
+
+except Exception, e:
```
yaakovi: error handling should be `return_error` and not `raise`.
```suggestion
except Exception as e:
```

ronykoz:
@yaakovi seems strange, I’ll check that on Sunday.
@orlichter1 for the future you can just link the issue using one of the following 
Just copy the url without the fixes prefix 
Copy the url with the relates to prefix
This way the issue will not close once the PR is merged

ronykoz:
It’s because we are using re_create_id_set instead of updating the set, hence not catching the duplication, open an issue to fix this please

orlichter1:
@yaakovi We said we'll fix these issues in the malop-fix branch (which I already did). Please approve here :)

PR #4141, opened 2019-08-19
 - Gmail enhancement and context standard

roysagi:
## Status
In Progress

## Related Issues
https://github.com/demisto/etc/issues/18366
and possibly: 
https://github.com/demisto/etc/issues/17478
https://github.com/demisto/etc/issues/17979

## Description
Did several things:
1) added 6 commands:
  * gmail-hide-user-in-directory
  * gmail-set-password
  * gmail-get-autoreply
  * gmail-set-autoreply
  * gmail-delegate-user-mailbox
  * gmail-send-mail

2) updated the Context for Account and Email to fit the new standard: https://github.com/demisto/content/tree/master/docs/context_standards

3) updated the test playbook fro gmail to check the new commands

4) changed the way 'Date' field is set in Email/Gmail context - it used to be taken from the email payload and now it is taken from the attached timestamp. This is considered more reliable: https://developers.google.com/gmail/api/v1/reference/users/messages

5.0.0

   - Yes, sort of
       - Further details: it might create a several hour gap in Gmail incidents fetched - will expand in a comment.

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes

### DIFF
--- a/Integrations/Gmail/CHANGELOG.md
+++ b/Integrations/Gmail/CHANGELOG.md
@@ -1,0 +1,7 @@
+Added 6 new commands:
+- gmail-hide-user-in-directory - Hide a user in the Global Directory.
+- gmail-set-password - Set password for the user.
+- gmail-get-autoreply - Get the auto-reply message set for the user-account.
+- gmail-set-autoreply - Set auto-reply for the user.
+- gmail-delegate-user-mailbox - Adds or removes a delegate to the mailbox, without sending any verification email.
+- gmail-send-mail - Send mail using Gmail.
--- a/Integrations/Gmail/Gmail.py
+++ b/Integrations/Gmail/Gmail.py
@@ -6,15 +6,20 @@ import re
 import json
 import base64
 from datetime import datetime, timedelta
-import httplib2
-import urlparse
 from distutils.util import strtobool
 import sys
 from HTMLParser import HTMLParser, HTMLParseError
 from htmlentitydefs import name2codepoint
+from email.mime.audio import MIMEAudio
+from email.mime.base import MIMEBase
+from email.mime.image import MIMEImage
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+import mimetypes
 
 from apiclient import discovery
 from oauth2client import service_account
+
 
 
 ''' GLOBAL VARS '''
@@ -157,6 +162,72 @@ def parse_mail_parts(parts):
     return body, html, attachments
 
 
+def days_of_the_week(day):
+    if day == 1:
+        return 'Mon'
+    elif day == 2:
+        return 'Tue'
+    elif day == 3:
+        return 'Wed'
+    elif day == 4:
+        return 'Thu'
+    elif day == 5:
+        return 'Fri'
+    elif day == 6:
+        return 'Sat'
+    elif day == 7:
+        return 'Sun'
+
+
+def month_abbr(mon_num):
+    if mon_num == 1:
+        return 'Jan'
+    elif mon_num == 2:
+        return 'Feb'
+    elif mon_num == 3:
+        return 'Mar'
+    elif mon_num == 4:
+        return 'Apr'
+    elif mon_num == 5:
+        return 'May'
+    elif mon_num == 6:
+        return 'Jun'
+    elif mon_num == 7:
+        return 'Jul'
+    elif mon_num == 8:
+        return 'Aug'
+    elif mon_num == 9:
+        return 'Sep'
+    elif mon_num == 10:
+        return 'Oct'
+    elif mon_num == 11:
+        return 'Nov'
+    elif mon_num == 12:
+        return 'Dec'
+
+
+def time_reformat(timestamp_date):
+    timestamp_date = timestamp_date.split(' ')
+    exact_time = timestamp_date[1]
+    date = timestamp_date[0]
+    date = date.split('-')
+    year = date[0]
+    mon = month_abbr(int(date[1]))
+    day = date[2]
+    return day+" "+mon+" "+year+" "+exact_time
+
+
+def utc_extract(base_time):
+    utc = base_time[-5:]
+    if utc[0] is not '-' and utc[0] is not '+':
+        return '-0000', 0
+    for ch in utc[1:]:
+        if not ch.isdigit():
+            return '-0000', 0
+    delta_in_seconds = int(utc[0] + utc[1:3]) * 3600 + int(utc[0] + utc[3:]) * 60
+    return utc, delta_in_seconds
+
+
 def get_email_context(email_data, mailbox):
     context_headers = email_data.get('payload', {}).get('headers', [])
     context_headers = [{'Name': v['name'], 'Value':v['value']}
@@ -166,7 +237,12 @@ def get_email_context(email_data, mailbox):
     body = body.encode('ascii') if body is not None else ''
     parsed_body = base64.urlsafe_b64decode(body)
 
-    context = {
+    utc, delta_in_seconds = utc_extract(str(headers.get('date', '')))
+    base_time = datetime.fromtimestamp((int(str(email_data.get('internalDate'))[:10]))) + timedelta(seconds=delta_in_seconds)
+    day = days_of_the_week(int(base_time.isoweekday()))  # gets day of week from date
+    base_time = day + ", " + time_reformat(str(base_time)) + " " + utc
+    #return_error(str(headers.get('date', ''))+" ### "+base_time)
+    context_gmail = {
         'Type': 'Gmail',
         'Mailbox': ADMIN_EMAIL if mailbox == 'me' else mailbox,
         'ID': email_data['id'],
@@ -187,21 +263,42 @@ def get_email_context(email_data, mailbox):
         # only for incident
         'Cc': headers.get('cc', []),
         'Bcc': headers.get('bcc', []),
-        'Date': headers.get('date', ''),
+        'Date': unicode(base_time),
+        #'Date': headers.get('date', ''),
         'Html': None,
     }
 
-    if 'text/html' in context['Format']:  # type: ignore
-        context['Html'] = context['Body']
-        context['Body'] = html_to_text(context['Body'])
+    context_email = {
+        'Headers': context_headers,
+        'Attachments': {'entryID': email_data.get('payload', {}).get('filename', '')},
+        # only for format 'raw'
+        'RawData': email_data.get('raw'),
+        # only for format 'full' and 'metadata'
+        'Format': headers.get('content-type', '').split(';')[0],
+        'Subject': headers.get('subject'),
+        'From': headers.get('from'),
+        'To': headers.get('to'),
+        # only for format 'full'
+        'Body/Text': unicode(parsed_body, 'utf-8'),
 
-    if 'multipart' in context['Format']:  # type: ignore
-        context['Body'], context['Html'], context['Attachments'] = parse_mail_parts(
+        'CC': headers.get('cc', []),
+        'BCC': headers.get('bcc', []),
+        'Date': unicode(base_time),
+        #'Date': headers.get('date', ''),
+        'Body/HTML': None,
+    }
+
+    if 'text/html' in context_gmail['Format']:  # type: ignore
+        context_gmail['Html'] = context_gmail['Body']
+        context_gmail['Body'] = html_to_text(context_gmail['Body'])
+
+    if 'multipart' in context_gmail['Format']:  # type: ignore
+        context_gmail['Body'], context_gmail['Html'], context_gmail['Attachments'] = parse_mail_parts(
             email_data.get('payload', {}).get('parts', []))
-        context['Attachment Names'] = ', '.join(
-            [attachment['Name'] for attachment in context['Attachments']])  # type: ignore
+        context_gmail['Attachment Names'] = ', '.join(
+            [attachment['Name'] for attachment in context_gmail['Attachments']])  # type: ignore
 
-    return context, headers
+    return context_gmail, headers, context_email
 
 
 TIME_REGEX = re.compile(r'^([\w,\d: ]*) (([+-]{1})(\d{2}):?(\d{2}))?[\s\w\(\)]*$')
@@ -210,7 +307,6 @@ TIME_REGEX = re.compile(r'^([\w,\d: ]*) (([+-]{1})(\d{2}):?(\d{2}))?[\s\w\(\)]*$
 def parse_time(t):
     # there is only one time refernce is the string
     base_time, _, sign, hours, minutes = TIME_REGEX.findall(t)[0]
-
     if all([sign, hours, minutes]):
         seconds = int(sign + hours) * 3600 + int(sign + minutes) * 60
         parsed_time = datetime.strptime(
@@ -243,8 +339,8 @@ def create_incident_labels(parsed_msg, headers):
 def emails_to_entry(title, raw_emails, format_data, mailbox):
     emails = []
     for email_data in raw_emails:
-        context, _ = get_email_context(email_data, mailbox)
-        emails.append(context)
+        context_gmail, _, context_email = get_email_context(email_data, mailbox)
+        emails.append(context_gmail)
 
     headers = {
         'minimal': ['Mailbox', 'ID', 'Labels', 'Attachment Names', ],
@@ -259,12 +355,13 @@ def emails_to_entry(title, raw_emails, format_data, mailbox):
         'Contents': emails,
         'ReadableContentsFormat': formats['markdown'],
         'HumanReadable': tableToMarkdown(title, emails, headers[format_data]),
-        'EntryContext': {'Gmail(val.ID && val.ID == obj.ID)': emails}
+        'EntryContext': {'Gmail(val.ID && val.ID == obj.ID)': emails,
+                         'Email(val.Date == obj.Date && val.Subject == obj.Subject)': context_email}
     }
 
 
 def mail_to_incident(msg, service, user_key):
-    parsed_msg, headers = get_email_context(msg, user_key)
+    parsed_msg, headers, _ = get_email_context(msg, user_key)
 
     file_names = []
     command_args = {
@@ -295,7 +392,8 @@ def mail_to_incident(msg, service, user_key):
         'name': parsed_msg['Subject'],
         'details': parsed_msg['Body'],
         'labels': create_incident_labels(parsed_msg, headers),
-        'occurred': parse_time(parsed_msg['Date']),
+        'occurred': datetime.strptime(parsed_msg['Date'][:-6], '%a, %d %b %Y %H:%M:%S').isoformat() + 'Z',
+        #'occurred': parse_time(parsed_msg['Date']),
         'attachment': file_names,
         'rawJSON': json.dumps(parsed_msg),
     }
@@ -315,9 +413,18 @@ def users_to_entry(title, response):
             'Gmail': {'Address': user_data.get('primaryEmail')},
             'Group': user_data.get('kind'),
             'CustomerId': user_data.get('customerId'),
+            'Groups': user_data.get('kind'),
+            'Domain': user_data.get('primaryEmail').split('@')[1],
+            'Username': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
+                                 if (user_data.get('organizations') and len(user_data.get('organizations')) > 0
+                                     and 'name' in user_data.get('organizations')[0]) else None)
+
         })
     headers = ['Type', 'ID', 'UserName',
-               'DisplayName', 'Email', 'Group', 'CustomerId']
+               'DisplayName', 'Email', 'Group', 'CustomerId', 'Email',
+               'Groups', 'Domain', 'Username', 'OrganizationUnit']
 
     return {
         'ContentsFormat': formats['json'],
@@ -326,6 +433,91 @@ def users_to_entry(title, response):
         'ReadableContentsFormat': formats['markdown'],
         'HumanReadable': tableToMarkdown(title, context, headers),
         'EntryContext': {'Account(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
+    }
+
+
+def hide_users_to_entry(title, response):
+    context = []
+    for user_data in response:
+        context.append({
+            'Type': 'Google',
+            'ID': user_data.get('id'),
+            'UserName': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'DisplayName': (user_data.get('name').get('fullName')
+                            if user_data.get('name') and 'fullName' in user_data.get('name') else None),
+            'Email': {'Address': user_data.get('primaryEmail')},
+            'Gmail': {'Address': user_data.get('primaryEmail')},
+            'Group': user_data.get('kind'),
+            'CustomerId': user_data.get('customerId'),
+            'VisibleInDirectory': user_data.get('includeInGlobalAddressList'),
+            'Groups': user_data.get('kind'),
+            'Domain': user_data.get('primaryEmail').split('@')[1],
+            'Username': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
+                                 if (user_data.get('organizations') and len(user_data.get('organizations')) > 0
+                                     and 'name' in user_data.get('organizations')[0]) else None)
+
+        })
+    headers = ['Type', 'ID', 'UserName',
+               'DisplayName', 'Email', 'Group', 'CustomerId', 'Visible In Directory', 'CustomerId',
+               'Groups', 'Domain', 'Username', 'OrganizationUnit']
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, context, headers),
+        'EntryContext': {'Account(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
+    }
+
+
+def autoreply_to_entry(title, response):
+    context = []
+    for autoreply_data in response:
+        context.append({
+            'Type': 'Google',
+            'EnableAutoReply': autoreply_data.get('enableAutoReply'),
+            'ResponseBody': autoreply_data.get('responseBodyPlainText'),
+            'ResponseSubject': autoreply_data.get('responseSubject'),
+            'RestrictToContact': autoreply_data.get('restrictToContacts'),
+            'RestrcitToDomain': autoreply_data.get('restrictToDomain'),
+
+        })
+    headers = ['Type', 'EnableAutoReply', 'ResponseBody',
+               'ResponseSubject', 'RestrictToContact', 'RestrcitToDomain']
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, context, headers),
+        'EntryContext': {'Autoreply(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
+    }
+
+
+def mail_results_to_entry(title, response):
+    context = []
+    for mail_results_data in response:
+        context.append({
+            'Type': 'Google',
+            'MessageId': mail_results_data.get('id'),
+            'Label': mail_results_data.get('labelIds'),
+            'ThreadId': mail_results_data.get('threadId'),
+        })
+    headers = ['Type', 'MessageId', 'Label',
+               'ThreadId']
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, context, headers),
+        'EntryContext': {'SentMail(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
     }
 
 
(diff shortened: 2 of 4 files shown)


roysagi:
fetch-incidents BC possible issue:

Before the change the 'Date' field of the mail was taken from the email payload itself - it looked like so: "Mon, 19 Aug 2019 09:21:33 +0300".

This string already takes UTC time into account - meaning the GMT is "Mon, 19 Aug 2019 *6*:21:33" (3 hours backwards - as indicated by the "+0300" part of the string = this is the correction made to bring it to local time).

Now I see that this time string is sent to the function parse_time() which implements the UTC correction again. Considering the former example, it will move it to: "Mon, 19 Aug 2019 *12*:21:33" - another 3 hours forward.

Not only that it seems the correction is implemented *once more* in the Demisto server itself so when the incident is created it has a 'Date' field of: "Mon, 19 Aug 2019 *15*:21:33"

for instance:
This is a mail I sent to test this issue
image

and this is the incident created from it, note the 'Occurred' field:
image

In order not to create BC, I've created the same original string ("Mon, 19 Aug 2019 09:21:33 +0300") in the same format using the timestamp that we get in the HTTP response, thus eliminating the reliance on the way the date string was formatted in the mail itself.

Now we have two options:
1) Leave it as is meaning the time indicated in the Demisto incidents is incorrect.
2) I can fix this "time jump" mistake - simply by *sending the GMT* time to Demisto and it will correct itself.
But, fixing it will probably create a BC issue in the day we release it - now incidents will get the actual time (as originally intended I presume) but this might create a time when new incidents cannot be created for a while.

Consider this: 
I sent my mail at 9:21 ( = 6:21 GMT), it got transformed to 12:21 using the parse_time function and this is set as the last fetch time. 
now I correct the bug, and when I send a new mail at 9:31, I will send the GMT time (knowing it will be corrected by Demisto), meaning the time will be set to 6:31 and *no incident will be made because the last fetch time is set to 12:21!*

@ronykoz @yaakovi - we need to discuss what should be done...

roysagi:
Talked to @glicht about the issue - and the correction is implemented as follows:
* The date displayed in the incident 'occurred' time field will be the correct time.
* Internally the last_fetch is set to the previous time before the correction (GMT + 2*UTC), so this change will not cause the integration to miss incoming incidents - no BC issues.

i've added comments near the date changes for anyone who might approach this integration in the future.

Integrations/Gmail/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+Added 6 new commands:
```
yaakovi: fix file format as specified in the docs(https://github.com/demisto/internal-content/tree/master/documentation/change-log)

Integrations/Gmail/Gmail.py
```diff
@@ -157,6 +163,72 @@ def parse_mail_parts(parts):
     return body, html, attachments
 
 
+def days_of_the_week(day):
+    if day == 1:
```
yaakovi: use a mapping dictionary

Integrations/Gmail/Gmail.py
```diff
@@ -157,6 +163,72 @@ def parse_mail_parts(parts):
...
+    elif day == 5:
+        return 'Fri'
+    elif day == 6:
+        return 'Sat'
+    elif day == 7:
+        return 'Sun'
+
+
+def month_abbr(mon_num):
+    if mon_num == 1:
```
yaakovi: same here

Integrations/Gmail/Gmail.py
```diff
@@ -259,12 +361,13 @@ def emails_to_entry(title, raw_emails, format_data, mailbox):
         'Contents': emails,
         'ReadableContentsFormat': formats['markdown'],
         'HumanReadable': tableToMarkdown(title, emails, headers[format_data]),
-        'EntryContext': {'Gmail(val.ID && val.ID == obj.ID)': emails}
+        'EntryContext': {'Gmail(val.ID && val.ID == obj.ID)': emails,
+                         'Email(val.Date == obj.Date && val.Subject == obj.Subject)': context_email}
```
yaakovi: `context_email` contain only the last email (in case of a list, you ignore all except the last one).

Integrations/Gmail/Gmail.py
```diff
@@ -315,9 +421,56 @@ def users_to_entry(title, response):
...
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
+                                 if (user_data.get('organizations') and len(user_data.get('organizations')) > 0
+                                     and 'name' in user_data.get('organizations')[0]) else None)
+
+        })
+    headers = ['Type', 'ID', 'UserName',
+               'DisplayName', 'Email', 'Group', 'CustomerId', 'Email',
+               'Groups', 'Domain', 'Username', 'OrganizationUnit']
+
+    return {
```
yaakovi: update to `return_outputs`

Integrations/Gmail/Gmail.py
```diff
@@ -315,9 +421,56 @@ def users_to_entry(title, response):
...
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, context, headers),
+        'EntryContext': {'Account(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
+    }
+
+
+def hide_users_to_entry(title, response):
```
yaakovi: it's the same as `users_to_entry` except for a single field, join them together.

Integrations/Gmail/Gmail.py
```diff
@@ -315,9 +421,56 @@ def users_to_entry(title, response):
...
+            'Username': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
+                                 if (user_data.get('organizations') and len(user_data.get('organizations')) > 0
+                                     and 'name' in user_data.get('organizations')[0]) else None)
+
         })
     headers = ['Type', 'ID', 'UserName',
-               'DisplayName', 'Email', 'Group', 'CustomerId']
+               'DisplayName', 'Email', 'Group', 'CustomerId', 'Visible In Directory', 'CustomerId',
```
yaakovi: 'Visible In Directory' does not appear in the dictionary and will always be empty, use the `headerTransform` in the `tableToMarkdown` in order to use the spacing.

Integrations/Gmail/Gmail.py
```diff
@@ -468,6 +668,201 @@ def get_user(user_key, view_type, projection, customer_field_mask=None):
...
+    command_args = {
+        'userKey': user_key if user_key != 'me' else ADMIN_EMAIL,
+        'body': {
+            'password': password,
+        }}
+    credentials = get_credentials(additional_scopes=['https://www.googleapis.com/auth/admin.directory.user'])
+    service = discovery.build('admin', 'directory_v1', credentials=credentials)
+    service.users().update(**command_args).execute()
+
+    return 'User %s password has been set.' % (command_args['userKey'],)
```
yaakovi: don't use this kind of string formatting, use `.format()`

Integrations/Gmail/Gmail.py
```diff
@@ -1105,6 +1500,12 @@ def main():
         'gmail-add-delete-filter': add_delete_filter_command,
         'gmail-list-filters': list_filters_command,
         'gmail-remove-filter': remove_filter_command,
+        'gmail-hide-user-in-directory': hide_user_command,
+        'gmail-set-password': set_user_password_command,
+        'gmail-get-autoreply': get_autoreply_command,
+        'gmail-set-autoreply': set_autoreply_command,
+        'gmail-delegate-user-mailbox': delegate_user_mailbox_command,
+        'gmail-send-mail': send_mail_command,
```
yaakovi: also add the generic `send-mail` command as an alias for the `gmail-send-mail`

Integrations/Gmail/Gmail.py
```diff
@@ -165,8 +237,17 @@ def get_email_context(email_data, mailbox):
...
     parsed_body = base64.urlsafe_b64decode(body)
+    context_email = {}  # type: Dict
+    if email_data.get('internalDate') is not None:
+        utc, delta_in_seconds = utc_extract(str(headers.get('date', '')))
+        base_time = datetime.fromtimestamp((int(str(email_data.get('internalDate'))[:10]))) + timedelta(seconds=delta_in_seconds)
+        day = days_of_the_week(int(base_time.isoweekday()))  # gets day of week from date
+        base_time = day + ", " + time_reformat(str(base_time)) + " " + utc
+    # return_error(str(headers.get('date', ''))+" ### "+base_time)
+    else:
+        base_time = None
```
glicht: Not sure you want to set this to None. Might want to fallback to original behaviour in this case. Also, add a warning as if I understand correctly it shouldn't happen.

Integrations/Gmail/Gmail.py
```diff
@@ -187,37 +268,57 @@ def get_email_context(email_data, mailbox):
         # only for incident
         'Cc': headers.get('cc', []),
         'Bcc': headers.get('bcc', []),
-        'Date': headers.get('date', ''),
+        'Date': unicode(base_time),
```
glicht: Why are you using unicode? In case base_time == None you will have `None` string.

Integrations/Gmail/Gmail.py
```diff
@@ -187,37 +268,57 @@ def get_email_context(email_data, mailbox):
...
+        'Format': headers.get('content-type', '').split(';')[0],
+        'Subject': headers.get('subject'),
+        'From': headers.get('from'),
+        'To': headers.get('to'),
+        # only for format 'full'
+        'Body/Text': unicode(parsed_body, 'utf-8'),
+
+        'CC': headers.get('cc', []),
+        'BCC': headers.get('bcc', []),
+        'Date': unicode(base_time),
```
glicht: Same comment as above about unicode

Integrations/Gmail/Gmail.py
glicht: Why not add to Email Context the ID field to make sure we rely on unique values as is done with Gmail Context?

Integrations/Gmail/Gmail.py
```diff
@@ -315,9 +421,56 @@ def users_to_entry(title, response):
...
+            'Group': user_data.get('kind'),
+            'CustomerId': user_data.get('customerId'),
+            'VisibleInDirectory': user_data.get('includeInGlobalAddressList'),
+            'Groups': user_data.get('kind'),
+            'Domain': user_data.get('primaryEmail').split('@')[1],
+            'Username': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
+                                 if (user_data.get('organizations') and len(user_data.get('organizations')) > 0
+                                     and 'name' in user_data.get('organizations')[0]) else None)
```
glicht: Seems like duplicate code with users_to_entry. Try to consolidate.

Integrations/Gmail/Gmail.py
```diff
@@ -315,9 +421,56 @@ def users_to_entry(title, response):
             'Gmail': {'Address': user_data.get('primaryEmail')},
             'Group': user_data.get('kind'),
             'CustomerId': user_data.get('customerId'),
+            'Groups': user_data.get('kind'),
+            'Domain': user_data.get('primaryEmail').split('@')[1],
+            'Username': (user_data.get('name').get('givenName')
+                         if user_data.get('name') and 'givenName' in user_data.get('name') else None),
+            'OrganizationUnit': (user_data.get('organizations')[0].get('name')
```
glicht: Why only the first organization?

roysagi:
Found another date bug:
if mails a received from different time zones, this could create a situation where some mails are not displayed as incidents.
I checked this behaviour and it happens in the original integration as well.

The only way to fix this bug is to use the GMT timezone as the setter for the Last_fetch parameter. 
This requires the Last_fetch to be reset and @ronykoz came up with a good idea to simply create a new key in the in the LastRun dictionary - this will silently reset the Last_fetch with no need for action from the costumers.

Only side-effect I can think of is that some mails could create 2 incident the day this fix is released.

Integrations/Gmail/Gmail.py
```diff
@@ -1028,22 +1330,297 @@ def remove_filter(user_id, _id):
...
+        entry_number += 1
+
+    return attachments
+
+
+def handle_html(htmlBody):
+    """
+    Extract all data-url content from within the html and return as separate attachments.
+    Due to security implications, we support only images here
+    We might not have Beautiful Soup so just do regex search
```
ronykoz: ?

Integrations/Gmail/Gmail.py
```diff
@@ -1028,22 +1330,297 @@ def remove_filter(user_id, _id):
...
-
-    # handle first time fetch
+    last_fetch = last_run.get('gmt_time')
+    # handle first time fetch - gets current GMT time -1 day
     if last_fetch is None:
-        last_fetch = datetime.now() - timedelta(days=1)
-    else:
-        last_fetch = datetime.strptime(
-            last_fetch, '%Y-%m-%dT%H:%M:%SZ')
+        last_fetch = (datetime.utcfromtimestamp(int(time.time())) - timedelta(days=1)).isoformat() + 'Z'
```
ronykoz: Make this value configurable, with default to 1 day

TestPlaybooks/playbook-GmailTest.yml
```diff
@@ -969,8 +969,710 @@ tasks:
...
+  "35":
+    id: "35"
+    taskid: ed3b5ef4-d0e3-46d5-8a3b-7432480c123c
+    type: regular
+    task:
+      id: ed3b5ef4-d0e3-46d5-8a3b-7432480c123c
+      version: -1
+      name: gmail-set-autoreply
+      description: Set auto-reply for the user
+      script: Gmail|||gmail-set-autoreply
```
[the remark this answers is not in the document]
roysagi: yeah

Integrations/Gmail/Pipfile
```diff
@@ -15,6 +15,13 @@ google-api-python-client = "*"
 google-auth-httplib2 = "*"
 google-auth-oauthlib = "*"
 httplib2 = "*"
+urlparse = "*"
```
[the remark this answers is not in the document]
roysagi: I don't think we do because it does work locally and in Circle without an updated docker.
The pipfile is used for the unit test

Integrations/Gmail/Gmail.py
```diff
@@ -157,6 +165,29 @@ def parse_mail_parts(parts):
     return body, html, attachments
 
 
+def utc_extract(time_from_mail):
```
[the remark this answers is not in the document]
roysagi: it isn't really a general function - it is very much focused on extracting it from date string in mail

Integrations/Gmail/Gmail.py
```diff
@@ -165,8 +196,20 @@ def get_email_context(email_data, mailbox):
...
+    context_gmail = {}  # type: Dict
+    if email_data.get('internalDate') is not None:
+        base_time = create_base_time(email_data.get('internalDate'), str(headers.get('date', '')))
 
-    context = {
+    else:
+        # in case no internalDate field exists will revert to extracting the date from the email payload itself
+        # Note: this should not happen in any command other than other than gmail-move-mail which doesn't return the
+        # email payload nor internalDate
+        LOG("No InternalDate timestamp found - getting Date from mail payload - msg ID:" + str(email_data['id']))
```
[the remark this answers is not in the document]
roysagi: this could be used for fetch as well - it is mostly for future debugging just in case something goes wrong

Integrations/Gmail/Gmail.py
```diff
@@ -187,32 +230,59 @@ def get_email_context(email_data, mailbox):
...
-        'Date': headers.get('date', ''),
+        'Date': base_time,
+        # 'Date': headers.get('date', ''),
         'Html': None,
     }
 
-    if 'text/html' in context['Format']:  # type: ignore
-        context['Html'] = context['Body']
-        context['Body'] = html_to_text(context['Body'])
+    context_email = {
```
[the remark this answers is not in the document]
roysagi: context_gmail is what originally existed - all of the fields are as they were before in order not to break BC.
context_email was added to fit the current context standard.

Integrations/Gmail/Gmail.py
```diff
@@ -315,20 +406,99 @@ def users_to_entry(title, response):
...
+        'HumanReadable': tableToMarkdown(title, context, headers, removeNull=True),
         'EntryContext': {'Account(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': context}
     }
 
 
+def autoreply_to_entry(title, response, user_id):
+    autoreply_context = []
+    for autoreply_data in response:
+        autoreply_context.append({
+            'Type': 'Google',
```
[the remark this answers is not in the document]
roysagi: Originally because AutoReply was it's own thing. Now it's under Accoung.Gmail so I guess it's pointless - I'l take it out

Integrations/Gmail/Gmail.py
```diff
@@ -315,20 +406,99 @@ def users_to_entry(title, response):
...
+        "AutoReply": autoreply_context
+    }
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': autoreply_context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, autoreply_context, headers, removeNull=True),
+        'EntryContext': {'Account.Gmail(val.Address == obj.Address)': account_context}
```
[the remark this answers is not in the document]
roysagi: When we talked we decided on Account.Gmail because the auto-replay is account based.
It is changeable but I do think it makes sense it stays there

Integrations/Gmail/Gmail.py
```diff
@@ -315,20 +406,99 @@ def users_to_entry(title, response):
...
+               'ThreadId']
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': gmail_context,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': tableToMarkdown(title, gmail_context, headers, removeNull=True),
+        'EntryContext': {'Gmail.SentMail(val.ID && val.Type && val.ID == obj.ID && val.Type == obj.Type)': gmail_context,
+                         'Email(val.ID && val.ID == obj.ID)': email_context}
```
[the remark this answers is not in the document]
roysagi: MailSender does not put any outputs.
I do think it makes sense to put something out for the user and again I used the context standard here for Email.

Integrations/Gmail/Gmail.py
roysagi: Taken straight from MailSender - It seems to extract attachments which are embedded in the htmlBody.

Integrations/Gmail/Gmail.py
```diff
@@ -289,46 +375,132 @@ def mail_to_incident(msg, service, user_key):
...
-                            if user_data.get('name') and 'fullName' in user_data.get('name') else None),
+            'UserName': username,
+            'DisplayName': display,
             'Email': {'Address': user_data.get('primaryEmail')},
             'Gmail': {'Address': user_data.get('primaryEmail')},
             'Group': user_data.get('kind'),
             'CustomerId': user_data.get('customerId'),
+            'Groups': user_data.get('kind'),
+            'Domain': user_data.get('primaryEmail').split('@')[1],
+            'Username': username,
```
yaakovi: move this next to original and add a comment specifying this is new standard context

Integrations/Gmail/Gmail.py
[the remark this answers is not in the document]
yaakovi: follow the suggested change

Integrations/Gmail/Gmail.py
```diff
@@ -468,6 +640,139 @@ def get_user(user_key, view_type, projection, customer_field_mask=None):
...
+    if delegate_token:  # guardrails-disable-line
+        command_args = {
+            'userId': user_id if user_id != 'me' else ADMIN_EMAIL,
+            'body': {
+                'delegateEmail': delegate_email,
+            }
+        }
+
+        service.users().settings().delegates().create(**command_args).execute()
+        return 'Email %s has been delegated' % delegate_email
```
yaakovi: use format instead of old string formatting

Integrations/Gmail/Gmail.py
```diff
@@ -647,7 +952,8 @@ def search_command(mailbox=None):
     mails, q = search(user_id, subject, _from, to, before, after, filename, _in, query,
                       fields, label_ids, max_results, page_token, include_spam_trash, has_attachments)
 
-    return emails_to_entry('Search in %s:\nquery: "%s"' % (mailbox, q, ), mails, 'full', mailbox)
+    res = emails_to_entry('Search in %s:\nquery: "%s"' % (mailbox, q, ), mails, 'full', mailbox)
```
yaakovi: same

Integrations/Gmail/Gmail.py
```diff
@@ -1028,22 +1333,361 @@ def remove_filter(user_id, _id):
...
-    # handle first time fetch
+    last_fetch = last_run.get('gmt_time')
+    # handle first time fetch - gets current GMT time -1 day
     if last_fetch is None:
-        last_fetch = datetime.now() - timedelta(days=1)
-    else:
-        last_fetch = datetime.strptime(
-            last_fetch, '%Y-%m-%dT%H:%M:%SZ')
-    current_fetch = last_fetch
+        last_fetch = (datetime.utcfromtimestamp(int(time.time())) - timedelta(days=FETCH_TIME)).isoformat() + 'Z'
```
yaakovi: be consistent with other integrations, use `parse_date_range` on the value of `FETCH_TIME`

Integrations/Gmail/Gmail.py
```diff
@@ -1028,22 +1333,361 @@ def remove_filter(user_id, _id):
...
 def fetch_incidents():
     params = demisto.params()
     user_key = params.get('queryUserKey')
     user_key = user_key if user_key else ADMIN_EMAIL
     query = '' if params['query'] is None else params['query']
     last_run = demisto.getLastRun()
-    last_fetch = last_run.get('time')
-
-    # handle first time fetch
+    last_fetch = last_run.get('gmt_time')
```
yaakovi: this will cause customer duplicate incidents since after the update it will not find the `gmt_time` as the last run contains only the `time` value.

@kirbles19 
This MUST be documented properly in the RN

Integrations/Gmail/Gmail.yml
```diff
@@ -1,37 +1,28 @@
+category: Authentication
```
yaakovi: i did not gone over this file until the order of fields is fixed.

Integrations/Gmail/Gmail_description.md
```diff
@@ -34,4 +34,22 @@ To get an Immutable Google Apps ID (or customerId):
 
 You will see there URLs in the format:
 [https://accounts.google.com/o/saml2/idp?idpid=Cxxxxxxxx](https://accounts.google.com/o/saml2/idp?idpid=Cxxxxxxxx)
-Cxxxxxxxx is your Immutable Google Apps ID (customerId).
\ No newline at end of file
+Cxxxxxxxx is your Immutable Google Apps ID (customerId).
+
+In order to get auto-replay messages from a user, please authorize this api too:
```
yaakovi: move all of those above line 28 to be near the rest of the commands

Integrations/Gmail/Gmail.py
```diff
@@ -165,8 +208,20 @@ def get_email_context(email_data, mailbox):
...
+    if email_data.get('internalDate') is not None:
+        base_time = create_base_time(email_data.get('internalDate'), str(headers.get('date', '')))
 
-    context = {
+    else:
+        # in case no internalDate field exists will revert to extracting the date from the email payload itself
+        # Note: this should not happen in any command other than other than gmail-move-mail which doesn't return the
+        # email payload nor internalDate
+        LOG("No InternalDate timestamp found - getting Date from mail payload - msg ID:" + str(email_data['id']))
+        LOG.print_log()
```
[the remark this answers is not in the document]
roysagi: Consulted with @ronykozon this matter - because it can appear in fetch incidents - a LOG is necessary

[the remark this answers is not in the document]
yaakovi: you can use `demisto.info` in `fetch-incidents` as well

ronykoz:
@kirbles19 could you go over 
https://github.com/demisto/content/blob/gmail_add_commands/Integrations/Gmail/Gmail.yml

https://github.com/demisto/content/blob/gmail_add_commands/Integrations/Gmail/Gmail_description.yml

https://github.com/demisto/content/blob/gmail_add_commands/Integrations/Gmail/CHANGELOG.yml

Integrations/Gmail/Gmail.yml
```diff
@@ -4,6 +4,7 @@ commonfields:
...
...bR7U+wDqJgSQS8o8eEzxI+gbEEA27efBjOOfQAAZtZsHT/yj5BkJIKc282De8U8ggKzi58HM459AAHnFzoPtPm02AQHkFjMP5h//BALILnweTPnE+bYIIL/QeXDQh8pkQwAFhM2DRcY/gQBKCJkHmz5YLhMCKKNpHiw1/gkEUMjgebDY+CcQQCmD5sFy459AAMX0nwfDP2BeHwGU028efP4zZAwQQEm95sGy459AAEXJebDw+CcQQFmXzoPXFB7/BAIobPE8WHz8EwiguOfnQYPxTyCA8i7Mgxbjn0AABubnQZPxTyAAC3PzoM34JxCAidWrVtmMfwIBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBOAcAThHAM4RgHME4BwBOEcAzhGAcwTgHAE4RwDOEYBzBODc0AdwaNgvYCwygBPW5xWOE0CSyAAesz6vsJEAkkQGsNz6vMIyAkgSGcDd1ucVxgkgSWQAh63PKywhgCSRAZyyPq8wNkMAKeICmDlrfV7pJgJIERfAaevj9jBJACniArja+rg9TGwggARRAUztsT5uL5cRQIKoANZan7anLVME0F5MALv+aX3a3v5FAO3FBHCF9WH7mN5JAK1FBPDcfuvD9nPm3wTQVngAu05an7W/3dsJoKXgALb/x/qog5wjgJaCA5i0Pulg/w39PcD6oF0TGMDMtdYHbbJ7FwG0ERbA1k7//j/v5I0E0EJQAM8dsT5miNWTWwkgWkAAU9umrU8Z6MB48y5gfcauaQxg6nBH3//raeJ/txFAlMEBzJy+tZP7zyBn7zh8bPOKQwQQpl8Ah44/eWztqVusjwcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQBH/BwUO4nZfxiAEAAAAAElFTkSuQmCC
```
ronykoz: remove

Integrations/Gmail/Gmail.yml
```diff
@@ -34,20 +35,28 @@ configuration:
...
   type: 8
+  required: false
 - display: Incident type
   name: incidentType
   defaultvalue: ""
   type: 13
   required: false
+- display: First fetch timestamp, in days.
+  name: fetch_time
+  defaultvalue: 1 days
```
ronykoz:
```suggestion
  defaultvalue: 1
```

Integrations/Gmail/Gmail.yml
```diff
@@ -838,9 +1082,265 @@ script:
...
+      required: true
+      description: The email address to remove from delagation
+    description: Removes a delegate from the mailbox, without sending any verification
+      email. The delegate user must be a member of the same G Suite organization as
+      the delegator user.  Note that a delegate user must be referred to by their
+      primary email address, and not an email alias.
   dockerimage: demisto/google-api:1.0
   isfetch: true
   runonce: false
-tests:
```
ronykoz: Why did you remove it?

Integrations/Gmail/Gmail.yml
```diff
@@ -838,9 +1082,265 @@ script:
...
+      description: The email Mailbox
+      type: String
+    description: Send mail using Gmail
+  - name: gmail-remove-delegated-mailbox
+    arguments:
+    - name: user-id
+      required: true
+      description: The user's email address. The special value me can be used to indicate
+        the authenticated user.
+    - name: remove-mail
```
ronykoz:
```suggestion
    - name: removed-mail
```

Integrations/Gmail/Gmail.py
```diff
@@ -24,7 +32,7 @@
 SCOPES = ['https://www.googleapis.com/auth/admin.directory.user.readonly']
 PROXY = demisto.params().get('proxy')
 DISABLE_SSL = demisto.params().get('insecure', False)
-
+FETCH_TIME = demisto.params().get('fetch_time', '1 days')
```
ronykoz:
```suggestion
FETCH_TIME = demisto.params().get('fetch_time', '1')
```

roysagi: @yaakovi said I should keep the convention of using '1 days' in fetch time and the parse_date_range function

Integrations/Gmail/Gmail.yml
roysagi: I did not - I'll add again though it

Integrations/Gmail/Gmail.py
```diff
@@ -1382,35 +1378,31 @@ def transient_attachments(transientFile, transientFileContent, transientFileCID)
...
+    if transientFileContent is None:
+        transientFileContent = []
+
+    if transientFileCID is None:
+        transientFileCID = []
+
     attachments = []
-    entry_number = 0
-    for file_name in transientFile:
+    for triplet in it.izip_longest(transientFile, transientFileContent, transientFileCID):
```
yaakovi: also notice that you are changing the default here from empty string to `None` for `file_data` and `file_cid`, is that ok?
```suggestion
    for file_name, file_data, file_cid in it.izip_longest(transientFile, transientFileContent, transientFileCID):
```

richardbluestone:
Hi I'm going through the yml file and noticed that some arguments seem to be missing the "required" key:value. For example, in the **gmail-list-users** command,  the projection, domain, customer,  fields, etc do not say whether they are required.

yaakovi:
@richardbluestone 
incase the `required` field is not specified in the yml, it means that the argument is optional.

yaakovi:
@roysagi please trigger a build without the BC-checks

richardbluestone:
I've updated the yml and md files. As discussed earlier, the Get a New Private Key section in the Zen Desk documentation needs updating

Integrations/Gmail/Gmail.py
roysagi: I had a look at the RN after the approval and issue is not mentioned - @kirbles19 @richardbluestone - if you can please take another look and add this important clarification.
https://github.com/demisto/content/blob/gmail_add_commands/Integrations/Gmail/CHANGELOG.md

PR #4151, opened 2019-08-20
better logging of "Mismatch between playbookID...."

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15038

## Description
we hold all the conf.json file and can better log the instance name and playbook ID so the developer can see if the error is related to the mismatch or to a schema in the TPB.

   - No

### DIFF
--- a/Tests/test_integration.py
+++ b/Tests/test_integration.py
@@ -12,9 +12,6 @@ DEFAULT_TIMEOUT = 60
 DEFAULT_INTERVAL = 20
 ENTRY_TYPE_ERROR = 4
 
-INC_CREATION_ERR = 'Failed to create incident. Possible reasons are:\nMismatch between playbookID in conf.json and ' \
-                   'the id of the real playbook you were trying to use, or schema problems in the TestPlaybook.'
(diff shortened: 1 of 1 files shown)


bakatzir:
@yuvalbenshalom 
```
------ Test playbook: Bluecat Address Manager non existent test with integration(s): BluecatAddressManager start ------ (Mock: Recording)
Proxy process up and running. Took 2 seconds
Configuring instance for BluecatAddressManager (instance name: BluecatAddressManager)
Create integration BluecatAddressManager succeed
Failed to create incident for integration names: [u'BluecatAddressManager'] and playbookID: Bluecat Address Manager non existent test.Possible reasons are:
Mismatch between playbookID in conf.json and the id of the real playbook you were trying to use,or schema problems in the TestPlaybook.
Failed: playbook: Bluecat Address Manager non existent test with integration(s): BluecatAddressManager failed
------ Test playbook: Bluecat Address Manager non existent test with integration(s): BluecatAddressManager end ------
```

PR #4159, opened 2019-08-20
Removed indicator fields filename and signature

DeanArbel:
## Status
Ready

## Description
The indicator fields `Signature` and `File Name` were added to the content repository by mistake in https://github.com/demisto/content/pull/4047.
Because of an unrelated bug (that is being investigated by me), these fields do not appear in indicator fields tab. In another manual test they did appear, but had no mapping.

### No filename:
image
#### Filename without mapping:
image

### No Signature:
image
#### Signature without mapping:
image


DeanArbel:
Implemented in https://github.com/demisto/content/pull/4203

PR #4167, opened 2019-08-20
Add files via upload

vibhuabharadwaj:
## Status
Ready

## Description
Script to remotely handle Demisto Lists using the API with functionality to create, rewrite, append and delete the lists. The script contains just functions which can be used to handle lists.

4.5.x

## Does it break backward compatibility
No

##Testing
Was tested in 4.5 and in customer environment, and was found to be working fine.


Itay4:
if this is no meant to run in demisto, then why do you want to add it to demisto ?

PR #4172, opened 2019-08-21
Added context paths to standard outputs

DeanArbel:
## Status
Ready

## Description
Added new context standards outputs to our docs.

### DIFF
--- a/docs/context_standards/README.MD
+++ b/docs/context_standards/README.MD
@@ -21,7 +21,6 @@ The following is the format for a file.
         "SHA1": "STRING, SHA1 Hash of the file",
         "SHA256": "STRING, SHA256 hash of the file",
         "SHA512": "STRING SHA512 hash of the file",
-        "Blocked": "BOOLEAN Whether the indicator is blocked, for example, in the firewall or blacklisted",
         "SSDeep": "STRING, The ssdeep value for the file (same as we display in file entries)",
         "Extension": "STRING, the file extension e.g. xls",
         "Type": "STRING, File type as determined by libmagic (same as we display in file entries)",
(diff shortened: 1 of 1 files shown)


docs/context_standards/README.MD
```diff
@@ -305,6 +310,27 @@ The following is the format for a process.
...
+    "ID": "STRING, The unique ID within the tool retrieving the host.",
+    "DHCPServer": "STRING, DHCP Server.",
+    "IP": "STRING, IP Address of the host.",
+    "MACAddress": "STRING, Host MAC Address.",
+    "Memory": "STRING, Host memory.",
+    "Model": "STRING, Host model.",
+    "OS": "STRING, Host operating system.",
+    "OSVersion": "STRING, Host operating system version.",
+    "Processor": "STRING, Host processor.",
+    "Processors": "STRING, Host processors"
```
yaakovi: why do we have both single and plural of processor?
```suggestion
    "Processors": "STRING, Host processors." 
```

DeanArbel: Because the second field is the Processors number - I'll update the description to better reflect this

DeanArbel:
@kirbles19 Can you please review the (single) file changed?
I'm not sure you had a chance to review this documentation in the page, but if not then now would be a good time to do it, since we're updating our integrations according to this standard, and now we're in the middle of an epic to update all existing integrations.

docs/context_standards/README.MD
```diff
@@ -305,6 +305,26 @@ The following is the format for a process.
...
+## Host
+The following is the format for a host.
+```python
+"Host": {
+    "Domain": "STRING, Domain of the host.",
+    "Hostname": "STRING, Name of the host.",
+    "BIOVersion": "STRING, BIOS version of the host.",
+    "ID": "STRING, The unique ID within the tool retrieving the host.",
+    "DHCPServer": "STRING, DHCP Server.",
+    "IP": "STRING, IP Address of the host.",
```
kirbles19: "IP": "STRING, IP address of the host.",

kirbles19:
@DeanArbel One comment made inline, all done.

docs/context_standards/README.MD
DeanArbel: @kirbles19 I'll fix that. Except for that, are the descriptions ok?

kirbles19:
Updated.

PR #4177, opened 2019-08-21
GetByIncidentId Script

LPettay:
## Description
This is the opposite of SetByIncidentId which was written by me several months ago. Gets specific context data from another incident and sets it into the current incident.

Without this script there isn't a user-friendly way to grab context from other incidents. The only way I know if is the hidden getContext API command which will require you to write a custom script to pull out the exact keys you want.

This script will make it easy to pull a specific key, and set it into any value the user wants.

## Screenshots
Screenshot of context key stored in incident A (ID:61)
image

Running the following command in another incident will grab that value from the other incident and store it into its own context:
`!GetByIncidentId id=61 getKey=IncidentA.GetByIncidentId.Test setKey=IncidentB.KeyFromIncidentA`

image

Leaving setKey blank will store it in the same path as it was stored in the original incident.


Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
+commonfields:
+  id: a85a213b-2c41-40d7-8d31-ccef5008e980
```
Itay4:
```suggestion
  id: GetByIncidentId
```

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+  version: 16
```
Itay4:
```suggestion
  version: -1
```

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+  entry_context = {set_key: res}
+
+  demisto.results({'Type' : entryTypes['note'],
+                   'ContentsFormat': formats['text'],
+                   'EntryContext' : entry_context,
+                   'Contents': res
+  })
+type: python
+tags:
+- DemistoAPI
```
Itay4: why do we need this ?

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+comment: Gets a value from the given incident's context.
+enabled: true
+args:
+- name: id
+  default: true
+  description: Incident to get context values from (Default is current incident)
+- name: getKey
+  required: true
+  description: The key to get
+- name: setKey
```
Itay4: why are we setting keys?

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  try:
+      context = demisto.executeCommand("getContext", {"id":incident_id})[0]['Contents']['context']
+      res = reduce(lambda x,y: x[y], keys, context)
+  except:
+      res = "Cannot find {} in Incident #{}".format(key, incident_id)
+
+  entry_context = {set_key: res}
+
+  demisto.results({'Type' : entryTypes['note'],
```
Itay4: use `return_outputs`

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+  key = demisto.args()['getKey']
+  set_key = demisto.args().get('setKey', key)
+
+  keys = key.split('.')
+
+  try:
+      context = demisto.executeCommand("getContext", {"id":incident_id})[0]['Contents']['context']
+      res = reduce(lambda x,y: x[y], keys, context)
+  except:
+      res = "Cannot find {} in Incident #{}".format(key, incident_id)
```
Itay4: this should be returned as error

Scripts/script-GetByIncidentId.yml
```diff
@@ -0,0 +1,41 @@
...
+type: python
+tags:
+- DemistoAPI
+comment: Gets a value from the given incident's context.
+enabled: true
+args:
+- name: id
+  default: true
+  description: Incident to get context values from (Default is current incident)
+- name: getKey
```
Itay4:
```suggestion
- name: get_key
```

Scripts/script-GetByIncidentId.yml
LPettay: It's in case we don't want to use the same path as in the original (e.g. avoid overwriting, or we just want to rename). It's optional.

Itay4:
@LPettay any update here?

LPettay:
@Itay4 I'm not sure how to write tests for this script. It seems like I would need to overwrite the behavior of getContext so I can pull mock data. Is there a better way to go about this?

Itay4:
@LPettay you can write a unit test for that
have a look here https://github.com/demisto/content/tree/master/docs/tests/unit-testing

Scripts/script-GetByIncidentId.yml
LPettay: Not needed, I've removed it

Scripts/script-GetByIncidentId.yml
LPettay: Done

Scripts/script-GetByIncidentId.yml
LPettay: Done

LPettay:
I'm not sure if my package format is correct, but all the necessary changes have been made. I finished the test cases as well.

Scripts/GetByIncidentId/GetByIncidentId.yml
```diff
@@ -0,0 +1,56 @@
+commonfields:
+  id: GetByIncidentId
+  version: -1
+name: GetByIncidentId
+script: |
+  from functools import reduce
```
Itay4: remove script from the yml
it should be `script: '-'`

Scripts/GetByIncidentId/GetByIncidentId.yml
```diff
@@ -0,0 +1,56 @@
...
+  default: true
+  description: Incident to get context values from (Default is current incident)
+- name: get_key
+  required: true
+  description: The key to get
+- name: set_key
+  description: The key to set (will default to get_key)
+scripttarget: 0
+runonce: false
+dockerimage: demisto/python3:3.7.3.221
```
Itay4:
```suggestion
dockerimage: demisto/python3:3.7.4.1900
```

Itay4:
@kirbles19 please review the yml

kirbles19:
@LPettay Done.

Itay4:
@LPettay note the CircleCI build is failing
1. you should add in the yml the following:
```
tests:
  - No test
```
2. please fix the linting issues

ronykoz:
@LPettay Could you please address the following error that come up from flake8
image

And please pull from master

PR #4180, opened 2019-08-21
[Internal] Added schema tests

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18683

## Description
Added a structure validation test to all schema-relevant files when a schema is changed.

   - No


yuvalbenshalom:
@yaakovi @DeanArbel

PR #4186, opened 2019-08-22
XDR - adding default query when no args are given

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18728

4.1.0

   - No

## Must have
- [x] Tests
- [ ] Code Review


bakatzir:
@kirbles19 can you review the error message added:
`Please provide a query for the incidents.
For example: "!xdr-get-incidents since_creation_time="1 year" sort_by_creation_time="desc" limit=10"`

kirbles19:
@bakatzir adding here: Specify a query for the incidents. For example: "!xdr-get-incidents since_creation_time="1 year" sort_by_creation_time="desc" limit=10".

PR #4195, opened 2019-08-23
Improved logging for mail sender new: test-module

glicht:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17165

## Description
Include protocol level debugging when doing test-module. Allows better understanding any issues encountered during configuration.

Includes a change in CommonServer to support using the LOG object as a file like object so it can be used to replace stderr in smtplib.

All supported

   - No

## Must have
- [x] Tests - See unit tests
- [ ] Code Review

## Additional changes
Updated common_server_docs.py to include demistomock (as demisto) when generating docs.


Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -558,30 +559,69 @@ class IntegrationLogger(object):
 
     def __init__(self, ):
         self.messages = []  # type: list
-
-    def __call__(self, message):
+        self.write_buf = []  # type: list
+        self.replace_strs = []  # type: list
+        if (not os.getenv('COMMON_SERVER_NO_AUTO_REPLACE_STRS') and hasattr(demisto, 'getParam')
```
yaakovi: need to add a comment here about `COMMON_SERVER_NO_AUTO_REPLACE_STRS`

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -301,28 +305,33 @@ def main():
     global SERVER
     FROM = demisto.getParam('from')
     FQDN = demisto.params().get('fqdn')
+    stderr_org = smtplib.stderr  # type: ignore
     try:
-        SERVER = SMTP(demisto.getParam('host'), int(demisto.getParam('port')), local_hostname=FQDN)
+        if demisto.command() == 'test-module':
```
yaakovi: why are we triggering this only for test-module?

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -558,30 +559,69 @@ class IntegrationLogger(object):
...
-                self.messages.append("Failed encoding message with error: {}".format(ex))
+                res = "Failed encoding message with error: {}".format(ex)
+        for s in self.replace_strs:
+            res = res.replace(s, '<XX_REPLACED>')
+        return res
+
+    def __call__(self, message):
+        self.messages.append(self.encode(message))
+
+    def add_repalce_strs(self, *args):
```
yaakovi: logger docstring (for the script helper) should recommend user to call `add_repalce_strs` with the credentials.

we also need to add it to code standard (next to `handle_proxy`)

yaakovi:
@kirbles19 please review CHANGELOG files.

kirbles19:
@yaakovi Done. When approved, please add the `docs-approved` label

Integrations/MailSenderNew/MailSenderNew.py
glicht: for send-mail can affect memory

Scripts/CommonServerPython/CommonServerPython.py
glicht: added comments

Scripts/CommonServerPython/CommonServerPython.py
glicht: added a comment

glicht:
@yaakovi updated the comments

PR #4200, opened 2019-08-25
removed process printing

ShahafBenYakir:
## Status
Ready

## Description
Added processes data to slack messages


Tests/scripts/run_tests.sh
```diff
@@ -21,4 +21,4 @@ USERNAME=$temp
 
 [ -n "${NIGHTLY}" ] && IS_NIGHTLY=true || IS_NIGHTLY=false
```
yaakovi: need to do the same for the `MEM_CHECK` as in regular build this is not define

Tests/test_content.py
```diff
@@ -44,8 +44,12 @@ def options_handler():
     parser.add_argument('-b', '--buildNumber', help='The build number', required=True)
     parser.add_argument('-g', '--buildName', help='The build name', required=True)
     parser.add_argument('-i', '--isAMI', type=str2bool, help='is AMI build or not', default=False)
+    parser.add_argument('-m', '--memCheck', type=str2bool,
+                        help='Should trigger memory checks or not. The slack channel to check the data is: dmst_content_nightly_memory_data',
```
yaakovi: break this line, it fails the build

Tests/test_content.py
```diff
@@ -44,8 +44,12 @@ def options_handler():
     parser.add_argument('-b', '--buildNumber', help='The build number', required=True)
     parser.add_argument('-g', '--buildName', help='The build name', required=True)
     parser.add_argument('-i', '--isAMI', type=str2bool, help='is AMI build or not', default=False)
+    parser.add_argument('-m', '--memCheck', type=str2bool,
+                        help='Should trigger memory checks or not. The slack channel to check the data is: '
+                             'dmst_content_nightly_memory_data',default=False)
```
yaakovi:
```suggestion
                             'dmst_content_nightly_memory_data', default=False)
```

PR #4212, opened 2019-08-26
Add hook validation that a domain/email/file/ip/url command has an a …

bakatzir:
…non required and default argument with the same name. add relevant tests.

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16369

## Must have
- [x] Tests
- [x] Documentation - none needed
- [x] Code Review

### DIFF
--- a/Tests/scripts/hook_validations/tests/integration_test.py
+++ b/Tests/scripts/hook_validations/tests/integration_test.py
@@ -613,8 +613,8 @@ def test_is_default_arguments_non_default():
                     "arguments": [
                         {
                             "name": "file",
-                            "required": "true",
-                            "default": "false"
+                            "required": True,
+                            "default": False
                         },
                         {
                             "name": "verbose"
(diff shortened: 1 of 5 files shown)


bakatzir:
@ronykoz in the .yml the values are false, true
is it converted to boolean? or stays as string?

Tests/scripts/hook_validations/tests/integration_test.py
```diff
@@ -601,3 +620,84 @@ def test_valid_subtype_lies():
...
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+    validator.old_integration = None
+
+    assert validator.is_default_arguments() is False, \
+        "The integration validator did not find invalid arg (needed to be default)"
```
ronykoz:
```suggestion
        "The integration validator did not find invalid arg (needed to be default and not required)"
```

PR #4216, opened 2019-08-26
Build shift

ShahafBenYakir:
## Status
Ready

## Description
- Skipped ParseEmailHeaders test as it fails https://github.com/demisto/etc/issues/18815
- LogRhythm was fixed and was removed from skipped https://github.com/demisto/content/pull/4053
- EclecticIQ Platform has an instance, but now it has issues with the API thus was moved from `NO INSTANCE` section to `skipped Integrations` section.


ShahafBenYakir:
@yuvalbenshalom Not sure what test you mean. This test fails the build and thus was moved to skipped. For that reason, I don't want the build to trigger it.

PR #4228, opened 2019-08-27
AbuseDB - Packaging

orlichter1:
## Status
Ready

## Description
Just packaging of AbuseDB

PR #4229, opened 2019-08-27
Abuse indicators

orlichter1:
## Status
Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/17949

## Description
Fixed Indicators to match the new format
## Related PRs

branch | PR
------ | ------
AbuseDB-Packaging|https://github.com/demisto/content/pull/4228

   - No

- [x] [Docuentation](https://github.com/demisto/etc/issues/19376)


orlichter1:
@yaakovi in the packaging PR I've had to change some things using the plugin. I believe there it'll change to this format so after I'll merge the packaging PR I'll see if there's a need to change. In general, maybe we should run this sorted thing the plugin is using when running the packaging script.

Integrations/AbuseDB/AbuseDB.yml
```diff
@@ -0,0 +1,282 @@
...
+    name: abuseipdb-get-categories
+    outputs:
+    - contextPath: AbuseIPDB.Categories
+      description: List of AbuseIPDB categories
+      type: string
+  isfetch: false
+  runonce: false
+  script: '-'
+  type: python
+  subtype: python3
```
yaakovi: not sure it is python3, looks more like 2

orlichter1: How do you know?
I usually look for the " Exception as e" and by that I know

yaakovi: `Exception as e` works in both versions.
it looks like an old integration (before we switched to python3

orlichter1:
@kirbles19 Should we generate the .yml again? The difference from the previous version is getting the Indicator fields up to standards.

richardbluestone:
> @kirbles19 Should we generate the .yml again? The difference from the previous version is getting the Indicator fields up to standards.

Hi I've reviewed and updated the .yml file. Please generate the docs.

PR #4230, opened 2019-08-27
added new widget for showing new incidents per week

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18865

## Description
New widget for showing new incidents per week

### DIFF
--- a/Scripts/IncidentsPerWeek/CHANGELOG.md
+++ b/Scripts/IncidentsPerWeek/CHANGELOG.md
@@ -1,0 +1,3 @@
+## [Unreleased]
+  - 
+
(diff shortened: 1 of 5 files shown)


ronykoz:
@kirbles19 could you please go over everything but the .py file please?
BTW the empty changelog is meant to be this way, because it is used only for the widget itself

aviadl:
@ronykoz this is not the correct way to do this (I think) 
You can come over an i will show you what i did in some other places

kirbles19:
Done.

yaakovi:
@ronykoz any updates here?

Itay4:
@ronykoz ?

yaakovi:
@ronykoz stale?

PR #4232, opened 2019-08-28
Context paths DBotScore and a few outputs hook validations

bakatzir:
@DeanArbel FYI

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/16390

## Description
In order to make sure our context standard integrations changes for 5.0 will be top-notch,
added a few checks in our hook validations.
Please see error example in the referenced issue.

   - No

## Must have
- [x] Tests
- [x] Documentation - no need
- [ ] Code Review

### DIFF
--- a/Tests/scripts/hook_validations/integration.py
+++ b/Tests/scripts/hook_validations/integration.py
@@ -55,13 +55,16 @@ class IntegrationValidator(object):
         self.is_there_duplicate_params()
         self.is_changed_subtype()
 
+        # will move to is_valid_integration after https://github.com/demisto/etc/issues/17949
+        self.is_outputs_for_reputations_commands_valid()
+
         return self._is_valid
 
     def is_valid_integration(self):
         """Check whether the Integration is valid or not, update the _is_valid field to determine that"""
         self.is_valid_subtype()
         self.is_default_arguments()
-        self.is_outputs_for_reputations_commands_valid()
+
 
         return self._is_valid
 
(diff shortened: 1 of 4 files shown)


Tests/scripts/hook_validations/integration.py
```diff
@@ -87,12 +88,50 @@ def is_default_arguments(self):
...
+
+        return self._is_valid
+
     def is_valid_subtype(self):
         """Validate that the subtype is python2 or python3."""
         type_ = self.current_integration.get('script', {}).get('type')
         if type_ == 'python':
             subtype = self.current_integration.get('script', {}).get('subtype')
-            if not subtype or subtype not in ['python3', 'python2']:
+            if subtype not in ['python3', 'python2']:
```
yaakovi: let's move the subtype options to the constants.py file and use them as a set

bakatzir: @yaakovi fixed also for the script.py

Tests/scripts/hook_validations/tests/integration_test.py
```diff
@@ -701,3 +701,156 @@ def test_is_default_arguments_ok():
...
+                            "description": "The indicator type.",
+                            "type": "string"
+                        },
+                        {
+                            "contextPath": "DBotScore.Vendor",
+                            "description": "Vendor used to calculate the score.",
+                            "type": "string"
+                        },
+                        {
+                            "contextPath": "DBotScore.Sc0re",
```
yaakovi:
```suggestion
                            "contextPath": "DBotScore.Score",
```

bakatzir: @yaakovi it should be sc0re
the assert is False

yaakovi:
as we have a standard for the description syntax, we can verify them as well

ronykoz:
We already talked about it, @bakatzir didn’t you say you will add it here?

bakatzir:
@yaakovi @ronykoz 
indeed. Added that and treat a wrong description as warning, not as an error.

bakatzir:
reference: https://github.com/demisto/etc/issues/17949

PR #4240, opened 2019-08-28
Remedy On Demand - remove trailing slash from login url

Itay4:
## Status
Ready

## Description
Removed trailing slash from login URL, since it caused bad request response returned from remedy.

Any

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
None


anara123:
@Itay4 if there is an open issue , please add link to the issue

Itay4:
there isn't an issue

PR #4241, opened 2019-08-28
Mimecast v2 fix

yardensade:
## Status
In Progress - waiting first for packigify to merge into master
Fixes issue when not having access_key configured blocks the integration from running any command due to auto token refresh that needs it.

Fix: If there is no access key, won't run auto token refresh since it won't work anyway.
Also reflects in test error in case access key is not provided, that it cannot actually test the integration.

### DIFF
--- a/Integrations/MimecastV2/MimecastV2.py
+++ b/Integrations/MimecastV2/MimecastV2.py
@@ -219,6 +219,8 @@ def parse_query_args(args):
 
 
 def test_module():
+    if not ACCESS_KEY:
+        return_error('Cannot test valid connection without the Access Key parameter.')
     list_managed_url()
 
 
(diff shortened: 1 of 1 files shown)


Integrations/MimecastV2/MimecastV2.py
```diff
@@ -219,6 +219,8 @@ def parse_query_args(args):
 
 
 def test_module():
+    if not ACCESS_KEY:
+        return_error('Cannot test valid connection without the Access Key parameter.')
```
yardensade: @kirbles19 could you review this specific line plz :)

yardensade:
Can't run all tests due to instance, but seems to still work, which is important.
image

yardensade:
@kirbles19 please review this, for changes needed i will PR again.

PR #4246, opened 2019-08-29
netskope-alerts raw response fix

bakatzir:
@shkarni validated with the customer the solution since we have no instance.

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18696

   - No

## Must have
- [x] Tests - no instance
- [ ] Code Review


bakatzir:
@kirbles19 
bug fix, only RN is needed:
https://github.com/demisto/content/blob/c2c8be72fccab988075bd3185f19590df3ad791f/Integrations/integration-Netskope_CHANGELOG.md

PR #4259, opened 2019-09-01
Fix beta non pkg

yuvalbenshalom:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/18748

## Description
added beta to display field and beta disclaimer to long description of an packaged beta integrations.

   - No


yuvalbenshalom:
fixed in https://github.com/demisto/content/pull/4302

PR #4264, opened 2019-09-01
Panorama logs multiple args and refactor TPB

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18908
reference: https://github.com/demisto/etc/issues/18246

4.0.0

   - No

## Must have
- [x] Tests
- [x] Code Review

## Dependencies
the panorama Query Logs PB, which was tested with the changes.


Integrations/Panorama/Panorama.py
```diff
@@ -3410,48 +3410,67 @@ def panorama_get_traffic_logs_command():
 ''' Logs '''
 
 
+def build_array_query(query, arg_string, string, operator):
```
yuvalbenshalom: how do you handle a single parameter?

bakatzir: @yuvalbenshalom  argToList will still make a list of one item in it.
and it's handled in the function.

PR #4269, opened 2019-09-01
Add 'Sort' param to Incident Severity by Type Widget

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#18400

## Description
Now sorts the severity type by ascending values.

## Screenshots
Screen Shot 2019-09-01 at 14 12 30

   - No


amshamah419:
Should be pretty quick to implement. Should we identify if there are other widgets that need this?

Bars92:
Not much. Probably only phase widget will suffer the same issue

PR #4278, opened 2019-09-01
Fix intsights bug when URL is wrong

adi88d:
## Status
Ready
## Related Issues
fixes: https://github.com/demisto/etc/issues/18055

## Description
When the url parameter is different than `https://api.intsights.com/`, the commands crash with the error : 

> No JSON object could be decoded

## Screenshots
64049784-d094fa00-cb2a-11e9-92b9-73fd3a6bc4ce

### DIFF
--- a/Integrations/IntSight/IntSight.py
+++ b/Integrations/IntSight/IntSight.py
@@ -53,8 +53,8 @@ SEVERITY_LEVEL = {
 def is_json(myjson):
     demisto.info('walla')
     try:
-        json_object = json.loads(myjson)
-    except ValueError as e:
+        json.loads(myjson)
+    except ValueError:
         return False
     return True
 
--- a/Integrations/IntSight/IntSight.py
+++ b/Integrations/IntSight/IntSight.py
@@ -50,7 +50,16 @@ SEVERITY_LEVEL = {
 }
 
 
-def req(method, path, json_data=None, params=None):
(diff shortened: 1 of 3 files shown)


adi88d:
Intsights instance has issue -Account Expired (issue #16351), the test playbook doesn't run during the build but it passed when i did it locally:

image

Integrations/IntSight/IntSight.py
```diff
@@ -59,6 +59,17 @@ def req(method, path, json_data=None, params=None):
...
+        return []  # type: ignore
+
+    if json_response:
+        try:
+            r = r.json()
+        except ValueError:
+            return_error('Error in API call to IntSights service - please check you URL address')
+        return r
+
     return r
```
yaakovi:
```suggestion
    return `r.content`
```
or
```suggestion
    return `r.text`
```

Integrations/IntSight/IntSight.py
```diff
@@ -59,6 +59,17 @@ def req(method, path, json_data=None, params=None):
...
+
+    if r.status_code == 204:
+        return []  # type: ignore
+
+    if json_response:
+        try:
+            r = r.json()
+        except ValueError:
+            return_error('Error in API call to IntSights service - please check you URL address')
+        return r
```
yaakovi: unreachable code, please remove

PR #4284, opened 2019-09-02
fixed clickable links

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17812

## Description
Hiding links


ronykoz:
@yaakovi i don't see the need in UT for simple replace.
And thats the fix:
image

yaakovi:
@ronykoz please fix the change log before merging to master (after the the last commit the format is not correct)

ronykoz:
@yaakovi not sure what do you mean by wrong format here, seems ok to me

yaakovi:
```
## [Unreleased]
Removed the hyperlink from links.

## [19.9.0] - 2019-09-04
  - Improved EML file type detection.
  - Added the *Email.AttachmentNames* output, which contains a list of the names of all email attachments.
```

PR #4287, opened 2019-09-02
moved the location of ? in the url regex to catch url query, unskip "E…

bakatzir:
…xtract Indicators From File - Generic v2 - Test"

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18795

4.1.0

   - No

## Must have
- [x] Tests
- [x] Code Review


bakatzir:
@kirbles19 RN change(should be the same string in both):
priot to 5.0:
https://github.com/demisto/content/blob/e65bc8fd4f3600032e3b5fd04ab6ed90cbbe6511/Misc/reputations_CHANGELOG.md
5.0:
https://github.com/demisto/content/blob/e65bc8fd4f3600032e3b5fd04ab6ed90cbbe6511/Misc/reputation-url_CHANGELOG.md

Tests/scripts/constants.py
```diff
@@ -30,6 +30,7 @@
 TEST_SCRIPT_REGEX = r"{}.*script-.*\.yml$".format(TEST_PLAYBOOKS_DIR)
 SCRIPT_REGEX = r"{}.*script-.*\.yml$".format(SCRIPTS_DIR)
 INTEGRATION_PY_REGEX = r"{}.*\.py$".format(INTEGRATIONS_DIR)
+INTEGRATION_PY_UNIT_TESTS_REGEX = r"{}.*\_test.py$".format(INTEGRATIONS_DIR)
```
yaakovi: why is that here?
and why do you escape `_`?

PR #4298, opened 2019-09-03
Small cleanup / fix for fields

idovandijk:
## Status
Ready

## Description
Fixed some small errors with new fields

## Does it break backward compatibility?
   - No

## Must have
- [x] Documentation (with link to it)
- [x] Code Review

PR #4301, opened 2019-09-03
Fixed error in master build caused by Phishing Sub-type incident field

DeanArbel:
## Status
Ready

## Description
Fixed error in master build caused by Phishing Sub-type incident field file name not matching demisto standards. Also added release notes CHANGELOG.

### Error in master:
image

   - No


DeanArbel:
@idovandijk fyi

DeanArbel:
This will be implemented in https://github.com/demisto/content/pull/4298

PR #4304, opened 2019-09-03
fixed 2 fields to have content=true

idovandijk:
## Status
Ready

## Description
The fields should have `content: true`

   - No

## Must have
- [x] Documentation
- [ ] Code Review


idovandijk:
No need for this

PR #4306, opened 2019-09-04
ATD - mypy score

bakatzir:
## Status
Ready

## Description
before:
Screen Shot 2019-09-04 at 12 39 07
after:
Screen Shot 2019-09-04 at 13 06 47

   - No

## Must have
- [x] Tests
- [x] Code Review

### DIFF
--- a/Integrations/McAfee_Advanced_Threat_Defense/CHANGELOG.md
+++ b/Integrations/McAfee_Advanced_Threat_Defense/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## [Unreleased]
+-
+
+## [19.9.0]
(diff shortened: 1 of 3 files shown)


Integrations/McAfee_Advanced_Threat_Defense/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## [Unreleased]
+-
+
+## [19.9.0]
```
yardensade: We shouldn't be adding version tags to changelogs, it's messing up the clear release notes script, we handle these cases automatically

yardensade: +1 from bark

PR #4307, opened 2019-09-04
[Fixed] ZeroFox - Wrong credentials bug

guyfreund:
## Status
Ready

## Description
Fixed the bug in configuring instance with wrong creds

   - No

## Additional Information
All user outputs added were confirmed by @kirbles19. There's no new outputs shown to the user.

### DIFF
--- a/Integrations/ZeroFox/ZeroFox.py
+++ b/Integrations/ZeroFox/ZeroFox.py
@@ -309,10 +309,10 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
         err_msg = 'Failure verying SSL certificate. Try selecting \'Trust any certificate\' in' \
                   ' the integration configuration.'
         raise Exception(err_msg)
-    except requests.exceptions.ProxyError:
-        err_msg = 'Proxy Error - try clearing \'Use system proxy\' in the integration configuration if it has been' \
-                  ' selected.'
-        raise Exception(err_msg)
+    # except requests.exceptions.ProxyError:
+    #     err_msg = 'Proxy Error - try clearing \'Use system proxy\' in the integration configuration if it has been' \
+    #               ' selected.'
+    #     raise Exception(err_msg)
     except requests.exceptions.ConnectionError as e:
         # Get originating Exception in Exception chain
         while '__context__' in dir(e) and e.__context__:
--- a/Integrations/ZeroFox/ZeroFox.py
+++ b/Integrations/ZeroFox/ZeroFox.py
@@ -290,9 +290,9 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
         else:
             try:
                 if res.status_code not in {200, 201}:
-                    # case of wrong credentials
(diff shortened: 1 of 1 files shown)


guyfreund:
Please note that the msg of the commit [59b3111](https://github.com/demisto/content/pull/4307/commits/59b31115665744398372368dcf8788d187c1a1ac) is wrong and needs to be changed to: "fixed creds bug"

Integrations/ZeroFox/ZeroFox.py
```diff
@@ -288,7 +290,14 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
         else:
             try:
                 if res.status_code not in {200, 201}:
-                    raise Exception('Failure resolving URL.')
+                    # case of wrong credentials
```
Itay4: is it's case of wrong creds, why are we eventually raising `'Failure resolving URL.'`

Integrations/ZeroFox/ZeroFox.py
```diff
@@ -288,7 +290,14 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
         else:
             try:
                 if res.status_code not in {200, 201}:
-                    raise Exception('Failure resolving URL.')
+                    res_data = json.loads(res.text)
```
Itay4: this is risky, if res.text is not a valid json, it will raise an exception

Integrations/ZeroFox/ZeroFox.py
```diff
@@ -300,17 +309,20 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
         err_msg = 'Failure verying SSL certificate. Try selecting \'Trust any certificate\' in' \
                   ' the integration configuration.'
         raise Exception(err_msg)
-    except requests.exceptions.ProxyError:
-        err_msg = 'Proxy Error - try clearing \'Use system proxy\' in the integration configuration if it has been' \
-                  ' selected.'
-        raise Exception(err_msg)
+    # except requests.exceptions.ProxyError:
```
Itay4: uncomment this

Integrations/ZeroFox/ZeroFox.py
```diff
@@ -300,17 +309,20 @@ def http_request(method: str, url_suffix: str, params: Dict = None, data: Union[
...
+    #               ' selected.'
+    #     raise Exception(err_msg)
     except requests.exceptions.ConnectionError as e:
         # Get originating Exception in Exception chain
         while '__context__' in dir(e) and e.__context__:
             e = cast(Any, e.__context__)
-        err_msg = f'\nMESSAGE: {e.strerror}\n' \
-                  f'ADVICE: Check that the Server URL parameter is correct and that you' \
-                  f' have access to the Server from your host.'
+        if not isinstance(e, TypeError):
```
Itay4: is this needed?

guyfreund: Nope, it has been removed

PR #4311, opened 2019-09-04
AD Query v2

dantavori:
## Status
Ready

## Related Issues
fixes: [https://github.com/demisto/etc/issues/18316](https://github.com/demisto/etc/issues/18316)

## Description
1. Fixed a bug in create-contact command
2. Added some detailed description for arguments in search and create commands.

   - No (but users must be aware of the correct search filter syntax)

### DIFF
--- a/Integrations/Active_Directory_Query/Active_Directory_Query.py
+++ b/Integrations/Active_Directory_Query/Active_Directory_Query.py
@@ -570,8 +570,8 @@ def create_contact():
             )
 
     # set common user attributes
-    if args.get('diaply-name'):
-        attributes['displayName'] = args['diaply-name']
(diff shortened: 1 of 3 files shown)


kirbles19:
@dantavori Please only assign me, not as a reviewer.

dantavori:
@ronykoz we can define another syntax that we decode and encode (with regex), but basically the search filter syntax defines a regular language with an infinite cardinality while our command receives in each execution a constant number of arguments (and lists won't help us much)

ronykoz:
@dantavori yeah i know, but it feels strange to ask the user to fill in \\28 instead of brackets

richardbluestone:
I've reviewed both the yml and md files.

PR #4328, opened 2019-09-05
[New Integration] AttackIQ firedrill

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17305

## Description
Implemented AttackIQ firedrill integration.

## Must have
- [x] Tests
- [ ] Code Review


Integrations/AttackIQFireDrill/AttackIQFireDrill.yml
```diff
@@ -0,0 +1,509 @@
...
+    - contextPath: AttackIQ.Assessment.AssessmentTemplate.Id
+      description: Assessment template ID.
+      type: String
+    - contextPath: AttackIQ.Assessment.AssessmentTemplate.TemplateName
+      description: Assessment template Name.
+      type: String
+    - contextPath: AttackIQ.Assessment.AssessmentTemplate.TemplateDescription
+      description: Assessment template description.
+      type: String
+    - contextPath: AttackIQ.Assessment.AssessmentTemplate.AssessmentTemplateType.Id
```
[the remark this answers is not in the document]
DeanArbel: This proved to be problematic. let's go over this together with @Arsenikr

Integrations/AttackIQFireDrill/AttackIQFireDrill.yml
```diff
@@ -0,0 +1,509 @@
...
+      type: Number
+    - contextPath: AttackIQ.Assessment.MasterJobCount
+      description: Number of tests that ran in the assessment.
+      type: Number
+  - arguments:
+    - default: false
+      defaultValue: '1'
+      description: Assessment page. Each page contains by default up to 10 assessments.
+      isArray: false
+      name: page_number
```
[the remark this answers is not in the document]
DeanArbel: Added `Assessment.Count` to context, we can also add `Assessment.Remaining`.

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -0,0 +1,475 @@
...
+    page = demisto.args().get('page_number')
+    try:
+        page_size = int(demisto.args().get('page_size', DEFAULT_PAGE_SIZE))
+        page = int(page)
+    except (ValueError, TypeError):
+        page_size = DEFAULT_PAGE_SIZE
+    raw_assessments = get_assessments(page=page)
+    assessments_res = build_transformed_dict(raw_assessments.get('results'), ASSESSMENTS_TRANS)
+    ass_cnt = raw_assessments.get('count')
+    total_pages = (ass_cnt // page_size) + 1
```
yuvalbenshalom: I think that you should add 1 to `(ass_cnt // page_size)`  if `(ass_cnt % page_size) > 0`,
and not all the time.

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -0,0 +1,475 @@
...
+    return http_request('GET', '/v1/assessments', params={'page': page})
+
+
+def list_assessments_command():
+    """ Implements attackiq-list-assessments command
+    """
+    page = demisto.args().get('page_number')
+    try:
+        page_size = int(demisto.args().get('page_size', DEFAULT_PAGE_SIZE))
+        page = int(page)
```
yuvalbenshalom: what if you get an exception on `page = int(page)`?
there code afterwards dose not handle it.

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -0,0 +1,475 @@
...
+    except (ValueError, TypeError):
+        page_size = DEFAULT_PAGE_SIZE
+    raw_assessments = get_assessments(page=page)
+    assessments_res = build_transformed_dict(raw_assessments.get('results'), ASSESSMENTS_TRANS)
+    ass_cnt = raw_assessments.get('count')
+    total_pages = (ass_cnt // page_size) + 1
+    try:
+        remaining_pages = total_pages - page
+    except (ValueError, TypeError):
+        remaining_pages = None
```
yuvalbenshalom: if the user did not insert a value to `page_number ` the result should not be that the remaining pages is none.
If the user ran the command for the first time it Is natural that the page number will not be specified, but we still expect to return the remaining pages. 

I think it is better to put zero(or 1) in the default, and return an error if there is a non neural number there.
the same for `page_size`

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: replaced with [math.ceil()](https://docs.python.org/2/library/math.html) function

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: Changed behaviour so it'll raise an error to the user in that case

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: replaced implementation, so now it'll always be `integers`, so the math operation will never fail

DeanArbel:
> see my comments about the paging mechanism.

Resolved

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -316,6 +317,30 @@ def build_test_results_hr(test_results, test_id):
...
+
+    """
+    page = args.get('page_number', 1)
+    try:
+        page = int(page)
+    except (ValueError, TypeError):
+        return_error(f'Error: Invalid page_number value. {page} Is not valid. Please enter a positive integer.')
+    try:
+        page_size = int(args.get('page_size', DEFAULT_PAGE_SIZE))
+    except (ValueError, TypeError):
```
yuvalbenshalom: return a similar error here if it gets a page size which is not a number or not natural.

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -316,6 +317,30 @@ def build_test_results_hr(test_results, test_id):
...
+    Get arguments page_number and page_size from args
+    Args:
+        args (dict): Argument dictionary, with possible page_number and page_size keys
+
+    Returns (int, int): Return a tuple of (page_number, page_size)
+
+    """
+    page = args.get('page_number', 1)
+    try:
+        page = int(page)
```
yuvalbenshalom: make sure that page is also natural.(bigger than 1)

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -414,12 +435,20 @@ def build_tests_hr(assessment_res):
 def list_tests_by_assessment_command():
     """ Implements attackiq-list-tests-by-assessment command
     """
+    page, page_size = get_page_number_and_page_size(demisto.args())
```
yuvalbenshalom: Seems the page_size is never used

DeanArbel: you're 100% right, fixed it

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: done and added a check for page_size as well

DeanArbel:
@kirbles19 Hey please review:

Integrations/AttackIQFireDrill/AttackIQFireDrill.yml
Integrations/AttackIQFireDrill/CHANGELOG.md

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -345,16 +351,28 @@ def get_test_results_command(args=demisto.args()):
...
+        'page': page,
+        'page_size': page_size,
         'test_id': test_id,
         'show_last_result': args.get('show_last_result') == 'True'
     }
     try:
         raw_test_res = http_request('GET', '/v1/results', params=params)
         test_res = build_transformed_dict(raw_test_res['results'], TEST_RESULT_TRANS)
+        test_cnt = raw_test_res.get('count')
+        total_pages = math.ceil(test_cnt / page_size)
```
yuvalbenshalom: image

I think we have a problem with this devision

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -410,38 +430,51 @@ def get_assessment_by_id_command():
...
-    raw_res = http_request('GET', f'/v1/tests', params={'project': ass_id})
+    params = {
+        'project': ass_id,
+        'page_size': page_size,
+        'page': page
+    }
+    raw_res = list_tests_by_assessment(params)
     test_cnt = raw_res.get('count')
     tests_res = build_transformed_dict(raw_res.get('results'), TESTS_TRANS)
     total_pages = math.ceil(test_cnt / page_size)
```
yuvalbenshalom: problematic div also

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
```diff
@@ -386,6 +404,8 @@ def list_assessments_command():
     ass_cnt = raw_assessments.get('count')
     total_pages = math.ceil(ass_cnt / page_size)
```
yuvalbenshalom: problematic div too

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: irrelevant

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: irrelevant

Integrations/AttackIQFireDrill/AttackIQFireDrill.py
DeanArbel: irrelevant

richardbluestone:
I've reviewed and updated the yaml file.

PR #4331, opened 2019-09-06
CommonServerPython - parse_date_string function

Itay4:
## Status
Ready

## Description
Added a function `parse_date_string` which parses a date string to a datetime object.
The function extracts the requested date format based on the error message raised when trying to parse with wrong format.
It can be useful mainly in `fetch_incidents` functions which gets a date string and parses it to a datetime object.
This should handle bugs we get for errors raised in fetch_incidents such as: 
Screenshot 2019-09-06 at 16 50 45
There are builtin functions for that in packages like `datetime` but only for Python3 and `dateutil` but it is not in our default docker image, so the added function can be used until we can start using these builtin ones.

Note that this function should be able to handle common date formats, but does not handle uncommon ones like ones that include ISO weeks.

Any

   - No

## Must have
- [x] Tests
- [x] Documentation
- [ ] Code Review

## Dependencies
None


Itay4:
@kirbles19 please review the changelog and the function [docstring](https://github.com/demisto/content/blob/fbc8e78c5f0e3b86225d9cc4befa8f4ba080d9ce/Scripts/CommonServerPython/CommonServerPython.py#L1814-L1832), which will be displayed in script helper

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1808,3 +1808,70 @@ def get_demisto_version():
...
+
+            if unconverted_data_remains_match:
+                # found unconverted_data_remains
+                # example of caught error message:
+                # "unconverted data remains: 22Z"
+                sliced_time_data = unconverted_data_remains_match[0]
+
+        if not sliced_time_data:
+            # did not catch expected error
+            raise ValueError(e)
```
yaakovi: inorder to preserve original exception data
```suggestion
            raise
```

kirbles19:
@Itay4 All done.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1808,3 +1808,70 @@ def get_demisto_version():
         return demisto.demistoVersion()
     else:
         raise AttributeError('demistoVersion attribute not found.')
+
+
+def parse_date_string(date_string, date_format='%Y-%m-%dT%H:%M:%S'):
+    """
+        Parses the date_string function to the corresponding datetime object.
+
```
yuvalbenshalom: lets add comment here that suggest to use better functions in datetime-Python3 or dateutil when possible.

Itay4:
@kirbles19 i've added another note to the docstring, please review again

Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -617,3 +617,25 @@ def test_get_demisto_version(mocker):
...
+    assert parse_date_string('2019-09-17T06:16:39Z') == datetime(2019, 9, 17, 6, 16, 39)
+
+    # test unconverted data remains: .22Z
+    assert parse_date_string('2019-09-17T06:16:39.22Z') == datetime(2019, 9, 17, 6, 16, 39, 220000)
+
+    # test time data without ms does not match format with ms
+    assert parse_date_string('2019-09-17T06:16:39Z', '%Y-%m-%dT%H:%M:%S.%f') == datetime(2019, 9, 17, 6, 16, 39)
+
+    # test time data with timezone Z does not match format with timezone +05:00
+    assert parse_date_string('2019-09-17T06:16:39Z', '%Y-%m-%dT%H:%M:%S+05:00') == datetime(2019, 9, 17, 6, 16, 39)
```
ronykoz: wouldn't we want to convert it to UTC maybe?

Itay4: not sure i understand what you mean

kirbles19:
@Itay4 Updated the new strings.

Scripts/CommonServerPython/CommonServerPython_test.py
ronykoz: this time is represented in local time rather than in UTC, if we want to calculate the UTC we will decrease 5 hours of the time we got

Itay4:
> @Itay4 add more use cases in unit tests
> for date formats like:
> 2019-09-17
> 2019-09-17 06:13:39
> 2019-09-17 06:13
> 2019/09/17 06:13:39
> 17/09/2019

@anara123 the function does not cover these formats
can easily add them, but as discussed, only common date strings formats can be handled by this function
if you think these are such, ill add them, what do you say ?

PR #4336, opened 2019-09-08
[Fixed] Bug in pkg_dev_test_tasks.py

guyfreund:
## Status
Ready

## Description
When I was trying to run the script "pkg_dev_test_tasks.py" in the integration package I was getting the next error message: 

> [FAILED /Users/gfreund/dev/demisto/content/Integrations/AlienVault_USM_Anywhere] Error: Command '['pipenv', 'lock', '-r', '-d']' returned non-zero exit status 1.

@ronykoz helped me to backtrack the flow of the "pipenv lock -r -d" command and we've noticed that the pipfile is taken from Tests/scripts/dev_envs/default_python3/2 and not from the integration package. We've changed the code accordingly in the related function ״get_dev_requirements״.

## Screenshots
Screen Shot 2019-09-08 at 15 11 09

   - No

## Must have
- [x] Tests
- [x] Code Review

## Additional changes
Changed the name of all project_dir variables to package_dir, package is better associated in the context of integration than project.

## Additional Information
No user outputs changed, therefore "docs-approved" label was added by me.

### DIFF
--- a/Tests/scripts/pkg_dev_test_tasks.py
+++ b/Tests/scripts/pkg_dev_test_tasks.py
@@ -78,7 +78,7 @@ def get_pipenv_dir(py_version):
     return "{}{}".format(ENVS_DIRS_BASE, int(py_version))
 
 
-def get_dev_requirements(py_version):
+def get_dev_requirements(py_version, package_dir):
     """
     Get the requirements for the specified py version.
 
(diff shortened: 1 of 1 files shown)


ronykoz:
@DeanArbel you can follow up here

glicht:
It is by design taken from there to make sure we run with the exact dependencies in docker.

glicht:
The problem you had was that you were not running the script from the content root dir. If you want to make a change so the script can also be run from within a package dir, you are welcome. But don't change the pipenv being used to build the docker for testing.

PR #4339, opened 2019-09-08
Urlscan BugFixes

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#17768
fixes: demisto/etc#19044

## Description
- Enabled global override of the public/private argument via a parameter.
- URLScan now uses the `verdicts` entry in the response.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Integrations/UrlScan/URLScan.py
```diff
@@ -146,7 +146,7 @@ def poll(target, step, args=(), kwargs=None, timeout=None, max_tries=None, check
 
 def urlscan_submit_url():
     submission_dict = {}
-    if demisto.args().get('public') == 'public':
+    if demisto.params().get('privacy') == 'public':
```
yuvalbenshalom: Seems like you disabled the argument of the command.
We should talk about it again because this breaks BC anyway.

amshamah419:
> we should think again on the public/private defaults because it will break BC anyway.
> and seems like WHOIS files are in this PR by mistake.

I think I know what happened here... In my mind, I already merged the Whois stuff. I did not yet. Checking with Alex to see if we can get a green light on docs. Then the diff should be the same.

Integrations/UrlScan/URLScan.py
amshamah419: I agree. I'm not sure how we would both change the default and not break BC. I could change the argument do do something like:
```
if demisto.args().get('public') == 'public' and demisto.params().get('override) is not True:
   do the thing
```

Integrations/UrlScan/URLScan.yml
```diff
@@ -28,9 +28,15 @@ configuration:
   defaultvalue: "1"
   type: 0
   required: false
+- display: Enable public submissions by default.
+  name: is_public
+  defaultvalue: false
```
yuvalbenshalom: change to true

Integrations/UrlScan/URLScan_CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## [Unreleased]
+  - URLScan now leverages the _Verdict_ result from the API.
```
yuvalbenshalom: add comment about the new parameter

amshamah419:
@kirbles19 Can you review these? 
https://github.com/demisto/content/blob/urlscan-bf/Integrations/UrlScan/URLScan_CHANGELOG.md

https://github.com/demisto/content/blob/urlscan-bf/Integrations/UrlScan/URLScan.yml

kirbles19:
@amshamah419 looks like this integration needs updates for context standards.

amshamah419:
@kirbles19 Looks like they were there in the code, just not in the yml. I updated the yml to have the context outputs.

kirbles19:
@amshamah419 All done.

PR #4340, opened 2019-09-08
run validate integration on new integration files

yuvalbenshalom:
## Status
Ready

## Description
Just found out that validate integration do not run on new integrations.
Pretty sure it should.


ronykoz:
@yaakovi @anara123 please force merge this(Due to the build problems)

PR #4343, opened 2019-09-09
fixed integration/script code files regex

yaakovi:
## Status
Ready

## Related Issues
related: https://circleci.com/gh/demisto/content/29036
@DeanArbel FYI

## Description
fixed regex for code files in integration/script packages.
this regex will only catch code files in the package directories (excluding the unit test files).

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Additional changes
as a followup to this fix, i opened dev task to validate the rest of the regex in that file:
https://github.com/demisto/etc/issues/19073

### DIFF
--- a/Tests/scripts/constants.py
+++ b/Tests/scripts/constants.py
@@ -1,4 +1,5 @@
 import re
+import os
 
 # dirs
 INTEGRATIONS_DIR = "Integrations"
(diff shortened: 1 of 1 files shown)


Itay4:
why do we have the js regex ?

yaakovi:
we support packages for JS as well, no one is using this feature but it is supported.

PR #4346, opened 2019-09-09
Added batching to Cylance

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/15766

## Description
Added a batching logic to the delete-devices command.
Requests will now be batched.

   - No

## Must have
- [ ] Code Review


Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
```diff
@@ -40,7 +40,11 @@ configuration:
 - defaultvalue: '-59'
   display: File Threshold
   name: file_threshold
+  type: 0
   required: false
+- display: Deletion Batch Size
```
ronykoz: If it's only for one command why is it a param and not an argument of that function?

guykeller: Because it was supposed to be used in other commands as well, but it seemed to me after reviewing that this is the only command that uses the mechanism.
Obviously mentioned it in the issue - but no adequate response (they are very unresponsive and uninterested in this fix)

ronykoz: i think that this should be done in an argument, seems un-natural this way

Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
```diff
@@ -1155,8 +1155,15 @@ def delete_devices():
             'Device Name': device_name,
             'Deletion status': True
         })
+    batch_size = demisto.args().get("batch_size", 20)
+    try:
```
ronykoz: why not batch_size.isdigit() ?

Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
```diff
@@ -962,10 +961,13 @@ script:
     - contextPath: Cylance.Device.Deleted
       description: A boolean to check if the device was deleted.
       type: boolean
+    description: Delete one or more devices from an organization.
+    execution: true
```
ronykoz: whats that?

Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
guykeller: because that would fail on an int.
I can assign the default get value to be "20" and then it should work, but this seems more robust, don't you think?

Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
guykeller: It yelled at me because the yml didn't have those, and when I downloaded it from my Demisto instance they did exist so copied them from there

ronykoz:
@kirbles19 new arg was added for batch delete, so just that line on the yml and the CHANGELOG

kirbles19:
@guykeller All done.

PR #4349, opened 2019-09-09
Create integration-MCAS.yml

Infosec-L3:
Integration for Microsoft Cloud App Security (MCAS). Allows listing of alerts by severity, querying alerts/activity by IP, and querying alerts by username.

## Status
Ready

## Related Issues
fixes: link to the issue

## Description
Supported arguments:
mcas-list-alerts (Arguments: limit; severity): Presents recent alerts by defined limit (i.e., last 20 alerts) and severity (Low, Medium, High, All). Alerts are saved to context with pertinent information (i.e., user, IP, description, etc.)

mcas-lookup-ip (Arguments: IP): Looks up associated users with recent activity/alerts by IP address.

mcas-lookup-user (Arguments: username): Looks up recent alerts by username/email address.

## Screenshots

## Related PRs
N/A

branch | PR
------ | ------

4.5.0

   - No

## Must have
- [ ] Tests
- [ ] Code Review -- Scanned code in Checkmarx with 0 findings.

## Dependencies
N/A

## Additional changes
Addressed suggested changes. Added http_request function for specific requests; removed deprecated quotation marks; updated defaults and comments in integration settings; updated test function for integration.

### DIFF
--- a/Integrations/integration-MCAS.yml
+++ b/Integrations/integration-MCAS.yml
@@ -17,7 +17,7 @@ configuration:
   defaultvalue: ""
   type: 4
   required: true
-- display: Ignore server certificate
+- display: Trust any certificate (not secure)
   name: insecure
   defaultvalue: ""
   type: 8
--- a/Integrations/integration-MCAS.yml
+++ b/Integrations/integration-MCAS.yml
@@ -22,7 +22,7 @@ configuration:
   defaultvalue: ""
   type: 8
   required: false
-- display: Use System Proxy
+- display: Use system proxy settings
   name: proxy
   defaultvalue: ""
   type: 8
--- a/Integrations/integration-MCAS.yml
+++ b/Integrations/integration-MCAS.yml
@@ -1,0 +1,265 @@
+commonfields:
+  id: Microsoft Cloud App Security
+  version: -1
+name: Microsoft Cloud App Security
+display: Microsoft Cloud App Security
+category: Network Security
+image: data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAbcAAAG3BAMAAADFuQbYAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAAAElBMVEXz8/PzUyWBvAYFpvD/ugj///98Xc9SAAAAAWJLR0QF+G/pxwAAAAd0SU1FB+IECxAWCcy7n+sAAAGSSURBVHja7c9BCQAgEABBxQJawQSCFeyfyZcRTu4xW2CZUiRJkiRJylodwfV3mtHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcF9wUmSJEmSlK62g1vvdKKDg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4P7gpMkSZIkKUsX5n/TJVBjPXQAAAAldEVYdGRhdGU6Y3JlYXRlADIwMTgtMDQtMTFUMTY6MjI6MDkrMDA6MDD3dzx7AAAAJXRFWHRkYXRlOm1vZGlmeQAyMDE4LTA0LTExVDE2OjIyOjA5KzAwOjAwhiqExwAAAABJRU5ErkJggg==
+description: Microsoft Cloud App Security (MCAS)
+configuration:
+- display: Server URL
+  name: url
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: API Token
+  name: token
+  defaultvalue: ""
+  type: 4
+  required: true
+- display: Ignore server certificate
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Use System Proxy
+  name: proxy
+  defaultvalue: ""
+  type: 8
+  required: false
+script:
+  script: |-
+    import requests
+    import json
+    import yaml
+
+    SERVER = demisto.params()['url'][:-1] if demisto.params()['url'].endswith('/') else demisto.params()['url']
+    TOKEN = demisto.params().get('token')
+    USE_SSL = not demisto.params().get('insecure', False)
+
+    if not demisto.params()['proxy']:
+        del os.environ['HTTP_PROXY']
+        del os.environ['HTTPS_PROXY']
+        del os.environ['http_proxy']
+        del os.environ['https_proxy']
+
+    def mcas_list_alerts():
+        limit = demisto.args()['limit']
+        severity_raw = demisto.args()['severity']
+        if severity_raw == "Low":
+            severity = "0"
+        elif severity_raw == "Medium":
+            severity = "1"
+        elif severity_raw == "High":
+            severity = "2"
+        else:
+            severity = "2,1,0"
+        url = SERVER + '/cas/api/v1/alerts/'
+        data = '{"skip":0,"limit":' + limit + ',"filters":{"resolutionStatus":{"eq":[0]},"severity":{"eq":[' + severity + ']}},"sortField":"date","sortDirection":"desc","performAsyncTotal":true}'
+        headers = {'Authorization': 'Token ' + TOKEN, 'Content-type': 'application/json'}
+        r = requests.post(url=url, headers=headers, data=data, verify=USE_SSL)
+        j = yaml.load(json.dumps(r.json()))
+        x = (len(j['data']))
+        results = []
+        demisto.results(j)
+        print('Number of fields: ' + str(x))
+        for i in range(0, x):
+            ip = "unknown"
+            user = "unknown"
+            email = "unknown"
+            label = "unknown"
+            country = "unknown"
+            policyType = "unknown"
+            policyName = "unknown"
+            url = j['data'][i]['URL']
+            root_id = j['data'][i]['_id']
+            context_id = j['data'][i]['contextId']
+            description_raw = j['data'][i]['description']
+            description = description_raw.replace('<p>'," ").replace('</p>'," ").replace('<br>'," ")
+            title = j['data'][i]['title']
+            timestamp = j['data'][i]['timestamp']
+            entities = j['data'][i]['entities']
+            y = (len(entities))
+            for k in range(0, y):
+                if 'em' in entities[k]:
+                    email = entities[k]['em']
+                    user = entities[k]['label']
+                if 'type' in entities[k]:
+                    if entities[k]['type'] == "user":
+                        email = entities[k]['label']
+                    if entities[k]['type'] == "account":
+                        user = entities[k]['label']
+                    if entities[k]['type'] == "ip":
+                        ip = entities[k]['label']
+                    if entities[k]['type'] == "country":
+                        country = entities[k]['label']
+                    if entities[k]['type'] == "policyRule":
+                        policyType = entities[k]['policyType']
+                        policyName = entities[k]['label']
+            results.append({'ID': root_id, 'title': title, 'description': description, 'policyName': policyName,
+            'policyType': policyType, 'timestamp': timestamp, 'user': user, 'email': email, 'ip': ip, 'country': country})
+        entry = {'Type' : entryTypes['note'],
+                      'Contents' :results,
+                      'ContentsFormat' : formats['json'],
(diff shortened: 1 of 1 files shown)


Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+  name: url
+  defaultvalue: ""
+  type: 0
+  required: true
+- display: API Token
+  name: token
+  defaultvalue: ""
+  type: 4
+  required: true
+- display: Ignore server certificate
```
Itay4:
```suggestion
- display: Trust any certificate (not secure)
```

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+  name: insecure
+  defaultvalue: ""
+  type: 8
+  required: false
+- display: Use System Proxy
```
Itay4:
```suggestion
- display: Use system proxy settings
```

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+                      'ContentsFormat' : formats['json'],
+                      'HumanReadable': tableToMarkdown('MCAS list', results),
+                      'ReadableContentsFormat' : formats['markdown'],
+                      'EntryContext' :{'mcas_username': results}
+                     }
+            return entry
+    try:
+        # The command demisto.command() holds the command sent from the user.
+        if demisto.command() == 'test-module':
+            # This is the call made when pressing the integration test button.
```
Itay4: Please remove this comment.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+            demisto.results('ok')
+            sys.exit(0)
```
Itay4: Please remove as there is no need for it.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+                      'HumanReadable': tableToMarkdown('MCAS list', results),
+                      'ReadableContentsFormat' : formats['markdown'],
+                      'EntryContext' :{'mcas_username': results}
+                     }
+            return entry
+    try:
+        # The command demisto.command() holds the command sent from the user.
+        if demisto.command() == 'test-module':
+            # This is the call made when pressing the integration test button.
+            demisto.results('ok')
```
Itay4: Please add some HTTP query here, so the test module will actually test something.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+            # Export results to context
+            entry = {'Type' : entryTypes['note'],
+                      'Contents' :results,
+                      'ContentsFormat' : formats['json'],
+                      'HumanReadable': tableToMarkdown('MCAS list', results),
+                      'ReadableContentsFormat' : formats['markdown'],
+                      'EntryContext' :{'mcas_username': results}
+                     }
+            return entry
+    try:
```
Itay4: Please wrap the execution `try-except` block in a `main()` function

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+  script: |-
+    import requests
+    import json
+    import yaml
+
+    SERVER = demisto.params()['url'][:-1] if demisto.params()['url'].endswith('/') else demisto.params()['url']
+    TOKEN = demisto.params().get('token')
+    USE_SSL = not demisto.params().get('insecure', False)
+
+    if not demisto.params()['proxy']:
```
Itay4: Please use `handle_proxy` function instead.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+        elif severity_raw == "Medium":
+            severity = "1"
+        elif severity_raw == "High":
+            severity = "2"
+        else:
+            severity = "2,1,0"
+        url = SERVER + '/cas/api/v1/alerts/'
+        data = '{"skip":0,"limit":' + limit + ',"filters":{"resolutionStatus":{"eq":[0]},"severity":{"eq":[' + severity + ']}},"sortField":"date","sortDirection":"desc","performAsyncTotal":true}'
+        headers = {'Authorization': 'Token ' + TOKEN, 'Content-type': 'application/json'}
+        r = requests.post(url=url, headers=headers, data=data, verify=USE_SSL)
```
Itay4: It is best practise to use `json` instead of `data` in this case.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+        j = yaml.load(json.dumps(r.json()))
```
Itay4: It is a bit risky to call `.json()` like that, as it might raise an encoding exception.
It is better to wrap it in `try-except`, catching `ValueError`

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+        x = (len(j['data']))
+        results = []
+        demisto.results(j)
+        print('Number of fields: ' + str(x))
```
Itay4: Please remove the print

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+        else:
+            severity = "2,1,0"
+        url = SERVER + '/cas/api/v1/alerts/'
+        data = '{"skip":0,"limit":' + limit + ',"filters":{"resolutionStatus":{"eq":[0]},"severity":{"eq":[' + severity + ']}},"sortField":"date","sortDirection":"desc","performAsyncTotal":true}'
+        headers = {'Authorization': 'Token ' + TOKEN, 'Content-type': 'application/json'}
+        r = requests.post(url=url, headers=headers, data=data, verify=USE_SSL)
+        j = yaml.load(json.dumps(r.json()))
+        x = (len(j['data']))
+        results = []
+        demisto.results(j)
```
Itay4: Please remove this

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+        print('Number of fields: ' + str(x))
+        for i in range(0, x):
+            ip = "unknown"
```
Itay4: Not sure if we need these `unknown` defaults, why not get the value from the API and not do anything with it if it's not there?

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
...
+  version: -1
+name: Microsoft Cloud App Security
+display: Microsoft Cloud App Security
+category: Network Security
+image: data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAbcAAAG3BAMAAADFuQbYAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAAAElBMVEXz8/PzUyWBvAYFpvD/ugj///98Xc9SAAAAAWJLR0QF+G/pxwAAAAd0SU1FB+IECxAWCcy7n+sAAAGSSURBVHja7c9BCQAgEABBxQJawQSCFeyfyZcRTu4xW2CZUiRJkiRJylodwfV3mtHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcF9wUmSJEmSlK62g1vvdKKDg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4P7gpMkSZIkKUsX5n/TJVBjPXQAAAAldEVYdGRhdGU6Y3JlYXRlADIwMTgtMDQtMTFUMTY6MjI6MDkrMDA6MDD3dzx7AAAAJXRFWHRkYXRlOm1vZGlmeQAyMDE4LTA0LTExVDE2OjIyOjA5KzAwOjAwhiqExwAAAABJRU5ErkJggg==
+description: Microsoft Cloud App Security (MCAS)
+configuration:
+- display: Server URL
+  name: url
+  defaultvalue: ""
```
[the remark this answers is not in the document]
Infosec-L3: This is organization specific. There is no default URL.

Integrations/integration-MCAS.yml
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
```diff
@@ -0,0 +1,265 @@
+commonfields:
+  id: Microsoft Cloud App Security
+  version: -1
+name: Microsoft Cloud App Security
+display: Microsoft Cloud App Security
+category: Network Security
+image: data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAbcAAAG3BAMAAADFuQbYAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6JgAAgIQAAPoAAACA6AAAdTAAAOpgAAA6mAAAF3CculE8AAAAElBMVEXz8/PzUyWBvAYFpvD/ugj///98Xc9SAAAAAWJLR0QF+G/pxwAAAAd0SU1FB+IECxAWCcy7n+sAAAGSSURBVHja7c9BCQAgEABBxQJawQSCFeyfyZcRTu4xW2CZUiRJkiRJylodwfV3mtHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcF9wUmSJEmSlK62g1vvdKKDg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4P7gpMkSZIkKUsX5n/TJVBjPXQAAAAldEVYdGRhdGU6Y3JlYXRlADIwMTgtMDQtMTFUMTY6MjI6MDkrMDA6MDD3dzx7AAAAJXRFWHRkYXRlOm1vZGlmeQAyMDE4LTA0LTExVDE2OjIyOjA5KzAwOjAwhiqExwAAAABJRU5ErkJggg==
+description: Microsoft Cloud App Security (MCAS)
+configuration:
+- display: Server URL
```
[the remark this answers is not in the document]
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
Infosec-L3: Updated in offline copy; will upload after subsequent changes fixed.

Integrations/integration-MCAS.yml
Infosec-L3: Each query results may or may not contain the value IP. Since each record in the response is being added to context, if I were to ignore fields that were not present, there would need to be multiple versions of the 'results' array, or "null" values in the context. Setting defaults to 'unknown' ended up resulting in much less code and makes playbook construction easier since there is always a string present in every context label. Should I still restructure?

Infosec-L3:
439px-Microsoft_logo

Infosec-L3:
439px-Microsoft_logo

Integrations/integration-MCAS.yml
Itay4: Is `unknown` also a value returned by the API, or a custom default value you set?

Infosec-L3: It is a custom value. If I don't set the default as a string then the context will contain a `null` value for some records.

Itay4: So it is better not to return it and make sure to remove the `null`
We have builtin functions for that:
For human readable, you can pass `removeNull=True` to `tableToMarkdown`.
Same argument exist for `createContext`

Infosec-L3: Thanks @Itay4. `removeNull=True` is a great mention -- I didn't realize that was an option. However, I still need to assign the default string value as the json fields vary based on the type of alert returned. For instance, 5 different alert types may have IP, description, username, and time, but 3 other ones may have username, time, location, and reputation. If I only assign the set the variable when the field is present, the next iteration of the loop will have the previous value for IP instead of being null. Now, I could fully mitigate this will different case statements for each alert type, but since there are so many alert types this is much more practical. What are your thoughts?

Itay4: Let's schedule an execution review and go over this together and see what would be the right thing to do here.
Will reach out to you in DFIR.

PR #4350, opened 2019-09-09
Tanium bug fix 

dantavori:
Added a missing line in deploy_action() that should fix the issue with 'get_results' argument

## Status
In Progress

## Related Issues
fixes: [https://github.com/demisto/etc/issues/18871](https://github.com/demisto/etc/issues/18871)

## Description
The 'get_results' argument should now work as supposed to, but still - when it is set to True, the deployment will be executed but in Demisto there will be a script timeout for executions longer than 5 minutes.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/Tanium/Tanium.py
+++ b/Integrations/Tanium/Tanium.py
@@ -303,7 +303,7 @@ def deploy_action(handler):
     for key, value in demisto.args().items():
         kwargs[key] = value
 
-    kwargs["get_results"] = True if kwargs.get('get_results') == 'True' else False
+    kwargs["get_results"] = True if kwargs.get('get_results').lower() == 'true' else False
 
     callbacks = {}
     callbacks['PreAddAction'] = handle_cgs
--- a/Integrations/Tanium/Tanium.py
+++ b/Integrations/Tanium/Tanium.py
@@ -303,7 +303,7 @@ def deploy_action(handler):
     for key, value in demisto.args().items():
         kwargs[key] = value
 
-    kwargs["get_results"] = True if kwargs.get('get_results').lower() == 'true' else False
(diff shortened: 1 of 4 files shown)


dantavori:
We have no Tanium license to verify the changes so we won't merge the branch until we make sure it fixed the problem.

Integrations/Tanium/Tanium.py
```diff
@@ -302,6 +302,9 @@ def deploy_action(handler):
 
     for key, value in demisto.args().items():
         kwargs[key] = value
+
+    kwargs["get_results"] = True if kwargs.get('get_results') == 'True' else False
```
ronykoz: Safer this way
```suggestion
    kwargs["get_results"] = True if kwargs.get('get_results').lower() == 'true' else False
```

Integrations/Tanium/Tanium.py
```diff
@@ -302,6 +302,9 @@ def deploy_action(handler):
 
     for key, value in demisto.args().items():
         kwargs[key] = value
+
+    kwargs["get_results"] = True if kwargs.get('get_results').lower() == 'true' else False
```
ronykoz: My bad for the first time, it might raise an exception if you don't enter get_results at all.
```suggestion
    kwargs["get_results"] = True if kwargs.get('get_results', '').lower() == 'true' else False
```

ronykoz:
@kirbles19 just the CHANGELOG.md please

kirbles19:
@dantavori Done.

ronykoz:
@dantavori please fix the lint

dantavori:
need to approve the PR and merge, customer [says](https://github.com/demisto/etc/issues/18871#issuecomment-537053037) the bug is fixed

PR #4353, opened 2019-09-10
Update sane doc reports automation

agamm:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19058
fixes: https://github.com/demisto/etc/issues/18556

## Description
Updates the sane-doc-reports image tag.

## Note
It will fail because we are changing the docker tag.


agamm:
@anara123 / @ronykoz ready.

ronykoz:
@anara123 we need force merge due to BC break(but this is from 5.0, so no biggie)

anara123:
@ronykoz dont he missing CHANGELOG?

ronykoz:
@anara123 he has a Changelog file there, but i don't see the need in updating it for this change as customers still don't use this

ronykoz:
@anara123 ?

PR #4371, opened 2019-09-11
add printout of output when a subprocess fails

glicht:
## Status
Ready

## Description
Add print of output when running unit tests / pylint and there is a failure. Should make things clearer for users.


ronykoz:
Added couple of `\n` to make it more readable

PR #4372, opened 2019-09-12
Contributing docs organizing

Itay4:
## Status
Ready

## Description
we have two contributing docs, removed the one from the docs directory, and referred all docs to the one in the root


glicht:
@Itay4 added section about cla

PR #4375, opened 2019-09-12
mention -v option when failing

glicht:
## Status
Ready

## Description
ex.output is None because of the way we run subprocesses in non verbose mode. Simply added a printout suggesting to run in verbose mode if failing. Should make life easier when encountering failures.

### DIFF
--- a/Tests/scripts/pkg_dev_test_tasks.py
+++ b/Tests/scripts/pkg_dev_test_tasks.py
@@ -314,7 +314,7 @@ Will lookup up what docker image to use and will setup the dev dependencies and 
         except subprocess.CalledProcessError as ex:
             sys.stderr.write("[FAILED {}] Error: {}\n".format(project_dir, str(ex)))
             if not LOG_VERBOSE:
-                sys.stderr.write("Need a more detailed log?" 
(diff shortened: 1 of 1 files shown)


Tests/scripts/pkg_dev_test_tasks.py
```diff
@@ -312,7 +312,10 @@ def main():
                 docker_image_created = docker_image_create(docker, requirements)
                 docker_run(project_dir, docker_image_created, args.no_test, args.no_pylint, args.keep_container, args.root)
         except subprocess.CalledProcessError as ex:
-            sys.stderr.write("[FAILED {}] Error: {}.\nOutput:\n{}\n".format(project_dir, str(ex), ex.output))
+            sys.stderr.write("[FAILED {}] Error: {}\n".format(project_dir, str(ex)))
+            if not LOG_VERBOSE:
```
ronykoz: but this way if you face an exception you will not display its full contents

glicht: It doesn't display anyway unless you run with -v because the way subprocesses are run. This fix seems to be sufficient.

PR #4379, opened 2019-09-12
IdentifyAttachedEmail: context output improvement

glicht:
## Status
Ready

## Description
Following cases that context was not set moved to away from deprecated setContext method.

@idovandijk please use this script for your tests


kirbles19:
@glicht Done.

PR #4381, opened 2019-09-12
EWSv2

dantavori:
## Status
Ready

## Related Issues
fixes: [https://github.com/demisto/etc/issues/18380](https://github.com/demisto/etc/issues/18380)

## Description
removed unnecessary try except in move_item_between_mailboxes function 

   - No


Integrations/EWSv2/EWSv2.yml
```diff
@@ -60,8 +60,8 @@ configuration:
   name: defaultServerVersion
   required: false
   type: 0
-- defaultvalue: 'true'
-  display: Trust any certificate (unsecure)
+- defaultvalue: ''
```
IkaDemisto: Changing the default value to false can cause ssl verification error at users that already configured `EWS v2`, so it should be taken into consideration.

dantavori: After an instance's parameter value was set (with default value or not), changing the parameter's default value in the .yml file won't affect the current value.

kirbles19:
@dantavori Done.

Itay4:
@dantavori ?

jpeng19:
@jpeng19 
https://demisto.zendesk.com/agent/tickets/6983

PR #4387, opened 2019-09-13
WildFire - bug fix

dantavori:
## Status
Ready

## Related Issues
fixes: [https://github.com/demisto/etc/issues/19077](https://github.com/demisto/etc/issues/19077)

## Description
Some executions of "get-report" returned the error "string indices must be integers" - it happened when there was only one report in the result, and it was returned as a single json and not an array, which caused an error.

   - No


ronykoz:
@kirbles19 only the changelog please, @dantavori left it in the same format as the one we already had there

kirbles19:
@dantavori Done.

PR #4394, opened 2019-09-15
GmailTest playbook - user existence check

roysagi:
## Status
Ready

## Description
Updating GmailTest playbook - now checks for existing users before creation.
If they exist - it deletes them - if not, continues as normal.


roysagi:
I didn't want to touch the original playbook flow when adding new commands - so I created a different user to reset his password

PR #4405, opened 2019-09-15
Update circleci config for a limited build on forks

avidan-H:
## Status
In Progress

## Related Issues
https://github.com/demisto/etc/issues/18464
https://github.com/demisto/etc/issues/17791

## Description
Run a limited build for forked PRs.

any

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Make sure the build process for unforked PRs is unchanged


avidan-H:
Will need to update the circleci settings for the `content` repo to ensure that `Build forked pull requests` option is _on_ and that `Pass secrets to builds from forked pull requests` is _off_
image

.circleci/config.yml
```diff
@@ -43,45 +59,70 @@ jobs:
...
             pytest ./Tests/scripts/test_configure_tests.py -v
+      - run:
+          name: Spell Checks
+          when: always
+          command: |
+            python ./Tests/scripts/circleci_spell_checker.py $CIRCLE_BRANCH
+      - run:
+          name: Common Server Documentation
+          when: always
+          command: ./Documentation/commonServerDocs.sh
```
ronykoz: Why do we want this in validation_and_testing?

ronykoz:
@avidan-H did you check if we expose secret keys this way?

glicht:
Can we keep the name of the job as `build` and not `validation_and_testing`?

image

.circleci/config.yml
glicht: It is good to have it. To make sure the common server docs are generated properly. I've had cases that this failed as a result of changing common server.

avidan-H:
> Can we keep the name of the job as `build` and not `validation_and_testing`?
> 
> image

Well it's actually two separate jobs so in practice it actually looks like so. 
image
It's no problem to change the name of the job. I think `build` will work (The only reason I'm not certain is that it looks like it is a reserved job name when one isn't using workflows)
image

but given that there are actually two jobs do you still want the first one to be called `build`?

avidan-H:
> @avidan-H did you check if we expose secret keys this way?

@ronykoz I'm fairly certain we are fine in that regard if you look at the documentation [here](https://circleci.com/docs/2.0/oss/#security) and [here](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests)

ronykoz:
@avidan-H i think that GitHub requires a `build` to pass as well as you can see in the following image.
image

Itay4:
@glicht @anara123 ok from you to merge ?

glicht:
Not sure artifacts are being stored during the `build` stage. See: https://circleci.com/gh/demisto/content/29894#artifacts/containers/0

Itay4:
@glicht @anara123 please review

PR #4426, opened 2019-09-18
[Fixed] Snowflake - Fetch Incidents Bug

guyfreund:
## Status
Ready

## Related Issues
we think it fixes: https://github.com/demisto/etc/issues/18992

## Description
We think the bug's reason is wrong use of strftime func - i.e. using this function with wrong time format. In order to solve it we use isoformat func.

## Other changes within the code
I've added support for correct error msg when entering an integer value of demisto arg.

   - No (We've just changed the way the occurred value of incident looks)

## Must have
- [ ] Tests
- [ ] Code Review

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/db5a700bebd8cba8057bac755e9102f2513bbfd4/Integrations/Snowflake/CHANGELOG.md)

### DIFF
--- a/Integrations/Snowflake/CHANGELOG.md
+++ b/Integrations/Snowflake/CHANGELOG.md
@@ -1,0 +1,2 @@
+## [Unreleased]
+-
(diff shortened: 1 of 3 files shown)


Itay4:
@guyfreund add release notes and then ask @kirbles19 to review

guyfreund:
@kirbles19 please review the CHANGELOG file

PR #4430, opened 2019-09-18
Added email authenticity check field to the phishing layout of 4.5 (summary page)

idovandijk:
# Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19339
related to: https://github.com/demisto/etc/issues/17721

## Description
Added the field that was added last release, to the layout of phishing in 4.5. For 5.0, there will be a different PR.

4.5 (will be on 4.1 too, but field doesn't exist there, so it won't be shown)
Separate issue for all 5.0 fields and layout: https://github.com/demisto/etc/issues/17841

   - No

## Must have
- [x] Code Review


idovandijk:
Docs were approved when the field was added. Now it's just added to the layout.

PR #4432, opened 2019-09-18
Limit the number of emails ingested per fetch

omercnet:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18903

## Description
Throttle number of incidents created per fetch 

## Screenshots

## Related PRs

branch | PR
------ | ------

x.x.x

   - Yes
       - Further details:
   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


Integrations/EWSv2/EWSv2.py
```diff
@@ -103,6 +104,10 @@
 AUTO_DISCOVERY = False
 SERVER_BUILD = ""
 MARK_AS_READ = demisto.params().get('markAsRead', False)
+MAX_FETCH = int(demisto.params().get('maxFetch', 50))
```
BenJoParadise: `demisto.params().get('maxFetch', 50)` can return `None`, which will cause exception (`TypeError: int() argument must be a string or a number, not 'NoneType'`) and is generally a bad idea (e.g. did you know that `None < 0` is true in python?)

To avoid this, split into two parts:
```
MAX_FETCH = demisto.params().get('maxFetch')
MAX_FETCH = MAX_FETCH if MAX_FETCH and 0 < MAX_FETCH <= 50 else 50
```

The bonus is you don't need the if statement in the next 2 lines.

Integrations/EWSv2/EWSv2.py
```diff
@@ -873,7 +878,7 @@ def get_last_run():
         last_run = {
             LAST_RUN_TIME: None,
             LAST_RUN_FOLDER: FOLDER_NAME,
-            LAST_RUN_IDS: None
+            LAST_RUN_IDS: []
```
BenJoParadise: Backwards compatibility: This will still be `None` on customer envs, make sure you handle both cases.

Integrations/EWSv2/EWSv2.py
```diff
@@ -1150,25 +1157,32 @@ def parse_incident_from_item(item, is_fetch):
...
 def fetch_emails_as_incidents(account_email, folder_name):
-    start_time = EWSDateTime.now(tz=EWSTimeZone.timezone('UTC'))
     last_run = get_last_run()
 
     try:
         account = get_account(account_email)
         last_emails = fetch_last_emails(account, folder_name, last_run.get(LAST_RUN_TIME), last_run.get(LAST_RUN_IDS))
 
-        ids = []
+        ids = deque(last_run.get(LAST_RUN_IDS, []), maxlen=LAST_RUN_IDS_QUEUE_SIZE)
```
BenJoParadise: Same as previous comments, here it is pretty much guaranteed that `None` will be returned after the content update. Use the same method as above to ensure that `None` is replaced with `[]`.

Integrations/EWSv2/EWSv2.py
```diff
@@ -1150,25 +1157,32 @@ def parse_incident_from_item(item, is_fetch):
...
+        last_run_time = incident.get('occurred', last_run.get(LAST_RUN_TIME))
+        if isinstance(last_run_time, EWSDateTime):
+            last_run_time = last_run_time.ewsformat()
+
         new_last_run = {
-            LAST_RUN_TIME: start_time.ewsformat(),
+            LAST_RUN_TIME: last_run_time,
             LAST_RUN_FOLDER: folder_name,
-            LAST_RUN_IDS: ids,
+            LAST_RUN_IDS: list(ids),
```
BenJoParadise: This will not work for if `ids` is `None`: `TypeError: 'NoneType' object is not iterable`

Integrations/EWSv2/EWSv2.yml
```diff
@@ -77,6 +77,11 @@ configuration:
   name: requestTimeout
   required: false
   type: 0
+- display: Max incidents per fetch
+  name: maxFetch
+  defaultvalue: "50"
+  type: 0
+  required: true
```
BenJoParadise: According to the hooks, this breaks backwards compatibility.
Even though this has a default value here, the existing integration configs in the DB don't have this key.
This should be OK, since you handle that case in the code, but:
1. The hook doesn't check that, so you will need to bypass it somehow...
2. There is no guarantee that an existing configuration will survive the content update (the integration may still fail simply because the server logic requires all mandatory fields to have a value, and the default value is not applied to existing configurations, this needs further testing).

The simpler solution is to set `required: false` (and handle the case where it is empty in the code, which you already do).

Integrations/EWSv2/EWSv2.py
```diff
@@ -103,6 +104,8 @@
 AUTO_DISCOVERY = False
 SERVER_BUILD = ""
 MARK_AS_READ = demisto.params().get('markAsRead', False)
+MAX_FETCH = min(50, int(demisto.params().get('maxFetch', 50)))
```
omercnet: @BenJoParadise see this solution instead?

Integrations/EWSv2/EWSv2.py
omercnet: fixed in https://github.com/demisto/content/pull/4432/files#diff-f868b5eac3f32b0f92afbc8ec448c70fR886

Integrations/EWSv2/EWSv2.py
omercnet: fixed in https://github.com/demisto/content/pull/4432/files#diff-f868b5eac3f32b0f92afbc8ec448c70fR886

omercnet:
why would it be lost?

the incident list stops appending after hitting the limit
https://github.com/demisto/content/blob/feature/ewsfetchlimit/Integrations/EWSv2/EWSv2.py#L1177-L1178

and then the last run timestamp is of the last fetched email
https://github.com/demisto/content/blob/feature/ewsfetchlimit/Integrations/EWSv2/EWSv2.py#L1180-L1182

so next time it runs it'll fetch from the last timestamp (i.e., the 50th email) and will get the remaining 20

Integrations/EWSv2/EWSv2.py
```diff
@@ -873,11 +876,15 @@ def get_last_run():
...
             LAST_RUN_FOLDER: FOLDER_NAME,
-            LAST_RUN_IDS: None
+            LAST_RUN_IDS: []
         }
     if LAST_RUN_TIME in last_run and last_run[LAST_RUN_TIME] is not None:
         last_run[LAST_RUN_TIME] = EWSDateTime.from_string(last_run[LAST_RUN_TIME])
-
+    
+    # In case we have existing last_run data
+    if last_run.get(LAST_RUN_IDS) === None:
```
IkaDemisto: === is invalid syntax

Itay4:
@kirbles19 please review yaml and changelog

kirbles19:
@omercnet @Itay4 Done.

Itay4:
@omercnet build fails on linting

omercnet:
fixed 🤞

yaakovi:
@omercnet @IkaDemisto 
can this be merged?

PR #4445, opened 2019-09-20
Domain Tools Iris - First Release

ChuckWoodraska:
## Status
Ready

4.1.0

   - No

## Must have
- [x] Tests
- [x] Documentation (with link to it) Not sure where this goes in the repo.
- [ ] Code Review

## Additional changes
Add List named tags for the IrisTags playbook

## Technical writer review
- [x] [YAML file](https://github.com/ChuckWoodraska/content/blob/master/Integrations/DomainTools_Iris/iris.yml)
- [x] [CHANGELOG](https://github.com/ChuckWoodraska/content/blob/master/Integrations/DomainTools_Iris/CHANGELOG.md)


Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
+from CommonServerPython import *
+import requests
+
+# disable insecure warnings
+requests.packages.urllib3.disable_warnings()
+
+''' GLOBALS '''
+
+BASE_URL = 'http://api.domaintools.com'
```
Itay4: Is this the only URL possible?

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+YOUNG_DOMAIN_TIMEFRAME = int(demisto.params().get('young_domain_timeframe'))
+
+''' HELPER FUNCTIONS '''
+
+
+def http_request(method, path, other_params=None):
+    """
+    HTTP request helper function
+    """
+    params = {'app_partner': 'demisto', 'app_name': 'Iris Plugin', 'app_version': '1', 'api_username': USERNAME,
```
Itay4: Please lint properly

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+              'api_key': API_KEY}
+    if other_params:
+        params.update(other_params)
+    url = '{}{}'.format(BASE_URL, path)
+    res = requests.request(
+        method=method,
+        url=url,
+        params=params
+    )
+    res_json = res.json()
```
Itay4: This might raise `JsonDeocdeError`, please wrap it in `try-except` block

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    if not res.ok:
+        txt = 'error in URL {} status code: {} reason: {}'.format(url, res.status_code,
+                                                                  res_json['error'].get('message'))
+        demisto.error(txt)
+        raise Exception(txt)
+
+    try:
+        return res_json.get('response')
```
Itay4: Why is this wrapped in `try-except` ? What can fail here ?

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    except Exception as ex:
+        demisto.debug(str(ex))
+        demisto.results({"Type": entryTypes["error"], "ContentsFormat": formats["text"], "Contents": res.text})
```
Itay4: An Exception should be raised, no `demisto.results()`

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+
+def test_module():
+    res = http_request('GET', '/v1/iris-investigate/', {'domain': 'demisto.com'})
+    if res.get('results_count'):
+        demisto.results('ok')
+    else:
+        demisto.results('No results found.')
+
+
+try:
```
Itay4: Wrap this in a `main` function

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    else:
+        demisto.results('No results found.')
+
+
+def test_module():
+    res = http_request('GET', '/v1/iris-investigate/', {'domain': 'demisto.com'})
+    if res.get('results_count'):
+        demisto.results('ok')
+    else:
+        demisto.results('No results found.')
```
Itay4: This will fail the test, is this the expected behaviour ?

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+            if not isinstance(v, int) and (v is None or len(v) == 0):
+                data_obj[k] = None
+    elif isinstance(data_obj, list) and (data_obj is None or len(data_obj) != 0):
+        for index, item in enumerate(data_obj):
+            convert_empty_to_null(item)
+            if not isinstance(item, int) and (item is None or len(item) == 0):
+                data_obj[index] = None
+
+
+def create_context_output(domain_result):
```
Itay4: Please change to more specific function name, and not so generic one, something like `create_domain_context_outputs`

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    return domain_context
+
+
+def format_domain_profile(response):
+    domain_result = response['results'][0]
+    domain_context = create_context_output(domain_result)
+    outputs = {"Domain(val.Name == obj.Name)": domain_context}
+    convert_empty_to_null(outputs)
+
+    human_readable = '### DomainTools Profiled {}. See context data for more info.'.format(domain_result['domain'])
```
Itay4: Users should not look at the context data for outputs, let's return a table here.

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+               "Domain Age (in days)",
+               "Website Response",
+               "Google Adsense",
+               "Google Analytics",
+               "Alexa Rank",
+               "Tags"]
+    human_readable = tableToMarkdown('DomainTools Domain Analytics for {}.'.format(domain_result['domain']),
+                                     human_readable_data,
+                                     headers=headers)
+    return human_readable, outputs
```
Itay4: You are always returning empty dictionary here in `outputs`, what are you trying to do ?

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    else:
+        demisto.results('No results found.')
+
+
+def domain_pivot():
+    res = None  # type: ignore
+    search_type = ""
+    search_value = ""
+    if demisto.args().get('ip'):
+        res = http_request('GET', '/v1/iris-investigate/', {'ip': demisto.args().get('ip')})
```
Itay4: Let's just initialise the parameters here and call `http_request` after the conditions block

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+    elif demisto.command() == 'domain':
+        domain_profile()
+    elif demisto.command() == 'domain-analytics':
+        domain_analytics()
+    elif demisto.command() == 'threat-profile':
+        threat_profile()
+    elif demisto.command() == 'domain-pivot':
+        domain_pivot()
+except Exception as e:
+    return_error('Unable to perform command : {}, Reason: {}'.format(demisto.command(), e))
```
Itay4:
```suggestion
    return_error('Unable to perform command : {}, Reason: {}'.format(demisto.command(), str(e)))
```

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+                                     headers=headers)
+    return human_readable, outputs
+
+
+def format_domain_pivot(response, search_type, search_value):
+    domain_list = [x['domain'] for x in response['results']]
+    human_readable_data = domain_list
+    domain_context_list = []
+
+    for domain_result in response['results']:
```
Itay4: There are two iterations over `response['results']`, no need for that, let's do it in once.

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+
+def get_threat_component(components, threat_type):
+    for component in components:
+        if component['name'] == threat_type:
+            return component
+    else:
+        return None
+
+
+def convert_empty_to_null(data_obj):
```
Itay4: Not sure why we need this function. There's an argument `removeNull` of `tableToMarkdown`, try using it instead

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+requests.packages.urllib3.disable_warnings()
+
+''' GLOBALS '''
+
+BASE_URL = 'http://api.domaintools.com'
+USERNAME = demisto.params().get('username')
+API_KEY = demisto.params().get('apikey')
+RISK_THRESHOLD = int(demisto.params().get('risk_threshold'))
+PIVOT_COUNT_THRESHOLD = int(demisto.params().get('pivot_count_threshold'))
+YOUNG_DOMAIN_TIMEFRAME = int(demisto.params().get('young_domain_timeframe'))
```
Itay4: This is not used

Integrations/DomainTools_Iris/iris.py
```diff
@@ -0,0 +1,393 @@
...
+# disable insecure warnings
+requests.packages.urllib3.disable_warnings()
+
+''' GLOBALS '''
+
+BASE_URL = 'http://api.domaintools.com'
+USERNAME = demisto.params().get('username')
+API_KEY = demisto.params().get('apikey')
+RISK_THRESHOLD = int(demisto.params().get('risk_threshold'))
+PIVOT_COUNT_THRESHOLD = int(demisto.params().get('pivot_count_threshold'))
```
Itay4: This is not used

Integrations/DomainTools_Iris/iris.py
ChuckWoodraska: Yes.

Integrations/DomainTools_Iris/iris.py
ChuckWoodraska: This cleans up the DomainTools context data. `removeNull` would not do what I am doing.

Integrations/DomainTools_Iris/iris.py
ChuckWoodraska: My linter doesn't see a problem here and the circle ci also didn't point anything out. Do you have an example of how you want this to look?

Integrations/DomainTools_Iris/iris.py
ChuckWoodraska: Do you have an example of this? I don't see any other integration doing this.

Itay4: Yes, many.
Look at [Proofpoint](https://github.com/demisto/content/tree/master/Integrations/ProofpointTAP_v2) for one

ChuckWoodraska: When I wrap it in main the code no longer works in demisto. The test module can't run and all the other commands leave dbot hanging. Is there anything special you have to do to make it work with wrapping in main?

Itay4: Did you add these 2 lines to the code:
```
if __name__ in ('__main__', 'builtin', 'builtins'):
    main()
```

ChuckWoodraska: Yes.

Itay4: Can you push the code so I can see it?

ChuckWoodraska: Code pushed.

ChuckWoodraska:
I believe all the suggested changes for the integration are done other than what I have left unresolved. Still working on updating playbooks and scripts due to changes in the integration.

Integrations/DomainTools_Iris/iris.py
Itay4: im able to get results with the following code:
```
def main():
    """
    Main Demisto function.
    """
    try:
        if demisto.command() == 'test-module':
            test_module()
        elif demisto.command() == 'domain':
            domain_profile_command()
        elif demisto.command() == 'domaintoolsiris-analytics':
            domain_analytics_command()
        elif demisto.command() == 'domaintoolsiris-threat-profile':
            threat_profile_command()
        elif demisto.command() == 'domaintoolsiris-pivot':
            domain_pivot_command()
    except Exception as e:
        return_error('Unable to perform command : {}, Reason: {}'.format(demisto.command(), str(e)))

if __name__ in ['__main__', 'builtin', 'builtins']:
    main()
```
image

ChuckWoodraska: Ya I have that and the test module fails and the integration commands never return. Would version of Demisto or anything like that affect it? I'm on version 4.5.0 Build 40783

ChuckWoodraska:
Also can you point to examples of testing automation scripts and playbooks? I'm not seeing much in the docs on the best way to do this.

Itay4:
> Also can you point to examples of testing automation scripts and playbooks? I'm not seeing much in the docs on the best way to do this.

Example for automation script test: [ParseEmailFiles](https://github.com/demisto/content/blob/master/Scripts/ParseEmailFiles/parse_email_files_test.py)

Example for playbook test: [Phishing playbook](https://github.com/demisto/content/blob/master/TestPlaybooks/playbook-Phishing_v2_Test_-_Inline.yml)

Integrations/DomainTools_Iris/iris.py
Itay4: Demisto version shouldn't matter.
What does the test module fail on?

ChuckWoodraska: I get this message when I click the test button in the integration settings. image

It works fine wrapping in main.

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        domain_result: DomainTools domain data
+
+    Returns: Dict with context data
+
+    """
+    domain = domain_result.get('domain')
+    ip_addresses = ', '.join([x.get('address').get('value') for x in domain_result.get('ip')])
+    create_date = domain_result.get('create_date').get('value')
+    expiration_date = domain_result.get('expiration_date').get('value')
+    domain_name_servers = ', '.join([x.get('host').get('value') for x in domain_result.get('name_server')])
```
Itay4: it is better to return a list to the context, and not a string

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    Creates all the context data necessary given a domain result
+    Args:
+        domain_result: DomainTools domain data
+
+    Returns: Dict with context data
+
+    """
+    domain = domain_result.get('domain')
+    ip_addresses = ', '.join([x.get('address').get('value') for x in domain_result.get('ip')])
+    create_date = domain_result.get('create_date').get('value')
```
Itay4:
```suggestion
    create_date = domain_result.get('create_date', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    expiration_date = domain_result.get('expiration_date').get('value')
```
Itay4:
```suggestion
    expiration_date = domain_result.get('expiration_date', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    domain_name_servers = ', '.join([x.get('host').get('value') for x in domain_result.get('name_server')])
+    name_servers = domain_result.get('name_server')
+    domain_status = domain_result.get('active')
+
+    proximity_risk_score = 0
+    threat_profile_risk_score = 0
+    threat_profile_threats = ''
+    threat_profile_evidence = ''
+
+    overall_risk_score = domain_result.get('domain_risk').get('risk_score')
```
Itay4:
```suggestion
    overall_risk_score = domain_result.get('domain_risk', {}).get('risk_score')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    risk_components = domain_result.get('domain_risk').get('components')
```
Itay4:
```suggestion
    risk_components = domain_result.get('domain_risk', {}).get('components')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    if len(risk_components):
```
Itay4:
```suggestion
    if risk_components:
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        blacklist_data = get_threat_component(risk_components, 'blacklist')
+        if proximity_data:
+            proximity_risk_score = proximity_data.get('risk_score')
+        elif blacklist_data:
+            proximity_risk_score = blacklist_data.get('risk_score')
+        threat_profile_data = get_threat_component(risk_components, 'threat_profile')
+        if threat_profile_data:
+            threat_profile_risk_score = threat_profile_data.get('risk_score')
+            threat_profile_threats = ', '.join(threat_profile_data.get('threats', []))
+            threat_profile_evidence = ', '.join(threat_profile_data.get('evidence', []))
```
Itay4: it is better to return a list to the context, and not a string

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+
+    website_response = domain_result.get('website_response')
+    google_adsense = domain_result.get('adsense').get('value')
```
Itay4:
```suggestion
    google_adsense = domain_result.get('adsense', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    google_analytics = domain_result.get('google_analytics').get('value')
```
Itay4:
```suggestion
    google_analytics = domain_result.get('google_analytics', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    alexa_rank = domain_result.get('alexa')
+    tags = domain_result.get('tags')
+
+    registrant_name = domain_result.get('registrant_name').get('value')
+    registrant_org = domain_result.get('registrant_org').get('value')
+    domain_registrant_contact = {
+        'Country': domain_result.get('registrant_contact').get('country').get('value'),
+        'Email': ', '.join([x.get('value') for x in domain_result.get('registrant_contact').get('email')]),
```
Itay4: it is better to return a list to the context, and not a string

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+            threat_profile_threats = ', '.join(threat_profile_data.get('threats', []))
+            threat_profile_evidence = ', '.join(threat_profile_data.get('evidence', []))
+
+    website_response = domain_result.get('website_response')
+    google_adsense = domain_result.get('adsense').get('value')
+    google_analytics = domain_result.get('google_analytics').get('value')
+    alexa_rank = domain_result.get('alexa')
+    tags = domain_result.get('tags')
+
+    registrant_name = domain_result.get('registrant_name').get('value')
```
Itay4:
```suggestion
    registrant_name = domain_result.get('registrant_name', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    registrant_org = domain_result.get('registrant_org').get('value')
```
Itay4:
```suggestion
    registrant_org = domain_result.get('registrant_org', {}).get('value')
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    domain_registrant_contact = {
+        'Country': domain_result.get('registrant_contact').get('country').get('value'),
```
Itay4: no need to get `registrant_contact` so many times, it is better to init it first `registrant_contact = domain_result.get('registrant_contact', {}` and then use it to get the required fields

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    }
+    technical_contact = {
+        'Country': domain_result.get('technical_contact').get('country'),
+        'Email': domain_result.get('technical_contact').get('email'),
+        'Name': domain_result.get('technical_contact').get('name'),
+        'Phone': domain_result.get('technical_contact').get('phone'),
+    }
+
+    billing_contact = {
+        'Country': domain_result.get('billing_contact').get('country'),
```
Itay4: no need to get `registrant_contact` so many times, it is better to init it first `billing_contact = domain_result.get('billing_contact', {}` and then use it to get the required fields

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        'Email': domain_result.get('billing_contact').get('email'),
+        'Name': domain_result.get('billing_contact').get('name'),
+        'Phone': domain_result.get('billing_contact').get('phone'),
+    }
+    soa_email = [x.get('value') for x in domain_result.get('soa_email')]
```
Itay4: please use meaningful variable names and not `x`

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    ssl_email = [x.get('value') for x in domain_result.get('ssl_email')]
+    email_domains = [x.get('value') for x in domain_result.get('email_domain')]
+    additional_whois_emails = domain_result.get('additional_whois_email')
+    domain_registrant = domain_result.get('registrar').get('value') if isinstance(domain_result.get('registrar'),
```
Itay4:
```suggestion
    domain_registrant = domain_result.get('registrar', {}).get('value') if isinstance(domain_result.get('registrar'),
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+                                                                                  dict) else domain_result.get(
+        'registrar')
+    registrar_status = domain_result.get('registrar_status')
+    ip_data = domain_result.get('ip')
+    ip_country_code = domain_result.get('ip')[0] if len(domain_result.get('ip')) else ''
```
Itay4:
```suggestion
    ip_country_code = domain_result.get('ip')[0] if domain_result.get('ip') else ''
```

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    mx_servers = domain_result.get('mx')
+    spf_info = domain_result.get('spf_info')
+    ssl_certificates = domain_result.get('ssl_info')
+    redirects_to = domain_result.get('redirect')
+
+    domain_tools_context = {
+        'Name': domain,
+        'LastEnriched': datetime.now().strftime('%Y-%m-%d'),
```
Itay4: are we sure the string date format will always be this ?

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+
+    if not res.ok:
+        error_message = res_json.get('error').get('message')
+        txt = 'error in URL {} status code: {} reason: {}'.format(url, res.status_code, error_message)
+        demisto.error(txt)
+
+    return res_json.get('response')
+
+
+def get_dbot_score(proximity_score, age, threat_profile_score):
```
Itay4: DBot Score should be a number, not a string

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        'DNS': ip_addresses,
+        'Vendor': 'DomainTools',
+        'CreationDate': create_date,
+        'RiskScore': overall_risk_score,
+        'DomainStatus': domain_status,
+        'ExpirationDate': expiration_date,
+        'NameServers': domain_name_servers,
+        'Registrant': domain_registrant_contact
+    }
+    if overall_risk_score and overall_risk_score >= RISK_THRESHOLD:
```
Itay4: it is better to check if the DBot Score is 3, and then init the `Malicious` object

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    Profiles domain and gives back all relevant domain data
+    Args:
+        domain:
+
+    Returns: All data relevant for Demisto command.
+
+    """
+    results = {'human_readable': 'No results found.', 'context': {}, 'raw': None}
+    response = http_request('GET', '/v1/iris-investigate/', {'domain': domain})
+    results['raw'] = response
```
Itay4: all these logic of building the outputs should be in the command function

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        context = create_domain_context_outputs(domain_result)
+        outputs = {'Domain(val.Name && val.Name == obj.Name)': context.get('domain'),
+                   'DomainTools.Domains(val.Name && val.Name == obj.Name)': context.get('domaintools'),
+                   'DBotScore': context.get('dbotscore')}
+        convert_empty_to_null(outputs)
+
+        human_readable_data = {
+            'Name': domain,
+            'Last Enriched': datetime.now().strftime('%Y-%m-%d'),
+            'Overall Risk Score': context.get('domaintools').get('Analytics').get('OverallRiskScore'),
```
Itay4: same here about initializing an object and then getting from it

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    Args:
+        domain (str):
+
+    Returns: All data relevant for Demisto command.
+
+    """
+    results = {'human_readable': 'No results found.', 'context': {}, 'raw': None}
+    response = http_request('GET', '/v1/iris-investigate/', {'domain': domain})
+    results['raw'] = response
+    if response.get('results_count'):
```
Itay4: all these logic of building the outputs should be in the command function

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    Args:
+        domain:
+
+    Returns: All data relevant for Demisto command.
+
+    """
+    results = {'human_readable': 'No results found.', 'context': {}, 'raw': None}
+    response = http_request('GET', '/v1/iris-investigate/', {'domain': domain})
+    results['raw'] = response
+    if response.get('results_count'):
```
Itay4: all these logic of building the outputs should be in the command function

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+
+    Returns: All data relevant for Demisto command.
+
+    """
+    results = {'human_readable': 'No results found.', 'context': {}, 'raw': None}
+    response = http_request('GET', '/v1/iris-investigate/', search_data)
+    results['raw'] = response
+    human_readable_data = []
+    domain_context_list = []
+    if response.get('results_count'):
```
Itay4: all these logic of building the outputs should be in the command function

Integrations/DomainTools_Iris/DomainTools_Iris.yml
```diff
@@ -0,0 +1,489 @@
...
+- defaultvalue: '7'
+  display: Young Domain Timeframe (within Days)
+  name: young_domain_timeframe
+  required: true
+  type: 0
+- display: Trust any certificate (not secure)
+  name: insecure
+  required: false
+  type: 8
+- display: Use system proxy settings
```
Itay4: this is not used in the code, please do so with `handle_proxy`

Integrations/DomainTools_Iris/DomainTools_Iris.yml
```diff
@@ -0,0 +1,489 @@
...
+  display: High-Risk Threshold
+  name: risk_threshold
+  required: true
+  type: 0
+- defaultvalue: '7'
+  display: Young Domain Timeframe (within Days)
+  name: young_domain_timeframe
+  required: true
+  type: 0
+- display: Trust any certificate (not secure)
```
Itay4: this is not used in the code, please do so with passing `verify` to `requests`

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+            'RedirectsTo': redirects_to,
+            'GoogleAdsenseTrackingCode': google_adsense,
+            'GoogleAnalyticTrackingCode': google_analytics
+        }
+    }
+
+    domain_context = {
+        'Name': domain,
+        'DNS': ip_addresses,
+        'Vendor': 'DomainTools',
```
Itay4: this shouldn't be an output under `Domain`, please remove

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        'CreationDate': create_date,
+        'RiskScore': overall_risk_score,
```
Itay4: this shouldn't be an output under `Domain`, please remove

Integrations/DomainTools_Iris/DomainTools_Iris.yml
```diff
@@ -0,0 +1,489 @@
...
+    - contextPath: DomainTools.PivotedDomains.Hosting.IPCountryCode
+      description: IP Country Code
+      type: Unknown
+    - contextPath: DomainTools.PivotedDomains.Hosting.MailServers
+      description: Mail Servers Info
+      type: Unknown
+    - contextPath: DomainTools.PivotedDomains.Hosting.SPFRecord
+      description: SPF Record Info
+      type: Unknown
+    - contextPath: DomainTools.PivotedDomains.Hosting.NameServers
```
Itay4: this seem to a list of objects, need to specify these here

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
```diff
@@ -0,0 +1,194 @@
+id: eb5bb557-9de5-4bfc-8dbc-f1bebddd342a
```
Itay4:
```suggestion
id: DomainTools Iris - Test
```

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
```diff
@@ -0,0 +1,194 @@
...
+version: -1
+name: DomainToolsTestPlaybook
```
Itay4:
```suggestion
name: DomainTools Iris - Test
```

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
```diff
@@ -0,0 +1,194 @@
...
+        "width": 1200,
+        "x": 160,
+        "y": -90
+      }
+    }
+  }
+inputs:
+- key: domain
+  value:
+    simple: ${Domain.Name}
```
Itay4: you should provide a domain here

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        'NameServers': domain_name_servers,
+        'Registrant': domain_registrant_contact
+    }
+    if overall_risk_score and overall_risk_score >= RISK_THRESHOLD:
+        domain_context['Malicious'] = {
+            'Vendor': 'DomainTools',
+            'Description': threat_profile_evidence
+        }
+    dbot_score = 'Unknown'
+    if create_date != '':
```
Itay4: what is the logic here ? how is `create_date` related to the score?

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+
+    """
+    for component in components:
+        if component.get('name') == threat_type:
+            return component
+    else:
+        return None
+
+
+def convert_empty_to_null(data_obj):
```
Itay4: this function leads to return `null` to the war room and it doesn't look good, or to the context, which loads it with empty values
in case of null, the relevant cell in the table should be either empty or not appear at all, and should be populated to the context data

Integrations/DomainTools_Iris/DomainTools_Iris.py
ChuckWoodraska: `create_date` lets us find how old the domain is. Younger domains are considered riskier to DomainTools.

Itay4: but in that case, if there is no `creation_date`, then any risk score will get an `Unknown` score, for example:
image

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+        'RiskScore': overall_risk_score,
+        'DomainStatus': domain_status,
+        'ExpirationDate': expiration_date,
+        'NameServers': domain_name_servers,
+        'Registrant': domain_registrant_contact
+    }
+    if overall_risk_score and overall_risk_score >= RISK_THRESHOLD:
+        domain_context['Malicious'] = {
+            'Vendor': 'DomainTools',
+            'Description': threat_profile_evidence
```
Itay4: Need to make sure the description is not empty or `null`

Integrations/DomainTools_Iris/DomainTools_Iris.py
ChuckWoodraska: We are creating so we should know.

Integrations/DomainTools_Iris/DomainTools_Iris.yml
ChuckWoodraska: I don't understand the requested change here. Is there a type that is list? I only see string, number, boolean, date, and unknown as available types.

Integrations/DomainTools_Iris/DomainTools_Iris.py
ChuckWoodraska: Changed the function to prune any empty or null attributes.

ChuckWoodraska:
Is the repo for docker images an entirely different repo? Or where does that image/dockerfile have to be?

Integrations/DomainTools_Iris/DomainTools_Iris.yml
Itay4: what im saying is that the output looks like this:
image
if you wont specify `domain.value` for example as an output, this wont be usable

Itay4:
> Is the repo for docker images an entirely different repo? Or where does that image/dockerfile have to be?

you can see the python 3 docker image [here](https://github.com/demisto/dockerfiles/tree/master/docker/python3)
in order to use it in your integration, you need to specify in the yml `dockerimage: demisto/python3:3.7.4.2718`

Integrations/DomainTools_Iris/DomainTools_Iris.yml
ChuckWoodraska: Hmm, talking to my PM we would like to keep it the way it is and our playbooks and documentation would give examples of how to interact with the items not explicitly listed. Would this be alright?

Itay4: sorry, but not really
without defining the context path properly, the integration won't be usable

ChuckWoodraska: Maybe I'm missing something, but how does it make it unusable? The data is still there and you can still access it.

Itay4: it is not usable in sense that the user has to manipulate the output in order to use it, and that only if he first runs the command in the war room and look at the context data to look at how the output looks
this is not good ux

Playbooks/playbook-DomainTools_Domain_Auto_Enrichment.yml
```diff
@@ -0,0 +1,170 @@
+id: DomainTools_Domain_Auto_Enrichment
```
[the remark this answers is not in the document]
EngageCloud: Hi @idovandijk, thanks again for your help. I'm the responsible PM here at DT.

This playbook (and some others below e.g. Guided_Pivot_Enrichment, Tags) are specific use cases that DT users commonly execute within an IR. The default auto-extract may not work specifically in parsing our somewhat extensive data-format, due to it being extensive and complex. With this playbook, we wanted to give the user an option to further extend the Indicator table (a small set of custom fields are added at installation time) to make this possible.

I'm glad to hear fo the auto-extract capability, however not been successful in adding to the Indicator from the War-room. Also, thinking of instances when the enrichment is not performed on an existing Indicator.

Lastly, there are automation scripts that were written to support this for the user. These 2 playbooks will significantly reduce customization/ development time for our user-base and in parsing through our extensive dataset.

Hope this helps. I will let @ChuckWoodraska respond back to the remaining items. I just wanted to ensure I'm able to color the requirements here.

Playbooks/playbook-DomainTools_Iris_Tags.yml
```diff
@@ -0,0 +1,145 @@
+id: DomainTools_Iris_Tags
```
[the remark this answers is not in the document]
EngageCloud: @idovandijk referencing my above comment to playbook.

The intent was to provide the user with a template (a micro-playbook as you suggest) that they can further extend. The use-case is if an Indicator/ observable is already known malicious, the playbook changes the severity of the Incident. However, user can customize this to send out a Service ticket or create human approval. The playbook demonstrates how user can parse through the tags and retrieve the metadata. (P.S. - We are extending the Tags metadata further in Q1, and this will have a lot more actionable details).

If this seems trivial enough to be considered a playbook, how would you suggest we package this? I would have considered command, but need to be automated as a playbook for scaleability.

Playbooks/playbook-DomainTools_Guided_Pivot_Enrichment.yml
```diff
@@ -0,0 +1,616 @@
+id: DomainTools_Guided_Pivot_Enrichment
```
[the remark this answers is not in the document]
ChuckWoodraska: > What does Identity.TechnicalContact.Email.value} represent? What is the purpose of checking whether the .count is less than 500? Referring to task #4 in the playbook.

The value is the actual email address we want to pivot on. The count is how many other domains share that e-mail in our dataset. The check for less than 500 gives us a good pivot point threshold that gives us back domains that share this somewhat connected data attribute depending on what you pivot on.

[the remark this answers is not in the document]
ChuckWoodraska: > The task "SetIndicatorTableData" should be removed, as explained earlier. Referring to task #2.

We are setting custom fields in the indicator table using this task. Is there a way to do that with the auto extractor?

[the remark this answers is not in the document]
ChuckWoodraska: > Please divide the inputs so it can be used with any data (and not just from domaintools). So rather than having the input of ${Domain.DomainTools}, let's add individual, generic inputs, such as Email.Address, IP.Address and so on.

I'm not sure I can make this more generic as the domaintools data is needed because it contains the counts and actual values.

Playbooks/playbook-DomainTools_Domain_Auto_Enrichment.yml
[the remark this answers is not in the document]
idovandijk: Hi again. I'm happy to be of help. I will comment on the other playbooks separately.
As for this playbook, I feel like I should clear things up a bit:
As you said, a dependency of using your specific script that adds the indicator in this playbook, is custom fields at installation time, so we cannot release it as an out-of-the-box playbook that all customers will be using (as it won't work for them, unless they explicitly added those fields). As for the data retention - the way we data related to the investigation is through context. If the data returned from `!domain` is of use to you during an investigation, you shouldn't bother adding it all to the indicator data. The data returned from `!domain` will remain in context. The indicator stores specific information in some pre-existing fields, such as when the indicator was first seen, its score and a couple more.

Can you please elaborate on "not been successful in adding the indicator from the war-room"?
I will explain the flow. Whether the indicator existed before or not, once a URL or an email is involved in an incident in any way, auto-extract will extract the domain from it, and run `!domain` using all active supporting integrations (including DomainTools Iris). This will add all the data returned from `!domain` to the context. The user can then utilize this data to their needs. Subplaybooks can have this data as their inputs to work further on that data too. Furthermore, the indicator will automatically be stored in the "Indicators" page in Demisto. There is no need to actually store that information manually in the indicator itself.

Just an example would be your pivoting playbook. An investigation involving an interesting domain can take place, and then running your pivoting table, which will take the data returned from `!domain`, can use it to further enrich the data.

If you wish to retain the logic of that playbook due to the "last enriched" check,  then it is possible for you to wrap that in the code, or add another command such as `domaintools-enrich-if-outdated` (or something of the sort. Remember that you can still create this playbook locally for yourself, but I still was not convinced that this is an important playbook that all customers should be provided.

As a test for auto-extract, you can go to the war-room of any incident (or the playground). Use the `Print` command, just to print a URL of interest. Set the `auto-extract` setting to `inline`, which means the command will complete once auto-extract finished doing its job. Then look at your context and see that you have all the data returned from `!domain`.
image
Also, make sure your integration is enabled, otherwise you definitely won't have any data returned.

I hope this clarifies things a bit. If there are additional questions here or if you think I misunderstood, feel free to ask!

Playbooks/playbook-DomainTools_Guided_Pivot_Enrichment.yml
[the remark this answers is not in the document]
idovandijk: I further explained my point about setting the indicator data manually and auto-extract, above. If you still have questions regarding that, feel free to ask here.

Making this playbook generic is very much possible, I'll explain how it works.:

Once you change the inputs to the generic ones as I showed, by doing this, amongst other changes (the rest of the inputs should be added too):
image

then you will have to modify your tasks to utilize the data from the input, as follows:
image
image

At this point you have a good flow of playbook inputs going into task inputs.
Now, how do we make the playbook receive the inputs we want it to?
Well, the inputs of any playbook **can be modified** to fit the place where they are used. Let's use your test playbook as an example. In the following pictures you will see on the left - the generic inputs I set for that playbook, and on the right - the modification I did to the inputs in the context of a subplaybook.
image

What I did was change the inputs of the subplaybooks only when used within the context of my main playbook.

Itay4:
@ChuckWoodraska please pull from the base branch

Playbooks/playbook-DomainTools_Guided_Pivot_Enrichment.yml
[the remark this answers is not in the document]
ChuckWoodraska: I guess I'm still confused by the use of generic inputs in this playbook. If I used generic inputs, the playbook wouldn't do anything because it wouldn't know when to pivot. I see they can be modified, but the playbook makes no sense without the use of the counts available through DomainTools context data.

Playbooks/playbook-DomainTools_Iris_Tags.yml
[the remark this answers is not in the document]
idovandijk: Hey there. The use-case is clear to me, but we cover that in already existing playbooks :)
In this case your script sets the severity according to the domain tags, but that could affect the severity of an existing incident for the worst, too. E.g., if you were to have a "critical" incident which gets a lower severity of "high" because of this script when it sees a bad tag in a domain.
The wa we handle this is we have the `Calculate Severity - Generic v2` playbook. Among other assessments, it looks at the DBotScore of all indicators of the incident and decides the final severity of the incident accordingly.
What you should do is output a DBotScore with this command instead, so if a domain with a bad tag was involved, then it will be handled and taken into account with all other IOCs in `Calculate Severity - Generic v2` automatically (not need to edit that generic playbook).

As for the metadata of tags - the answer is the same as I explained in one of my other comments. Your task provides the output data, and this data can then be used anywhere in any playbook and any task, so there is no need to tie a user specific to a flow where incident severity is affected immediately and solely based on domain tags.

If and when additional possible logic / flow / use case arises, regarding tags of domain, we will definitely be willing and happy to add it as an out-of-the-box playbook for customers.
If there is additional data that you would like to utilize from the CheckTags script, then it should output it as well as a DBotScore, but not set a severity.

To summarize: it seems that we agree that the tags should affect the severity. I'm saying that by outputting DBotScore, our existing OOTB playbooks will be able to take the bad tags into account when calculating a severity for the incident. As such, it seems that the purpose of this playbook is already covered elsewhere :)

Playbooks/playbook-DomainTools_Guided_Pivot_Enrichment.yml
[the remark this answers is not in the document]
idovandijk: One of the inputs could be a hardcoded "IPAddressCount" / "IPPivotThreshhold" which will be 500. Then your task could use this playbook input.
Imagine someone wants to use this playbook for pivoting on indicators that he got from another integration - then you will have nothing under `${Domain.DomainTools}`, and nothing in your playbook input `domaintools_data` . To make it generic, the tasks should take the indicators from the playbook inputs, and the playbook inputs should take the indicators from standard context paths (Email.Address, IP.Address, File.MD5 and so on). 
Finally, when a playbook wants to use the guided pivot playbook, it will call it and modify its inputs to take the IP Addresses, URLs, Domains and so on, from wherever they want.
For example, Playbook A has `
domaintools_data.IPAddresses.address.value` in context. Playbook A calls Guided Pivot, so it will have to modify the inputs of Guided Pivot to take `
domaintools_data.IPAddresses.address.value` and put it under its `IPAddress` input. Guided Pivot will then pass `inputs.IPAddress` into its "IP Pivot" task.
Does the flow make more sense to you now? If there are still any hesitations or questions, ask me :)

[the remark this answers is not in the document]
ChuckWoodraska: I think I understand now. Thanks for your patience in explaining this.

ChuckWoodraska:
I can't get Circle CI to play nice with my typing import and if I don't have the import it complains about not having it. Any ideas how to solve this?

Itay4:
@ChuckWoodraska updated the docker image and seem to be ok now

ChuckWoodraska:
Anything else that needs to be changed?

Itay4:
@kirbles19 please review:
https://github.com/demisto/content/blob/dd2b0ddb52ee05fe0f1ec4629975ce43b817ad69/Integrations/DomainTools_Iris/DomainTools_Iris.yml
https://github.com/demisto/content/blob/dd2b0ddb52ee05fe0f1ec4629975ce43b817ad69/Integrations/DomainTools_Iris/DomainTools_Iris_description.md
https://github.com/demisto/content/blob/dd2b0ddb52ee05fe0f1ec4629975ce43b817ad69/Playbooks/playbook-Indicator_Pivoting-DomainTools_Iris.yml
https://github.com/demisto/content/blob/dd2b0ddb52ee05fe0f1ec4629975ce43b817ad69/Integrations/DomainTools_Iris/README.md

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
```diff
@@ -0,0 +1,1177 @@
+id: DomainTools Iris - Test
```
idovandijk: The test looks much better now. Good job!
Still need to check for some outputs. Choose some outputs that you expect to see from the call for `Indicator Pivoting - DomainTools Iris` to make sure that the playbook outputs what its supposed to. No need to test ALL the outputs though.
Also, is it incidental that there is nothing in `NameServerIPAddress` and others? Or are you inputting the wrong data?
image

Notice that 2 tasks ended up not running:
image

Playbooks/playbook-Indicator_Pivoting-DomainTools_Iris.yml
```diff
@@ -0,0 +1,1072 @@
+id: Indicator Pivoting - DomainTools Iris
```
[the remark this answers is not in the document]
idovandijk: Also please add a bit to the description of the playbook - what the pivoting points are for example

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+            if not isinstance(v, int) and (v is None or len(v) == 0):
+                items_to_prune.append(k)
+            elif k == 'count' and v == 0:
+                items_to_prune.append(k)
+        for k in items_to_prune:
+            del data_obj[k]
+    elif isinstance(data_obj, list) and len(data_obj):
+        for index, item in enumerate(data_obj):
+            prune_context_data(item)
+            if not isinstance(item, int) and (item is None or len(item) == 0):
```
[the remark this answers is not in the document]
ChuckWoodraska: This has to be the way I have it. If the terminal point is `[]` then this doesn't catch it.

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+    Does a deep dive through a data object to prune any null or empty items. Checks for empty lists, dicts, and strs.
+    Args:
+        data_obj: Either a list or dict that needs to be pruned
+    """
+    items_to_prune = []
+    if isinstance(data_obj, dict) and len(data_obj):
+        for k, v in data_obj.items():
+            if isinstance(data_obj[k], dict) or isinstance(data_obj[k], list):
+                prune_context_data(data_obj[k])
+            if not isinstance(v, int) and (v is None or len(v) == 0):
```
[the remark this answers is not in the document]
ChuckWoodraska: This has to be the way I have it. If the terminal point is `{}` then this doesn't catch it.

Integrations/DomainTools_Iris/DomainTools_Iris.py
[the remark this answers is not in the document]
Itay4: `if []` is `false`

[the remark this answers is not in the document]
ChuckWoodraska: Oops you are correct. It needs to be `not item` though.

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
ChuckWoodraska: I am testing fewer things in the pivot playbook to save time. Hopefully as this is just a demonstration of what the playbook can do this is alright. Otherwise the playbook takes awhile to run even in a test environment.

Integrations/DomainTools_Iris/DomainTools_Iris.py
```diff
@@ -0,0 +1,646 @@
...
+
+def test_module():
+    """
+    Tests the API key for a user.
+    """
+    try:
+        http_request('GET', '/v1/iris-investigate/', {'domain': 'demisto.com'})
+        demisto.results('ok')
+    except Exception as test_error:
+        return_error('Unable to perform command : {}, Reason: {}'.format(demisto.command(), str(test_error)))
```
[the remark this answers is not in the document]
Itay4: what i meant is `return_error` is not needed here, just raise

Itay4:
@ChuckWoodraska looks good, just note 1 comment i've unresolved
pretty much done on my side, waiting for @idovandijk approval and @kirbles19 review

TestPlaybooks/playbook-DomainTools_Iris_Test.yml
idovandijk: What you did is exactly what was needed :)

Playbooks/playbook-Indicator_Pivoting-DomainTools_Iris.yml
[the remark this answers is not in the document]
idovandijk: You happened to miss my point for the generic input paths. We have standard paths we use for IP & email addresses. Please add `IP.Address` and `Account.Email.Address` respectively, with the "Unique" Transformer.
image

Once that is done, we are good to go!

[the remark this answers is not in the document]
ChuckWoodraska: Weird, I see the inputs in my yaml file. I had to make a change anyway for the email as I was using just Email.Address so I pushed the updates.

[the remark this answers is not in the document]
idovandijk: Perfect

ChuckWoodraska:
Is there an ETA for the final review being done?

Itay4:
> Is there an ETA for the final review being done?

mid next week

richardbluestone:
I've reviewed the yml and md files.

richardbluestone:
All docs have been reviewed.

PR #4446, opened 2019-09-20
Split incidentfields.json file to individual files

yaakovi:
## Status
Ready

## Description
in order to simplify the validations done on incident fields.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


yaakovi:
no change logs - everything here is same as original

PR #4451, opened 2019-09-22
Microsoft Teams - single port support and integration health command

Itay4:
## Status
Ready

## Description
- Added support for single port mapping
- Added integration health command, which checks connectivity to Microsoft APIs and lists the mirrored channels

5.0

   - No

## Must have
- [x] Tests
- [ ] Documentation
- [ ] Code Review

## Dependencies
- [x] MicrosoftTeamsAsk

## Technical writer review
@kirbles19 please review:
- [ ] [YAML file](https://github.com/demisto/content/blob/bd2285bc9c3095dfe304395e6a01a493344b90d6/Integrations/MicrosoftTeams/MicrosoftTeams.yml)
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/bd2285bc9c3095dfe304395e6a01a493344b90d6/Integrations/MicrosoftTeams/CHANGELOG.md)

### DIFF
--- a/Integrations/MicrosoftTeams/MicrosoftTeams.py
+++ b/Integrations/MicrosoftTeams/MicrosoftTeams.py
@@ -432,7 +432,9 @@ def get_graph_access_token() -> str:
             return access_token
     tenant_id: str = integration_context.get('tenant_id', '')
     if not tenant_id:
-        raise ValueError('Tenant ID not found')
+        raise ValueError(
+            'Did not receive tenant ID from Microsoft Teams, verify the messaging endpoint is configured correctly.'
+        )
     url: str = f'https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token'
     data: dict = {
         'grant_type': 'client_credentials',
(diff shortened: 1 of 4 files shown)


kirbles19:
@Itay4 Done.

Integrations/MicrosoftTeams/MicrosoftTeams.yml
```diff
@@ -54,7 +54,7 @@ configuration:
   name: longRunning
   required: false
   type: 8
-- display: Port mapping, e.g. <host port>:<docker port> (Required for investigation
+- display: Port mapping, e.g. <port> or <host port>:<docker port> (Required for investigation
```
glicht: Let's remove the host:docker port option for this description. We can have this in the more detailed description. I think it will save confusions.
```suggestion
- display: Listen port, e.g. 7000 (Required for investigation
```

Integrations/MicrosoftTeams/MicrosoftTeams.py
```diff
@@ -1376,18 +1424,23 @@ def long_running_loop():
...
+    certificate_path = str()
+    private_key_path = str()
+
     try:
         port_mapping: str = PARAMS.get('longRunningPort', '')
+        port: int
         if port_mapping:
-            port: int = int(port_mapping.split(':')[0])
+            if ':' in port_mapping:
+                port = int(port_mapping.split(':')[0])
```
glicht: I think we should be taking the second port in this case and not the first. docker port is the second argument. First one is the host port.
```suggestion
                port = int(port_mapping.split(':')[1])
```

PR #4457, opened 2019-09-23
Revert "Update circleci config for a limited build on forks"

yaakovi:
Reverts demisto/content#4405

reverting for now until proper testing will be done to this change:
1. nightly doesn't run `instance_testing` job, only `build` job.
2. same for trigger scripts.
3. unit tests doesn't run due to lacking of python.

extra:
* `instance_testing` should output as artifacts all files (not just server logs).

note:
having two jobs that developers should keep an eye on can be confusing. we should consider different approach. 🤔

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,30 +1,36 @@
 version: 2
-
 jobs:
   build:
     docker:
       - image: devdemisto/content-build-2and3:2.0.0.1785  # disable-secrets-detection
+    resource_class: medium+
     environment:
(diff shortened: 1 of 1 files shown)


yaakovi:
1. we can switch our trigger (from a job in demisto) to `schedule` in the circle config:
https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow

yaakovi:
2. seems it is not supported for now...
https://ideas.circleci.com/ideas/CCI-I-192

PR #4473, opened 2019-09-24
Lowercase 'MD' to match file extension of destination URL

amshamah419:
As title suggests.

## Status
Ready

## Description
"The best bug fixes are the most simple bug fixes." - George Washington (Circa 1995)


only-says-nice:
> I would say nice

Nice

PR #4281, opened 2019-09-02
Template: Analytics and SIEM

jochman:
# Status
Ready

## Description
New integration templates

## Does it break backward compatibility?
   - No

## Must have
- [X] Tests
- [ ] Documentation (with link to it)
- [ ] Code Review

PR #4487, opened 2019-09-25
Update base client

jochman:
## Status
Ready

## Description
Added the ability to send `None` as `outputs` to `return_outputs`.
added the `xml` as `resp_type` option.

## Related PRs

branch | PR
https://github.com/demisto/content/tree/new-templates | https://github.com/demisto/content/pull/4281

   - No


Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1488,7 +1488,9 @@ def return_outputs(readable_output, outputs, raw_response=None):
         "Contents": raw_response,
         "EntryContext": outputs
     }
-
+    # Return 'ok' to test module
+    if readable_output == 'ok' and not outputs and not raw_response:
+        demisto.results('ok')
```
only-says-nice:
```suggestion
        demisto.results('Nice')
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1488,7 +1488,9 @@ def return_outputs(readable_output, outputs, raw_response=None):
         "Contents": raw_response,
         "EntryContext": outputs
     }
-
+    # Return 'ok' to test module
+    if readable_output == 'ok' and not outputs and not raw_response:
```
only-says-nice:
```suggestion
    if readable_output == 'Nice' and not outputs and not raw_response:
```

Scripts/CommonServerPython/CommonServerPython.py
anara123: when we return in test any string other than `ok` then the test should fail and present that error. Please check that scenario

kirbles19:
@jochman Done.

PR #4489, opened 2019-09-25
Added condition to cases where only headline exist in the HR

ShahafBenYakir:
## Status
Ready

## Description
If a command has only headline as a HR output the automation will fail.

### DIFF
--- a/Scripts/HTMLDocsAutomation/HTMLDocsAutomation.py
+++ b/Scripts/HTMLDocsAutomation/HTMLDocsAutomation.py
@@ -290,6 +290,8 @@ def human_readable_example_to_html(hr_sample):
             title = hr_sample
             if '\n' in hr_sample:
                 title, hr_sample = hr_sample.split('\n', 1)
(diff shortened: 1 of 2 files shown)


Scripts/HTMLDocsAutomation/HTMLDocsAutomation.py
```diff
@@ -287,7 +287,9 @@ def human_readable_example_to_html(hr_sample):
     hr_html: list = []
     while hr_sample:
         if hr_sample.startswith('#'):
-            title, hr_sample = hr_sample.split('\n', 1)
+            title = hr_sample
+            if '\n' in hr_sample:
```
yaakovi: in case `hr_sample` does not include `\n` you should do 2 things:
1) `title` should consist with the value of `hr_sample` (which you did)
2) `hr_sample` should be empty (which you didn't do)

kirbles19:
@ShahafBenYakir Done.

PR #4504, opened 2019-10-01
adding validation for unique file name

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19330

## Description
added a validation in the content_creator that checks that all file names are unique


content_creator.py
```diff
@@ -85,11 +91,17 @@ def copy_dir_yml(dir_name, version_num, bundle_pre, bundle_post, bundle_test):
     print ' - total post files: %d' % (post_files, )
 
 
-def copy_dir_json(dir_name, version_num, bundle_pre, bundle_post, bundle_test):
+def copy_dir_json(flie_names, dir_name, version_num, bundle_pre, bundle_post, bundle_test):
     # handle *.json files
     scan_files = glob.glob(os.path.join(dir_name, '*.json'))
     for path in scan_files:
         dpath = os.path.basename(path)
+        if odpath in file_names:
```
ronykoz:
```suggestion
        if dpath in file_names:
```

yaakovi:
this is not the problem from the issue.
at this point you are most likely never get conflicts of file names since each type of entity has a different prefix (the only collision may occur is with the BetaIntegration folder).
the problem specified in the issue is when handling the "packagify" of integration/script we are creating a new file that may override an existing file (since we are using `open(..., 'w')` and not checking if the file already exist)

Itay4:
@ronykoz can we close this and open a new PR in the SDK repo ?

PR #4517, opened 2019-10-02
add id == name validation

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19275

### DIFF
--- a/Tests/scripts/hook_validations/integration.py
+++ b/Tests/scripts/hook_validations/integration.py
@@ -69,6 +69,7 @@ class IntegrationValidator(object):
         self.is_proxy_configured_correctly()
         self.is_insecure_configured_correctly()
         self.is_valid_category()
+        self.is_valid_id()
 
         return self._is_valid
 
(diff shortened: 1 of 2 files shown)


Tests/scripts/hook_validations/integration.py
```diff
@@ -488,3 +489,14 @@ def is_docker_image_changed(self):
                 return True
 
         return False
+
+    def is_valid_id(self):
+        """Validate that the id equals to the name."""
+        _id = self.current_integration.get('commonfields', {}).get('id')
```
yaakovi: should be `id_`

https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments

Tests/scripts/hook_validations/script.py
```diff
@@ -80,7 +80,8 @@ def is_backward_compatible(self):
     def is_valid_script(self):
         """Check whether the Integration is valid or not, update the _is_valid field to determine that"""
         is_script_valid = any([
-            self.is_valid_subtype()
+            self.is_valid_subtype(),
+            self.is_valid_id()
```
yaakovi: this is a problem since we currently have some script in that form.
this should be applied to only new files

Itay4:
@ronykoz can we close this and open a new PR in the SDK repo ?

Tests/scripts/hook_validations/script.py
reutshal: @yaakovi  after the fix of @GalRabin this should be fine right?

yaakovi: should be fine now

ronykoz:
@reutshal as discussed please move this to the SDK, closing this PR for now

PR #4528, opened 2019-10-03
change guid in scripts

GalRabinDemisto:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/19289

## Description
Change id to readable id by following constraints:
1. only js scripts.
2. only not deprecated.
3. id isn't readable.

   - No

I used the following script:
```python3
import glob
import re
from tempfile import mkstemp
from shutil import move
from os import fdopen, remove

def is_deprecated(filename):
    with open(filename,mode='r') as file:
        for line in file:
            if 'deprecated: true' in line:
                return True
    return False

def is_wrong_id(filename):
    with open(filename,mode='r') as file:
        name = ''
        id = ''
        for idx ,line in enumerate(file):
            name_compiled = re.compile(r'.*name: (.*)\n')
            if name_compiled.search(line):
                name = name_compiled.search(line).group(1)

            wrong_id_compiled = re.compile(r'.*id: .*-.*-.*-.*-.*\n')
            if wrong_id_compiled.search(line):
                id = wrong_id_compiled.search(line)
                line_to_edit = idx

            if id and name:
                break

        if id:
            return name

def is_js(filename):
    with open(filename, mode='r') as file:
        for line in file:
            if 'type: javascript' in line:
                return True
    return False

def replace(filename, name):
    with open(filename, "r+") as f:
        data = f.read()
        f.seek(0)
        f.write(re.sub(r'(.*id: ).*-.*-.*-.*-.*\n', rf"\1{name}\n", data))
        f.truncate()

def main():
    for filepath in glob.iglob('/Users/grabin/dev/demisto/content/Scripts/*.yml'):
        if not is_deprecated(filepath) and is_wrong_id(filepath) and is_js(filepath):
            name = is_wrong_id(filepath)
            replace(filepath, name)

if __name__ == '__main__':
    main()

```


GalRabinDemisto:
@yaakovi 
Hi,
FYI I changed the validations files and it working now.
I will appreciate your review.
Thanks

Scripts/script-BlockIP.yml
```diff
@@ -1,6 +1,73 @@
 commonfields:
-  id: bffdcf72-2061-4767-83d5-3ff2a9e8afe7
-  version: 5
+  id: BlockIP
+  version: -1
+name: BlockIP
```
yaakovi: you have duplicates here, please fix.
same for all scripts

Scripts/script-BlockIP_CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## [Unreleased]
+- Change guid to readable format
```
yaakovi: shouldn't be included in the releasenotes. same for all change logs
```suggestion
-
```

PR #4556, opened 2019-10-07
Added support for European timestamp

guykeller:
## Status 
Ready

## Description
Added support for European timestamp.

   - No


yaakovi:
@kirbles19 we are merging this as it fixed broken build in master.
if you want to update it, contact @guykeller

kirbles19:
@guykeller @yaakovi Looks good

PR #4565, opened 2019-10-08
fix not returning entries

jochman:
## Status
Ready

## Description
fixes: https://github.com/demisto/etc/issues/19506

## Screenshots
test module:
image
general output:
image

  - No

## Must have
- [X] Tests
- [ ] Code Review

### DIFF
--- a/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Scripts/CommonServerPython/CommonServerPython.py
@@ -1491,11 +1491,11 @@ def return_outputs(readable_output, outputs=None, raw_response=None):
     # Return 'readable_output' only if needed
     if readable_output and not outputs and not raw_response:
         demisto.results(readable_output)
+        return
     elif outputs and raw_response is None:
         # if raw_response was not provided but outputs were provided then set Contents as outputs
         return_entry["Contents"] = outputs
-    else:
-        demisto.results(return_entry)
+    demisto.results(return_entry)
 
 
 def return_error(message, error='', outputs=None):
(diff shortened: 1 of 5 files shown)


jochman:
Done

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1491,10 +1491,10 @@ def return_outputs(readable_output, outputs=None, raw_response=None):
     # Return 'readable_output' only if needed
     if readable_output and not outputs and not raw_response:
         demisto.results(readable_output)
```
anara123: The better solution will be if you change here
```
return_entry["Contents"] = readable_output
return_entry["ContentsFormat"] = formats["text"]
demisto.results(return_entry)
```

anara123: this way if the test returns "ok" it will set ok to contents.
And in case someone uses the return_outputs to return just markdown to the warroom, the function will return markdown as it suppose to do, instead of just doing demisto.results()

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1490,12 +1490,13 @@ def return_outputs(readable_output, outputs=None, raw_response=None):
     }
     # Return 'readable_output' only if needed
     if readable_output and not outputs and not raw_response:
-        demisto.results(readable_output)
+        return_entry["Contents"] = readable_output
+        return_entry["ContentsFormat"] = formats["text"]
+        demisto.results(return_entry)
```
glicht: remove. You have the results call in the end. make sure the unit test check that results is called only once.

PR #4610, opened 2019-10-17
GetDockerImageLatestTag fixed bug

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19534

## Description
Now script works as intended and returns the latest tag

## Screenshots
This is the result given to demisto/python
image

Which is correct in the time of writing:
image

   - No

## Must have
- [ ] Code Review

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/docker_image_script_fix/Scripts/GetDockerImageLatestTag/CHANGELOG.md)


glicht:
If using the default registry (docker hub) then use the following logic as we do in `dockerfiles` project: 

https://github.com/demisto/dockerfiles/blob/master/docker/create_new_python_image.py#L38

Otherwise sort tags according to number logic instead of lex.

glicht:
Maybe worth checking parse_version: 

For example:
```
from pkg_resources import parse_version
print(parse_version('3.7.4.300') < parse_version('3.7.4.1000'))
```

kirbles19:
@roysagi Done.

PR #4614, opened 2019-10-17
API Token param -> encrypted

orenzohar:
## Status
Ready

## Description
change API token param type to encrypted in JIRA v2


Itay4:
@orenzohar did we verify the content installation effects a configured instance as expected ?

orenzohar:
@Itay4 yes, it just masks the parameter and doesn't delete the old value, so everything keeps working.

PR #4619, opened 2019-10-18
Change Script from Python 2.7 to Python3 to support ASCII Characters

Xboarder56:
Changed to Python3 as Python 2.7 does not support ASCII characters and causes the regexgroups function to fail. (Ex: Ïº)

## Status
Ready

## Related Issues
fixes: Issue with Python 2.7 not supporting ASCII input causing the python container to error. 

## Description
Changed to Python3 as Python 2.7 does not support ASCII characters and causes the regexgroups function to fail. (Ex: Ïº)

Tested on 4.5

   - No, This is utilizing the default demisto container for demisto/Python3:latest


anara123:
@Xboarder56 Thank you for the contribution.
The build fails, you need to make the build to pass.
The build fails because we are no longer keep the integration in yml file, we moved to package format.
Follow the guide  https://github.com/demisto/content/tree/master/docs/package_directory_structure.

Let me know if you any questions

Xboarder56:
@anara123 it looks like the build is failing due to the demisto:latest container containing python3.3 which specifically the build script looks for `pkg_dev_tes_tasks.py`. The container demisto/python3:latest is over 2 years old and leverages python 3.3 which causes the build to fail. 

Any advice on this? The automation works correctly on 3.3 (demisto/python3:latest).

Itay4:
@Xboarder56 you should use `demisto/python3:3.7.4.2728` and not `demisto/python3:latest`

Xboarder56:
@Itay4 this would work but then if a new container is released the package would need to be update. Also does that container always ship with demisto?

Itay4:
> @Itay4 this would work but then if a new container is released the package would need to be update. Also does that container always ship with demisto?

why would it need an update?
demisto pulls the requested docker image

Xboarder56:
> > @Itay4 this would work but then if a new container is released the package would need to be update. Also does that container always ship with demisto?
> 
> why would it need an update?
> demisto pulls the requested docker image

Potentially newer python features or fixes. I tagged it with 3.7.4.2728 instead of latest based on your input and the build passed the validation phase on CircleCI.

Itay4:
@Xboarder56 you are first declaring `matches = []` (a list) and then assigning `matches = dict(zip(dict_keys, matches))` (a dict) - what should it be? a list or a dict?

Xboarder56:
> @Xboarder56 you are first declaring `matches = []` (a list) and then assigning `matches = dict(zip(dict_keys, matches))` (a dict) - what should it be? a list or a dict?

Yeah depending on the path the script can be either or. I changed the way the output to demisto.results is handled to account for this. Additionally, error handling was added if the user chooses to specify an index that is out of range of the capture group.

anara123:
@kirbles19 can you go over the PR please
https://github.com/Xboarder56/content/blob/master/Scripts/RegexGroups/CHANGELOG.md
https://github.com/Xboarder56/content/blob/master/Scripts/RegexGroups/RegexGroups.yml

kirbles19:
@anara123 done.

glicht:
@anara123 I think we can merge?

Xboarder56:
Nope everything is good on my side, this can be merged. Thanks for the help guys!

PR #4632, opened 2019-10-20
Update type of values in last run dict

guyfreund:
## Status
Ready

## Description
Added important note of the value of the dict assigned to last run obj.

## Technical Writer files to review:
[README.md](https://github.com/demisto/content/blob/4b0f1091591c0d35100bc81cc37badbc39bedede/docs/fetching_incidents/README.md)

### DIFF
--- a/docs/fetching_incidents/README.md
+++ b/docs/fetching_incidents/README.md
@@ -84,6 +84,7 @@ When we are fetching incidents, it is important to include the ```rawJson``` key
 
 ### Setting Last Run
 When the last of the events have been retrieved, we need to save the new last run time into the integration context. This timestamp will be used the next time the ```fetch-incidents``` function is ran.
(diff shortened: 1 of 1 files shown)


docs/fetching_incidents/README.md
```diff
@@ -84,6 +84,7 @@ When we are fetching incidents, it is important to include the ```rawJson``` key
 
 ### Setting Last Run
 When the last of the events have been retrieved, we need to save the new last run time into the integration context. This timestamp will be used the next time the ```fetch-incidents``` function is ran.
+One more thing important to know when setting the last run object is that the values of the dictionairy must be of type string.
```
amshamah419:
```suggestion
Something important to note is that when setting the last run object, the values of the dictionary must be of the type `string`.
```

kirbles19:
@guyfreund Done.

PR #4633, opened 2019-10-21
ip indicator class - opinionated way to create indicators in demisto

anara123:
## Status
Done

## Related Issues
fixes: no issue right now- but I can open one later if we want to proceed with this PR

## Description
opinionated way to create indicators in Demisto integrations.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


glicht:
enum is python 3.4 and up. We probably can't use it in common.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1463,7 +1511,84 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+            ip_context['ASN'] = self.asn
+
+        if self.hostname:
+            ip_context['Hostname'] = self.hostname
+
+        if self.geo_latitude or self.geo_country or self.geo_description:
+            ip_context['Geo'] = {}
+
+            if self.geo_latitude and self.geo_longitude:
+                ip_context['Geo']['Location'] = f'{self.geo_latitude}:{self.geo_longitude}'
```
yaakovi: f-string can't be used in python2
```suggestion
                ip_context['Geo']['Location'] = '{}:{}'.format(self.geo_latitude, self.geo_longitude)
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1463,7 +1511,84 @@ def is_ip_valid(s, accept_v6_ips=False):
         return True
 
 
-def return_outputs(readable_output, outputs=None, raw_response=None):
+class DBotScoreEntry:
```
yaakovi: this is not an entry, main use is the context feature

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1463,7 +1511,84 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+    CONTEXT_PATH = 'DBotScore'
```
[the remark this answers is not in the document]
anara123: I actually want to remove the whole OutputPath class. Each indicator class will contain its own context_path. No need to keep it in separate class.

Scripts/CommonServerPython/CommonServerPython.py
anara123: I had naming collusion.
DBotScore class for the context or DBotScore for enum.
Or I can combine them together.
I would like to know your thoughts

ronykoz: Not sure we need this class, you only use the to_context section of it

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -12,6 +12,7 @@
 import re
 import base64
 import logging
+from enum import Enum
```
ronykoz: remove this as discused

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1463,7 +1511,84 @@ def is_ip_valid(s, accept_v6_ips=False):
         return True
 
 
-def return_outputs(readable_output, outputs=None, raw_response=None):
+class DBotScoreEntry:
+    CONTEXT_PATH = 'DBotScore'
+
+    def __init__(self, indicator, indicator_type: IndicatorType, vendor, score: DBotScore):
```
glicht: Add default to the args that are optional.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1463,7 +1511,84 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+            'Vendor': self.vendor,
+            'Score': self.score
+        }
+
+
+class IPIndicator:
+    CONTEXT_PATH = 'IP(val.Address && val.Address == obj.Address)'
+
+    def __init__(self, vendor, ip, asn, hostname, geo_latitude, geo_longitude, geo_country, geo_description,
+                 detection_engines, positive_engines, dbot_score: DBotScore, malicious_description):
```
glicht: add defaults to args that are optional

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -44,13 +47,33 @@
...
-    'plagroundError': 8,
     'playgroundError': 8,
     'entryInfoFile': 9,
     'warning': 11,
     'map': 15,
     'widget': 17
 }
+
+
+class EntryType:
```
jochman: make each class inherit from `(object)` for py2 compatibility

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -78,15 +114,21 @@
...
+    """
+    Enum: contains all the indicator types, used DBotScore.Type)
+    """
+    IP = 'ip'
+    HASH = 'hash'
+    DOMAIN = 'domain'
+    URL = 'url'
+
+    @staticmethod
+    def is_valid_type(_type):
```
jochman: Suggests the function below, doesn't require maintenance. also, we should check that the key exists and not the value.
```python
@classmethod
def is_valid_type(cls, _type):
    # type: (str) -> book
    types = [
        var
        for var in cls.__dict__.keys()
        if not var.startswith("__")
    ]
    return _type in types
```

if you still think it should be by value, replace top `var` with ```cls.__dict__[var]```

Scripts/CommonServerPython/CommonServerPython.py
jochman: Inherit from `object` to maintain py2 compatibility.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1492,6 +1534,205 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+    """
+    def __init__(self):
+        pass
+
+    def to_context(self):
+        pass
+
+class CommandResults:
+    def __init__(self, output_prefix, uniq_field, outputs, indicators=None):
+        # type: (str, str, Any, List[Indicator]) -> None
```
jochman: Add docstring to clarify the class use cases and inputs

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1492,6 +1534,205 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+        if self.dbot_score:
+            ret_value.update(self.dbot_score.to_context())
+
+        return ret_value
+
+
+class Domain(Indicator):
+    CONTEXT_PATH = 'Domain(val.Name && val.Name == obj.Name)'
+
+    def __init__(self, domain, dns):
```
jochman: add docstring

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1492,6 +1534,205 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+        self.domain = domain
+        self.dns = dns
+
+        # DBotScore fields
+        self.dbot_score = None
+
+    def set_dbot_score(self, dbot_score):
```
jochman: can use `@property` annotator instead of creating getter and setter

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1492,6 +1534,205 @@ def is_ip_valid(s, accept_v6_ips=False):
         return True
 
 
+class Indicator:
+    """
+    interface class
+    """
+    def __init__(self):
```
jochman: don't need this. just inherit from `object`

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1492,6 +1534,205 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+        pass
+
+    def to_context(self):
```
jochman: add `@abstractmethod` annotator.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -78,15 +114,21 @@
...
-    'Unknown': 0,
-    'Informational': 0.5
-}
+
+
+class IndicatorType:
+    """
+    Enum: contains all the indicator types, used DBotScore.Type)
+    """
+    IP = 'ip'
```
jochman: suggests merging my branch here https://github.com/demisto/content/pull/5368
Already added more indicators type (that needed for TIP integrations).
You should preserve the name of this class as its singular.

ronykoz:
@anara123 where are we on this?

anara123:
still in progress, didn't have time to work on it.
I hope to continue to work on it next week

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -79,15 +128,28 @@
...
-    'Informational': 0.5
-}
+
+
+class DBotScoreType(object):
+    """ ignore docstring
+    Enum: contains all the indicator types, used DBotScore.Type)
+    """
+    IP = 'ip'
+    HASH = 'hash'
```
ronykoz: Pretty sure that we should mark it as a file rather then hash
```suggestion
    FILE = 'file'
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -45,13 +48,33 @@
     'pinned': 5,
     'userManagement': 6,
     'image': 7,
-    'plagroundError': 8,
```
ronykoz: We should keep it there if we are deprecating it.
The new version will be without it

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -79,15 +128,28 @@
...
-    'High': 3,
-    'Medium': 2,
-    'Low': 1,
-    'Unknown': 0,
-    'Informational': 0.5
-}
+
+
+class DBotScoreType(object):
+    """ ignore docstring
```
ronykoz: ignore docsting?

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -79,15 +128,28 @@
     'vtPositives': 10,
     'vtPositiveUrlsForIP': 30
 }
-# The dictionary below does not represent DBot Scores correctly, and should not be used
-dbotscores = {
```
ronykoz: Why did we remove this?

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1770,14 +2399,13 @@ def replace_str(src_str):
...
+ipv6cidrRegex = r'\b(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))(\/(12[0-8]|1[0-1][0-9]|[1-9][0-9]|[0-9]))\b'  # noqa
 emailRegex = r'\b[^@]+@[^@]+\.[^@]+\b'
 hashRegex = r'\b[0-9a-fA-F]+\b'
 urlRegex = r'(?:(?:https?|ftp|hxxps?):\/\/|www\[?\.\]?|ftp\[?\.\]?)(?:[-\w\d]+\[?\.\]?)+[-\w\d]+(?::\d+)?' \
            r'(?:(?:\/|\?)[-\w\d+&@#\/%=~_$?!\-:,.\(\);]*[\w\d+&@#\/%=~_$\(\);])?'
-cveRegex = r'(?i)^cve-\d{4}-([1-9]\d{4,}|\d{4})$'
```
ronykoz: why are we removing it?

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1583,6 +1645,572 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+        return return_entry
+
+
+class DBotScore(Indicator):
+    NONE = 0
+    GOOD = 1
+    SUSPICIOUS = 2
+    BAD = 3
+
+    CONTEXT_PATH = 'DBotScore(val.Indicator && val.Indicator == obj.Indicator && val.Vendor == obj.Vendor)'
```
ronykoz: i think that this path is ok only from 5.5

Scripts/CommonServerPython/CommonServerPython.py
anara123: I think this was auto merge problem

Scripts/CommonServerPython/CommonServerPython.py
anara123: it is not used anywhere, so I deleted it

Scripts/CommonServerPython/CommonServerPython.py
anara123: I added in the common_server_docs.py condition that if the docstring contains `ignore docstring` it will ignore generating docs for it.
Till now if user wanted to ignore/skip function from being in Script Helper documentation then he had to add it under `pyPrivateFuncs`.

Scripts/CommonServerPython/CommonServerPython.py
anara123: It is good from 5.0, no?

ronykoz: @hirshbergshachar on which version did you add the support of this? 
And @anara123 we should have two versions then, one for 5.0 and one for 5.5

Scripts/CommonServerPython/CommonServerPython.py
ronykoz: Lets return it, it seems to be located in master

Scripts/CommonServerPython/CommonServerPython.py
hirshbergshachar: @ronykoz fix was merged only to 5.5 as far as I can see

ronykoz: @anara123 So we should address this in the code.
Let's use the get_demisto_version function to check that we are >= 5.5 for this CONST
And otherwise, we should use the one without DT

Scripts/CommonServerPython/CommonServerPython.py
anara123: I deleted all the ignore docstring, instead I will generate docstring for classes too

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1694,6 +1768,779 @@ def is_ip_valid(s, accept_v6_ips=False):
...
+class CommandResults:
+    """
+    CommandResults class - use to return results to warroom
+
+    :type outputs_prefix: ``str``
+    :param outputs_prefix: should be identical to the prefix in the yml contextPath in yml file. for example:
+            CortexXDR.Incident
+
+    :type outputs_key_field: ``str``
+    :param outputs_key_field: primary key field in the main object
```
ronykoz: I think that we should mention that it is used for the DT, because the current explanation is a bit vague

anara123:
@kirbles19 https://github.com/demisto/content/blob/indicator-classes/Packs/Base/Scripts/CommonServerPython/CHANGELOG.md

PR #4634, opened 2019-10-21
pytest parallelization

Itay4:
## Status
Ready

## Description
Added the option to run pytest unit tests in parallel using pytest-xdist

### DIFF
--- a/Integrations/ImageOCR/CHANGELOG.md
+++ b/Integrations/ImageOCR/CHANGELOG.md
@@ -1,0 +1,2 @@
+## [Unreleased]
+-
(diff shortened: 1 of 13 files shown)


glicht:
I am not sure this increases performance on circle (as we already run the full validation done in docker in parallel between tests). Also, from running on my local laptop didn't see a difference with 4 cpu or 1 cpu for CommonServerPython and ParseEmailFiles. 

Linting is failing as a result of upgrade to pylint. It looks like they've introduced a bug in check_output validation. I've reported an issue at: https://github.com/PyCQA/astroid/issues/715

Itay4:
@glicht yea it needs a big test suite for it to actual improve performance, just thought to add support with `0` value which actually doesn't do anything for now, but didnt consider circle
prefer to close this for now ?

PR #4635, opened 2019-10-22
Fixed array handling in Splunk notable to incident

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19572

## Description
Splunk was producing an error  when the rule name was an array.
This case in now handled correctly.

   - No

## Must have
- [ ] Code Review


kirbles19:
@guykeller done

PR #4636, opened 2019-10-22
Tenable.io - packegify and raw response fix

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19207

## Description
no instance, validated with customer - please see referenced issue

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Technical writer review
- [x] [YAML file](link) - no change
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/13a831d7bc1176e927bdd0e7d40c5c3991a80ac5/Integrations/Tenable_io/CHANGELOG.md)


Integrations/Tenable_io/Tenable_io.py
```diff
@@ -129,7 +129,7 @@ def intersection(lst1, lst2):
 
     return {
         'Type': entryTypes['note'],
-        'Contents': context_obj,
+        'Contents': obj,
```
Itay4: lets return here the object that is not null filtered

kirbles19:
@bakatzir Done.

Integrations/Tenable_io/Tenable_io.py
bakatzir: @Itay4 changed accordingly in the begging of the function.

PR #4651, opened 2019-10-23
Support appending to existing rule and moving rules in DAG & EDL playbooks

Arsenikr:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/19552

## Description
At the moment, the DAG playbook creates a block rule if the given tag does not exist. this rule is added to the bottom of the ruleset by default, and this can make it obsolete in cases that another rule above it overrides it.

after consulting with Francesco from PANW, this addition to the flow was suggested:

Provide an existing rule name as a parameter and then add the DAG as the destination to it.

If the above is not provided, a new policy is added to the top or to bottom (or before/after a certain rule) according to the customer's choice but the customer needs to keep in mind that adding to the bottom might not work depending on the ruleset.

Do the same with EDL setup playbook as well

   - No

## Must have
- [ x] Tests
- [ x] Documentation (with link to it)
- [ x] Code Review

## Technical writer review
- [x] [YAML file](link)
- [x] [CHANGELOG](link)


Arsenikr:
@michalgold please review
@kirbles19 please go over the descriptions etc..

yaron-libman:
Updated texts and descriptions.

PR #4657, opened 2019-10-23
Rasterize wait

orenzohar:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19688

## Description
new parameter to wait a set number of seconds before taking a screenshot

5.0.0

   - No

## Must have
- [x] Tests
- [ ] Code Review


Itay4:
why did you set `docs-approved` ?

Itay4:
please add test

Itay4:
i dont think this is a good solution, as the user will probably have no idea how long it takes to a page to load
we should try to understand how to use [waits](https://selenium-python.readthedocs.io/waits.html) better

orenzohar:
@Itay4 Implicit wait was already implemented but doesn't cover all cases, such as the one from the issue. As for the lack of tests, we're not testing URLs in rasterize, and this argument is only relevant for rasterizing a URL.

Integrations/rasterize/rasterize.py
```diff
@@ -66,22 +67,23 @@ def init_driver(offline_mode=False):
...
+    :param wait_time: time in seconds to wait before taking a screenshot
     """
     driver = init_driver(offline_mode)
 
     try:
         demisto.debug(f'Navigating to path. Mode: {"OFFLINE" if offline_mode else "ONLINE"}')
 
         driver.get(path)
-        driver.implicitly_wait(5)
+        time.sleep(wait_time) if wait_time > 0 else driver.implicitly_wait(5)  # type: ignore
```
glicht: I think the sleep should be done after the `implicitly_wait`. So we first make sure the page loaded and then if needed add another sleep to allow the page to fully render (Javascript stuff).

orenzohar:
@Itay4 it will require too many changes in the image (installing tika pkg and java), as we said - we need to do some re-thinking about this specific case.

glicht:
Do a playbook test and verify that the generated image contains the expected text with the ImageOCR integration.

PR #4669, opened 2019-10-27
Convert Build Process to use Demisto-Py V2

amshamah419:
## Status
Ready pending review

## Related Issues
fixes: demisto/etc#19954

## Description
Demisto-Py v1 is deprecated and build process should use v2

x.x.x

   - Ideally not


Integrations/dnstwist/dnstwist.yml
```diff
@@ -63,3 +63,4 @@ script:
   runonce: false
   script: '-'
   type: python
+  subtype: python2
```
amshamah419: I would like to just revert this since DNSTwist is unrelated here.

Tests/test_integration.py
```diff
@@ -320,7 +375,8 @@ def test_integration(client, integrations, playbook_id, options=None, is_mock_ru
         print('Create integration %s succeed' % (integration_name,))
 
     # create incident with playbook
-    incident, inc_id = __create_incident_with_playbook(client, 'inc_%s' % (playbook_id,), playbook_id, integrations)
+    incident, inc_id = __create_incident_with_playbook(client, 'inc_%s' % (playbook_id,),
```
yaakovi: use `.format`

Tests/mock_server.py
```diff
@@ -196,7 +201,8 @@ def configure_proxy_in_demisto(self, proxy=''):
                 },
             'version': -1
         }
-        return self.demisto_client.req('POST', '/system/config', data)
+        return demisto_client.generic_request_func(self=self.client, path='/system/config',
```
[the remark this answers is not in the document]
amshamah419: This is more involved than a quick fix. It'll require a design change in the `configure()` function of demisto-py. It's a bit out of scope, but I'll open an issue with it for demisto-py

Integrations/dnstwist/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+- Added subtype to integration yaml.
```
glicht: If this is the only change you can leave the release notes empty. Simply put `-`

amshamah419:
@glicht Made the above changes and build passes. When should we plan to push this?

glicht:
Let's push after the branch of the release

glicht:
Still see demisto-py installed in: https://github.com/demisto/content/blob/api-testing-branch/.circleci/build-requirements.txt . Please remove.

amshamah419:
> Still see demisto-py installed in: https://github.com/demisto/content/blob/api-testing-branch/.circleci/build-requirements.txt . Please remove.

Just caught that. It's now removed. Was looking for demisto-py==2.0.6 originally.

.circleci/build-requirements.txt
```diff
@@ -7,5 +7,4 @@ rsa==3.4.2
 paramiko==2.6.0
 boto3==1.10.0
 pyminizip===0.2.4
-git+https://github.com/demisto/demisto-py.git@v1.0.1 # disable-secrets-detection
```
yaakovi: are you sure we don't need to add the new `demisto-py` here instead?

glicht: we have it in `dev-requirements*.txt` file. I think it is ok.

PR #4701, opened 2019-10-28
ARIAPacketIntelligence renaming

Itay4:
## Status
Ready

## Description
Renamed file names to be same as directory
   - No


yaakovi:
new integration, no need to add CHANGELOG

PR #4711, opened 2019-10-29
packgeify ZipFile

roysagi:
## Status
Ready

## Description
packaged ZipFile script


roysagi:
added 'docs approved' because I just packaged the script with no changes to the descriptions

PR #4713, opened 2019-10-29
TruSTAR - no license

dantavori:
## Status
Ready

## Related Issues
fixes: [https://github.com/demisto/etc/issues/19756](https://github.com/demisto/etc/issues/19756)

## Description
Added issue number in conf.json

   - No


dantavori:
@ShahafBenYakir

PR #4696, opened 2019-10-28
PAN-OS EDL enhancements

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19753

## Does it break backward compatibility?
   - No

## Must have
- [x] Tests
- [ ] Documentation - soon
- [x] Code Review

## Technical writer review
Mention and link to the files that require a technical writer review.
- [ ] [YAML file](https://github.com/demisto/content/blob/ca9200efb9b6bf8e05fdd108ada5ced81922bd1b/Integrations/PaloAltoNetworks_PAN_OS_EDL_Management/PaloAltoNetworks_PAN_OS_EDL_Management.yml)
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/ca9200efb9b6bf8e05fdd108ada5ced81922bd1b/Integrations/PaloAltoNetworks_PAN_OS_EDL_Management/CHANGELOG.md)

PR #4746, opened 2019-11-02
Update CHANGELOG.md

spearmin10:


bakatzir:
@spearmin10 cahgnelog changes needs to be in the same PR.
see this for reference: https://github.com/demisto/content/pull/4696

PR #4753, opened 2019-11-03
Circle build: use virtual env

glicht:
## Status
Ready

## Description
Move to use virtualenv for our build. Docker image is a minimal one based on circle's docker without the python deps. Python deps are managed in a virtual env and cached by circle. If there are no changes to dependencies loading the cache and setting up the venv takes 7 seconds.

Dependencies which are needed only for the build are managed at: `.circleci/build-requirements.txt`

## Additional changes
Mypy fixes for the integrations that were failing mypy after the mypy upgrade.


dev-requirements-py2.txt
```diff
@@ -4,5 +4,7 @@ PyYAML==5.1.2
 pykwalify==1.7.0
 demisto-py==2.0.5
 pytest==4.6.5
+pytest-mock==1.11.1
 beautifulsoup4==4.8.1
-pypdf2==1.26.0
\ No newline at end of file
+pypdf2==1.26.0
+git+https://github.com/coyo8/parinx.git@6493798ceba8089345d970f71be4a896eb6b081d
```
yaakovi: why don't we need this also in py3?

glicht: It is used by docs script only which is py2. Not sure this package even supports py3.

yaakovi:
from now on, we should avoid changing the build docker?

glicht:
Correct. I added you to the docker or.

PR #4755, opened 2019-11-03
Check if image is not default image - Hooks

guyfreund:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18704

## Description
Added hook test to check if the image is not the default image

## Must have
- [x] Tests
- [x] Code Review


Tests/scripts/hook_validations/image.py
```diff
@@ -86,3 +90,45 @@ def is_existing_image(self):
             return False
 
         return True
+
+    def is_not_default_image(self):
+        """Check if the image is the default one"""
+        if re.match(IMAGE_REGEX, self.file_path, re.IGNORECASE):
```
Itay4: add comment what this condition handles

Tests/scripts/hook_validations/image.py
```diff
@@ -86,3 +90,45 @@ def is_existing_image(self):
...
+                    encoded_image = encoded_image.decode("utf-8")
+                if encoded_image == DEFAULT_IMAGE_BASE64:  # disable-secrets-detection
+                    print_error("{} is the default image, please change to the "
+                                "integration image.".format(self.file_path))
+                    self._is_valid = False
+                    return False
+                else:
+                    return True
+
+        else:
```
Itay4: and add comment what this condition handles

Tests/scripts/hook_validations/image.py
guyfreund: fixed

Tests/scripts/hook_validations/image.py
guyfreund: fixed

PR #4773, opened 2019-11-04
Cortex new

teizenman:
### Status
Ready

## Related Issues
fixes: demisto/etc#19634

## Description
- Since there are Raise Exception in the query_loggings function, demisto.results('ok') will be reached if no errors caught before.
- Last branch has been ruined

### DIFF
--- a/Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
+++ b/Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
@@ -1119,8 +1119,6 @@ def convert_log_to_incident(log):
 
 
 def test_module():
-    global FETCH_QUERY
-
     if demisto.params().get('isFetch'):
         last_fetched_event_timestamp, _ = parse_date_range(FIRST_FETCH_TIMESTAMP)
     test_args = {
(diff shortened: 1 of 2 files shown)


Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
```diff
@@ -1117,6 +1118,18 @@ def convert_log_to_incident(log):
 ''' COMMANDS FUNCTIONS '''
 
 
+def test_module():
+    if demisto.params().get('isFetch'):
+        last_fetched_event_timestamp, _ = parse_date_range(FIRST_FETCH_TIMESTAMP)
+    test_args = {
+        "query": "SELECT * FROM panw.threat LIMIT 1",
```
Itay4: better to query the table that was chosen for the fetch

Integrations/PaloAltoNetworksCortex/PaloAltoNetworksCortex.py
```diff
@@ -1117,6 +1118,20 @@ def convert_log_to_incident(log):
 ''' COMMANDS FUNCTIONS '''
 
 
+def test_module():
+    global FETCH_QUERY
+
+    if demisto.params().get('isFetch'):
+        last_fetched_event_timestamp, _ = parse_date_range(FIRST_FETCH_TIMESTAMP)
+    test_args = {
+        "query": FETCH_QUERY_DICT[FETCH_QUERY] + 'LIMIT 1',
```
Itay4: use f-strings and don't use `"` but `'`

PR #4776, opened 2019-11-04
Added new script TransformerTemplateProcessor.

tundisto:
Added new script TransformerTemplateProcessor and unit test.

## Status
Ready

## Related Issues
N/A

## Description
Allows simple one-argument value= transformers to be run on a string using template syntax {{my_context_var | my_transformer}}, allowing many transformers to be run from a single command, effectively turning a string template into a compiled string.

## Screenshots
N/A

## Related PRs
N/A

4.x

   - No

## Must have
- [X] Tests
- [ ] Code Review

## Dependencies
N/A

## Additional changes
N/A

## Technical writer review
- [ ] [YAML file](https://github.com/demisto/content/blob/0e6b0bdb0ae7aef886d2738e899b930b827b972f/Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.yml)


Itay4:
please review the [YAML](https://github.com/demisto/content/blob/0e6b0bdb0ae7aef886d2738e899b930b827b972f/Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.yml)

tundisto:
Please don’t merge today (though I doubt there’s much danger of that :) ). I’m going to make a change to the script to add the ability to output to a context key.

kirbles19:
@Itay4 Done.

tundisto:
The 'key' parameter has been added.  All ready for your review, @Itay4 .

Itay4:
not sure why we need this one, why not just apply multiple transformers?

tundisto:
Why?  Because the native transformer UX for building large customised strings can be quite cumbersome.  Need to insert a transformer or re-order the flow?  Be prepared to delete every transformer and start all over again.

Templates are part of many products, because they’re quite useable — the {{var | command}} syntax came from Angular.  

I wrote this to build a complex HTML email body for a training scenario, which required many transformers to be run on variables, such as converting many timer fields to human readable format. The body required many changes before it was right. This allowed me to simply change a string.  It would have been a small nightmare to do this if I had to use the native transformers UX, requiring either potentially many tasks to make it manageable, or having to delete and re-add transformers many times.

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
```diff
@@ -0,0 +1,83 @@
...
+            # no transformer was found
+            transformed_value = context_lookup(context_key)
+
+        res_body = res_body.replace(m.group(0), transformed_value, 1)  # insert result into template
+
+    # return completed template
+    if not key:
+        demisto.results(res_body)
+    else:
+        demisto.results({
```
Itay4: use `return_outputs` instead

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
```diff
@@ -0,0 +1,83 @@
...
+        return res
+
+    try:
+        return demisto.get(context, key)
+    except Exception:
+        return None
+
+
+def main(value, key=None):
+    BODY = value
```
Itay4: use lower case for non const vars

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
```diff
@@ -0,0 +1,83 @@
...
+            if not context_value:
+                transformed_value = 'None'
+
+        elif context_key and transformer:
+            # found a key and a transformer
+            context_value = context_lookup(context_key)
+
+            if context_value:
+                # now run the transformer
+                transformed_value = demisto.executeCommand(transformer, {'value': str(context_value)})[0]['Contents']
```
Itay4: use `is_error` to check if response is ok

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
```diff
@@ -0,0 +1,83 @@
...
+            'Type': entryTypes['note'],
+            'ContentsFormat': formats['text'],
+            'Contents': res_body,
+            'EntryContext': {
+                key: res_body
+            }
+        })
+
+
+if __name__ == "__builtin__" or __name__ == "builtins":
```
Itay4:
```suggestion
if __name__ in ('__builtin__', 'builtins'):
```

yaakovi:
@tundisto @Itay4 
what is the status here?

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
tundisto: Changed in latest commit.

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
tundisto: Changed in latest commit.

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
tundisto: Changed in latest commit.

Scripts/TransformerTemplateProcessor/TransformerTemplateProcessor.py
tundisto: Changed in latest commit.

tundisto:
@yaakovi @Itay4   Sorry for the long delay, gents.  Requested changes have been made, back to you.

PR #4787, opened 2019-11-05
Archer - fixed fetch incidents offset bug problem

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18463

## Description
There were several issues with Archer fetch incidents, causing it to fail on some edge cases.
Those should be fixed now.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/integration-ArcherRSA.yml
+++ b/Integrations/integration-ArcherRSA.yml
@@ -2358,7 +2358,6 @@ script:
         // This function verifies the fields specified in the fieldsToValue argument are known by Archer to be compatible with the given app ID.
         var userSpecifiedFields = Object.keys(fieldsToValueJSON);
         var actualFieldsOfGivenApp = Object.keys(appFieldMapping);
-        logInfo("\n\n\n" + "type of actualFieldsOfGivenApp: " + JSON.stringify(actualFieldsOfGivenApp) + "\n\n\n");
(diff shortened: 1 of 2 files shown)


hirshbergshachar:
Don’t we add an int to a time object?

guykeller:
There are several possible cases:
In all cases the offset is an int, as it is only defined once using getTime().
1. This is our first run. In which case, the lastTime is also defined using getTime(), thus is an int.
2. This is not our first run, but no incidents returned. In which case, the lastTime is taken from lastRun, and inductively we assume that it is correct.
3. This is not our first run, some incidents returned and are in European time - this is the case in our own instance, in which case I can guarantee that this is a dateTime object and thus used getTime() - which means it is an int.
4. The only case I'm not sure about is the one using non-European time (namely, American).
Thus in this issue (https://github.com/demisto/etc/issues/18463) I've asked Eden to send the fix to the customer for testing before merging it, and have now added the inprogress label.
However, if you know from experience the American time is also successfully translated to a date by createIncidentsArr, I'll add a getTime() there as well - just let me know 🙂 
@hirshbergshachar

kirbles19:
@guykeller Done.

guykeller:
Closing this since I combined two confusing PRs to one PR, since it is the same function.

PR #4790, opened 2019-11-05
fix - xdr-update-incident should not execute with empty arguments

anara123:
## Status
Ready

## Description
fix - xdr-update-incident should not execute with empty arguments

   - No

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/xdrsync-fix/Scripts/XDRSyncScript/CHANGELOG.md)


anara123:
@kirbles19 please review

kirbles19:
@anara123 Done.

PR #4791, opened 2019-11-06
Archer - fixed fetch incidents details bug

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19592

## Description
Until now, when several incidents were fetched on the same minutes, the details of some incidents were concatenated with the details of others.
This issue is now fixed.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/integration-ArcherRSA.yml
+++ b/Integrations/integration-ArcherRSA.yml
@@ -2079,6 +2079,7 @@ script:
 
         if (records && records.Contents && records.Contents.length > 0) {
             for (var i = 0; i < records.Contents.length; i++) {
+                details = "";
                 var res = records.Contents[i];
                 var labels = [];
                 var fields = Object.keys(res.Fields);
--- a/Integrations/integration-ArcherRSA.yml
+++ b/Integrations/integration-ArcherRSA.yml
@@ -2358,7 +2358,6 @@ script:
         // This function verifies the fields specified in the fieldsToValue argument are known by Archer to be compatible with the given app ID.
         var userSpecifiedFields = Object.keys(fieldsToValueJSON);
         var actualFieldsOfGivenApp = Object.keys(appFieldMapping);
-        logInfo("\n\n\n" + "type of actualFieldsOfGivenApp: " + JSON.stringify(actualFieldsOfGivenApp) + "\n\n\n");
         userSpecifiedFields.forEach(function(fieldSpecifiedByUser) {
             if (!(actualFieldsOfGivenApp.indexOf(fieldSpecifiedByUser) > -1)) {
                 var wrongFieldErrorMessage = "Error: you've specified the following field in the arg fieldsToValues: {0}, which does not exist for records of the app ID: {1}. \n".format(fieldSpecifiedByUser, appID);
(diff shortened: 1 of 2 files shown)


guykeller:
@hirshbergshachar @ronykoz 
Quick note - these files already include the previous fix, please ignore those.  
If you prefer, it might be easier to review this after the other fix is merged to master.

Integrations/integration-ArcherRSA.yml
```diff
@@ -2920,14 +2920,17 @@ script:
         }
         var offset = tz ? (60 * 1000 * new Date().getTimezoneOffset()) : 0;
         offset = tz ? parseInt(tz) : 0;
```
ronykoz: Don't think we need those two lines

Integrations/integration-ArcherRSA.yml
```diff
@@ -2944,10 +2947,15 @@ script:
         var incidents = createIncidentsArr(results, dataArgs.applicationId);
 
         if (incidents.length > 0){
+            shouldAddOffset = true;
             lastTime = incidents[incidents.length-1].occurred;
-        }
 
-        setLastRun({ lastTime: lastTime + offset});
+            if (params.useEuropeanTime) {
```
ronykoz: why do we need this condition? seems to be always true

Integrations/integration-ArcherRSA.yml
```diff
@@ -2920,14 +2920,17 @@ script:
         }
         var offset = tz ? (60 * 1000 * new Date().getTimezoneOffset()) : 0;
         offset = tz ? parseInt(tz) : 0;
+        offset = tz ? 60 * 1000 * parseInt(tz) : 60 * 1000 * new Date().getTimezoneOffset();
```
ronykoz: why do we need *1000 ?

Integrations/integration-ArcherRSA.yml
```diff
@@ -2944,10 +2947,15 @@ script:
...
+            shouldAddOffset = true;
             lastTime = incidents[incidents.length-1].occurred;
-        }
 
-        setLastRun({ lastTime: lastTime + offset});
+            if (params.useEuropeanTime) {
+                lastTime = lastTime.getTime()
+            }
+        }
+        var newLastTime = shouldAddOffset ? lastTime + offset : lastTime;
```
ronykoz: not sure i understood in which cases you will want to add the offset and why

guykeller: When you send the time from which you want to start fetching to Archer, you need to do adjustments so Archer would know the right time to start from (as there are differences between the time of your Archer instance and your local time).
However, we should only add the offset in two cases:
1. If Archer returns incidents, the time it returns is the time the incidents were created in ARCHER's time. Which means - you need to add the offset to get to the actual time they were created from your perspective.
2. If this is the first run you should add the offset, since it hasn't been added before

However, we shouldn't add the offset if there were no incidents fetched but this isn't our first run - since the offset is already included by induction.

Integrations/integration-ArcherRSA.yml
guykeller: It's in MS

Integrations/integration-ArcherRSA.yml
guykeller: I agree, thought it would be better to remove them in a rewrite

Integrations/integration-ArcherRSA.yml
guykeller: Because we're not sure yet if the createIncidentsArr function works perfectly for American time.
Sent a version without this condition to be tested by KPMG, if it works for them I'll update this version

guykeller:
Closing this since I combined two confusing PRs to one PR, since it is the same function.

PR #4800, opened 2019-11-06
Archer - fix details concat bug + offset bug

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19592
fixes: https://github.com/demisto/etc/issues/18463

## Description
Fixing two fetch incidents bugs:
1. Details of different incidents were  concatenated, now they aren't.Solution tested by customer and works.
2. Fetch was unstable and would sometimes fail due to a problem with the offset. Solution tested by customer and works.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/integration-ArcherRSA.yml
+++ b/Integrations/integration-ArcherRSA.yml
@@ -2081,6 +2081,7 @@ script:
 
         if (records && records.Contents && records.Contents.length > 0) {
             for (var i = 0; i < records.Contents.length; i++) {
+                details = "";
                 logDebug("Number of records is: {0}".format(records.Contents.length));
                 var res = records.Contents[i];
                 logDebug("Current value of res is: {0}".format(JSON.stringify(res)));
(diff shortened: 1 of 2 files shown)


guykeller:
@kirbles19 FYI - just the second point of the CHANGELOG was not approved by you yet.

Integrations/integration-ArcherRSA.yml
```diff
@@ -2948,16 +2948,18 @@ script:
         if (params.timeZone === '0') {
             tz = 0;
         }
-        var offset = tz ? (60 * 1000 * new Date().getTimezoneOffset()) : 0;
-        offset = tz ? parseInt(tz) : 0;
+
+        offset = tz ? 60 * 1000 * parseInt(tz) : 60 * 1000 * new Date().getTimezoneOffset();
```
ronykoz: add a comment explaining the 60 * 1000

Integrations/integration-ArcherRSA.yml
```diff
@@ -2081,6 +2081,7 @@ script:
 
         if (records && records.Contents && records.Contents.length > 0) {
             for (var i = 0; i < records.Contents.length; i++) {
+                details = "";
```
hirshbergshachar: @guykeller is this used for anything? If not, please remove

kirbles19:
@guykeller Updated all RNs. Done.

Integrations/integration-ArcherRSA.yml
guykeller: This is what fixes the concatenation issue.
As you can see, the "details" variable is declared outside the scope of the for loop, thus isn't re-written for each record.
This fixes it.

PR #4812, opened 2019-11-07
Pylint enable string checker

glicht:
## Status
Ready

## Description
Until now we were running pylint with errors only. This pr adds also warnings from the string checker except the `duplicate-string-formatting-argument` which seems to be extreme . See: https://docs.pylint.org/en/1.6.0/features.html#string-checker

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Additional changes
Fix the pylint errors in existing integrations.

## Technical writer review
No need.


glicht:
@yaakovi see Polyswarm validation failure in build: https://circleci.com/gh/demisto/content/33018?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Integrations/Whois/Whois.py
```diff
@@ -8285,7 +8285,7 @@ def whois_command():
                 '%d-%m-%Y')
             md['Expiration Date'] = whois_result.get('expiration_date')[0].strftime('%d-%m-%Y')
     except ValueError as e:
-        return_error('Date could not be parsed. Please check the date again.\n{error}'.format(type(e)))
+        return_error('Date could not be parsed. Please check the date again.\n{}'.format(e))
```
Itay4: i think this might cause issues cause of unicode
better to format as `str(e)`

glicht: It's the same. `format` internally does the same as `str` to the passed object.

glicht:
Not sure about pylint score. It will be 10, because we disable all non relevant stuff and enforce to fix the errors.

PR #4830, opened 2019-11-10
Fix cloaken

ShahafBenYakir:
## Status
Ready

## Description
Removed unstable verification


TestPlaybooks/playbook-Cloaken-Test.yml
```diff
@@ -230,53 +230,6 @@ tasks:
     note: false
     timertriggers: []
     ignoreworker: false
-  "9":
```
yaakovi: line 218:
switch to "10" - there is no task 9 anymore

PR #4834, opened 2019-11-10
added -sn flag support

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20003

## Description
Fixed bug where -sn flag would raise an error

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/nmap_bug/Integrations/Nmap/CHANGELOG.md)


roysagi:
> Not 100% sure how does it solve the -sn thing, could you explain please?

Sure. 
'-sn' flag basically disables the port scan done in nmap command.

This resulted is that the 'scan_type' parameter to not exist in the response and attempting to get it is what threw the KeyError indicated in the bug description.

In the fix I try getting the scan_type but if it fails on key error I catch it and continue.

Because of the way the response returns I couldn't use "get" method nor a "get"-like method (https://libnmap.readthedocs.io/en/latest/objects/nmapreport.html) - so this try and except was the best way I figured how to do it.

PR #4837, opened 2019-11-10
unskip Intezer Testing

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19853

## Description
Remove Intezer Testing from skipped


Integrations/integration-Intezer.yml
```diff
@@ -5,7 +5,7 @@ name: Intezer
...
-description: Malware detection and analysis based on code reuse
+description: Malware detection and analysis based on code reuse.
```
ShahafBenYakir: Why is this?

Integrations/integration-Intezer_CHANGELOG.md
```diff
@@ -0,0 +1 @@
+unskip Intezer Testing
```
ShahafBenYakir: This is not something we need to share in content release notes, you can erase.

Integrations/integration-Intezer.yml
adi88d: it should be in the end of the description

PR #4841, opened 2019-11-10
Fixed display name

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19943

## Description
Display names weren't clear enough and tended to be confusing.
That is now fixed.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


Integrations/integration-ArcherRSA.yml
```diff
@@ -1926,10 +1926,21 @@ script:
             var user = usersField['User'];
             if (Array.isArray(usersField['User'])){
                 usersField['User'].forEach(function (user){
-                    users += user['-firstName'] + " , " + user['-lastName'] + " " + user['-middleName'];
+                    users += user['-firstName'] + " " + user['-middleName'] + " " + user['-lastName'];
```
ronykoz: why do we need this line then?

PR #4851, opened 2019-11-11
Modified script to handle zero as value and not as None

teizenman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18798

## Description
Before the change, there was no handling to `0` as a value, but it was considered as `None`


Scripts/PositiveDetectionsVSDetectionEngines/CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 ## [Unreleased]
 - Added PositiveDetectionsVSDetectionEngines automation. Tagged as dynamic-indicator-section. Shows a bar chart of the number of Positive Detections out of overall detections.
+- Fixed an issue for zero-values treated as None.
```
ronykoz: i think that we might want a better msg here

Scripts/PositiveDetectionsVSDetectionEngines/Test_PositiveDetectionsVSDetectionEngines.py
```diff
@@ -0,0 +1,80 @@
+from __future__ import print_function
+import pytest
+from PositiveDetectionsVSDetectionEngines import extract_engines_data_from_indicator
+
+'''INDICATORS DATA FOR TESTING'''
+
+detect_positive_not_zero_1 = {'CustomFields': {'detectionengines': 71, 'positivedetections': 10},
```
ronykoz: make all the dictionaries multiline

teizenman:
@kirbles19 Can you please verify RN at: https://github.com/demisto/content/blob/URLIndicatorCommScore/Scripts/PositiveDetectionsVSDetectionEngines/CHANGELOG.md

PR #4854, opened 2019-11-11
CB Enterprise - correcting bug in non decompress get-binary command

roysagi:
## Status
Ready

## Description
Correcting bug in get-binary non decompress flow

   - No


roysagi:
No changes in descriptions nor parameters so added docs-approved

orenzohar:
why empty string? we need to ask someone who knows what if the purpose of `Contents:` and make sure it's the right use.

roysagi:
Considering fileResults function - it seems Contents should return an empty string:
return {'Contents': '', 'ContentsFormat': formats['text'], 'Type': file_type, 'File': filename, 'FileID': temp}

PR #4857, opened 2019-11-11
add fromversion to test playbook

anara123:
## Status
Ready

## Description
fixes the build, the integration is from version 4.5.0 but the test has no fromversion, that is why the build fails on server with version prior than 4.5.0

## Related PRs

   - No


glicht:
I think it needs to go in the conf json as we do here: https://github.com/demisto/content/blob/master/Tests/conf.json#L11

anara123:
there is a problem with SecurityAdvisor instance. from time to time it returns Internal Error 500. Not related to the fromversion problem. merging the PR

PR #4861, opened 2019-11-12
Phishing - Core - playbok + test fix

idovandijk:
## Status
Ready

## Related Issues
- fixes: https://github.com/demisto/etc/issues/20087
- fixes test failing issue as well (no issue opened)

## Description
- The test was failing because the scheduled command that should automatically complete the manual task in the `Phishing - Core` playbook, was referring to the wrong playbook ID (most likely due to changes in the test). I fixed the ID and increased the frequency and number of times that the manual task completion runs.

- I also spotted a small bug in the playbook in the Rasterize task - it would attempt to run even if inactive. Fixed as well.

4.5.0

   - No

### DIFF
--- a/Playbooks/playbook-Phishing_-_Core.yml
+++ b/Playbooks/playbook-Phishing_-_Core.yml
@@ -203,6 +203,14 @@ tasks:
                       value:
                         simple: Rasterize
                     ignorecase: true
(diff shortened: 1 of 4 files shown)


idovandijk:
Green build

PR #4863, opened 2019-11-12
Remove secrets check for forked PRs

Itay4:
## Status
Ready

## Description
Added check if forked PR or not, and run secrets checks accordingly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,13 +77,8 @@ jobs:
               find . -maxdepth 1 -type d -not \( -path . -o -path ./Integrations -o -path ./Scripts -o -path ./Beta_Integrations -o -path ./venv \) | xargs ./Tests/scripts/pyflake.sh
 
               [ -n "${BACKWARD_COMPATIBILITY}" ] && CHECK_BACKWARD=false || CHECK_BACKWARD=true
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -77,7 +77,13 @@ jobs:
               find . -maxdepth 1 -type d -not \( -path . -o -path ./Integrations -o -path ./Scripts -o -path ./Beta_Integrations -o -path ./venv \) | xargs ./Tests/scripts/pyflake.sh
 
               [ -n "${BACKWARD_COMPATIBILITY}" ] && CHECK_BACKWARD=false || CHECK_BACKWARD=true
-              python ./Tests/scripts/validate_files.py -c true -b $CHECK_BACKWARD
+              if [[ $CIRCLE_BRANCH =~ pull/[0-9]+ ]] ;
```
ronykoz: i think that it would be better to do this valiidation in the validate_files file, we already extract the branch name there, and we have the external regex that Fruend created few days ago

yaakovi:
add `fromversion: "4.5.0"` to `SecurityAdvisor` test in `Tests/conf.json`

.circleci/config.yml
Itay4: done

PR #4874, opened 2019-11-13
WildFire-v2 - changed the server param description

dantavori:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19467

## Description
Changed the server param description.
Should also be changed in the integration documentation.

   - No


Integrations/Palo_Alto_Networks_WildFire_v2/Palo_Alto_Networks_WildFire_v2.yml
```diff
@@ -3,7 +3,7 @@ commonfields:
   id: WildFire-v2
   version: -1
 configuration:
-- display: Server URL (e.g. https://192.168.0.1)
+- display: Server base URL (e.g. https://192.168.0.1/publicapi)
```
anara123: why publicapi is not part of the base url in the code?

dantavori: Because there are users that have already configured it themselves.
If you think it should be changed I can check in the code if they added it and if not then add it, but I wasn't sure if it's necessary

ronykoz: you can add a simple code check to make sure that it is there, and if not add a corresponding error msg or just add that yourself

dantavori: adding an error msg in test-module or something is not such a good idea if we want to allow both options (with/without the suffix), so I just add the suffix if it's not there

ronykoz:
@dantavori add RN please, it fails the build at the moment

Integrations/Palo_Alto_Networks_WildFire_v2/Palo_Alto_Networks_WildFire_v2.py
```diff
@@ -28,6 +33,11 @@
     'sample': '/get/sample'
 }
 
+if not URL.endswith('/publicapi'):
```
ronykoz: why do we have that twice

kirbles19:
@dantavori Looks good, no changes.

ronykoz:
force merging due to lack of license.

PR #4887, opened 2019-11-15
jira-issue-upload-file with ASCII filenames

orenzohar:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19707

## Description
`jira-issue-upload-file` fails with error 500 if the file to be uploaded contains special (non-ASCII). added a new argument **attachmentName** that will override the original filename when trying to upload.

   - No

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


kirbles19:
@orenzohar Done, no changes.

PR #4899, opened 2019-11-17
Joe security output fix

yaakovi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20167

## Description
no code changes necessary, already exports to that output.

## Screenshots
image

   - No


yaakovi:
force-push,
build fails only on outputpath changes

PR #4914, opened 2019-11-18
Fix create phishing classifier test

erezh31:
## Status
Ready

fixes https://github.com/demisto/etc/issues/20174


erezh31:
@yardensade run 3 times, seems fine

PR #4924, opened 2019-11-18
SumList Fix + UT

teizenman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20221


Scripts/SumList/SumList.py
```diff
@@ -1,33 +1,35 @@
...
-elif isinstance(value, str):
-    try:
-        result = int(value)
-        demisto.results(result)
-    except TypeError:
-        return_error('The string does not represent a number.')
-
-else:
-    return_error('This transformer applies only to a list of numbers.')
+            value = [int(item) for item in value.split(',')]
```
ronykoz: what about floats?

teizenman: Fixed

Scripts/SumList/SumList.py
```diff
@@ -1,33 +1,37 @@
...
-        else:
-            try:
-                result = int(value[0])
-                demisto.results(result)
-            except TypeError:
-                return_error('The string does not represent a number.')
-    else:
+
+def sum_value(value):
+    if isinstance(value, int):
```
ronykoz: Again, it could be a float

Scripts/SumList/SumList.py
```diff
@@ -1,33 +1,37 @@
...
+            return 'error', 'error'
+
+    elif isinstance(value, list):
+        try:
+            value = [float(item) for item in value]
+        except ValueError:
+            return 'error', 'error'
+
+    result = sum(value)
+    if result == int(result):
```
ronykoz: Why do we need this?

teizenman: Just to prettify the result, but it doesn't really matter

PR #4932, opened 2019-11-20
Qradar fix search results format

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20000

## Description
Fixed formatting of the get-search-results output.

   - No

## Must have
- [ ] Tests
- [ ] Code Review


guykeller:
@ronykoz The customers ran several playbooks and confirmed they are very happy with the results, and it's purely a "Prettier Format" bug - so I think we should go with them on this one.
What do you think?

PR #4935, opened 2019-11-20
added constants and get_matching_regex function

jochman:
## Status
ready


yardensade:
Do not merge stuff without review next time, it's important to keep track so we won't have to re-review everything later on.

PR #4941, opened 2019-11-20
fixing issue with update_id_set

roysagi:
## Status
Ready

## Description
fixing issue with update_id_set where it says the keys are unsubscribe


roysagi:
putting docs approved as this has no description/RN changes.

PR #4959, opened 2019-11-21
Slack notification to include unittests

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18846

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Tests/scripts/run_parallel_pkg_dev_tasks.py
+++ b/Tests/scripts/run_parallel_pkg_dev_tasks.py
@@ -92,7 +92,7 @@ def main():
             res = future.result()
             handle_run_res(res, fail_pkgs, good_pkgs)
     if fail_pkgs:
-        create_result_files(fail_pkgs.append(res[1]))
+        create_result_files(fail_pkgs)
         print_color("\n******* FAIL PKGS: *******", LOG_COLORS.RED)
         print_color("\n\t{}\n".format("\n\t".join(fail_pkgs)), LOG_COLORS.RED)
     if good_pkgs:
--- a/Tests/scripts/run_parallel_pkg_dev_tasks.py
+++ b/Tests/scripts/run_parallel_pkg_dev_tasks.py
@@ -20,7 +20,7 @@ def run_dev_task(pkg_dir: str, params: Optional[List[str]]) -> Tuple[subprocess.
     # color stderr in red and remove the warning about no config file from pylint
     cmd_line += r" 2> >(sed '/No config file found, using default configuration/d' | sed $'s,.*,\x1B[31m&\x1B[0m,'>&1)"
     res = subprocess.run(cmd_line, text=True, capture_output=True, shell=True, executable='/bin/bash')
-    return (res, pkg_dir)
+    return res, pkg_dir
 
 
 def should_run_pkg(pkg_dir: str) -> bool:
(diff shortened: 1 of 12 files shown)


Tests/scripts/run_parallel_pkg_dev_tasks.py
```diff
@@ -4,8 +4,11 @@
 import subprocess
 import concurrent.futures
 from typing import List, Optional, Tuple
+
 from pkg_dev_test_tasks import get_dev_requirements
 
+DEMISTO_GREY_ICON = 'https://3xqz5p387rui1hjtdv1up7lw-wpengine.netdna-ssl.com/wp-content/uploads/2018/07/Demisto-Icon-Dark.png'
```
Itay4: why is this needed here?

Tests/scripts/slack_notifier.py
```diff
@@ -40,9 +40,9 @@ def options_handler():
 
 
 def get_attachments(build_url, env_results_file_name):
-    content_team_fields, content_fields, failed_tests = get_fields()
+    content_team_fields, content_fields, failed_tests, failed_unittests = get_fields()
     color = 'good' if not failed_tests else 'danger'
-    title = 'Content Build - Success' if not failed_tests else 'Content Build - Failure'
+    title = 'Content Build - Success' if not (failed_tests or failed_unittests) else 'Content Build - Failure'
```
Itay4: shouldn't the operator in the condition be `and` ?

Tests/scripts/run_parallel_pkg_dev_tasks.py
```diff
@@ -82,6 +90,7 @@ def main():
             res = future.result()
             handle_run_res(res, fail_pkgs, good_pkgs)
     if fail_pkgs:
+        create_result_files(fail_pkgs)
```
Itay4: can you explain the flow here?

Tests/scripts/run_parallel_pkg_dev_tasks.py
```diff
@@ -4,8 +4,11 @@
 import subprocess
 import concurrent.futures
 from typing import List, Optional, Tuple
+
```
Itay4: remove empty line

Tests/scripts/run_parallel_pkg_dev_tasks.py
```diff
@@ -4,8 +4,11 @@
...
 from pkg_dev_test_tasks import get_dev_requirements
 
+DEMISTO_GREY_ICON = 'https://3xqz5p387rui1hjtdv1up7lw-wpengine.netdna-ssl.com/wp-content/uploads/2018/07/Demisto-Icon-Dark.png'
+
```
Itay4: remove empty line

Tests/scripts/run_parallel_pkg_dev_tasks.py
Shellyber: It's not, removing it now

Tests/scripts/run_parallel_pkg_dev_tasks.py
Shellyber: I want to get only the fail packages. In the circle it will look like this:
image

So I'm taking this packages and creating a file that later the slack_notifier will read and execute.

Tests/scripts/slack_notifier.py
Shellyber: No, because if one of the failures won't exist and the other yes, it will return a success with the `and` operator.

Shellyber:
Also will look like this:
image

Tests/scripts/run_parallel_pkg_dev_tasks.py
Itay4: ok got it
lets improve function name,and add docstring to it

Itay4:
@Shellyber looks good 👍 
see my comment above

PR #4964, opened 2019-11-21
test_utils:get_file - return in case of error should be empty dict

guyfreund:
## Status
Ready

## Description
return in case of error should be empty dict

   - No

## Must have
- [ ] Code Review


Tests/test_utils.py
```diff
@@ -160,7 +160,7 @@ def get_file(method, file_path, type_of_file):
             except Exception as e:
                 print_error(
                     "{} has a structure issue of file type{}. Error was: {}".format(file_path, type_of_file, str(e)))
-                return []
+                return {}
```
Itay4: dont we have test for this?

Itay4:
@guyfreund can we merge this?

PR #4969, opened 2019-11-24
Add Bandit to pkd_dev_tasks

Itay4:
## Status
Ready

## Description
added [Bandit](https://bandit.readthedocs.io/en/latest/) to run after the linters.
it does not fail the build in any case, but it prints the security issues found, as seen below in the screenshot

## Screenshots
image


glicht:
I like bandit. Would like to find the right level to enforce on. I think we can do level high confidence high as a start.

Itay4:
@glicht done

image

glicht:
We need to run with the proper python interpreter. Probably simplest is to give up on the run_bandit.sh file and simply call directly the proper python version  as we do with flake8. 

See: https://github.com/PyCQA/bandit#under-which-version-of-python-should-i-install-bandit

Tests/scripts/pkg_dev_test_tasks.py
```diff
@@ -260,6 +261,16 @@ def run_mypy(project_dir, py_num):
     print("mypy completed")
 
 
+def run_bandit(project_dir, py_num):
+    lint_files = get_lint_files(project_dir)
+    print("========= Running bandit on: {} ===============".format(lint_files))
+    python_exe = 'python2' if py_num < 3 else 'python3'
+    print_v('Using: {} to run flake8'.format(python_exe))
```
glicht:
```suggestion
    print_v('Using: {} to run bandit'.format(python_exe))
```

Tests/scripts/pkg_dev_test_tasks.py
```diff
@@ -260,6 +261,16 @@ def run_mypy(project_dir, py_num):
...
+    sys.stdout.flush()
+    subprocess.check_call([python_exe, '-m', 'bandit', '-lll', '-iii', lint_files], cwd=CONTENT_DIR)
```
glicht: Let's add `-q`. It prints out a lot of stuff when there is no need.
```suggestion
    subprocess.check_call([python_exe, '-m', 'bandit', '-lll', '-iii', '-q', lint_files], cwd=CONTENT_DIR)
```

PR #5005, opened 2019-11-26
Archer - Fixed retrieval of app IDs for applications with reverse field mapping. 

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19721

## Description
For several app IDs, such as 411, the field mapping which returns from Archer is reversed (the field ID is mapped to the field name instead of vice versa).
These cases used to cause an error, and are now handled.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/integration-ArcherRSA.yml
+++ b/Integrations/integration-ArcherRSA.yml
@@ -2754,6 +2754,18 @@ script:
         };
     }
 
+
+    function getFieldIdInReverseMapping(idToNameMapping, fieldNameToSearch) {
+        for (var fieldID in idToNameMapping) {
+            var currentFieldData = idToNameMapping[fieldID];
+            if ("Name" in currentFieldData && currentFieldData["Name"] === fieldNameToSearch) {
+                return fieldID;
+            }
+        }
+        return '';
+    }
+
+
     function searchRecords(command, dataArgs) {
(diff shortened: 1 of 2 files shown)


Integrations/integration-ArcherRSA.yml
```diff
@@ -2817,12 +2829,20 @@ script:
         var fieldIdToSearchOn;
         var fieldNameToSearchOn;
         if (fieldToSearchOn && searchValue) { // the user entered search parameters
-            fieldIdToSearchOn = nameToFieldMapping[fieldToSearchOn] ? nameToFieldMapping[fieldToSearchOn].Id : '';
+            isFieldNameInMapping = nameToFieldMapping[fieldToSearchOn] && typeof nameToFieldMapping[fieldToSearchOn] !== 'undefined';
+            fieldIdToSearchOn = isFieldNameInMapping ? nameToFieldMapping[fieldToSearchOn].Id : '';
+            if (fieldIdToSearchOn === '') {
+                fieldIdToSearchOn = getFieldIdInReverseMapping(nameToFieldMapping, fieldToSearchOn)
+            }
+            throw fieldIdToSearchOn;
```
hirshbergshachar: Why do we have this?

PR #5011, opened 2019-11-26
docxCriticalAndHighIncidentsReport - Fixed the capitalisation of the table's headers

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19889

## Description
The headers of the table's column s in the report were not capitalised.
Now they are.

## Screenshots
Before (on my mac):
Screen Shot 2019-11-26 at 17 51 06

After (notice the problem is fixed):
Screen Shot 2019-11-26 at 17 53 45

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Reports/report-docx-criticalAndHighIncidentsReport.json
+++ b/Reports/report-docx-criticalAndHighIncidentsReport.json
@@ -243,6 +243,15 @@
           "status",
           "dueDate"
         ],
+        "readableHeaders": {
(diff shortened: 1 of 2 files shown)


agamm:
Maybe it would be smart to try to generate the other default reports and see if their column headers are "readable too".

PR #5015, opened 2019-11-27
Fixing elasticsearch v2 test

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20334

## Description
Disabled warnings for insecure connection

### DIFF
--- a/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
+++ b/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
@@ -10,9 +10,11 @@ from elasticsearch_dsl.query import QueryString
 from datetime import datetime
 import json
 import requests
+import warnings
 
 # Disable insecure warnings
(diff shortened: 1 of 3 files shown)


Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -10,9 +10,11 @@
 from datetime import datetime
 import json
 import requests
+import warnings
```
ShahafBenYakir: Is this part of default python? Or you needed to install it?

roysagi: didn't need to install it.

ShahafBenYakir: Sounds good!

roysagi:
adding docs approved as there is no change in Changelog/yml/description.

PR #5019, opened 2019-11-27
Content version check

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20528

## Description
Added a function to update content version in commonServerPython

   - No


yardensade:
I think we should add it to the pre-commit hook stage, checks if content version in .yml is changed it will change the value in common server python and this will be pushed into the repo as well.
In circle it can test that the versions match as confirmation.
Changing artifacts with no validation could potentially render content broken and we can easily avoid that since we manually modify the content version anyway.

content_creator.py
```diff
@@ -146,9 +148,23 @@ def copy_test_files(bundle_test):
             shutil.copyfile(path, os.path.join(bundle_test, os.path.basename(path)))
 
 
-def main(circle_artifacts):
-    print('Starting to create content artifact...')
+def update_content_version(key, new_value):
```
yardensade: Should add try/except in case something here fails, print user friendly message that explains what went wrong and where to look for.

content_creator.py
```diff
@@ -146,9 +148,23 @@ def copy_test_files(bundle_test):
...
+    f = open('./Scripts/CommonServerPython/CommonServerPython.py', 'r')
+    lines = f.readlines()
+    f.close()
+    for i, line in enumerate(lines):
+        if line.split('=')[0].strip(' \n') == key:
+            lines[i] = f'{key} = "{new_value}"\n'
+    f = open('./Scripts/CommonServerPython/CommonServerPython.py', "w")
```
yardensade: No need to open the file twice, you can just open it once for both read and write

content_creator.py
```diff
@@ -146,9 +148,23 @@ def copy_test_files(bundle_test):
             shutil.copyfile(path, os.path.join(bundle_test, os.path.basename(path)))
 
 
-def main(circle_artifacts):
-    print('Starting to create content artifact...')
+def update_content_version(key, new_value):
+    f = open('./Scripts/CommonServerPython/CommonServerPython.py', 'r')
+    lines = f.readlines()
+    f.close()
+    for i, line in enumerate(lines):
```
yardensade: should either use regex over file content here or read the yml into a dict and pull the relevant key

Shellyber:
@yardensade The design for the the place of the function was made with @yaakovi 
Better discuss also with him.

yardensade:
Yes we talked about the 2 methods, there are pros and cons for both ways but it matters less which way was done as long as we have a proper way to test that everything is ok after this operation is done.

In our case we can have a test for the editing function, that will ensure we have the desired result and ease our minds that nothing broke in the process.
```
take the original file
run the function
compare the results
make sure a single line has changed verify that the line is what you expect it to be
```
This should do the trick.

content_creator.py
```diff
@@ -146,9 +148,23 @@ def copy_test_files(bundle_test):
...
+            lines[i] = f'{key} = "{new_value}"\n'
+    f = open('./Scripts/CommonServerPython/CommonServerPython.py', "w")
+    f.write(''.join(lines))
+    f.close()
+
 
+def main(circle_artifacts, content_version):
+
+    # update content_version in commonServerPython
+    update_content_version('CONTENT_RELEASE_VERSION', content_version)
```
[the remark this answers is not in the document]
Shellyber: @yaakovi what for?

content_creator.py
jochman: there's no flag for doing so. if so - please let me know!

content_creator.py
```diff
@@ -146,9 +149,26 @@ def copy_test_files(bundle_test):
...
+        with open(path, 'r+') as f:
+            content = f.read()
+            content = re.sub(regex, f"CONTENT_RELEASE_VERSION = '{content_ver}'", content, re.M)
+            f.seek(0)
+            f.write(content)
+    except Exception as ex:
+        print_warning(f'Could not open CommonServerPython File - {ex}')
 
+
+def main(circle_artifacts, content_version):
```
jochman:
```suggestion
def main(circle_artifacts, content_version, *args, **kwargs):
```
we don't know if someone will pass something into the file, so that's a good way to prevent errors.

content_creator.py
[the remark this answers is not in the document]
yardensade: Lichtman asked for it, it is important for knowing when not to use this feature.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -24,6 +24,8 @@
 except Exception:
     pass
 
+CONTENT_RELEASE_VERSION = '0.0.0'
+CONTENT_BRANCH_NAME = 'master'
```
yardensade: I don't see any handling for branch name, should be in same place as update content release version

Shellyber: @yaakovi What should be done here?

content_creator.py
```diff
@@ -1,4 +1,6 @@
+
```
yaakovi: remove this line

content_creator.py
```diff
@@ -146,9 +149,24 @@ def copy_test_files(bundle_test):
...
+            f.write(content)
+    except Exception as ex:
+        print_warning(f'Could not open CommonServerPython File - {ex}')
+
 
+def main(circle_artifacts, content_version):
+
+    # update content_version in commonServerPython
+    update_content_version(content_version)
+    print(content_version)
```
yaakovi: add a better informative print

Scripts/CommonServerPython/CommonServerPython.py
yaakovi: replace `'master'` with current branch name.
branch name can be retrieved using:
```
branches = run_command('git branch')
branch_name_reg = re.search(r'\* (.*)', branches)
branch_name = branch_name_reg.group(1)
```

Scripts/CommonServerPython/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+    -
```
yaakovi:
```suggestion
-
```

Shellyber:
@yardensade @yaakovi All done! Please review and approve

Shellyber:
Image from the instance 
image

Itay4:
@yaakovi please review again

PR #5021, opened 2019-11-27
Virustotal dbotscore duplication fix

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18022

## Description
Added VirusTotal 5.5 - now able to use DT with DbotScore.

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


roysagi:
To easily understand the fix I only changed:
ec.DBotScore => ec["DBotScore(val.Indicator && val.Indicator === obj.Indicator && val.Vendor === obj.Vendor && val.Type === obj.Type)"] wherever it appeared.

@guykeller - to streamline your CR
@mayagoldb -FYI

roysagi:
Also a question - though it looks like I added a new integration this has the same features and descriptions as VirusTotal.
Also the bug fix will only take affect from Demisto 5.5 - so I don't know if it should appear in current release notes.

Integrations/integration-VirusTotal_5.5.yml
```diff
@@ -0,0 +1,1317 @@
+commonfields:
+  id: VirusTotal 5.5
```
anara123:
```suggestion
  id: VirusTotal
```

Integrations/integration-VirusTotal_5.5.yml
```diff
@@ -0,0 +1,1317 @@
...
+  version: -1
+name: VirusTotal 5.5
```
anara123:
```suggestion
name: VirusTotal
```

Integrations/integration-VirusTotal.yml
```diff
@@ -3,6 +3,7 @@ commonfields:
   version: -1
 name: VirusTotal
 display: VirusTotal
+toversion: 5.5.0
```
anara123:
```suggestion
toversion: 5.1.0
```

Integrations/integration-VirusTotal_5.5.yml
```diff
@@ -0,0 +1,1317 @@
...
+display: VirusTotal 5.5
```
anara123:
```suggestion
display: VirusTotal
```

anara123:
@roysagi please create self service with demisto master and upload the content.zip to make sure that VT is uploaded and everything works properly

ronykoz:
@roysagi before merging please do a diff between the VT we have in master to check that only the DTs were updated

dantavori:
The previous version of the 5.5 integration didn't contain the last update in VT (batch support for generic commands) and the TPB tested these changes. That's the reason for the build failure. Now it is fixed.

dantavori:
Ran `!ip` multiple times in a self-service instance, and it seems to do the job (incidents are created and there are no duplicates).
FYI, I got this error once: `No content received. Possible API rate limit reached.`. From searching in Slack it seems that our VT instance isn't very stable, just want to make sure.

ronykoz:
@anara123 so are we ok with this?

anara123:
Lets merge then. Make sure that in release notes we mention that this fix
will be available only from 5.5

ronykoz:
@roysagi lets add changelog and state that the fix is only from Demisto version 5.5 please

roysagi:
Added RN and also made sure again that all the changes against the version in master are of DT only - all good.

Integrations/integration-VirusTotal_CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## [Unreleased]
   - Added batch support for the **ip** and **url** and **domain** commands.
+  - Fixed an issue Where DBotScore would create duplications in context. This will take effect from Demisto version 5.5.
```
amshamah419:
```suggestion
  - Fixed an issue where the DBotScore would create duplications in the incident context. This effects Demisto version 5.5 and higher.
```

PR #5023, opened 2019-11-27
Fixed issue with Qradar scripts

guykeller:
## Status 
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20289

## Description
The Qradar scripts were only compatible for one kind of CRE name default value.
This is now fixed.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


Scripts/QRadarGetCorrelationLogs/QRadarGetCorrelationLogs.py
```diff
@@ -1,86 +1,106 @@
...
+                "identityip, CATEGORYNAME(category), PROTOCOLNAME(protocolid), " \
+                "LOGSOURCENAME(logsourceid){1} from events " \
+                "where \"CRE Name\" <> NULL AND INOFFENSE({2}) START '{3}'"
+    return query
+
+
+def main():
+    d_args = demisto.args()
+    is_cre_name_null = demisto.args().get("is_cre_name_null", "True")
+    QUERY = get_query(is_cre_name_null)
```
anara123: there is no need to QUERY be with upper case

Scripts/QRadarGetOffenseCorrelations/QRadarGetOffenseCorrelations_test.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from QRadarGetOffenseCorrelations import get_query
+
+
+@pytest.mark.parametrize('bool_val', ["False"])
+def test_get_query_cre_name_null_false(bool_val):
+    query_from_false = get_query(bool_val)
```
anara123: I think in tests you better mock the demisto.args() - pass is_cre_name_null=True and mock demisto.executeCommand and validate that the passed query contains `"\"CRE NAME\" <> NULL"` or you can compare the whole query

Scripts/QRadarGetCorrelationLogs/QRadarGetCorrelationLogs.py
guykeller: It was like that before and I'm not refactoring the script, just re wrote it to be more easily testable

Scripts/QRadarGetOffenseCorrelations/QRadarGetOffenseCorrelations_test.py
guykeller: How would that improve the test from your perspective?

anara123: it will make it more end to end test.

PR #5028, opened 2019-11-28
Misp - added data filtering

teizenman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20277

## Description
Most of the data is stored in `Attributes` section which exists for every event and every related event.


Integrations/MISP_V2/MISP_V2.py
```diff
@@ -349,7 +369,13 @@ def get_dbot_level(threat_level_id: str) -> int:
     return 0
 
 
-def check_file():
+def get_files_events():
+    files = argToList(demisto.args().get('file'), ',')
+    for file in files:
```
ronykoz: File is a python saved word, use something else please

Integrations/MISP_V2/MISP_V2.py
```diff
@@ -231,6 +232,24 @@ def replace_keys(obj_to_build: Union[dict, list, str]) -> Union[dict, list, str]
...
+
+    # each event has it's own attributes
+    for attribute in context_data[0]['Attribute']:
+        for key in list(attribute.keys()):
+            if key not in DATA_KEYS_TO_SAVE:
+                del attribute[key]
+
+    # each related event has it's own attributes
+    for obj in context_data[0]['Object']:
+        for attribute in obj['Attribute']:
```
ronykoz: Seems like code duplication, separate to another function

Integrations/MISP_V2/MISP_V2.yml
```diff
@@ -19,6 +19,29 @@ configuration:
   name: insecure
   required: false
   type: 8
+- display: Select which event's attribute data fields will be saved to context. Default
```
ronykoz: Not clear for how the user will choose “all” maybe mention that by leaving it empty it will take all the fields

teizenman:
@kirbles19 Please verify RN and arguments at:
1. https://github.com/demisto/content/blob/misp/Integrations/MISP_V2/CHANGELOG.md
2. https://github.com/demisto/content/blob/misp/Integrations/MISP_V2/MISP_V2.yml

ronykoz:
@teizenman maybe you should add a test for this new logic, just to verify it will not break in the future

Integrations/MISP_V2/test_constants.py
```diff
@@ -0,0 +1,328 @@
+full_response_before_filtering = [
```
ronykoz: We shouldn’t have another py file in the integration directory, if you want to add files that are relevant to the tests add a directory called test_data and put those files in there

ronykoz:
You should fix the import in the test as well for this to work

teizenman:
It finds the import, the UT passed successfully both locally and in circle.

kirbles19:
@teizenman Done.

PR #5029, opened 2019-11-28
Fixed an issue with URL extraction from pdf

reutshal:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20178

## Description
Added another way to extract urls from a pdf file (the old way missed some urls from a specific kind of pdf) 

   - No

## Must have
- [ ] Tests
- [X] Code Review

## Technical writer review
- [ ] [YAML file](link)
- [X] [CHANGELOG](https://github.com/demisto/content/blob/readPDFFile-url_extracting_fix/Scripts/ReadPDFFileV2/CHANGELOG.md)


Scripts/ReadPDFFileV2/ReadPDFFileV2.py
```diff
@@ -213,6 +213,21 @@ def build_readpdf_entry_object(pdf_file, metadata, text, urls, images):
     return results
 
 
+def get_urls_from_binary_file(file_path):
+    """Reading from the binary pdf in the pdf_text_output_path and returns a list of the urls in the file"""
+    with open(file_path, 'rb') as file:
+        # the urls usually appear in the form: '/URI (url)'
+        urls = re.findall(r'/URI \((.*?)\)', str(file.read()))
+
+    binary_file_urls = list()
```
ronykoz: How can we do union between list and set?

Scripts/ReadPDFFileV2/ReadPDFFileV2.py
```diff
@@ -213,6 +213,21 @@ def build_readpdf_entry_object(pdf_file, metadata, text, urls, images):
     return results
 
 
+def get_urls_from_binary_file(file_path):
+    """Reading from the binary pdf in the pdf_text_output_path and returns a list of the urls in the file"""
+    with open(file_path, 'rb') as file:
+        # the urls usually appear in the form: '/URI (url)'
+        urls = re.findall(r'/URI \((.*?)\)', str(file.read()))
```
ronykoz: How does that find url when reading in binary? Shouldn’t it be text?

Scripts/ReadPDFFileV2/ReadPDFFileV2.yml
```diff
@@ -128,6 +128,6 @@ subtype: python3
 runas: DBotWeakRole
 runonce: false
 tests:
-- Extract Indicators From File - test
+- Extract Indicators From File - Generic v2 - Test
```
ronykoz: Why are we changing this?

reutshal: This test is using "Extract Indicators From File - Generic" playbook, which calls the readPdfFile script (the old one).
This means that until now this test didn't even use the readPdfFileV2 script.
Thought to update the used playbook to "Extract Indicators From File - Generic v2",
but after consulting @idovandijk we decided that it will be better to use "Extract Indicators From File - Generic v2 - Test" which covers more cases (and of-course using the new readPdfFile script).

Scripts/ReadPDFFileV2/ReadPDFFileV2.py
reutshal: A weird error was thrown when I tried to convert to set, checked this union in python console and it worked. Now I tried it again and the error is gone, but I don't think this casting is mandatory. 
Can do both ways.
Can also initiate it to be set rather than a list

ronykoz: I think that set would be nicer here, so please do that

PR #5039, opened 2019-12-01
Added RN structure validation + UTs

teizenman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18954


ronykoz:
Not sure if structure is the best place as we want to check RN for all content entities, which can be packs/playbooks/scripts/integrations/incidentfields/.... maybe consult with @yardensade on the best approach here

teizenman:
Moved to sdk.

PR #5059, opened 2019-12-02
Fix Threatq test

dantavori:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19468

   - No


dantavori:
@ShahafBenYakir basically the issue should be fixed. However, on the first build of this PR, the TPB failed on status code 500 returned from an API request. It doesn't happen a lot but it did happen a few times.
What do you think? is it possible to run the integration multiple times? should the test be skipped forever? or maybe there is a way to stabilize the instance?

orenzohar:
@dantavori why are we getting status code 500?

dantavori:
@orenzohar 500 is an internal server error - it just happens sometimes without a specific reason, and in the next request it will work (99%). Have you encountered it?

ShahafBenYakir:
@dantavori 
Regarding your question about the 500 error, there are some integrations where this occurred but mainly on the test module part and not while running the playbook (for example in Mcafee TIE). 
I see that you introduced some changes to the test playbook that includes some new tasks, but no changes to the existing tasks or the integration itself. Are these changes suppose to solve most of the fail issues we had so far (except the 500)? If so, let's remove it from skipped and see what is happening. If not I think we need to look deeper into this and understand the behavior better, for example, Is it a time-based issue(occurs every x minutes)? maybe this occurs only in specific commands? and so on. Maybe we'll consider retires or other creative solutions that could also help our customers to avoid bugs. If we will not see any line logic within these failures then I guess we'll have no other choice then skipping it forever.

dantavori:
@ShahafBenYakir the last test failure was due to duplications of indicators with the same value (which occurred for a reason I haven't completely understood - probably when a test fails at the moment between creation and deletion of an indicator). TL;DR - in the fix I prevent duplications by checking if the indicator already exists and if so, I pass over the creation command (did the same for all the types of objects, just in case).
We can merge now and see if this fix is enough, and if we have a failure again and it's the 500 error, then several retries of the TPB run could be a good solution.

dantavori:
Here's an example of the 500 error from this PR's build:
image

If I run the same exact command locally:
image
i.e the object was already deleted, which makes me wonder if maybe the deletion was executed, but for some reason the 200 response was not received properly to Demisto.

Generally, the command works.
image

orenzohar:
@dantavori maybe change this task to not fail on error, and add a check after it to make sure the indicator was deleted?

dantavori:
The problem is it's not necessarily in deletion... It could be in any command

orenzohar:
@dantavori okay then, @ShahafBenYakir isn't the mocking should cove cases like this?

ShahafBenYakir:
@orenzohar You are right this is exactly the idea of mocks however ThreatQ (much like many other integrations) still does not work good enough with mocks. 
@dantavori The extra verification you added is a good idea for cases of duplicates (as occurred yesterday) but correct me if I'm wrong, this error is only a consequence of the real problem discussed here - the 500 internal server error issue. Let's try to dig a bit deeper together, with either Rony or me if you don't care to wait for me to be back from Korea.

dantavori:
Rony and I agreed that we should try the three tries idea. In addition, I'll share the 500 error issue with ThreatQ support.

PR #5074, opened 2019-12-04
Cisco align image fix

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20488

## Description
Updated the image to align with the rest of cisco integrations

   - No


Integrations/integration-IronPort.yml
```diff
@@ -4,9 +4,9 @@ commonfields:
...
 description: Cisco Email Security protects against ransomware, business email compromise,
-  spoofing, and phishing
+  spoofing, and phishing.
```
yardensade:
```suggestion
  spoofing, and phishing
```

PR #5080, opened 2019-12-04
remove Get File Sample By Hash - Generic - Test

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19841

## Description
The playbook `Get File Sample From Hash - Generic` use the playbook `Get File Sample By Hash` for the deprecated integration Carbon Black Enterprise Response(Go version) and for Cylance protect.

No need to test it in Carbon Black and for testing in Cylance Protect we have `Get File Sample By Hash - Cylance Protect - Test`

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -1823,9 +1823,6 @@
             "playbookID": "efc817d2-6660-4d4f-890d-90513ca1e180"
         },
         {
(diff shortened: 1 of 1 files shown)


ShahafBenYakir:
I think this is not the right issue. Isn't it demisto/etc#19841?

PR #5084, opened 2019-12-04
Splunk - Changed command structure to new format

guykeller:
## Status
Ready

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/SplunkPy/SplunkPy.py
+++ b/Integrations/SplunkPy/SplunkPy.py
@@ -205,129 +205,77 @@ def request(url, message, **kwargs):
     }
 
 
-service = None
-proxy = demisto.params()['proxy']
-if proxy:
+# COMMANDS
+
+def parse_raw_command():
+    raw = demisto.args()['raw']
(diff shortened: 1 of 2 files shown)


Integrations/SplunkPy/SplunkPy.py
```diff
@@ -205,44 +205,159 @@ def request(url, message, **kwargs):
...
+    raw = demisto.args()['raw']
+    rawDict = rawToDict(raw)
+    ec = {}
+    ec['Splunk.Raw.Parsed'] = rawDict
+    demisto.results({"Type": 1, "ContentsFormat": "json", "Contents": json.dumps(rawDict), "EntryContext": ec})
+    sys.exit(0)
+
+
+def notable_event_edit_command():
+    if not proxy:
```
ronykoz: why not handle_proxy()

guykeller: Because I'm not changing any logic on this PR, just changing the structure. Logic will be later, there are many such changes to make

PR #5088, opened 2019-12-04
Fixed description for ir-get-alarm command - LogRhythm

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20409

## Description
Fixed description for ir-get-alarm command

   - No

## Technical writer review
- [ ] [YAML file

### DIFF
--- a/Integrations/integration-LogRhythm.yml
+++ b/Integrations/integration-LogRhythm.yml
@@ -536,8 +536,12 @@ script:
             start.setDate(start.getDate()-30);
             break;
         case 'Custom':
-            if(!startArg || !endArg){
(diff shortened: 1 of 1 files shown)


Shellyber:
@kirbles19 Please review the description change in ir-get-alarm command

yardensade:
Just to be sure, either one of the arguments is required, not both in the same time, right?

Shellyber:
@yardensade According to the issue, you can filter this command only by status and not by entity.
So fixed the description according to that

kirbles19:
@Shellyber Done.

PR #5105, opened 2019-12-05
Build shift

ShahafBenYakir:
## Status
Ready

## Description
Skipping for life

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -2063,7 +2063,7 @@
 
         "_comment": "~~~ INSTANCE ISSUES ~~~",
         "VxStream": "Exceeded quota, remove in 1/1/19",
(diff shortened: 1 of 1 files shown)


Tests/conf.json
```diff
@@ -2000,6 +2000,8 @@
         }
     ],
     "skipped_tests": {
+        "Test-Detonate URL - Crowdstrike": "Issue 19439",
```
guyfreund: @ShahafBenYakir - https://github.com/demisto/etc/issues/19439 has been closed, are you sure it's related? Seems like a mistake in the issue number.

ShahafBenYakir: Seems like a mistake indeed, you can run locally and if passed you can un-skip and see that it passes.

PR #5106, opened 2019-12-06
Slack - polling incremental backoff

liorblob:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20529

## Description
Added polling intervals according to amount of time passed

   - No

## Must have
- [X] Tests
- [ ] Code Review


Integrations/Slack/Slack.py
```diff
@@ -649,6 +663,28 @@ def check_for_answers(now: datetime):
     set_to_latest_integration_context('questions', questions)
 
 
+def get_poll_minutes(current_time: datetime, received: Optional[str]):
```
Itay4:
```suggestion
def get_poll_minutes(current_time: datetime, received: Optional[str]) -> int:
```

Integrations/Slack/Slack.py
```diff
@@ -1250,6 +1286,7 @@ def save_entitlement(entitlement, thread, reply, expiry, default_response):
         'entitlement': entitlement,
         'reply': reply,
         'expiry': expiry,
+        'received': datetime.strftime(get_current_utc_time(), DATE_FORMAT),
```
Itay4: i think `sent` is more clear than `received`

Integrations/Slack/Slack.py
```diff
@@ -98,6 +102,13 @@ def test_module():
     demisto.results('ok')
 
 
+def get_current_utc_time():
```
Itay4:
```suggestion
def get_current_utc_time() -> datetime.datetime:
```

PR #5111, opened 2019-12-08
Packagify unzipfile

jochman:
## Status
Ready

## Description
Just packagify

   - No


Scripts/UnzipFile/UnzipFile.yml
```diff
@@ -0,0 +1,50 @@
...
+  id: UnzipFile
+  version: -1
+enabled: true
+name: UnzipFile
+outputs:
+- contextPath: ExtractedFiles
+  description: A list of file names that were extracted from the ZIP file.
+  type: Unknown
+script: '-'
+subtype: python3
```
Itay4:
```suggestion
subtype: python2
```

Scripts/UnzipFile/UnzipFile.yml
```diff
@@ -0,0 +1,50 @@
...
+system: false
+tags:
+- Utility
+- file
+timeout: '0'
+type: python
+dockerimage45: demisto/7zip
```
Itay4: why ?

jochman: no tag, so I'm preserving bc.

Itay4: `latest` is redundant in this case, lets keep the docker image as is

PR #5119, opened 2019-12-08
Fix opening threads ewsv2

jochman:
# Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/20566

   - No

## Must have
- [X] Tests
- [X] Documentation (with link to it)
- [X] Code Review

### DIFF
--- a/Integrations/EWSv2/CHANGELOG.md
+++ b/Integrations/EWSv2/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+Fixed issue where threads not closed after executing commands.
 
 ## [19.11.1] - 2019-11-26
 - Logging improvement.
--- a/Integrations/EWSv2/EWSv2.py
+++ b/Integrations/EWSv2/EWSv2.py
@@ -2087,6 +2087,7 @@ def main():
     finally:
         try:
             if isinstance(config, Configuration):
+                # Sometimes new threads are created but never killed, so killing them manually.
                 config.protocol.thread_pool.terminate()
         except Exception:
             demisto.debug("Error was found in terminating threads in config.protocol, ignoring.")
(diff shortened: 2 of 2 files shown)


Itay4:
@jochman @IkaDemisto lets make sure this makes it to the release

IkaDemisto:
@glicht @Itay4 we don't gain any improvement in `EWS v2` integration with this change in contrast to `EWS Mail Sender`. Thread numbers remains fixed size with/without this change and there is no impact on memory usage inside docker container. However terminating thread pool will have negative effect and will break all integration's commands that use indirectly `EWSPooledMixIn` class (is using protocol.thread_pool also https://github.com/ecederstrand/exchangelib/blob/10fc6d8f73894f11c4f131da2d7d983947fbfac9/exchangelib/services/common.py). 
error
Probably there is a solution to close/terminate thread pool in less aggressive way without breaking other functionality of `exchagelib` but my suggestion is not to merge this changes.

jochman:
> @glicht @Itay4 we don't gain any improvement in `EWS v2` integration with this change in contrast to `EWS Mail Sender`. Thread numbers remains fixed size with/without this change and there is no impact on memory usage inside docker container. However terminating thread pool will have negative effect and will break all integration's commands that use indirectly `EWSPooledMixIn` class (is using protocol.thread_pool also https://github.com/ecederstrand/exchangelib/blob/10fc6d8f73894f11c4f131da2d7d983947fbfac9/exchangelib/services/common.py).
> error
> Probably there is a solution to close/terminate thread pool in less aggressive way without breaking other functionality of `exchagelib` but my suggestion is not to merge this changes.

I think that @IkaDemisto is right, We've already seen that there's a problem with this change and I'm afraid that the change that we did in EWSMailSender is also problematic, but we need to wait for feedback.

jochman:
@IkaDemisto, Shall I merge?

PR #5132, opened 2019-12-09
UnitTestFilters

dantavori:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20568
fixes: https://github.com/demisto/etc/issues/19892

## Description
- Added support for test filters in packs.
- configure_tests.py will now ignore deprecated integrations/scripts/playbooks.

   - No


Tests/scripts/configure_tests.py
```diff
@@ -228,13 +242,26 @@ def get_test_ids(check_nightly_status=False):
...
     for integration in integration_set:
         integration_id = list(integration.keys())[0]
         integration_data = list(integration.values())[0]
         if integration_id in integration_ids:
-            integration_to_command[integration_id] = integration_data.get('commands', [])
+            integration_commands = set(integration_data.get('commands', []))
+            integration_deprecated_commands = set(integration_data.get('deprecated_commands', []))
+            if integration_deprecated_commands:
+                deprecated_names = ', '.join(integration_deprecated_commands)
+                deprecated_commands_string += '{}: {}\n'.format(integration_id, deprecated_names)
```
ronykoz: it can be done in line 260

Tests/scripts/configure_tests.py
```diff
@@ -347,17 +375,100 @@ def collect_changed_ids(integration_ids, playbook_names, script_names, modified_
...
+            script_data = list(script.values())[0]
+            if script_data.get('deprecated', False):
+                deprecated_scripts_string += '   ' + script_name + '\n'
+                script_names.remove(script_name)
+
+    if deprecated_scripts_string:
+        deprecated_message = '  The following scripts are deprecated and are not taken ' \
+                             'into account in the test collection:\n{}\n'.format(deprecated_scripts_string)
+
+    return deprecated_message
```
ronykoz: i think that all of those functions are generic, maybe we should create one function for all of them

ronykoz:
Looks good!

Tests/scripts/configure_tests.py
dantavori: outside the for loop?

ronykoz:
Lets just change the printing as discussed, and check the commands for integrations

PR #5133, opened 2019-12-09
Final fix thread ewsmailsender

jochman:
## Status
Ready

## Description
Used a less aggressive way to kill threads.

   - No

TESTED AND WORKING


glicht:
Looking at: https://github.com/ecederstrand/exchangelib/blob/master/exchangelib/protocol.py#L359 I see that exchangelib later changed to use terminate and close+join.

jochman:
@glicht it failed on a different playbook everytime, might need to force merge it.

jochman:
so I messed things up, opening a new pr.

PR #5134, opened 2019-12-09
Deprecated Intezer v1 and fixed Intezer v2 test

DeanArbel:
## Status
Ready

## Description
Deprecated Intezer v1, and changed the Intezer test to reflect that they now support `md5`.

   - No


DeanArbel:
Removed `docs-approved` due to adding additional changes to the playbooks

PR #5144, opened 2019-12-10
fix Cofense Test func and fetch and outputs in yml

orenzohar:
## Status
Ready

## Related Issues
fixes wrong outputs in yml
fixes: https://github.com/demisto/etc/issues/20700

## Description
fix for cofense triage test function and context outputs paths. also updated output context paths


orenzohar:
@yuvalbenshalom test passed in demo4 
image

yuvalbenshalom:
outputs in yml changed because they wear wrong in the first place. this fix correct them to be the same as in the code.
force merging because it fails the BC check.

PR #5160, opened 2019-12-12
Bug fix: GetTime js script

yardensade:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20639

Then:
image

Now:
image


Scripts/script-GetTime.yml
```diff
@@ -4,8 +4,14 @@ commonfields:
...
 script: |-
   var d = new Date();
-  if ((args.date) && (args.date.length > 0)){
-      d =  new Date(args.date);
+  if (args.date){
+      if (args.date.length > 0){
+          // args.date is string
+          d =  new Date(args.date);
+      } else {
+          // args.date is object (Time.time in Go)
```
Itay4: can we check if its actually an object, and else throw exception?

yardensade: we can but it wouldn't really provide any value since this arg can only be used as string or object, anything else would fail anyway. Even if it's object but not Time.time in Go.

yardensade: And there is no actual way for me to determine that this object is Go's time.time object

Itay4: ok leave as is and lets just add test for the case is `Time.time`

Scripts/script-GetTime_CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## [Unreleased]
+Fixed a bug where providing date input from context returned current date instead of provided date.
```
yaakovi:
```suggestion
Fixed an issue where providing date input from context returned current date instead of provided date.
```

Playbooks/playbook-GetTime-ObjectVsStringTest.yml
```diff
@@ -0,0 +1,262 @@
+id: GetTime-ObjectVsStringTest
```
Itay4:
```suggestion
id: GetTime - Test
```

Playbooks/playbook-GetTime-ObjectVsStringTest.yml
```diff
@@ -0,0 +1,262 @@
...
+version: -1
+contentitemfields:
+  propagationLabels: []
+name: GetTime-ObjectVsStringTest
```
Itay4:
```suggestion
name: GetTime - Test
```

Itay4:
oh i see theres already a test playbook, why didnt we enhance it ?

yardensade:
we basically simulated unit test in test playbook, was hard enough to make sure the data is reliable and I didn't want to mix it with existing context to prevent mistakes. (we need to get object from context instead of string and then be able to test it as string as well even tho it does not exist as string, way to do this was getting occured time from incident context)

yardensade:
@kirbles19 please review the changelog.

Scripts/script-GetTime_CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## [Unreleased]
+Fixed an issue where providing date input from context returned current date instead of provided date.
```
amshamah419:
```suggestion
Fixed an issue where providing a date input from context returned current date instead of provided date.
```

PR #5163, opened 2019-12-12
changed handle proxy spot to be after the get token operation

ShahafBenYakir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18744

## Description
Changed handle proxy spot to be after the get token operation

   - No

## Dependencies
- RSA NetWitness v11.1

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/4d993465bdc5133c29b5430e592186a4fbbd9e50/Integrations/RSANetWitness_v11_1/CHANGELOG.md)

### DIFF
--- a/Integrations/RSANetWitness_v11_1/CHANGELOG.md
+++ b/Integrations/RSANetWitness_v11_1/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## [Unreleased]
+  - Fixed an issue where environment proxy effects the integration when even though no proxy should be used.
 
 
 ## [19.11.0] - 2019-11-12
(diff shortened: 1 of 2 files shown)


ado120:
May have tagged the wrong Alex

yuvalbenshalom:
reviewd the RN with @Itay4

ShahafBenYakir:
There is a small grammar issue here. We should change this to:
**Fixed an issue where environment proxy effects the integration even though no proxy should be used.**
(Removed the “when”)

Itay4:
@ShahafBenYakir its ok, leave it for now, we will handle it in the rn

PR #5166, opened 2019-12-14
IntSights close-alert handle arguments prorperly

Itay4:
## Status
Ready

## Related Issues
fixes https://github.com/demisto/etc/issues/20715

## Description
`is-hidden` and `rate` in the command `close-alert` were not handled well when checked if provided.
didn't add test because there is no command to create an alert.

## Screenshots
image

Any

   - No

## Must have
- [ ] Code Review

## Dependencies
None

### DIFF
--- a/Integrations/IntSight/IntSight.py
+++ b/Integrations/IntSight/IntSight.py
@@ -441,7 +441,7 @@ def close_alert():
     alert_id = demisto.getArg('alert-id')
     reason = demisto.getArg('reason')
     free_text = demisto.getArg('free-text')
-    is_hidden = demisto.getArg('is-hidden')
+    is_hidden = demisto.getArg('is-hidden') == 'True'
     rate = demisto.getArg('rate')
     close_details = {'ID': alert_id, 'Close Reason': reason, 'Closed FreeText': free_text, 'Closed Rate': rate,
                      'IsHidden': is_hidden}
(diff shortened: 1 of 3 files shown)


Integrations/IntSight/IntSight.yml
```diff
@@ -656,7 +661,7 @@ script:
       description: IP address.
       type: String
     - contextPath: Domain.Name
-      description: 'The domain name. For example, "google.com".'
+      description: The domain name. For example, "google.com".
```
yardensade: Looks like some of the description got the `'`sign removed while others not, any idea why?

Integrations/IntSight/IntSight.py
```diff
@@ -441,7 +441,7 @@ def close_alert():
     alert_id = demisto.getArg('alert-id')
     reason = demisto.getArg('reason')
     free_text = demisto.getArg('free-text')
-    is_hidden = demisto.getArg('is-hidden')
+    is_hidden = demisto.getArg('is-hidden') == 'True'
```
yardensade: Why not strtobool()?

Integrations/IntSight/IntSight.yml
Itay4: pycharm does that automatically, doesnt really matter

Integrations/IntSight/IntSight.py
Itay4: dont really like `strtobool` in command args, since the user can pass whatever there and can cause weird issues, so prefer that as idiot-proof

PR #5170, opened 2019-12-15
Crowdstrike duplicates problem -refix

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19464

## Description
1) last fetch time now takes milliseconds into affect.
2) Last detection id fetched is now saved to prevent refetching it.


Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -664,18 +670,32 @@ def fetch_incidents():
...
+
+        # make sure we do not fetch the same detection again.
+        if last_detection_id == detections_ids[0]:
+            first_index_to_fetch = 1
+
+            # if this is the only detection - dont fetch.
+            if len(detections_ids) == 1:
+                return incidents
+
+        else:
```
ronykoz: add a comment that explains that this is for the first fetch basically

Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -686,10 +706,11 @@ def fetch_incidents():
                 if incident_date_timestamp > last_fetch_timestamp:
                     last_fetch = incident_date
                     last_fetch_timestamp = incident_date_timestamp
+                    last_detection_id = json.loads(incident['rawJSON']).get('detection_id')
```
ronykoz: is it numerical?

roysagi: no - it's a string with both letters and numbers

Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
roysagi: it could be more than just the first fetch - I'll add a comment simply saying "if the first detection in this pull is different than the last detection fetched we bring it as well".

ronykoz:
@yaakovi can we merge that? i saw you reverted some changed here today

ronykoz:
@roysagi i think you need to update the Rn you added(the location of it) and then merge this
@Itay4 FYI

roysagi:
@ronykoz - I merged master into this branch - I don't see why I need to change the RN?

speaking of which - @amshamah419 - Any chance you can go through the changelog?

PR #5179, opened 2019-12-15
Add headers to oproxy requests

adi88d:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20182

## Description
Adding the headers `X-Content-Version` and `X-Content-Version` to oproxy requests contains the integration brand and content release version.


liorblob:
I think it's kind of redundant because we're going to have a unified API for this...

PR #5213, opened 2019-12-19
Build updating fixes

avidan-H:
## Status

## Description
Only perform configure and update on non-nightly master builds. Reconfigure client before disabling instances since authorization has probably expired.

## Screenshots

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,9 @@ jobs:
             demisto-sdk create -a $CIRCLE_ARTIFACTS
             # create zip with contents of content_new.zip and content_test.zip for use in updating content on instances
             cp "$CIRCLE_ARTIFACTS/content_new.zip" "$CIRCLE_ARTIFACTS/all_content.zip"
-            zip -j "$CIRCLE_ARTIFACTS/all_content.zip" "$CIRCLE_ARTIFACTS/bundle_test/*"
+            unzip -q "$CIRCLE_ARTIFACTS/content_test.zip" -d "$CIRCLE_ARTIFACTS/test_content"
+            zip -j "$CIRCLE_ARTIFACTS/all_content.zip" $CIRCLE_ARTIFACTS/test_content/*
+            rm -r $CIRCLE_ARTIFACTS/test_content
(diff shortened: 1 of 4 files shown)


Tests/configure_and_test_integration_instances.py
```diff
@@ -612,6 +612,8 @@ def main():
         if not success:
             postupdate_fails.add((instance_name, integration_of_instance))
 
+    # reinitiate the client since its authorization has probably expired by now
+    client = demisto_client.configure(base_url=server, username=username, password=password, verify_ssl=False)
```
yaakovi: switch to api key instead of user-password combo.
do it in another PR

.circleci/config.yml
```diff
@@ -173,7 +173,9 @@ jobs:
             demisto-sdk create -a $CIRCLE_ARTIFACTS
             # create zip with contents of content_new.zip and content_test.zip for use in updating content on instances
             cp "$CIRCLE_ARTIFACTS/content_new.zip" "$CIRCLE_ARTIFACTS/all_content.zip"
-            zip -j "$CIRCLE_ARTIFACTS/all_content.zip" "$CIRCLE_ARTIFACTS/bundle_test/*"
+            unzip -q "$CIRCLE_ARTIFACTS/content_test.zip" -d "$CIRCLE_ARTIFACTS/test_content"
```
yaakovi: change the destination of the temp folder (use CIRCLE_ARTIFACTS only for actual artifacts).
do it in another PR.

avidan-H: what does it matter if I’m deleting it in the same build step anyway?

yaakovi: if step crashed before it removed the folder, it will be posted as an artifact.
don't know what are the effects for so many files as artifacts and i don't want to find out...

avidan-H: legit. It shall be as you say sensei. 
image

PR #5239, opened 2019-12-23
Skipped "attachment" phishing test (inline will still run)

idovandijk:
## Status
Ready

## Description
Phishing v2 attachment test is still running on the old URL which is currently not flagged as malicious - and makes builds fail on timeout.

   - No


idovandijk:
Maya cancelled my build

PR #5243, opened 2019-12-23
Fixed bug with query-file

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20870

## Description
Fixed an issue with `query-file` 

   - No

## Must have
- [ ] Code Review

## Additional changes
Removed unnecessary loop in response

## Technical writer review
- [ ] [CHANGELOG](link)


Shellyber:
@kirbles19 Please review rn

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,10 +1027,22 @@ def query_file_command():
...
+            if 'createdTime' in simple_values:
+                created_time = timestamp_to_datestring(simple_values['createdTime']['values'][0])
+
+            modified_time = None
+            if 'modifiedTime' in simple_values:
+                modified_time = timestamp_to_datestring(simple_values['modifiedTime']['values'][0])
+
+            is_signed = None
+            if 'isSigned' in simple_values:
+                is_signed = simple_values['isSigned']['values'][0]
```
Itay4: you changed the value is `is_signed` here, as the API returns a string, and we return a boolean

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,10 +1027,22 @@ def query_file_command():
             if 'companyName' in simple_values:
                 company_name = simple_values['companyName']['values'][0]
 
+            created_time = None
+            if 'createdTime' in simple_values:
+                created_time = timestamp_to_datestring(simple_values['createdTime']['values'][0])
```
Itay4: lets change the direct key access to `values` to `.get()` and to index `0` by checking the value of `values` is a list

Integrations/Cybereason/Cybereason.py
```diff
@@ -996,11 +996,9 @@ def query_file_command():
         endpoint_outputs = []
         files = data.get('resultIdToElementDataMap')
         for file in files.keys():
```
Itay4:
```suggestion
        for file_ in files.keys():
```

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,10 +1027,22 @@ def query_file_command():
...
+            if 'isSigned' in simple_values:
+                is_signed = simple_values['isSigned']['values'][0]
+
             cybereason_outputs.append({
                 'Name': file_name,
-                'CreationTime': timestamp_to_datestring(simple_values['createdTime']['values'][0]),
-                'ModifiedTime': timestamp_to_datestring(simple_values['modifiedTime']['values'][0]),
+                'CreationTime': created_time,
+                'ModifiedTime': modified_time,
                 'Malicious': files[file]['isMalicious'],
```
Itay4:
```suggestion
                'Malicious': files[file].get('isMalicious'),
```

kirbles19:
@Shellyber Done

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
             if 'companyName' in simple_values:
                 company_name = simple_values['companyName']['values'][0]
 
+            created_time = None
+            if 'createdTime' in simple_values:
+                created_time = timestamp_to_datestring(simple_values.get('createdTime').get('values')[0])
```
Itay4:
```suggestion
                created_time = timestamp_to_datestring(simple_values.get('createdTime', {}).get('values')[0])
```

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
...
+
+            modified_time = None
+            if 'modifiedTime' in simple_values:
+                modified_time = timestamp_to_datestring(simple_values.get('modifiedTime').get('values')[0])
```
Itay4:
```suggestion
                modified_time = timestamp_to_datestring(simple_values.get('modifiedTime', {}).get('values')[0])
```

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
...
+
+            is_signed = None
+            if 'isSigned' in simple_values:
+                is_signed = True if simple_values['isSigned']['values'][0] == 'true' else False
```
Itay4:
```suggestion
                is_signed = True if simple_values.get('isSigned', {}).get('values')[0] == 'true' else False
```

Tests/secrets_white_list.json
```diff
@@ -1189,7 +1189,8 @@
       "walla.com",
       "www.securityadvisor.io",
       "fp.tools",
-      "www.securityadvisor.io"
+      "www.securityadvisor.io",
+      "https://integration.cybereason.net"
```
Itay4:
```suggestion
      "integration.cybereason.net"
```

Integrations/Cybereason/Cybereason_test.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def test_query_file(mocker):
+    mocker.patch.object(demisto, 'params', return_value=params)
+    mocker.patch.object(demisto, 'args', return_value={'file_hash': '4778901e54f55d54435b2626923054a8'})
+    mocker.patch('Cybereason.client_certificate', side_effect=lambda: None, autospec=False)
+    mocker.patch('Cybereason.get_file_machine_details', return_value=MACHINE_OUTPUTS)
+    mocker.patch('Cybereason.query_file', return_value=FILE_OUTPUTS)
+    mocker.patch.object(demisto, 'results')
+    import Cybereason
```
Itay4: remove the import from here, and do it at the top of the code at the import section as follows:
`from Cybereason import query_file_command`

Tests/secrets_white_list.json
Shellyber: @Itay4  Did that at the beginning, kept yelling for secrets until I added the https

Itay4: need to understand why, we shouldnt add the protocol here
@jochman can you have a look please?

jochman: we should not do it after a problem we encountered with excluding domains (we can by accident push email in a@<secret-excluded-domain>.)
But, this change is fine, it's unlikely that there's someone under <name>@integration.cybereason.net.

Integrations/Cybereason/Cybereason_test.py
Shellyber: Can't change that now since I'm mocking this object, so it's won't work

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
...
+            if 'createdTime' in simple_values:
+                created_time = timestamp_to_datestring(simple_values.get('createdTime', {}).get('values')[0])
+
+            modified_time = None
+            if 'modifiedTime' in simple_values:
+                modified_time = timestamp_to_datestring(simple_values.get('modifiedTime', {}).get('values')[0])
+
+            is_signed = None
+            if 'isSigned' in simple_values:
+                is_signed = True if simple_values.get('isSigned', {}).get('values')[0] == 'true' else False
```
ShahafBenYakir: This could result in an exception (if `isSigned` is not in `simple_values` and then `.get('values') ` would return None, which after, using this: `[0]` you are trying to grab the first element of `None` value)
I would suggest doing simethong like this:
```
value_list = demisto.get(simple_values, 'isSigned.values')[0] if  demisto.get(simple_values, 'isSigned.values') else 'false'
is_signed = True if value_list == 'true' else False

```

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
             if 'companyName' in simple_values:
                 company_name = simple_values['companyName']['values'][0]
 
+            created_time = None
+            if 'createdTime' in simple_values:
+                created_time = timestamp_to_datestring(simple_values.get('createdTime', {}).get('values')[0])
```
ShahafBenYakir: What will happen if `createdTime` is not in `simple_values` I think it could result in an exception no?.

Integrations/Cybereason/Cybereason.py
```diff
@@ -1029,11 +1027,23 @@ def query_file_command():
...
+
+            modified_time = None
+            if 'modifiedTime' in simple_values:
+                modified_time = timestamp_to_datestring(simple_values.get('modifiedTime', {}).get('values')[0])
```
ShahafBenYakir: Same exception issue (like the comment at line 1036).

Integrations/Cybereason/Cybereason_test.py
ShahafBenYakir: @Itay4 As @Shellyber said, I don't think we have another option here. Will merge like this if we will find something else will change after.

Integrations/Cybereason/CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
 ## [Unreleased]
-
+Fixed an issue with the ***cybereason-query-file*** command.
```
ShahafBenYakir: Maybe it is worthwhile to add a word on the issue? Say something like:  `Fixed an issue where  ***cybereason-query-file*** command did not pulled specific hashes`
Just so the customer would know what was the issue fixed.

Integrations/Cybereason/Cybereason.py
Shellyber: No need. It will not show this field if its not there.
Don't think an exception is needed

PR #5244, opened 2019-12-23
fix keys in yml files

jochman:
#Status
In progress

   - No

waiting for @roysagi to fix the sdk so the build will pass


ronykoz:
@Itay4 this what's left in order for us to release content

Itay4:
@ronykoz lets discuss in private

PR #5254, opened 2019-12-24
Update demistomock with createIndicators

DeanArbel:
## Status
Ready

## Description
Added `createIndicators` to `demistomock.py`.


yuvalbenshalom:
@DeanArbel just don't merge before the checkout

yuvalbenshalom:
fails because of TestSafeBreach that is un related. force merging.

PR #5272, opened 2019-12-26
Zendesk bug fix

guyfreund:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20876

## Description
Changed the api call

   - No

## Must have
- [ ] Code Review

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/a9cde00fa348022b7ab1e35cecf0e41e3d42242e/Integrations/integration-Zendesk_CHANGELOG.md)

### DIFF
--- a/Integrations/integration-Zendesk.yml
+++ b/Integrations/integration-Zendesk.yml
@@ -25,10 +25,10 @@ configuration:
   required: true
 - display: Use system proxy settings
   name: proxy
-  defaultvalue: "true"
+  defaultvalue: ''
   type: 8
   required: false
-- display: Trust any certificate (unsecure)
+- display: Trust any certificate (not secure)
   name: insecure
   defaultvalue: "false"
   type: 8
@@ -345,7 +345,7 @@ script:
                     email: args.email
                 }
         };
-        res = sendRequest('POST', 'users/create_or_update.json', JSON.stringify(body));
+        res = sendRequest('POST', 'users.json', JSON.stringify(body));
         md = 'Zendesk user added successfuly - ' + args.name + ' (' + args.email + ')\n';
         return ( {'ContentsFormat': formats.json, 'Type': entryTypes.note, 'Contents': res, "HumanReadable": md} );
     };
--- a/Integrations/integration-Zendesk_CHANGELOG.md
+++ b/Integrations/integration-Zendesk_CHANGELOG.md
@@ -1,0 +1,2 @@
+## [Unreleased]
+Fixed an issue where the ***zendesk-add-user*** command didn't return an error when a user is already exists.


guyfreund:
@kirbles19 if there's an issue with the release notes pleace notice and we'll change them.

fvigo:
@guyfreund , wouldn't this break compatibility? 

The previous API call is an "upsert" (create-or-update), this is update only. If playbooks have been built that are based on the "upsert" logic, the change might create unexpected behavior.

guyfreund:
@fvigo - The command is to add a user, not to update it. The command wasn't implemented correctly, if there's a need to update a user we shall implement that as a separate command.
As far as I know, wrong use in a command should not be supported.

fvigo:
@guyfreund that's correct, but the way it was implemented implied that the consumer didn't have to check specific return values to know whether the user already existed or not because the command was mapping the 'create-or-update' API that returns success in either case.

With this change, what is the error that gets returned in case a user already exists? And how is it different from an error that is returned if there is another issue (i.e. API unreachable or API Key invalid).

guyfreund:
@fvigo - if a user already exists the Zendesk API returns the following error: 
image
Which isn't optimal, since the error message is not beautiful, but it does indicates what is the cause of the error. There wasn't error parsing since there's so error parsing across the whole integration (just forwarding the messages returned from the API).

fvigo:
Thanks @guyfreund . I'm still convinced that this would break compatibility in existing Playbooks (it broke mine! :) ) 

Would it be possible at least to add a new command that uses the 'create-or-update' API and still exposes the old behavior?

guyfreund:
@fvigo - Yeah sure
@ShahafBenYakir - Do we to open feature request issue and prioritize?

ShahafBenYakir:
@guyfreund Let's do something like this, instead of changing the path by default we'll add an argument called check_if_user_exists with default set to false. Only If the user sets it to true we will use the new path, otherwise, we will use the old path. This way old users will not experience braking PB.
@fvigo Regarding editing a user with this command, as Guy mentioned, we do not support such actions, so I would not advise using this command for such action. You can open an enhancement request to add a new command and we will take a look at it.

fvigo:
That would be great @ShahafBenYakir . I don't need to update at the moment, but the "add" without error check for duplicates has been used in some of the customer/partner onboarding playbooks from BD and Marketing, so the change you suggested should fix the issue without breaking the playbooks. 

Do you think it is possible to have it applied before the next content release ships? Otherwise they'll break.

Thanks!

Thanks!

guyfreund:
@fvigo - Yes, the fix will be applied before the content release next week.

ShahafBenYakir:
Reopening. As discussed over the PR,  instead of changing the path by default we'll add an argument called check_if_user_exists with the default set to false. Only If the user sets it to true we will use the new path, otherwise, we will use the old path. This way old users will not experience braking PB.

PR #5273, opened 2019-12-26
Fixed ThreatQ TPB

dantavori:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19468

## Description
Improved the TPB flow and added 3 retries to all the ThreatQ tasks.
FYI, still waiting for a response from ThreatQ support regarding the cause of the `500 internal server error`.

   - No


ShahafBenYakir:
@dantavori Mock shouldn't really be a problem as our build also tries to record new responses in cases of failure. I would suggest testing the above on one of our Demo servers (5/4/3/...)

dantavori:
Tested on demo4/5 multiple times, and it worked (can't login to demo3 for some reason). Any other ideas? The last option would be simply removing the upload file part, but it didn't happen before so I think there's a better option

ronykoz:
@dantavori seem to work after @orenzohar s change, please fix the UT and lets merge this

dantavori:
CircleCI build skipped the unit testing, so I tested @orenzohar's changes in the UT locally and it passes.
The fails in the CircleCI tests are:
1. In `Server Master`, not related to ThreatQ
2. In `Two before GA`, I think that v4.1 doesn't support retries there.

@ronykoz

ShahafBenYakir:
@dantavori You can pull from master (the PANW PB is skipped there)

PR #5276, opened 2019-12-26
G vault groups hold bug fix

ShahafBenYakir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20777

## Description
Added email option for the create hold command

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Technical writer review
@kirbles19 
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/92a433548b3337f4930f39c6da0bc255b6b22636/Integrations/GoogleVault/CHANGELOG.md)

Before fix:
#885 Traps test - Work Plan 2019-12-27 12-20-19

After fix:
#885 Traps test - Work Plan 2019-12-27 12-14-32

### DIFF
--- a/Integrations/GoogleVault/GoogleVault.py
+++ b/Integrations/GoogleVault/GoogleVault.py
@@ -202,7 +202,8 @@ def create_hold_query(hold_name, corpus, accounts, terms, time_frame="", start_t
     if mail_query:
         request['query'] = {'mailQuery': mail_query}  # Adding the ready mail query
     for acc_id in accounts:
-        accounts_for_query.append({'accountId': acc_id})
+        acc_entry = {'accountId': acc_id} if '@' not in acc_id else {'email': acc_id}
+        accounts_for_query.append(acc_entry)
     request['accounts'] = accounts_for_query
     return request
 
--- a/TestPlaybooks/playbook-GoogleVault-test.yml
+++ b/TestPlaybooks/playbook-GoogleVault-test.yml
@@ -8,10 +8,10 @@ starttaskid: "0"
 tasks:
   "0":
     id: "0"
-    taskid: 84da32d3-2769-407c-8c01-6593e0ca01ce
+    taskid: 7a2c8c3b-ce55-4853-815b-3c1aed3776f6
     type: start
     task:
-      id: 84da32d3-2769-407c-8c01-6593e0ca01ce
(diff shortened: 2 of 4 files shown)


TestPlaybooks/playbook-GoogleVault-test.yml
```diff
@@ -235,85 +201,19 @@ tasks:
...
-    timertriggers: []
-    ignoreworker: false
-  "14":
-    id: "14"
-    taskid: 40a9a9b8-043e-4b60-8a2e-95947b9f8030
-    type: regular
-    task:
-      id: 40a9a9b8-043e-4b60-8a2e-95947b9f8030
-      version: -1
-      name: Get "First matter" Matter
```
Itay4: why did we remove that one?

Integrations/GoogleVault/GoogleVault.py
```diff
@@ -202,7 +202,8 @@ def create_hold_query(hold_name, corpus, accounts, terms, time_frame="", start_t
     if mail_query:
         request['query'] = {'mailQuery': mail_query}  # Adding the ready mail query
     for acc_id in accounts:
-        accounts_for_query.append({'accountId': acc_id})
+        acc_entry = {'accountId': acc_id} if '@' not in acc_id else {'email': acc_id}
```
Itay4: not too familiar with accounts in google, but are we sure they cannot contain `@`?
if so, the fix is fine, if not, it is better to detect if `acc_id` is actually an email address by using the `emailRegex` we have in common server

ShahafBenYakir: Account IDs can only include numbers so I guess regex was needed here.

TestPlaybooks/playbook-GoogleVault-test.yml
ShahafBenYakir: This test is quite old so it wasn't so optimal. For example, it had hardcoded IDs that are no longer exist making it fail. I had to change several flows. :/

ShahafBenYakir:
Merging.
@kirbles19 if you find issues with the release notes let us know and we will fix them.

kirbles19:
@ShahafBenYakir What was actually added to support this? New argument? New parameter? How should we update docs?

ShahafBenYakir:
We added the ability to support emails in addition to account IDs (like in every other google product).

PR #5278, opened 2019-12-29
slack notifier for container 1

Shellyber:
## Status
Ready

## Description
Removed an If statement to allow `slack-notifier` for unit tests(container 1)

 - No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -387,6 +387,11 @@ jobs:
               echo "Skipping instance tests for forked PRs"
               exit 0
             fi
+            if [ -n "${INSTANCE_TESTS}" ];
+              then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -387,11 +387,6 @@ jobs:
               echo "Skipping instance tests for forked PRs"
               exit 0
             fi
-            if [ -n "${INSTANCE_TESTS}" ] || [ $CIRCLE_NODE_INDEX -ne 0 ] ;
```
ronykoz: maybe we should keep the INSTANCE_TESTS section?

Shellyber: Right, Fixed

PR #5282, opened 2019-12-29
WF-v2 refactor of Not found for report and sample

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20310

## Description
added a test case for webpage as well
No change of docs/RN

## Screenshots
Screen Shot 2019-12-29 at 12 59 59

4.0.0

   - No

## Must have
- [x] Tests
- [x] Documentation - no cahnge
- [ ] Code Review


Integrations/Palo_Alto_Networks_WildFire_v2/Palo_Alto_Networks_WildFire_v2.py
```diff
@@ -707,8 +714,13 @@ def wildfire_get_sample_command():
     inputs = hash_args_handler(sha256, md5)
 
     for element in inputs:
-        result = wildfire_get_sample(element)
-
+        result, not_found = wildfire_get_sample(element)
```
yuvalbenshalom: seems you do not use `result` when `not_found` is true.
so you dont need to catch `NotFoundError` in `wildfire_get_sample` and no need to pass not_found, but just catch it here and return the message hen you catch it.

Integrations/Palo_Alto_Networks_WildFire_v2/Palo_Alto_Networks_WildFire_v2.py
```diff
@@ -707,8 +714,13 @@ def wildfire_get_sample_command():
...
+        if not_found:
+            demisto.results(
+                'Sample was not found. '
+                'Please note that grayware and benign samples are available for 14 days only. '
+                'For more info contact your WildFire representative.')
+            sys.exit(0)
```
yuvalbenshalom: I think you can just return

Integrations/Palo_Alto_Networks_WildFire_v2/Palo_Alto_Networks_WildFire_v2.py
```diff
@@ -620,17 +616,22 @@ def wildfire_get_report(file_hash):
...
     hash_type = 'SHA256' if sha256Regex.match(file_hash) else 'MD5'
     entry_context = {hash_type: file_hash}
-
-    if not json_res:
+    if not_found or not json_res:
         entry_context['Status'] = 'NotFound'
         demisto.results({
             'Type': entryTypes['note'],
             'HumanReadable': 'Report not found.',
+            'Contents': json_res,
```
yuvalbenshalom: check what happens when `json_res` is `None`.

PR #5300, opened 2019-12-30
[Bug Fix] QRadar fetch incidents having issues with newlines

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20960

## Description
Removed newlines from incident names for incidents with newlines in their description.

   - No

## Must have
- [x] Tests
- [x] Code Review

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/edit/qradar-fetch-fix/Integrations/QRadar/CHANGELOG.md?pr=%2Fdemisto%2Fcontent%2Fpull%2F5300)


Integrations/QRadar/QRadar.py
```diff
@@ -525,8 +525,9 @@ def create_incident_from_offense(offense):
     labels = []
     for i in range(len(keys)):
         labels.append({'type': keys[i], 'value': convert_to_str(offense[keys[i]])})
+    formatted_description = offense['description'].replace('\n', '') if offense['description'] else ''
```
yuvalbenshalom:
```suggestion
    formatted_description = offense['description'].replace('\s\n', ' ').replace('\n', ' ') if offense['description'] else ''
```

this way you will replace the new lines with space and will concat words(or create double spaces)
what do you think?

DeanArbel: Good idea, this should work but with `re.sub(r'\s\n', ' ', s)`. I'll commit that change.

PR #5302, opened 2019-12-30
Removed riskscore duplicate incident field Extrahop

jochman:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20972
related: https://github.com/demisto/etc/issues/20965

## Description
Duplicate incident field from 19.11.0 release removed and integrated into one.

5.0.0

   - Yes
       - Further details: It'll remove an incident field from 4.5.0 and ExtraHop type was `number`, not it's `shortText` as in PrismaCloud. There is no playbook that uses this field in the content repo.

## Must have
- [ ] Code Review


anara123:
@michalgold @jochman  riskscore should be shortText or number by design?

jochman:
> @michalgold @jochman riskscore should be shortText or number by design?

I've consulted @mayagoldb and she said it should be shortText.
But because it's not in use in any of our playbooks, we can change it to number.

mayagoldb:
@jochman @anara123 The field is used in Prisma Cloud classification (as shortText), but I don't think changing to number should affect the classifier.

Dan-at-Extrahop:
@anara123 @jochman @mayagoldb changing this to shortText from number ended up being a breaking change for the ExtraHop integration since we're using the REST API directly to push in incidents.  Is it possible to change it back to type number?

FYI REST API error:
`{"id":"incCreateErr","status":400,"title":"Failed creating incident","detail":"Failed creating incident","error":"Cannot convert type float64 of '33' to type string, field: riskscore (8902)","encrypted":false,"multires":null}`

Furthermore, if any of the ExtraHop incident field types are changed this will result in our integration breaking, so how can we avoid this type of change in the future? Thanks

anara123:
@Dan-at-Extrahop I suggest to create fields with ExtraHop prefix. So in this case `ExtraHop-RiskScore` and make it of type number.
But in general we try to avoid changes like this as much as we can. But this case we had to.

PR #5305, opened 2019-12-30
Refactored job create command

guykeller:
## Status
Ready

## Description
Refactored Splunk's job-create command for improved readability.

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Integrations/SplunkPy/SplunkPy.py
+++ b/Integrations/SplunkPy/SplunkPy.py
@@ -345,16 +345,23 @@ def splunk_search_command():
 
 
 def splunk_job_create_command():
-    searchquery_normal = demisto.args()['query']
-    if not searchquery_normal.startswith('search'):
-        searchquery_normal = 'search ' + searchquery_normal
-    kwargs_normalsearch = {"exec_mode": "normal"}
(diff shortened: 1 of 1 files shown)


Integrations/SplunkPy/SplunkPy.py
```diff
@@ -345,16 +345,23 @@ def splunk_search_command():
...
-                     "Contents": "Splunk Job created with SID: " + job.sid, "EntryContext": ec})
+    entry_context = {
+        'Splunk.Job': search_job.sid
+    }
+    demisto.results({
+        "Type": 1,
+        "ContentsFormat": formats['text'],
+        "Contents": "Splunk Job created with SID: " + search_job.sid,
+        "EntryContext": entry_context
+    })
```
ronykoz: lets use return_outputs already, will be easier to test later on.

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -345,16 +345,23 @@ def splunk_search_command():
...
     sys.exit(0)
```
ronykoz: remove this, from everywhere in the code

Integrations/SplunkPy/SplunkPy.py
ronykoz: ?

PR #5310, opened 2019-12-31
Fixed access to non existing key

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20978

## Description
When fetching incidents from an index that isn't notable, a KeyError was sometimes thrown if the index does not include a "_time" field.
This issue is now fixed.

   - No

## Must have
- [ ] Tests
- [ ] Code Review

### DIFF
--- a/Integrations/SplunkPy/SplunkPy.py
+++ b/Integrations/SplunkPy/SplunkPy.py
@@ -184,7 +184,10 @@ def notable_to_incident(event):
         incident["severity"] = severity_to_level(event['urgency'])
     if demisto.get(event, 'rule_description'):
         incident["details"] = event["rule_description"]
-    incident["occurred"] = event["_time"]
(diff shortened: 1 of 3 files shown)


ronykoz:
@roysagi please go over the RN when you have the time(Not urgent)

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -184,7 +184,10 @@ def notable_to_incident(event):
         incident["severity"] = severity_to_level(event['urgency'])
     if demisto.get(event, 'rule_description'):
         incident["details"] = event["rule_description"]
-    incident["occurred"] = event["_time"]
+    if demisto.get(event, "occurred"):
```
ronykoz: just saw that
```suggestion
    if demisto.get(event, "_time"):
```

PR #5323, opened 2020-01-01
Removed Archer from skipped

guykeller:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20335

## Description
Removing Archer from skipped since our instance was fixed.

   - No


ShahafBenYakir:
Did we change something that should've solves this?

guykeller:
@ShahafBenYakir Yes, went on a call with Archer support

PR #5337, opened 2020-01-02
zendesk fix

guyfreund:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20876

## Description
Fixing the issue without breaking BC

   - No

## Must have
- [x] Code Review

## Technical Writer Review
- [x] [YML](https://github.com/demisto/content/blob/c1b87393ce483d4641e478a41c373c06e6cf831a/Integrations/integration-Zendesk.yml) - description in lines 786-787 needs to be reviewed


kirbles19:
@guyfreund Done. Why is there no changelog.md ?

guyfreund:
@kirbles - There’s no chanelog since this bug has been reopened and logs for this bug are already in the changelog file. When I’ll resolve the conflicts it’ll appear.

guyfreund:
Since we don't have an instance for development, the fix was tested on zendesk production env.
As can see down below, the fix solves the issue and doesn't break BC.
image

Integrations/integration-Zendesk_CHANGELOG.md
```diff
@@ -1,2 +1,5 @@
 ## [Unreleased]
- -
\ No newline at end of file
+-
+
+## [19.12.1] - 2019-12-25
+Fixed an issue where the ***zendesk-add-user*** command didn't return an error when a user is already exists.
```
ShahafBenYakir: Why is this on previous release?

guyfreund:
@kirbles19 - Please review changelog

kirbles19:
@guyfreund Done.

kirbles19:
@guyfreund @ShahafBenYakir We have no documentation for this integration at all. Please generate documentation (MD) for the entire integration.

guyfreund:
@kirbles19 Please open a documentation issue.

kirbles19:
I thought all docs are supposed to be part of the integration package.

ShahafBenYakir:
@kirbles19 It is, but in this case, as you mentioned, no docs yet exist and thus someone should write them from scratch i.e. a new task which is not related to this bug fix made by @guyfreund.

kirbles19:
@ShahafBenYakir Actually, technically, there was an addition made which requires updates to documentation, therefore docs are definitely related to the PR. Docs should've been generated without me having to request it.

guyfreund:
@kirbles19 If the docs already existed, then this PR is supposed also to update the docs. But since they do not exist, the docs is another issue that has to be done from scratch, therefore has to prioritized.

kirbles19:
@guyfreund So create the docs, and i'll create an issue for it. It's a high priority, if the integration was important enough to update, it's important enough to have documentation for.

What's the ETA?

PR #5339, opened 2020-01-04
Update sdk version

ronykoz:
## Status
Ready

Updating the build to fully use the demisto-sdk

### DIFF
--- a/dev-requirements-py3.txt
+++ b/dev-requirements-py3.txt
@@ -1,5 +1,5 @@
 flake8==3.7.8
-git+https://github.com/demisto/demisto-sdk.git@fix-commands
+demisto-sdk>=0.3.0
 demisto-py>=2.0.7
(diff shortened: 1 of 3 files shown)


ronykoz:
@yaakovi please take a look at the `Run Unit Testing and Lint` not sure why i need to pass the --dir option while on my sdk branch i removed it as required and it works without it locally

dev-requirements-py3.txt
```diff
@@ -1,6 +1,6 @@
 flake8==3.7.8
-demisto-sdk==0.2.6
-demisto-py==2.0.6
+git+https://github.com/demisto/demisto-sdk.git@fix-commands
```
yaakovi: don't forget to update this after the sdk release

ronykoz:
Force merging the SDK version update

PR #5353, opened 2020-01-05
added-regex-cider

jochman:
## Status
Ready

## Description
added regexes for ipv4 and ipv6 with CIDR.

   - No

## Must have
- [X] Tests
- [x] Code Review


yuvalbenshalom:
@jochman UT fail, please check

yuvalbenshalom:
force merging: this is an addition an there is no code dependant on it, plus UT passed.
No point to wait for the unrelated palaybook tests(that triggered by change in common server) to end.

PR #5359, opened 2020-01-06
Hello world to pack

anara123:
## Status
Done

## Related Issues
fixes: https://github.com/demisto/etc/issues/21072

## Description
Move Hello World example to pack format


anara123:
@yaakovi the build fails because of a bug in demisto-sdk.

anara123:
we came to conclusion that it will only confuse contributors. so we keep just single version

PR #5368, opened 2020-01-06
Add indicatortype enum to commonsp

jochman:
## Status
Ready

## Description
added enum of indicatortypes to commonsp

   - No

## Must have
- [ ] Code Review


jochman:
@bakatzir

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -102,6 +102,30 @@
...
+    MD5 = "File MD5"
+    SHA1 = "File SHA-1"
+    Host = "Host"
+    IP = "IP"
+    CIDR = "CIDR"
+    IPv6 = "IPv6"
+    IPv6CIDR = "IPv6CIDR"
+    Registry = "Registry Key"
+    SSDeep = "ssdeep"
+    URL = "URL"
```
avidan-H: Why is this a class?

avidan-H: just so you can call it like `IndicatorTypes.Account`?

bakatzir:
reference: https://github.com/demisto/content/pull/5260

jochman:
@avi

> Just a unit-test as we discussed that make sure the enum mapping is current with the indicators' `name` to `details` fields values and then you're good.

This test can't be done in CommonsServerPython tests as it has no connection to the global repo. I'm trying to add its functionality to Demisto-SDK. meanwhile, we can merge this pr.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -102,6 +102,31 @@
     'ctph': 'file',
     'ssdeep': 'file'
 }
+
+
+class IndicatorTypes(object):
```
anara123: I think maybe you should change the name of your class to `FeedIndicatorType` and I will change to `DBotScoreIndicatorType` this way it will be clear which class used where.

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -124,6 +124,27 @@ class IndicatorTypes(object):
     SSDeep = "ssdeep"
     URL = "URL"
 
+    @classmethod
+    def is_valid_type(cls, _type):
+        return _type in (
+            cls.Account,
```
anara123: it is not a class method. change it to be staticmethod

PR #5374, opened 2020-01-07
Remove astrix in reputation TPB to allow content download repo in windows

bakatzir:
@David-BMS FYI


bakatzir:
@avidan-H Test name is per PB id, which have stayed the same.

bakatzir:
@avidan-H Test name is per PB id, which have stayed the same.
`tests.add('Indicators reputation-.json Test')`

PR #5380, opened 2020-01-07
adding autofocus-get-export-list-indicators command

roysagi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21093

## Description
Adding the command: autofocus-get-export-list-indicators
The command gets an export list name (called "label" in AutoFocus) processes them to fit the Indicator "scheme" of Type and Value and returns ExportListIndocator elements to context.

The implementation is based on the former AutoFocus miner seen here: https://github.com/PaloAltoNetworks/minemeld-core/blob/master/minemeld/ft/autofocus.py

We can also add additional arguments as seen here: https://docs.paloaltonetworks.com/autofocus/autofocus-api/perform-direct-searches/export-list#id1b0b954a-f74d-462d-805b-8916455bf7b0_id259466ba-307f-42b0-8e1d-f99067576c84
But other then the time ones I don't think we really want to include them

The indicator type resolution is with accordance to: https://github.com/demisto/content/pull/5427

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Technical writer review
- [ ] [YAML file](https://github.com/demisto/content/blob/autofocus_add_command/Integrations/AutofocusV2/AutofocusV2.yml)
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/autofocus_add_command/Integrations/AutofocusV2/CHANGELOG.md)


Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -22,6 +24,8 @@
 
 VENDOR_NAME = 'AutoFocus V2'
 
+DOMAIN_RE = re.compile(r'^[a-zA-Z\d-]{,63}(\.[a-zA-Z\d-]{,63})*$')
```
ronykoz: we have that in commonserverpython

Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -1234,6 +1298,29 @@ def search_file_command(file):
     get_indicator_outputs(indicator_type, indicator_details, 'SHA256')
 
 
+def get_export_list_command(args):
+    data = {
+        'label': args.get('label'),
+        'panosFormatted': True,
```
ronykoz: add an explanation for this and why don;t you need an api key

Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -1234,6 +1298,29 @@ def search_file_command(file):
...
+    for indicator_value in results.get('export_list'):
+        indicator_type = type_of_indicator(indicator_value)
+        indicators.append({
+            'Type': indicator_type,
+            'Value': indicator_value,
+        })
+
+    hr = tableToMarkdown(f"Export list {args.get('label')}", indicators, headers=['Type', 'Value'])
+
+    return_outputs(hr, {'AutoFocus.ExportListIndicator(val.Value == obj.Value && val.Type == obj.Type)': indicators},
```
ronykoz: why under ExportListIndicator

Integrations/AutofocusV2/AutofocusV2.yml
```diff
@@ -969,6 +969,24 @@ script:
...
+      description: The label of the exported list.
+      isArray: false
+      name: label
+      required: true
+      secret: false
+    deprecated: false
+    description: Gets export list indicators from AutoFocus.
+    execution: false
+    name: autofocus-get-export-list-indicators
+    outputs:
```
ronykoz: This is all the data we need here?
@Arsenikr take a look at this pls

Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -973,6 +977,66 @@ def get_indicator_outputs(indicator_type, indicators, indicator_context_output):
     return_outputs(readable_output=human_readable, outputs=ec, raw_response=raw_res)
 
 
+def check_for_ip(indicator):
```
ronykoz: shouldn't it be in commonserverpython?

Integrations/AutofocusV2/AutofocusV2.py
roysagi: I can't find it there - are you sure?

Integrations/AutofocusV2/AutofocusV2.py
roysagi: we can add this... but it might be specific to this implementation (it is taken from the miner itself with some minor modifications)

Integrations/AutofocusV2/AutofocusV2.py
roysagi: No api key needed here as it is added in the http_request command.
added explanation

Integrations/AutofocusV2/AutofocusV2.yml
roysagi: this is all the info given in the miner... (the other given is the api-key but that is given to the integration itself)

Integrations/AutofocusV2/AutofocusV2.py
roysagi: simply Indicator? I wanted to differentiate this from other indicators

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: My bad, we should add it there then @reutshal what is the most updated regex that we have for domain?

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: i think that we should have something for that in the commonserverpython, @bakatzir maybe you already done something for this?

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: so maybe we should remove this?

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: @avidan-H do we have a convention for this?

Integrations/AutofocusV2/AutofocusV2.py
bakatzir: +1
I haven't since office365 can also have urls so it's a bit different.
but we should implement it in commonserverpython using the enums from this pr:
https://github.com/demisto/content/pull/5368

Integrations/AutofocusV2/AutofocusV2.py
reutshal: We don't actually have a regex for domains because it is too general and can match to non-domain texts (every text in the form  _._ )
in the indicator extraction, we use url or email regex and then extracting the domain from the result.

ronykoz: My guess is that we need to do something like that here too

Integrations/AutofocusV2/AutofocusV2.py
avidan-H: No we don't. We should discuss if it makes sense to establish one. Also, shouldn't output entries for each of the indicator entities (different for each indicator type) be included in the outputs? https://github.com/demisto/content-docs/blob/master/docs/context-standards.md

Integrations/AutofocusV2/AutofocusV2.py
avidan-H: While I think we should definitely move something like this to commonserverpython, this uses the `ipaddress` module which comes only with python3. If we move to commonserverpython we should make sure it's 2&3 compatible.

Integrations/AutofocusV2/AutofocusV2.py
roysagi: when I tried to run without this line it raised an error... I'll try again...

Integrations/AutofocusV2/AutofocusV2.py
roysagi: Well the command literally returns a list of indicators with no additional info like you can see on the context standards. 

You can see an output example here: https://docs.paloaltonetworks.com/autofocus/autofocus-api/perform-direct-searches/export-list#id1b0b954a-f74d-462d-805b-8916455bf7b0_id259466ba-307f-42b0-8e1d-f99067576c84

Also considering this is supposed to be "minerish" shouldn't be up to the standards set to other miners 
(type: *, value: *)?

Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -973,6 +977,66 @@ def get_indicator_outputs(indicator_type, indicators, indicator_context_output):
...
+    if '-' in indicator:
+        # check for address range
+        a1, a2 = indicator.split('-', 1)
```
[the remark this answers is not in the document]
roysagi: taken from the autofocus miner - Im still trying to create an IP list...not as simple as it sounds...
Once done I'll test it.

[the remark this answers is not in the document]
roysagi: image
seems this is how ip indicators come form autofocus - explaining the split

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: @roysagi ?

Integrations/AutofocusV2/AutofocusV2.py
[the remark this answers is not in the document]
ronykoz: it seems like CIDR to me

Integrations/AutofocusV2/AutofocusV2.py
ronykoz: You can and need to add whatever you can get from the product and populate the relevant fields from the context standards.
like ip.address

Integrations/AutofocusV2/AutofocusV2.py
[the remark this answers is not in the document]
roysagi: isn't CIDR addresses that are of the form: X.X.X.X/Y  ?

Integrations/AutofocusV2/AutofocusV2.py
roysagi: done

roysagi: done

roysagi:
For prosperity:
Also mentioned in this issue: https://github.com/demisto/etc/issues/21174

ronykoz:
@kirbles19 could you please review the yml file 
https://github.com/demisto/content/blob/autofocus_add_command/Integrations/AutofocusV2/AutofocusV2.yml

Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -1234,6 +1319,91 @@ def search_file_command(file):
...
+                resolved_address = resolve_ip_address(indicator_value.split(":", 1)[0])
+                semicolon_suffix = indicator_value.split(":", 1)[1]
+                slash_suffix = None
+
+            else:
+                resolved_address = resolve_ip_address(indicator_value.split("/", 1)[0])
+                slash_suffix = indicator_value.split("/", 1)[1]
+                semicolon_suffix = None
+
+            if resolved_address:
```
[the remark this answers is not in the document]
roysagi: no this is correct - you want to see if an address was resolved and if so create the new value

Integrations/AutofocusV2/AutofocusV2.yml
```diff
@@ -969,6 +969,36 @@ script:
...
+    - contextPath: AutoFocus.ExportListIndicator.Type
+      description: 'The type of the indicator in the export list. '
```
kirbles19: The indicator type in the export list.

Integrations/AutofocusV2/AutofocusV2.yml
```diff
@@ -969,6 +969,36 @@ script:
...
+      description: The value of the indicator in the export list.
+      type: String
+    - contextPath: IP.Address
+      description: The IP address.
+      type: String
+    - contextPath: URL.Data
+      description: The URL address.
+      type: String
+    - contextPath: File.SHA256
+      description: SHA256 hash of the file.
```
kirbles19: The SHA256 hash of the file.

kirbles19:
@roysagi Done, comments inline.

Integrations/AutofocusV2/AutofocusV2.yml
ronykoz:
```suggestion
      description: 'The indicator type in the export list. '
```

Integrations/AutofocusV2/AutofocusV2.yml
ronykoz:
```suggestion
      description: The SHA256 hash of the file.
```

PR #5393, opened 2020-01-08
Script for setting that can handle a null value in the context path.

David-BMS:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/21061

## Description
This script handles setting to context even when no value exists in the value arg.
This is useful when trying to set results from searches where its unknown if there's a value in the returned query fields. This fixes the existing issue with the !set command that provides an error when setting an empty value and is a better solution than creating conditional tasks or continue on error

## Screenshots

x.x.x

   - Yes
       - Further details:
   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Additional changes

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
```diff
@@ -0,0 +1,41 @@
+commonfields:
+  id: SetAndHandleEmpty
+  version: -1
+name: SetAndHandleEmpty
+fromversion: 4.5.0
+type: python
+tags:
+- Utility
+comment: Checks if value exists in context. If the value exists it will be set in context,
```
kirbles19: Checks if the specified value exists in context. If the value exists, it will be set in context, otherwise no value will be set it context.

@David-BMS Is this still technically accurate?

Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
```diff
@@ -0,0 +1,41 @@
...
+  if not nothing will be set.
+enabled: true
+args:
+- name: key
+  required: true
+  default: true
+  description: The key to set
```
kirbles19: The key to set in context.

Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
```diff
@@ -0,0 +1,41 @@
...
+- name: value
+  description: The value to set to the key. Can be an array. Usually, a dq expression.
```
kirbles19: The value of the key to set in context. Can be an array. The value is usually a DQ expression.

Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
```diff
@@ -0,0 +1,41 @@
...
+  isArray: true
+- name: append
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: If false then then the context key will be overwritten. If set to true
```
kirbles19: Whether to append the new context key to the existing context key. If "false", then the existing context key will be overwritten with the new context key.

Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
```diff
@@ -0,0 +1,41 @@
...
+    then the script will append to existing context key
+- name: stringify
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: Whether the argument should be saved as a string.
```
kirbles19: Whether to save the argument as a string. The default value is "false".

kirbles19:
@David-BMS Done. Shouldn't there be a changelog file?

Scripts/SetAndHandleEmpty/SetAndHandleEmpty.yml
David-BMS: It sounds confusing but the intention is that the script checks for a content value in one location and if it exists sets it in another location which is needed for the context outputs

David-BMS:
@kirbles19  Added changelog

PR #5417, opened 2020-01-12
SplunkPy new command

Shellyber:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19657

   - No

## Must have
- [ ] Tests
- [ ] Code Review

## Dependencies
- [ ] Dependency 1
- [ ] Dependency 2
- [ ] Dependency 3

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)


Shellyber:
@kirbles19 Please review yml and changelog file.
Added one command - `splunk-submit-event-hec`

kirbles19:
@Shellyber Done.

Integrations/SplunkPy/SplunkPy.yml
```diff
@@ -259,8 +281,61 @@ script:
...
+        JSON example: {"event": {"message":"Access log test message"}}
+        String example: "event": "Access log test message."
+      isArray: false
+      name: event
+      required: true
+      secret: false
+    - default: false
+      description: Fields for indexing that do not occur in the event payload itself.
+      isArray: false
+      name: fields
```
ShahafBenYakir: Is it CSV list of fields?

Integrations/SplunkPy/SplunkPy.yml
```diff
@@ -259,8 +281,61 @@ script:
...
+  - arguments:
+    - default: false
+      description: |-
+        Event payload key-value. Value can be a string or a JSON object.
+        JSON example: {"event": {"message":"Access log test message"}}
+        String example: "event": "Access log test message."
+      isArray: false
+      name: event
+      required: true
+      secret: false
```
ShahafBenYakir: By this description, it seems like a valid run of the command could be: `!splunk-submit-event-hec {"event": {"message":"Access log test message"}}`. It is not correct right?

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -489,6 +489,60 @@ def splunk_submit_event_command():
...
+
+    response = requests.post(baseurl + '/services/collector/event', data=json.dumps(args), headers=headers,
+                             verify=VERIFY_CERTIFICATE)
+    return response
+
+
+def splunk_submit_event_hec_command():
+
+    hec_token = demisto.params().get('hec_token')
+    hec_port = demisto.params().get('hec_port')
```
ShahafBenYakir: In order to prevent connection exceptions let's also make sure the port is also defined```
```
if not hec_port:
    raise Exception(...)
```

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -489,6 +489,60 @@ def splunk_submit_event_command():
...
+    if host:
+        args['host'] = host
+    if index:
+        args['index'] = index
+    if source_type:
+        args['sourcetype'] = source_type
+    if source:
+        args['source'] = source
+    if time_:
+        args['time'] = time_
```
ShahafBenYakir: Why not use `assign_params` from commonServerPython instead? it does exactly this:
```
assign_params(a='1', b=True, c=None, d='')

>>>   {'a': '1', 'b': True}
```

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -489,6 +489,60 @@ def splunk_submit_event_command():
...
+
+    headers = {
+        'Authorization': 'Splunk %s' % hec_token,
```
ShahafBenYakir: Is there a special reason to use `%` and not `.format()` here?

Shellyber: No special reason.. it's just the way it was written in the integration so far.
Will change that

Integrations/SplunkPy/SplunkPy.yml
Shellyber: According to the docs, you can send it both ways

Integrations/SplunkPy/SplunkPy.yml
Shellyber: No. any field can come here. 
It will add it to the event as the following:
image

ShahafBenYakir: How do the user expected to pass the list of fields.

Shellyber: Comma separated. will update the description and add an example to the docs afterwards

PR #5423, opened 2020-01-13
Fixed bizarre playbook locations

idovandijk:
## Status
Ready
Pushed without hooks due to https://github.com/demisto/etc/issues/21203

## Related Issues
fixes: https://github.com/demisto/etc/issues/21191

## Description
Fixes an issue where the playbook tasks where not centered when the playbook or workplan were opened. The tasks were sometimes not even visible on the screen unless you scrolled to them.

5.0.0

   - No


kirbles19:
@idovandijk Done

PR #5442, opened 2020-01-15
update indicatorfields name

ronykoz:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21253

## Description
Updating indicatorField names due to server issues


glicht:
@kirbles19 see release notes. they are all the same.

Misc/reputation-registryKey.json
```diff
@@ -24,7 +24,7 @@
     "contextValue": "",
     "excludedBrands": [],
     "defaultMapping": {
-        "registrykeyname": {
+        "namefield": {
```
Bars92: is that on purpose?

Bars92:
I think it should be fine. Always mind the ids and the cli name should also match. since we do not want to (temporarily) remove the system fields we must add them to content

kirbles19:
@glicht Do we want to add section to the top of the RN: Breaking Changes. We can add this there?

glicht:
We will need to add to the top too.

IndicatorFields/incidentfield-id.json
```diff
@@ -11,8 +11,8 @@
   "script": "",
   "isReadOnly": false,
   "commitMessage": "",
-  "system": false,
-  "content": true,
+  "system": true,
```
Bars92: remove this file

Misc/reputation-registryKey.json
jochman: yeah, it's now pointing to `IndicatorFields/incidentfield-namefield.json`

IndicatorFields/incidentfield-indicatorid.json
```diff
@@ -1,11 +1,11 @@
 {
   "sortValues": null,
   "closeForm": false,
-  "cliName": "id",
+  "cliName": "indicatoridentification",
```
Bars92: change the file name

PR #5443, opened 2020-01-15
Add more efficient integer count function.

amshamah419:
## Status
Ready

## Description
`add()` wasn't slow enough. We have fixed that for now.

### DIFF
--- a/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Scripts/CommonServerPython/CommonServerPython.py
@@ -15,6 +15,7 @@ import logging
 from collections import OrderedDict
 import xml.etree.cElementTree as ET
 from datetime import datetime, timedelta
+from time import sleep, time
 
 import demistomock as demisto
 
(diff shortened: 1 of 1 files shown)


Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -166,6 +167,23 @@ def is_valid_type(_type):
...
+
+    Known limitations:
+     - The sum of both numbers cannot exceed 300.
+     - May take some time to count.
+    :param num1: First number.
+    :param num2: Second number.
+    :return: result of two numbers
+    """
+    start = time()
+    sleep(num1)
```
yardensade:
```suggestion
    sleep(num1*num2)
```

Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -166,6 +167,23 @@ def is_valid_type(_type):
 # ====================================================================================
 
 
+def add(num1, num2):
```
jochman:
```suggestion
def add(nice1, nice2):
```

Scripts/CommonServerPython/CommonServerPython.py
amshamah419: Will add multiply function to address this usecase

PR #5444, opened 2020-01-15
Unittest slack notifier

ShahafBenYakir:
## Status
Ready

## Description
Separated unit test slack messages to a different message


ShahafBenYakir:
@yaakovi This could only be merged after you release the SDK update

ShahafBenYakir:
@yaakovi can you approve this? Now that the new SDK version is released we can merge this to get the unit test changes.

Tests/scripts/download_demisto_conf.sh
```diff
@@ -45,9 +45,11 @@ if [ "$?" != "0" ]; then
...
     rm -rf ./test_configuration.zip
+    echo "ERROR: Found a branch with the same name in contest-test-conf conf.json - $UNDERSCORE_CIRCLE_BRANCH.\n Merge it in order to merge the current branch into content repo."
+    exit 1
+
 fi
 
 set -e
-echo "using instance:"
-cat instance.json
+rm instance.json
```
yaakovi: remove this line

Tests/scripts/slack_notifier.py
```diff
@@ -149,18 +170,21 @@ def get_fields(build_number):
...
+def slack_notifier(build_url, slack_token, env_results_file_name, container):
     branches = run_command("git branch")
     branch_name_reg = re.search(r'\* (.*)', branches)
     branch_name = branch_name_reg.group(1)
 
     if branch_name == 'master':
         print_color("Starting Slack notifications about nightly build", LOG_COLORS.GREEN)
         print("Extracting build status")
-        content_team_attachments, _ = get_attachments(build_url, env_results_file_name)
+        if container:
```
yaakovi: add comments
container 0: test playbooks
container 1: unit tests

Tests/scripts/download_demisto_conf.sh
ShahafBenYakir: Why? We said we don't want it

yaakovi: we are not using the instance so the print message shouldn't be there.
but no need to delete it.
instead, you can remove lines 39-45

PR #5453, opened 2020-01-16
pan-os-dag out of skipped, moved to be first

guyfreund:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19205

## Description
Out of skipped tests, moved to be first test running, needs to check if PB fails on nightly

   - No

## Must have
- [ ] Code Review

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -3,6 +3,12 @@
     "testInterval": 20,
     "tests": [
         {
+            "playbookID": "PAN-OS DAG Configuration Test",
+            "integrations": "Panorama",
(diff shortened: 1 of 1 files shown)


guyfreund:
@Arsenikr - what is the status with this PR? Can we take it out of skipped?

Arsenikr:
keep it in skipped, waiting for this to be fixed so the build will work: https://github.com/demisto/etc/issues/21495

it happens on panorama instances with no address groups etc

guyfreund:
So closing this PR.

PR #5458, opened 2020-01-19
[Fix] Bypass exclusion list for TAXII and Office365 feeds

DeanArbel:
## Status
Ready

## Description
Added the UI fix to TAXII and Office365 feed integrations implemented for bypass exclusion list yaml field.

## Screenshots
image

5.5.0

   - No


DeanArbel:
> Shouldn't you remove `skipExclusionList` in the Office365 yml?

You're right. Replaced `skipExclusionList` with `bypassExclusionList`

PR #5468, opened 2020-01-20
My py fixes

yaakovi:
## Status
Ready

## Description
update build to work with a specific mypy version instead of latest.
made some recommended changes to integration from the nightly build (with the newer version of mypy)

   - No


dev-requirements-py3.txt
```diff
@@ -1,12 +1,12 @@
 flake8>=3.7.8
-demisto-sdk>=0.3.3
-demisto-py>=2.0.7
+demisto-sdk==0.3.3
+demisto-py==2.0.8
```
liorblob: not sure about these... @ronykoz what do you think?

yaakovi: 2.0.8 introduced updates regarding ssl certificate, it's not mandatory to increase the version but shouldn't cause any issues.

liorblob: We might want the SDK to be always latest

yaakovi: we want the option to release a new SDK version but keep working on a stable version for a few more days. (wouldn't like to release a new sdk in middle of content release and break the build).

liorblob: every SDK version is stable!

PR #5473, opened 2020-01-20
Update BaseClient _http_request Function

avidan-H:
## Status
Ready

## Description
Fixes the issue if the user wanted to provide a `full_url` instead of using the `base_url` together with a `url_suffix` when calling the BaseClient's `_http_request` method that the user still needed to provide a value for the `url_suffix` argument.

any

   - No

## Must have
- [x] Tests
- [ ] Code Review

## Dependencies
- [ ] Everything

## Additional changes
Added a unit-test first. You can see it fail in the first commit. Passes in the second.

## Technical writer review
- [ ] [YAML file](link)
- [ ] [CHANGELOG](link)

### DIFF
--- a/Scripts/CommonServerPython/CommonServerPython_test.py
+++ b/Scripts/CommonServerPython/CommonServerPython_test.py
@@ -907,6 +907,11 @@ class TestBaseClient:
         response.status_code = 400
         assert not self.client._is_status_code_valid(response)
 
+    def test_http_request_full_url_no_url_suffix(self, requests_mock):
+        requests_mock.get('http://fullurl.com/without/url/suffix', text=json.dumps(self.text))
+        res = self.client._http_request('get', full_url='http://fullurl.com/without/url/suffix')
+        assert res == self.text
+
 
 def test_parse_date_string():
     # test unconverted data remains: Z
--- a/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Scripts/CommonServerPython/CommonServerPython.py
@@ -2315,7 +2315,7 @@ if 'requests' in sys.modules:
             if not proxy:
                 self._session.trust_env = False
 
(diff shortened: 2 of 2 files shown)


Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -907,6 +907,11 @@ def test_is_valid_ok_codes_from_self_false(self):
         response.status_code = 400
         assert not self.client._is_status_code_valid(response)
 
+    def test_http_request_full_url_no_url_suffix(self, requests_mock):
+        requests_mock.get('http://fullurl.com/without/url/suffix', text=json.dumps(self.text))
```
yaakovi: use `example.com` instead of `fullurl.com` or add it to secrets.

yaakovi: also add a test where neither of them is passed to the function

anara123:
my problem with this approach is in 90% of the cases the users will have base_url and url_suffix. If url_suffix has default value, then external contributor can miss this input argument

avidan-H:
@anara123 If the contributor/developer isn't passing the `url_suffix` or the `full_url` argument then a request will be made to the `base_url` and assuming this wasn't the intended endpoint for a command request, the request's response will be incorrect (either an error or the wrong response). What's the problem with that? It doesn't make any sense that when you are passing the `full_url` argument that you should have to supply the `url_suffix` as well since the `full_url` argument is intended as a mechanism to bypass using the client's configured `base_url` and the `url_suffix` argument is only relevant to usage in conjunction with the `base_url`.

Itay4:
@anara123 what do we think about this one?

anara123:
We already talked with @avidan-H . Closing for now.

PR #5479, opened 2020-01-20
Revert "Revert "Limit query time range for better index usage - Find similar incidents search v2""

meirwah:
Reverts demisto/content#5478

@erezh31 plz fix build, it fails on some validation


meirwah:
image

PR #5498, opened 2020-01-21
elasticsearch dateutil parse enhancement

roysagi:
## Status
Ready

fixes: https://github.com/demisto/etc/issues/21510

## Description
ElasticSearch now uses dateutil package to fetch Simple-Date incidents without requiring the Time Format from the user - immensely simplifying the instance configuration.

   - No

## Technical writer review
- [ ] [CHANGELOG](https://github.com/demisto/content/blob/elastic_dateutil/Integrations/Elasticsearch_v2/CHANGELOG.md)


roysagi:
> why do we need all of those `* 100`?

So after testing in an active Elasticsearch with a customer - it appears that for dates ES only accepts some time formats and ***Epoch Timestamps in milliseconds***. 

So my solution is to change all the datetime objects to timestamps for the query itself - hence the *1000 stuff.

roysagi:
@kirbles19 went through my docs - adding docs approved

PR #5504, opened 2020-01-21
fix release notes generator

yaakovi:
## Status
- [ ] In Progress
- [ ] Ready

## Description
in json files from version key is with capital V (`fromVersion`).

## Screenshots
image

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


release_notes.py
```diff
@@ -43,6 +43,9 @@
 
 
 def add_dot(text):
+    if not text:
```
Itay4: What’s that for ?

yaakovi: incase the text is empty, the function code crashes (saw that in another PR).
should never happen, since we have validations for that (in the validator step).

PR #5505, opened 2020-01-21
doc-howto for server howto documentation

glicht:
## Status
- [x] Pending on demisto-sdk

## Description
doc-howto.json for use by the server. 

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


yaakovi:
why is the body escaped as unicode instead of using the actual characters (like `<>`)?

glicht:
@yaakovi it's how the Server provided it. Once all is in place we can check if non Unicode escaping works.

dorsha:
@yaakovi we don't see this file as part of the content, do you know why?

PR #5533, opened 2020-01-23
Url concat multi param

jochman:
## status 
Ready

added an `url_concat` function so you can pass unlimited strings to it.


Itay4:
@jochman whats the difference between that and `urljoin` ?

jochman:
It's receives unlimited number of arguments and not only url and suffix

PR #5546, opened 2020-01-26
[Enhancement] Elasticsearch fetch enriched indicators

DeanArbel:
## Status
- [x] In Progress
- [ ] Ready

## Description
Adds treatment to enriched indicators in the `fetch-indicators` flow.


Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -557,6 +551,18 @@ def get_indicators_search_scan():
...
 
 
+def extract_indicators_from_insight_hit(hit):
+    ioc_lst = []
+    ioc = results_to_indicator(hit)
+    if ioc.get('value'):
+        ioc_lst.append(ioc)
+        module_to_feedmap = ioc.pop('moduleToFeedMap', {})
+        for key, val in module_to_feedmap.items():
+            val['isEnrichment'] = True
```
hirshbergshachar: This is incorrect.
What we should do is check if `val['isEnrichment'] == True`, and if so append the val.
In addition, we don't want to pop the `moduleToFeedMap`, just the relevant keys (the ones which are enrichments)

DeanArbel: I see I understood you differently. Will make these changes

PR #5563, opened 2020-01-28
Jira - jira-get-issue using getattachemnt argument is faulty

bakatzir:
Fixed an issue where trying to get the attachment of an issue using ***jira-get-issue*** raised an error.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21544

## Description
in the referenced issue

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests - added in the TPB

### Tech Docs
- [x] [CHANGELOG](https://github.com/demisto/content/blob/79a39e1f463bebbf6b0dd177e626cc0a836111b9/Integrations/JiraV2/CHANGELOG.md)


Integrations/JiraV2/JiraV2.py
```diff
@@ -394,9 +394,16 @@ def get_issue(issue_id, headers=None, expand_links=False, is_update=False, get_a
...
+    # handle issues were we allowed incorrect values of true
+    if get_attachments == "true" or get_attachments == "\"true\"":
+        get_attachments = True
+    if get_attachments and attachments:
+        attachment_urls = [attachment['content'] for attachment in attachments]
+        for attachment_url in attachment_urls:
+            attachment = f"secure{attachment_url.split('/secure')[-1]}"
+            filename = attachment.split("/")[-1]
+            attachments_zip = jira_req(method='GET', resource_url=attachment).content
+            demisto.results(fileResult(filename=filename, data=attachments_zip))
```
yuvalbenshalom: @bakatzir did you mange to reproduce the error? if so, when dose it happens and when dose the old code work?
multi attachments maybe?

bakatzir: @yuvalbenshalom 
It did not work even with a single attachment. the url we requested was wrong.
i've tested this code with one and multiple attachments.
test pb tests with one attachment.

PR #5564, opened 2020-01-28
Changed ML field associations to be associated with Phishing

idovandijk:
## Status
Ready

## Description
Changed `DBotPrediction` and `DBotPredictionProbability` field associations to be associated only with Phishing incidents.

   - Yes(?) - if the fields were utilized in other incident types, though they shouldn't have.


idovandijk:
> why are we changing that?

The fields are currently only used in Phishing incidents as far as I know, and are displayed for any new incident type right now, which makes no sense. @meirwah @erezh31 is that the only reason?

meirwah:
correct @idovandijk

kirbles19:
@idovandijk Done.

idovandijk:
@anara123 can we force merge?

PR #5566, opened 2020-01-28
fix MISP is storing too much data on indicator

yuvalbenshalom:
## Status
- [ ] In Progress
- [ ] Ready
- [x] In Hold - searching for the best default filtering to use, because it is expected to break BC

## Related Issues
fixes: https://github.com/demisto/etc/issues/20277

## Description
add defaults to `context_select` param. when there are no defaults it will put all of the data to the indicators which can crush the demisto client and slow down the server

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?

### DIFF
--- a/Integrations/MISP_V2/CHANGELOG.md
+++ b/Integrations/MISP_V2/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased]
-Fixed an issue where MISP indicators data was not filtered correctly by default
+Fixed an issue where MISP indicators data was not filtered correctly by default.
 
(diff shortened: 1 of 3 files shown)


yuvalbenshalom:
closing because this is not a full solution to the bug and demisto 5.0 dose not support default values to multi select integration parameters.

PR #5575, opened 2020-01-29
Add tags field of type tags

amitdavidson234:
## Status
- [X] Ready

## Description
After the removal of the tags field in the upgrade process was handled, t's possible to add the field back again

## Minimum version of Demisto
- [X] 5.5.0

   - [X] No


jochman:
we should wait for our CircleCI to get fixed and then will merge from the master. Then we can merge it.

PR #5579, opened 2020-01-29
Fix nightly content instalation

ShahafBenYakir:
## Status
- [x] Ready

## Description
Changed content instalation to run parallel on all servers when running nightly.


.circleci/config.yml
```diff
@@ -403,7 +403,6 @@ jobs:
                 echo "Skipping - not running in INSTANCE_TESTS build"
                 exit 0
             fi
-            ./Tests/scripts/slack_notifier.sh $CIRCLE_NODE_INDEX ./env_results.json
```
yaakovi: why did you remove this?

ShahafBenYakir: I didn't want to spam the content team channel so I removed the notifications while I fixed this. Good thing you caught it.

PR #5584, opened 2020-01-30
[Fix] !cs-resolve-detection not returning results

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18497

## Description
Changed `cs-resolve-detection` to return the same result as cs-detection-details command (previously didn't return any result), and added status to Human Readable.

## Screenshots
image

   - [x] No

## Must have
- [x] Tests


kirbles19:
@DeanArbel Done.

Integrations/integration-CrowdStrikeHost.yml
```diff
@@ -823,7 +937,7 @@ script:
     - name: first_behavior
       description: 'First Behavior of the detection. Example: 2017-01-31T22:36:11Z'
     outputs:
-    - contextPath: CrowdStrikeHost.Detections.detectionID
+    - contextPath: CrowdStrikeHost.Detections.detection_id
```
yuvalbenshalom: @DeanArbel Was this a mistake in the outputs?

DeanArbel: Yeah

DeanArbel:
This PR is done, however the build fails due to the following errors:
- **Output was wrong. I changed the yaml to reflect real output** Integrations/integration-CrowdStrikeHost.yml: Possible backwards compatibility break, You've changed the context in the file,please undo. the command is:
cs-detection-search
- **Unrelated test** The playbook Endpoint Enrichment - Generic v2 - Test does not appear in the conf.json file, which means no test with it will run.please update the conf.json file accordingly

PR #5588, opened 2020-01-30
[Fix] ParseEmailFiles Content-Type validation bug

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21548

## Description
Added new argument `ignore_content_type` that will ignore the contet-type of the email, and will default on text content type. This will enable the user to run the command while ignoring `Content-Type` check in case his eml is missing a `Content-Type` (this should not normally happen, but a user encountered such a case).

   - [x] No

## Must have
- [x] Tests


kirbles19:
@DeanArbel Done.

DeanArbel:
Removed docs-approved since I'm changing the implementation, and will need to rewrite the release notes.

kirbles19:
@DeanArbel Done.

PR #5607, opened 2020-02-02
Added Domain Command to Whois

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#21494

## Description
Added `domain` command and validated existing outputs

## Must have
- [ ] Tests
- [ ] Documentation


Integrations/Whois/Whois.py
```diff
@@ -8245,8 +8245,10 @@ def get_whois(domain, normalized=None):
 
 
 def whois_command():
-
-    domain = demisto.args().get('query')
+    if demisto.args().get('query'):
+        domain = demisto.args().get('query')
+    else:
+        domain = demisto.args().get('domain')
```
yuvalbenshalom: please move the context and hr to a create context and hr function, than create a `whois_command()` and a `domain_command()` and they will use the context and hr functions.
so you will not differentiate by checking the args.

Integrations/Whois/Whois.yml
```diff
@@ -28,6 +28,210 @@ script:
...
+      type: string
+    - contextPath: DBotScore.Type
+      description: The indicator type.
+      type: string
+    - contextPath: DBotScore.Vendor
+      description: The vendor used to calculate the score.
+      type: string
+    - contextPath: Domain.Name
+      description: The domain name.
+      type: string
```
amshamah419: @kirbles19 Can you please check here? It is copypasta from the other command since they are identical but just wanted your input.

kirbles19:
@amshamah419 Updated changelog, YML looks great, no changes.

PR #5617, opened 2020-02-03
release fix to ignore content version mismatch for release branches

avidan-H:
## Status
- Ready

## Description
Fixes issue where during the build the expected content version on the Demisto instances during the `Setup Instances` step is for the new release number but because the installation of new content does not occur until the `Run Tests` steps, the version of content actually found on the instances does not match.

## Minimum version of Demisto
- Any

   - No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Tests/scripts/wait_until_server_ready.py
+++ b/Tests/scripts/wait_until_server_ready.py
@@ -4,6 +4,7 @@ import json
 import ast
 import argparse
 import time
+import re
 from time import sleep
 import datetime
 import requests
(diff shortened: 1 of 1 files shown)


Tests/scripts/wait_until_server_ready.py
```diff
@@ -89,7 +99,6 @@ def is_correct_content_installed(ips, content_version, api_key):
                 err_msg += "Server response: {}".format(resp_json)
             print_error(err_msg)
             return False
-    print_color("Content was installed successfully on all of the instances! :)", LOG_COLORS.GREEN)
```
yaakovi: keep the print incase of success

avidan-H: We already print success for each instance. It is stam here. Plus in the case of release branch it's not even true - it's just misleading.

avidan-H:
> before merging to master, make sure this fixes the release build.
> from this branch, checkout to a new branch and change the content version in `config.yml`

I created a new branch from master and checked out the `wait_until_server_ready.py` from the `fix-release` branch (this branch) and created a PR to merge to master from there.

PR #5619, opened 2020-02-03
Added content installation to instance test too

ShahafBenYakir:
## Status
- [x] Ready

## Description
Added content installation to instance test too.

### DIFF
--- a/Tests/instance_notifier.py
+++ b/Tests/instance_notifier.py
@@ -25,7 +25,7 @@ def options_handler():
 
 
 def install_new_content(client, server):
-     prints_manager = ParallelPrintsManager(1)
+    prints_manager = ParallelPrintsManager(1)
     update_content_on_demisto_instance(client, server, prints_manager, 0)
     prints_manager.execute_thread_prints(0)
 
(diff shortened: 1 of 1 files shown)


Tests/instance_notifier.py
```diff
@@ -43,6 +48,7 @@ def test_instances(secret_conf_path, server, username, password):
 
     for integration in integrations:
         c = demisto_client.configure(base_url=server, username=username, password=password, verify_ssl=False)
+        install_new_content(c, server)
```
yaakovi: why is that inside the loop?

Tests/instance_notifier.py
```diff
@@ -24,6 +24,11 @@ def options_handler():
     return options
 
 
+def install_new_content(client, server):
+    prints_manager = ParallelPrintsManager(0)
+    update_content_on_demisto_instance(client, server, prints_manager, 0)
```
yaakovi: print the collected messages after calling the function:
`prints_manager.execute_thread_prints(0)`

Tests/instance_notifier.py
ShahafBenYakir: True

ShahafBenYakir: True

Tests/instance_notifier.py
ShahafBenYakir: True

Tests/instance_notifier.py
```diff
@@ -27,6 +27,7 @@ def options_handler():
 def install_new_content(client, server):
     prints_manager = ParallelPrintsManager(0)
```
yaakovi: missed that the first time
```suggestion
    prints_manager = ParallelPrintsManager(1)
```

yaakovi:
forced pushed, failed on flashpoint test (unrelated to changes in the PR)

PR #5625, opened 2020-02-04
SumoLogic: do not exit early when fetching records

mwek:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
N/A

## Description
This would cause incorrect results to be returned by the Sumo integration, based only on the subset of messages being scanned.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
   - [X] No

## Must have
- [X] Tests
- [X] Documentation

### DIFF
--- a/Integrations/integration-SumoLogic.yml
+++ b/Integrations/integration-SumoLogic.yml
@@ -260,9 +260,10 @@ script:
                 query = beforeHttp + url + afterWhitespace
             }
             var headers = 'headers' in args ? argToList(args.headers) : undefined;
+            var waitForSearchComplete = args.waitForSearchComplete == 'true';
             var s = search(query, args.from, args.to, a2i(args.limit, defaultLimit), a2i(args.offset, 0), args.timezone,
                 a2i(args.maxTimeToWaitForResults, defaultSearchTimeout) * 60, args.byReceiptTime, a2i(params.sleepBetweenChecks, defaultSleep),
-                args.waitForSearchComplete);
+                waitForSearchComplete);
             var md = '';
             var ec = {};
             if (s.messages && s.messages.length > 0) {
--- a/Integrations/integration-SumoLogic.yml
+++ b/Integrations/integration-SumoLogic.yml
@@ -133,7 +133,7 @@ script:
         return {body: result.Body, obj: obj, statusCode: result.StatusCode, cookies: result.Cookies};
     };
 
-    var search = function(query, from, to, limit, offset, timezone, maxTimeToWaitForResults, byReceiptTime, sleep, fetchRecords) {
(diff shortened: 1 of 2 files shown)


Integrations/integration-SumoLogic.yml
```diff
@@ -261,7 +261,8 @@ script:
             }
             var headers = 'headers' in args ? argToList(args.headers) : undefined;
             var s = search(query, args.from, args.to, a2i(args.limit, defaultLimit), a2i(args.offset, 0), args.timezone,
-                a2i(args.maxTimeToWaitForResults, defaultSearchTimeout) * 60, args.byReceiptTime, a2i(params.sleepBetweenChecks, defaultSleep));
+                a2i(args.maxTimeToWaitForResults, defaultSearchTimeout) * 60, args.byReceiptTime, a2i(params.sleepBetweenChecks, defaultSleep),
+                params.fetchRecords);
```
anara123: I think in this case we better add new argument to `search` command, `waitForSearchComplete` with default value false, so user will have the flexibility on a command level decide if he wants to wait or not.
params.fetchRecords usually we refer to `fetch-incidents`

mwek: I’m happy to implement this, but I think this lays a trap for unsuspecting users. The reason we have discovered the flaw was that our indicator enhancement did not return any values, while the Sumo search ran manually returned it correctly.
The “terminate shortly” logic is flawed when you want to have accurate aggregate results: the API response returns the aggregations correctly, but over incomplete data. So whoever sets `fetchAggregations` without `waitForSearchComplete=true` ends up receiving incomplete results

Integrations/integration-SumoLogic.yml
```diff
@@ -215,7 +215,7 @@ script:
             }
 
             var s = search(params.fetchQuery, lastRun.time, now, a2i(params.limit, defaultLimit), 0, params.timeZone, a2i(params.maxTimeout, 180), false,
-                a2i(params.sleepBetweenChecks, defaultSleep));
+                a2i(params.sleepBetweenChecks, defaultSleep), params.fetchRecords);
```
ronykoz: why do we need the addition of params.fetchRecords here?
Shouldn't we want this to be `true` for both records and messages?

ronykoz:
@kirbles19 could you go over the new argument and the changelog please?

Integrations/integration-SumoLogic.yml
mwek: No, when fetching messages, we are still only returning the first (params.limit / defaultLimit) entries. In this case, we can return early from the search.
For aggregate records, we want to iterate over all messages, as otherwise the aggregations will be computed over a partial data set only.

kirbles19:
@ronykoz Done.

PR #5662, opened 2020-02-06
[Fix] ReadPDFFileV2 doesn't extract indicators correctly

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21826

## Description
- Fixed a bug where emails would be labeled as urls.
- Added Email standard context.

   - [x] No

## Must have
- [x] Tests


amshamah419:
@yuvalbenshalom nice.

PR #5685, opened 2020-02-09
AWS Feed fix

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
Fixed the mapping to be cli name and not name

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 
- [ ] Code Review


liorblob:
I wasn't aware of the cliname thing, maybe we should check that the region is mapped correctly in the UI.

PR #5693, opened 2020-02-10
Mcafee web gateway - obscure secrets

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21620

## Description
in issue

## Minimum version of Demisto
- [x] 4.5.0

   - [x] No

### DIFF
--- a/Integrations/integration-McAfeeWebGateway.yml
+++ b/Integrations/integration-McAfeeWebGateway.yml
@@ -65,7 +65,7 @@ script:
 
         if (res.StatusCode < 200 || res.StatusCode >= 300) {
             if (res.Status.indexOf('pass=') !== -1) {
-                throw 'Request Failed.\nStatus code: ' + res.StatusCode + '.';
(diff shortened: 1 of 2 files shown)


Integrations/integration-McAfeeWebGateway.yml
```diff
@@ -64,6 +64,9 @@ script:
         );
 
         if (res.StatusCode < 200 || res.StatusCode >= 300) {
+            if (res.Status.indexOf('pass=') !== -1) {
+                throw 'Request Failed.\nStatus code: ' + res.StatusCode + '. Check your Server URL and Port.';
+            }
             throw 'Request Failed.\nStatus code: ' + res.StatusCode + '.\nBody: ' + JSON.stringify(res) + '.';
```
yuvalbenshalom: take a look at the example of the bad test print you putted in the original issue.
there is no "Body:" there. so where did it came from?

PR #5711, opened 2020-02-11
CSV Feed - Fix test

liorblob:
## Status
- [ ] In Progress
- [x] Ready


liorblob:
@kirbles19 can you review?

Packs/FeedCSV/Integrations/FeedCSV/FeedCSV.yml
```diff
@@ -83,8 +83,7 @@ configuration:
   name: ignore_regex
   required: false
   type: 0
-- additionalinfo: Name of the field names in the CSV feed. If several are given, will
-    use "indicator" as the default indicator value field.
+- additionalinfo: The names to give the fields in the CSV feed. If several are given, "value" will be used as the default indicator value field.
```
avidan-H: I'm not in love with this tooltip but if you feel strongly then do you. I think we need to make it _super_ clear that the field called "value" is the one passed as the value of an indicator.

liorblob: Moving to https://github.com/demisto/content/pull/5766

PR #5716, opened 2020-02-12
Update incidentfield-TargetFirewallVersion.json

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
remove required field from incident field.
related PR: https://github.com/demisto/content/pull/5682

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Additional Info
@anara123 @bakatzir 
should this field be associated to all incident types?


anara123:
@yaakovi I think the main issue here is that this field should be assosiated to specifc incident type

PR #5731, opened 2020-02-13
fixed none type error

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/20719#issuecomment-58550929


IkaDemisto:
Should fix this error:
Screen Shot 2020-02-13 at 15 18 14

anara123:
@IkaDemisto but what is this field `integrationScript` inside of `configuration`? I don't see that any integration have it

IkaDemisto:
@anara123 it is returned from the rest api when creating integration instance using `demisto-py`:
`
        res = demisto_client.generic_request_func(self=client, method='PUT',
                                                  path='/settings/integration',
                                                  body=module_instance)
`

yaakovi:
@anara123 this is the `script` dict in integrations, the key is taken from the server representation of integrations: https://github.com/demisto/server/blob/master/domain/module.go#L336

PR #5740, opened 2020-02-13
fixed an issue with as-get-entries

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21904

## Description
Fixed an issue where ***as-get-entries*** return output in incorrect format for results with a large number of objects.

   - [x] No

## Must have
- [x] Tests


kirbles19:
@DeanArbel Done.

PR #5772, opened 2020-02-18
Cofense mapping

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
Cofense mapping

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
+++ b/Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
@@ -157,8 +157,27 @@ class Client(BaseClient):
             block["value"] = value
             block["type"] = indicator
             block["threat_id"] = threat_id
+            malware_family: dict = block.get("malwarefamily", {})
+            ip_detail: dict = block.get("ipDetail", {})
             if indicator:
-                results.append({"value": value, "type": indicator, "rawJSON": block})
+                results.append({
+                    "value": value,
+                    "type": indicator,
+                    "rawJSON": block,
+                    "fields": {
+                        "name": threat_id,
+                        "malwarefamily": malware_family.get("familyName"),
+                        "description": malware_family.get("description"),
+                        "feedoriginalseverity": block.get("impact"),
+                        "feedthreattypes": {
+                            "threatcategoryconfidence": block.get("confidence"),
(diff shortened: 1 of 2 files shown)


Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
```diff
@@ -157,8 +157,27 @@ def process_item(cls, threat: dict) -> List[dict]:
...
-                results.append({"value": value, "type": indicator, "rawJSON": block})
+                results.append({
+                    "value": value,
+                    "type": indicator,
+                    "rawJSON": block,
+                    "fields": {
+                        "name": threat_id,
+                        "malwarefamily": malware_family.get("familyName"),
+                        "description": malware_family.get("description"),
+                        "feedoriginalseverity": block.get("impact"),
```
avidan-H: Please check what happens if these values are `None`. If it's problematic then you'll need to set the default argument to `get()` to be of the appropriate type that given indicator field expects.

Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
```diff
@@ -157,8 +157,27 @@ def process_item(cls, threat: dict) -> List[dict]:
...
+                        "feedthreattypes": {
+                            "threatcategoryconfidence": block.get("confidence"),
+                            "threatcategory": block.get("role")
+                        },
+                        "geocountry": ip_detail.get("countryIsoCode"),
```
avidan-H: The same goes for here as well.

Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
guyfreund: Checked, no exception is raised. Indicators with missing fields are being fetched as usual.

Packs/FeedCofense/Integrations/FeedCofense/FeedCofense.py
guyfreund: Checked, no exception is raised. Indicators with missing fields are being fetched as usual.

PR #5796, opened 2020-02-19
azurefeed sdk check

bakatzir:
fixes: https://github.com/demisto/etc/issues/22233


roysagi:
no docs adding docs approved

ronykoz:
@bakatzir seems to be failing on hooks validations

bakatzir:
@ronykoz We know, solution is here: https://github.com/demisto/demisto-sdk/pull/192
and will not fail after the force-merge in master.

PR #5808, opened 2020-02-20
Updated filterByList

ronykoz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22216


ronykoz:
Already created the v2, but after that i realised that the v1 change will not break BC, @Itay4 what do you think?

Itay4:
we're just adding an output, no?

ronykoz:
@Itay4 well yes and no, the path remains the same, but the structure in the context changes.
But i don't think it will cause any trouble, am i right?

PR #5811, opened 2020-02-20
Added new Default playbook starting in 5.0

idovandijk:
## Status
Ready - **Note that the ID is different than the name, for BC purposes. Therefore, a force-merge is needed.**

## Related Issues
fixes: https://github.com/demisto/etc/issues/22114

## Description
- New `Default` playbook starting in 5.0. Uses v2 playbooks and `extractIndicators` instead of the deprecated `Extract Indicators - Generic` playbook
- Test playbook for the new `Default` playbook

## Screenshots
image

## Minimum version of Demisto
5.0.0

Yes, but only in very rare occasions where someone used the `Default` playbook and created a scheduled task to automatically close the manual task there, because the new manual task is a condition task, and won't be closed in the same way (although both were manual).

## Must have
- [x] Tests
- [ ] Code Review
- [ ] Documentation
https://github.com/demisto/content/blob/481957435a74561be308b2d92665c72f03001e56/Playbooks/playbook-Default_5.0.yml
lines 41-42
lines 85-86
lines 144-145
lines 208-211

Changelogs.


idovandijk:
@yaakovi I added you to make sure it's fine to merge with ID that is different than the name

yaron-libman:
Updated texts.

PR #5816, opened 2020-02-21
[risksense-835] RiskSense Integration

crestdatasystems:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: NA

## Description
This is the initial commit for RiskSense Integration.
RiskSense is a cloud-based platform that provides vulnerability management and prioritization to measure and control cybersecurity risk.

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

## Demisto Partner?
- [x] Crest Data Systems


Integrations/RiskSense/RiskSense.py
```diff
@@ -0,0 +1,1975 @@
...
+
+def test_module(client):
+    """
+        Performs basic GET request
+    :param client:
+    :return: None
+    """
+    client_name = demisto.params().get('client_name', '-')
+    resp_json = client.http_request('GET', url_suffix='')
+    clients = resp_json.get('_embedded', {}).get('clients', [])
```
ronykoz: better to put this in a try, except as the URL might be faulty

crestsantoshgiri: @ronykoz if URL is faulty then it will fall under main method's catch exception and will use `return_error` method

crestsantoshgiri:
> is the CVE list unique for each finding?
@ronykoz Yes, CVE is the unique key for` !risksense-get-unique-cves` command

Integrations/RiskSense/RiskSense.py
ronykoz: I understand, but we want to make sure that the error will be clear enough for the user, so he will just have to read it and will know what to search for/do in order to fix it

crestsantoshgiri: @ronykoz we can do that but, we are getting DemistoException (CommonServerPython) hence won't be able to differentiate the type of exception (in this case ConnectionError). Let us know if we have missed anything.

ronykoz: If you wrap this code section with try except you should be able to catch it here, and then return_error with a good msg attached to it

crestsantoshgiri: okay @ronykoz, will that in `test_module`

Integrations/RiskSense/RiskSense.py
```diff
@@ -0,0 +1,1975 @@
...
+    # Prepares filter request body
+    data = prepare_filter_payload(args, 'detail')
+
+    resp = client.http_request('POST', url_suffix=url_suffix, json_data=data)
+    page_number, total_element, total_pages = fetch_page_details(resp)
+
+    if total_element == 0:
+        return 'No host(s) found for the given argument.', {}, {}
+    if page_number >= total_pages:
+        raise ValueError('Invalid page navigation.')
```
ronykoz: This will end more gracefully this way
```suggestion
       return_error('Invalid page navigation.')
```

Integrations/RiskSense/RiskSense.py
```diff
@@ -0,0 +1,1975 @@
...
+def get_risksense_host_context(href, resp_host):
+    """
+    Prepare context data under RiskSense host.
+
+    :param href:  Reference link from host resp.
+    :param resp_host: Response from host command.
+    :return: None
+    """
+    return {
+        'Id': resp_host.get('id', ''),
```
ronykoz: lets change this to ID please, it's a convention, and much easier to read.
```suggestion
        'ID': resp_host.get('id', ''),
```

Integrations/RiskSense/RiskSense.py
```diff
@@ -0,0 +1,1975 @@
...
+        resp_list_hostfinding = resp.get('_embedded', {}).get('hostFindings', [])
+
+        host_finding_details_hr = []  # type: List[Dict[str, Any]]
+        risksense_host_finding_context = []  # type: List[Dict[str, Any]]
+
+        for resp_hostfinding in resp_list_hostfinding:
+            host_finding_details_hr.append(get_host_finding_hr(resp_hostfinding))
+            risksense_host_finding_context.append(get_risksense_host_finding_context(resp_hostfinding))
+
+
```
ronykoz: this command uses the exact same logic as the opened one, it will be easier to maintain just one function - not a must.

Integrations/RiskSense/RiskSense.py
crestsantoshgiri: @ronykoz  we have made given changes in entire source code for above ValueError exception.

Integrations/RiskSense/RiskSense.py
crestsantoshgiri: @ronykoz this has been taken care

Integrations/RiskSense/RiskSense.py
crestsantoshgiri: @ronykoz this has been taken care

crestdatasystems:
@ronykoz - Did you get chance to review the updated app after we submitted the changes last week?

ronykoz:
@kirbles19 Could you please go over the integration?

crestsantoshgiri:
> 
> 
> Looks good, just one thing:
> I think that the between argument should be a separate one and not part of the operator.
> And you might just want to split into two different ones, start_time, end_time or something of that sort

The between operator is applicable to date and numeric data type, hence we have added it as an operator. Also it matches with the RiskSense API format.

kirbles19:
@ronykoz Done.

PR #5822, opened 2020-02-22
fix proxy ignore. fix chrome driver log file. add debug-mode option

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20702

## Description
* Fix so proxy is not used when the settings checkbox is NOT clicked.
* Add logging to a log file and display the content of the file when in debug-mode
* add no_proxy to localhost host to fix: https://github.com/demisto/etc/issues/20702 . When proxy env variables where defined the shutdown command was actually being sent to the proxy and not to webdriver (I think this command goes through a different flow in selenium).

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation  (changelog)


glicht:
@kirbles19 also check the description. Added a note about proxies. See: https://github.com/demisto/content/pull/5822/files#diff-19cc0194bd7ef4000491240d867ba570

kirbles19:
@glicht Done.

glicht:
@IkaDemisto I've added support for pid and memory threshold support at the docker image level. So it is possible to specify the thresholds per docker image instead of per test. Please take a look.

PR #5830, opened 2020-02-23
Removed url regex in office365 feed 

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21988#event-3059767135

## Description
Removed url regex in check of indicator type.


bakatzir:
Office365 contains AFAIK only domain and domainGlob.
we will add the domainGlob to Office 365 integration after the indicator will be merged.

bakatzir:
@IkaDemisto If you can add a comment that the feed only contains IPs(v4,v6,cidr) and Domains in the code or documentation, might be easier for support and upcoming coders.

PR #5836, opened 2020-02-23
First draft for powershell remoting over ssh integration

yardensade:
This is a first non yet working draft, just so we have a sense of what it would look like more or less.

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22231

## Description
Few points to discuss:
- ~~I need a prepared docker image with openssh client in order to test this, this code is not yet tested and will take some time to make it work correctly.~~ -https://github.com/demisto/dockerfiles/pull/1366
- ~~Currently have few pre-commit issues that integration is not passing mostly about some yml fields, needs to be solved.~~
- ~~Image is awaiting design by UI, already gave them whats needed.~~
- In order to have a test playbook for this we will need 24/7 access to a machine, preferably where nothing harmful can happen.
- What's running inside is essentially a block of code, should discuss if we want to be able to run multiple commands one after another or just a single block of code each time. (normally this integration is ideal for long running but has security aspect to it that needs to be addressed)
- Output currently is json entry, we don't have table to markdown function for powershell, could output the content as Out-String that will prettify it.
- Need to look about implementing insecure/proxy params.

## Screenshots
Non yet

## Minimum version of Demisto
- [X] 5.5.0

   - [x] No

## Must have
- [ ] Tests - Good luck with that
- [ ] Documentation - Good luck with that


Itay4:
Closing this due to changes done in `master` branch.
Please reopen in a new branch.

PR #5847, opened 2020-02-24
fix for panaroma-edit-rule with the behaviour option add  or remove

ssudheerk99:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
in the present integration, if we select the behavior as add or remove. first it fetches the current objects and ideally it  has to make a list of present IPs But, instead of making list of Just IPs the present integration is making list of some metadata as well. so edited the fucntion to return only list of present IPs

## Screenshots
image
after editing:

image

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

### DIFF
--- a/Integrations/Panorama/Panorama.py
+++ b/Integrations/Panorama/Panorama.py
@@ -2726,7 +2726,7 @@ def panorama_get_current_element(element_to_change: str, xpath: str) -> list:
     if 'list' in current_object:
         current_objects_items = argToList(current_object['list']['member'])
     elif 'member' in current_object:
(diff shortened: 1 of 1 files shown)


bakatzir:
@ssudheerk99 Thanks for the contribution.
Our pan-os process is based on the fact that all objects must be committed prior to editing them.
The root cause is that we do not want to allow to multiple people to edit the same objects in the same time.
Hence, we will not accept this change.

PR #5867, opened 2020-02-25
adding url params support and offset to Export Indicators

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22124

## Description
Added offset param to the Export Indicators.
Added support of url in line params in the format stated here: https://live.paloaltonetworks.com/t5/MineMeld-Articles/Parameters-for-the-output-feeds/ta-p/146170
Added the following behavior - in case limit, offset or format change - the results are refreshed regardless of cache time.
Added offset to ei-update command.
Added query change support to URL params - enhancement over the original.

Also added an accompanying unit test for the offset.


roysagi:
@DeanArbel - FYI

roysagi:
Added the following behavior - in case limit, offset or format change - the results are refreshed regardless of cache time.

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -204,6 +230,37 @@ def validate_basic_authentication(headers: dict, username: str, password: str) -
 ''' ROUTE FUNCTIONS '''
 
 
+def get_request_args(params):
+    limit = request.args.get('n', None)
```
ronykoz:
```suggestion
    limit = try_parse_integer(request.args.get('n', params.get('list_size')), CTX_LIMIT_ERR_MSG)
```
Same goes for the rest, but what happens if the number is not present anywhere?

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -204,6 +230,37 @@ def validate_basic_authentication(headers: dict, username: str, password: str) -
...
+    offset = request.args.get('s', None)
+    out_format = request.args.get('v', None)
```
ronykoz:
```suggestion
    out_format = request.args.get('v', params.get('format'))
```

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -64,31 +66,48 @@ def get_params_port(params: dict = demisto.params()) -> int:
...
-def find_indicators_with_limit(indicator_query: str, limit: int) -> list:
+def find_indicators_with_limit(indicator_query: str, limit: int, offset: int) -> list:
     """
     Finds indicators using demisto.searchIndicators
     """
-    iocs, _ = find_indicators_with_limit_loop(indicator_query, limit)
-    return iocs[:limit]
+    # calculate the starting page (each page holds 200 entries)
+    if offset:
+        next_page = int(offset / 200)
```
ronykoz:
```suggestion
        next_page = int(offset / PAGE_SIZE)
```

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -64,31 +66,48 @@ def get_params_port(params: dict = demisto.params()) -> int:
...
+
+        # set the offset from the starting page
+        parsed_offset = offset - (200 * next_page)
```
ronykoz:
```suggestion
        parsed_offset = offset - (PAGE_SIZE * next_page)
```

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -64,31 +66,48 @@ def get_params_port(params: dict = demisto.params()) -> int:
...
+        'last_limit': limit,
+        'last_offset': offset,
+        'last_format': out_format,
+        'last_query': query
+    })
     demisto.setIntegrationContext(out_dict)
 
 
-def find_indicators_with_limit(indicator_query: str, limit: int) -> list:
+def find_indicators_with_limit(indicator_query: str, limit: int, offset: int) -> list:
```
ronykoz: the terminology of page offset and parsed offset is confusing here

roysagi:
Added changes:
offset - no longer parameter of instance configuration.

on demand behavior: 
 - indicator list is not refreshed until eis-update command is given.
 - changes in limit, offset and format are done on the existing list.

Error handling- now throws informative error output.

In case offset is larger than the list returned - returns an empty output.

roysagi:
@kirbles19 - if you can please review the following:
[Changelog](https://github.com/demisto/content/blob/export_indicators_url_params/Packs/ExportIndicators/Integrations/ExportIndicators/CHANGELOG.md)
[yml](https://github.com/demisto/content/blob/export_indicators_url_params/Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.yml)

ronykoz:
@anara123 do we want another URL argument to refresh the list even if the on demand checkbox is checked?

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -247,32 +274,36 @@ def route_list_values() -> Response:
...
-        offset=offset,
-        last_update_data=demisto.getLastRun(),
-        indicator_query=query,
-        cache_refresh_rate=params.get('cache_refresh_rate')
-    )
-
-    mimetype = get_outbound_mimetype()
-    return Response(values, status=200, mimetype=mimetype)
+    except Exception as e:
+        return Response(str(e), status=500, mimetype='text/plain')
```
ronykoz:
```suggestion
        return Response(str(e), status=400, mimetype='text/plain')
```
What do you think?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -169,29 +175,50 @@ def get_outbound_ioc_values(on_demand, limit, offset, indicator_query='', out_fo
...
+            return ''
+
+        iocs = iocs[offset: limit + offset]
+        returned_dict = create_values_out_dict(iocs, new_format)
+        current_cache = demisto.getIntegrationContext()
+        current_cache['last_output'] = returned_dict
+        demisto.setIntegrationContext(current_cache)
+
+    else:
+        returned_dict = demisto.getIntegrationContext().get('last_output')
```
ronykoz: can we have a case in which in the first fetch this will be empty?

kirbles19:
@roysagi Done. The yml looked fine.

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
roysagi: yes it can happen if the query returns nothing or if the offset requested is larger than the amount brought back.

ronykoz: Then it could break the return line(line 221), you should have a default value of {}

Packs/ExportIndicators/Integrations/ExportIndicators/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 ## [Unreleased]
+  - Added the *offset* parameter to the ***eis-update*** command.
+  - Added support for the following inline URL parameters.
+    - n - The number of indicators to fetch.
+    - s - The first index from which to fetch indicators.
+    - v - The output format for indicators.
+    - q - The query that defines which indicators to fetch.
```
DeanArbel: This is an extensive addition, this should be documented in the readme. Please add it there as well as to the detailed description, as this is not trivial for a user with no past experience with these arguments.

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -64,31 +66,54 @@ def get_params_port(params: dict = demisto.params()) -> int:
...
+    iocs = find_indicators_with_limit(indicator_query, limit, offset)  # poll indicators into list from demisto
     out_dict = create_values_out_dict(iocs, out_format)
     out_dict[CTX_MIMETYPE_KEY] = 'application/json' if out_format == FORMAT_JSON else 'text/plain'
-    save_context(now, out_dict)
+    save_context(now, limit, offset, out_format, indicator_query, iocs, out_dict)
     return out_dict[CTX_VALUES_KEY]
 
 
-def save_context(now: datetime, out_dict: dict):
+def save_context(now: datetime, limit: int, offset: int, out_format: str, query: str, iocs: list, out_dict: dict):
```
DeanArbel: This function is pretty redundant, as it only contains a single line of code, which calls  `demisto.setIntegrationContext`. Consider removing the function.

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -136,36 +161,64 @@ def create_values_out_dict(iocs: list, out_format: str) -> dict:
...
+
     else:
-        if last_run:
+        if last_update:
+            # takes the cache_refresh_rate amount of time back since run time.
             cache_time, _ = parse_date_range(cache_refresh_rate, to_timestamp=True)
-            if last_run <= cache_time:
-                values_str = refresh_outbound_context(indicator_query, out_format, limit=limit)
+            if last_update <= cache_time or last_limit != limit or \
+                    last_offset != offset or last_format != out_format or indicator_query != last_query:
```
DeanArbel: Consider extracting this along with lines 182-183 to a boolean function such as `is_request_change`, it might make the code a bit longer, however, it'll make it easier to understand what you're checking here, and to test this logic. Please make sure to add a unit test for this logic condition.

anara123:
@ronykoz no need for another parameter for now

Packs/ExportIndicators/Integrations/ExportIndicators/CHANGELOG.md
roysagi: done

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -136,36 +156,80 @@ def create_values_out_dict(iocs: list, out_format: str) -> dict:
 
 def get_outbound_mimetype() -> str:
     """Returns the mimetype of the export_iocs"""
-    ctx = demisto.getIntegrationContext()
+    ctx = demisto.getIntegrationContext().get('last_output')
     return ctx.get(CTX_MIMETYPE_KEY, 'text/plain')
 
 
-def get_outbound_ioc_values(on_demand, limit, indicator_query='', out_format='text', last_run=None,
+def is_request_change(limit, offset, out_format=FORMAT_TEXT, last_update_data={}) -> bool:
```
DeanArbel: Nice 👍

PR #5868, opened 2020-02-25
Added a check that prevents empty email addresses from being enriched.

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20787

## Description
Added a check that will prevent empty email addresses from being enriched.
A separate fix should be done to Autoextract regex (if that is what caused empty email extraction).

## Minimum version of Demisto
4.1.0

   - [x] No

## Must have
- [x] Tests
- [x] Code Review
- [ ] Documentation


idovandijk:
Coordinated docs with Alex. Merging.

PR #5875, opened 2020-02-26
Return GetTime test playbooks

jochman:
## Status
- [ ] In Progress
- [x] Ready


ShahafBenYakir:
@jochman What was fixed? 🤔

jochman:
Nothing, it's working (this playbook is running script with no API calls.)

ShahafBenYakir:
SO why did it fail? Plus, aren't these two TPB?

jochman:
Two PB that are dependent on the same script.
Server issue most likely.

ShahafBenYakir:
Add a `fixes: ...` for the two issues, otherwise, these issues will remain open

PR #5881, opened 2020-02-26
New integraion - Fidelis EDR

Shellyber:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20495

   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation


Itay4:
running `list-alerts` with no `limit` given does seem to work:
image

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1577 @@
...
+def list_alerts_command(client: Client, args: dict):
+    """get information about alerts. """
+
+    skip = args.get('skip', 0)
+    limit = args.get('limit')
+    sort = args.get('sort')
+    facet_search = args.get('facet_search')
+    start_date = args.get('start_date')
+    end_date = args.get('end_date')
+    headers = ['Name', 'EndpointName', 'EndpointID', 'Source', 'ArtifactName', 'IntelName', 'Severity', 'ID',
```
Itay4: make sure `ID` is the first header in all commands outputs
add `CreateDate` to headers

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+      type: String
+    - contextPath: FidelisEndpoint.Script.Description
+      description: The script description.
+      type: String
+    - contextPath: FidelisEndpoint.Script.TimeoutSeconds
+      description: Script timout in seconds
+      type: Number
+  - arguments:
+    - default: false
+      description: Endpoint IP.
```
Itay4: where can this be retrieved from?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1577 @@
...
+
+    if os == 'macOS':
+        script_id = '020114c2-d000-4876-91b0-97f41a83b067'  # disable-secrets-detection
+
+    response = client.list_process(script_id, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Script(val.ID && val.ID === obj.ID)': context}
```
Itay4: the outputs here do not seem relevant to this command

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+    - default: false
+      description: Endpoint IP. The endpoint to run the script on.
+      isArray: false
+      name: endpoint_ip
+      required: true
+      secret: false
+    - default: false
+      description: What script to run. Get the answer from script-manifest command).
+      isArray: false
+      name: answer
```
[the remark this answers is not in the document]
Shellyber: Wanted to improve that but after discussion with the designer, he said to leave it as it is

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint_test.py
```diff
@@ -0,0 +1,25 @@
+from FidelisEndpoint import Client, say_hello_command,\
```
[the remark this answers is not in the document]
Shellyber: Why unittests are needed here?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1578 @@
...
+        self._password = password
+        self._proxies = handle_proxy() if proxy else None
+        self._token = self._generate_token()
+        self._headers = headers
+
+        if self._token:
+            token = self._token['data'].get('token')
+            headers.update({'Authorization': 'bearer ' + token})
+
+    def _generate_token(self) -> str:
```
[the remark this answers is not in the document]
Shellyber: yes. I'm taking only the token part and converting it to string

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1577 @@
...
+    }
+    human_readable = tableToMarkdown('Fidelis Endpoint file results metadata', contents, removeNull=True)
+
+    return human_readable, entry_context, response
+
+
+def get_file(client: Client, args: dict):
+    file_id = args.get('file_id')
+    response = client.get_file(file_id)
+    attachment_file = fileResult('Fidelis_Endpoint.txt', response)
```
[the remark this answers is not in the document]
Shellyber: No...

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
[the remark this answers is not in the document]
Itay4: what is this argument? what does it represent? how is it named in the ui?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint_test.py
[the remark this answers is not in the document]
Itay4: if none is needed then why do we need this file? remove it

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1577 @@
...
+def list_process_command(client: Client, args: dict):
+    endpoint_ip = argToList(args.get('endpoint_ip'))
+    endpoints = client.convert_ip_to_endpoint_id(endpoint_ip)
+    endpoint_id = endpoints.get('data')
+    time_out = args.get('time_out')
+    os = args.get('os')
+    script_id = ''
+
+    if os == 'Windows':
+        script_id = '2d32a530-0716-4542-afdc-8da3bd47d8bf'  # disable-secrets-detection
```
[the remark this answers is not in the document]
Shellyber: Yes, It's general. 
Will make it global

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1577 @@
...
+
+    entry_context = {'FidelisEndpoint.FileSearch(val.JobID && val.JobID === obj.JobID)': contents}
+
+    human_readable = f'Fidelis Endpoint file search status is: {status}'
+
+    return human_readable, entry_context, data
+
+
+def file_search_reasult_metadata(client: Client, args: dict):
+    """Get the job results metadata (max 50 results)"""
```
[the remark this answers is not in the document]
Shellyber: yeah, according to the API docs

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Shellyber: This is the only output that returns (the data) the ID is just the unique key to add to the right context path.
The data is basically the same as Job ID. (its a workaround from Fidelis)

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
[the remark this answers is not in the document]
Itay4: but you're doing it in the `__init__` function, which means you are returning a `Dict` here, no?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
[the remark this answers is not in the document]
Itay4: ok so lets put it in the command description, no need for it to be in the code

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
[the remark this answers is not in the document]
Itay4: ok so this name is problematic, as it is not a text file but an executable if i understand correctly
lets set the name to be something like `FidelisEndpoint_{file_id}.exe`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Itay4: yea but listing processes, not scripts, so why are we populating these under `Script` ?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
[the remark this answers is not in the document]
Shellyber: It will also be problematic, since the file id is really long and messy string. Also the file doesn't have to be .exe

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Shellyber: Will check it again with design. Basically all this commands are script executions.

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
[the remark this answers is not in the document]
Itay4: what other file types can there be?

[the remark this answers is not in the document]
Shellyber: No limitation, can be exe, txt, pdf, etc..

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Shellyber: Changed.

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1775 @@
...
+            'name': f"Fidlie Endpoint alert: {alert.get('id')}",
+            'occurred': alert.get('createDate'),
+            'severity': alert_severity_to_dbot_score(alert.get('severity')),
+            'rawJSON': json.dumps(alert)
+        } for alert in alerts if alert.get('id') > last_incident_id and alert.get('severity') == severity]
+        # New incidents fetched
+        if incidents:
+            last_incident_timestamp = incidents[-1].get('occurred')
+            last_incident_id = alerts[-1].get('id')
+            new_last_run = {'time': last_incident_timestamp, 'id': last_incident_id}
```
Itay4: why are we storing the timestamp?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1619 @@
...
+  name: isFetch
+  required: false
+  type: 8
+- defaultvalue: 3 days
+  display: First fetch timestamp (<number> <time unit>, e.g., 12 hours, 7 days, 3
+    months, 1 year)
+  name: fetch_time
+  required: false
+  type: 0
+- display: severity
```
Itay4: improve display name

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1619 @@
...
+  name: severity
+  required: false
+  type: 0
```
Itay4: do we know if the severity value can be one of `High`, `Medium` and `Low`, set this param to be multi select

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1619 @@
...
+  required: false
+  type: 8
+- defaultvalue: 3 days
+  display: First fetch timestamp (<number> <time unit>, e.g., 12 hours, 7 days, 3
+    months, 1 year)
+  name: fetch_time
+  required: false
+  type: 0
+- display: severity
+  name: severity
```
Itay4: add default value

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
Itay4: add limit parameter for the fetch

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1619 @@
...
+  type: 9
+- display: Incident type
+  name: incidentType
+  required: false
+  type: 13
+- display: Trust any certificate (not secure)
+  name: insecure
+  required: false
+  type: 8
+- display: Use system proxy settings
```
Itay4: move proxy and insecure to be below all other parameters

Tests/secrets_white_list.json
```diff
@@ -62,6 +62,10 @@
       "8.8.8.8",
       "3.7.3.221",
       "192.168.1.136",
+      "10.128.0.1",
```
Itay4: use pack secrets for this

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1619 @@
...
+      description: Event index.
+      type: Number
+    - contextPath: FidelisEndpoint.Query.IndexingTime
+      description: Indexing time.
+      type: Date
+    - contextPath: FidelisEndpoint.Query.EntityType
+      description: Entity type.
+      type: Number
+  dockerimage: demisto/python3:3.7.4.2245
+  isfetch: false
```
Itay4:
```suggestion
  isfetch: true
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1775 @@
...
+    end_time = args.get('end_time')
+    logic = args.get('logic')
+    entity_type = args.get('entity_type')
+    column = args.get('column')
+    value = args.get('value')
+    operator = args.get('operator')
+    limit = args.get('limit')
+    additional_filter = args.get('additional_filter')
+    if additional_filter:
+        add_filter = additional_filter.split()
```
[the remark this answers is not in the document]
Shellyber: Adding another filter to the query, then splitting the values according to the dict the request suppose to get.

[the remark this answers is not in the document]
Itay4: but shouldnt you provide some delimiter to split on? like `=` ?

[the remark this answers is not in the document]
Shellyber: No, its will split on space. the delimiter is part of the query, it goes to the operator value.
So the user will need to enter for example: "pid = 1234" (its written in the description)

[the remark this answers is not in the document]
Itay4: in the description there are not whitespaces:
image
which by the way make more sense than with spaces, so lets do it without spaces and need to add verification the length of the array from the split result so we wont get index error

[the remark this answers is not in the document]
Shellyber: Can't, the operator doesn't have to be =
So I'll fix the description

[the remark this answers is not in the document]
Itay4: you can keep `=`, just change to `add_filter = additional_filter.split('=')`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1710 @@
...
+# Remove network isolation
+REMOVE_NETWORK_ISOLATION_WINDOWS = '99bbaea5-df18-40cc-8759-b5fb61527d5a'  # disable-secrets-detection
+REMOVE_NETWORK_ISOLATION_MAC_LINUX = '5e252298-4c50-4cdd-94c0-d6997b79157c'  # disable-secrets-detection
+
+
+class Client(BaseClient):
+    """
+    Client to use in the Fidelis Endpoint integration. Overrides BaseClient
+    """
+    def __init__(self, server_url: str, username: str, password: str, verify: bool, proxy: bool, headers: dict):
```
Itay4: remove `headers` from here as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1710 @@
...
+
+        super().__init__(base_url=server_url, verify=verify, proxy=proxy, headers=headers)
```
Itay4: lets change it to something like this?
```
token = self._generate_token(username, password)
headers = {'Authorization': f'Bearer {token}'}
super().__init__(base_url=server_url, verify=verify, proxy=proxy, headers=headers)
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1710 @@
...
+
+        return self._http_request('GET', url_suffix, params=params)
+
+    def get_host_info(self, host_name: Union[None, str], ip_address: Union[None, str]):
+
+        if host_name:
+            url_suffix = f'/endpoints/v2/0/100/hostname%20Ascending?accessType=3&search={{%22searchFields%22: \
+                             [{{%22fieldName%22:%22HostName%22,%22values%22:[{{%22value%22:%22{host_name}%22}}]}}]}}'
+
+        if ip_address:
```
Itay4: in all cases you need either `host_name` or `ip_address`, please change to `elif`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            startDate=start_date,
+            endDate=end_date
+        )
+
+        return self._http_request('GET', url_suffix, params=params)
+
+    def get_host_info(self, host_name: Union[None, str], ip_address: Union[None, str]):
+
+        if host_name:
+            url_suffix = f'/endpoints/v2/0/100/hostname%20Ascending?accessType=3&search={{%22searchFields%22: \
```
Itay4: better to pass the url query params to `params` argument of `_http_request`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+
+        body = ip
+
+        return self._http_request('POST', url_suffix, json_data=body)
+
+    def convert_name_to_endpoint_id(self, endpoint_name: list = None) -> Dict:
+
+        url_suffix = '/endpoints/endpointidsbyname'
+
+        body = [endpoint_name]
```
Itay4:
```suggestion
        body = endpoint_name
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+def query_events(client: Client, args: dict) -> Tuple[str, Dict, Dict]:
+    start_time = args.get('start_time')
+    end_time = args.get('end_time')
+    logic = args.get('logic')
+    entity_type = args.get('entity_type')
+    column = args.get('column')
+    value = args.get('value')
+    operator = args.get('operator')
+    limit = args.get('limit')
+    additional_filter = args.get('additional_filter')
```
Itay4:
```suggestion
    additional_filter_string = args.get('additional_filter')
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    if additional_filter:
+        add_filter = additional_filter.split()
```
Itay4:
```suggestion
        additional_filter_split = additional_filter_string.split()
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        add_filter = {
```
Itay4:
```suggestion
        additional_filter = {
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'filterType': 'criteria',
+            'column': add_filter[0],
```
Itay4: need to add some validation here so we wont get an `IndexError` here

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    job_id = args.get('job_id')
+    job_result_id = args.get('job_result_id')
+    headers = ['ID', 'FileName', 'FilePath', 'MD5Hash', 'FileSize', 'HostName', 'HostIP', 'AgentID']
+
+    response = client.file_search_results_metadata(job_id, job_result_id)
+    if not response.get('success'):
+        return f'Could not find results for this job ID.', {}, {}
+    data = response.get('data', {}).get('jobResultInfos', [])
+    if not data:
+        return 'Check the job status, it might be still running', {}, {}
```
Itay4:
```suggestion
        return 'No results found.\nCheck the job status, it might be still running.', {}, {}
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    }
+    human_readable = tableToMarkdown('Fidelis Endpoint file results metadata', contents, headers=headers, removeNull=True)
+
+    return human_readable, entry_context, response
+
+
+def get_file_command(client: Client, args: dict):
+    file_id: str = str(args.get('file_id'))
+    response = client.get_file(file_id)
+    attachment_file = fileResult('Fidelis_Endpoint.zip', response)
```
Itay4: why did we decide about `zip` ?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    response = client.get_script_job_status(job_result_id)
+
+    results = response.get('data', {}).get('targets', [])
+
+    for result in results:
+        contents.append({
+            'JobResultID': result.get('jobResultId'),
+            'Name': result.get('name'),
+            'Status': result.get('status'),
+            'JobName': response.get('data').get('jobName')  # type: ignore
```
Itay4:
```suggestion
            'JobName': response.get('data', {}).get('jobName')
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'ProcessStartTime': event.get('processStartTime'),
+            'EventIndex': event.get('eventIndex'),
+            'IndexingTime': event.get('indexingTime'),
+            'NetworkDirection': event.get('networkDirection'),
+            'EntityType': event.get('entityType'),
+            'StartTime': event.get('startTime')
+        })
+
+    entry_context = {'FidelisEndpoint.Query(val.ParentID && val.ParentID === obj.ParentID)': context}
+    human_readable = tableToMarkdown('Fidelis Endpoint query results for the DNS request', contents,
```
Itay4: add `headers`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'ProcessStartTime': event.get('processStartTime'),
+            'EventIndex': event.get('eventIndex'),
+            'IndexingTime': event.get('indexingTime'),
+            'NetworkDirection': event.get('networkDirection'),
+            'EntityType': event.get('entityType'),
+            'StartTime': event.get('startTime')
+        })
+
+    entry_context = {'FidelisEndpoint.Query(val.TargetID && val.TargetID === obj.TargetID)': context}
+    human_readable = tableToMarkdown('Fidelis Endpoint query results for the DNS request by server IP', contents,
```
Itay4: add `headers`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'DnsAnswer': event.get('dnsAnswer'),
+            'ProcessStartTime': event.get('processStartTime'),
+            'EventIndex': event.get('eventIndex'),
+            'IndexingTime': event.get('indexingTime'),
+            'EntityType': event.get('entityType'),
+            'StartTime': event.get('startTime')
+        })
+
+    entry_context = {'FidelisEndpoint.Query(val.PID && val.PID === obj.PID)': context}
+    human_readable = tableToMarkdown('Fidelis Endpoint query events result', contents,
```
Itay4: add headers

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    incidents: list = list()
+    response = client.list_alerts()
+    alerts = response.get('data', {}).get('entities', [])
+    if alerts:
+        last_incident_id = last_run.get('id', 0)
+        # Creates incident entry
+        incidents = [{
+            'name': f"Fidlie Endpoint alert: {alert.get('id')}",
+            'occurred': alert.get('createDate'),
+            'severity': alert.get('severity'),
```
Itay4: removed the severity

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'rawJSON': json.dumps(alert)
+        } for alert in alerts if alert.get('id') > last_incident_id and alert.get('severity') == severity]
+        # New incidents fetched
+        if incidents:
+            last_incident_id = alerts[-1].get('id')
+            new_last_run.update({'id': last_incident_id})
```
Itay4: lets store the latest fetch time and fetch by it

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    if not host and not ip_address:
+        return f'You must provide either ip_address or host', {}, {}
+
+    contents = []
+    context_standards = []
+    headers = ['ID', 'HostName', 'IpAddress', 'OS', 'MacAddress', 'Isolated', 'LastContactDate', 'AgentInstalled',
+               'AgentVersion', 'OnNetwork', 'AV_Enabled', 'Groups', 'ProcessorName']
+    response = client.get_host_info(host, ip_address)
+    hosts = response.get('data', {})
+    if not hosts:
```
Itay4: need to check if `entities` is empty for no results

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'AgentVersion': host.get('agentVersion'),
+            'AV_Enabled': host.get('aV_Enabled'),
+            'Isolated': host.get('isolated'),
+            'OnNetwork': host.get('onNetwork'),
+            'Groups': host.get('groups'),
+            'ProcessorName': host.get('processorName')
+        })
+
+        context_standards.append({
+            'HostName': host.get('hostName'),
```
Itay4:
```suggestion
            'Hostname': host.get('hostName'),
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+            'ID': host.get('id'),
+            'IpAddress': host.get('ipAddress'),
```
Itay4:
```suggestion
            'IPAddress': host.get('ipAddress'),
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+    endpoint_id = get_endpoint_id(client, endpoint_ip, endpoint_name)
+
+    response = client.execute_script(script_id, endpoint_id, answer, time_out, additional_answer)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Script(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        script_id = LIST_PROCESSES_MACOS
+
+    response = client.list_process(script_id, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Process(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        script_id = KILL_PROCESS_MAC_LINUX
+
+    response = client.kill_process(script_id, pid, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Process(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        script_id = DELETE_FILE_MAC_LINUX
+
+    response = client.delete_file(script_id, file_path, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Script(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        script_id = NETWORK_ISOLATION_MAC_LINUX
+
+    response = client.network_isolation(script_id, allowed_server, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Isolation(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1707 @@
...
+        script_id = REMOVE_NETWORK_ISOLATION_MAC_LINUX
+
+    response = client.remove_network_isolation(script_id, time_out, endpoint_id)
+    context = {
+        'ID': script_id,
+        'JobID': response.get('data')
+    }
+    entry_context = {'FidelisEndpoint.Isolation(val.ID && val.ID === obj.ID)': context}
+
+    return 'The job has been executed successfully', entry_context, response
```
Itay4: lets return the job ID in the human readable as well

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1583 @@
...
+      type: Boolean
+  - arguments:
+    - default: false
+      description: The host to search the file in (for multiple hosts use (comma separated)
+      isArray: true
+      name: host
+      required: false
+      secret: false
+    - default: false
+      description: MD5 to search for (for multiple hashes use (comma separated)
```
Itay4: lets suggest here how this can be retrieved

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Shellyber: How can we do that? since we need all this escaped characters. 
Also why do it in first place?

Itay4: see my last commit ive pushed, i think it makes the code more readable

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+def fetch_incidents(client: Client, fetch_time: str, fetch_limit: str, last_run: Dict) -> Tuple[List, Dict]:
+    # handle first time fetch
+    last_run = last_run.get('time')
+    if not last_run:
+        last_run, _ = parse_date_range(fetch_time, to_timestamp=True)
+
+    current_fetch = last_run
+    incidents = []
+    last_fetch_date_string = timestamp_to_datestring(last_run, '%Y-%m-%dT%H:%M:%S.%fZ')
+    start_time = parse_timestamp(last_fetch_date_string)
```
Itay4: if i understand correctly, this function strips the milliseconds from the timestamp?
if so, why not just pass `%Y-%m-%dT%H:%M:%SZ` to be the date format?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+        }
+        incident_date = int(dateutil.parser.parse(alert.get('createDate')).timestamp())*1000
+        # update last run
+        if incident_date > last_run:
+            incidents.append(incident)
+
+        if incident_date > current_fetch:
+            current_fetch = incident_date
+
+    demisto.setLastRun({'time': current_fetch})
```
Itay4: `setLastRun` should not be called here, but in the main function
return `last_run` instead

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+    for alert in alerts:
+        alert_id = str(alert.get('id'))
+        incident = {
+            'name': f"Fidlie Endpoint alert: {alert_id}",
+            'occurred': alert.get('createDate'),
+            'rawJSON': json.dumps(alert)
+        }
+        incident_date = int(dateutil.parser.parse(alert.get('createDate')).timestamp())*1000
+        # update last run
+        if incident_date > last_run:
```
Itay4: why is this needed? we are querying only for incidents that were created after the last fetch

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+    current_fetch = last_run
+    incidents = []
+    last_fetch_date_string = timestamp_to_datestring(last_run, '%Y-%m-%dT%H:%M:%S.%fZ')
+    start_time = parse_timestamp(last_fetch_date_string)
+    response = client.list_alerts(limit=fetch_limit, start_date=start_time)
+    alerts = response.get('data', {}).get('entities', [])
+    for alert in alerts:
+        alert_id = str(alert.get('id'))
+        incident = {
+            'name': f"Fidlie Endpoint alert: {alert_id}",
```
Itay4:
```suggestion
            'name': f"Fidelis Endpoint alert {alert_id}",
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+            'occurred': alert.get('createDate'),
+            'rawJSON': json.dumps(alert)
+        }
+        incident_date = int(dateutil.parser.parse(alert.get('createDate')).timestamp())*1000
```
Itay4:
```suggestion
        alert_create_date_timestamp = int(dateutil.parser.parse(alert.get('createDate')).timestamp())*1000
```

Itay4: you're getting `createDate` twice, once for `occurred` and once in order to convert it to timestamp
instead initialize a variable with it

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+
+def parse_timestamp(time_):
+    if len(time_[time_.rfind('.')+1:time_.rfind('Z')]) > 3:
+        new_time = time_[:time_.rfind('.') + 4] + 'Z'
+    return new_time
+
+
+def fetch_incidents(client: Client, fetch_time: str, fetch_limit: str, last_run: Dict) -> Tuple[List, Dict]:
+    # handle first time fetch
+    last_run = last_run.get('time')
```
Itay4: dont use same variable name.. change to something like `last_fetch_alert_time`

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
...
+    # get the service API url
+    base_url = urljoin(demisto.params().get('url'), '/Endpoint/api')
+
+    verify_certificate = not demisto.params().get('insecure', False)
+
+    proxy = demisto.params().get('proxy', False)
+
+    LOG(f'Command being called is {demisto.command()}')
+    try:
+        client = Client(base_url, username=username, password=password, verify=verify_certificate, proxy
```
Itay4:
```suggestion
            fetch_limit = demisto.params().get('fetch_limit', '50')
```

Itay4:
```suggestion
            fetch_time = demisto.params().get('fetch_time', '3 days')
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
+import dateutil
+
+import demistomock as demisto
+from CommonServerPython import *  # noqa: E402 lgtm [py/polluting-import]
+from CommonServerUserPython import *  # noqa: E402 lgtm [py/polluting-import]
+"""GLOBALS/PARAMS"""
```
Itay4: remove this comment or place it above the globals

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1753 @@
+import dateutil
```
Itay4:
```suggestion
import dateutil.parser
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
Shellyber: Not completely. It strirps the last milliseconds and leaves only 3 digits after the .
It changes the timestamp so the integration will know how to handle it. 
Basically instead of this: 2020-02-15T14:23:38.000000Z
We will get: 2020-02-15T14:23:38.000Z

Itay4: in this case just dont pass a date format and it will use the default `"%Y-%m-%dT%H:%M:%S.000Z"` and will return you the requested result

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1746 @@
...
+            body['criteriaV3']['filter']['filters'].append(additional_filter)  # type: ignore
+
+        response = self._http_request('POST', url_suffix, params=params, json_data=body)
+        if response.get('error'):
+            raise Exception(response.get('error'))
+
+        return response
+
+
+def get_endpoint_id(client: Client, endpoint_ip=None, endpoint_name=None):
```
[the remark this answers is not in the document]
Shellyber: Here it's a List.

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+      description: The host to search the file in (for multiple hosts use (comma separated)
+      isArray: true
+      name: host
+      required: true
+      secret: false
+    - default: false
+      description: MD5 to search for (for multiple hashes use (comma separated)
+      isArray: true
+      name: md5
+      required: true
```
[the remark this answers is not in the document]
Shellyber: No. The Md5 is required according to the API doc.
Host is not.

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+      - Windows
+      - Linux
+      - macOS
+      required: true
+      secret: false
+    - default: false
+      description: The server that can communicate with the isolated endpoint.
+      isArray: false
+      name: allowed_server
+      required: true
```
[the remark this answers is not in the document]
Shellyber: The input here will be IP. and this will be the server that can communicate with the isolated endpoint.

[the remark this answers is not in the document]
Shellyber: I changed also the description of the command. So now this argument will be more clear to the user

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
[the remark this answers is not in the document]
Shellyber: No name in the UI for this. 
Basically each script has question with should be the argument and answer which is the Value of the argument. 
Should I change it to value instead?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+      description: The logic of the query.
+      isArray: false
+      name: logic
+      predefined:
+      - and
+      - or
+      required: true
+      secret: false
+    - default: false
+      description: The file hash to search for.
```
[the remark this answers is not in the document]
Shellyber: It's what we talked about.
The script results command also will return file hash but not always. It depends on the script

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.yml
```diff
@@ -0,0 +1,1505 @@
...
+      - process
+      - registry
+      - remoteThread
+      - script
+      - usb
+      - windowsevent
+      required: true
+      secret: false
+    - default: false
+      description: 'Column to query (for example: hash, name, remoteIP, dnsQuestion,
```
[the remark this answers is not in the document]
Shellyber: So each event type has its own fields - it's a long list so can't put it here but I will add it to the documentation.
Should I refer the user there?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1801 @@
...
+    start_time = args.get('start_time')
+    end_time = args.get('end_time')
+    logic = args.get('logic')
+    entity_type = args.get('entity_type')
+    column = args.get('column')
+    value = args.get('value')
+    operator = args.get('operator')
+    limit = args.get('limit')
+    additional_filter_string = args.get('additional_filter')
+    additional_filter = ''
```
Itay4:
```suggestion
    additional_filter = None
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1801 @@
...
+            'DnsQuestion': event.get('dnsQuestion'),
+            'DnsAnswer': event.get('dnsAnswer'),
+            'ProcessStartTime': event.get('processStartTime'),
+            'EventIndex': event.get('eventIndex'),
+            'IndexingTime': event.get('indexingTime'),
+            'EntityType': event.get('entityType'),
+            'StartTime': event.get('startTime')
+        })
+
+
```
Itay4: `current_fetch` is not being updated

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1801 @@
...
+                    'Name': obj.get('name'),
+                    'MD5': obj.get('mD5Hash'),
+                    'Path': obj.get('filePath'),
+                    'Size': obj.get('fileSize'),
+                    'Hostname': item.get('hostName')
+                }
+
+    entry_context = {
+        'FidelisEndpoint.File(val.ID && val.ID === obj.ID)': contents,
+        'File': file_standards
```
Itay4:
```suggestion
        outputPaths['file']: file_standards
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1801 @@
...
+            'Path': event.get('path'),
+            'Hostname': event.get('endpointName'),
+            'SHA1': event.get('hashSHA1'),
+            'SHA256': event.get('hashSHA256'),
+            'FileVersion': event.get('fileVersion')
+        })
+
+    entry_context = {
+        'FidelisEndpoint.Query(val.Hash && val.Hash === obj.Hash)': context,
+        'File': file_standards
```
Itay4:
```suggestion
        outputPaths['file']: file_standards
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1801 @@
...
+            'ID': host.get('id'),
+            'IpAddress': host.get('ipAddress'),
+            'OS': host.get('os'),
+            'MACAddress': host.get('macAddress'),
+            'Processor': host.get('processorName')
+        })
+
+    entry_context = {
+        'FidelisEndpoint.Host(val.ID && val.ID === obj.ID)': contents,
+        'Endpoint': context_standards
```
Itay4:
```suggestion
        'Endpoint(val.ID && val.ID === obj.ID)': context_standards
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1802 @@
...
+            'LocalIP': event.get('localIP'),
+            'LocalPort': event.get('localPort'),
+            'RemoteIP': event.get('remoteIP'),
+            'RemotePort': event.get('remotePort'),
+            'DnsQuestion': event.get('dnsQuestion'),
+            'DnsAnswer': event.get('dnsAnswer'),
+            'ProcessStartTime': event.get('processStartTime'),
+            'EventIndex': event.get('eventIndex'),
+            'IndexingTime': event.get('indexingTime'),
+            'EntityType': event
```
Itay4: if there will be no alerts fetched, we wont get in the loop, and `latest_alert_create_date` will be referenced before assignment
lets initalize `latest_alert_create_date = last_fetch_alert_time` before the loop

Itay4: add a comment here why this condition is needed

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1802 @@
...
+        outputPaths['file']: file_standards
+    }
+    human_readable = tableToMarkdown('Fidelis Endpoint file results metadata', contents, headers=headers, removeNull=True)
+
+    return human_readable, entry_context, response
+
+
+def get_file_command(client: Client, args: dict):
+    file_id: str = str(args.get('file_id'))
+    file_name: str = args.get('file_name')
```
Itay4:
```suggestion
    file_name: str = args.get('file_name', '')
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1802 @@
...
+        'FidelisEndpoint.File(val.ID && val.ID === obj.ID)': contents,
+        outputPaths['file']: file_standards
+    }
+    human_readable = tableToMarkdown('Fidelis Endpoint file results metadata', contents, headers=headers, removeNull=True)
+
+    return human_readable, entry_context, response
+
+
+def get_file_command(client: Client, args: dict):
+    file_id: str = str(args.get('file_id'))
```
Itay4:
```suggestion
    file_id: str = args.get('file_id', '')
```

Packs/FidelisEndpoint/TestPlaybooks/playbook-Fidelis_Endpoint.yml
```diff
@@ -0,0 +1,1077 @@
...
+      name: Get file
+      description: Get the file stream and download the file.
+      script: '|||fidelis-endpoint-get-file'
+      type: regular
+      iscommand: true
+      brand: ""
+    nexttasks:
+      '#none#':
+      - "14"
+    scriptarguments:
```
Itay4: please add `file_name` here

kirbles19:
@Shellyber Done. Need to regenerate the README from the updated YML.

Packs/FidelisEndpoint/README.md
```diff
@@ -0,0 +1,2375 @@
+## Overview
+---
+
+Use the Fidelis Endpoint integration for advanced endpoint detection and response (EDR) across Windows, Mac and Linux OSes for faster threat remediation.
+This integration was integrated and tested with version xx of Fidelis EDR
```
Itay4: replace `xx` with version

Packs/FidelisEndpoint/README.md
```diff
@@ -0,0 +1,2375 @@
+## Overview
+---
+
+Use the Fidelis Endpoint integration for advanced endpoint detection and response (EDR) across Windows, Mac and Linux OSes for faster threat remediation.
```
Itay4: are there any roles or permissions required to use the api ?

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1820 @@
...
+            'Isolated': host.get('isolated'),
+            'OnNetwork': host.get('onNetwork'),
+            'Groups': host.get('groups'),
+            'ProcessorName': host.get('processorName')
+        })
+
+        context_standards.append({
+            'HostName': host.get('hostName'),
+            'ID': host.get('id'),
+            'IpAddress': host.get('ipAddress'),
```
Itay4:
```suggestion
            'IPAddress': host.get('ipAddress'),
```

Packs/FidelisEndpoint/Integrations/FidelisEndpoint/FidelisEndpoint.py
```diff
@@ -0,0 +1,1820 @@
...
+            'AgentVersion': host.get('agentVersion'),
+            'AV_Enabled': host.get('aV_Enabled'),
+            'Isolated': host.get('isolated'),
+            'OnNetwork': host.get('onNetwork'),
+            'Groups': host.get('groups'),
+            'ProcessorName': host.get('processorName')
+        })
+
+        context_standards.append({
+            'HostName': host.get('hostName'),
```
Itay4:
```suggestion
            'Hostname': host.get('hostName'),
```

PR #5883, opened 2020-02-26
Fixing splunk test module

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22202

## Description
change test module to a lite query.

   - [ ] Yes
   - [x] No


Integrations/SplunkPy/SplunkPy.py
```diff
@@ -555,18 +555,11 @@ def splunk_parse_raw_command():
...
-        t = datetime.utcnow() - timedelta(days=3)
+        t = datetime.utcnow() - timedelta(hours=1)
         time = t.strftime(SPLUNK_TIME_FORMAT)
         kwargs_oneshot = {'count': 1, 'earliest_time': time}
         searchquery_oneshot = demisto.params()['fetchQuery']
-        oneshotsearch_results = service.jobs.oneshot(searchquery_oneshot, **kwargs_oneshot)  # type: ignore
-        reader = results.ResultsReader(oneshotsearch_results)
-        for item in reader:
-            if item:
-                demisto.results('ok')
```
esharf: if the query will be invalid we get an exceptin.
so we don't need to check results.

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -555,18 +555,11 @@ def splunk_parse_raw_command():
 
 def test_module(service):
     if demisto.params().get('isFetch'):
-        t = datetime.utcnow() - timedelta(days=3)
```
esharf: see the comment about the invalid query.
we don't need really results so one hour is enough.

Integrations/SplunkPy/SplunkPy.py
```diff
@@ -602,6 +595,7 @@ def main():
     # The command demisto.command() holds the command sent from the user.
     if demisto.command() == 'test-module':
         test_module(service)
+        demisto.results('ok')
```
esharf: Connect is already checked.

Integrations/SplunkPy/SplunkPy.py
IkaDemisto: But the main point in this test is to validate that the search query input is valid. Don't think that we should remove it

IkaDemisto: What will happen if query is empty, it is not mandatory field (for some reason) and has no default value. Try to run it  with empty query and see the error

esharf: `isFetch` should be set to `False`

kirbles19:
@esharf Done.

PR #5730, opened 2020-02-13
EDL & ExportIndicators enhancements

guyfreund:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/21955
fixes: https://github.com/demisto/etc/issues/21957

## Description
Both integrations now support basic authentication.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 
- [ ] Code Review

## Technical Writer Review
- [ ] [EDL CHANGELOG](https://github.com/demisto/content/blob/80a7486105405d381691f9f70fd40725eda4260c/Integrations/EDL/CHANGELOG.md)
- [ ] [EDL YML](https://github.com/demisto/content/blob/80a7486105405d381691f9f70fd40725eda4260c/Integrations/EDL/EDL.yml) - line 58 needs to be reviewed

- [ ] [ExportIndicators CHANGELOG](https://github.com/demisto/content/blob/80a7486105405d381691f9f70fd40725eda4260c/Packs/ExportIndicators/Integrations/ExportIndicators/CHANGELOG.md)
- [ ] [ExportIndicators YML](https://github.com/demisto/content/blob/80a7486105405d381691f9f70fd40725eda4260c/Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.yml) - line 57 needs to be reviewed
- [ ] [ExportIndicators DOCS](https://github.com/demisto/content/blob/e083f2398dbc29f74b6f82734606f08b3e6f5fcb/Packs/ExportIndicators/Integrations/ExportIndicators/README.md) - lines 30-31 needs to be reviewed

PR #5922, opened 2020-03-03
EDL - improve test module functionality 

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22035

## Description
 - changed SSL args to be passed as `SSLContext` object to `gevent` and catch relevant exception
 - improved the test module functionality as was approved in https://github.com/demisto/content/pull/5730

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No


Itay4:
changes in non related files are cause of windows/mac line seperator

PR #5930, opened 2020-03-04
Remove port parameter from BPA

darkushin:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#21483
fixes: demisto/etc#22659

## Description
Remove port parameter from panorama while preserving BC

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


bakatzir:
@amshamah419 please think about the CHANGELOG. when saying that you remove the port from the config, maybe state somehow that it won't affect existing customers. we would'nt want anyone to panic without a need.

amshamah419:
Linted and fixed an issue with the old incapsula yml so the build will pass.

PR #5957, opened 2020-03-08
[Fix] NOT the latest content release error message

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22626

## Description
Added an error message for when the release notes check fails to fetch latest release tag, urging the user to merge from master.

## Screenshots
image


release_notes.py
```diff
@@ -632,41 +632,45 @@ def main():
...
+                beta_res.append(beta_ans)
+
+        release_notes = "\n---\n".join(res)
+        beta_release_notes = "\n---\n".join(beta_res)
+        create_content_descriptor(args.version, args.asset_id, release_notes, args.github_token, beta_rn=beta_release_notes)
+
+        if missing_release_notes:
+            print_error("Error: some release notes are missing. See previous errors.")
+            sys.exit(1)
+    except RuntimeError:
```
anara123: RuntimeError is too generic. We can miss other exceptions here.
which line exactly causes this exception/error?

release_notes.py
```diff
@@ -632,41 +632,45 @@ def main():
...
+        print_error('Failed to get latest release tag. This may happen if your branch is not updated with '
```
anara123: For external contributions we either should ignore this error or give them exact instructions how to pull from our origin master. 
This error doesn't contain such instructions.

release_notes.py
DeanArbel: @anara123 it shouldn't be too generic as it uses the exception that is being thrown in `run_command` function, which is called in line 636

release_notes.py
DeanArbel: I added the following instructions:
`Please run "git origin/master"`

DeanArbel:
@kirbles19 would you mind giving in your 2 cents for the readability of the error message? (see the screenshot above)

PR #5974, opened 2020-03-09
Fix Whois Test Playbook

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#22660


ShahafBenYakir:
So we used the retry mechanism? Will this solve the issue? Didn't we say that after the first attempt fails all trailing attempts will fail as well no?

PR #5990, opened 2020-03-10
Add `TIMELINE_CATEGORY` Constant

avidan-H:
## Status
- [x] Ready

## Description
~~Create a constant variable in CSP to be used in all integration commands that return timeline data as the value for the `"Category"` field (since it should always be `"Integration Update"`).~~
When timeline data is returned via the `return_outputs` convenience function, if the timeline dict(s) do not include the `"Category"` key and value, the dict(s) will be updated with that key set with the value of `"Integration Update"`.

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No - server version 5.5.0 hasn't been released yet.

## Must have
- [ ] Tests
- [ ] Documentation


Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -25,6 +25,9 @@
 except Exception:
     pass
 
+# when returning an entry using `demisto.results` or `return_outputs` that has timeline data,
+# use the TIMELINE_CATEGORY constant as the value for the "Category" field of the timeline dictionary
+TIMELINE_CATEGORY = 'Integration Update'
```
Itay4: if this is always the case, why are we not setting it ourselves in `demisto.results` ?

avidan-H: Because there are, in theory, cases where you wouldn't use this category value. For instance, let's say you're doing stuff (calling `demisto.results()` or `return_outputs`) from a script - the category would probably be different (could be `"Automation Update"` maybe? Who knows?). Plus, I imagine that new categories will be added in the future.

Itay4: if there will be such case we will handle it, but for now i want to ease the process of development
if the only possible value for the field is currently, that is demisto 5.5, `Integration Update` then lets set it in `results`

avidan-H: tov 👍

avidan-H:
Closing and reopening PR from the purged master branch.

PR #5993, opened 2020-03-10
Added base documentation for F5 Firewall integration

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22381

## Description
base documentation only because we don't have instance.


kirbles19:
@yaakovi I reviewed and updated, but it still needs a bunch of info (I assume we're waiting for an instance). Not sure if I should add the docs-approved label...

yaakovi:
added `docs-approved` label.
once we get access to an instance, we can improve the docs.

yaakovi:
forced push as the build fails on BC check for argument name change.
not a BC break since the original argument name was invalid (include spaces)

PR #6003, opened 2020-03-11
Added O365 fields and mapping

teizenman:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22141

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [X] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No


Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -169,7 +150,11 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
                     "rawJSON": raw_data,
                     "fields": {
                         "port": argToList(item.get('tcpPorts', '')),
-                        "subfeed": item.get('serviceArea', '')
+                        "subfeed": item.get('serviceArea', ''),
```
avidan-H: Aren't you changing `subfeed` to `service` in your other PR? Worthwhile to preemptively change it here even if that PR hasn't been merged yet.

Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -169,7 +150,11 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
...
+                        "office365expressroute": item.get('expressRoute', ''),
```
avidan-H: We should only set the boolean values if they are returned, so don't set a default, and don't map to the indicator field _unless_ a non-`None` boolean value is returned from the `.get()`.

Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -169,7 +150,11 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
...
+                        "office365category": item.get('category', ''),
+                        "office365required": item.get('required', ''),
```
avidan-H: Same here.

teizenman:
@avidan-H all were fixed

Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -159,22 +155,34 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
...
                 for key, val in item.items():
                     if key not in ['ips', 'urls']:
                         raw_data.update({key: val})
+
+                indicator_mapping_fields = {
+                    "port": argToList(item.get('tcpPorts', '')),
+                    "service": item.get('serviceArea', '')
+                }
+
+                if item.get('expressRoute', None):
```
avidan-H:
```suggestion
                if item.get('expressRoute'):
```
The default value of `.get()` is `None` - no need to explicitly set it to `None`. The same goes for wherever else you explicitly set the default value to `None`.

PR #6029, opened 2020-03-13
Pack metadata fixes

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready

## Description
We should never add empty string as tag in `pack_metadata.json`. Need to add validation for it in later step, it causes the server to display empty data as tags in UI.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/Active_Directory_Query/pack_metadata.json
+++ b/Packs/Active_Directory_Query/pack_metadata.json
@@ -10,9 +10,7 @@
     "categories": [
         "Data Enrichment & Threat Intelligence"
     ],
-    "tags": [
-        ""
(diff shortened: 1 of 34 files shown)


liorblob:
@IkaDemisto this comes from the SDK, need to fix it there.

amshamah419:
Nice

amshamah419:
Guardrails is down? Really?

IkaDemisto:
yes  👎

liorblob:
This corona virus man...

amshamah419:
Next we will have to quarantine the build...

PR #6037, opened 2020-03-15
Fix typos in AlienVault

liorblob:
## Status
- [ ] In Progress
- [X] Ready

Is it just me or the `t` comes naturally after `alien`?


amshamah419:
I'm glad we are all in agreement

PR #6048, opened 2020-03-15
Packify Slack

liorblob:
## Status
- [ ] In Progress
- [X] Ready


liorblob:
@Itay4 there's a separate README for SlackAsk as well...

Itay4:
yea but we need to have one README for a pack, so lets just append the two readmes into one

PR #6057, opened 2020-03-16
Remove `hidden` fields from AWS Feed YAML file

avidan-H:
## Status
- [x] Ready

## Description
A few of the `AWS Feed` integration's parameters were accidentally set as hidden. Removed all the `hidden` fields from the YAML file.

## Screenshots
image
As you can see, the `feedExpirationPolicy` parameter is missing from the integration configuration panel.

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No - Server version 5.5.0 hasn't been released yet.


avidan-H:
@Itay4 with regards to the changelog, I don't think this is a fix that we want to appear in the content release notes (right?), so I just put a `-`, is that correct?

Itay4:
> @Itay4 with regards to the changelog, I don't think this is a fix that we want to appear in the content release notes (right?), so I just put a `-`, is that correct?

correct

PR #6061, opened 2020-03-16
Fixed aws-ec2-describe-vpcs command when the command returns empty list

adi88d:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22751

## Description
Fixed aws-ec2-describe-vpcs command - return the message `No VPCs were found.` when there is no results.

## Screenshots
image


Integrations/AWS-EC2/AWS-EC2.py
```diff
@@ -583,6 +583,10 @@ def describe_vpcs_command(args):
 
     response = client.describe_vpcs(**kwargs)
 
+    if len(response['Vpcs'] == 0):
```
amshamah419:
```suggestion
    if len(response['Vpcs']) == 0:
```

PR #6093, opened 2020-03-19
Bambenek update

teizenman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22740

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [X] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/FeedBambenekConsulting/Integrations/FeedBambenekConsulting/FeedBambenekConsulting.py
```diff
@@ -39,9 +39,57 @@ def main():
             'mapping': {
                 'description': 'malwarefamily'
             }
+        },
+        'https://faf.bambenekconsulting.com/feeds/dga-feed.gz': {
+            'fieldnames': ['value', 'description',
+                           'date_created',
+                           'info'],
```
avidan-H: Why aren't we mapping all these other fields and only description? Off the top of my head, `date_created` could probably map to `firstseenbysource`. Please consult with @David-BMS - we should be mapping as many fields here as possible (this comment applies for all the new feeds you added here to Bambenek).

Packs/ApiModules/Scripts/CSVFeedApiModule/CSVFeedApiModule.py
```diff
@@ -155,6 +156,23 @@ def build_iterator(self, **kwargs):
...
+
+        Args:
+            url: Current feed's url.
+            raw_response: The raw response from the feed's url.
+
+        Returns:
+            List. List of lines from the feed content.
+        """
+        if self.feed_url_to_config and self.feed_url_to_config.get(url).get('is_zipped_file'):  # type: ignore
+            response_content = gzip.decompress(raw_response.content)
```
avidan-H: I think by using the `.decompress()` method, you could still run into the issue I opened [here](https://github.com/demisto/etc/issues/22325) of a memory error. Test against the dga zip and see - but if it does give you memory errors as I suspect, then look at using `gzip.open`.

teizenman: Ive tested all of them locally and it seemed to work just fine.
Do you think it still would be better doing it in another way?

Packs/FeedBambenekConsulting/Integrations/FeedBambenekConsulting/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+-
```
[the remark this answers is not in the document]
teizenman: I know, was waiting to see if the old ones need deprecation before updating it.

Packs/FeedBambenekConsulting/Integrations/FeedBambenekConsulting/FeedBambenekConsulting.py
teizenman: @avidan-H We would like to use @David-BMS mapping for the old feeds as well?

avidan-H: Yeah, speak to @David-BMS - we should be mapping as much data as possible in the code.

Packs/ApiModules/Scripts/CSVFeedApiModule/CSVFeedApiModule.py
avidan-H: See this [comment](https://github.com/demisto/etc/issues/22325#issuecomment-602117258) - if everything is working as is and you don't want to implement the enhancement right now, I'll still approve the PR after the updates to the mapping.

teizenman: I read about it for a little, but it doesnt seem to be different since I have to receive the full request and then read it fully.
So, for now it seems to be working fine so Ill keep it that way.

Packs/FeedBambenekConsulting/Integrations/FeedBambenekConsulting/FeedBambenekConsulting.yml
```diff
@@ -78,6 +69,24 @@ configuration:
...
+  options:
+  - http://osint.bambenekconsulting.com/feeds/c2-ipmasterlist.txt
+  - http://osint.bambenekconsulting.com/feeds/c2-ipmasterlist-high.txt
+  - http://osint.bambenekconsulting.com/feeds/c2-dommasterlist.txt
+  - http://osint.bambenekconsulting.com/feeds/c2-dommasterlist-high.txt
+  - https://faf.bambenekconsulting.com/feeds/dga-feed.gz
+  - https://faf.bambenekconsulting.com/feeds/dga-feed-high.gz
+  - https://faf.bambenekconsulting.com/feeds/dga/c2-masterlist.txt
+  - https://faf.bambenekconsulting.com/feeds/dga/c2-masterlist-high.txt
+  - https://faf.bambenekconsulting.com/feeds/sinkhole/latest.csv
```
ronykoz: maybe we should take the names from their site, please check this option
https://osint.bambenekconsulting.com/feeds/

PR #6096, opened 2020-03-19
Sane pdf reports - docker

agamm:
## Status
- [X] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/22495

## Description
The automation that passes data form the server reports to the docker container that runs the sane-reports binary. **This is an internal automation**.

## Technical Writing
https://github.com/demisto/content/blob/5d54bc4f3b73e01111d3371fd67f4904de8506d4/Scripts/SanePdfReport/CHANGELOG.md

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [X] 5.5.0

   - [X] No

## Must have
- [ ] Tests (on server)
- [ ] Documentation (deprecated - used internally)

## Additional Information
See `Scripts/SaneDocReportV2` for reference of the docx automation.
There were some minor changes to the SaneDoc Reports.


Scripts/SanePdfReport/SanePdfReport.py
```diff
@@ -0,0 +1,82 @@
...
+        'utf-8')
+    orientation = demisto.args().get('orientation', 'portrait')
+    resourceTimeout = demisto.args().get('resourceTimeout', '4000')
+    reportType = demisto.args().get('reportType', 'pdf')
+    headerLeftImage = demisto.args().get('customerLogo', '')
+    headerRightImage = demisto.args().get('demistoLogo', '')
+    pageSize = demisto.args().get('paperSize', 'letter')
+    disableHeaders = demisto.args().get('disableHeaders', '')
+
+    if DISABLE_LOGOS:
```
hirshbergshachar: Do we have an option that it won't be true?

agamm: When we fix the main repo (sane-reports).

agamm:
> Python

@hirshbergshachar any procedure? is it ok because we run in docker? (still could be a problem if someone has a kernel vuln)

Scripts/SanePdfReport/SanePdfReport.py
Bars92: please fix sane reports before or ignore the bug... there is no reason to add this const or code to address that

Scripts/SanePdfReport/SanePdfReport.py
```diff
@@ -0,0 +1,82 @@
...
+import subprocess
+from pathlib import Path
+
+WORKING_DIR = Path("/app")
+INPUT_FILE_PATH = 'sample.json'
+OUTPUT_FILE_PATH = 'out{id}.pdf'
+DISABLE_LOGOS = True  # Bugfix before sane-reports can work with image files.
+
+
+def fix_base64(base64str, filename):
```
Bars92: 👌

agamm: ❤️

Scripts/SanePdfReport/SanePdfReport.py
agamm: > please fix sane reports before or ignore the bug... there is no reason to add this const or code to address that

This literally ignores the bug. What did you have in mind? 
@hirshbergshachar are logos in the reports a must right now or can they wait for a future PR in `sane-reports`?

hirshbergshachar: @agamm it can wait for future PR

agamm: @Bars92 agreed?

PR #6098, opened 2020-03-19
Update Tanium V2 to support question text with parameters

adi88d:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22492

## Description
- Update parse_question method to support question text with parameters instead of using the `parameters` argument in `tn-ask-question` command, which means that now its possible to call this command:
`!tn-ask-question question-text="Get Folder Contents[c:\] from all machines"`

instead of this:
`!tn-ask-question question-text="Get Folder Contents from all machines" parameters="Folder Contents{folderPath=c:\}"`

- Fix parse_question_results method to support list in single column result.

## Screenshots
image

### DIFF
--- a/Integrations/Tanium_v2/Tanium_v2.py
+++ b/Integrations/Tanium_v2/Tanium_v2.py
@@ -649,7 +649,6 @@ def ask_question(client, data_args):
     parameters = data_args.get('parameters')
 
     body = client.parse_question(question_text, parameters)
-    print(body)
     id_, res = client.create_question(body)
     context = {'ID': id_}
     context = createContext(context, removeNull=True)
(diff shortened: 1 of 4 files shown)


Integrations/Tanium_v2/Tanium_v2.py
```diff
@@ -108,7 +108,38 @@ def parse_question(self, text, parameters):
...
+                            if param == '""':
+                                param = ''
+                            key = sensor_response['Parameters'][index]['Key']
+                            tmp_item['parameters'].append({
+                                'key': '||' + key + '||',
+                                'value': param
+                            })
+
+                        index = index + 1
+                    parameters_condition.append(tmp_item)
```
yaakovi: original looked like C...

```suggestion
                    for param, sensor_key in zip(parameters, sensor_response['Parameters']):
                        if param:
                            if param == '""':
                                param = ''
                            tmp_item['parameters'].append({
                                'key': '||' + sensor_key['Key'] + '||',
                                'value': param
                            })

                    parameters_condition.append(tmp_item)
```

kirbles19:
@adi88d Done.

PR #6112, opened 2020-03-22
Taxii collection list

teizenman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21581

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [X] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -511,6 +517,31 @@ def __init__(self, insecure: bool = True, polling_timeout: int = 20, initial_int
                 self.username = credentials.get('identifier', None)
                 self.password = credentials.get('password', None)
 
+        self.all_collections = self.get_all_collections()
+        if collection is None or collection == '':
+            return_error(f"No collection set. Here is a list of all accessible collections: "
+                         f"{str(self.all_collections)}")
```
DeanArbel: It seems like this is the only place you're using `self.all_collections`. If you're not using it I think it doesn't need to be a member. While at it, please move the call to `get_all_collections()` inside the condition scope, since we can save the request time it if the user provided a collection.

Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -511,6 +517,31 @@ def __init__(self, insecure: bool = True, polling_timeout: int = 20, initial_int
...
+        if self.discovery_service:
+            taxii_client = cabby.create_client(discovery_path=self.discovery_service)
+            if self.username:
+                taxii_client.set_auth(username=str(self.username), password=self.password, verify_ssl=self.verify_cert)
+            elif self.api_key:
+                taxii_client.set_auth(username=str(self.api_key), verify_ssl=self.verify_cert)
+            else:
+                taxii_client.set_auth(verify_ssl=self.verify_cert)
+
+            all_collections = taxii_client.get_collections()
```
DeanArbel: `get_collections` can trigger an exception, please check if the error returned by such an exception is informative when its triggered, see [the documentation](https://cabby.readthedocs.io/en/stable/api.html):
image

teizenman: Done

Integrations/FeedTAXII/FeedTAXII.py
teizenman: Well, I kept it that way so I can validate it in the test module as well.

Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -896,6 +931,10 @@ def interval_in_sec(val):
 
 
 def test_module(client, args):
+    if client.collection not in client.all_collections:
```
DeanArbel: Good job on adding this check, but as it is I still feel that `client.all_collections` can be a function instead of a member, as it'll be called only once every time the code runs, and will either exit right after or ignore its value.

Keep in mind this call alone can take a few seconds on a slow TAXII server (such as hailataxii).

teizenman: You are right

Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -896,6 +931,11 @@ def interval_in_sec(val):
 
 
 def test_module(client, args):
+    all_collections = client.get_all_collections()
+    if client.collection not in all_collections:
```
DeanArbel: Please add another level of error handling here - my main concern here is if there's a problem with the `Collection Management Service`, while the user provided a valid collection, he'll get an error though his instance is usable.
This is a valid scenario since all services are optional. See http://taxiiproject.github.io/releases/1.1/TAXII_Services_Specification.pdf

teizenman: According to the specification it seems like we can't really know if the list we are receiving is incomplete. If the issue was that we get an error then we could handle that, but an incomplete list is an issue that we can't really handle.
So, maybe a better error message will do the trick?

teizenman:
@kirbles19 Can you please go over the changes in the following files:

- https://github.com/demisto/content/blob/taxii-collection-list/Integrations/FeedTAXII/CHANGELOG.md

- https://github.com/demisto/content/blob/taxii-collection-list/Integrations/FeedTAXII/FeedTAXII_description.md

avidan-H:
@teizenman not a chance I'm reviewing this PR like this...
image
You can close this PR and reopen it as Lichtman suggested in #dmst-content-team channel and I'll review it then.

PR #6115, opened 2020-03-23
Revert "ip indicator class - opinionated way to create indicators in demisto"

anara123:
Reverts demisto/content#4633


glicht:
I am going to push a commit so all unit test run when CommonServer has been modified

PR #6117, opened 2020-03-23
Move id set to sdk

liorblob:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22897

## Description
Move the creation of the ID set in the build to the SDK.


liorblob:
@yuvalbenshalom Need to remove as part of https://github.com/demisto/etc/issues/22938, not currently because stuff might break, as we still have other files that depend on the id set file.

According to the build it looks ok, but I will do some more testing like you said.

PR #6120, opened 2020-03-23
Taxii Feed Fix

avidan-H:
## Status
- [x] Ready

## Description
Fixes a problem where instance variables were uninitialized in the `init` method of the `TAXIIClient` class, causing an `AttributeError` to be raised when making requests.

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


avidan-H:
@Itay4 you can see the unit-test I added in my first commit fails in the associated build.

avidan-H:
@kirbles19 can you please take a look at the CHANGELOG file linked [here](https://github.com/demisto/content/edit/taxii-feed-fix/Integrations/FeedTAXII/CHANGELOG.md?pr=%2Fdemisto%2Fcontent%2Fpull%2F6120)?

kirbles19:
@avidan-H Done, no change.

Integrations/FeedTAXII/FeedTAXII_test.py
```diff
@@ -95,3 +96,13 @@ def test_fetch_indicators(self, mocker):
             with open('FeedTAXII_test/TestCommands/indicators_results.json', 'r') as exp_f:
                 expected = json.load(exp_f)
                 assert res == expected
+
+
+class TestTaxiiClient:
+    # verify that unassigned instance variables are still instantiated
+    def test__send_request(self):
```
Itay4:
```suggestion
    def test_send_request(self):
```

Integrations/FeedTAXII/FeedTAXII_test.py
```diff
@@ -95,3 +96,13 @@ def test_fetch_indicators(self, mocker):
...
+        from FeedTAXII import TAXIIClient
+        client = TAXIIClient()
+        with requests_mock.Mocker() as m:
+            m.post('http://www.example.com', text='response does not matter')
```
Itay4: lets also add a test which asserts the right auth params are initialized

Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -523,7 +526,7 @@ def _send_request(self, url, headers, data, stream=False):
             data=data
         )
 
-        if self.username is not None and self.password is not None:
+        if self.username and self.password:
```
Itay4: shouldnt be `elif` of the prev condition ?

Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -503,16 +503,19 @@ def __init__(self, insecure: bool = True, polling_timeout: int = 20, initial_int
...
-        if credentials:
-            if '_header:' in credentials.get('identifier', None):
-                self.api_header = credentials.get('identifier', None).split('_header:')[1]
-                self.api_key = credentials.get('password', None)
-            else:
-                self.username = credentials.get('identifier', None)
-                self.password = credentials.get('password', None)
+        if not credentials:
+            credentials = {}
+        self.api_head, self.api_key, self.username, self.password = '', '', '', ''
```
Itay4: i prefer we do it in the condition
so in the condition we will init `username` and `password` to be empty
in the `else` clause we will init `api_header` and `api_key`

avidan-H: Unless I'm misunderstanding you - that still seems problematic - you could still have an `AttributeError` occur. Instance variables should be initialized in all cases, even if it's just to the empty string. I don't think we should be defining instance attributes on the fly. If I get an instance of a class, regardless of when or under what condition, there should be a well-defined list of instance attributes for that class, whether or not those instance attributes have non-null/0/empty-string values assigned to them or not, the _attributes_ themselves should always exist.

Integrations/FeedTAXII/FeedTAXII_test.py
avidan-H: I think they should be initialized in all cases.

avidan-H: ... so I don't think this makes sense

avidan-H:
Closing and reopening in a new branch based off of the purged master.

PR #6123, opened 2020-03-23
Improved base client exception parsing

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Description
When the base client gets a text error response- trying to get `res.json()` will cause ValueError and hide the true cause of the error which can be taken from `res.text`

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CHANGELOG.md
+++ b/Packs/Base/Scripts/CommonServerPython/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [Unreleased]
   - Added the argument *ignore_auto_extract* to the ***return_outputs*** command.
   - Added a default value to the indicator timeline field **Category** when a value is not provided in an entry's timeline data.
+  - Improved error message parsing of HTTP response in BaseClient.
+
 
 ## [20.3.4] - 2020-03-30
 - Added support for successful empty responses (status code 204) in the base client.
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -2567,8 +2567,9 @@ if 'requests' in sys.modules:
                         error_entry = res.json()
                         err_msg += '\n{}'.format(json.dumps(error_entry))
                         raise DemistoException(err_msg)
-                    except ValueError as exception:
-                        raise DemistoException(err_msg, exception)
+                    except ValueError:
+                        err_msg += '\n{}'.format(res.text)
+                        raise DemistoException(err_msg)
 
                 is_response_empty_and_successful = (res.status_code == 204)
                 if is_response_empty_and_successful and return_empty_response:
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
@@ -8,6 +8,7 @@ import sys
 import requests
 from pytest import raises, mark
 import pytest
+
 from CommonServerPython import xml2json, json2xml, entryTypes, formats, tableToMarkdown, underscoreToCamelCase, \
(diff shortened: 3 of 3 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2557,13 +2557,16 @@ def _http_request(self, method, url_suffix, full_url=None, headers=None,
                 if not self._is_status_code_valid(res, ok_codes):
                     err_msg = 'Error in API call [{}] - {}' \
                         .format(res.status_code, res.reason)
+                    if 'text/plain' in res.headers.get('Content-type'):
```
Itay4: i think we can just append `res.text` in the `except` clause, regardless of the content type
what do you say?

hod-alpert: I think not, because if the content type is `json` it will not be a readable string it will be a `json.dumps` string.
And the header should be a good indication of whether it's a text format or a json format string.

Itay4: yea but it was to be a valid json, then it wouldnt fail when trying to deserialize it..
im ok with it not being readable, but its better than the nothing we return now

Packs/Base/metadata.json
```diff
@@ -0,0 +1 @@
+[]
```
Itay4: remove this file

Packs/Base/Scripts/CommonServerPython/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+Fixed exception parsing for BaseClient
```
Itay4:
```suggestion
Improved error message parsing of HTTP response in BaseClient.
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2562,8 +2562,9 @@ def _http_request(self, method, url_suffix, full_url=None, headers=None,
                         error_entry = res.json()
                         err_msg += '\n{}'.format(json.dumps(error_entry))
                         raise DemistoException(err_msg)
-                    except ValueError as exception:
-                        raise DemistoException(err_msg, exception)
+                    except ValueError:
+                        err_msg += '\n{}'.format(res.text)
+                        raise DemistoException(err_msg)
```
Itay4:
```suggestion
                        raise DemistoException(err_msg, exception)
```

hod-alpert: Why is this necessary? Why should we care about the `ValueError`?
Don't you think it's blinding?

Itay4: what do you mean by blinding? it just prints it to the logs

hod-alpert: I mean that the value Error is not really the error we encountered, it's a consequence of the real error.

In the UI it looks like this
image

rather then like this:
image

hod-alpert:
About the unit test- I could add those, but there are no tests at all of the base_client as it is now, and the `_http_request` is a private method, and it's not affecting the behavior, just the string in the exception.

Is it really necessary? Seems a little like an overkill

Itay4:
> About the unit test- I could add those, but there are no tests at all of the base_client as it is now, and the `_http_request` is a private method, and it's not affecting the behavior, just the string in the exception.
> 
> Is it really necessary? Seems a little like an overkill

yea we have to have more unit tests for this, a lot of integrations use it

hod-alpert:
@Itay4 Added unittests

kirbles19:
@hod-alpert Done.

Itay4:
force merging due to non-related failing tests

PR #6129, opened 2020-03-23
CrowdStrike Falcon Streaming v2

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/11887

## Description
 - Tried to avoid using `integrationContext` as it would probably lead to many server calls. It can be done, but not sure we want to.
 - Set it as beta for now. I would like to gain some experience with it and get some feedback from users, as it is widely used.

## Minimum version of Demisto
- [x] 5.0.0 - long running container

   - [x] No

## Must have
- [ ] Tests - test is a bit tricky here, still looking what would be the right way to do it
- [x] Documentation


Beta_Integrations/CrowdStrikeFalconStreamingV2/CrowdStrikeFalconStreamingV2.py
```diff
@@ -0,0 +1,354 @@
...
+        stream.set_refresh_token(refresh_token)
+        async for event in stream.fetch_event(offset=offset, event_type=event_type):
+            event_metadata = event.get('metadata', {})
+            event_type = event_metadata.get('eventType', '')
+            event_offset = event_metadata.get('offset', '')
+            demisto.info(f'Fetching event with offset: {event_offset}')
+            incident_name = f'{event_type} - offset {event_offset}'
+            incident = [{
+                'name': incident_name,
+                'details': json.dumps(event)
```
yaakovi: shouldn't this be in `rawJson`?

Beta_Integrations/CrowdStrikeFalconStreamingV2/CrowdStrikeFalconStreamingV2.py
```diff
@@ -0,0 +1,354 @@
...
+                    elif res.status == CREATED_STATUS_CODE:
+                        try:
+                            body = await res.json()
+                            break
+                        except json.decoder.JSONDecodeError:
+                            raise RuntimeError(
+                                f'Failed to decode successful token retrieval response: {str(res.content)}'
+                            )
+                    else:
+                        body = await res.json()
```
yaakovi: remove (you have it inside the try section)

Beta_Integrations/CrowdStrikeFalconStreamingV2/CrowdStrikeFalconStreamingV2.py
```diff
@@ -0,0 +1,354 @@
...
+                        except json.decoder.JSONDecodeError:
+                            raise RuntimeError(
+                                f'Failed to decode token retrieval failure response: {str(res.content)}'
+                            )
+            if not body:
+                raise RuntimeError('Failed to retrieve token')
+            token = body.get('access_token')
+            self.expiry_time = body.get('expires_in', MINUTES_30) - TIME_BUFFER_1_MINUTE
+        if not token:
+            raise RuntimeError('Failed to retrieve token')
```
yaakovi: same error message as in line 237, might be hard to to debug from logs in case of error

kirbles19:
@Itay4 Done.

Beta_Integrations/CrowdStrikeFalconStreamingV2/CrowdStrikeFalconStreamingV2.py
Itay4: no, because the `rawJSON` is being deleted in later stage by the server unless given a server config flag

Itay4:
> great debug logs, really spread in the entire integration. have you checked if `@logger` play nice with `async` functions?

dont think it works well, but can use python native logging async functionality
this its needed here?

yaakovi:
> > great debug logs, really spread in the entire integration. have you checked if `@logger` play nice with `async` functions?
> 
> dont think it works well, but can use python native logging async functionality
> this its needed here?

no need, i was just wondering if you tried

PR #6132, opened 2020-03-23
Added a new playbook for failed login attempts with Slack

altmannyarden:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22959

## Description
Added a v2.1 playbook for failed logins with Slack interactions. The new playbook accepts both username and email, includes cosmetic changes, and prevents the unlikely scenario where the playbook would complete before all tasks are done.

## Minimum version of Demisto
5.0.0

No

## Must have
- [ ] Code Review
- [ ] Documentation


idovandijk:
Went over the playbook with @altmannyarden. Basically done a review, looks good.
@yaakovi I was wondering why she didn't have the checkbox to run hooks during commit. Is that because of the new SDK commands that already format+validate?
Also, @bakatzir - it's currently failing because of an issue with the SDK, I know you are aware of this issue (it looks for an Else path for conditions, but we don't need such path here). How should we proceed?

bakatzir:
There are 2 issues:
1. schema validation. the `sla` is part of the schema but the `slareminder` is not (yet). we can add it upon a github issue request(so please do if needed), as we try to keep the schema to the only needed fields.
```
sla:
      hours: 2
      days: 0
      weeks: 0
    slareminder:
      hours: 1
      days: 0
      weeks: 0
```
``
in <SchemaError: error code 2: Schema validation failed:
 - Key 'slareminder' was not defined. Path: '/tasks/2'.: Path: '/'>
```

2. the conditions. AFAIK, @DeanArbel has tended to all the issues, htough I did not review the PRs. deam - care to have a look?
```
Playbook conditional Ask task with id:2 has task with unreachable next task condition "yes". Please remove this task or add this condition to condition task with id:2.
Playbook conditional Ask task with id:2 has unhandled condition: Yes
Playbook conditional Ask task with id:6 has task with unreachable next task condition "no". Please remove this task or add this condition to condition task with id:6.
Playbook conditional Ask task with id:6 has task with unreachable next task condition "yes". Please remove this task or add this condition to condition task with id:6.
Playbook conditional Ask task with id:6 has unhandled condition: Yes,No
```

yaakovi:
* checkbox for running hooks will appear automatically once there is a hook file under `.git/hooks/pre-commit`. if yarden uses windows, give her your copy as we know it works. otherwise she can create a symlink in that path to `.hooks/pre-commit` .

* all condition paths should be filled, what do you expect the playbook do in the case of else clause? (nothing? if so, connect to end of playbook)

* change logs for new content entities is generated in the content release process (clear release notes stage).

idovandijk:
> * checkbox for running hooks will appear automatically once there is a hook file under `.git/hooks/pre-commit`. if yarden uses windows, give her your copy as we know it works. otherwise she can create a symlink in that path to `.hooks/pre-commit` .
> * all condition paths should be filled, what do you expect the playbook do in the case of else clause? (nothing? if so, connect to end of playbook)
> * change logs for **new** content entities is generated in the content release process (clear release notes stage).

@altmannyarden please note Yaakovi's first point. Also, let's open an issue for the issue with `slareminder` (tell me if you need help)
@yaakovi I'm not sure your 2nd point is always correct.. we have cases where the condition is one or the other, but there is no "else" option, as @Arsenikr had in one of his playbooks if I remember correctly. Was the final decision to leave that check there?

yaakovi:
@idovandijk @Arsenikr what was the playbook?

Arsenikr:
@yaakovi pan os commit configuration

DeanArbel:
@idovandijk went over the errors. Seems they're valid, and you need to fix the playbook. see my screenshots:
image
image

idovandijk:
@DeanArbel 
Actually this is case **INSENSITIVE**, and in addition to that - the labels always change everything to uppercase which would make it very confusing to have failures due to case sensitivity. The playbook works with the answer of Yes, and takes the "yes" path and was also tested by @altmannyarden 
image

DeanArbel:
@idovandijk more like a dev-bug for the sdk :)
I'm on it.

altmannyarden:
Created a new PR: https://github.com/demisto/content/pull/6190

PR #6143, opened 2020-03-24
MS Teams - Adding ring-user command

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22329

## Description
Added ring-user command which rings a user on Teams.
This does not create an actual call but just a ring.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?


roysagi:
@Arsenikr @Zivsolo - FYI

Integrations/MicrosoftTeams/MicrosoftTeams.yml
```diff
@@ -225,7 +225,33 @@ script:
...
+      required: true
+      secret: false
+    - default: false
+      description: The tenant ID for the Teams.
+      isArray: false
+      name: tenant_id
+      required: true
+      secret: false
+    - default: false
+      description: The callback URI for the call. The URI determins who will recieve
```
ronykoz: how will you get that as a user?

Integrations/MicrosoftTeams/MicrosoftTeams.py
```diff
@@ -1495,6 +1495,74 @@ def messages() -> Response:
...
+
+    Notes:
+        This is a ring only! no media plays in case the generated call is answered.
+
+    Returns:
+        None.
+    """
+    bot_id = demisto.params().get('bot_id')
+    tenant_id = demisto.args().get('tenant_id')
+    callback_uri = demisto.args().get('callback_uri', 'https://callback.url')
```
ronykoz: uri or url?
Add better explanation

Itay4:
looks like we missed `microsoft-teams` prefix for most of the commands here
all of the commands but the following should have it: `send-notification`, `mirror-investigation` and `close-channel`
please add it to the and deprecate the others and also add it to the ring command you added

Integrations/MicrosoftTeams/MicrosoftTeams.yml
roysagi: you are right there is no way I found to get this info. you can put a bot's callback_uri if you want the call to redirect to a different bot for some reason - I'll edit the description accordingly

Integrations/MicrosoftTeams/MicrosoftTeams.py
roysagi: it is called callback URI in the api but I can change to URL if you think it is more understandable

PR #6144, opened 2020-03-24
name -> indicatorName

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22945

## Description
`name` cliname is a system name, changed to indicatorName

## Minimum version of Demisto
- [x] 4.5.0

   - [x] No
This field was never in Demisto (XSOAR)


Itay4:
lets change also file name
are there other content entities that use that field ? if so, we need to update them as well

jochman:
> lets change also file name
> are there other content entities that use that field ? if so, we need to update them as well

I was unable to find any entities using it. Care to have a look?

Itay4:
also change the `name` and `prevName` to `Indicator Name`

avidan-H:
@jochman It looks like you already did something with name related indicator field... ?
image

jochman:
Field, not name.

avidan-H:
> Field, not name.

I don't understand what this means.

PR #6149, opened 2020-03-24
[dev-task] CircleCI: smarter test selection

DeanArbel:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18080
fixes: https://github.com/demisto/etc/issues/22709

## Description
Made the test collection a bit smarter:

1. When there are no collected tests, collect 3 random tests that are runnable
a. The tests and dependent integrations/scripts are not skipped.
b. The tests and dependent integrations/scripts don't have a fromversion which is greater than the tested env.
2. When there are tests, but they're all skipped / unrannble - collect 3 random tests in accordance with the previous step.

The way CommonServer test should be run is as follows:
1. Run all Content unit tests.
2. Run test playbooks in accordance with the changed files alongside CommonServer change.
3. If no change was introduced other than CommonServer, then choose 3 random test playbooks from python integrations, that are not skipped, and run them.

added: https://github.com/demisto/etc/issues/22709 implementation

   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation 

## Additional Changes
Rewrote UT in `test_configure_tests.py` to be able to test more easily.


DeanArbel:
@yuvalbenshalom @yaakovi Please note in the last build, 3 random tests were picked that should be able to run in `Two Before GA` - that's the intended behavior in this case.

image

DeanArbel:
continued in https://github.com/demisto/content/pull/6181

PR #6150, opened 2020-03-24
Update integration logos

Itay4:


Itay4:
build fails due to schema formatting of deprecated integration and non related test
merging

PR #6203, opened 2020-03-26
Mail sender Full TLS

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22549

## Description
So far Mail Sender integration supported semi TLS connection (the first creation on the SERVER object was without SSL) such that some client's mailboxes failed to configured in such a way. Added an option for Full TLS including the creation of the SERVER object.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests: Working on getting our own instance to run test PB 
- [ ] Documentation


ShahafBenYakir:
@glicht According to Sylvain Meras - smeras@paloaltonetworks.com (Don't think he has a Github user) who asked for this enhancement, there are some customers that needs Full TLS connection wheres ours is almost full. 

I added two different instances to https://dev3.demisto.works/#/settings/servers - Mail Sender Shahaf. 
The first supports regular TlS as we had so far where the second supports Full TLS and will fail for the regular TLS. You can go ahead and test that.

glicht:
See my comments at: https://github.com/demisto/etc/issues/22549#issuecomment-605658573

ShahafBenYakir:
@glicht Changed the implementation to be with drop-down instead of checkboxes.
The only possible issue is that for users that already configured the integration with TLS 
it will appear like this: 
Settings 2020-03-29 21-33-53

glicht:
It looks ok. Please do an integration upgrade check with 5.0 too.

Integrations/MailSenderNew/MailSenderNew.yml
```diff
@@ -25,11 +25,14 @@ configuration:
   name: fqdn
   required: false
   type: 0
-- defaultvalue: 'false'
-  display: Use TLS for connection
+- display: Use TLS for connection
```
glicht: Please use display name `Connection Security` with options:
* None
* STARTTLS
* SSL/TLS

By default None should be selected.

ShahafBenYakir:
@glicht Should be fine now, have a look.
I did an integration update and will do the same on 5.0.

ShahafBenYakir:
@kirbles19 Please go over the MD and YML. 
We will also need to add an explanation of the new dropdown configuration parameter in the integration docs which replaces the old one called `Use TLS for connection`.
Something like this should work:
### connection security: 
- `None` will have no connection security at all. 
- `SSL/TLS` first establishes a secure `SSL/TLS` connection and conducts SMTP over that connection.
- STARTTLS starts with unencrypted SMTP and then switches to SSL/TLS to ensure security.

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -333,19 +334,27 @@ def main():
...
-        SERVER = SMTP(demisto.getParam('host'), int(demisto.params().get('port', 0)), local_hostname=FQDN)
-        SERVER.ehlo()
+
         # TODO - support for non-valid certs
-        if demisto.getParam('tls'):
-            SERVER.starttls()
+        if TLS == 'SSL/TLS':
+            SERVER = SMTP_SSL(demisto.getParam('host'), int(demisto.params().get('port', 0)), local_hostname=FQDN)
+        else:
+            SERVER = SMTP(demisto.getParam('host'), int(demisto.params().get('port', 0)), local_hostname=FQDN)  # type: ignore
```
glicht, line 348: When using `type: ignore` use the syntax: `# type: ignore[<error-name>]`  See: https://xsoar.pan.dev/docs/integrations/linting#mypy

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -333,19 +334,27 @@ def main():
...
+
+        SERVER.ehlo()   # type: ignore
```
glicht, line 350: same remark regarding mypy ignore

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -333,19 +334,27 @@ def main():
...
+        # For BC purposes where TLS was a checkbox (no value only true or false) if TLS=True or TLS='STARTTLS' we enter
+        # this condition, otherwise it means TLS is not configured (TLS=False) or is set to 'SSL/TLS' or 'None'.
+        if TLS is True or TLS == 'STARTTLS':
+            SERVER.starttls()   # type: ignore
```
glicht, line 354: fix mypy ignore

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -333,19 +334,27 @@ def main():
...
         user, password = get_user_pass()
         if user:
-            SERVER.login(user, password)
+            SERVER.login(user, password)    # type: ignore
```
glicht, line 357: fix mypy ignore

Integrations/MailSenderNew/MailSenderNew.py
```diff
@@ -360,8 +369,8 @@ def main():
             msg['Subject'] = 'Test mail from Demisto'
             msg['From'] = FROM
             msg['To'] = FROM
-            SERVER.sendmail(FROM, [FROM], msg.as_string())
-            SERVER.quit()
+            SERVER.sendmail(FROM, [FROM], msg.as_string())  # type: ignore
+            SERVER.quit()   # type: ignore
```
glicht, line 373: fix mypy ignore

kirbles19:
@ShahafBenYakir Done.

PR #6220, opened 2020-03-26
[dev-task] CircleCI: smarter test selection

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18080
fixes: https://github.com/demisto/etc/issues/22709

## Description
Made the test collection a bit smarter:

1. When there are no collected tests, collect 3 random tests that are runnable
a. The tests and dependent integrations/scripts are not skipped.
b. The tests and dependent integrations/scripts don't have a fromversion which is greater than the tested env.
~2. When there are tests, but they're all skipped / unrannble - collect 3 random tests in accordance with the previous step.~

The way CommonServer test should be run is as follows:
1. Run all Content unit tests.
2. Run test playbooks in accordance with the changed files alongside CommonServer change.
3. If no change was introduced other than CommonServer, then choose 3 random test playbooks from python integrations, that are not skipped, and run them.

added: https://github.com/demisto/etc/issues/22709 implementation

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation 

## Additional Changes
Rewrote UT in `test_configure_tests.py` to be able to test more easily.


Tests/scripts/configure_tests.py
```diff
@@ -714,7 +872,9 @@ def create_test_file(is_nightly, skip_save=False):
             last_commit, second_last_commit = commit_string.split()
             files_string = run_command("git diff --name-status {}...{}".format(second_last_commit, last_commit))
 
-        tests = get_test_list(files_string, branch_name)
+        with open('./Tests/ami_builds.json', 'r') as ami_builds:
+            two_before_ga = json.load(ami_builds).get('TwoBefore-GA', '0').split('-')[0]
```
ronykoz: Add a comment for why you are doing this

Tests/scripts/configure_tests.py
```diff
@@ -670,30 +825,33 @@ def get_test_list(files_string, branch_name):
...
+            tests = tests.union(get_runnable_tests(tests_num=RANDOM_TESTS_NUM, conf=conf, id_set=id_set,
+                                                   server_version=two_before_ga_ver))
         else:
             print_warning("Running Sanity check only")
+            tests = get_runnable_tests(tests_num=RANDOM_TESTS_NUM, conf=conf, id_set=id_set,
+                                       server_version=two_before_ga_ver)
             tests.add('DocumentationTest')  # test with integration configured
             tests.add('TestCommonPython')  # test with no integration configured
 
+    elif changed_common:
```
ronykoz: otherwise, you will add this only when there are no tests
```suggestion
    if changed_common:
```

Tests/scripts/configure_tests.py
```diff
@@ -646,14 +685,130 @@ def get_test_from_conf(branch_name):
...
+        if not conf:
+            with open("./Tests/conf.json", 'r') as conf_file:
+                conf = json.load(conf_file)
+        for test_id in test_ids:
+            if is_test_runnable(test_id, id_set, conf, server_version):
+                return True
+    return False
+
+
+def get_runnable_tests(tests_num, conf=None, id_set=None, server_version='0'):
```
ronykoz:
```suggestion
def get_random_tests(tests_num, conf=None, id_set=None, server_version='0'):
```

Tests/scripts/configure_tests.py
```diff
@@ -646,14 +685,130 @@ def get_test_from_conf(branch_name):
...
+        with open("./Tests/conf.json", 'r') as conf_file:
+            conf = json.load(conf_file)
+    skipped_integrations = conf['skipped_integrations'].keys()
+    # check if all integrations are skipped
+    if all(integration_id in skipped_integrations for integration_id in integration_ids):
+        return False
+    return True
+
+
+def is_any_test_runnable(test_ids, conf=None, id_set=None, server_version='0'):
```
ronykoz: not used

Tests/scripts/configure_tests.py
```diff
@@ -646,14 +685,130 @@ def get_test_from_conf(branch_name):
     return tests
 
 
-def get_test_list(files_string, branch_name):
+def is_test_runnable(test_id, id_set, conf, server_version):
```
ronykoz: i want to get notified when my test is skipped, lets fail the build on this

Tests/scripts/configure_tests.py
```diff
@@ -608,7 +618,35 @@ def update_test_set(tests, tests_set):
...
+    test_conf = next((test_conf for test_conf in test_conf_lst if (
+        test_conf.get('playbookID') == test_id
+        and is_runnable_in_server_version(from_v=test_conf.get('fromversion', '0'),
+                                          server_v=server_version,
+                                          to_v=test_conf.get('toversion', '99.99.99'))
+    )), None)
+    return test_conf
+
+
+def get_obj_from_obj_set(obj_id, obj_set, server_version='0'):
```
ronykoz: give a better name, what obj for what..

Tests/scripts/configure_tests.py
DeanArbel: As it  is this has to be elif, otherwise we'll have `TestCommonPython` twice. We can however remove 850, and do as you suggested, would that be better?

Tests/scripts/configure_tests.py
DeanArbel: true, was used, and will be used in a future branch, but I'll just copy it to it

Tests/scripts/configure_tests.py
DeanArbel: I'm not sure this is what we're looking for...
@yuvalbenshalom what do you think?

Tests/scripts/configure_tests.py
DeanArbel: `extract_matching_object_from_id_set`

DeanArbel:
@ronykoz you're right, that's why @yuvalbenshalom and I decided to not implement bullet 2 (I crossed it out).

Please see https://github.com/demisto/etc/issues/23172

After we merge this PR I'll open a PR for ignoring filtered envs - i.e. skip their creation and use.

ronykoz:
So we should also remove the fixes from the first issue

DeanArbel:
@ronykoz the fixes for the first issue still make sense, as this is a decision that was made in regards to it - after reviewing again what we seek in the content builds. 

I'll update the issue to reflect this.

yaakovi:
Bringing up instances will also check the content update.
For example, Updating a layout will not generate any test but might find a problem with the file

liorblob:
@DeanArbel Please move the tests files from hook validations to infrastructure tests. They do not belong there and this directory will be removed from content.

PR #6224, opened 2020-03-26
Update TLP Indicator Field's type to singleSelect

avidan-H:
## Status
- [x] Ready

## Description
This indicator field shouldn't be a free text field - should be single select from predetermined values of `RED`, `AMBER`, `GREEN`, `WHITE`.

## Minimum version of Demisto
- [x] 5.5.0

## Does it break backward compatibility?
   - [x] No - Server version 5.5.0 hasn't been released yet.

## Must have
- [ ] Tests
- [ ] Documentation

PR #6243, opened 2020-03-26
Delete Old TLP Indicator Field

avidan-H:
## Status
- [x] Ready

## Related Issues
related to https://github.com/demisto/content/pull/6224

## Description
Need to delete the old trafficlightprotocol indicator field before can make the new version of it that is of type singleSelect.

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No - server version 5.5.0 hasn't been released yet.


avidan-H:
Adding the `docs-approved` label seeing as there are no documents to approve.

PR #6266, opened 2020-03-28
MongoDB integration

jochman:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20026

## Minimum version of Demisto
- [x] 4.5.0
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

Documentation:
https://github.com/demisto/content/blob/mongodb-final/Integrations/MongoDB/README.md


Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -0,0 +1,347 @@
...
+        self,
+        urls: List[str],
+        username: str,
+        password: str,
+        database: str,
+        ssl: bool = False,
+        insecure: bool = False
+    ):
+        if insecure and not ssl:
+            return_error(f'"Trust any certificate (not secure)" must be ticked with "Use TLS/SSL secured connection"')
```
Itay4: why is this restriction? where is that documented on pymongo?

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -0,0 +1,347 @@
...
+    _ = client.create_collection(collection)
+    return (
+        f'MongoDB: Collection \'{collection}\' has been successfully created.',
+        None,
+    )
+
+
+def list_collections_command(client: Client, **kwargs) -> Tuple[str, dict, list]:
+    raw_response = client.db.list_collection_names()
+    if len(raw_response):
```
Itay4:
```suggestion
    if raw_response:
```

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -0,0 +1,347 @@
...
+            'MongoDB: All collections in database', raw_response, headers=['Collection']
+        )
+        outputs = {
+            'MongoDB.Collection(val.Name === obj.Name)': [
+                {'Name': collection} for collection in raw_response
+            ]
+        }
+        return readable_outputs, outputs, raw_response
+    else:
+        return "MongoDB: No results found", {}, raw_response
```
Itay4:
```suggestion
        return 'MongoDB: No results found', {}, raw_response
```

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
jochman: when you're using them both, that's the error raised.

Itay4: what error?

jochman:
```
ConfigurationError: ssl has not been enabled but the following ssl parameters have been set: ssl_cert_reqs. Please set `ssl=True` or remove.
```

PR #6270, opened 2020-03-29
Revert "Delete Old TLP Indicator Field"

avidan-H:
Reverts demisto/content#6243


avidan-H:
Adding `docs-approved` since there are no documents to be reviewed.

PR #6279, opened 2020-03-29
Fix Alienvault USM fetch incidents

adi88d:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22756

## Description
- Set page parameter in search alarms request to 0 instead of 1 to gets all the incidents.
**page parameter documentation:**
image

- Add one second to last incident occurred time to avoid fetching the last incident again.


Integrations/AlienVault_USM_Anywhere/AlienVault_USM_Anywhere.py
```diff
@@ -422,6 +422,12 @@ def fetch_incidents():
     if incidents:
         #  updating according to latest incident
         time_str = str(incidents[-1].get('occurred'))
+
+        # add one second to last incident occurred time to avoid duplications
+        occurred = datetime.strptime(time_str, '%Y-%m-%dT%H:%M:%S.000Z')
```
yaakovi: will `time_str` always have zero milliseconds? if not, truncate the string and parse only the beginning

kirbles19:
@adi88d Done.

PR #6282, opened 2020-03-29
Script LookupCSV 

adambaumeister:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
None

## Description
This PR adds a basic CSV lookup script, which provides a command for retrieving a specific row (or rows) from a CSV based on a provided column/value pair.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0
- Unknown - developed on 5.0.0

   - [ ] Yes
       - Further details:
   - [ ] No
   - Unknown. No?

## Must have
- [X] Tests
- [X] Documentation


adambaumeister:
re-opened with included changes from Master

Scripts/LookupCSV/LookupCSV.yml
```diff
@@ -0,0 +1,31 @@
...
+enabled: true
+args:
+  - name: entryID
+    required: true
+    description: EntryID of CSV file.
+  - name: header_row
+    description: CSV file has a header row
+  - name: column
+    description: Column to search for value in
+  - name: value
```
ronykoz, line 19: please make this required, as without it there is nothing to search for

Scripts/LookupCSV/LookupCSV.yml
```diff
@@ -0,0 +1,31 @@
...
+  - Utility
+comment: Parses a CSV and looks for a specific value in a specific column, returning
+  a dict of the entire matching row.
+enabled: true
+args:
+  - name: entryID
+    required: true
+    description: EntryID of CSV file.
+  - name: header_row
+    description: CSV file has a header row
```
ronykoz, line 16: please add optional values as it may seem that the meaning is the headers themselves,
And could you add the option to specify the headers in the arguments, as some CSV might come without it

Scripts/LookupCSV/LookupCSV.yml
```diff
@@ -0,0 +1,31 @@
...
+  - name: column
+    description: Column to search for value in
```
ronykoz, line 18: Should be stated that this could be an integer when no headers, and column name otherwise

Scripts/LookupCSV/LookupCSV.py
```diff
@@ -0,0 +1,115 @@
...
+            csv_data = search_lists(search_column, search_value, csv_data)
+
+    if not csv_data:
+        output = {
+            'LookupCSV.missing': search_value
+        }
+    else:
+        output = {
+            'LookupCSV.result': csv_data
+        }
```
ronykoz, line 104:
```suggestion
    if not csv_data:
        output = {
            'LookupCSV.missing': search_value
        }
    else:
        output = {
            'LookupCSV': {
                'FoundResult': True if csv_data else False,
                'Result': csv_data
            }
        }
```

Scripts/LookupCSV/LookupCSV.yml
adambaumeister: If you don't specify column, it simply extracts the entire CSV into json format and still works. Unfortunately (as far as I know) there's no way to have demisto args be dependent on each other so I can't make value be required.

Scripts/LookupCSV/LookupCSV.py
adambaumeister: Added.

Scripts/LookupCSV/LookupCSV.yml
adambaumeister: Good idea, have updated this description.

Scripts/LookupCSV/LookupCSV.yml
adambaumeister: without the headers it still parses - do you mean allow people to arbitarily map headers? But could get a little tricky as if the user passes in less than the number of cols in the row it will result in unexpected (to them) behavior. Thoughts?

ronykoz: I think that we can catch this case very easily, with a len(entered_header) == len(first_line.split(',')

But this one is not a must, your call

Scripts/LookupCSV/LookupCSV.yml
ronykoz: But then the script doesn't exactly do what it was meant to, I think that we should not return a result for this, or at least mention that no results were found(Both in the War room and in the context)

Scripts/LookupCSV/LookupCSV.yml
adambaumeister: This is a good idea, not a lot of extra code for a potential big boost in functionality. Added.

Scripts/LookupCSV/LookupCSV.yml
```diff
@@ -0,0 +1,37 @@
...
+      - "false"
+  - name: column
+    description: Column to search for value in, if not specified, entire CSV is parsed into the context.
+  - name: value
+    description: value to search for
+  - name: add_header_row
+    description: Extra row, in CSV format, to function as header if original does not contain headers
+outputs:
+  - contextPath: LookupCSV.result
+    description: List of result objects; either a list of dicts (with header_row) or a list of lists (no header row)
```
ronykoz, line 29:
```suggestion
  - contextPath: LookupCSV. Result
    description: List of result objects; either a list of dicts (with header_row) or a list of lists (no header row)
  - contextPath: LookupCSV. FoundResult
    description: Boolean, for whether the result was found in the CSV or not.
```

Scripts/LookupCSV/LookupCSV.py
```diff
@@ -0,0 +1,128 @@
...
+            for line in lines:
+                d = {}
+                row = line.split(",")
+                for i, h in enumerate(headers):
+                    d[h] = row[i]
+                csv_data.append(d)
+        elif add_row:
+            if len(lines[0].split(",")) != len(add_row.split(",")):
+                return_error(
+                    "Added row via add_row has invalid length.")
```
ronykoz, line 84:
```suggestion
                    "Added row via add_header_row has invalid length.")
```

Scripts/LookupCSV/LookupCSV.yml
adambaumeister: Added!

Scripts/LookupCSV/LookupCSV.py
adambaumeister: Fixed!

adambaumeister:
I've re-added the LookupCSV.missing key output as well as we use this to create and maintain a list in the context within a playbook of the stuff that fails to be found within a given CSV.

kirbles19:
@ronykoz Done.

ronykoz:
@adambaumeister as we talked about i'm closing this now,it's moved to another PR which will be merged in a few minutes https://github.com/demisto/content/pull/6597

PR #6283, opened 2020-03-29
Add default value when reading keys from data dictionary

adi88d:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22716

## Description
Add default value when reading keys from data dictionary after execute the commands: `sandblast-query` and `sb-query`


ronykoz:
@adi88d we should update the changelog please

kirbles19:
@adi88d Done.

PR #6295, opened 2020-03-29
Fixed instance_notifier to work with new PreGA

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Description
Changed instance notifier to work with pre_GA instead of Master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -287,7 +287,7 @@ jobs:
                   else
                     export IFRA_ENV_TYPE=Content-Env # disable-secrets-detection
                 fi
-                python ./Tests/scripts/awsinstancetool/aws_instance_tool.py -envType $IFRA_ENV_TYPE -outfile ./env_results.json
+                python ./Tests/scripts/awsinstancetool/aws_instance_tool.py -envType "$IFRA_ENV_TYPE" -outfile ./env_results.json
 
               else
                 python ./Tests/scripts/awsinstancetool/aws_instance_tool.py -envType CustomBuild -outfile ./env_results.json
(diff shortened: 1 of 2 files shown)


Tests/instance_notifier.py
```diff
@@ -131,14 +131,11 @@ def slack_notifier(slack_token, secret_conf_path, server, user, password, build_
     if options.instance_tests:
         with open('./env_results.json', 'r') as json_file:
             env_results = json.load(json_file)
-            for env in env_results:
-                if env["Role"] == "Server Master":
-                    server = SERVER_URL.format(env["InstanceDNS"])
-                    break
+            server = SERVER_URL.format(env_results[0]["InstanceDNS"])
```
yaakovi: not sure if you will get a list in case there is only one instance

.circleci/config.yml
```diff
@@ -277,9 +277,12 @@ jobs:
             fi
             if ./Tests/scripts/is_ami.sh ;
               then
-                if [ -n "${NIGHTLY}" ] || [ -n "${INSTANCE_TESTS}" ] ;
+                if [ -n "${NIGHTLY}" ] ;
                   then
                     export IFRA_ENV_TYPE=Nightly # disable-secrets-detection
+                elif [ -n "${INSTANCE_TESTS}" ] ;
+                  then
+                    export IFRA_ENV_TYPE=Demisto PreGA # disable-secrets-detection
```
yaakovi: better safe than sorry
```suggestion
                    export IFRA_ENV_TYPE="Demisto PreGA" # disable-secrets-detection
```

Tests/instance_notifier.py
ShahafBenYakir: I think I will, but in any case, will run an instance build on this branch.

.circleci/config.yml
ShahafBenYakir: :yaakovi:

PR #6311, opened 2020-03-30
update domain indicator type id

barchen1:
## Status
- [x] Ready

## Description
Update the indicator type ID

## Minimum version of Demisto
- [x] 5.5.0


barchen1:
I also checked it on self service and the layout is updated
FYI @reutshal

ronykoz:
@barchen1 check the cherwell failure in the build please

PR #6320, opened 2020-03-30
Change prevname of incidentfield-namefield.json

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22945


Itay4:
@anara123 @ronykoz please keep me sane here and verify this makes sense and we're not missing anything here

Itay4:
@jochman please access the build demisto instance and make sure both `name` (the system one) and `namefield` exist as expected

anara123:
looks ok. make sure that the error not happens and as @Itay4 mentioned both fields exists

jochman:
before upgrading
image
after upgrading
image

Seems like we're breaking BC with this one. Looking forward for another solution.

jochman:
It looks like now it's fine.
image
The indicator is installed.
image
Nothing contains `VCStore` In the logs.

Itay4:
@jochman any reason to keep the `name` field?

jochman:
> @jochman any reason to keep the `name` field?
Somehow, someone is maybe using it.

jochman:
according to @sagylell all of those keys should not be in the file:
"vcShouldIgnore"
shouldPublish"
"shouldCommit"
"shouldPush"
"commitMessage"
"modified"
"prevName"
"prevDetails"
"prevKind"
"prevTypeId"
"prevType"

deleting all of those and updating.

jochman:
after removing those keys, the indicator is inside demisto
image

logs:
image
nothing found

sagylell:
Haven't tested manually myself, but it should be ok

PR #6327, opened 2020-03-30
add deprecated msg

ronykoz:
Fixes https://github.com/demisto/etc/issues/23223


Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -1926,21 +1927,43 @@ def all_alerts_to_incidents(alerts, latest_creation_time, existing_ids):
...
+    Args:
+        aware_timestamp(date): The alert creation time after parse to aware timestamp
+
+    Returns:(date). Naive timestamp for alert creation time
+    """
+    iso_aware = aware_timestamp.isoformat()
+    # Deal with timestamp like: 2020-03-26T17:24:58.441093
+    if '.' in iso_aware:
+        iso_aware = iso_aware.split('.')[0]
+    # Deal with timestamp like: 2020-03-14T22:11:20+0000
```
ronykoz, line 1961: what happens if the + will be with 0300

barchen1: isoformat is changing the timestamp format, so for example if we have 2020-03-14T22:11:20+0300 in the aware_timestamp, after using isoformat it will change the +0300 to +0000

PR #6334, opened 2020-03-31
Cortex xdr bug fix

barchen1:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23083

## Description
Add more conditions to check the isolate and unisolate commands and return errors if needed.
for example:
Screen Shot 2020-03-31 at 17 25 51

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


Packs/CortexXDR/Integrations/PaloAltoNetworks_XDR/PaloAltoNetworks_XDR.py
```diff
@@ -932,25 +932,34 @@ def isolate_endpoint_command(client, args):
...
             None,
             None
         )
-
+    if endpoint_status == 'DISCONNECTED':
+        return_error(
+            f'Error: Endpoint {endpoint_id} is disconnected and therefore can not be isolated.'
+        )
+    if is_isolated == 'AGENT_PENDING_ISOLATION_CANCELLATION':
+        return_error(
```
anara123: instead of `return_error` raise `ValueError`

anara123: there is a main try/except that will catch this exception and return_error.
The main reason why we prefer not to use return_error everywhere in the code because it has side effects - it uses sys.exit()

barchen1: done

anara123: I still dont see that code has changed and that you `raise ValueError`

barchen1: im working on the unittest, cause its different with raise ValueError

ronykoz:
@barchen1 you have some guardrails/scan failures, please have a look

kirbles19:
@barchen1 Done.

ronykoz:
Force merging due to Server master thing

PR #6336, opened 2020-03-31
added 'builtin' to the '__name__' checks on startup

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready


Packs/GetLicenseID/Scripts/GetLicenseID/GetLicenseID.py
```diff
@@ -12,5 +12,5 @@ def get_license_id():
     return human_readable, {'Demisto License': license_id}
 
 
-if __name__ == '__main__':
+if __name__ in ('__main__', 'builtins'):
```
Itay4:
```suggestion
if __name__ in ('__main__', '__builtin__', 'builtins'):
```

hod-alpert: But this uses a docker with python 3 and `__builtin__` is from python 2

Itay4: user can override the docker image

PR #6352, opened 2020-04-01
Export indicators - on-demand error messages and csv to text functionality

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23287

## Description
Added the following:
1. Added a message in case no results return - and it is not an error.
2. In case of On-Demand mode - a message telling the user to run !eis-update for the first time.
3. Tooltip for On-Demand param to tell the user to run the !eis-update command for the first time.
4. Added functionality to output csv and xsoar-csv as text by:
a. Integration configuration checkbox.
b. csv_text argument in !eis-update command.
c. tx URL parameter

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -770,6 +770,13 @@ def route_list_values() -> Response:
             request_args=request_args
         )
 
+        if demisto.getIntegrationContext() == {} and params.get('on_demand'):
```
teizenman, line 773:
```suggestion
        if demisto.getIntegrationContext() and params.get('on_demand'):
```
would do the same isnt it?

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -770,6 +770,13 @@ def route_list_values() -> Response:
...
+            values = 'You are running in On-Demand mode - please run !eis-update command to initialize the ' \
+                     'export process'
+
+        elif values == '':
```
teizenman, line 777: same here

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -879,7 +886,12 @@ def update_outbound_command(args, params):
             '"Update exported IOCs On Demand" is off. If you want to update manually please toggle it on.')
     limit = try_parse_integer(args.get('list_size', params.get('list_size')), CTX_LIMIT_ERR_MSG)
     print_indicators = args.get('print_indicators')
+
     query = args.get('query')
+    # in case no query is entered take the query in the integration params
+    if len(query) == 0:
```
teizenman, line 892: same here

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -893,8 +905,13 @@ def update_outbound_command(args, params):
                                     category_default, category_attribute, collapse_ips)
 
     indicators = refresh_outbound_context(request_args)
-    hr = tableToMarkdown('List was updated successfully with the following values', indicators,
-                         ['Indicators']) if print_indicators == 'true' else 'List was updated successfully'
+    if len(indicators) > 0:
```
teizenman, line 908: It seems from `refresh_outbound_context` that the returned dict will never be of length 0, are you sure thats the correct way to do that?

roysagi:
> In general, I dont know the integration that good, but I think that if you have the ability to run the missing command for the user without ever breaking anything, I think that might be better than an error message.

Theoretically I agree but when I discussed this with @ronykoz he pointed there might be a user that wants to control even the first fetch since this is on demand so we came up with this choice.

Also could not find a way to run the first fetch myself - tried doing it with the test-module but I think integration context is not saved at that point so it did not work...

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
roysagi: this is used if you are not in the first situation - "on demand mode and empty integration context" and you still got no results - that means the integration returned no results.

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
roysagi: in case no indicators are fetched (lets say for example that the query did not return any results or the offset given is larger that the indicator list length) - then refresh_outbound_context would return an empty string.

Here I separate the two possibilities - I'll make the "not" change here as well

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
roysagi: done

teizenman:
Looks good to me

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -893,8 +905,13 @@ def update_outbound_command(args, params):
...
     indicators = refresh_outbound_context(request_args)
-    hr = tableToMarkdown('List was updated successfully with the following values', indicators,
-                         ['Indicators']) if print_indicators == 'true' else 'List was updated successfully'
+    if not indicators:
+        hr = tableToMarkdown('List was updated successfully with the following values', indicators,
+                             ['Indicators']) if print_indicators == 'true' else 'List was updated successfully'
+
+    else:
+        hr = "No Results Found For the Query"
+
```
ronykoz, line 914: Shouldn't this be the other way around?

roysagi: right my bad

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py
```diff
@@ -685,6 +694,7 @@ def get_request_args(params):
     category_default = request.args.get('cd', params.get('category_default', 'bc_category'))
     category_attribute = request.args.get('ca', params.get('category_attribute', ''))
     collapse_ips = request.args.get('tr', params.get('collapse_ips', DONT_COLLAPSE))
+    csv_text = request.args.get('tx', params.get('csv_text', False))
```
ronykoz, line 697: let's change it to
```suggestion
    csv_text = request.args.get('text', params.get('csv_text', False))
```

Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators_description.md
```diff
@@ -41,3 +42,8 @@ Use the following arguments in the URL to change the request:
     * 1 - Collapse to ranges.
     * 2 - Collapse to CIDRs.
  * Example: https://{demisto_instance}/instance/execute/{ExportIndicators_instance_name}?q="type:ip and sourceBrand:my_source"&tr=1
+11. **tx** - Whether to output `csv` or `xsoar-csv` formats as textual web pages.
```
ronykoz, line 45: let's update the name here too

roysagi: several places - on it...

kirbles19:
@roysagi Done.

ronykoz:
Merging due to master problems

PR #6355, opened 2020-04-01
Shift management

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23254

## Description
 - Added `onCall` argument to `AssignAnalystToIncident` script and to all playbooks that use it
 - Added `On-Call Users` widget
 - Added `Roles Per Shift` widget

## Screenshots
image

image

image

## Minimum version of Demisto
- [x] 5.5.0

## Must have
- [x] Tests
   - Added UT for the widget scripts
   - Adding test playbook for `AssignAnalystToIncident` will require adding roles and users, can be done either in the AMI or with Demisto REST API in the playbook - still need to be added
- [ ] Documentation 

## Additional changes
Modified `Pipfile.lock` to include `3.8` for `importlib-metadata` and `cpython`


Packs/ShiftManagement/Scripts/GetRolesPerShift/GetRolesPerShift.py
```diff
@@ -0,0 +1,73 @@
+import demistomock as demisto
+from CommonServerPython import *
+from CommonServerUserPython import *
+
+from typing import List
+
+HOURS_DAYS_HEADER = 'Hours / Days'
+SUNDAY_HEADER = 'Sunday'
```
slavikm, line 8: Wonder if there is a way to get them from i18n - probably not...

Playbooks/playbook-Access_Investigation_-_Generic.yml
```diff
@@ -918,6 +921,11 @@ inputs:
     simple: Administrator
   required: true
   description: The default role to assign the incident to.
+- key: OnCall
+  value:
+    simple: false
+  required: false
+  description: Set to true to assign only the users that are currently on shift.
```
slavikm, line 928: Should we specify that this is only 5.5 and above?

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -27,6 +26,7 @@ script: |
   }
 
   assignBy = args.assignBy || 'random';
+  var onCallArg = true ? args.onCall === 'true' : false;
```
slavikm, line 29: Probably should ignore case, no?

Scripts/script-AssignAnalystToIncident.yml
```diff
@@ -36,28 +36,29 @@ script: |
...
-              var usersRes = executeCommand('getUsers', { roles: args.roles, online: true });
+              var usersRes = executeCommand('getUsers', { roles: args.roles, online: true, onCall: onCallArg });
               if (isError(usersRes[0])) {
                   return usersRes[0];
               }
               pickRandomUser(usersRes);
               break;
           case 'current':
-              var usersRes = executeCommand('getUsers', { current: true });
+              var usersRes = executeCommand('getUsers', { current: true, onCall: onCallArg });
```
slavikm, line 46: It's not relevant for current but it's also not a problem to pass the flag

slavikm:
@Itay4 can we make each role in the roles widget a different color?

slavikm:
@Itay4 actually, the script `GetRolesPerShift` is wrong - DO NOT MERGE. Fixing.

Packs/ShiftManagement/Scripts/GetRolesPerShift/GetRolesPerShift.py
Itay4: nope

Scripts/script-AssignAnalystToIncident.yml
Itay4: there are predefined values, so i think its ok

Itay4:
> @Itay4 can we make each role in the roles widget a different color?

@slavikm seems like we dont support html in widgets

Itay4:
@glicht added two more widgets, please have a look
also, please review the change done in the default `Pipfile.lock` to work with python 3.8

Itay4:
@kirbles19 please note that the added argument `onCall` to the script, and added inputs `OnCall` to the playbooks, is available from version 5.5
how would you like to mention that in the argument/inputs description ?

Packs/ShiftManagement/Scripts/GetOnCallHoursPerUser/GetOnCallHoursPerUser.py
```diff
@@ -0,0 +1,60 @@
...
+
+def main():
+    get_roles_response: List = demisto.executeCommand('getRoles', {})
+    if is_error(get_roles_response):
+        demisto.error(f'Failed to get roles: {str(get_error(get_roles_response))}')
+    else:
+        hours_per_user: Dict[str, int] = {}
+        roles = get_roles_response[0]['Contents']
+        for role in roles:
+            get_users_response: List = demisto.executeCommand('getUsers', {'roles': role.get('name')})
```
slavikm, line 35: Should check if there are shifts for this role and only then retrieve uses. We might have 1000s of roles from AD, etc. and this can be extremely slow.

Itay4: But then we might miss users, and we want to display all users, including ones with no on-call hours
We can change the script to first get all users, initialize them all with 0 on-call hours and then query for users who are assigned to roles with shifts as you suggested
You think that will be better?

slavikm: Yes, sounds like a good suggestion. In that case, you might not even need to call `getUsers` for a role because you already have the data in your hands from the first `getUsers` call...

Itay4: done

PR #6361, opened 2020-04-02
Delete Windows-ATP entries from conf.json and add the test playbook t…

barchen1:
…o non-circle-tests

## Status
- [x] Ready

## Description
Delete Windows-ATP from conf.json and add Windows-ATP-Test_playbook to non circle testplaybook - Deprecated integration


Tests/conf.json
```diff
@@ -1335,17 +1335,6 @@
         {
             "integrations": "Cisco Meraki",
             "playbookID": "Cisco-Meraki-Test"
-        },
```
ShahafBenYakir, line 1338: This line should not be deleted

barchen1: done

PR #6362, opened 2020-04-02
Revert "Revert "Delete Old TLP Indicator Field""

avidan-H:
Reverts demisto/content#6270


avidan-H:
Adding the `docs-approved` label since there are no documents to review.

PR #6369, opened 2020-04-02
intsights-get-alert-activities - None err when missing some alert data

bakatzir:
## Status
- [x] Ready

reference: https://github.com/demisto/etc/issues/22185


Integrations/IntSight/IntSight.py
```diff
@@ -328,45 +330,48 @@ def get_alert_activity():
...
-            'Initiator': demisto.get(activity, 'Initiator'),
-            'CreatedDate': demisto.get(activity, 'CreatedDate'),
-            'UpdateDate': demisto.get(activity, 'UpdateDate'),
-            'RemediationBlocklistUpdate': extract_remediation(
-                demisto.get(activity, 'AdditionalInformation.RemediationBlocklistUpdate')),
-            'AskTheAnalyst': {'Replies': demisto.get(activity, 'AdditionalInformation.AskTheAnalyst.Replies')},
-            'Mail': extract_mail(demisto.get(activity, 'AdditionalInformation.Mail.Replies')),
-            'ReadBy': demisto.get(activity, 'ReadBy')
+    if not response:
+        demisto.results('Alert activities were not found.')
```
ShahafBenYakir, line 334: I am afraid this could break BC. What do you think?

ShahafBenYakir, line 334: I am afraid this could break BC. What do you think?

bakatzir: if there is no response, there is nothing to iterate on in response, which can reproduce an error as well.

bakatzir:
@ShahafBenYakir We get 401 now :/

PR #6370, opened 2020-04-02
demisto/etc#23148: Wrapped vulnDB integration in base client 

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#23148

## Description
Wrapped vulnDB integration in base client

### DIFF
--- a/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
+++ b/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased]
--
+Improve exception parsing when api quota has exceeded to be more informative 
 
 ## [20.2.0] - 2020-02-04
 Added the ***cve*** command, which returns CVE information.
--- a/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
+++ b/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## [Unreleased]
 -
 
-
 ## [20.2.0] - 2020-02-04
 Added the ***cve*** command, which returns CVE information.
--- a/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
+++ b/Packs/Legacy/Integrations/VulnDB/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased]
-Improve exception parsing when api quota has exceeded to be more informative 
+Improved exception parsing when the API quota is exceeded.
 
 ## [20.2.0] - 2020-02-04
 Added the ***cve*** command, which returns CVE information.
--- a/Packs/Legacy/Integrations/VulnDB/VulnDB.py
+++ b/Packs/Legacy/Integrations/VulnDB/VulnDB.py
@@ -10,57 +10,42 @@ import urllib.parse
 # Disable insecure warnings
 requests.packages.urllib3.disable_warnings()
 
-''' GLOBALS/PARAMS '''
-
-# Remove trailing slash to prevent wrong URL path to service
-API_URL = demisto.params()['api_url'].rstrip('/')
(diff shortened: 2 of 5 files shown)


Itay4:
did we verify test playbook passes?

Integrations/VulnDB/VulnDB.py
```diff
@@ -421,31 +379,51 @@ def vulndb_get_cve_command():
...
+    # Remove trailing slash to prevent wrong URL path to service
+    api_url = params['api_url'].rstrip('/')
+    client_id = params['client_id']
+    client_secret = params['client_secret']
+    use_ssl = not params.get('insecure', False)
+    proxy = params.get('proxy', False)
+    client = Client(proxy, use_ssl, api_url, client_id, client_secret)
+    args = demisto.args()
+    command = demisto.command()
+    LOG('Command being called is %s' % (demisto.command()))
```
Itay4, line 394: lets use f-strings here

Integrations/VulnDB/VulnDB.py
```diff
@@ -421,31 +379,51 @@ def vulndb_get_cve_command():
...
-    vulndb_get_version_command()
-elif demisto.command() == 'vulndb-get-updates-by-dates-or-hours':
-    vulndb_get_updates_by_dates_or_hours_command()
-elif demisto.command() == 'cve':
-    vulndb_get_cve_command()
+
+def main():
+    params = demisto.params()
+    # Remove trailing slash to prevent wrong URL path to service
+    api_url = params['api_url'].rstrip('/')
```
Itay4, line 386: i think `BaseClient` already handles that

Integrations/VulnDB/VulnDB.py
```diff
@@ -10,57 +10,42 @@
...
+        return res.get('access_token')
+
+    def http_request(self, url_suffix, size=None):
+        params = {'size': size} if size else None
+        res = self._http_request("GET",
+                                 url_suffix=url_suffix,
+                                 params=params)
+        for error_attribute in ['error', 'details']:
+            if error_attribute in res:
+                return_error(res[error_attribute])
```
Itay4, line 47: raise exception instead of `return_error`

Integrations/VulnDB/VulnDB.py
```diff
@@ -421,31 +379,51 @@ def vulndb_get_cve_command():
...
+        elif command == 'vulndb-get-product':
+            vulndb_get_product_command(args, client)
+        elif command == 'vulndb-get-version':
+            vulndb_get_version_command(args, client)
+        elif command == 'vulndb-get-updates-by-dates-or-hours':
+            vulndb_get_updates_by_dates_or_hours_command(args, client)
+        elif command == 'cve':
+            vulndb_get_cve_command(args, client)
+    except Exception as e:
+        error_message = str(e)
```
Itay4, line 424:
```suggestion
        error_message = f'Failed to execute {command} command. Error: {str(e)}')
```

Integrations/VulnDB/VulnDB.py
```diff
@@ -10,57 +10,42 @@
...
+                                 },
+                                 full_url=oath_url)
+        return res.get('access_token')
+
+    def http_request(self, url_suffix, size=None):
+        params = {'size': size} if size else None
+        res = self._http_request("GET",
+                                 url_suffix=url_suffix,
+                                 params=params)
+        for error_attribute in ['error', 'details']:
```
Itay4, line 45: add comment here that its either one of the two

hod-alpert:
@Itay4 Added unit tests and the playbook passes when there is quota, when there isn't it will display the correct error.

In addition- @ShahafBenYakir changed its configurations so that it will only run at nightly to avoid exceeding the limit

kirbles19:
@hod-alpert Done.

PR #6375, opened 2020-04-03
Added the argument *ignore_auto_extract* to the ***return_outputs*** …

bakatzir:
…command

## Status
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/15009

## Description
in order to complete the referenced issue, need to add this func to `return_output` command.

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1723,6 +1726,7 @@ def return_outputs(readable_output, outputs=None, raw_response=None, timeline=No
         "ContentsFormat": formats["json"],
         "Contents": raw_response,
         "EntryContext": outputs,
+        'IgnoreAutoExtract': ignore_auto_extract,
```
yaakovi, line 1729: add this key only when `ignore_auto_extract` is True.

yaakovi: or fix the failing tests

PR #6384, opened 2020-04-05
Deprecate postgreSQL - using GenericSQL instead

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

Deprecating PostgreSQL - using GenericSQL instead.

reference: https://github.com/demisto/etc/issues/23355


Integrations/integration-PostgreSQL.yml
```diff
@@ -4,8 +4,9 @@ commonfields:
 name: PostgreSQL
 display: PostgreSQL
```
bakatzir, line 5:
```suggestion
display: PostgreSQL (Deprecated)
```

Integrations/integration-PostgreSQL.yml
```diff
@@ -4,8 +4,9 @@ commonfields:
...
-description: Query PostgreSQL Database
+description: Deprecated. Use Generic SQL
```
glicht, line 9: Small syntax update suggestion:
```suggestion
description: Deprecated. Use The Generic SQL integration instead.
```

PR #6416, opened 2020-04-07
Bug fix find similiar incidents

barchen1:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22648

## Description
Added a fix for list values in context keys.

## Screenshots
For example the contexts for some incidents here are
 {out: {test: [1, 2]}} and others are {out: {test: 2, 1]}} and it shows that they are duplicate (didn't happen until this fix)

Screen Shot 2020-04-07 at 13 16 01

## Minimum version of Demisto
- [x] 4.5.0

   - [x] Yes

## Must have
- [x] Tests
- [ ] Documentation


Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.py
```diff
@@ -283,22 +290,29 @@ def main():
...
-                                                        raise_error=RAISE_ERROR_MISSING_VALUES)
+    incident_similar_context = demisto.context()
+    original_context_map = {}
+
+    if incident_similar_context:
+        for key in SIMILAR_CONTEXT_MAP.keys():
+            response = demisto.dt(incident_similar_context, key)
+            original_context_map[key] = response
+            if not response and RAISE_ERROR_MISSING_VALUES:
+                return_error("Missing context field for incident: %s" % key)
```
erezh31, line 301: rename, "context key" -> "context field"

Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.yml
```diff
@@ -146,7 +146,7 @@ tags:
 - incidents
 timeout: 300ns
 type: python
-dockerimage: demisto/dateutil
+dockerimage: demisto/python:2.7.17.6981
```
erezh31, line 149: Sure we want to do that? I know that usually we don't change the docker image because it can lead to backward compatibility issues, @ronykoz ?

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
```diff
@@ -170,6 +173,44 @@ def test_similar_context(mocker):
...
     assert result['EntryContext']['similarIncidentList'][1]['rawId'] == 2
+
+    args.update({'similarIncidentFields': 'name', 'similarContextKeys': 'dictListValue'})
+    result = main()
+    assert len(result['EntryContext']['similarIncidentList']) == 1
+    assert result['EntryContext']['similarIncidentList'][0]['rawId'] == 2
+
+
+def dt_res(context, key):
+    if key == 'simpleValue':
```
erezh31, line 192: I would prefer to implement a simple dt logic, so if someone will change the values of context, it will still work.
A simple logic can be:
split by "." and return the nested values for that. You can use on of the methods here:
https://stackoverflow.com/questions/25833613/python-safe-method-to-get-value-of-nested-dictionary
I think `safeget` or `deep_get` both will be good for you

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
```diff
@@ -170,6 +173,44 @@ def test_similar_context(mocker):
...
+    result = main()
     assert len(result['EntryContext']['similarIncidentList']) == 2
     assert result['EntryContext']['similarIncidentList'][0]['rawId'] == 3
     assert result['EntryContext']['similarIncidentList'][1]['rawId'] == 2
+
+    args.update({'similarIncidentFields': 'name', 'similarContextKeys': 'dictListValue'})
+    result = main()
+    assert len(result['EntryContext']['similarIncidentList']) == 1
+    assert result['EntryContext']['similarIncidentList'][0]['rawId'] == 2
+
```
erezh31, line 189: Please add one more test for empty condition to context - means that the incident does not have the context key that specified in the SimilarContextKeys. This test should return error

Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.py
```diff
@@ -326,27 +342,34 @@ def main():
...
+            if other_incident_context:
+                other_incident_context_map = {}
+
+                for key in SIMILAR_CONTEXT_MAP.keys():
+                    response = demisto.dt(other_incident_context, key)
+                    if response:
+                        other_incident_context_map[key] = response
+                        if verify_map_equals(original_context_map, other_incident_context_map, SIMILAR_CONTEXT_MAP):
+                            filter_by_context.append(c)
         duplicate_incidents = filter_by_context
```
ronykoz, line 359: Why do you need both `filter_by_context` and `duplicate_incidents`

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
ronykoz: Yes i agree

Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.yml
ronykoz: From 5.0 we are ok with changing that, but it should be:
```
dockerimage45: demisto/dateutil
dockerimage: demisto/python:2.7.17.6981
```
But why is it python2?
@erezh31 from which version do we have the demisto.dt in the server?

Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.py
ronykoz, line 359: why do you need them both

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
ronykoz: yes i agree

barchen1: done

Packs/Legacy/Scripts/FindSimilarIncidentsV2/FindSimilarIncidentsV2.py
barchen1: the for loop is on duplicate_incidents, so we cant append to it and that is why we need a different list

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
```diff
@@ -182,35 +209,66 @@ def test_similar_context(mocker):
...
+        if context_key_value:
+            if isinstance(context_key_value, list):
+                list_value = []
+                for value in context_key_value:
+                    if isinstance(context_key_value, dict):
+                        if value.get(key):
+                            list_value.append(value[key])
+                    else:
+                        list_value.append(value)
+                context = context_key_value = list_value
```
erezh31, line 257: why this?
why not just `context_key_value = list_value`

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
```diff
@@ -182,35 +209,66 @@ def test_similar_context(mocker):
...
+            elif isinstance(context_key_value, dict):
+                dict_list_value = []
+                for dict_key in context_key_value.keys():
+                    if dict_key == key:
+                        if isinstance(context_key_value[dict_key], list):
+                            dict_list_value.append(sorted(context_key_value[dict_key]))
+                        else:
+                            dict_list_value.append(context_key_value[dict_key])
+
+                context = context_key_value = dict_list_value
```
erezh31, line 268: why not just context_key_value = list_value

barchen1: done

Packs/Legacy/Scripts/FindSimilarIncidentsV2/find_similar_incidents_test.py
barchen1: you are right
I will change it

PR #6424, opened 2020-04-07
added md files check

orhovy:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23339

## Description
Added a check if file has been loaded.

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


release_notes.py
```diff
@@ -503,11 +503,14 @@ def handle_deleted_file(full_file_name, git_sha1):
...
     data = get_remote_file(full_file_name, git_sha1)
-    name = data.get('name') or data.get('name') or full_file_name
-    file_type = full_file_name.split("/")[0]
-    file_type_mapping = RELEASE_NOTE_GENERATOR.get(file_type)
-    if file_type_mapping is not None:
-        file_type_mapping.add('D', name)
+    # If the data that returns is {} than the file is a md file,
+    # for such files we will not have a release note generator and we will skip
+    if data:
+        name = data.get('name') or data.get('name') or full_file_name
```
jochman, line 509:
```suggestion
        name = data.get('name') or full_file_name
```

orhovy:
@jochman If no data returned from get_remote_file func then there are 2 options:
First , the deleted file had no actual data in it, empty yml or empty json, in this case we dont want to get its name as if there was not actual data deleted, only empty file.
Second, the deleted file was a changelog or readme file, in this case , there is no reason to add it to release notes as if there is not actual functionally that has been deleted.

PR #6426, opened 2020-04-07
Packs uploader - Update index with private packs

liorblob:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
https://github.com/demisto/etc/issues/23240

## Description
This updates the index with the packs from the private bucket, as well as adding them to the `index.json` with the price. 
Tested locally and it works.
@IkaDemisto @yuvalbenshalom Please have a look before I start with unit tests.


Tests/Marketplace/upload_packs.py
```diff
@@ -15,7 +16,7 @@
 from demisto_sdk.commands.common.tools import run_command, print_error, print_warning, print_color, LOG_COLORS
 
 
-def get_modified_packs(specific_packs=""):
+def get_modified_packs(specific_packs="", private_packs_path=""):
```
IkaDemisto, line 19: why do you need this changes?

Tests/Marketplace/upload_packs.py
```diff
@@ -203,6 +212,24 @@ def upload_index_to_storage(index_folder_path, extract_destination_path, index_b
     print_color(f"Finished uploading {GCPConfig.INDEX_NAME}.zip to storage.", LOG_COLORS.GREEN)
 
 
+def get_private_packs(private_index_path):
+    metadata_files = glob.glob(f"{private_index_path}/**/metadata.json")
+    private_packs = []
+
+    for metadata_file_path in metadata_files:
+        with open(metadata_file_path, "r") as metadata_file:
+            file_data = metadata_file.read()
```
IkaDemisto, line 221: json.load() instead?

IkaDemisto:
You need to copy the folders inside the private index folder to public (it includes pack README, Changelog.json and metadata.json).

Tests/Marketplace/upload_packs.py
```diff
@@ -383,15 +415,16 @@ def main():
             pack.cleanup()
             continue
 
-        # detected index update
-        index_was_updated = True
         pack.status = PackStatus.SUCCESS.name
 
+    # Add private packs to index
+    for d in os.scandir(private_index_path):
```
IkaDemisto, line 421: I think better to copy it before the loop

Tests/Marketplace/upload_packs.py
```diff
@@ -203,6 +211,44 @@ def upload_index_to_storage(index_folder_path, extract_destination_path, index_b
...
+    Add the private packs to the index folder.
+    Args:
+        index_folder_path: The index folder path.
+        private_index_path: The path for the index of the private packs.
+
+    """
+    for d in os.scandir(private_index_path):
+        if os.path.isdir(d.path):
+            update_index_folder(index_folder_path, d.name, d.path)
+    shutil.rmtree(private_index_path)
```
yuvalbenshalom, line 249: I think it is better to delete outside of this function, because this is not what this function states it dose.
if you move it, consider wrapping the commands after the creation of `private_index_path` with `try`, and call the deletion with `finally`

Tests/Marketplace/upload_packs.py
```diff
@@ -293,23 +349,35 @@ def main():
...
+    storage_client = init_storage_client(service_account)
+    storage_bucket = storage_client.bucket(storage_bucket_name)
+    private_storage_bucket = storage_client.bucket(private_bucket_name)
+    index_folder_path, index_blob = download_and_extract_index(storage_bucket, extract_destination_path)
+    private_index_path, _ = download_and_extract_index(private_storage_bucket,
+                                                       os.path.join(extract_destination_path, 'private'))
+
     # detect new or modified packs
     modified_packs = get_modified_packs(specific_packs)
+    private_packs = get_private_packs(private_index_path)
```
yuvalbenshalom, line 369: since there is no use for `private_packs` until line 374,
please move this line down, just before the `try` so no exceptions can happen in between and live the directory un cleaned

Tests/Marketplace/upload_packs.py
```diff
@@ -284,6 +339,7 @@ def option_handler():
                         action='store_true', required=False)
     parser.add_argument('-k', '--key_string', help="Base64 encoded signature key used for signing packs.",
                         required=False)
+    parser.add_argument('-P', '--private_bucket_name', help="Private storage bucket name", required=False)
```
IkaDemisto, line 342: Bad name `-P`. Additionally, where in ci yml you pass `--private_bucket_name` to script?

liorblob: Couldn't think of a better letter for that... And I haven't added that to the config because the server doesn't support encrypted packs yet

IkaDemisto: It is failing without it with following error: `ValueError: Cannot determine path without bucket name`. `private_storage_bucket` will be None and `download_and_extract_index` will raise exception. You need to skip this task in case no private bucket name was supported. `-p` for public and `-P` for private is confusing, better to remove `-P` shortcut at all.

IkaDemisto: Ignore my comment regarding letter...

liorblob: Let's decide what to do with the config in the daily/afterwards

PR #6372, opened 2020-04-03
[sixgill-195] sixgill darkfeed integration

BenSterenson:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: link to the issue

## Description
New integration of sixgills darkfeed
deprecating old sixgill-get-indicators command

## Screenshots
Paste here any images that will help the reviewer

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

## Does it break backward compatibility?
   - [x ] Yes
       - Further details:
removed sixgill-get-indicators and created a new integration of darkfeed
   - [ ] No

## Must have
- [x] Tests
- [x] Documentation 

## Demisto Partner?
- [x] The title must be in the following format: **[YOUR_PARTNER_ID] short description**

PR #6439, opened 2020-04-09
dont fail on skipped integrations

anara123:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23258

## Description
This PR contains two fixes for two different scenarios:
1. I have a test in skip integrations, I don't intend to add test playbook to it (partner). The validation (configure_tests) was failing on enforcing the user to configure some test playbook. 
Suggested fix in the PR is - if the integration is in skipped_integrations then don't require a test playbook

2. I have integration that have only fetch-incidents. I configured a test playbook and the test playbook uses FetchFromInstance script only for testing. Currently the validation (configure_tests) fails, because according to id_set no test playbook uses that integration commands.
Suggested fix in the PR is - don't fail on it and run the correct test playbook. Check if the integration  has `isfetch: true` and if there is test playbook that configured together with this integration and uses FetchFromInstance script then don't fail.

PR for examples that this happens
https://github.com/demisto/content/pull/6372

Build
https://circleci.com/gh/demisto/content/47041?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


Tests/scripts/configure_tests.py
```diff
@@ -266,7 +275,60 @@ def update_missing_sets(catched_intergrations, catched_playbooks, catched_script
...
+
+    def get_tests(self):
+        return self._conf.get('tests', {})
+
+    def get_test_playbook_ids(self, check_nightly_status=False):
+        conf_tests = self._conf['tests']
+        test_ids = []
+
+        for t in conf_tests:
+            if not check_nightly_status or not t.get('nightly', False):
```
ronykoz, line 300: we don't really care if it's nightly or not anymore, we can remove this

PR #6417, opened 2020-04-07
AF v2 - sample-search artifact argument addition

orhovy:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/23139

## Description
added a new argument to search-samples command.
changed context and human readable of samples-result command. now every entry in table will be set in a new entry in war room , and for each entry its artifacts will be added.

## Minimum version of Demisto
- [x] 4.5.0

## Does it break backward compatibility?
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

PR #6447, opened 2020-04-12
AutoFocus v2 - added support for upper case hash

roysagi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23356

## Description
added support for upper case hash

Also added a test for this in the playbook
image


Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -838,11 +838,11 @@ def search_indicator(indicator_type, indicator_value):
...
-            return_error(f'Request Failed with status: {result.status_code}.\n'
-                         f'Reason is: {str(error_message)}.')
+            raise ValueError(f'Request Failed with status: {result.status_code}.\n'
+                             f'Reason is: {str(error_message)}.')
         elif str(result.status_code) in ERROR_DICT:
-            return_error(f'Request Failed with status: {result.status_code}.\n'
-                         f'Reason is: {ERROR_DICT[str(result.status_code)]}.')
+            raise ValueError(f'Request Failed with status: {result.status_code}.\n'
+                             f'Reason is: {ERROR_DICT[str(result.status_code)]}.')
         else:
```
ronykoz, line 846: why did you change it?

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -1320,7 +1320,14 @@ def search_file_command(file):
     file_list = argToList(file)
     indicator_details = []
     for _file in file_list:
-        raw_res = search_indicator('sha256', _file)
+        try:
```
ronykoz, line 1323: why not to do lower to begin with?
And lets add a UT for this

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
roysagi: when it was return_error it skipped the 'except' I entered in the search_file_command function bellow.
This is a way to use it and not change the integration behavior.

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
roysagi: I looked it up and there are cases of "case sensitive" hash - https://stackoverflow.com/questions/23786939/what-kind-of-hash-returns-lower-and-upper-case-characters

This way they will also be supported as long as AutoFocus supports them.

Also I didn't find a way to test my change through a UT without changing more parts of the code - hence the addition of a command to the playbook as a replacement.

roysagi: seems this command can only get SHA256 which is not case sensitive - https://stackoverflow.com/questions/11892053/is-sha-256-case-insensitive

Packs/AutoFocus/Integrations/AutofocusV2/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
--
+Fixed an issue where ***!file*** only accepted lower case hash.
```
amshamah419, line 2:
```suggestion
Fixed an issue where ***!file*** only accepted a lower case hash.
```

roysagi:
Fails for the same reason as this PR: https://github.com/demisto/content/pull/6417

ronykoz:
Merging due to Master

PR #6449, opened 2020-04-12
ExportIndicators - fix outbound mimetype bug

Itay4:
## Status
- [x] Ready

## Description
Fixed bug in which `get_outbound_mimetype` raised `NoneType has no get attribute`

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests


roysagi:
should be noted that this issue is covered here: https://github.com/demisto/content/pull/6352 - with additional explanations about the on-demand mode and readme.

PR #6459, opened 2020-04-12
Removing https or http before URL or IP - Pan-os management

GalRabin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23463

## Description
Palo Alto Networks PAN-OS EDL Management Doesn't omit HTTP/s from URL entries which causing failure in loading entries to the EDL object.

## Screenshots
The first URL before my change.
The second-third URL after my change.
image

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Legacy/Integrations/PaloAltoNetworks_PAN_OS_EDL_Management/PaloAltoNetworks_PAN_OS_EDL_Management.py
```diff
@@ -124,6 +126,45 @@ def scp_execute(file_name: str, file_path: str):
...
+
+    Examples:
+        >>> parse_url('http://google.com')
+        'google.com'
+        >>> parse_url('https://google.com')
+        'google.com'
+        >>> parse_url('https://google.com/hello_world')
+        'google.com/hello_world'
+        >>> parse_url('not url')
+        'google.com/hello_world'
```
yaakovi, line 146: fix the example (it will return `not url`)

yaakovi:
do we need to support the params as well?
`https://google.com/search?q=hello`

GalRabin:
> do we need to support the params as well?
> `https://google.com/search?q=hello`

I will test it

ronykoz:
@GalRabin where are we on this?

GalRabin:
https://github.com/demisto/content/pull/6459#issuecomment-617598294 ,
There is bug with the secret command, can't move forward due to build failure

PR #6460, opened 2020-04-12
postgre - name fix

bakatzir:
## Status
- [x] Ready
@ohaim1008 we should cahnge only the display, not the name, as it might break BC.


bakatzir:
force-merging as master is currently broken, and this fix is necessary to not break BC.

bakatzir:
fixes: https://github.com/demisto/etc/issues/23355

PR #6468, opened 2020-04-12
added traceback for all errors

ronykoz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23530

Added traceback to all errors


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1755,6 +1756,7 @@ def return_error(message, error='', outputs=None):
         :return: Error entry object
         :rtype: ``dict``
     """
+    message = "{}\n\nTrace:{}".format(message, traceback.format_exc())
```
Itay4, line 1759: `format_exc()` already returns some thing like `Traceback (most recent call last):` no?
if so, lets remove the string `Trace:` from here

glicht:
@ronykoz We should add a function argument: `include_stacktrace=False`. If it is false the stacktrace shouldn't be added to the message. There are a lot of integrations/scripts already handling the stacktrace.

ronykoz:
@glicht i rather have the message twice in the logs, it will help us solve bugs a lot faster. Because going over all the integrations/scripts and adding this is not feasible.
So I will put a default value of True, and again it will only show in the logs

anara123:
@ronykoz how about added also validation if args contains `debug-mode`

ronykoz:
@roysagi can you go over the RN here?

PR #6492, opened 2020-04-13
abuse.ch SSL Blacklist Feed - switching creationdate to firstseenbyfeed

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23569

## Description
The mapping of creationdate field is wrong as it is only associated with domain.
This should be firstseenbyfeed as seen here: https://docs.google.com/spreadsheets/d/1F6KeP-z5yLH0bhEE7MN_xRoTZvsTe5epkExFkbr6Kpo/edit#gid=0 (line 40)

## Screenshots
Here is the result of the get-indicators command to test the output fields
image


roysagi:
@Nitzperetz - FYI

ronykoz:
@roysagi please check that the indicator is created correctly in the indicators page, the time format might cause issues

roysagi:
@ronykoz - indicators are created but yes I see an issue with the first seen time
image

But this is because this is the way info is returned from the feed. To change this, I'll need to dive into the CSVAPIModule.

ronykoz:
@roysagi lets do that please, let’s make sure the solution covers the last seen as well

roysagi:
image
done

roysagi:
continued here: https://github.com/demisto/content/pull/6535

PR #6509, opened 2020-04-16
Genericsql  bugs: mysql.user table is available and context fixed

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23354, https://github.com/demisto/etc/issues/23575

## Description
Fixing 5 bugs:
1. Context outputs not using skip and limit.
2. Tables_in_mysql (such as user) were not available due to unreadable values in the table.
3. In case there are connection parameters and we are using mssql db, the connection string should not contain "?" twice.
4. When query doesn't return results there we get the error "list index out of range". changing the error by including the stack trace and the query.
5. Changing docker image in order to support Oracle db.


ohaim1008:
Bug fixed with same docker image.

Packs/GenericSQL/Integrations/GenericSQL/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased]
-
+-
```
bakatzir, line 2: please add a changelog message

bakatzir:
@roysagi can you help us with the CHANGELOG?

bakatzir:
@roysagi Orel added a few more lines to the CHANGELOG. please have a 2nd look.

ohaim1008:
Build is failing due to docker image change which is needed in this PR.
Please consider force merge. @bakatzir

bakatzir:
@ohaim1008 Please file an issue for a fix in the demisto-sdk validation.

PR #6512, opened 2020-04-16
Fixed CheckEmailAuthenticity to run on Headers rather than HeadersMap

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19561

## Description
`CheckEmailAuthenticity` script was running on the wrong headers output (Email.HeadersMap) instead of the one it can parse (Email.Headers).

## Minimum version of Demisto
4.5.0

No

## Must have
- [x] Tests


kirbles19:
@idovandijk Done, perfect, no changes.

PR #6518, opened 2020-04-16
Fixed an issue in which the *job_id* argument in the ***panorama-down…

bakatzir:
…load-panos-status*** command was misspelled.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23649

## Description
not adding UT as this is a contrib and we do not upgrade panos in the lab.
opening a request for pan-os v2 to be created with the class to allow easier UT writing.
reference: https://github.com/demisto/etc/issues/17429

## Minimum version of Demisto
- [x] 3.5.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


bakatzir:
failed on an unrelated automation `GenerateInvestigationSummaryReport`. force merging.

PR #6530, opened 2020-04-19
Zipfile escape illegal chars in file name

teizenman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23089

## Description
Replaces all appearances of illegal characters with`-`.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests


teizenman:
Needs a force merge since Ive updated the `dockerimage` key

Packs/Legacy/Scripts/ZipFile/ZipFile.py
```diff
@@ -2,100 +2,117 @@
...
+    # copying the file to current location
+    shutil.copy(filePath, fileCurrentName)
+    # zipping the file
+    if password:
+        pyminizip.compress(fileCurrentName, zipName, password, 5)
+
+    else:
+        zf = zipfile.ZipFile(zipName, mode='w')
+        try:
+            zf.write(fileCurrentName, compress_type=compression)
```
roysagi, line 82: If you changed the fileCurrentName above in relation to how it is "written" in the system (`res[0]['Contents']['name']`) - how does the zip know to fetch the file here?

Packs/Legacy/Scripts/ZipFile/ZipFile.py
```diff
@@ -2,100 +2,117 @@
...
-    }]
-
-demisto.results(results)
+ESCAPE_CHARACTERS = r'[/\<>"|?*]'
+
+
+def escape_illegal_characters_in_file_name(file_name):
+    if file_name:
+        file_name = re.sub(ESCAPE_CHARACTERS, '-', file_name)
+        file_name = re.sub('-+', '-', file_name)
```
roysagi, line 17:
```suggestion
        file_name = re.sub(r'-+', '-', file_name)
```
shouldn't it be like so? so it will be perceived as a pattern.
Also a note explaining this is to prevent subsequent dashes could be useful.

Packs/Legacy/Scripts/ZipFile/ZipFile.py
teizenman: `shutil.copy(filePath, fileCurrentName)` copies the file from the Demisto entry id to the `fileCurrentName` which was altered prior to it.
So, in the `zf.write` the file name is the right now.

Packs/Legacy/Scripts/ZipFile/ZipFile.yml
```diff
@@ -36,5 +36,5 @@ tags:
 timeout: '0'
 type: python
 subtype: python3
-dockerimage: demisto/python_zipfile
+dockerimage: demisto/python_zipfile:latest
```
ronykoz, line 39: Don't we have a tag number?

PR #6536, opened 2020-04-19
Netskope - fixed fetch bug

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22205

## Description
Fixed a logical error in Netskope's fetch incidents functionality.

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Legacy/Integrations/integration-Netskope.yml
```diff
@@ -293,13 +295,18 @@ script:
                 incident.name = item.alert_type + " - " + item.alert_name;
                 incident.rawJSON = JSON.stringify(item);
                 incidents.push(incident);
-                last_run = item.timestamp;
-                if (incidents >= Math.min(50, params.maxFetch)) {
+                if (latestIncidedentTimeInt == null || parseInt(item.timestamp) > latestIncidedentTimeInt) {
```
ronykoz, line 298: wouldn't it cause problems if latestIncidedentTimeInt is null?

guykeller: If it is null the first condition will be met and the second one won't even be checked

ronykoz:
@guykeller add RN please

PR #6550, opened 2020-04-20
SNOW scripts - packify 101

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23719

## Description
moved to the snow pack
packigify the scripts
removed the depends on ServiceNow, kept the depends on the commands
kept the servicenow tag, added a tag for servicenowv2

## Minimum version of Demisto
- [x] 4.5.0

   - [x] No


bakatzir:
reference: https://github.com/demisto/etc/issues/23708

PR #6558, opened 2020-04-20
 demisto/etc#23214: Added retry mechanism

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#23214

## Description
Since api.paloaltonetworks.com has a some performance issue it causes test failures such as [this](https://github.com/demisto/etc/issues/23718).
Implementing retry mechanism should help get over these performance issues.

The retry mechanism will not change previous behavior and will be executed only when the specifically use the relevant arguments.
  
## Screenshots

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -3275,7 +3277,21 @@ def __init__(self, base_url, verify=True, proxy=False, ok_codes=tuple(), headers
             if not proxy:
                 self._session.trust_env = False
 
-        def _http_request(self, method, url_suffix, full_url=None, headers=None,
+        def implement_retry(self, retries=0, status_list_to_retry=None):
+            retry = Retry(
+                total=retries,
+                read=retries,
+                connect=retries,
+                backoff_factor=10,
```
Itay4, line 3285: i think `10` is too much, lets set it to `5`

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -3275,7 +3277,21 @@ def __init__(self, base_url, verify=True, proxy=False, ok_codes=tuple(), headers
             if not proxy:
                 self._session.trust_env = False
 
-        def _http_request(self, method, url_suffix, full_url=None, headers=None,
+        def implement_retry(self, retries=0, status_list_to_retry=None):
```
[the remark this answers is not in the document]
hod-alpert: Added below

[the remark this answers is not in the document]
hod-alpert: Added below

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
Itay4: also, lets make this configurable with `5` as default value

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -3337,6 +3430,7 @@ def _http_request(self, method, url_suffix, full_url=None, headers=None,
                 address = full_url if full_url else urljoin(self._base_url, url_suffix)
                 headers = headers if headers else self._headers
                 auth = auth if auth else self._auth
+                self._implement_retry(retries, status_list_to_retry, backoff_factor, raise_on_redirect, raise_on_status)
```
anara123, line 3433: lets add a condition here, and only if `retries` parameter is provided - greater than 1 then the code will go into  `_implement_retry`.
This way all the existing code will not use that code (which potentially can introduce new bugs).

hod-alpert: That's actually is not a good idea.

The implementation changes the session's configurations and the session is the class's attribute, not the method.

So assuming I did one call with retry configuration- the next call I'll do with retries=0 will also retry like the previous call because the configuration part that should reset the retries is not performed now.

In any case- setting retries to 0 will fail the call after the first failure and will cause no background changes.

Itay4:
fails due to https://github.com/demisto/etc/issues/23880
merging

PR #6579, opened 2020-04-21
Create a pack for ThreatExchange

barchen1:
## Status
- [x] Ready

## Description
Create a pack for ThreatExchange and add readme file 

   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/ThreatExchange/Integrations/ThreatExchange/integration-ThreatExchange.yml
```diff
@@ -62,7 +62,7 @@ script:
                 Body: method == 'POST' ? encodeToURLQuery(parameters).substring(1) : ''
             },
             params.insecure,
-            params.useproxy
+            params.useproxyf
```
ronykoz, line 65: proxyf?

Packs/ThreatExchange/README.md
```diff
@@ -0,0 +1,402 @@
+## Overview
+---
+
+Receive threat intelligence about applications, IP addresses, URLs and hashes, a service by Facebook
+This integration was integrated and tested with version xx of ThreatExchange
```
ronykoz, line 5: update the version

ronykoz:
@kirbles19 could you please go over the readme?

Packs/ThreatExchange/README.md
barchen1: I dont think there is a version for it cause its a service by facebook

ronykoz: they should have a version.. if not let remove this line

kirbles19:
@barchen1 Is there anything to review here, or was the integration just put into package format?

barchen1:
@kirbles19 yes, The Readme file.

https://github.com/demisto/content/blob/1e98d06731a6cdd26f2a73aeea53b03cb27393be/Packs/ThreatExchange/README.md

kirbles19:
@barchen1 I think the changes should be made to the YML and then regenerate the README after.

ronykoz:
@kirbles19 this is not a new entity, it's an old integration which we converted to the new format and added readme to, if you think that we need to fix the yml as well that's ok, but i believe you already checked it out at some point.

kirbles19:
> @kirbles19 this is not a new entity, it's an old integration which we converted to the new format and added readme to, if you think that we need to fix the yml as well that's ok, but i believe you already checked it out at some point.

Even I can be edited.
I think it's worth taking a look at the YML and regenerating README.

ronykoz:
@kirbles19 sounds good to me

kirbles19:
@barchen1 @ronykoz Done. Can regenerate README now.

PR #6585, opened 2020-04-21
Remove default value from long running port parameters

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23744

## Description
Removed default values from long-running port parameters

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


DeanArbel:
@kirbles19 all the files of this PR require technical writer review

kirbles19:
@DeanArbel Done.

DeanArbel:
> Might break BC won't it?

Not really, since instances that were already created with that value should retain it.
Only newly created instances will have their value changed.

ronykoz:
@yaakovi is this right? Not too sure about this one

PR #6590, opened 2020-04-21
fixed get-report trendmicro dda

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22481

fixes get-report


Packs/TrendMicroDDA/Integrations/integration-TrendMicroDDA.yml
```diff
@@ -436,6 +452,8 @@ script:
                 hr['Detection Name'] = file_analyze_report['VirusName']['#text']
             #set the filename
             filename = hr['Detection Name'] if ('Detection Name' in hr) else file_analyze_report['FileSHA1']
+            if '.' not in filename:
+                filename = filename + ".txt"
```
bakatzir, line 456: I think it is not needed, as you set in in 533 and do not use the value until there

ronykoz: thought so too, but it's an if/else so two cases

PR #6596, opened 2020-04-22
PhishLabs fetch-fix

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23491
fixes: https://github.com/demisto/etc/issues/23926

## Description
fixed fetch-incidents command to not work all the time with the preset FETCH_TIME but to update according to the time passed since running the last time.


Packs/Legacy/Integrations/PhishLabsIOC/PhishLabsIOC.py
```diff
@@ -587,28 +588,47 @@ def get_feed_request(since: str = None, limit: str = None, indicator: list = Non
...
+    last_fetch_time: datetime = (datetime.strptime(last_fetch, '%Y-%m-%dT%H:%M:%SZ') if last_fetch
+                                 else datetime.strptime(NONE_DATE, '%Y-%m-%dT%H:%M:%SZ'))
 
     incidents: list = []
     count: int = 1
     limit = int(FETCH_LIMIT)
-    feed: dict = get_feed_request(since=FETCH_TIME)
-    last_fetch_time: datetime = (datetime.strptime(last_fetch, '%Y-%m-%dT%H:%M:%SZ') if last_fetch
-                                 else datetime.strptime(NONE_DATE, '%Y-%m-%dT%H:%M:%SZ'))
+    if not last_fetch:
```
ronykoz, line 621: so when is the NONE_DATE used?

roysagi: on first fetch only

roysagi:
> Why did we delete the offset logic here?

since it was wrong in the original code and it is unnecessary in this fix. I'll explain.

In the old code the fetch time was fixed let's say it 1 hour back.
Now I fetch 10 incidents my offset is now 9.

10 min pass and I fetch again - as I said the fetch_time was fixed so I again fetch 1 hour back - in this 10 min gap the fetched incident list could have changed as well.
Let's say one incident  that was fetched in the original fetch does not appear in the results of this new fetch.

my offset was 9 but one incident from the original bunch was not fetched so if I start processing from the 10th incident I actually skip an incident = hence the bug.

In this fix I simply fetch only the new incidents since the last fetched incident until now - without the arbitrary fetch time - hence offset is pointless.

Packs/Legacy/Integrations/PhishLabsIOC/PhishLabsIOC.py
```diff
@@ -587,28 +588,47 @@ def get_feed_request(since: str = None, limit: str = None, indicator: list = Non
     return response
 
 
+def get_min_time_delta(last_fetch_time):
+    # try in UTC first
+    fetch_delta = datetime.utcnow() - last_fetch_time
+    fetch_delta_in_min = fetch_delta.seconds
```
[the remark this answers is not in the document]
roysagi: sec - will change

Packs/Legacy/Integrations/PhishLabsIOC/PhishLabsIOC.py
```diff
@@ -587,28 +588,47 @@ def get_feed_request(since: str = None, limit: str = None, indicator: list = Non
...
+
 def fetch_incidents():
     """
     Fetches incidents from the PhishLabs user feed.
     :return: Demisto incidents
     """
     last_run: dict = demisto.getLastRun()
     last_fetch: str = last_run.get('time', '') if last_run else ''
-    last_offset: str = last_run.get('offset', '0') if last_run else '0'
+    last_fetch_time: datetime = (datetime.strptime(last_fetch, '%Y-%m-%dT%H:%M:%SZ') if last_fetch
```
ronykoz, line 615: Seems that you don't have last fetch you will go to NONE_DATE and will never use FETCH_TIME am i missing something?

roysagi: FETCH_TIME is how far back in time to fetch for the first time.
NON_DATE is a base date = '0001-01-01T00:00:00Z' - to be saved in ***last_fetch_time*** on the first fetch (that is if ***last_fetch*** is None).

The first fetch happens in line 621 of the code in case ***last_fetch*** is none - there you can see it uses FETCH_TIME to bring the first batch of incidents.

***last_fetch_time*** is used later to make sure you only bring back new incidents (line 639) but this is irrelevant for the first fetch and only comes into play on second fetch.

roysagi:
@ronykoz - added a fix for this: https://github.com/demisto/etc/issues/23926 so if you could please review again

roysagi:
@ronykoz - added the email fix to PhishLabsPopulateIndicators if you want to have a look (same code as in PhishLabsIOC

roysagi:
@amshamah419 - if you can please check my RN.

amshamah419:
Release notes look good

roysagi:
@ronykoz - only fails on the validate of the metadata file

PR #6599, opened 2020-04-22
[Test Filter] Added id_set initialization at the beginning of get_test_list

DeanArbel:
## Status
- [x] Ready

## Description
Added id_set initialization at the beginning of `get_test_list` to make sure it gets passed correctly.


Tests/scripts/configure_tests.py
```diff
@@ -931,6 +931,9 @@ def get_test_list(files_string, branch_name, two_before_ga_ver='0', conf=None, i
     """Create a test list that should run"""
     (modified_files, modified_tests_list, changed_common, is_conf_json, sample_tests, is_reputations_json,
      is_indicator_json) = get_modified_files(files_string)
+    if not id_set:
```
anara123, line 934: instead of loading id_set from file here, lets load it once in `create_test_file` file and pass it to get_test_list

PR #6604, opened 2020-04-22
MicrosoftApiModule http _request - Handling HTTP 206

Bargenish:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
reference: demisto/etc#20893

## Description
Added status code verification inside http_request method to catch HTTP 206 special Microsoft Graph case.
Added support for provided headers.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?


Packs/ApiModules/Scripts/MicrosoftApiModule/MicrosoftApiModule.py
```diff
@@ -75,20 +75,44 @@ def __init__(self, tenant_id: str = '',
...
+        if headers:
+            default_headers.update(headers)
+
+        response = super()._http_request(   # type: ignore[misc]
+            resp_type="response", headers=default_headers, *args, **kwargs)
+
+        # 206 indicates Partial Content, reason will be in the warning header.
+        # In that case, logs with the warning header will be written.
+        if response.status_code == 206:
+            demisto.debug(str(response.headers))
```
yaakovi, line 100: could headers contain personal data? better print only the relevant keys

Packs/ApiModules/Scripts/MicrosoftApiModule/MicrosoftApiModule.py
```diff
@@ -75,20 +75,44 @@ def __init__(self, tenant_id: str = '',
         self.auth_type = SELF_DEPLOYED_AUTH_TYPE if self_deployed else OPROXY_AUTH_TYPE
         self.verify = verify
 
-    def http_request(self, *args, **kwargs):
+    def http_request(self, headers=None, *args, **kwargs):
```
yaakovi, line 78: you can't pass keyword argument before positional args
```suggestion
    def http_request(self, *args, headers=None, **kwargs):
```

Packs/ApiModules/Scripts/MicrosoftApiModule/MicrosoftApiModule.py
Bargenish: No, in this case, it would contain fields that relate to the partial range download, such as `Content-Range` or `Content-Type`.
e.g:

```
HTTP/1.1 206 Partial Content
Content-Range: bytes 0-1023/2048
Content-Type: application/octet-stream

<first 1024 bytes of file>
```

kirbles19:
@Bargenish Done, added changelog description.

PR #6607, opened 2020-04-22
[CircleCi Build] Skip Env Creation If Skipped

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23172

## Description
Added file `filter_envs.json` which is a map of envs that have runnable tests, as created by `configure_tests.py`. This file is to be used in the env creation, in order to skip the creation of that env.


DeanArbel:
@ronykoz please review https://github.com/demisto/content-test-conf/pull/799 as well

PR #6611, opened 2020-04-22
AbuseDB - fixed rate limit behavior

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23385

## Description
A bad error message would be thrown when the API quota was reached if the "Disregard quota errors" was set to True.
This is now fixed.

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


guykeller:
Closing this and opening a new one.

PR #6420, opened 2020-04-07
Added an argument for different blacklisted url handling

guykeller:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/23344

## Description
Until now, if any url in your list was blacklisted, an error would have been returned.
That will now depend on an argument

## Does it break backward compatibility?
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?
- [ ] The title must be in the following format: **[YOUR_PARTNER_ID] short description**

PR #6618, opened 2020-04-23
Urlscan - fixed error concerning blacklisted urls

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23344

## Description
Until now, if any url in your list was blacklisted, an error would have been returned.
That will now depend on an argument

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


guykeller:
Exactly the same as this one:
https://github.com/demisto/content/pull/6420
You both already approved it - FYI

@ronykoz @roysagi

PR #6636, opened 2020-04-23
Marketplace - update index fix

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
In case that new pack is uploaded, need to create new pack folder in index. `shutill.copy` doesn't handle it. Received error: [Errno 20] Not a directory: `/tmp/tmp.ZbcYkPjdus/index/MicrosoftGraphCalendar/metadata-1.0.0.json`

https://circleci.com/gh/demisto/content/49515#artifacts/containers/0


IkaDemisto:
This should be merged before the phases of integrations, otherwise all new packs won't be updated to `index`.

IkaDemisto:
@liorblob I will add unit test for it in separate PR, the build finally past the setup instance part. Don't want to waste it 😃

PR #6641, opened 2020-04-24
CommonServerPython: Dbot score Class add indicator type to DT

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
We should take into account the indicator type when doing the DT. It is possible to have two indicators from different types with the same value.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests (Unit)
- [ ] Documentation 

## Demisto Partner?


glicht:
Didn't update the changelog as the entry we have is good.

anara123:
@glicht I think if you already opened this PR and in that area. Please add this fix too. Replace the `get_context_path` function in `DBotScore`
```python
        def get_context_path():
            try:
                if get_demisto_version().get('version') >= '5.5.0':
                    return Common.DBotScore.CONTEXT_PATH
            except AttributeError:
                return Common.DBotScore.CONTEXT_PATH_PRIOR_V5_5
```

glicht:
@anara123 see the fix I put in for dbotscore and I added is_demisto_version_ge function

glicht:
Lint failed on a bug in Demisto sdk

PR #6649, opened 2020-04-26
Archer - fixed display of type 4 fields in search records

guykeller:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23125

## Description
Type 4 fields were not returned to context, nor shown in the human readable, when running the search records command.
This is now fixed.

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Legacy/Integrations/integration-ArcherRSA.yml
+++ b/Packs/Legacy/Integrations/integration-ArcherRSA.yml
@@ -1467,9 +1467,9 @@ script:
 
             var fields = verifyOrMakeArray(record.Field);
             fields.forEach(function (field) {
-                if (field['#text'] != null) {
+                if (field['#text'] !== null) {
                     toContext.Fields[GLOBAL_MAPPING[field['-id']].Name] = field['#text'];
-                } else if (field.ListValues != null && field.ListValues['#text'] != null) {
+                } else if (field.ListValues !== null && field.ListValues['#text'] !== null) {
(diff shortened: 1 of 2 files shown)


Packs/Legacy/Integrations/integration-ArcherRSA.yml
```diff
@@ -1467,7 +1467,13 @@ script:
 
             var fields = verifyOrMakeArray(record.Field);
             fields.forEach(function (field) {
-                toContext.Fields[GLOBAL_MAPPING[field['-id']].Name] = field['#text'];
+                if (field['#text'] != null) {
+                    toContext.Fields[GLOBAL_MAPPING[field['-id']].Name] = field['#text'];
+                } else if (field.ListValues != null && field.ListValues['#text'] != null) {
+                    toContext.Fields[GLOBAL_MAPPING[field['-id']].Name] = field.ListValues['#text'];
+                } else {
+                    toContext.Fields[GLOBAL_MAPPING[field['-id']].Name] = field['#text'];
```
ronykoz, line 1475: this seems identical to the first condition

guykeller: Fixed now - did not change the logic (this condition is  equivalent to null at this point, which is the desired behavior) but wrote `null` explicitly for code readability and clarity

guykeller:
Opened new PR.

PR #6661, opened 2020-04-26
AbuseDB - fixed quota limit handling logic

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23385

## Description
A bad error message would be thrown when the API quota was reached if the "Disregard quota errors" was set to True.
This is now fixed.

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


guykeller:
@ronykoz same as https://github.com/demisto/content/pull/6611 but I had to delete the Legacy copy of it, apparently it shouldn't exist (consulted with @IkaDemisto )

PR #6680, opened 2020-04-26
Revert " demisto/etc#23214: Added retry mechanism"

Itay4:
Reverts demisto/content#6558

need to revert this as some docker images don't have module requests.adapters


Itay4:
fails due to non-relavent validation
merging

PR #6699, opened 2020-04-27
Integrations to PAN-OS Pack

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
Moved the following from Legacy to PAN-OS Pack:
- BPA, PAN_OS_EDL, Panorama Integrations
- Relevant playbooks, scripts, IndicatorTypes, IndicatorFields, TPBs


Packs/PAN-OS/pack_metadata.json
```diff
@@ -0,0 +1,23 @@
...
+    "support": "xsoar",
+    "currentVersion": "1.0.0",
+    "author": "Cortex XSOAR",
+    "url": "https://www.paloaltonetworks.com/cortex",
+    "email": "",
+    "categories": [
+        "Utilities", "Network Security"
+    ],
+    "tags": [
+        "PAN-OS"
```
bakatzir, line 13:
```suggestion
        "PAN-OS",
        "NetOps",
        "Panorama"
```

bakatzir: Not sure if it should go to here or to key words.
but we definitely want to add use cases as well to this pack.

bakatzir:
I think all these Playbooks should move to tho the pack, as they are pan-os exclusive.
but let's. wait to see what Michal opinion is as well.

michalgold:
@dantavori Exclude BPA and EDL. They should not be in the PAN-OS pack but should have a dedicated one.

bakatzir:
@michalgold 
Could you elaborate why?
Is not the `BPA` and the `PAN-OS EDL management` integrations are being using exclusively for pan-os products(panorama+firewall)?

michalgold:
1. We'll probably have EDL management deprecated by 6.0. Replaced by EDL service. 
2. PAN-OS users are not necessary using BPA (most of them are not). BPA will have a dedicated layout and maybe dashboard, it's playbook is not at all depended on PAN-OS. It should have it's own pack.

PR #6700, opened 2020-04-27
Integrations to packs - Phase 4 (Dan)

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Description
Moved the following integrations from Legacy pack to separate packs:
- CofenseTriage
- Confluence
- CounterCraft
- CounterTack
- CrowdStrikeFalcon
- Cybereason
- Cylance_Protect_v2
- Digital_Defense_FrontlineVM
- dnstwist
- DomainTools_Iris


Packs/Confluence/Integrations/Confluence/Confluence.yml
```diff
@@ -29,6 +29,7 @@ configuration:
 script:
   script: ''
   type: python
+  subtype: python3
```
yuvalbenshalom, line 32: I think that since there is no docker, this may actually be python2 by default.
need to check this.

dantavori, line 32: @yuvalbenshalom so it needs to be python2?

dantavori, line 32:
```suggestion
  subtype: python2
```

dantavori:
@yuvalbenshalom I think it's a bug in configure test filter, I'm pulled from master

PR #6701, opened 2020-04-27
SNOW v2 -Added the *system_params* argument to the ***servicenow-quer…

bakatzir:
…y-table*** command.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23429

## Description
per design discussed with @aneeshamore in the referenced issue

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


amshamah419:
RN looks good

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -617,18 +617,23 @@ def upload_file(self, ticket_id: str, file_id: str, file_name: str, ticket_type:
...
         table_name: table name
         sys_param_limit: limit the number of results
         sys_param_offset: offset the results
         sys_param_query: the query
+        system_params: system parameters
 
         Returns:
             Response from API.
         """
+        if system_params:
```
liorblob, line 634: Why not append them to the query params below?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -1582,6 +1588,7 @@ def main():
             demisto.results(get_ticket_command(client, args))
         elif command in commands:
             md_, ec_, raw_response, ignore_auto_extract = commands[command](client, args)
+            demisto.log(str(ec_))
```
liorblob, line 1591: oops

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -1546,6 +1546,14 @@ script:
       name: offset
       required: false
       secret: false
+    - default: false
+      description: 'System parameters. overrides the query, limit and offset arguments.
```
liorblob, line 1550: it should not override the other query params IMO. Just add more params.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
bakatzir: Because we said that we want to override all the system params e.g offset, limit and. query. no?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
bakatzir: If we want it not. to override at all and just add, we need to make it key=val, and then check waht happens if one of these keys are, e.g `sysparm_limit`.
I think this is not the preferred way.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
liorblob: If so, the params need to be sent as dict to the request.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
liorblob: I think that usually users will want to add additional params, they will not like to add the query and limit themselves because it's already there. @Arsenikr WDYT?

bakatzir:
@liorblob Agreed.
refactored accordingly, and also added the option to have `key=va=l` custom field.
please re review.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -1546,6 +1546,13 @@ script:
       name: offset
       required: false
       secret: false
+    - default: false
```
liorblob, line 1549: We should add it to the query-tickets command as well

liorblob: And possibly to the computers/groups/users command maybe, not mandatory

bakatzir:
@liorblob correct again. adding it only to `query-ticketts` for now, as if the customer is using a more "boxed" command e.g the query computers/users, I do no think it is needed.

bakatzir:
fixes: https://github.com/demisto/etc/issues/24113

PR #6705, opened 2020-04-28
Securonix detailed desc

bakatzir:
## Status
- [x] Ready

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Itay4:
lets also add it to the integration doc

Packs/Securonix/Integrations/Securonix/README.md
```diff
@@ -2,6 +2,11 @@
 Use the Securonix integration to manage incidents and watchlists.
 Integration was build and tested with SNYPR Version 6.3.
```
Itay4, line 3: do we know if this is the version for both cloud and on-prem ?

bakatzir: No, as we get access to  cloud base only. I know for a fact that it is working for an on prem of version 6.2 as well.

bakatzir: according to the vendor, API should be compatible for both deployments.

Packs/Securonix/Integrations/Securonix/README.md
```diff
@@ -1,6 +1,11 @@
 ## Overview
 Use the Securonix integration to manage incidents and watchlists.
-Integration was build and tested with SNYPR Version 6.3.
+Integration was build and tested with SNYPR Versions: 6.2, 6.3.
```
amshamah419, line 3: Integration was built and tested with SNYPR Versions: 6.2, 6.3.

PR #6708, opened 2020-04-28
GenericSQL - fix ssl connection

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23743

## Description
Fixed the option to connect with ssl 
in order to make it user friendly the option was added as a button

   - [ ] Yes
       - Further details:
   - [x] No


Packs/GenericSQL/Integrations/GenericSQL/GenericSQL.py
```diff
@@ -63,6 +64,11 @@ def _create_engine_and_connect(self) -> sqlalchemy.engine.base.Connection:
             elif self.connect_parameters and self.dialect != "Microsoft SQL Server":
                 # a "?" was already added when the driver was defined
                 db_preferences += f'?{self.connect_parameters}'
+
+            if self.ssl_connect:
```
bakatzir, line 68: can you have connect_parameteres and ssl_connect?
If so, this code will not allow it.

ohaim1008: yes, tried and it looks ok, I tried to add the ssl dict to the connect_parameters in the way we currently use - "?arg1=something" but it doesn't support nested dictionaries. 
why won't it? the other connection_args are in the db_preferences string.

bakatzir: You are right. can we have 1 return statement, and `ssl_connection` would be an empty dict if the `ssl_connect` is not true?
It feels redundant to have different return statement for this flow.

ohaim1008: sure, it will just be empty.
change was made.

bakatzir:
@kirbles19 we have added a new parameter to the integration, and would love your help.

kirbles19:
@ohaim1008 Done

PR #6729, opened 2020-04-30
Integrations to packs phase 7

liorblob:
Integrations:
~~PaloAltoNetworks_Traps~~
~~PaloAlto_MineMeld~~
Palo_Alto_Networks_WildFire_v2
McAfee_Advanced_Threat_Defense
McAfee-TIE
MaxMind_GeoIP2
MailSenderNew
Looker
LogRhythmRest
MicrosoftGraphMail
MicrosoftGraphListener
MicrosoftGraphFiles

@michalgold FYI on the PANW integrations


michalgold:
1. Minemeld should be deprecated. Not relevant from version 5.5. 
2. Traps might need to be deprecated as well, I'll check. 
3. Don't move the following playbooks as they are old and we need to review and approve them:
- TIE- IOC Hunt
- Search endpoints by hash- TIE.

PR #6737, opened 2020-04-30
Archer - fixed numerous issues

guykeller:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19721

## Description
Archer fetch incidents, create records and search records were not working for the 411 app ID.
That problem is now fixed.

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ArcherRSA/Integrations/integration-ArcherRSA.yml
```diff
@@ -3545,22 +3562,36 @@ script:
         return 'ok';
     }
 
+    function mergeDicts(dictA, dictB) {
```
ronykoz, line 3565: what happens if you have the same key in both dicts

PR #6742, opened 2020-04-30
Securonix int incident is not needed, fetch is limited by the vendor.

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23956

## Description
Securonix limit the fetch to the last 10 incidents. to chnge to the `limit` or `order` is yet available.
Added to the readme, customer will communicate this to the vendor to fix ASAP.

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Securonix/Integrations/Securonix/Securonix.py
```diff
@@ -1135,23 +1136,31 @@ def fetch_incidents(client: Client, fetch_time: Optional[str], incident_status:
...
+        if total_incidents > 10:
+            demisto.info(f'Fetching Securonix incidents. With offset: {total_incidents-10}')
+            securonix_incidents = client.list_incidents_request(from_epoch, to_epoch, incident_status,
+                                                                int(total_incidents - 10))
+
         incidents_items = list(securonix_incidents.get('incidentItems'))  # type: ignore
-        last_incident_id = last_run.get('id', '0')
+        # 1st incident is the latest
+        if incidents_items:
+            now = timestamp_to_datestring(incidents_items[0].get('lastUpdateDate'))
```
yuvalbenshalom, line 1149: 1) is lastUpdateDate is the correct value to use? shouldn't it be creation time?
2)you are using the exact time dose the query in Securonix brings every thing that is ">= now"
or "> now"?

bakatzir: Asked the vendor both, still waiting for reply.

bakatzir: See comments in the referenced issue, there are issues with the current way listing the incidents was implemented by the vendor.

PR #6743, opened 2020-04-30
fix xfe suggestions

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24020

## Description
fix xfe suggestions


yaakovi:
@DeanArbel why `HelloWorld-Test` is collected here? it is not relevant for any of the affected items

DeanArbel:
@yaakovi I suspect it's because it implements `ip` and `domain` commands.
There is an open issue to fix this. https://github.com/demisto/etc/issues/22708

PR #6746, opened 2020-04-30
Address a Few RedCanary Integration Bugs

blakemotl:
## Status
- [x] In Progress
- [ ] Ready

## Description

Bug #1: No incidents are created due to a bug in the exclusion logic around acknowledgement of an incident in RedCanary. Looking at this [line](https://github.com/blakemotl/content/blob/master/Packs/Legacy/Integrations/RedCanary/RedCanary.py#L323) last_acknowledged_at and last_acknowledged_by are guaranteed to exist as confirmed by RedCanary directly, but default to null. As a result this line always evaluates to false.

Bug #2: There is a bug in the integration that results in the last incident in RedCanary, being repeatedly created in Demisto. If a new incident is created in RedCanary, the old incident will stop repeating and the new one starts repeating. I was not able to figure out exactly whats causing it, but it looks to be some kind of rounding error, since the integration cuts off the millisecond part of the timestamp. Additionally it is possible, but unlikely, that an incident can be skipped due to a timing error since the integration stores a timestamp that skips a second ahead to attempt to prevent duplication. However RedCanary already exposes a way to enforce uniqueness via the ID field. That field is unique per incident just like the incident ID in Demisto is unique. So the easiest way to fix this bug is to keep track of the latest ID that was created within Demisto and store that in addition to the time. That change starts [here](https://github.com/blakemotl/content/blob/master/Packs/Legacy/Integrations/RedCanary/RedCanary.py#L522).

## Minimum version of Demisto
- [ ] 4.5.0
- [X] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Testing
We have deployed this in our Demisto environment and both bugs above have been resolved.

### DIFF
--- a/Packs/Legacy/Integrations/RedCanary/RedCanary.py
+++ b/Packs/Legacy/Integrations/RedCanary/RedCanary.py
@@ -1,7 +1,3 @@
-import demistomock as demisto
-from CommonServerPython import *
-from CommonServerUserPython import *
-
 ''' IMPORTS '''
 import requests
 
@@ -66,14 +62,12 @@ def http_request(requests_func, url_suffix, **kwargs):
     params = kwargs.get('params')
     headers = kwargs.get('headers', {})
     data = kwargs.get('data', {})
-
     res = requests_func(BASE_URL + url_suffix,
                         verify=USE_SSL,
                         params=params,
                         headers=headers,
                         data=data
                         )
-
     if res.status_code == 403:
         raise Exception('API Key is incorrect')
 
@@ -162,7 +156,7 @@ def get_full_timeline(detection_id, per_page=100):
                            'per_page': per_page,
         })
 
-        if len(res['data']) == 0:
+        if len(res['data']) == 0 or True:
             done = True
 
         activities.extend(res['data'])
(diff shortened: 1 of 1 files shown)


reutshal:
> [CLA assistant check](https://cla-assistant.io/demisto/content?pullRequest=6746)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/demisto/content?pullRequest=6746) before we can accept your contribution.
> 
> **Blake Motl** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/demisto/content?pullRequest=6746) it.

@blakemotl Can you please sign the `Contributor License Agreement` so we can accept your contribution? Thanks

reutshal:
@blakemotl Also, there are some conflicts in the PR, which caused by some re-organization from our side- the integration was moved to `Packs/RedCanary/Integrations/RedCanary/RedCanary.py`. We have a few options:
1. You can update your PR - the best way will be to save the changes and paste them in the right place.
2. You can open a new PR from the updated content.
3. We can take it from here and add your changes by ourselves.
Please tell us what you would like to do. Thanks in advance!

ronykoz:
@blakemotl could you please sign the CLA? we need that in order to merge the PR

ronykoz:
@blakemotl we are closing this PR due to it's being in-active, and we can't add these changes without the CLA signing

PR #6750, opened 2020-05-01
Integrations to packs - Phase 6 (Dan) 

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
Moved the following from Legacy to packs:
- integration-Aella_StarLight.yml
- integration-Algosec.yml
- integration-AlienVailtOTX.yml
- integration-AlphaSOC_Network_Behavior_Analytics.yml
- integration-AlphaSOC_Wisdom.yml
- integration-AMP.yml
- integration-AnomaliThreatStream.yml
- integration-ArcherRSA.yml
- integration-ArcSightLogger.yml
- integration-CarbonBlackDefense.yml


dantavori:
@yuvalbenshalom I'm not sure if the following repeated error in validate is truly an error:
```You have added an image in the yml file, please update the package Packs/AlienVault_OTX_v2/Integrations```
If so, what should we do about them?

In addition to that, we have `AlienVailt OTX` integration which I'm not sure about, first of all - should it be in a pack? It's not deprecated but we also have `AlienVault OTX` (which is deprecated and a bit newer) and `AlienVault OTX v2`, not to mention the typos in the filename ("Vailt") and in the display name ("Valut"). It also has an error in the URL reputation command output which I'm not sure if I need to handle.

yuvalbenshalom:
`AlienVailt OTX` has `toversion: 3.0.1` so you can move it to non supported along side with deprecated `AlienVault OTX`.

and I think we should chang the pack name from "AlienVault OTX v2" to "AlienVault OTX" including the path. 

after that we will need to delete the v2 from the bucket.

PR #6753, opened 2020-05-03
PCAPHTTPExtractor - issue with file containing a response but not a request.

reutshal:
## Status
- [] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23807

## Description
Fixed an issue in a pcap file which contains http response but not a request.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
```diff
@@ -206,27 +219,34 @@ def get_http_flows(pcap_file_path):
...
-        #         #     res["HTTP"].data = decode_gzip(res["HTTP"].data)
-        #         # except zlib.error:
-        #         #     res["HTTP"].data = "Couldn't decompress gzip data (incomplete or truncated stream)."
+        sanitized_req = req
+        sanitized_res = res
+
+        if req:
+            req_fields = req['HTTP'].__dict__["_all_fields"].keys()
+            if 'http.response' in req_fields:
+                sanitized_req = None
```
ronykoz, line 244: why are we overriding the request here?

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
```diff
@@ -206,27 +219,34 @@ def get_http_flows(pcap_file_path):
...
+                sanitized_res = req
```
ronykoz, line 245: res = req? why? add an explanation comment maybe

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
```diff
@@ -195,6 +195,19 @@ def fix_order(http_packets):
...
+def process_tcp_stream(tcp_packets):
+    stream = []
+    for t in tcp_packets:
+        if 'segment_data' in dir(t.tcp):
+            tcp_data = {}
+            segment_hex = re.sub('[:]', '', t.tcp.segment_data)
+            tcp_data['data'] = bytearray.fromhex(segment_hex).decode()
+            tcp_data['stream_number'] = t.tcp.stream
+            stream.append(tcp_data)
+    print(len(stream))
```
ronykoz, line 207: remove

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
reutshal: The bug here is that sometimes the file is containing only response and not a request.
In that case, the response will appear inside the `req` variable.
So, if that is the case, we put None in the req (because it doesn't really exist) and putting the response in the res variable as it should be.

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
reutshal: The same explanation as to the above

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
```diff
@@ -195,6 +195,18 @@ def fix_order(http_packets):
     return new_order
 
 
+def process_tcp_stream(tcp_packets):
```
teizenman, line 198: Im not familiar with some of the functions used here, any reason that they will fail? 
If so, maybe add some error handling mechanism here?

Packs/Legacy/Scripts/PcapHTTPExtractor/PcapHTTPExtractor.py
```diff
@@ -206,27 +218,34 @@ def get_http_flows(pcap_file_path):
 
     # Filter all non HTTP packets
     http_packets = [p for p in capture_object if "HTTP" in p]
-
     http_packets = fix_order(http_packets)
 
+    tcp_packets = [q for q in capture_object if 'http' not in q]
+    tcp_stream = process_tcp_stream(tcp_packets)
+    markdown_output = tableToMarkdown("TCPStreamInfo", tcp_stream)
```
teizenman, line 225: maybe add headers to the table?

reutshal:
> Actually, why are we parsing the TCP packets here for?

Removing as discussed.

reutshal:
@roysagi Can you please go over the CHANGELOG file? https://github.com/demisto/content/blob/PCAPHTTPExtractor-bug_fix/Packs/Legacy/Scripts/PcapHTTPExtractor/CHANGELOG.md
thanks

PR #6757, opened 2020-05-03
Genericsql - when query returns an empty table don't return an error

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23354

## Description
When a sql query return an empty table an error return - fixing it to a valid output "No entries.".

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/GenericSQL/Integrations/GenericSQL/GenericSQL_test.py
+++ b/Packs/GenericSQL/Integrations/GenericSQL/GenericSQL_test.py
@@ -1,6 +1,15 @@
 from GenericSQL import Client, sql_query_execute
 import pytest
 
+
+class ConnectionMock:
+    def __init__(self):
+        self.execute = []
+
+    def fetchall(self):
+        return []
+
+
 args1 = {
     'query': "select Name from city",
     'limit': 5,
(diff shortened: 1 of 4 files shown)


ohaim1008:
making a small change in the unitest - will take couple of min and then will be ready.

Packs/GenericSQL/Integrations/GenericSQL/GenericSQL_test.py
```diff
@@ -163,10 +172,41 @@ def test_sql_queries(command, args, response, expected_result, header, mocker):
...
+    - validate the expected_result and the created context
     """
     mocker.patch.object(Client, '_create_engine_and_connect')  # needed in order not to make a connection in tests
     mocker.patch.object(Client, 'sql_query_execute_request', return_value=(response, header))
     client = Client('sql_dialect', 'server_url', 'username', 'password', 'port', 'database', "")
     result = command(client, args)
     assert expected_result == result[1]  # entry context is found in the 2nd place in the result of the command
+
+
+empty_output = {
```
bakatzir, line 184: should no this be inside `test_sql_queries_with_empty_table`?

Packs/GenericSQL/Integrations/GenericSQL/GenericSQL_test.py
```diff
@@ -163,10 +172,41 @@ def test_sql_queries(command, args, response, expected_result, header, mocker):
...
+    - mock the database result
+    Then
+    - convert the result to human readable table
+    - create the context
+    - validate the expected_result and the created context
+    """
+    mocker.patch.object(Client, '_create_engine_and_connect')
+    client = Client('sql_dialect', 'server_url', 'username', 'password', 'port', 'database', "")
+    mocker.patch.object(client.connection, 'execute', return_value=ConnectionMock())
+    result = sql_query_execute(client, args1)
```
bakatzir, line 211: you can hard code the args, though response is mocked to demonstrate that it will return no data, as using a variable that is defined outside of this function is not a best practice.

PR #6762, opened 2020-05-03
Snow v2 service catalog UC implementation

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22557

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -341,12 +371,14 @@ class Client(BaseClient):
     Client to use in the ServiceNow integration. Overrides BaseClient.
     """
 
-    def __init__(self, server_url: str, username: str, password: str, verify: bool, proxy: bool, fetch_time: str,
-                 sysparm_query: str, sysparm_limit: int, timestamp_field: str, ticket_type: str, get_attachments: bool):
+    def __init__(self, server_url: str, sc_server_url: str, username: str, password: str, verify: bool, proxy: bool,
```
liorblob, line 374: We don't want another client for the service catalog?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -2021,6 +2021,112 @@ script:
     - contextPath: ServiceNow.Ticket.TagID
       description: Ticket tag ID.
       type: String
+  - arguments:
+    - default: false
+      description: Query by name. Match by contains and not by exact name.
```
liorblob, line 2026: It can also be matched by exact name... this is a bit confusing.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
bakatzir: Since it is one more arg in he ini I think we can do without. Do you think it is needed?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
bakatzir: Agreed. I'll ask @amshamah419 to the rescue.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
liorblob: might be overkill. Just I'm not sure that extra arg is very clear. We should add some comments then.

bakatzir: Added in `send_request` and in the relevant client methods.

PR #6763, opened 2020-05-03
Remove incident filter log from tests

Itay4:
## Status
- [x] Ready

## Description
Removed the incident filter print in the tests log


Itay4:
those actually might have value when we manage to reproduce the issue, so yea we need them

PR #6764, opened 2020-05-03
Fix claroty readme

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
moved integration README to right location.
added an empty README file for the pack.

adding a `docs-approved` label as I didn't change the content of the file, just the location.


asafshen:
Can we check how it looks in a new server?

yaakovi:
should be available once the build finishes `Setup Instance` step

PR #6774, opened 2020-05-04
search and install packs

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
Improved search_and_install_packs.py


dantavori:
@yuvalbenshalom 
```Packs Successfully Installed!```
Also the packs installed are the packs found in the prints above, and if there are dependencies it will be printed as well (in our example there aren't such).

PR #6779, opened 2020-05-04
SNOW v2 - Clarified descs for  impact, urgeny args in CRUD ticket commands

bakatzir:
in the commands:
    - ***servicenow-update-ticket***
    - ***servicenow-create-ticket***

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21294

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


liorblob:
Did not understand the descriptions. What are you trying to explain?

bakatzir:
@liorblob That you can put free text in these arguments if you do not want to use the pre defined in the list. and if you have it configured in your ServiceNow, it will work. 
can you / @amshamah419 help me out?

liorblob:
@bakatzir I might say something like `Either choose from the dropdown or specify any other value, e.g., "Urgent" or "5".` (just an example).

Packs/ServiceNow/Integrations/ServiceNowv2/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## [Unreleased]
+  - Clarified the documentation for usage of the ***impact*** and the ***urgency*** arguments in the commands:
+    - ***servicenow-update-ticket***
+    - ***servicenow-create-ticket***
```
amshamah419, line 4:
```suggestion
  - Improved documentation regarding the usage of the ***impact*** and the ***urgency*** arguments for the commands:
    - ***servicenow-update-ticket***
    - ***servicenow-create-ticket***
```

amshamah419: I'm inclined to agree with @liorblob. The most important message to convey is that the end-user can select from the prepopulated values, or override it using some sort of template. 

I would suggest the following:

> 'Ticket urgency. You may select from the predefined options, or define a custom field by using the following syntax. "Number - Label". (For example, "4 - Medium")

amshamah419:
I'm inclined to agree with @liorblob. The most important message to convey is that the end-user can select from the prepopulated values, or override it using some sort of template. 

I would suggest the following:

> 'Ticket urgency. You may select from the predefined options, or define a custom field by using the following syntax. "Number - Label". (For example, "4 - Medium")

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -213,8 +213,8 @@ script:
       secret: false
     - auto: PREDEFINED
       default: false
-      description: 'Ticket urgency. Giving a custom value which predefined in ServiceNow,
-        e.g: ''5 - Informational'' is available.'
+      description: 'Ticket urgency. You may select from the predefined options, or
+        define a custom field by using other value. e.g: "Urgent", "1 - Informational".'
```
liorblob, line 217: @bakatzir @amshamah419 I think `custom field` term will confuse SNOW users. 
```suggestion
         use another value. e.g: "Urgent" or "5".'
```

amshamah419: Is it just "$Label" or can it be both "$INT - $Label"?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -768,7 +768,7 @@ script:
     - auto: PREDEFINED
       default: false
       description: 'Ticket severity. You may select from the predefined options, or
```
liorblob, line 770: Pretty sure we already had severity

PR #6794, opened 2020-05-05
[Test Fix] Rapid 7 Nexpose

DeanArbel:
## Status
- [x] Ready

## Description
Fixed Rapid 7 Nexpose test playbook.


DeanArbel:
@yaakovi Since it's flaky - our instance sometimes recognizes the vulnerabilities and other times it doesn't, so the task would randomly fail, since the instance had no vulnerabilities.

PR #6804, opened 2020-05-05
[Test Collection] Change random test collection seed

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22711

## Description
Changed random generation seed to be based on branch name + git diff.


Tests/scripts/configure_tests.py
```diff
@@ -975,7 +974,8 @@ def get_test_list(files_string, branch_name, two_before_ga_ver='0', conf=None, i
         tests = tests.union(get_test_from_conf(branch_name, conf))
 
     if not tests:
-        tests = get_random_tests(tests_num=RANDOM_TESTS_NUM, conf=conf, id_set=id_set, server_version=two_before_ga_ver)
+        rand = random.Random(files_string + branch_name)
```
yaakovi, line 977: why not only branch name?

glicht: I also think it is better just the branch name.

DeanArbel: It was like that before, but I'll switch to just branch name
```suggestion
        rand = random.Random(branch_name)
```

PR #6817, opened 2020-05-06
Fetch incident

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: demisto/etc/issues/18143

## Description
Adding fetch-incidents option to the integration

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

## Demisto Partner?


Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
```diff
@@ -95,6 +95,61 @@ def get_user(self, user_id):
...
+    if incidents:
+        count = 0
+        incidents = sorted(incidents, key=lambda k: k['createdDateTime'])  # sort the incidents by time-increasing order
+        last_incident_time = last_run.get('time', '0')
+        demisto.debug(f'Incidents times: {[incidents[i]["createdDateTime"] for i in range(len(incidents))]}\n')
+        for incident in incidents:
+            incident_time = incident.get('createdDateTime')
+            if incident_time > last_incident_time and count < fetch_limit:
+                demisto_incidents.append({
+                    'name': incident.get('title'),
```
yuvalbenshalom, line 137: is title unique to every alert? if so good, if not wi should consider adding an id of the alert to it as the incident name.
also we should consider adding the integration name. lets talk about this.

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
```diff
@@ -95,6 +95,61 @@ def get_user(self, user_id):
...
+                    'severity': severity_map.get(incident.get('severity', ''), 0),  # todo: check with Arseny
+                    'rawJSON': json.dumps(incident)
+                })
+                count += 1
+        if demisto_incidents:
+            last_incident_time = demisto_incidents[-1].get('occurred')
+            new_last_run.update({'time': last_incident_time})
+
+    if not demisto_incidents:
+        new_last_run.update({'time': now})
```
yuvalbenshalom, line 148: if you got no demisto incidents, you can keep the last run avoid the race condition that an alert was created just now.(so you can miss it).

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
```diff
@@ -95,6 +95,61 @@ def get_user(self, user_id):
         return response
 
 
+def fetch_incidents(client: MsGraphClient, fetch_time: str, fetch_limit: int, filter_param: str, providers_param: str) \
```
yuvalbenshalom, line 98: we should consider calling fetch incidents in test module(without calling last run) when fetch incidents is marked. this wy you will see the errors in the test if there are problems with the fetch configs.

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.yml
```diff
@@ -21,20 +21,54 @@ configuration:
...
+  display: How many incidents to fetch each time
+  hidden: false
+  name: fetch_limit
+  required: false
+  type: 0
+- display: Fetch incidents of the given providers only. Multiple providers can be
+    inserted, seperated by a comma, e.g. "{first_provider},{second_provioder}". If
+    empty, incidents of all providers will be fetched.
+  hidden: false
+  name: fetch_providers
```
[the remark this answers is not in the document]
darkushin: If the providers are not valid providers then we won't get any incidents returned

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
darkushin: The title isn't unique, but there is an unique ID for every alert. I updated the name of the incident to be 'title + id', for example: 'test alert - da637218501473413212_-1554891308'.
What do you mean by adding the integration name?

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
darkushin: I considered it again after our talk. 'now' is a variable that is defined to the current time **before** I call the search_alerts command. The 'time_to' variable of the search alerts is defined according to 'now', so updating the new_last_run according to 'now' is actually the same as updating it with 'time_to', which is the last possible incident that was fetched. Do you still see an option for race condition and missing an alert?

Packs/MicrosoftGraphSecurity/Integrations/MicrosoftGraphSecurity/MicrosoftGraphSecurity.py
darkushin: Done. When trying to fetch incidents with an invalid filter format an error is returned as we wanted. Currently I just show the error returned by MS Graph (as was done in other parts in the integration as well), i.e. an error 400 is displayed. Do you want me to handle the error and display a different message or is it informative enough?
image

Packs/MicrosoftGraphSecurity/ReleaseNotes/2_0_0.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+- __Microsoft Graph__
+Added fetch-incidents function and unit test
```
amshamah419, line 4:
```suggestion
  - Added fetch-incidents function and unit test
```

kirbles19:
@darkushin Done.

PR #6839, opened 2020-05-07
Added delimiter argument

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#24185

## Description
Since we can configure Demisto lists to have other delimiters besides "," we need to add that option to Filterbylist script

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


ronykoz:
@hod-alpert Server has added the option to add a delimiter to a list, please check if we get the delimiter when we get the list, and if so please take that into consideration when splitting it

hod-alpert:
Checked, no delimiter in the list response

kirbles19:
@hod-alpert Done.

PR #6848, opened 2020-05-07
Common server python  - appendContext bug

reutshal:
## Status
- [] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24177

## Description
Fixed the method logic and added unit tests.

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
-        strBased = isinstance(data, STRING_TYPES) and isinstance(existing, STRING_TYPES)
-        if strBased:
-            data = data.split(',')
-            existing = existing.split(',')
-        newVal = data + existing
+        if isinstance(existing, STRING_TYPES):
+            if isinstance(data, STRING_TYPES):
+                new_val = data + ',' + existing
+            else:
+                return_error("Cannot append data to the existing context - \n The data is of instance {} while the "
```
ronykoz, line 1165:
```suggestion
                return_error("Cannot append data to the existing context - \n The data is of type {} while the "
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
+                             "context in the specified path is of instance {}.".format(type(data), type(existing)))
+
+        if isinstance(existing, dict):
+            if isinstance(data, dict):
+                existing.update(data)
+                new_val = existing
+            else:
+                return_error("Cannot append data to the existing context - \n The data is of instance {} while the "
```
ronykoz, line 1173:
```suggestion
                return_error("Cannot append data to the existing context - \n The data is of type {} while the "
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
+            else:
+                existing.append(data)
+            new_val = existing
+
         if dedup:
-            newVal = list(set(newVal))
-        if strBased:
-            newVal = ','.join(newVal)
-        demisto.setContext(key, newVal)
+            new_val = list(set(new_val))
```
ronykoz, line 1184: not sure how it behaves on a dict

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -1601,3 +1602,73 @@ def test_traceback_in_return_error_debug_mode_off(mocker):
...
+def get_set_context(key, val):
+    from CommonServerPython import return_error
+    return_error("key = {}, val = {}".format(key, val))
+
+
+@pytest.mark.parametrize('context_mock, data_mock, key, expected_answer', APPEND_CONTEXT_INPUT)
+def test_appendContext(mocker, context_mock, data_mock, key, expected_answer):
+    from CommonServerPython import demisto
+    mocker.patch.object(demisto, 'get', return_value=context_mock.get(key))
+    mocker.patch.object(demisto, 'setContext', side_effect=get_set_context)
```
ronykoz, line 1666: why not a function that just returns the key and value

ronykoz:
@anara123 could you make sure we are not missing some use case here?

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
reutshal: appendContext method does not return anything so I had to find some other way to validate this one

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
reutshal: Looks like it's just taking the keys.
image

The problem is on a list like this:
image
Not sure what exactly this `dedup` should do

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
-        strBased = isinstance(data, STRING_TYPES) and isinstance(existing, STRING_TYPES)
-        if strBased:
-            data = data.split(',')
-            existing = existing.split(',')
-        newVal = data + existing
+        if isinstance(existing, STRING_TYPES):
+            if isinstance(data, STRING_TYPES):
+                new_val = data + ',' + existing
+            else:
+                return_error("Cannot append data to the existing context - \n The data is of type {} while the "
```
anara123, line 1165: Don't use return_error in helper functions. Instead you can just try to add strings, it will raise `TypeError: can only concatenate list (not "str") to list` which is self explanatory

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
anara123: dedup should remove duplicates from the list.
if you have the same string more than once, it should keep only one string

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
+                             "context in the specified path is of instance {}.".format(type(data), type(existing)))
+
+        if isinstance(existing, dict):
+            if isinstance(data, dict):
+                existing.update(data)
+                new_val = existing
+            else:
+                return_error("Cannot append data to the existing context - \n The data is of type {} while the "
```
anara123, line 1173: same here

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
+        if isinstance(existing, STRING_TYPES):
+            if isinstance(data, STRING_TYPES):
+                new_val = data + ',' + existing
+            else:
+                return_error("Cannot append data to the existing context - \n The data is of type {} while the "
+                             "context in the specified path is of instance {}.".format(type(data), type(existing)))
+
+        if isinstance(existing, dict):
+            if isinstance(data, dict):
+                existing.update(data)
```
anara123, line 1170: I think this is not the expected behaviour.
If we have a 2 dictionaries, then I would expect to have list of 2 dictionaries in it.

anara123:
```
'a' + 'b' = 'a,b'
['a'] + 'b' = ['a','b']
['a']+['b'] = ['a','b']
{'a':1} + {'a':2} = [{'a':1},{'a':2}]
[{'a':1}] + {'a':2} = [{'a':1},{'a':2}]
1 + 2 = [1,2]
[1] + 2 = [1,2]
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
reutshal: Changed to a list of two dicts

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
reutshal: Done

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
reutshal: Done

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1156,17 +1156,34 @@ def appendContext(key, data, dedup=False):
...
-            existing = existing.split(',')
-        newVal = data + existing
-        if dedup:
-            newVal = list(set(newVal))
-        if strBased:
-            newVal = ','.join(newVal)
-        demisto.setContext(key, newVal)
+        if isinstance(existing, STRING_TYPES):
+            if isinstance(data, STRING_TYPES):
+                new_val = data + ',' + existing
```
ronykoz, line 1163: let's create a list of values here

reutshal: Are you sure? 
image
It's breaking the original behavior

ronykoz: if it's breaking then no need

reutshal:
@roysagi Can you please go over the CHANGELOG file? https://github.com/demisto/content/blob/CommonServerPython-appendContext_bug/Packs/Base/Scripts/CommonServerPython/CHANGELOG.md
Thanks!

PR #6863, opened 2020-05-08
[Marketplace] -  index update and images encode fix

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24075

## Description
- Integration images url with special character fix (added url encoding).
- Fixed inconsistency in updating index. In case that pack had initial `1.0.0` version and no release Notes (no ReleaseNotes folder), `changelog.json` was deleted from every second build (one was creating `changelog.json` properly and second build deleted it). 
Workaround for that is to copy `changelog.json` file from index to pack folder in such case (without modifying).


IkaDemisto:
@amshamah419 please take a look, instead of fixing this edge case in `update_index_folder`, just copied `changelog.json` in line 637 (it only happens in `1.0.0` version).

PR #6884, opened 2020-05-10
[Marketplace] - fixed changelog.json index update

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
Fixed index changelog.json update.


Tests/Marketplace/upload_packs.py
```diff
@@ -555,7 +555,6 @@ def main():
             continue
 
         pack.status = PackStatus.SUCCESS.name
-        pack.cleanup()
```
yuvalbenshalom, line 558: @IkaDemisto why do you remove the pack cleanup?

IkaDemisto: No need for it and because of it wrong uploaded version is printed in the summary.

PR #6901, opened 2020-05-10
Logrhythm - Added NTLM Support

guykeller:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23964

## Description
LogRhythm did not support windows authentication until now,  and such authentication would cause errors.
That problem is now fixed.

   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/LogRhythm/Integrations/integration-LogRhythm.yml
```diff
@@ -354,21 +382,45 @@ script:
     };
 
     var sendRequest = function(envelope, soapAction, apiUrl) {
+        username = params.Credentials.identifier.split('\\')[1]
```
ronykoz, line 385: why do we need the `\` ?

Packs/LogRhythm/Integrations/integration-LogRhythm.yml
```diff
@@ -47,6 +47,16 @@ configuration:
   name: incidentType
   type: 13
   required: false
+- display: Use Windows Authentication
+  name: use_ntlm
+  defaultvalue: "false"
+  type: 8
+  required: false
+- display: Domain of windows authentication user (used for windows authentication)
```
ronykoz, line 55: Shouldn't it be 
```suggestion
- display: Domain of windows authentication server (used for windows authentication)
```

Packs/LogRhythm/Integrations/integration-LogRhythm.yml
guykeller: Because that's how we are expected to send it when using NTLM for some reason

kirbles19:
@guykeller Done.

PR #6929, opened 2020-05-11
Check proxy and 'fields' argument in submit hec event

ag3ntsc4rn:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
The submit-hec command did not check for proxy before the request. This PR adds that check.
The fields argument was not being used as prescribed by Splunk. Fields should be a json payload. PR handles that.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -13,6 +13,7 @@ import requests
 import urllib3
 import io
 import re
+import time
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 # Define utf8 as default encoding
@@ -529,15 +530,18 @@ def splunk_submit_event_hec(hec_token, baseurl, event, fields, host, index, sour
     if hec_token is None:
         raise Exception('The HEC Token was not provided')
 
-    args = assign_params(
-        event=event,
-        host=host,
-        fields={'fields': fields} if fields else None,
-        index=index,
-        sourcetype=source_type,
-        source=source,
-        time=time_
-    )
+    args = {}
+    args["event"] = event
+    args["host"] = host
+    if fields:
+        args["fields"] = json.loads(fields)
+    args["index"] = index
+    args["sourcetype"] = source_type
+    args["source"] = source
+    if time_:
+        args["time"] = time_
+    else:
+        args["time"] = round(time.time())
 
     headers = {
         'Authorization': 'Splunk {}'.format(hec_token),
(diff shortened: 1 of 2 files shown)


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -549,8 +553,12 @@ def splunk_submit_event_hec(hec_token, baseurl, event, fields, host, index, sour
     return response
 
 
-def splunk_submit_event_hec_command():
-
+def splunk_submit_event_hec_command(proxy):
+    if not proxy:
```
guykeller, line 557: Is there a reason not to use `handle_proxy` here?

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -529,15 +530,18 @@ def splunk_submit_event_hec(hec_token, baseurl, event, fields, host, index, sour
...
+    args["host"] = host
+    if fields:
+        args["fields"] = json.loads(fields)
+    args["index"] = index
+    args["sourcetype"] = source_type
+    args["source"] = source
+    if time_:
+        args["time"] = time_
+    else:
+        args["time"] = round(time.time())
```
guykeller, line 544: please explain the goal of this change

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
ag3ntsc4rn: Thank you for the review comment. We can use handle_proxy. I was not aware of this common method mainly because the original author has not used it too. If you look at splunk_edit_notable_event_command, it did not use it hence i did not.  Please let me know if I should make the change.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
ag3ntsc4rn: Again, thank you for this too. We can use assign_params, but the way the fields param is parsed, is not correct according to what Splunk needs. Fields is supposed to contain key value pairs that is not part of the event.
Splunk reference: https://docs.splunk.com/Documentation/Splunk/8.0.3/Data/IFXandHEC
Snippet from that page:

> Assign the fields property to a JSON object that contains the custom fields to be indexed, as key-value pairs. For example, the following fields property, from within an HTTP request sent to the Splunk server, specifies two custom fields—club and wins:

So we should have something like this:
`args = assign_params(
        event=event,
        host=host,
        fields=json.loads(fields) if fields else None,
        index=index,
        sourcetype=source_type,
        source=source,
        time=time_
    )`

ag3ntsc4rn:
I have updated the code.

guykeller:
@ag3ntsc4rn 
Due to internal changes, your PR had to be closed.
Please delete the current fork, create a new one and add your changes there.
Apologies for the inconvenience

PR #6932, opened 2020-05-12
PAN-OS running ***panorama-get-url-category*** multiple times present…

bakatzir:
…ed in the war room data from the context.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24386

## Screenshots
in the issue

## Minimum version of Demisto
- [x] 3.5.0

   - [x] No

## Must have
- [x] Tests - no, as we do not have the urlfiltering license in all of our instances.
- [x] Documentation


Packs/Legacy/Integrations/Panorama/Panorama.py
```diff
@@ -2228,27 +2228,38 @@ def panorama_get_url_category_command(url_cmd: str):
     urls = argToList(demisto.args()['url'])
 
     categories_dict: Dict[str, list] = {}
+    categories_dict_hr: Dict[str, list] = {}
```
ronykoz, line 2231: not sure why do we need it if we have categories_dict

bakatzir: I want to have a separate one for the HR, as `categories_dict` is used to populate the data from the context, so we can add new urls to already pre defined categories in the context.

bakatzir:
Last build passed, files changed location as part of marketplace.

bakatzir:
conflicts with changelog due to release, merging.

PR #6933, opened 2020-05-12
Fixed ipv4Regex to not catch cidr regexs

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#24389

## Description
ipv4Regex in CommonServerPython catches also cidrs and not only ips
the fix allows only the string's end or any character that's not '/'

   - [x] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Tests
- [x] Documentation


kirbles19:
@hod-alpert Done.

PR #6935, opened 2020-05-12
Added IPv4 and CIDR indicators to the feed

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#24293

## Description
catching ip/cidrs indicators that are just in the same place as the valid domains that are fetched.
For example
The cidrs that are in the same line as microsoft.com here
image

## Note
Until [this PR](https://github.com/demisto/content/pull/6933) will be merged- ip indicators will have cidr indicators too


Itay4:
add unit test and check one indicator of each type

Packs/FeedMicrosoftIntune/Integrations/FeedMicrosoftIntune/FeedMicrosoftIntune.py
```diff
@@ -33,40 +33,40 @@ def build_iterator(self) -> List:
...
-        except ValueError as err:
-            demisto.debug(str(err))
-            raise ValueError(f'Could not parse returned data to Json. \n\nError massage: {err}')
+        table_rows = soup.select("tbody tr")
+        for row in table_rows:
+            found_domains = [string.strip() for string in row.strings if re.search(r'microsoft\.(com|net)', string)]
+            if found_domains:
+                domains += found_domains
+                ipv4s += [string.strip() for string in row.strings if re.match(ipv4Regex, string)]
+                ipv4cidrs += [string.strip() for string in row.strings if re.match(ipv4cidrRegex, string)]
```
avidan-H, line 50: Instead of using a list comprehension for each of these and essentially iterating over the contents twice, you can do one for loop for both of them. Then, if you check for `ipv4cidrRegex` first, you don't need to wait on the other PR you mentioned.

hod-alpert:
The description is very general and the change is not relevant to it unless I'll rewrite the whole thing, and the README is empty.
I'll leave it to @kirbles19 to decide

kirbles19:
@hod-alpert Done. I reviewed the RN.

PR #6939, opened 2020-05-12
Add ArcSight Logger images for docs

barchen1:
## Status
- [x] Ready

## Description
Added Arcsight logger images to docs/images.


ronykoz:
What about changing the image in the readme?

PR #6948, opened 2020-05-12
Created phishing pack

idovandijk:
## Status
- [x] In Progress - **testing on 5.5**
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24429

## Description
Moved all mandatory dependencies of phishing into one pack (the rest are optional, AKA not required by default or associated to other incident types as well)

No

## Must have
- [ ] Tests
- [x] Documentation


idovandijk:
build will fail until https://github.com/demisto/etc/issues/24451 is fixed

PR #6956, opened 2020-05-12
upload id-set to GCS

anara123:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related https://github.com/demisto/etc/issues/24216

## Description
Upload id_set.json to GCS. Making it publicly available.

### DIFF
--- a/Tests/Marketplace/upload_packs.py
+++ b/Tests/Marketplace/upload_packs.py
@@ -1,5 +1,6 @@
 import json
 import os
+import io
 import sys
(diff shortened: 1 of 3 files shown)


Tests/Marketplace/upload_packs.py
```diff
@@ -1,5 +1,6 @@
 import json
 import os
+import io
```
IkaDemisto, line 3: Don't see that it is used

PR #6958, opened 2020-05-13
PAN-OS - fix ***panorama-get-applications***

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24384

## Description
checked both on pan-os fw and panorama

## Screenshot
in the referenced issue

## Minimum version of Demisto
- [x] 3.0.0

   - [x] No - but I removed the key `Id` since pan-os. does not return data for it. kept it in the yml.

## Must have
- [x] Tests - yes. added to TPB.


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -3142,40 +3142,42 @@ def panorama_get_pcap_command():
...
 def panorama_list_applications():
     params = {
-        'type': 'op',
-        'command': '<show><objects></objects></show>',
+        'type': 'config',
+        'action': 'get',
+        'xpath': XPATH_OBJECTS + "application/entry",
         'key': API_KEY
     }
+    demisto.log(params['xpath'])
```
ronykoz, line 3167: let's remove this

bakatzir: I'm re reviewing this issue, will update today this as well and request review again.

PR #6966, opened 2020-05-13
FeedJSON handling flattened list inside a JSON object

teizenman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24461

## Description
Added handling for flattened list of indicators inside a JSON object.
Before it was treated as json objects that includes the value, now it also support a list of text indicators one after another. (you can see an example in the test)

## Must have
- [x] Tests
- [x] Documentation


roysagi:
@teizenman - Docs seems good to me

PR #6968, opened 2020-05-13
Making Content Great Again batch 5

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

##related issue: 
https://github.com/demisto/etc/issues/23712

## Description
fixing content files so that they pass validation- batch 5


hod-alpert:
The build fails for a lot of `You've added fromversion to an existing file in the system, this is not allowed, please undo` on files that are failing on `missing fromversion` error message

Itay4:
need to check those incident fields - lets discuss

PR #6973, opened 2020-05-13
Legacy scripts to DeprecatedContent - Part 3

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
Moved deprecated scripts from Legacy to DeprecatedContent


Packs/DeprecatedContent/Scripts/script-VMRayGetResults.yml
```diff
@@ -1,5 +1,5 @@
 commonfields:
-  id: 9364c36f-b1d6-4233-88c2-75008b106c31
```
yuvalbenshalom, line 2: not sure if its a good idea to change scripts ids.
let me check first.

yuvalbenshalom: @dantavori please change this one back to the original and I think we can merge

PR #6980, opened 2020-05-14
SNOW v2 - Added the *incident_name* parameter, to allow the user to c…

bakatzir:
…hoose the column from ServiceNow on which the fetched incidents will be named.

## Status
- [x] Ready
## Related Issues
fixes: https://github.com/demisto/etc/issues/17727

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


liorblob:
@yuvalbenshalom yes it can be done, it's already been done with the fetch timestamp parameter.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -47,6 +47,11 @@ configuration:
   name: timestamp_field
   required: false
   type: 0
+- display: ServiceNow ticket column to be set as the incident name
```
liorblob, line 50: add that the incident number is the default

PR #7051, opened 2020-05-18
SplunkPy - handle partial response returned from Splunk

Itay4:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/24269

## Description
`IncompleteRead` is raised by `httplib` and `splunk-sdk` does not handle it approritaly

when it is raised, a partial of the xml is returned in the exception object, which can be parsed and the indexes can be extracted from it

the xml is not well formatted, so need to iterate its elements until a `PraseError` is raised

   - [x] No

## Must have
- [x] Tests

see failing unit test at https://circleci.com/gh/demisto/content/53301#queue-placeholder/containers/1


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -502,12 +504,32 @@ def parse_time_to_minutes():
...
+    try:
+        for index in indexes:
+            index_json = {'name': index.name, 'count': index["totalEventCount"]}
+            indexes_names.append(index_json)
+    except IncompleteRead as e:
+        partial_xml_file = NamedTemporaryFile(delete=False)
+        partial_xml_file.write(str(e.partial))
+        partial_xml_file.close()
+        try:
+            tree = ET.iterparse(partial_xml_file.name)
```
anara123, line 517: import ET in the code , it is not clear where ET come from

Itay4:
closing this as we found better solution for this
will open a new PR

PR #7058, opened 2020-05-18
Move SendEmailToManager and StaticAnalyze

liorblob:
## Status
- [ ] In Progress
- [X] Ready

Moved SendEmailToManager to AD pack and changed the `AdSearch` command to `ad-search`.
Moved `StaticAnalyze` to D2 pack as it uses `D2PEdump`.


liorblob:
@yuvalbenshalom nothing depends on this script

yuvalbenshalom:
@anara123 @michalgold  WDYT?
should we open a separate pack for this or put it on AD pack?

anara123:
Move to AD pack

IkaDemisto:
`AdSearch` deprecated script of deprecated pack is used in this script.

liorblob:
@IkaDemisto is that a problem?

IkaDemisto:
Yes, that is why I moved this script temporary to `Legacy` pack.

liorblob:
Until when?

IkaDemisto:
Until it won't depend on script from `DeprecatedContent` pack. It's dependencies are ignored and new packs should not depend on this pack.

liorblob:
it's an old pack though, should we change the dependency?

IkaDemisto:
Yes, I think so.

liorblob:
@IkaDemisto done

PR #7061, opened 2020-05-18
SplunkPy - correct the usage of fields for hec submit event command

ag3ntsc4rn:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
For Splunk HEC submissions using the submit-event-hec command, fields param used to take in a comma separated string. So when you used the command, in splunk a field by the name "fields" will be created and its value will be the comma separated string. This is not the recommended usage of fields according to Splunk. Fields should actually be a JSON key value pair where you can create new fields (with the name provided in the param and not hardcoded 'fields') and values in Splunk. This PR addresses that issue.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ X] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -528,11 +528,18 @@ def splunk_submit_event_hec(hec_token, baseurl, event, fields, host, index, sour
 
     if hec_token is None:
         raise Exception('The HEC Token was not provided')
+    
+    parsed_fields = None
+    if fields:
+        try:
+            parsed_fields = json.loads(fields)
+        except:
+            parsed_fields = {'fields': fields}
 
     args = assign_params(
         event=event,
         host=host,
-        fields=json.loads(fields) if fields else None,
+        fields=parsed_fields,
         index=index,
         sourcetype=source_type,
         source=source,
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -532,7 +532,7 @@ def splunk_submit_event_hec(hec_token, baseurl, event, fields, host, index, sour
     args = assign_params(
         event=event,
         host=host,
-        fields={'fields': fields} if fields else None,
+        fields=json.loads(fields) if fields else None,
         index=index,
         sourcetype=source_type,
         source=source,
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
@@ -318,7 +318,7 @@ script:
       secret: false
     - default: false
       description: Fields for indexing that do not occur in the event payload itself.
-        Accepts json string of key value pairs.
+        Accepts json string of key value pairs or comma seperated values.
       isArray: false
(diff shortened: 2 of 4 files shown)


guykeller:
@ag3ntsc4rn 
This seems to me like a backwards compatibility break, and quite a major one since this is a mandatory argument, unless you are actually asking the user for the same syntax - in which case I don't understand the change.

Are you 100% sure the current syntax does not work?

@Shellyber Any input on this? since you wrote the command and this is a major change, thought you should be aware

ag3ntsc4rn:
@guykeller - So the current syntax works. It just does not work the way it is intended to by Splunk. In current syntax, when the logs are sent to Splunk, it creates one field called 'fields' with one or more comma separated values.

What I am suggesting is actually what Splunk recommends. You should pass a json payload to fields argument. The keys in the payload will be the fields in Splunk. Example.
`{"mykey":"myvalue"}`. This will create a field called mykey with a value of myvalue in Splunk and you can search in Splunk using.
`index=foo sourcetype=bar mykey::myvalue` (If mykey is not an indexed extraction)
`index=foo sourcetype=bar mykey=myvalue` (If mykey is an indexed field)

Having said that, I agree this is a big change for users who are not aware of this usage of fields. and code will break in json.fields. Maybe a json parse can be added to the logic. That way, if you send a valid json string, we can use the logic described here, if not a valid json, well continue with what we have. Open to suggestions.

guykeller:
@ag3ntsc4rn 
Thank you for the detailed explanation!
I agree with you - this is quite a big change.
Let's do it as you suggested - if the string isn't a valid json, use the old way. Otherwise, use your suggested way.

guykeller:
@ag3ntsc4rn 
Your current problem is that you don't have any release notes for the change you have introduced here. Let's add those and merge the change

ag3ntsc4rn:
Updated code as discussed via commit `ec6fbfe`
Not sure about how and where to update release notes. Need some guidance as this is my first ever contribution.

guykeller:
@ag3ntsc4rn 
Let's schedule a call so I would help you through it.
Please let me know which platform would you like to schedule on, Slack or email.

guykeller:
@ag3ntsc4rn Regarding the call I suggested - would love to schedule it.
Which platform would suit you best for scheduling it?

ag3ntsc4rn:
Same here. Submitted RN but CLA is unhappy.

anara123:
@ag3ntsc4rn who is `Manish Menon`?
 it seems that you signed

ag3ntsc4rn:
That’s my name actually. Ugh. I might have some time in the past switched to hiding my email for privacy reasons.

anara123:
@ag3ntsc4rn then you are good to go. You can ignore the CLA warning. It seems that you signed it

guykeller:
@ag3ntsc4rn 
As Anar said, we can ignore the CLA but it seems you still have some flake8 errors.
I re-ran your build but they seem straight forward enough to be fixed easily - would you like my help with that?

ag3ntsc4rn:
@guykeller - I may not be able to get to it over the weekend. If you have time, I don't mind you fixing it for me.

guykeller:
@ronykoz Could we do what we discussed here as well please?

PR #7074, opened 2020-05-19
updated kenna's Test playbook to single threaded.

orhovy:
## Status
- [X] Ready

## Related Issues
fixes:  https://github.com/demisto/etc/issues/21233

## Description
Changed Test playbook to single stream.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?


Packs/Kenna/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1 @@
+Fixed testing for the Kenna pack.
```
bakatzir, line 1: we changed only the test playbook, not sure we want rn. @amshamah419 how do we do that? `-`?

amshamah419: Right now RN will fail when they are empty. Let's use a simple description. I'm hoping to have this issue fixed in the future.

PR #7094, opened 2020-05-19
move validate to a script

ronykoz:
## Status
- [x] In Progress
- [x] Ready


Tests/scripts/validate.sh
```diff
@@ -0,0 +1,7 @@
+if [ "${CHECK_BACKWARD}" = "true" ] ;
```
yaakovi, line 1: need to forward the exit code from the sdk to the bash script (current state the script will always exit with 0
add `exit $?`

PR #7096, opened 2020-05-19
Add auto mapping script (used by the server)

erezh31:
- [X] Ready


Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.yml
```diff
@@ -0,0 +1,44 @@
+args:
```
lperryp, line 1: Perhaps add more initial values/list options to parameters?

erezh31: Not really needed, it's script that the server run internally, no one should run it manually, this is why is depracated.

kirbles19:
@erezh31 Done

PR #7110, opened 2020-05-20
Whois Queryresult is now determined by the "Domain Name:" string

teizenman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24594

### DIFF
--- a/Packs/Whois/Integrations/Whois/Whois.py
+++ b/Packs/Whois/Integrations/Whois/Whois.py
@@ -8277,7 +8277,7 @@ def get_domain_from_query(query):
 def create_outputs(whois_result, domain, query=None):
     md = {'Name': domain}
     ec = {'Name': domain,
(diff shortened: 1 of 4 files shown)


Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8277,7 +8277,7 @@ def get_domain_from_query(query):
 def create_outputs(whois_result, domain, query=None):
     md = {'Name': domain}
     ec = {'Name': domain,
-          'QueryResult': 'NOT FOUND' not in str(whois_result.get('raw', 'NOT FOUND'))}
+          'QueryResult': 'Domain Name:' in str(whois_result.get('raw', 'NOT FOUND'))}
```
ronykoz, line 8280: So it should be False if none found?

teizenman: Indeed

PR #7113, opened 2020-05-20
POST requests params should be passed through body and not uri

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#24535:

## Description
For some reason the server failed to return a proper response with status line, instead it sent a body without it.
That's why when the client tried to parse the first line of the body (the status line) it came across the first line of the following body:
```
<html>
<head><title>414 Request-URI Too Large</title></head>
<body bgcolor="white">
<center><h1>414 Request-URI Too Large</h1></center>
<hr><center></center>
</body>
</html>
```
And that's why the erroring status line was `BadStatusLine('<html>\r\n')`

Thing is when using a post request- the standard is to send the parameters through the body and not through the uri since the uri has a limit of =~ 2 kB which the body does not have.

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -480,7 +480,7 @@ def panorama_command():
     result = http_request(
         URL,
         'POST',
-        params=params
+        body=params
     )
 
     demisto.results({
@@ -502,7 +502,7 @@ def panorama_commit():
     result = http_request(
         URL,
         'POST',
-        params=params
+        body=params
     )
 
     return result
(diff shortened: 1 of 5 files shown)


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -2087,9 +2087,9 @@ def panorama_edit_custom_url_category(custom_url_category_name, type_, items, de
         'key': API_KEY
     }
     result = http_request(
-        URL,
+        'https://content.demisto.works:56111/api/',
```
Itay4, line 2090: revert this

hod-alpert: Oh, right

kirbles19:
@hod-alpert Done.

PR #7115, opened 2020-05-20
SplunkPy - add Python requests handler

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24269

## Description
Added Python requests handler, which handles HTTP requests better than `httplib.HTTPSConnection` which Splunk SDK uses.

   - [x] No

## Must have
- [x] Documentation


Itay4:
@glicht please have a look at my latest commits
added support for proxy using requests handler

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
```diff
@@ -101,6 +101,10 @@ configuration:
   name: fetch_time
   required: false
   type: 0
+- display: Use Python requests handler
+  name: use_requests_handler
+  required: false
```
anara123, line 106: don't we want it to be the requests handler by default?

kirbles19:
@Itay4 Done.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
Itay4: the splunk sdk is pretty fragile in regards to connection, and im concerned making this by default might break bc
i suggest we gain some experience with it and then maybe in set it to be by default

PR #7124, opened 2020-05-20
Make conf json redundant

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24412

## Description
Building conf.json in build time instead of updating it manually - still allowing to use it.


ronykoz:
I will add UT later, just want to make sure we are ok with this approach first.

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
+import os
+import yaml
+import json
+from distutils.version import LooseVersion
+
+PACKS_DIR = './Packs'
+CONF_JSON_PATH = './Tests/conf.json'
+INTEGRATIONS_DIR = 'Integrations'
+TEST_PLAYBOOKS_DIR = 'TestPlaybooks'
```
yaakovi, line 9: why not sdk constants?

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+    for pack_name in os.listdir(PACKS_DIR):
+        pack_path = os.path.join(PACKS_DIR, pack_name)
+
+        pack_integrations = []
+        pack_test_playbooks = []
+        for dir_name in os.listdir(pack_path):
+            dir_path = os.path.join(pack_path, dir_name)
+
+            if dir_name not in [INTEGRATIONS_DIR, TEST_PLAYBOOKS_DIR]:
+                continue
```
yaakovi, line 82: why use `listdir` if you only refer to those two directories?

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+
+            for file_or_dir in os.listdir(dir_path):
+                if os.path.isdir(os.path.join(dir_path, file_or_dir)) and dir_name == INTEGRATIONS_DIR:
+                    inner_dir_path = os.path.join(dir_path, file_or_dir)
+                    for integration_file in os.listdir(inner_dir_path):
+                        is_yml_file = integration_file.endswith('.yml')
+                        file_path = os.path.join(inner_dir_path, integration_file)
+                        if is_yml_file:
+                            if dir_name == INTEGRATIONS_DIR:
```
yaakovi, line 91: redundant due to the condition on line 85

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+                                pack_integrations.append(get_integration_data(file_path))
+                else:
+                    is_yml_file = file_or_dir.endswith('.yml')
+                    file_path = os.path.join(dir_path, file_or_dir)
+                    if is_yml_file:
+                        if dir_name == INTEGRATIONS_DIR:
+                            pack_integrations.append(get_integration_data(file_path))
+
+                        if dir_name == TEST_PLAYBOOKS_DIR:
```
yaakovi, line 100: make sure it's not a test script

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+
+    return max_fromversion
+
+
+def calc_conf_json_object(integrations, test_playbooks):
+    conf_objects = [{'playbookID': test_playbook} for test_playbook in test_playbooks]
+    fromverion = get_fromversion(integrations)
+    integrations = [integration_id for integration_id, _ in integrations]
+    if len(integrations) == 1:
+        integrations = integrations[0]
```
yaakovi, line 38: list of one should also work btw

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+    for integration_id, fromversion in integrations:
+        if LooseVersion(fromversion) > LooseVersion(max_fromversion):
+            max_fromversion = fromversion
+
+    return max_fromversion
+
+
+def calc_conf_json_object(integrations, test_playbooks):
+    conf_objects = [{'playbookID': test_playbook} for test_playbook in test_playbooks]
+    fromverion = get_fromversion(integrations)
```
yaakovi, line 35: I think we also have `fromversion` for the test playbook, not sure if still relevant (we had an integration we two test playbook that differs in a few tasks that appeared only for a specific version)

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
+import os
```
glicht, line 1: add the following at the start:
```
#!/usr/bin/env python3
```
and make the file executable. That way it is possible to call the file directly.

glicht:
Looks good. I like the idea. A few remarks:
* I think we should skip the dir: `Packs/DeprecatedContent/TestPlaybooks` and `Packs/NonSupported`.
* We can optimize the script to avoid parsing integration files if there are no test playbooks or the TestPlaybooks dir doesn't exist.
* Add print for each processed pack so it is easy to see that progress is being made
* Add a print of the total time it takes to generate so we can identify easily when this process gets slow and we need to further optimize (it seems that it takes around a minute currently)
* Copy conf.json to the artifacts dir as we do with id_set.json

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,112 @@
...
+        json.dump(conf, conf_file, indent=4)
+
+
+def load_test_data_from_conf_json():
+    test_playbooks = []
+    with open(CONF_JSON_PATH) as data_file:
+        conf = json.load(data_file)
+
+    for conf_test in conf['tests']:
+        test_playbooks.append(conf_test['playbookID'])
```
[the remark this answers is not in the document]
ronykoz: couldn't find it, where should i get it from?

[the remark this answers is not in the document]
anara123: Sorry, my bad. It was in `collect_tests_and_content_packs.py` in content repo.

Tests/scripts/update_conf_json.py
```diff
@@ -0,0 +1,127 @@
...
+            new_conf_json_objects.extend(calc_conf_json_object(pack_integrations, pack_test_playbooks))
+
+    add_to_conf_json(new_conf_json_objects)
+
+
+if __name__ == '__main__':
+    start_time = datetime.now()
+    run()
+    total_time = datetime.now() - start_time
+    print(f'Total time {total_time}')
```
glicht, line 127: Can we add here also total number of tests collected

ronykoz: @glicht Done

.circleci/config.yml
```diff
@@ -98,6 +98,7 @@ jobs:
                 exit 0
             fi
             demisto-sdk create-id-set -o ./Tests/id_set.json
+            python3 ./Tests/scripts/update_conf_json.py
```
glicht, line 101: Note that if `demisto-sdk create-id-set...` fails then this line will not be called.

ronykoz: If the id_set fails then the upload to marketplace/ configure test filter / configure packs to upload will all not work anyway
@glicht

glicht: @ronykoz  Thinking about the flow here.  I think we should do this step before the id_set step. So when id set runs it has all data and is able to include the relevant tests in its output.

glicht:
@ronykoz can you modify an integration, playbook and script to use the new setup so we can see that the whole flow works in the build.

ronykoz:
Added to the artifcats

ronykoz:
@glicht did you look at this?

glicht:
Looking good. Can you format that following json print:
```
Added the following objects to the conf.json:
[{"playbookID": "Test - Windows Defender Advanced Threat Protection", "fromversion": "4.5.0", "integrations": ["Windows Defender Advanced Threat Protection"]}, {"playbookID": "Cymulate Immediate Threats", "fromversion": "4.5.0", "integrations": ["Cymulate"]}, {"playbookID": "stripChars - Test", "fromversion": "4.5.0"}, {"playbookID": "GmailTest", "fromversion": "4.5.0", "integrations": ["Gmail"]}]
Total time 0:01:01.204081
```

ronykoz:
@glicht added the format + reverted the changes to the conf.json
Should be good to go

ronykoz:
@glicht can i merge once the build finish?

PR #7138, opened 2020-05-21
MongoDB - Processing datetime object

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24276

## Description
MongoDB returns datetime objects from the DB. so from now on, they will be converted to strings. Same as input, just put the date string and it'll automatically convert it to DateTime.

Added a step to both unitest and test playbook.

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


kirbles19:
@jochman Done.

PR #7141, opened 2020-05-21
Shodan - handle ip without info

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24776 https://github.com/demisto/etc/issues/23370

## Description
 - Return empty dict in case of 404 status code returned to handle no results found of an ip
 - Unskipped test

   - [x] No

## Must have
- [x] Tests
 
see test failing on error at https://circleci.com/gh/demisto/content/53863


Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
```diff
@@ -183,59 +184,62 @@ def ip_command():
...
-
-    ip_details = {
-        'ASN': res.get('asn', ''),
-        'Address': ip,
-        'Hostname': hostname,
-        'Geo': {
-            'Country': res.get('country_name', ''),
-            'Location': location
+    if not res:
+        demisto.results('No information available for that IP.')
```
ShahafBenYakir, line 188:
```suggestion
        demisto.results('No information available for the given IP.')
```

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
```diff
@@ -45,7 +45,8 @@ def http_request(method, uri, params=None, data=None, headers=None):
                            data=data,
                            headers=headers,
                            verify=USE_SSL)
-
+    if res.status_code == 404:
+        return {}
     if res.status_code != 200:
         error_msg = f'Error in API call {url} [{res.status_code}] - {res.reason}'
```
ShahafBenYakir, line 51: Why does this is not thrown?

Itay4: there isnt even main function here and dont want to change the whole integration

kirbles19:
@Itay4 Done.

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
```diff
@@ -20,12 +20,7 @@
 # Should we use SSL
 USE_SSL = not demisto.params().get('insecure', False)
 
-# Remove proxy if not set to true in params
-if not demisto.params().get('proxy'):
-    os.environ.pop('HTTP_PROXY', None)
-    os.environ.pop('HTTPS_PROXY', None)
-    os.environ.pop('http_proxy', None)
-    os.environ.pop('https_proxy', None)
+handle_proxy()
```
avidan-H, line 23: Are we sure that `requests.request` respects the proxy environment variables and that you don't need to explicitly pass the proxies dictionary returned by `handle_proxy` to the request created in `http_request`?

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
avidan-H: I think it probably makes more sense to `raise_for_status()` here and then use a `try-except` block below where if it's a 404 you can say no details were found instead of raising an error.

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
Itay4: yes, look at https://requests.readthedocs.io/en/master/user/advanced/#proxies

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
Itay4: i dont think it really matters at this point, it does the same thing

avidan-H: It moves all of the handlings of custom logic to the calling function instead of performing custom behavior in the function that makes the request itself.

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
avidan-H: I know. I read that prior to posting. But it seems odd that in our `BaseClient` we have to alter an instance variable to get the proxies working correctly but here we don't because `requests.get` etc. use a session under the hood... so it would seem to me that either our `BaseClient` or the integrations where we do `requests.(request/get/post/etc)` are doing something wrong or unnecessary.

Packs/Shodan/Integrations/Shodan_v2/Shodan_v2.py
Itay4: i prefer not to change this as `raise_for_status` accepts all status codes that are not in the range of 400-600, and some apis return 2xx and 3xx as incomplete results and stuff like that - we might break something here

PR #7143, opened 2020-05-21
Packs dependencies calculation - skip Legacy pack

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
Instead of skipping `Legacy` in sdk, we will cut `Legacy` subtree (if exists) in content dependencies creation of all packs. No need to release new sdk version and it is possible to rollback this changes in content.

Triggered build: https://circleci.com/gh/demisto/content/53909
(check `packs_dependencies.json` file)

### DIFF
--- a/Tests/Marketplace/packs_dependencies.py
+++ b/Tests/Marketplace/packs_dependencies.py
@@ -1,6 +1,7 @@
 import os
 import json
 import argparse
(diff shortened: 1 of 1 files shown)


IkaDemisto:
@ronykoz FYI

PR #7155, opened 2020-05-22
Microsoft Graph Mail Single User - set last run to be last fetched and not now time

Itay4:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/24502

## Description
In case no emails were fetched, now time was set as last fetch, and led to emails missed and not fetched

set the last run to be the last fetched

   - [x] No

## Must have
- [ ] Tests


kirbles19:
@Itay4 Done.

PR #7172, opened 2020-05-24
Update SNDBOX_Test Playbook

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24349

## Description
Removes `SNDBOX_Test` from the skip list and adds the `SNDBOX` integration to the unmockable integrations since it sends files for analysis which the mocking mechanism doesn't currently support. Update the test playbook so that it doesn't fail in two situations,
1. generic polling finished and the analysis is still pending - therefore the `download report` task does not execute and the playbook concludes successfully.
2. `SNDBOX` experienced an error in analyzing the file as in the attached screenshot, and therefore we do not execute the `download report` task and conclude the playbook successfully.

## Screenshots
image (1)

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No

### DIFF
--- a/Packs/SNDBOX/TestPlaybooks/playbook-SNDBOX_Test.yml
+++ b/Packs/SNDBOX/TestPlaybooks/playbook-SNDBOX_Test.yml
@@ -1,4 +1,3 @@
-elasticcommonfields: {}
 id: SNDBOX_Test
 version: -1
 name: SNDBOX_Test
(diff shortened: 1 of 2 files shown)


avidan-H:
The updated test playbook,
SNDBOX_Test_Sun_May_24_2020

Itay4:
build fails cause of a bug in the validation which is being handled
merging

PR #7175, opened 2020-05-24
ThreatMiner - handle IP address lookups without reports

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24817 https://github.com/demisto/etc/issues/20469

## Description
IP addresses with no reports, such as https://www.threatminer.org/host.php?q=35.24.216.211, were not handled appropriately.

   - [x] No

## Must have
- [x] Tests

see failing test at https://circleci.com/gh/demisto/content/54144
image


Packs/ThreatMiner/Integrations/ThreatMiner/ThreatMiner.py
```diff
@@ -0,0 +1,698 @@
...
+    return md
+
+
+def get_ip_report_from_threat_miner(ip_address):
+    return {
+        'raw_whois': get_ip_whois_rawdata(ip_address),
+        'raw_passive_dns': get_ip_passiveDNS_rawdata(ip_address),
+        'raw_ip_md5': get_ip_MD5_rawdata(ip_address),
+        'raw_ip_uris': get_ip_URI_rawdata(ip_address),
+        'raw_ip_ssl': get_ip_SSL_rawdata(ip_address)
```
avidan-H, line 331: The problem you fixed, returning an empty dictionary when there are no results returned from the request made instead of a list, does that problem also exist in all these other functions and not just `get_ip_whois_rawdata`? If so, shouldn't we fix it in all those other functions as well.

avidan-H: And what about the `file` and `domain` commands... Need to check if the same problem is present in those commands and if so fix it.

Itay4: actually the rest can be lists, they are handled differently than WHOIS results

for some reason `file` and `domain` are handled, they do return `{}` in case no results were found - the issue was only with the `ip` command

kirbles19:
@Itay4 Done.

PR #7177, opened 2020-05-24
Fixed a bug in query-logs command

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Description
When there were no records found for a given query in **!cdl-query-logs** command a `StopIteration` was raised and in the user side- there was no reason for the error at all.

So instead of getting table name from the records, table name will be taken from the query.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
+++ b/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
@@ -546,7 +546,7 @@ def query_logs_command(args: dict, client: Client) -> Tuple[str, Dict[str, List[
 
     records, raw_results = client.query_loggings(query)
 
-    table_name = get_table_name(records)
+    table_name = get_table_name(query)
     transformed_results = [common_context_transformer(record) for record in records]
     human_readable = tableToMarkdown('Logs ' + table_name + ' table', transformed_results, removeNull=True)
     ec = {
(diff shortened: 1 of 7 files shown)


Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
```diff
@@ -555,16 +555,20 @@ def query_logs_command(args: dict, client: Client) -> Tuple[str, Dict[str, List[
...
     Args:
-        records: Records to get the table name of
+        query: Query string, i.e SELECT * FROM firewall.threat LIMIT 1
 
     Returns:
-        The records table name
+        The query's table name
     """
-    return next(record.get('log_type', {}).get('value') for record in records if record.get('log_type')) or ''
+    FIND_TABLE_NAME_FROM_QUERY = r'(FROM `)(\w+.\w+)(`)'
```
Itay4, line 567: global consts should be at the top of the code

Packs/CortexDataLake/Integrations/CortexDataLake/README.md
```diff
@@ -776,7 +776,8 @@ Searches the Cortex panw.threat table, which is the threat logs table for PAN-OS
 | File.Type | String | The file type, as determined by libmagic (same as displayed in file entries). | 
 
 
-##### Command Example
+##### Command Examples
+```!cdl-query-threat-logs query="is_packet_capture = true AND severity = \"Critical\"" fields=pcap limit=10```
```
Itay4, line 780: lets remove backslashes

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
hod-alpert: But I only use it once in this method's scope 
Should I lowercase it?
or just use it as value in re.search?
Seems to much to make a const of something that's being used only once

Packs/CortexDataLake/Integrations/CortexDataLake/README.md
hod-alpert: Are you sure? It won't work without the backslashes..

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
Itay4: yea so lets lower case it

hod-alpert: Done

kirbles19:
@hod-alpert Done.

PR #7182, opened 2020-05-24
Content packs collection - init packs_to_install_string with empty str

Itay4:


Itay4:
build fails on non related test
merging

PR #7186, opened 2020-05-24
Checkpoint Firewall - add command arg to checkpoint cmd

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24636

## Description
Somehow the argument is required in the code but does not even exist in the yml, so it doesnt really break bc


kirbles19:
@Itay4 Done.

PR #7192, opened 2020-05-25
Snow readme fix

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24858

   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/ServiceNow/ReleaseNotes/1_1_1.md
+++ b/Packs/ServiceNow/ReleaseNotes/1_1_1.md
@@ -1,6 +1,4 @@
 
 #### Integrations
-- __README.md__
-Fixed the timestamp field parameter description.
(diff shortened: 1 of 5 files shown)


Packs/ServiceNow/ReleaseNotes/1_1_1.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+- __README.md__
```
glicht, line 3: We shouldn't have this entry. If sdk is enforcing this open a bug.

bakatzir: It does not enforce it. changed.

PR #7210, opened 2020-05-25
Fixing adding indicators to the context data

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24849

## Description
Indicator has no ID attribute, the actual value comes in `Value` field which should be the unique field.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Documentation

### DIFF
--- a/Packs/ThreatQ/.secrets-ignore
+++ b/Packs/ThreatQ/.secrets-ignore
@@ -14,3 +14,4 @@ https://raw.githubusercontent.com
 https://www.paloaltonetworks.com
 91.140.64.113
 goo@test.com
+10.50.0.1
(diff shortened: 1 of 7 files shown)


dantavori:
what if you have two indicators, say a URL with the value `https://www.google.com` and a string with the same value? @Arsenikr can you confirm/deny this case is extreme?

Arsenikr:
> what if you have two indicators, say a URL with the value `https://www.google.com` and a string with the same value? @Arsenikr can you confirm/deny this case is extreme?

if the value is the same it is the same indicator i guess.. am i missing something here?

hod-alpert:
He meant if there are two indicators with the same value but with different types
Anyways we agreed that there is no issue here since id the indicator has an ID it will be added to the context, and if it does not- then it's probably the same indicator.

kirbles19:
@hod-alpert @Itay4 Done.

PR #7214, opened 2020-05-25
Hello world fetch incidents addition

reutshal:
## Status
- [] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24619

## Description
* Added a time comparison in line 790.
* Changed the last-fetch time in the unit-test to a valid time (and not later than the incidents creation time).
   - [ ] Yes
       - Further details:
   - [X] No


Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.py
```diff
@@ -786,45 +786,48 @@ def fetch_incidents(client: Client, max_results: int, last_run: Dict[str, int],
...
-            # 'type': 'Hello World Alert',  # Map to a specific XSOAR incident Type
-            'severity': convert_to_demisto_severity(alert.get('severity', 'Low')),
-            # 'CustomFields': {  # Map specific XSOAR Custom Fields
-            #     'helloworldid': alert.get('alert_id'),
-            #     'helloworldstatus': alert.get('alert_status'),
-            #     'helloworldtype': alert.get('alert_type')
-            # }
-        }
+        # to prevent duplicates
+        if incident_created_time > last_fetch:  # type: ignore
```
fvigo, line 790: - Can you do the other way around to avoid unnecessary indentation? i.e.:
``` 
if incident_created_time <= last_fetch:
    continue
```
- We shouldn't use `  # type: ignore` unless really necessary: what mypy error are you getting?
- Can you add a comment/INTEGRAITON DEVELOPER TIP to explain the logic?

reutshal:
@francoisihry Nice thoughts.
We should also keep in mind that without the addition, as long as there aren't new incidents from the API, the last incident will be duplicated over and over again.
I'm not sure how common it is to have two incidents with the exact same time.

Another solution will be to keep list of incidents IDs from the last fetch and add only the incidents with the IDs that are not in the list.
But we should think if we want this in all of our integrations.
@ronykoz @anara123 What do you think?

anara123:
The probability for the same incident to occur exactly at the same time is very low. But usually I let the API handle it for me, so if I do `search_incidents(start_time=SOME_TIMESTAMP)` I assume that the API will return events that created after this time.
Because of that in some integrations, before I do `next_run = latest_incident_time + 1 # add one millisecond` or `client.search_alerts(start_time=LAST_TIME+1)` something like that.
But in this case I think all we need to do is a bit modify the Hello World API to return incidents with timestamp `start_time< X < start_time + 1 minute`

Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.py
anara123, line 790: I think instead of this change. Change line 763
https://github.com/demisto/content/pull/7214/files#diff-0c7a8de4fcba11b42a8433798afa9edbR763
And do:
`last_fetch = int(last_fetch) + 1`

Itay4:
@reutshal whats the status here?

reutshal:
> @reutshal whats the status here?

Will be closed today

reutshal:
@roysagi Can you please go over the changelog https://github.com/demisto/content/blob/HelloWorld-fetch_incidents_addition/Packs/HelloWorld/Integrations/HelloWorld/CHANGELOG.md
Thanks!

PR #7215, opened 2020-05-25
Marketplace - display first level integration images of pack dependencies

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
Instead of displaying all level integration images of pack dependencies, `displayedImages` will only have packs of first level dependencies.


IkaDemisto:
In case that product will take this idea, we can align sdk find-dependencies command instead of fixing that in content.

PR #7221, opened 2020-05-26
Unmock All AWS Integrations

avidan-H:
## Status
- [x] Ready

## Description
Adds all the AWS integration that use the `boto3` SDK to the unmockable list. See issue https://github.com/demisto/etc/issues/24926 for more details.

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No


avidan-H:
I added the `docs-approved` label since there is no documentation necessary for this PR.

PR #7239, opened 2020-05-26
[Marketplace] - upload packs changes

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Related Issues
related to: https://github.com/demisto/content/pull/7215

## Description
- Added cleanup in gcs in case pack is not part of content repo or paid packs (from private bucket)
- Fixed loading of first level dependencies integration images. Skipped integration images that are not part of dependency pack.


Tests/Marketplace/upload_packs.py
```diff
@@ -199,6 +201,47 @@ def update_index_folder(index_folder_path, pack_name, pack_path, pack_version=''
         return task_status
 
 
+def clean_non_existing_packs(index_folder_path, private_packs, storage_bucket):
```
yuvalbenshalom, line 204: can you create some unit tests for this?

IkaDemisto: Yes

IkaDemisto: @yuvalbenshalom  Added unit tests. Note that now`clean_non_existing_packs` function has detection that checks environment variables (circle CI env variables). And in order to mock it, I've used ` @patch.dict('os.environ',..` https://docs.python.org/3.4/library/unittest.mock.html#unittest.mock.patch.dict
Please take a look and let me know if it is o.k to use it.

Tests/Marketplace/upload_packs.py
```diff
@@ -169,6 +171,56 @@ def update_index_folder(index_folder_path, pack_name, pack_path, pack_version=''
...
+    Args:
+        index_folder_path (str): full path to downloaded index folder.
+        private_packs (list): priced packs from private bucket.
+        storage_bucket (google.cloud.storage.bucket.Bucket): google storage bucket where index.zip is stored.
+
+    Returns:
+        bool: whether cleanup was skipped or not.
+    """
+    if 'CI' not in os.environ or os.environ.get('CIRCLE_BRANCH') != 'master' \
+            or storage_bucket.name != GCPConfig.PRODUCTION_BUCKET:
```
yuvalbenshalom, line 188: in case we are NOT on the PRODUCTION_BUCKET, I think we should run the cleanup. this way wi will cleanup the build bucket and it will be a better representation of whats going to happen if a branch will be merged.

IkaDemisto: @yuvalbenshalom hmm I am not sure about non production bucket. The ci bucket is copied from master/production bucket (all master branches will have cleanup), so I don't see any case where it will ever enter the cleanup part on non production bucket.

IkaDemisto:
@yuvalbenshalom done

PR #7241, opened 2020-05-26
GetFirstObject transformer to CommonScripts

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
Moved GetFirstObject transformer to CommonScripts pack.


yaakovi:
this was a test script, it shouldn't be under the `Scripts` folder, should be under the `TestPlaybooks` folder.

dantavori:
@yaakovi are you sure? HelloWorld uses it as a transformer. Either way it should be in a pack.

PR #7255, opened 2020-05-27
New pack documentation suggestion

idovandijk:
## Status
In Progress - being discussed

## Related Issues
fixes: https://github.com/demisto/etc/issues/24012
related to: https://github.com/demisto/etc/issues/24932

## Description
Implemented some readme changes that include previously discussed suggestions and changes I saw fit:
- Reduced size of pack documentation by splitting it between pack  & playbook
- Stages/capabilities changed to "What can this pack do?" (more appealing). Reduced amount of text in that section, summarized it more
- Moved "Triggers" to be under playbook documentation instead of pack documentation, since the triggers trigger the playbook
- Added "Usage" title to playbook documentation. I moved the playbook usage description, the "Triggers", the "Configuration" and the "Best Practices & Suggestions" under the "Usage" section. After usage come the rest of the sections like "Dependencies", which the user should expect to see after the usage in my opinion.
- Adjusted header sizes in playbook documentation (the difference in size between headers was too small in my opinion).
- Removed "Pack Documentation" title in pack readme. Under that there was the pack description. I moved it to be in the "metadata" file of the pack, as the pack's description ("A pack used for the complete investigation of potential phishing incidents...") 
- Removed "Visualization" title in pack readme

Still need to think about what we visualize and where. Currently we see phishing playbook both in pack and in playbook documentation).

Playbook readme:
image

Pack readme:
image


anara123:
@idovandijk The pack readme looks good. I would only add section for reference links to other documentations like the main Phishing playbook documentation in this case.

michalgold:
@kirbles19 @yaron-libman please have a final look.

yaron-libman:
looks good - I like the split between the playbook and pack. let's see what customers have to say and we can always revisit if need be.

PR #7273, opened 2020-05-29
Signal Sciences - added verification to corp name param

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25049

## Description
According to SignalSciences API docs, the corp name must match the pattern `[0-9a-z_.-]+`

   - [x] No


kirbles19:
@Itay4 Done, no change.

PR #7274, opened 2020-05-30
ServiceNow v2 - init incident_name param with default value

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25052

## Description
`incident_name` exist in the `params` object, and hence the default value of `.get()` was not taken

also, added `defaultvalue` to the param in the yml file

   - [x] No

## Must have
- [x] Tests
See test failure at https://circleci.com/gh/demisto/content/54959#tests/containers/1

### DIFF
--- a/Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
+++ b/Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
@@ -1827,7 +1827,7 @@ def main():
     sysparm_limit = int(params.get('fetch_limit', 10))
     timestamp_field = params.get('timestamp_field', 'opened_at')
     ticket_type = params.get('ticket_type', 'incident')
-    incident_name = params.get('incident_name', 'number')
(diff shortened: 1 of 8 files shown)


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -1827,7 +1827,7 @@ def main():
     sysparm_limit = int(params.get('fetch_limit', 10))
     timestamp_field = params.get('timestamp_field', 'opened_at')
     ticket_type = params.get('ticket_type', 'incident')
-    incident_name = params.get('incident_name', 'number')
+    incident_name = params.get('incident_name', 'number') or 'number'
```
ShahafBenYakir, line 1830: Isn't this redundant? If there is no 'incident_name' param 'number` will always be the default.

kirbles19:
@Itay4 Done, no change.

PR #7275, opened 2020-05-30
Demisto REST API - handle uri given with slash as prefix

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25046

   - [x] No


kirbles19:
@Itay4 Done.

PR #7303, opened 2020-06-01
Test playbooks format fixes

roysagi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
related : https://github.com/demisto/etc/issues/23712

Fixing format for all test playbooks


hod-alpert:
I believe we decided to remove the validation for test playbooks so fixing test playbooks for validation is nice to have but not a must, right @anara123 ?

ronykoz:
@hod-alpert @anara123 checking the scheme should be important, not a must but important IMO

anara123:
@ronykoz right now schema of test playbook fails on non existing descriptions. There is no need to  require description for test playbooks. If we create separate schema for test playbooks, then it is ok. But if not, lets remove validation for test playbooks

roysagi:
ok closing this PR and I'll add a schema for test playbooks to my validation improvements

PR #7310, opened 2020-06-01
ME_Service_Desk_Plus_integration

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/17456

## Description
A new integration for Manage Engine Service Desk Plus

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.yml
```diff
@@ -0,0 +1,860 @@
...
+  name: server_url
+  options:
+  - United States
+  - Europe
+  - India
+  - China
+  - Australia
+  required: true
+  type: 15
+- additionalinfo: test additional
```
yuvalbenshalom, line 17: did you plan to write something else here?

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.yml
```diff
@@ -0,0 +1,860 @@
...
+  name: client_id
+  required: true
+  type: 4
+- display: Client Secret
+  name: client_secret
+  required: true
+  type: 4
+- display: Refresh Token
+  name: refresh_token
+  required: false
```
yuvalbenshalom, line 28: not required? what is the flow that this will be used?

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.yml
```diff
@@ -0,0 +1,860 @@
...
+  type: 4
+- display: Fetch incidents
```
yuvalbenshalom, line 30: lets move fetch down to be before the parameters relevant for fetch

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
```diff
@@ -0,0 +1,853 @@
...
+            fetch_filter = str(params.get('fetch_filter')) if params.get('fetch_filter') else ''
+            fetch_limit = int(params.get('fetch_limit', '10')) if params.get('fetch_limit') else 10
+
+            date_format = '%Y-%m-%dT%H:%M:%S'
+            time_from = date_to_timestamp(parse_date_range(fetch_time, date_format=date_format, utc=False)[0])
+            time_to = date_to_timestamp(datetime.now(), date_format=date_format)
+            list_info = create_fetch_list_info(str(time_from), str(time_to), fetch_status, fetch_filter, fetch_limit + 1)
+            params = {
+                'input_data': f'{list_info}'
+            }
```
yuvalbenshalom, line 763: this one needs to be re considered.
the idea is to call the fetch like it is called in every fetch, and the only thing that is done is that the fetch checks if it is a test and know not to get the last run. lets talk about this.

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
```diff
@@ -0,0 +1,853 @@
...
+    hr = f'### Successfully closed request {request_id}'
+    return hr, {}, result
+
+
+def fetch_incidents(client: Client, fetch_time: str, fetch_limit: int, status: str, fetch_filter: str) -> list:
+    date_format = '%Y-%m-%dT%H:%M:%S'
+    last_run = demisto.getLastRun()
+    if not last_run:  # if first time running
+        new_last_run = {
+            'time': date_to_timestamp(parse_date_range(fetch_time, date_format=date_format, utc=False)[0])
```
yuvalbenshalom, line 678: are you sure you will get a list here that is not empty in all cases?

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
```diff
@@ -0,0 +1,853 @@
...
+        last_run_id = last_run.get('id', '0')
+        last_incident_id = last_run.get('id', '0')
+        cur_time = new_last_run.get('time', 0)
+        incident_creation_time = new_last_run.get('time', 0)
+
+        for incident in incidents:
+            if count >= fetch_limit:
+                break
+            # Prevent fetching twice the same incident
+            if incident.get('id') == last_run_id:
```
yuvalbenshalom, line 706: shouldn't we break here?

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.yml
darkushin: No, it isn't required. The user can enter only the client id and secret and then use the integration to create the refresh token that will be used in the instance. I will explain in more details when we talk.

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
darkushin: No, the last incident that was previously fetched might be the first incident that will be returned the next time fetch is called. We want to skip this incident and use all others so we can't break here.

Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
darkushin: Moved the fetch parameters to the client and changed test_module to call fetch_incidents if isFetch

kirbles19:
@yuvalbenshalom Done.

PR #7311, opened 2020-06-01
Fix common server python test:

GalRabin:
## Status
- [ ] In Progress
- [x] Ready

## Description
Fix common server python unit-tests (test_http_requests_with_retry_negative_sanity):
1. Using 3-rd party.
2. Reduce backoff time in retries.

## Screenshots
Before (04:38 minutes)-
image

After (0.26 seconds)-
image

   - [x] No


Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -1249,7 +1249,7 @@ def test_http_requests_with_retry_sanity(self, method):
     ]
 
     @pytest.mark.parametrize('method, status', RETRIES_NEGATIVE_TESTS_INPUT)
-    def test_http_requests_with_retry_negative_sanity(self, method, status):
+    def test_http_requests_with_retry_negative_sanity(self, method, status, requests_mock):
```
hod-alpert, line 1252: Using requests_mock replaces the HTTPAdapter with a different one and basically make the test redundant since it is not testing the retry mechanism

hod-alpert: You can still reduce the backoff time of course.

hod-alpert: since the backoff is exponential, You can reduce the retries from 3 to 1, that will also reduce most of the time

GalRabin:
FYI - @glicht @yaakovi 
After consulting with @hod-alpert  and @yaakovi , If we mock the requests, it replaces the Adapter which makes the test unnecessary (it's not really retried), Because it does not really retry, although the test pass which is not ok.
By approval of @yaakovi, I am removing this test, because its 3-rd party implementation which we don't need to check that way. however, if further tests are required, we should implement it in different way (But I don't think we need one currently).

glicht:
> FYI - @glicht @yaakovi
> After consulting with @hod-alpert and @yaakovi , If we mock the requests, it replaces the Adapter which makes the test unnecessary (it's not really retried), Because it does not really retry, although the test pass which is not ok.
> By approval of @yaakovi, I am removing this test, because its 3-rd party implementation which we don't need to check that way. however, if further tests are required, we should implement it in different way (But I don't think we need one currently).

Put it in skip so people can still run it manually if they want.

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -1312,6 +1313,7 @@ def test_http_requests_with_retry_negative_sanity(self, method, status):
                                       '',
                                       full_url='http://httpbin.org/status/{}'.format(status),
                                       retries=3,
+                                      backoff_factor=0.0001,
```
hod-alpert, line 1316:
```suggestion
```
If it's manual anyways I think it's best to keep the backoff as default to avoid manual running of the test from failing on overloading.

PR #7339, opened 2020-06-02
Default - Fixed test + moved to pack

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25185

## Description
- Tweaked Default test to not throw errors because of AlienVault OTX v2 quota
- Moved test to the Default pack

## Minimum version of Demisto
5.0.0

No


ShahafBenYakir:
@idovandijk Can you post the images of the test before and after? since you moved the test I can't tell what is new and what is old

idovandijk:
> @idovandijk Can you post the images of the test before and after? since you moved the test I can't tell what is new and what is old

the playbook looks exactly the same, but one of the condition checks that the range of IP exists instead of geo location

PR #7345, opened 2020-06-02
Added tests to phishing pack

idovandijk:
## Status
Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/24429

## Description
Added the test playbooks/script to the phishing pack


idovandijk:
Verified with Ika that we can move the files

PR #7363, opened 2020-06-03
Active Directory Query - improve custom attributes parsing

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25227

## Description
The issue was trying to set to `Account` output fields that did not exist in the response

AD fields start with a lower-case letter, so added a try to look for the field that starts with one, otherwise just won't add to account outputs and continue silently

   - [x] No

## Must have
- [x] Tests

See failing test at https://app.circleci.com/pipelines/github/demisto/content/14268/workflows/ed66c1c4-3d41-4f06-97ee-080eca0bc5ce/jobs/55832


Packs/Active_Directory_Query/Integrations/Active_Directory_Query/Active_Directory_Query.py
```diff
@@ -99,7 +99,14 @@ def account_entry(person_object, custome_attributes):
     }
 
     for attr in custome_attributes:
-        account[attr] = person_object[attr]
+        try:
+            account[attr] = person_object[attr]
+        except Exception as e:
```
avidan-H, line 104:
```suggestion
        except KeyError as e:
```

Packs/Active_Directory_Query/Integrations/Active_Directory_Query/Active_Directory_Query.py
```diff
@@ -99,7 +99,14 @@ def account_entry(person_object, custome_attributes):
...
+            lower_cased_attr = attr[0].lower() + attr[1:]
+            if lower_cased_attr in person_object:
+                account[lower_cased_attr] = person_object[lower_cased_attr]
```
avidan-H, line 107: This only accounts for the case where the user has accidentally capitalized the first letter of the attribute name. I'm not familiar with AD User attributes, but if they are unique (when indifferent to capitalization), I think we could make this more robust against capitalization typos - we can compare using `.casefold()` and assign the value of the index of the matching key.

Itay4: this is the standard in AD, but i will apply your suggestion in any case

avidan-H:
@Itay4 Up to your discretion, but this seems like something you could have added a unit test for instead of updating the test playbook.

Itay4:
@avidan-H done
note i also changed the log level from `debug` to `error`

kirbles19:
@Itay4 Done.

PR #7379, opened 2020-06-03
Fix wrong mapper names

Itay4:
## Status
- [x] Ready

## Description
Fixed wrong classifier names

   - [x] No


Packs/CommonTypes/Classifiers/classifier-mapper-incoming-Mail-listener.json
```diff
@@ -1,6 +1,6 @@
 {
 	"id": "mail-listener-mapper",
```
yaakovi, line 2: again with the indentation?

PR #7388, opened 2020-06-04
Fixing content

roysagi:
fixing additional content
related: https://github.com/demisto/etc/issues/23712


roysagi:
Circle seems to fail on an unrelated issue - https://github.com/demisto/etc/issues/25306

PR #7396, opened 2020-06-04
Proofpoint Protection Server - use html.parser instead of lxml parser and update required admin role

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25023

## Description
 - Using `lxml` parser raised `FeatureNotFound: Couldn't find a tree builder with the features you requested: lxml. Do you need to install a parser library?`, so changed to using `html.parser` instead which worked as expected.
 - Updated docs and description with required admin role

   - [x] No


Itay4:
> Can we add a unit test?

dont have an example response, so would have to generate one somehow - dont think it worth the effort

ShahafBenYakir:
@Itay4 If we don't have an instance how did we validated this one?

kirbles19:
@Itay4 Done.

Itay4:
@kirbles19 please note i added another parameter to the integration and an entry to the release notes
can you please review again ?

Packs/ProofpointServerProtection/Integrations/ProofpointServerProtection/ProofpointServerProtection.py
```diff
@@ -312,6 +313,24 @@ def smart_search_command():
...
+    try:
+        cmd_url = '/admin?module=RPC&class=InputValidator&method=getSMD&pps_magic=' + pps_magic
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_0) '
+                          'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.131 Safari/537.36',
+            'X-Requested-With': 'XMLHttpRequest'
+        }
+        submit_search_response = http_request('GET', cmd_url, headers=headers)
+        submit_search_response_json = json.loads(submit_search_response)
+        service_url = submit_search_response_json.get('serviceURL')
```
yaakovi, line 326:
```suggestion
        service_url = submit_search_response_json.get('serviceURL', '')
```

PR #7407, opened 2020-06-04
timestamp_to_datestring uses `utcfromtimestamp`

hstern:
`timestamp_to_datestring`'s default date format includes Z for the time
zone. However, it uses `datetime.fromtimestamp` which is in localtime.
This yields incorrect results when the default time zone is anything
other than UTC.

The `epochToTimestamp` function in the same file does correctly use
`utcfromtimestamp`. This commit corrects and normalizes the timestamp
processing.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?

### DIFF
--- a/Packs/Base/ReleaseNotes/1_0_7.md
+++ b/Packs/Base/ReleaseNotes/1_0_7.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+Fixed incorrect time zone parsing for **timestamp_to_datestring**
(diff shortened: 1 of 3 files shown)


yaakovi:
CLA is failing on a Demisto developer.
force-pushing

hod-alpert:
See `epochToTimestamp` method in `CommonServerPython`
```
def epochToTimestamp(epoch):
    return datetime.utcfromtimestamp(epoch / 1000.0).strftime("%Y-%m-%d %H:%M:%S")
```
Where the method has no 'Z' in the end but still it parses it according to utc.

The local time in the old code could also be different from the customer's time without any indication.

PR #7409, opened 2020-06-04
Fix dependencies install

yuvalbenshalom:
## Status
- [ ] In Progress
- [x] Ready

## Description
try to fix the bug where dependent packs that should be updated are dropped from the installed list.
this sold solve unsuccessful installation of pacs by the build.

the same fix is applied for this branch "build_fix" were install failed because of packs that were dropped from the install list, active directory and CommonScripts. so before merge, the market place step on this build should be checked also:
https://app.circleci.com/pipelines/github/demisto/content/14673/workflows/a88705f8-3d3f-48c2-a504-a8f09c53fde1/jobs/56408/steps
## Screenshots

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -513,10 +513,6 @@ jobs:
                 echo "Skipping - not running in INSTANCE_TESTS build or unit-tests container"
                 exit 0
             fi
-            if [[ -z "${NIGHTLY}" ]]; then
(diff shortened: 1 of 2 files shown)


IkaDemisto:
@yuvalbenshalom please wait with this PR. I am able to run this script locally and reproduce.

PR #7411, opened 2020-06-05
FireEye Helix - fix headers arg processing in search cmd

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25339

   - [x] No

## Must have
- [x] Tests

See failing test at https://app.circleci.com/pipelines/github/demisto/content/14705/workflows/620f8e6b-dfe8-4b27-86b0-31a0c7cc5da1/jobs/56440/parallel-runs/1?filterBy=FAILED

### DIFF
--- a/Packs/FireEyeHelix/Integrations/FireEyeHelix/FireEyeHelix.py
+++ b/Packs/FireEyeHelix/Integrations/FireEyeHelix/FireEyeHelix.py
@@ -1614,7 +1614,11 @@ def search_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict]:
     Returns:
         Outputs
     """
-    query = build_mql_query(**args)
+    query = build_mql_query(
+        query=args.get('query'), start=args.get('start'), end=args.get('end'), page_size=args.get('page_size'),
+        limit=args.get('limit'), offset=args.get('offset'), groupby=args.get('groupby'), sort_by=args.get('sort_by'),
+        sort_order=args.get('sort_order')
+    )
     raw_response = client.search(query)
     headers = argToList(args.get('headers'))
     return build_search_result(raw_response, headers=headers)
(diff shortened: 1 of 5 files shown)


kirbles19:
@Itay4 Done.

Packs/FireEyeHelix/Integrations/FireEyeHelix/FireEyeHelix.py
```diff
@@ -1614,7 +1614,11 @@ def search_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict]:
     Returns:
         Outputs
     """
-    query = build_mql_query(**args)
+    query = build_mql_query(
+        query=args.get('query', ''), start=args.get('start'), end=args.get('end'), page_size=args.get('page_size'),
+        limit=args.get('limit'), offset=args.get('offset'), groupby=args.get('groupby'), sort_by=args.get('sort_by'),
+        sort_order=args.get('sort_order')
+    )
```
liorblob, line 1621: looks a bit messy to me. Also, is this the only place this fix is required? what about the other function calls?

Itay4: what do you mean messy? do you a suggestion here?

went over the other functions, it seems to be the only place to edit for 2 reasons:
1. other commands that have `headers` arg don't pass `**args` as in the search command
2. other commands that pass `**args` dont have the `headers` command

also, we dont have an instance so i prefer to do only required changes to fix the issue

liorblob: just declare the arguments outside I think

Itay4: i guess its matter of code style, but i really dont see how that makes it "less messy"

PR #7419, opened 2020-06-05
Fixed server version calculation

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
fixed server version calculation based on `env_results.json` in case of multiple servers (nightly)


yaakovi:
@yuvalbenshalom merging for now.
as long as this is the latest version, it should be ok.
we can update the AMI name so it will contain the right version number or we can add an edge-case check in this function

PR #5973, opened 2020-03-09
Fix Vulture reports in integrations

Itay4:
## Status
- [x] Ready
## Description
Fixed Vulture reports in integrations

## Does it break backward compatibility?
   - [x] No

PR #7421, opened 2020-06-07
Red Canary - bug fixes in fetch incidents

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25265

## Description
 - in https://github.com/demisto/content/pull/5973, a condition was fixed that led to all timeline of a detection will be fetched, instead of only the first page.
apparently a timeline can be huge, in such way that caused the issue and made the fetch times out. so I've removed the timeline retrieval from the fetch and that can be done in a playbook with a command.
 - also encountered a bug in which acknowledged detections were fetched, so fixed the condition there as well.

   - [x] No

### DIFF
--- a/Packs/RedCanary/Integrations/RedCanary/RedCanary.py
+++ b/Packs/RedCanary/Integrations/RedCanary/RedCanary.py
@@ -320,7 +320,7 @@ def get_unacknowledged_detections(t, per_page=50):
             attributes = detection.get('attributes', {})
             # If 'last_acknowledged_at' or 'last_acknowledged_by' are in attributes,
             # the detection is acknowledged and should not create a new incident.
-            if attributes.get('last_acknowledged_at') is None or attributes.get('last_acknowledged_by') is None:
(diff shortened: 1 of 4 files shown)


kirbles19:
@Itay4 Done.

Packs/RedCanary/Integrations/RedCanary/RedCanary.py
```diff
@@ -320,7 +320,7 @@ def get_unacknowledged_detections(t, per_page=50):
             attributes = detection.get('attributes', {})
             # If 'last_acknowledged_at' and 'last_acknowledged_by' are in attributes,
             # the detection is acknowledged and should not create a new incident.
-            if attributes.get('last_acknowledged_at') and attributes.get('last_acknowledged_by'):
+            if not (attributes.get('last_acknowledged_at') and attributes.get('last_acknowledged_by')):
```
yaakovi, line 323: if I remember correctly, I think that in some cases only one of them was filled out so better to check that both not defined.

PR #7434, opened 2020-06-07
increase sshd MaxStartups and restart sshd

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25349

## Description
For each server instance, we increase the MaxStartups to `20:30:100` and then restart the `sshd` service. Because the build runs in threads and the mocking mechanism executes commands on a given server instance via ssh, we may be hitting the `sshd` connection limit of the server instance. Can read more about it [here](https://appuals.com/fix-ssh_exchange_identification-connection-closed-by-remote-host/).

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No

### DIFF
--- a/Tests/scripts/wait_until_server_ready.py
+++ b/Tests/scripts/wait_until_server_ready.py
@@ -58,9 +58,13 @@ def main():
                 try:
                     res = requests.request(method=method, url=(host + path), verify=False)
                 except (requests.exceptions.RequestException, requests.exceptions.HTTPError) as exp:
-                    print_error(f'{ami_instance_name} encountered an error: {str(exp)}\n'
-                                f'Spot instance was dropped by amazon, if raised often - report to team leader.')
-                    instance_ips_to_poll.remove(ami_instance_ip)
+                    # print_error(f'{ami_instance_name} encountered an error: {str(exp)}\n'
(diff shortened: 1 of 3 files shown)


Tests/scripts/wait_until_server_ready.py
```diff
@@ -58,15 +58,23 @@ def main():
                 try:
                     res = requests.request(method=method, url=(host + path), verify=False)
                 except (requests.exceptions.RequestException, requests.exceptions.HTTPError) as exp:
-                    print_error(f'{ami_instance_name} encountered an error: {str(exp)}\n'
-                                f'Spot instance was dropped by amazon, if raised often - report to team leader.')
-                    instance_ips_to_poll.remove(ami_instance_ip)
+                    # print_error(f'{ami_instance_name} encountered an error: {str(exp)}\n'
+                    #             f'Spot instance was dropped by amazon, if raised often - report to team leader.')
+                    print_error(f'{ami_instance_name} encountered an error: {str(exp)}')
+                    if SETUP_TIMEOUT != 60 * 10:  # noqa: F823
```
yaakovi, line 64: declare as global at the beginning of the function

PR #7443, opened 2020-06-08
Improved bad response handling

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25395

## Description
Added a try except clause that catches a value error and return a valid exception instead.


Itay4:
@hod-alpert note the build fails on the integration tests

kirbles19:
@hod-alpert Done.

PR #7446, opened 2020-06-08
Hash incidents fields

yguezpa:
## Status
- [x] In Progress
- [ ] Ready

## Description
Get incidents and hash some of the fields


kirbles19:
@erezh31 @yguezpa Done, but seems there's a changelog file missing.

Packs/ML/Scripts/HashIncidentsFields/HashIncidentsFields.py
```diff
@@ -0,0 +1,74 @@
...
+    else:
+        entry = res[0]
+        incident_list = json.loads(entry['Contents'])
+        new_incident_list = []
+        for incident in incident_list:
+            for field in fieldsToHash_list:
+                if field in incident:
+                    incident[field] = hash_multiple(incident.get(field))
+            new_ctx = {}
+            ctx = get_context(incident['id'])
```
erezh31, line 47: please check if context is needed. if `contextKeys` args is empty -  no need to query the context at all

Packs/ML/Scripts/HashIncidentsFields/HashIncidentsFields.yml
```diff
@@ -0,0 +1,125 @@
...
+  name: limit
+  required: false
+  secret: false
+- auto: PREDEFINED
+  default: false
+  defaultValue: 'false'
+  description: Whether to query and fetch incident context. Can be "true" or "false".
+    The default is "false".
+  isArray: false
+  name: includeContext
```
erezh31, line 43: you don't need this argument, because you query the context separately, remove it

yguezpa:
@kirbles19 Added release note. Can you check it please ?

Packs/ML/Scripts/HashIncidentsFields/HashIncidentsFields.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        entry = res[0]
+        incident_list = json.loads(entry['Contents'])
+        new_incident_list = []
+        for incident in incident_list:
+            for field in fieldsToHash_list:
+                if field in incident:
+                    incident[field] = hash_multiple(incident.get(field))
+            new_ctx = {}
+            if demisto.args().get('contextKeys', '').split(","):
```
erezh31, line 47: This condition is not good, the expression always contains empty string.
you need first to extract the context keys like that:
`context_keys = [x for x in demisto.args().get('contextKeys', '').split(",") if x]`
and than you can use `if context_keys: ...`

PR #7449, opened 2020-06-08
Fix rasterize pack name

yuvalbenshalom:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25252

## Description
Rasterize part was without capital R, that caused it not to be loaded as core pack


anara123:
@yuvalbenshalom why we rename the pack name to be capital?

yuvalbenshalom:
will change the name of the pack in the core pack list and not the path

PR #7453, opened 2020-06-09
RTIR: fix ID header bug

jochman:
First, need to merge: https://github.com/demisto/content/pull/7452
Then can review this pr
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25429

## Description
Fix problem where there was an `ID` as a substring as a header.

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/RTIR/Integrations/RTIR/RTIR.yml
+++ b/Packs/RTIR/Integrations/RTIR/RTIR.yml
@@ -586,4 +586,3 @@ script:
 tests:
 - RTIR Test
 fromversion: 4.1.0
-fromversion: 4.1.0
--- a/Packs/RTIR/Integrations/RTIR/RTIR.yml
+++ b/Packs/RTIR/Integrations/RTIR/RTIR.yml
@@ -585,3 +585,5 @@ script:
   subtype: python2
 tests:
 - RTIR Test
+fromversion: 4.1.0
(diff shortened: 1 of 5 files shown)


Itay4:
lets also add a test for the given case, and verify it is not caught

Packs/RTIR/Integrations/RTIR/RTIR.yml
```diff
@@ -585,3 +585,5 @@ script:
   subtype: python2
 tests:
 - RTIR Test
+fromversion: 4.1.0
+fromversion: 4.1.0
```
Itay4, line 589: i think one is enough

Packs/RTIR/ReleaseNotes/1_0_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+- __RTIR__
+Lint and format fixes
```
Itay4, line 4: lets update to the actual rn here

Packs/RTIR/Integrations/RTIR/RTIR.yml
jochman: dsdk bug

PR #7468, opened 2020-06-09
Fix collect tests and content packs 

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Description
- Replaced DocumentationTest with HelloWorld-Test in case no tests were collected.
- Added DeveloperTools to packs_to_install and moved TestCommonPython to DeveloperTools pack.


Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1116,11 +1116,16 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
             print_warning('Collecting sample tests due to: {}'.format(','.join(sample_tests)))
         else:
             print_warning("Running Sanity check only")
-            tests.add('DocumentationTest')  # test with integration configured
+
             tests.add('TestCommonPython')  # test with no integration configured
+            packs_to_install.add("DeveloperTools")
```
yaakovi, line 1121: why do you need this? aren't we always installing that pack?

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1116,11 +1116,16 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
...
+
+            tests.add('HelloWorld-Test')  # test with integration configured
+            packs_to_install.add("HelloWorld")
 
     if changed_common:
         tests.add('TestCommonPython')
+        packs_to_install.add("DeveloperTools")
```
yaakovi, line 1128: same

Tests/scripts/collect_tests_and_content_packs.py
IkaDemisto: Installing but not re-uploading. @yaakovi see my comment on second pr

dantavori:
@yuvalbenshalom the build passed except for the validate errors which I opened an [issue](https://github.com/demisto/etc/issues/25480) about.

PR #7474, opened 2020-06-09
changing playbook name

roysagi:
## Description
fixing file name without comma


ronykoz:
@yaakovi this should not cause problems, right?

yaakovi:
changing file names are ok (just keep it short, under 80 chars)

PR #7477, opened 2020-06-09
Improve packs installing

reutshal:
## Status
- [] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25169

## Description
Added also packs with changed files that do not have any tests


Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1071,17 +1069,44 @@ def get_content_pack_name_of_test(tests: set, id_set: Dict = None) -> set:
...
+        if not file_data:
+            continue
+        file_status = file_data[0]
+        if file_status.lower().startswith('r'):
+            file_path = file_data[2]
+        else:
+            file_path = file_data[1]
+
+        # ignore files in ".circle", ".github" and ".hooks" directories and .gitignore
+        if not file_path.startswith('.'):
```
yaakovi, line 1087: since you look for specific prefixes (`Base`, `Documentation`), this is redundant.

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1071,17 +1069,42 @@ def get_content_pack_name_of_test(tests: set, id_set: Dict = None) -> set:
     return content_packs
 
 
+def get_modified_packs(files_string):
+    modified_packs = set([])
```
IkaDemisto, line 1073: Why not just set()?

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1071,17 +1069,42 @@ def get_content_pack_name_of_test(tests: set, id_set: Dict = None) -> set:
...
+    all_files = files_string.split('\n')
```
IkaDemisto, line 1074: You can use `splitlines()` instead

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1120,6 +1143,11 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
     if 'NonSupported' in packs_to_install:
         packs_to_install.remove("NonSupported")
 
+    # get all modified packs - not just tests related
+    modified_packs = get_modified_packs(files_string)
```
IkaDemisto, line 1147: Lets avoid adding `NonSupported` pack in case it was modified. So you can move 
```
  if 'NonSupported' in packs_to_install:
        packs_to_install.remove("NonSupported")
```
to the end. Or ignore appending this pack in `get_modified_packs` function.

reutshal: Good catch. moved to the end

PR #7486, opened 2020-06-10
Fixed print when GCS_MARKET_KEY is not set

hod-alpert:
## Description
When $GCS_MARKET_KEY is not set- print `GCS_MARKET_KEY not set aborting!` instead of `not set aborting!`

##Screenshots
image


Tests/scripts/prepare_content_packs_for_testing.sh
```diff
@@ -11,7 +11,7 @@ ID_SET=$CIRCLE_ARTIFACTS/id_set.json
 EXTRACT_FOLDER=$(mktemp -d)
 
 if [[ -z "$GCS_MARKET_KEY" ]]; then
```
Itay4, line 13: lets skip the whole step in forked PRs so we wont even run this script

@yuvalbenshalom FYI

Itay4:
build fails on marketplace
merging

PR #7497, opened 2020-06-10
fix for IsMaliciousIndicatorFound tpb

teizenman:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24398


teizenman:
Yes, the use of the integration before was not that relevant anyway.

PR #7512, opened 2020-06-11
Refactor packs xsoar author

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Related Issues
related to: https://github.com/demisto/etc/issues/25157

## Description
Refactored author of xsoar supported packs to `Palo Alto Networks`

### DIFF
--- a/Tests/Marketplace/marketplace_services.py
+++ b/Tests/Marketplace/marketplace_services.py
@@ -70,7 +70,7 @@ class Metadata(object):
     DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
     XSOAR_SUPPORT = "xsoar"
     XSOAR_SUPPORT_URL = "https://www.paloaltonetworks.com/cortex"  # disable-secrets-detection
-    XSOAR_AUTHOR = "Cortex XSOAR"
+    XSOAR_AUTHOR = "Palo Alto Networks"
     SERVER_DEFAULT_MIN_VERSION = "6.0.0"
     CERTIFIED = "certified"
 
@@ -369,8 +369,8 @@ class Pack(object):
         doesn't match xsoar default url, warning is raised.
 
         Args:
-            support_type (str): support type of pack, optional values are: partner, developer, xsoar and nonsupported
-            support_url (str): support full url
+            support_type (str): support type of pack.
+            support_url (str): support full url.
             support_email (str): support email address.
 
         Returns:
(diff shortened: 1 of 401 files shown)


Tests/Marketplace/marketplace_services.py
```diff
@@ -70,7 +70,7 @@ class Metadata(object):
     DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
     XSOAR_SUPPORT = "xsoar"
     XSOAR_SUPPORT_URL = "https://www.paloaltonetworks.com/cortex"  # disable-secrets-detection
-    XSOAR_AUTHOR = "Cortex XSOAR"
+    XSOAR_AUTHOR = "Palo Alto Networks"
```
IkaDemisto, line 73: @yuvalbenshalom  Pay attention also for this changes, you cannot see it diff due to large number of changed packs

asafshen:
@IkaDemisto the server is heavily relay on the `Cortex XSOAR` string

I am not sure that this is the right approach to change it from content side right now

IkaDemisto:
> @IkaDemisto the server is heavily relay on the `Cortex XSOAR` string
> 
> I am not sure that this is the right approach to change it from content side right now

@anara123 @yuvalbenshalom so how exactly we should rename xsoar author?

anara123:
@asafshen Why you rely specifically  on `Cortex XSOAR` string ?
I think this should be defined in the content side - or at least content should have the option to override whatever is defined in server.

Following logic:
if author provided, then display the author.
If author not provided and the support is `xsoar` then display `Palo Alto Networks`

asafshen:
@anara123 we use this constant in few places
1. to mark pack as contributed by us
image

2. to mark pack in marketplace with the top right icon
image

3. show this logo in the details tab
image

probably more places

image


this didn't came out from the blue :) we coordinate it somewhere.. anyway if we want to change this, we need to make the proper changes on our side

please open an issue so we can handle, I don't want to use the issue above which its subject is discalmer

asafshen:
you can override it, but then it won't show as "by cortex xsoar" and with logo in all places...

for this change we must involve product, this has high affect on the UX, if you change to palo alto and we show xsoar logo, its weird.

to be honest I don't think its practical to change it to the beta,
we still have bugs to handle, and we will have to get proper assets for panw...

if the priority of this will be critical we will of course.. but we must coordinate it cross teams

PR #7535, opened 2020-06-12
To ISO Format Transformer

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25351

## Description
This PR adds a transformer that takes a date string and converts it to ISO-8601 format. That transformer has then been added as an additional transformer to the `CreationDate` and `UpdatedDate` indicator fields for the `Domain` indicator type.

## Minimum version of Demisto
- [x] 5.5.0 (for the updated indicator type)

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/Scripts/DateStringToISOFormat/DateStringToISOFormat.yml
+++ b/Packs/CommonScripts/Scripts/DateStringToISOFormat/DateStringToISOFormat.yml
@@ -1,4 +1,3 @@
-
 commonfields:
   id: DateStringToISOFormat
   version: -1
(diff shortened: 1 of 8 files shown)


Packs/CommonScripts/Scripts/DateStringToISOFormat/DateStringToISOFormat.yml
```diff
@@ -0,0 +1,50 @@
+
```
Itay4, line 1: remove empty line

Packs/CommonScripts/Scripts/DateStringToISOFormat/DateStringToISOFormat.yml
```diff
@@ -0,0 +1,50 @@
...
+enabled: true
+args:
+- name: value
+  required: true
+  description: Date value to convert.
+- name: dayfirst
+  default: true
+  auto: PREDEFINED
+  predefined:
+  - "True"
```
Itay4, line 22: lets use `true` and `false`

Itay4:
cool looks good
we need to catch exceptions raised from `parse`, such as `ParserError` for unknown string format

avidan-H:
I set the default for parsing ambiguous 3-number dates such that `dayfirst` is `True`. I just want to confirm, that is the desired default behavior?

Itay4:
> I set the default for parsing ambiguous 3-number dates such that `dayfirst` is `True`. I just want to confirm, that is the desired default behavior?

yea sounds good

kirbles19:
@avidan-H Done, no change.

PR #7024, opened 2020-05-18
unit42 feed - taxii2.0

bakatzir:
## Status
- [x] Ready. ~~missing indicatorfields pr merge.~~

**Question from the meetings with unit42:**
Are all indicators bad? yes
Is the feed reliable? very much so
pagination? they will dm me how they implemented it
only one collection? yes
Did I miss some indicator types? no
Mapping? @mayagoldb will help - -> There is mapping of IOCs to attack-pattern, malware and campaing via relationship. all data in the referenced issue.

## Related Issues
fixes: https://github.com/demisto/etc/issues/21926

## Related PRs
moving the IOCs, indicatorfields: https://github.com/demisto/content/pull/7582

## Minimum version of Demisto
- [x] 5.5.0

## Does it break backward compatibility?
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

PR #7582, opened 2020-06-15
move mitre indicatortype and field from mitre pack to common pack

bakatzir:
## Status
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/21926, https://github.com/demisto/etc/issues/22586

## Description
I've seen that the mitre pack has NO dependencies.
moving the indicatortype and fields to common, so it can be used in other feeds, e.g DHS and Unit42.
Please validate me.

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No(Hopefully)

reference: https://github.com/demisto/content/pull/7024


bakatzir:
`- Cisco ASA - Test Playbook` fails, unrelated.
@yuvalbenshalom Please consider force-merging.

PR #7599, opened 2020-06-16
ThreatX - increase test timeout

Itay4:
fixes https://github.com/demisto/etc/issues/25599

the change done in https://github.com/demisto/content/pull/7557 increases the runtime of the test

increased the timeout to allow the test to finish running

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -340,7 +340,7 @@
         {
             "integrations": "ThreatX",
             "playbookID": "ThreatX-test",
-            "timeout": 3600
+            "timeout": 600
         },
         {
             "integrations": "Akamai WAF SIEM",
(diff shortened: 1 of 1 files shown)


Tests/conf.json
```diff
@@ -339,7 +339,8 @@
         },
         {
             "integrations": "ThreatX",
-            "playbookID": "ThreatX-test"
+            "playbookID": "ThreatX-test",
+            "timeout": 3600
```
ShahafBenYakir, line 343: an hour?!

Itay4: fine fine
changed to 600 secs

ShahafBenYakir: So magzim

PR #7615, opened 2020-06-16
Uploader - added support for overriding specific pack

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
Added support to override specific pack by adding:
```
"override": true
```
in `pack_metadata.json`


IkaDemisto:
@yaakovi FYI

IkaDemisto:
@yuvalbenshalom triggering build is complicated and will cause problems to developers that have no idea how marketplace flow works. This solution is much easier.

PR #7616, opened 2020-06-16
Run zipping packs only on master

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Related Issues
fixes: https://github.com/demisto/etc/issues/25443

## Description
Changed zip of all packs creation to run only on master branch and master bucket.


IkaDemisto:
@liorblob on master bucket we can run the script with `-rt False`, correct?

liorblob:
@IkaDemisto yes since the bucket doesn't have any test playbooks

PR #7618, opened 2020-06-16
Add 'multiple' Command Argument to Zscaler's 'url' command

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25684

## Description
Added an argument to the `url` command in `Zscaler` called `multiple`. By default, it's value will be `"true"` which will maintain backward compatibility. When the argument is set to `"false"`, the command won't interpret the value passed to the `url` command's `'url'` argument as a comma separated list of URLs. This will allow for the submission of singular URLs that have commas in them.

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Zscaler/.secrets-ignore
+++ b/Packs/Zscaler/.secrets-ignore
@@ -9,3 +9,4 @@ https://maybe_valid
 https://user-images.githubusercontent.com
 2.2.2.2
 3.3.3.3
+https://madeup.fake.com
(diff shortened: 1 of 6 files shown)


kirbles19:
@avidan-H Done.

PR #7619, opened 2020-06-16
Move RegPathReputationBasicLists test to D2 pack

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Description
Moved `RegPathReputationBasicLists_test` playbook to D2 pack.


IkaDemisto:
@dantavori @yuvalbenshalom  we are not overriding packs no more. Moving items between packs without new version has no effect on master bucket.

PR #7621, opened 2020-06-16
Add ML collect data script

erezh31:
## Status
- [ ] In Progress
- [x] Ready


Packs/Base/Scripts/DBotMLFetchData/DBotMLFetchData.py
```diff
@@ -0,0 +1,20 @@
+import demistomock as demisto
```
meirwah, line 1: why you need this?

Packs/Base/Scripts/DBotMLFetchData/DBotMLFetchData.yml
```diff
@@ -0,0 +1,14 @@
+comment: Collect machine learning data from the environment
```
meirwah, line 1: Collect telemetry is better here to say I think

Packs/Base/Scripts/DBotMLFetchData/DBotMLFetchData.py
erezh31: It's in every content script, this is the template (it's actually removed when the automation deployed)

erezh31:
@meirwah fixed comments

Packs/Base/Scripts/DBotMLFetchData/DBotMLFetchData.yml
```diff
@@ -11,4 +11,4 @@ tags:
   - ml
 timeout: '0'
 type: python
-deprecated: true
\ No newline at end of file
+deprecated: true
```
meirwah, line 14: why is this?

erezh31: we put this to all internal scripts that the user should not run.. like the auto mapper or sane doc report

PR #7624, opened 2020-06-16
Instance test enhancments

ShahafBenYakir:
## Status
- [x] In Progress
- [ ] Ready

## Description
Some small fixes to instance test step including the creation of an artifact that includes all failing tests in addition to the list.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -557,9 +557,11 @@ jobs:
       - run:
           name: Instance Test
           command: |
-            if [ -n "${INSTANCE_TESTS}" ] && [ $CIRCLE_NODE_INDEX -ne 1 ] ;
+#            if [ -n "${INSTANCE_TESTS}" ] && [ $CIRCLE_NODE_INDEX -ne 1 ] ;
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -560,6 +560,7 @@ jobs:
             if [ -n "${INSTANCE_TESTS}" ] && [ $CIRCLE_NODE_INDEX -ne 1 ] ;
               then
                 ./Tests/scripts/instance_test.sh
+                cp ./Tests/failed_instances.txt $CIRCLE_ARTIFACTS/failed_instances.txt
```
yaakovi, line 563: this is hiding the error code of the bash script, need to handle it like we do [here](https://github.com/demisto/content/pull/7624/files#diff-1d37e48f9ceff6d8030570cd36286a61R530-R532)

ShahafBenYakir: Is this an escape room? link leads to nowhere...

yaakovi: see lines 530-532 in this file.

ShahafBenYakir:
Failing on MP step. Force merging.
FYI @yuvalbenshalom

PR #7625, opened 2020-06-17
Mongodb: nested dicts fix

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25735

## Description
Nested dicts were not parsed properly when they had a datetime object.
Test introducing the bug:
image

   - [ ] No

## Must have
- [x] Tests

### DIFF
--- a/Packs/MongoDB/Integrations/MongoDB/MongoDB.py
+++ b/Packs/MongoDB/Integrations/MongoDB/MongoDB.py
@@ -169,7 +169,7 @@ def convert_id_to_object_id(
     return _convert(entries)
 
 
-def convert_str_to_datetime(entries: dict) -> dict:
+def convert_str_to_datetime(entries: Any) -> Any:
     """ Converts list or dict with date values of type str to Datetime.
 
     Args:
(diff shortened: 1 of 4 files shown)


Packs/MongoDB/Integrations/MongoDB/MongoDB_test.py
```diff
@@ -61,6 +61,11 @@ def test_convert_str_to_datetime_no_datetime_obj(self):
         res = convert_str_to_datetime(inputs)
         assert isinstance(res[1], int)
 
+    def test_dict_inside_dict(self):
```
Itay4, line 64: add test docstring according to template `Given-When-Then`

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -169,7 +169,7 @@ def _convert(entry: dict):
     return _convert(entries)
 
 
-def convert_str_to_datetime(entries: dict) -> dict:
+def convert_str_to_datetime(entries: Any) -> Any:
     """ Converts list or dict with date values of type str to Datetime.
```
Itay4, line 173: update docstring

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -180,13 +180,23 @@ def convert_str_to_datetime(entries: dict) -> dict:
...
+        matches = regex_for_timestamp.findall(entries)
+        if len(matches) == 1:
             try:
-                entries[key] = datetime.strptime(match, '%Y-%m-%dT%H:%M:%S.000Z')
-            except ValueError as e:
-                raise DemistoException(str(e))
-
+                return datetime.strptime(matches[0], time_format)
+            except ValueError:
+                # If could not parse the date, it may not a date. Return entries
```
Itay4, line 190: i think this is not handled appropriately
in what cases will we get here and what will be returned?

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
```diff
@@ -180,13 +180,23 @@ def convert_str_to_datetime(entries: dict) -> dict:
...
+                return entries
+        # If no match or more than 1, return itself
```
Itay4, line 192: why ?

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
jochman: I don't think that it could ever get to the except, it's there for ensuring that it'll never get there.
If somehow, someone could pass a string that is both fitting the regex and not fitting the time format, it'll return the string itself.

Packs/MongoDB/Integrations/MongoDB/MongoDB.py
jochman: because if there are two matches for the regex, which one would you pick as a timestamp?
if there are two valid timestamps in the string, it's not a timestamp (non-MongoDB valid)

kirbles19:
@jochman Done, no change.

PR #7628, opened 2020-06-17
Fix cfw extra arg

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25725

## Description
ipname argument has been never used and not needed for the command to work. So now it's deprecated to not break BC.

   - [ ] Yes
       - Further details:
   - [x] No


kirbles19:
@jochman Done.

PR #7647, opened 2020-06-18
Microsoft Defender ATP - set scope to default

Itay4:
##  Status
- [x] Ready

## Description
The default scope that was set was for graph, changed it to ATP one

   - [x] No


kirbles19:
@Itay4 Done.

PR #7654, opened 2020-06-18
[expanse-272] Updates for version 1.2.0

andrew-expanse:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Responding to customer and XSOAR dev feedback.
- Adding an additional playbook
- Adding additional Incident Fields
- Adding classifiers for different incident types
- Adding a Layout to support Expanse Behavior Incidents
- Updating integration to leverage the Demisto Integration Cache to allow adherence to the configured page size

## Screenshots
Expanse_Behavior_Severity_Update_Thu_Jun_18_2020

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No
      - Further details: Some irrelevant incident types were removed from IncidentFields they were mistakenly added to, however there was no classifier/mapping in place for these fields, so they were never populated. 

## Must have
- [X] Tests
- [X] Documentation 

## Demisto Partner?
- [X] The title must be in the following format: **[YOUR_PARTNER_ID] short description**


Packs/Expanse/Classifiers/classifier-Expanse_5_9_9.json
```diff
@@ -0,0 +1,95 @@
...
+	"brandName": "Expanse",
+	"custom": true,
+	"defaultIncidentType": "",
+	"feed": false,
+	"id": "95573d05-345e-43b6-82f1-a5a8ad402e69",
+	"incidentSamples": [],
+	"indicatorSamples": null,
+	"isDefault": false,
+	"keyTypeMap": {},
+    "fromVersion": "5.0.0",
```
Itay4, line 11: please fix indentation here

Packs/Expanse/IncidentFields/incidentfield-Expanse_Business_Unit.json
```diff
@@ -0,0 +1,48 @@
...
+    "unmapped": false,
+    "unsearchable": false,
+    "caseInsensitive": true,
+    "columns": null,
+    "defaultRows": null,
+    "sla": 0,
+    "threshold": 72,
+    "breachScript": "",
+    "validatedError": "",
+	"fromVersion": "5.0.0"
```
Itay4, line 47: please fix indentation here

Packs/Expanse/IncidentFields/incidentField-Expanse_Behavior_Rule.json
```diff
@@ -0,0 +1,47 @@
...
+    "unmapped": false,
+    "unsearchable": false,
+    "caseInsensitive": true,
+    "columns": null,
+    "defaultRows": null,
+    "sla": 0,
+    "threshold": 72,
+    "breachScript": "",
+    "validatedError": "",
+	"fromVersion": "5.0.0"
```
Itay4, line 46: please fix indentation here

Packs/Expanse/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
   "name": "Expanse",
   "description": "The Expanse App for Demisto leverages the Expander API to retrieve network exposures and create incidents in Demisto.",
   "support": "partner",
-  "currentVersion": "1.0.1",
+  "currentVersion": "1.2.0",
```
Itay4, line 5: lets bump to `1.1.0` and not `1.2.0`

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -698,29 +696,56 @@ def fetch_incidents_command():
...
+            incidents_to_send = stored_incidents[:PAGE_LIMIT]
+            del stored_incidents[:PAGE_LIMIT]
+        else:
+            incidents_to_send = list(stored_incidents)
+            stored_incidents = None
+        demisto.incidents(incidents_to_send)
+
+        # Update Cache
+        cache["incidents"] = stored_incidents
+        demisto.setIntegrationContext(cache)
```
Itay4, line 748: i dont really like the change here
the cache isnt meant for that and it could introduce bugs in edge cases
why do we need that and not limit in the request as it was?

andrew-expanse:
> looks good, general comments:
> 
> 1. no need for changelog files anymore, you can remove them all
> 2. please refactor `incidentField` to incidentfield` in incident fields filenames
> 3. why did we remove the associated type of behavior for some fields?

Hi @Itay4,
1. They can be removed for all of the different entities? Classifier, IncidentFields, etc?
2. 👍 
3. We removed the behavior type from some fields because we had initially applied it by mistake. However these fields were not being mapped to previously so I don't believe we should encounter any issues by removing it.

Packs/Expanse/Integrations/Expanse/Expanse.py
andrew-expanse: Maybe we could loop @anara123 or @fvigo into the conversation here. We spoke about this in Slack - the real issue is with the Expanse API not being able to do any kind of limiting/filtering to enable us to respect the incident page limit param.

This approach include pulling all applicable events once, and storing them in the cache and popping off then publishing a portion of them during each incident fetch. The alternative approach would be similar, but we would only store ids in the cache, and during each `fetch-incidents` we would pull all events, and use the next X ids from the cache to determine what will get published as an incident.

I like this approach more because we will only hit the API once (less network traffic/latency) and because it's slightly less complex.

andrew-expanse: I'm also more than happy to look into other options for accomplishing this - unfortunately the APIs we're using are unlikely to change to allow a more straight-forward approach to this.

Itay4:
> Hi @Itay4,
> 
> 1. They can be removed for all of the different entities? Classifier, IncidentFields, etc?
> 2. 👍
> 3. We removed the behavior type from some fields because we had initially applied it by mistake. However these fields were not being mapped to previously so I don't believe we should encounter any issues by removing it.

changelog can be removed for all

Packs/Expanse/Integrations/Expanse/Expanse.py
Itay4: @andrew-expanse how big is the response returned from the API ? trying to understand if there will be an actual impact on the performance if we query for all incidents every fetch

andrew-expanse: It will vary quite a bit by customer, but I'd anticipate it could be 5-10 requests to paginate through all of the data each time. Due to API limitations we're doing client-side filtering on severity, so a large majority of the data pulled is likely thrown away each time. 

From an earlier conversation about this I estimated that the max size we would be storing in the cache would be 
```
My guess on max size would be ~30MB assuming no filtering is done, but since we have a default filter in place I would expect things to be ~10MB on the high end, typically lower.
```
The response to this was 
```
Load balancing group should not have a problem.
10MB-30MB shouldnt be a problem.
But still worth to validate it.
But I would say 30MB is the limit
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -18,7 +18,7 @@
 
 
 API_KEY = demisto.params().get('api_key')
-PAGE_LIMIT = demisto.params().get('page_limit')
+PAGE_LIMIT = int(demisto.params().get('page_limit', '10'))
 FIRST_RUN = int(demisto.params().get('first_run', '7'))
 SERVER = 'https://expander.expanse.co'
 VERIFY_CERTIFICATES = not demisto.params().get('insecure')
```
Itay4, line 24: `PROXY` arg here is not used and can be removed since you're calling `handle_proxy()`

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
     return_outputs(human_readable, ec, exposures)
 
 
+def domains_for_certificate_command():
+    """
+
+    :return:
```
Itay4, line 963: please either fill in the docstring or remove it

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+
+    raw_domain_objects = context['DomainList']
+    del context['DomainList']  # Remove full objects from human readable response
+    human_readable = tableToMarkdown("Expanse Domains matching Certificate Common Name: {search}".format(search=search),
+                                     context)
+    context['DomainList'] = raw_domain_objects
+    return_outputs(human_readable, ec, matching_domains)
+
+
+def _fetch_certificates(params, token):
```
Itay4, line 1005: not sure whats the value of making it a private func

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+        if len(results['data']) > 0:
+            certificates += results['data']
+            next_page = results.get('pagination', {}).get('next', None)
+            while next_page is not None:
+                params['pageToken'] = get_page_token(next_page)
+                results = http_request('GET', 'assets/certificates', params, token=token)
+                certificates += results['data']
+                next_page = results.get('pagination', {}).get('next', None)
+        return certificates
+    except Exception:
```
Itay4, line 1025: is there no value to log here the exception ?

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    Returns details for a single certificate.
+
+    :param md5_hash: Search term for certificates
+    :param token: Expanse Refresh token
+    :return: Certificate details objects
+    """
+    try:
+        return http_request('GET', 'assets/certificates/{}'.format(md5_hash), {}, token=token)
+    except Exception:
+        demisto.results("No data found")
```
Itay4, line 1041: better to call `demisto.results` in the command func and here just return empty array

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+        "commonNameSearch": search
+    }
+    token = do_auth()
+
+    observed_ips = []
+    matching_domains = []
+
+    certificates = _fetch_certificates(params=params, token=token)
+    for certificate in certificates:
+        certificate_details = _fetch_certificate(md5_hash=certificate['certificate']['md5Hash'], token=token)
```
Itay4, line 976:
```suggestion
        certificate_details = _fetch_certificate(md5_hash=certificate.get('certificate', {}).get('md5Hash'), token=token)
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+            certificates += results['data']
+            next_page = results.get('pagination', {}).get('next', None)
+            while next_page is not None:
+                params['pageToken'] = get_page_token(next_page)
+                results = http_request('GET', 'assets/certificates', params, token=token)
+                certificates += results['data']
+                next_page = results.get('pagination', {}).get('next', None)
+        return certificates
+    except Exception:
+        demisto.results("No data found")
```
Itay4, line 1026: better to call `demisto.results` in the command func and here just return empty array

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    """
+    Returns details for a single certificate.
+
+    :param md5_hash: Search term for certificates
+    :param token: Expanse Refresh token
+    :return: Certificate details objects
+    """
+    try:
+        return http_request('GET', 'assets/certificates/{}'.format(md5_hash), {}, token=token)
+    except Exception:
```
Itay4, line 1040: is there no value to log here the exception ?

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+                results = http_request('GET', 'assets/certificates', params, token=token)
+                certificates += results['data']
+                next_page = results.get('pagination', {}).get('next', None)
+        return certificates
+    except Exception:
+        demisto.results("No data found")
+        return []
+
+
+def _fetch_certificate(md5_hash, token):
```
Itay4, line 1030: not sure whats the value of making it a private func

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    }
+    token = do_auth()
+
+    observed_ips = []
+    matching_domains = []
+
+    certificates = _fetch_certificates(params=params, token=token)
+    for certificate in certificates:
+        certificate_details = _fetch_certificate(md5_hash=certificate['certificate']['md5Hash'], token=token)
+        for ip in certificate_details['details']['recentIps']:
```
Itay4, line 977:
```suggestion
        for ip in certificate_details.get('details', {}).get('recentIps', []):
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+            observed_ips.append(ip['ip'])
```
Itay4, line 978:
```suggestion
            observed_ips.append(ip.get('ip'))
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+
+    for ip in observed_ips:
```
Itay4, line 980: i dont think we need that loop
we are already iterating all objects above when building `observed_ips` list, why not just do it there?

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    :return: Certificate details objects
+    """
+    try:
+        return http_request('GET', 'assets/certificates/{}'.format(md5_hash), {}, token=token)
+    except Exception:
+        demisto.results("No data found")
+        return {}
+
+
+def _fetch_ips(params, token):
```
Itay4, line 1045: not sure whats the value of making it a private func

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+        if len(results['data']) > 0:
+            ips += results['data']
+            next_page = results.get('pagination', {}).get('next', None)
+            while next_page is not None:
+                params['pageToken'] = get_page_token(next_page)
+                results = http_request('GET', 'assets/ips', params, token=token)
+                ips += results['data']
+                next_page = results.get('pagination', {}).get('next', None)
+        return ips
+    except Exception:
```
Itay4, line 1065: is there no value to log here the exception ?

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+        demisto.results("No data found")
```
Itay4, line 1066: better to call `demisto.results` in the command func and here just return empty array

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -617,15 +620,26 @@ def exposure_stats(exposures):
...
 
 
+def get_expanse_certificate_to_domain_context(common_name, data):
+    """
+    provides custom context information for domains looked up via certificate
+    """
+    return {
+        "SearchTerm": common_name,
+        "TotalDomainCount": len(data),
+        "FlatDomainList": [domain['domain'] for domain in data],
```
Itay4, line 630:
```suggestion
        "FlatDomainList": [domain.get('domain') for domain in data],
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -617,15 +620,26 @@ def exposure_stats(exposures):
     }
 
 
+def get_expanse_certificate_to_domain_context(common_name, data):
+    """
+    provides custom context information for domains looked up via certificate
```
Itay4, line 625: please add arguments and return to docstring

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+        demisto.results("No data found")
+        return
+
+    context = get_expanse_certificate_to_domain_context(common_name=search, data=matching_domains)
+
+    ec = {
+        'Expanse.IPDomains(val.SearchTerm == obj.SearchTerm)': context
+    }
+
+    raw_domain_objects = context['DomainList']
```
Itay4, line 997:
```suggestion
    raw_domain_objects = context.pop('DomainList')
```

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    del context['DomainList']  # Remove full objects from human readable response
```
Itay4, line 998: remove this line after applying the above suggestion

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -914,6 +957,116 @@ def exposures_command():
...
+    human_readable = tableToMarkdown("Expanse Domains matching Certificate Common Name: {search}".format(search=search),
+                                     context)
+    context['DomainList'] = raw_domain_objects
```
Itay4, line 1001: the logic here for building the human readable and context outputs here is weird
no need to remove the domain list from context and add it again, just initialize a list for the humanreadable while will be a copy of the context and remove from it the domainlist

Packs/Expanse/Integrations/Expanse/Expanse.yml
```diff
@@ -655,7 +655,31 @@ script:
     - contextPath: Expanse.Exposures.Exposures.Provider
       description: Provider details of the Exposure
       type: Unknown
-  dockerimage: demisto/python3:3.8.2.6981
+  - arguments:
+    - default: true
+      description: The certificate common name.
+      isArray: true
```
Itay4, line 661: should this is `true`? you're not handling it in the code

Packs/Expanse/Integrations/Expanse/Expanse.yml
```diff
@@ -655,7 +655,31 @@ script:
     - contextPath: Expanse.Exposures.Exposures.Provider
       description: Provider details of the Exposure
       type: Unknown
-  dockerimage: demisto/python3:3.8.2.6981
+  - arguments:
+    - default: true
+      description: The certificate common name.
```
Itay4, line 660: maybe add an example here? when can it be retrieved from?

Packs/Expanse/Integrations/Expanse/Expanse.py
andrew-expanse: good catch

Packs/Expanse/Integrations/Expanse/Expanse.yml
andrew-expanse: You're right, should be `false`.

Itay4:
@andrew-expanse the run time `!expanse-get-domains-for-certificate common_name=ma` seem to be about 3 minutes (because we query each certificate) - that's a lot

are we sure we want it to be like that? if so, lets mention that the run time can be long in the command description, as it just seem not to respond

Itay4:
also, please add to the new command to the test playbook

Packs/Expanse/Integrations/Expanse/Expanse.py
andrew-expanse: Good call, there was a better way to do this - I'd just pulled from an earlier method.

andrew-expanse:
> @andrew-expanse the run time `!expanse-get-domains-for-certificate common_name=ma` seem to be about 3 minutes (because we query each certificate) - that's a lot
> 
> are we sure we want it to be like that? if so, lets mention that the run time can be long in the command description, as it just seem not to respond

Sure, I can add some more details there. That is quite a long run time, but searching `ma` as a common name would hopefully be fairly uncommon in reality (the fuzzy matching on the certificate common name is likely returning dozens or hundreds of results we need to check individually).

Packs/Expanse/Integrations/Expanse/Expanse.py
```diff
@@ -698,35 +714,69 @@ def fetch_incidents_command():
...
-    # Save last_run
-    demisto.setLastRun({
-        "complete_for_today": True,
-        "start_time": yesterday
-    })
+    # Check if we've stored any events in the integration cache
+    cache = demisto.getIntegrationContext()
+    stored_incidents = cache.get("incidents")
+
+    if stored_incidents is None:
```
Itay4, line 721: it would be great if we could add a unit test for that flow to verify the cache mechanism works as expected

andrew-expanse:
@Itay4 Thanks for the thorough review - I appreciate it! 

I need to get the final unit test working and then I _think_ this should be good to go on my end.

Packs/Expanse/TestPlaybooks/playbook-Expanse-Test.yml
```diff
@@ -1,22 +1,21 @@
 id: test-Expanse
-version: -1
+version: 6
```
Itay4, line 2:
```suggestion
version: -1
```

Packs/Expanse/TestPlaybooks/playbook-Expanse-Test.yml
```diff
@@ -216,35 +220,113 @@ tasks:
...
+    quietmode: 0
+  "7":
+    id: "7"
+    taskid: aa2f5d92-1b67-4c54-89ea-d67aaec93703
+    type: regular
+    task:
+      id: aa2f5d92-1b67-4c54-89ea-d67aaec93703
+      version: -1
+      name: expanse-get-domains-for-certificate
+      script: Expanse_copy|||expanse-get-domains-for-certificate
```
Itay4, line 257:
```suggestion
      script: Expanse|||expanse-get-domains-for-certificate
```

Packs/Expanse/TestPlaybooks/playbook-Expanse-Test.yml
```diff
@@ -216,35 +220,113 @@ tasks:
...
+      type: regular
+      iscommand: true
+      brand: Expanse_copy
```
Itay4, line 260:
```suggestion
      brand: Expanse
```

Packs/Expanse/TestPlaybooks/playbook-Expanse-Test.yml
```diff
@@ -257,11 +339,17 @@ inputs:
...
   description: domain
+  playbookInputQuery: null
+- key: common_name
+  value:
+    simple: '*.enron.com'
+  required: false
+  description: A partial certificate common name
+  playbookInputQuery: null
 outputs: []
-description: ''
```
Itay4, line 266: revert these changes please, `description` and `fromversion` should still be here

Itay4:
@andrew-expanse the current domain used in the test playbook `atlas.enron.com` does not return results using our instance, can we please change it to something that does return results, such as `enron.com` ?

Packs/Expanse/Playbooks/Expanse_Incident_Playbook_CHANGELOG.md
```diff
@@ -1,9 +0,0 @@
-## [Unreleased]
```
mayagoldb, line 1: @andrew-expanse Why was this file deleted? Was the 'Expanse Incident Playbook' removed?

andrew-expanse: This was in response to an earlier review comment - https://github.com/demisto/content/pull/7654#issuecomment-648777603

I can add this back if Changelogs for playbooks should not be removed.

mayagoldb: Oh you are right, you can leave it like this 👌

Packs/Expanse/Playbooks/Expanse_Behavior_Severity_Update.yml
```diff
@@ -0,0 +1,1636 @@
...
+      type: regular
+      iscommand: true
+      brand: Expanse
+    nexttasks:
+      '#none#':
+      - "7"
+    scriptarguments:
+      ip:
+        complex:
+          root: ${Expanse
```
mayagoldb, line 55: @andrew-expanse you can remove the `${` from this task since you are using 'get'

Packs/Expanse/Playbooks/Expanse_Behavior_Severity_Update.yml
```diff
@@ -0,0 +1,1636 @@
...
+          accessor: IP.Address}
```
mayagoldb, line 56: @andrew-expanse remove the `}` from this row as well

PR #7656, opened 2020-06-18
Fix microsoft test playbook

barchen1:
## Status
- [x] Ready

## Description
Add 3 retries to create-alert task cause this task failes on connection errors.


Itay4:
@barchen1 lets add retries on all commands tasks

PR #7659, opened 2020-06-19
CDL - set temp creds dbfile

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25786

## Description
set a temp file in env var to be loaded in https://github.com/PaloAltoNetworks/pan-cortex-data-lake-python/blob/master/pan_cortex_data_lake/adapters/tinydb_adapter.py#L45

   - [x] No

### DIFF
--- a/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
+++ b/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
@@ -1,8 +1,5 @@
-from dateutil import parser
-
+""" IMPORTS """
 from CommonServerPython import *
-
-''' IMPORTS '''
 import os
 import requests
 import json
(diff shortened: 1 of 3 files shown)


Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
```diff
@@ -741,6 +740,8 @@ def fetch_incidents(client: Client,
 
 
 def main():
+    dbfile = NamedTemporaryFile(delete=False)
```
glicht, line 743: I would use a fixed file so it is the same between requests. Otherwise we are loosing the caching (not sure how important it is). Use tempfile.gettempdir() and a fixed file name.

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.py
```diff
@@ -789,6 +790,9 @@ def main():
     except Exception as e:
         error_message = str(e)
         return_error(error_message)
+    finally:
+        if dbfile:
```
glicht, line 794: As part of the previous remark. Let's leave the file for caching.

Itay4:
@glicht done

kirbles19:
@Itay4 Done, no change.

PR #7661, opened 2020-06-20
Microsoft Teams - handle notifications from server

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25819

## Description
 - handle notifications from server that come with notifications channel only in the arguments
 - add unclassified to incident severity notifications
 - update module health on webserver start - needed for case it crashed and spins up again
 - move integration_context init out of the try-except block, as it raised var referenced before var sometimes
 - added traceback print to logs

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests

### DIFF
--- a/Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
+++ b/Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
@@ -33,6 +33,7 @@ configuration:
   display: Minimum incident severity to send notifications to Teams by
   name: min_incident_severity
   options:
+  - Unclassified
   - Low
   - Medium
   - High
--- a/Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
+++ b/Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
@@ -286,7 +286,7 @@ script:
     description: Creates a new channel in a Microsoft Teams team.
     execution: false
     name: microsoft-teams-create-channel
-  dockerimage: demisto/teams:1.0.0.7727
+  dockerimage: demisto/teams:1.0.0.9271
   feed: false
   isfetch: false
   longRunning: true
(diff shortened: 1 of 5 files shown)


kirbles19:
@Itay4 Done.

Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
```diff
@@ -33,6 +33,7 @@ configuration:
   display: Minimum incident severity to send notifications to Teams by
   name: min_incident_severity
   options:
+  - Unclassified
```
liorblob, line 36: I think the term is `Unknown` but doesn't really matter

Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams_test.py
```diff
@@ -632,6 +632,118 @@ def test_send_message(mocker, requests_mock):
...
+    requests_mock.post(
+        f'{service_url}/v3/conversations/19:67pd3966e74g45f28d0c65f1689132bb@thread.skype/activities',
+        json={}
+    )
+    send_message()
+    results = demisto.results.call_args[0]
+    assert len(results) == 1
+    assert results[0] == 'Message was sent successfully.'
+
+
```
liorblob, line 746: Following the bug we had on slack, I would add a test for sending a message to the dedicated channel (the bug was that the channel was being deleted because it was the dedicated channel)

Itay4: dont think its relevant here

Packs/MicrosoftTeams/Integrations/MicrosoftTeams/MicrosoftTeams.yml
Itay4: fixed

PR #7673, opened 2020-06-21
Attivo Botsink - catch ValueError exception

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25832

   - [x] No


Itay4:
@roysagi please review rn

PR #7688, opened 2020-06-22
Pull Request template - add run as non-root user to must haves

Itay4:
## Status
- [x] Ready

## Description
Added run as non-root to must haves section


glicht:
I don't know if we need this. I don't want to overload the template.

PR #7694, opened 2020-06-22
snow v2 - Fixed the test button to work with debug mode

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25871

## Description
test_module did not return the bool value for `ignoreautoextract`
I've also uploaded andclocked `test` locally, it's working.

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests

### DIFF
--- a/Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
+++ b/Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
@@ -1794,7 +1794,7 @@ def fetch_incidents(client: Client) -> list:
     return incidents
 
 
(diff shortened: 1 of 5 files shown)


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -1794,7 +1794,7 @@ def fetch_incidents(client: Client) -> list:
     return incidents
 
 
-def test_module(client: Client, *_):
+def test_module(client: Client, *_) -> Tuple[Any, Dict[Any, Any], Dict[Any, Any], bool]:
```
liorblob, line 1797:
```suggestion
def test_module(client: Client, *_) -> tuple:
```

Might be simpler

PR #7699, opened 2020-06-22
fixed f string

jochman:
## Status
- [] In Progress
- [x] Ready


amshamah419:
We should probably write some unit tests for this change.

PR #7721, opened 2020-06-23
New phishing tests on 6.0, old tests removed

idovandijk:
## Status
- [ ] In Progress - Need to add test for `Phishing - Core` as well
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25914

## Description
Changed the way phishing is tested. This will apply only to 6.0. Previous tests were removed.
Also, disabled "continue on error" for the tasks that use `SetGridField`.
The constraints were:
- `PhishingIncident` script does not work (does not create a phishing incident)
- The phishing playbooks now use `SetGridField` which is dependent on an incident field that **only exists in the Phishing incident type.**
- Pre-6.0, it's impossible to create a Phishing incident type and investigate it automatically. What we can do is change the current type of the incident, but this will not create the incident fields necessary for the tests (Email Headers field is necessary for `SetGridField`). This has never been a problem because we used `setIncident` which is server-side, and does not cause problems for fields that do not exist.
- The only way to automatically investigate incidents pre-6.0 is to change the current default playbook for phishing incidents, and to set it to run automatically - this may affect customers negatively.
- In order to properly test phishing with the full flow, we cannot ignore errors with `SetGridField`, so the only way here is to run the tests only on 6.0. What I did was create a playbook that creates a new incident of type phishing. I set the incident to run a different, separate test playbook, and investigated it using the first playbook. Since we now won't continue when having errors on `SetGridField`, I had to remove the phishing tests from previous versions (before 6.0).

The tests now create a malicious indicator and do not depend on Minemeld.

## Minimum version of Demisto
6.0

Yes - `SetGridField` will stop on errors now. This should not have a big effect because it has only been less than 2 weeks and customers have not adapted to that script yet.

## Must have
- [ ] Tests


idovandijk:
Blocked by this https://github.com/demisto/etc/issues/25947

idovandijk:
Update: there is no workaround, the issue must be fixed

idovandijk:
@michalgold the Process Email - Generic test also fails because of the new grid, should we skip this test and use the phishing test to test process email, or bring back the "continue on error" for SetGridField?
Process Email - Generic test is more in depth than the phishing when it comes to processing the email

idovandijk:
RN approved with Yaron

PR #7732, opened 2020-06-24
Reverting restrictive api change

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25915

## Description
For some reason if there are no arguments configured for a command- all arguments are accepted and used in `demisto.args()`
But if there are arguments configured for a command- all non-configured arguments will be ignored

Therefore, the change that added this `command`argument have broke backwards compatibility, so reverting it back is the solution.


Itay4:
@roysagi can you please review rn here?

PR #7742, opened 2020-06-25
Search and install updates

liorblob:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25782

## Description
Ignoring `IGNORED_FILES` in packs search and install;
Raising exception in case a pack was not found in the bucket.

Did not manage to reproduce these scenarios in unit tests. If someone has an idea please share.


Tests/Marketplace/search_and_install_packs.py
```diff
@@ -138,9 +138,9 @@ def search_pack(client, prints_manager, pack_display_name):
 
             else:
                 print_msg = 'Did not find pack {} in bucket.\n'.format(pack_display_name)
-                prints_manager.add_print_job(print_msg, print_color, 0, LOG_COLORS.YELLOW)
+                prints_manager.add_print_job(print_msg, print_color, 0, LOG_COLORS.RED)
                 prints_manager.execute_thread_prints(0)
-                return {}
+                raise Exception(print_msg)
```
yuvalbenshalom, line 143: I thin that if we will raise exception here this will not keep going and try search the other packs. correct?
we need the build to fail in the end, but we need it to keep going until the end.

liorblob: This is running inside a thread, so it will crash when we do `thread.join`. That specific thread will stop working so some packs might be missed. That's why it's a little complicated to know the state of this process. @guykeller do you have an idea how to handle this situation similarly to the nightly build?

liorblob: @yaakovi do you have an idea maybe?

yaakovi: didn't find anything useful to handle it.
have you looked at `multiprocessing.Pool` docs instead of `multithreading.Thread`?

liorblob: @yaakovi I did, but that would take a lot of effort as a lot of the logic uses a thread index.

liorblob:
@yaakovi @yuvalbenshalom I added a global flag to indicate the state. please have a look

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -303,7 +318,8 @@ def search_and_install_packs_and_their_dependencies(pack_ids, client, prints_man
         prints_manager (ParallelPrintsManager): A prints manager object.
         thread_index (int): the thread index.
 
-    Returns (list): A list of the installed packs' ids, or an empty list if is_nightly == True.
+    Returns (list): A list of the installed packs' ids, or an empty list if is_nightly == True, in addition to a flag
+    that indicates if the operation succeeded or not.
```
yaakovi, line 322:
```suggestion
    Returns (list, bool):
        A list of the installed packs' ids, or an empty list if is_nightly == True. 
        A flag that indicates if the operation succeeded or not.
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -149,7 +156,13 @@ def search_pack(client, prints_manager, pack_display_name):
             raise Exception(err_msg)
     except Exception as e:
         err_msg = 'The request to search pack {} has failed. Reason:\n{}'.format(pack_display_name, str(e))
-        raise Exception(err_msg)
+        prints_manager.add_print_job(err_msg, print_color, 0, LOG_COLORS.RED)
+        prints_manager.execute_thread_prints(0)
```
yaakovi, line 160: if this is used under a multithreaded run, don't use 0 as the thread index

PR #7746, opened 2020-06-26
Panos threat

adambaumeister:
## Status
- [x] In Progress

## Description
Adds support for two new commands to the PAN-OS integration
* panorama-block-vulnerability
  * Allows threat IDs to be overridden by user-configurable actions  (drop/alert/reset-both etc.)
* panorama-get-predefined-threats-list
  * Retrieves the entire signature database from a PANOS device for reference purposes such as looking up CSVs

Credit enir@paloaltonetworks.com for panorama-get-predefined-threats list!

## Minimum version of Demisto
- [X] 3.0.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -4768,7 +4768,7 @@ def panorama_block_vulnerability():
     drop_mode = demisto.args().get('drop_mode', 'drop')
 
     threat = panorama_override_vulnerability(threatid, vulnerability_profile, drop_mode)
-    threat_output = {'ID': threatid, 'NewAction': drop_mode }
+    threat_output = {'ID': threatid, 'NewAction': drop_mode}
 
     demisto.results({
         'Type': entryTypes['note'],
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -4750,12 +4750,13 @@ def panorama_get_predefined_threats_list(target: str):
         'GET',
         params=params
     )
-    demisto.results(fileResult('predefined-threats.json', json.dumps(result['response']['result']).encode('utf-8')))
+    return result
 
 
 def panorama_get_predefined_threats_list_command():
     target = str(demisto.args()['target']) if 'target' in demisto.args() else None
-    panorama_get_predefined_threats_list(target)
+    result = panorama_get_predefined_threats_list(target)
+    demisto.results(fileResult('predefined-threats.json', json.dumps(result['response']['result']).encode('utf-8')))
 
 
 def panorama_block_vulnerability():
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -4750,7 +4750,7 @@ def panorama_get_predefined_threats_list(target: str):
         'GET',
         params=params
     )
-    demisto.results(result['response']['result'])
+    demisto.results(fileResult('predefined-threats.json', json.dumps(result['response']['result']).encode('utf-8')))
 
 
 def panorama_get_predefined_threats_list_command():
@@ -4764,12 +4764,10 @@ def panorama_block_vulnerability():
     """
     threatid = demisto.args()['threat_id']
     vulnerability_profile = demisto.args()['vulnerability_profile']
-    drop_mode = "drop"
-    if demisto.args().get('drop_mode'):
-        drop_mode = demisto.args().get('drop_mode')
+    drop_mode = demisto.args().get('drop_mode', 'drop')
 
     threat = panorama_override_vulnerability(threatid, vulnerability_profile, drop_mode)
-    threat_output = {'id': threatid}
+    threat_output = {'ID': threatid, 'NewAction': drop_mode }
 
     demisto.results({
         'Type': entryTypes['note'],
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -4714,8 +4714,12 @@ def panorama_add_static_route_command():
         'EntryContext': {"Panorama.StaticRoutes(val.Name == obj.Name)": static_route}
     })
 
+
 def panorama_override_vulnerability(threatid: str, vulnerability_profile: str, drop_mode: str):
-    xpath = "{}profiles/vulnerability/entry[@name='{}']/threat-exception/entry[@name='{}']/action".format(XPATH_OBJECTS, vulnerability_profile, threatid),
+    xpath = "{}profiles/vulnerability/entry[@name='{}']/threat-exception/entry[@name='{}']/action".format(
+        XPATH_OBJECTS,
+        vulnerability_profile,
+        threatid)
     params = {'action': 'set',
               'type': 'config',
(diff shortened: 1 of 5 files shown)


Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
```diff
@@ -1,17 +1,61 @@
...
+        <msg>command succeeded</msg>
+    </response>
+    """
+    requests_mock.post(base_url, text=mock_response_xml, status_code=200)
+    return requests_mock
+
+
+def test_panoram_get_os_version(patched_requests_mocker):
+    from Panorama import get_pan_os_major_version
+    get_pan_os_major_version()
```
bakatzir, line 53: shouldn't we use `assert` to test it's mocked value?

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
```diff
@@ -1,17 +1,61 @@
...
+
+
+def test_panoram_block_threatid(patched_requests_mocker):
+    from Panorama import panorama_block_vulnerability
+    panorama_block_vulnerability()
```
bakatzir, line 58: shouldn't we use `assert` to test it's mocked value?

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -199,6 +199,14 @@ script:
     description: Run any command supported in the API.
     execution: false
     name: panorama
+
+  - name: panorama-get-predefined-threats-list
```
bakatzir, line 203: no outputs?

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -556,6 +564,40 @@ script:
...
+    - default: false
+      description: Numberical Threat ID
+      isArray: false
+      name: threat_id
+      required: true
+      secret: false
+    deprecated: false
+    description: Sets a vulnerability signature to block mode
+    execution: false
+    name: panorama-block-vulnerability
```
bakatzir, line 599: no outputs?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4715,6 +4715,74 @@ def panorama_add_static_route_command():
...
+        'cmd': '<show><predefined><xpath>/predefined/threats</xpath></predefined></show>',
+        'target': target,
+        'key': API_KEY
+    }
+    result = http_request(
+        URL,
+        'GET',
+        params=params
+    )
+    demisto.results(result['response']['result'])
```
bakatzir, line 4753: we use the interior func to return it's result to the `command`, and do the `demisto.results()` from there.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4715,6 +4715,74 @@ def panorama_add_static_route_command():
...
+    params = {
+        'type': 'op',
+        'cmd': '<show><predefined><xpath>/predefined/threats</xpath></predefined></show>',
+        'target': target,
+        'key': API_KEY
+    }
+    result = http_request(
+        URL,
+        'GET',
+        params=params
```
bakatzir, line 4751: why do you send it in the params and not in the body?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4715,6 +4715,74 @@ def panorama_add_static_route_command():
...
+
+def panorama_block_vulnerability():
+    """
+    Ovverride a vulnerability signature such that it is in block mode
+    """
+    threatid = demisto.args()['threat_id']
+    vulnerability_profile = demisto.args()['vulnerability_profile']
+    drop_mode = "drop"
+    if demisto.args().get('drop_mode'):
+        drop_mode = demisto.args().get('drop_mode')
```
bakatzir, line 4769:
```suggestion
drop_mode = demisto.args().get('drop_mode', 'drop')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4715,6 +4715,72 @@ def panorama_add_static_route_command():
...
+
+    return http_request(
+        URL,
+        'POST',
+        body=params,
+    )
+
+
+@logger
+def panorama_get_predefined_threats_list(target: str):
```
bakatzir, line 4738: this function should return only the response form the api

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4715,6 +4715,72 @@ def panorama_add_static_route_command():
...
+    }
+    result = http_request(
+        URL,
+        'GET',
+        params=params
+    )
+    demisto.results(fileResult('predefined-threats.json', json.dumps(result['response']['result']).encode('utf-8')))
+
+
+def panorama_get_predefined_threats_list_command():
```
bakatzir, line 4756: while this function should call `demisto.results(fileResult(.....`

Packs/PAN-OS/Integrations/Panorama/Panorama.py
adambaumeister: Because it's a short OP command, there's no need to send via POST, but happy to change?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
adambaumeister: The predefined threat output is massive, probably way too much to load into the context, and fileResult makes it easy to return a File object to demisto i think. Thoughts?

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
adambaumeister: Added

Packs/PAN-OS/Integrations/Panorama/Panorama.py
adambaumeister: ah, disregard, i see what you mean :) added.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
bakatzir: Preferably, but not a must.

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
```diff
@@ -1,17 +1,63 @@
...
+            <model>Panorama</model>
+            <serial>FAKESERIALNUMBER</serial>
+        </result>
+    </response>
+    """
+    version_path = "{}{}{}".format(base_url, "?type=version&key=", integration_params['key'])
+    requests_mock.get(version_path, text=mock_version_xml, status_code=200)
+    mock_response_xml = """
+    <response status="success" code="20">
+        <msg>command succeeded</msg>
```
bakatzir, line 44:
```suggestion
    <msg>command succeeded</msg>
```

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
```diff
@@ -1,17 +1,63 @@
...
+    </response>
+    """
+    requests_mock.post(base_url, text=mock_response_xml, status_code=200)
+    return requests_mock
+
```
bakatzir, line 49:
```suggestion
```

PR #7754, opened 2020-06-28
pan-os - Added the *source_zone*, *destination_zone* arguments to the ***panor…

bakatzir:
…ama-create-rule*** command

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22954

## Description
the code for the args was there in the `prepare_security_rule_params` func, but never reached. so I added the args and checked.

## Minimum version of Demisto
- [x] 3.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


bakatzir:
Force merging as build fails on timeout and non related PBs.

PR #7755, opened 2020-06-28
Kafka v2 - set auto_offset_reset to intial offset to fetch from

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25999 https://github.com/demisto/etc/issues/25976

## Description
- Apparently `reset_offsets` does not set the first offset to fetch from as expected, so added the keyword arg `auto_offset_reset` according to https://pykafka.readthedocs.io/en/latest/usage.html?highlight=reset_offsets#setting-the-initial-offset
- Wrapped the available offset fetch, which seems to timeout in some cases, in try-except 

   - [x] No

### DIFF
--- a/Packs/Kafka/Integrations/Kafka_V2/Kafka_V2.py
+++ b/Packs/Kafka/Integrations/Kafka_V2/Kafka_V2.py
@@ -171,7 +171,7 @@ def print_topics(client):
     """
     Prints available topics in Broker
     """
-
+    include_offsets = demisto.args().get('include_offsets', 'true') == 'true'
     kafka_topics = client.topics.values()
     if kafka_topics:
         topics = []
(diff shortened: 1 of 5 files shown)


Packs/Kafka/Integrations/Kafka_V2/Kafka_V2.py
```diff
@@ -178,11 +178,15 @@ def print_topics(client):
         for topic in kafka_topics:
             partitions = []
             for partition in topic.partitions.values():
-                partitions.append({
-                    'ID': partition.id,
-                    'EarliestOffset': partition.earliest_available_offset(),
-                    'OldestOffset': partition.latest_available_offset()
-                })
+                partition_output = {'ID': partition.id}
+                try:
```
glicht, line 182: Maybe we should have a param if to perform this logic. It looks like for a system with a large number of topics that can take time.

Itay4:
@kirbles19 please review the release notes and the yml file only

kirbles19:
@Itay4 Done.

PR #7763, opened 2020-06-29
converting unicode response to string

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26007

## Description
Some responses res.json attributes can be none ascii. in this case python2 keeps it in bytes form and then when trying to use it as string it fails.

This seems to be a known issue in this integration since there is already a method that solves this recursivly- `unicode_to_str_recur`


kirbles19:
@hod-alpert Done.

hod-alpert:
@Itay4 The related issue is closed, let's merge this?

PR #7773, opened 2020-06-29
XSOAR mirroring

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24712

## Description
Adding XSOAR Mirroring server pack + XSOAR mirroring client pack

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


ronykoz:
Docs will be added tomorrow + a more thorough explanation about how to configure this with the two packs configuration.
The server will serve as the host and you could install a client pack on all the other instances you want to mirror from

Packs/Cymulate/Integrations/Cymulate/Cymulate.py
```diff
@@ -1,3 +1,4 @@
+
```
glicht, line 1: Why this line?

ronykoz: mistake, removing

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -3595,3 +3595,29 @@ def batch(iterable, batch_size=1):
 
 class DemistoException(Exception):
     pass
+
+
+class IncidentMirror(object):
```
glicht, line 3600: Do we really need this class? It looks to me like a simply function is sufficient. @yaakovi @anara123 what do you think?

ronykoz: We don't really need that, but we can make the return_results work with this object as well, which will make things less confusing for developers

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4047,48 @@ def update_integration_context(context, object_keys=None, sync=True):
...
+
+
+class ObjectMirror(object):
+    def __init__(self, object, entries):
+        self.object = object
+        self.entries = entries
+
+    def extract_for_local(self):
+        if self.object:
+            demisto.info(f'Updating object {self.object["id"]}')
```
anara123, line 4059: I think debug level should be fine

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4047,48 @@ def update_integration_context(context, object_keys=None, sync=True):
...
+class SchemeMappingField:
+    def __init__(self, name, description=''):
+        self.name = name
+        self.description = description
+
+
+class SchemeTypeMapping:
+    def __init__(self, type_name='', fields=None):
+        self.type_name = type_name
+        self.fields = fields if fields else []
```
anara123, line 4072: support also if user provides just simple dictionary of keys and values of field name and field description

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4047,48 @@ def update_integration_context(context, object_keys=None, sync=True):
...
+        self.object = object
+        self.entries = entries
+
+    def extract_for_local(self):
+        if self.object:
+            demisto.info(f'Updating object {self.object["id"]}')
+            return [self.object] + self.entries
+
+
+class SchemeMappingField:
```
anara123, line 4063: I don't need that we need a class for this. Just the function in SchemeTypeMapping will receive `def add_field(self, field_name, field_description):`

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4047,48 @@ def update_integration_context(context, object_keys=None, sync=True):
 
 class DemistoException(Exception):
     pass
+
+
+class ObjectMirror(object):
```
anara123, line 4052: I think maybe to change the name of the class to `GetRemoteDataResponse`

anara123: Similar class we need for inputs.
`parse_get_remote_data_args(args) -> GetRemoteDataArgs`
`GetRemoteDataArgs`

`parse_update_remote_system_args(args) -> UpdateRemoteSystemArgs`
`UpdateRemoteSystemArgs`
and also `UpdateRemoteSystemResponse`

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
ronykoz: in that case we don't really need this object, are you sure it's needed?

anara123: You can remove it. And just have a function that adds new type to SchemaTypes

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4047,114 @@ def update_integration_context(context, object_keys=None, sync=True):
...
+        otherwise throws an Exception
+    :rtype: ``Optional[int]``
+    """
+
+    if arg is None:
+        if required is True:
+            raise ValueError(f'Missing "{arg_name}"')
+        return -1
+
+    if isinstance(arg, str) and arg.isdigit():
```
roysagi, line 4083: if arg contains a millisecond epoch timestamp - this function will return it unchanged but it needs to be converted to seconds for consistency. Same for the "int/float" section lines 4096-4098

yaron-libman:
Reviewed and added docs-approved label.

anara123 (approved):

glicht (approved):
@ronykoz where are we with this?

glicht:
I just noticed that we have an `XSOAR Mirroring Server` and `XSOAR Mirroring Client` pack. I think this is over kill. I would just do one pack for both.

PR #7775, opened 2020-06-29
[code42-636] Include None in type hints for CommandResults object

unparalleled-js:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
PyCharm gives warnings when trying to use `None` for `CommandResults` arguments, giving the impression that it is not allowed. However, when observing `CommandResults` object in code, it seems to be allowed based on how it handles those situations. Therefore, the type hints should be relaxed to permit `None`. This PR adds the necessary changes to the type hints for this to happen.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ x] 5.5.0

   - [ ] Yes
       - Further details:
   - [x ] No

## Must have
- [ n/a] Tests
- [ x] Documentation 

## Demisto Partner?
- [x ] The title must be in the following format: **[YOUR_PARTNER_ID] short description**

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -2557,25 +2557,25 @@ class CommandResults:
     """
     CommandResults class - use to return results to warroom
 
-    :type outputs_prefix: ``str``
+    :type outputs_prefix: ``str`` or ``None``
     :param outputs_prefix: should be identical to the prefix in the yml contextPath in yml file. for example:
             CortexXDR.Incident
 
-    :type outputs_key_field: ``str``
+    :type outputs_key_field: ``str`` or ``None``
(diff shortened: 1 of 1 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2584,7 +2584,7 @@ class CommandResults:
     """
     def __init__(self, outputs_prefix=None, outputs_key_field=None, outputs=None, indicators=None, readable_output=None,
                  raw_response=None):
-        # type: (str, str, object, list, str, object) -> None
+        # type: (str or None, str or None, object or None, list or None, str or None, object or None) -> None
```
Itay4, line 2587: thanks for the contribution !

better to use `Optional` instead of `or None`, so for example you can use `Optional[str]`

please add release notes as well

PR #7779, opened 2020-06-30
Fixing false green builds

hod-alpert:
Fixed false green builds
fixes: https://github.com/demisto/etc/issues/26062


ShahafBenYakir:
Force merging

PR #7785, opened 2020-06-30
Checking if workflow holding a lock is still running

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Description
In case a build was canceled right after it has locked an integration, the locking attempt will check if the build workflow is still running, if it's not- it will override the lock file.

IMPORTANT: Once inside- this will break any tests that are not pulled from master because the file will now contain 3 fields delimited by ':' instead of two.


Tests/test_content.py
```diff
@@ -1207,6 +1209,32 @@ def safe_lock_integrations(test_timeout: int,
     return locked
 
 
+def workflow_still_running(workflow_id: str) -> bool:
+    """
+    This method takes a workflow id and checks if the workflow is still running
+    If given workflow ID is the same as the current workflow, will simply return True
+    else it will query circleci api for the workflow and return the status
+    Args:
+        workflow_id: The ID od the workflow
```
ShahafBenYakir, line 1218:
```suggestion
        workflow_id: The ID of the workflow
```

PR #7788, opened 2020-06-30
Update metadata dependencies

liorblob:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25951
fixes: https://github.com/demisto/etc/issues/25441

## Description
Merge user dependencies and generated dependencies 
Add core packs mandatory dependency validation


Tests/Marketplace/marketplace_services.py
```diff
@@ -1451,3 +1448,16 @@ def get_higher_server_version(current_string_version, compared_content_item, pac
...
     finally:
         return higher_version_result
+
+
+def set_pack_dependencies(pack_name, user_metadata, packs_dependencies_mapping):
+    pack_dependencies = packs_dependencies_mapping.get(pack_name, {}).get('dependencies', {})
+    if pack_name in GCPConfig.CORE_PACKS_LIST:
+        mandatory_dependencies = [dependency for dependency in pack_dependencies
+                                  if dependency.get('mandatory', False) is True]
+        if mandatory_dependencies:
```
yuvalbenshalom, line 1458: we should raise an exception if there is new mandatory dependency for a core pack if it answers the the terms:
1. the new dependency is not in the core packs group
2. the new dependency is not in the user_metadata dependencies.
3. the dependency is mandatory

only then we should raise an error and fail the build.
and it is better if we will print all the new problematic dependencies found and prompt all the list.

Tests/Marketplace/marketplace_services.py
```diff
@@ -1451,3 +1448,16 @@ def get_higher_server_version(current_string_version, compared_content_item, pac
                     f"Additional info:\n {e}")
     finally:
         return higher_version_result
+
+
+def set_pack_dependencies(pack_name, user_metadata, packs_dependencies_mapping):
```
IkaDemisto, line 1453: It can be part of Pack class

Tests/Marketplace/marketplace_services.py
IkaDemisto: ➕ 1️⃣

IkaDemisto: ➕ 1️⃣

liorblob:
Modified per https://github.com/demisto/etc/issues/26149#issuecomment-653850297

PR #7792, opened 2020-07-01
merge two content zip files from different builds

barchen1:
## Status
- [x] In Progress

## Related Issues
related: https://github.com/demisto/etc/issues/24383

## Description
This script gets 2 content_new.zip files and merge them into one.
It takes one zip file from the current running build and the other one from google cloud according to the given branch name.

## Must have
- [ ] Tests
- [ ] Documentation


Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def main(argv):
+    if len(argv) < 2:
+        print("")
+        sys.exit(1)
+    feature_branch_name = argv[0]
+    master_content_new_zip_path = argv[1]
+    if len(argv) == 3:
+        gc_service_account = argv[2]
```
ronykoz, line 155: use argparse

Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+    build_num = get_last_successful_build_num(feature_branch_name)
+    if not build_num:
+        print("Couldn't find successful build in this branch")
+    current_feature_content_zip_file_path = f'/content/{feature_branch_name}/{build_num}/0/content_new.zip'
+    extract_destination_path = f'{ARTIFACTS_PATH}/feature_content_new_zip'
+    new_feature_content_zip_file_path = download_zip_file_from_gc(current_feature_content_zip_file_path, extract_destination_path)
```
ronykoz, line 161: extract to another function

Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Returns:
+        Last successful build number of the given branch
+    """
+    recent_successful_builds = get_recent_builds_data_request(feature_branch_name)
+    last_successful_build_num = recent_successful_builds[0]['build_num']
+    return last_successful_build_num
+
+
+def download_zip_file_from_gc(current_feature_content_zip_file_path, extract_destination_path):
```
ronykoz, line 102:
```suggestion
def download_zip_file_from_gcp(current_feature_content_zip_file_path, extract_destination_path):
```

Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+
+    feature_branch_content_dir = z.ZipFile(feature_branch_content_zip_file_path, 'r')
+    master_content_zip = z.ZipFile(master_branch_content_zip_file_path, 'w')
+    feature_branch_content_dir.extractall(feature_branch_content_zip_file_path)
+    feature_content_zip_files = os.listdir(feature_branch_content_zip_file_path)
+
+    for file_name in feature_content_zip_files:
+        if file_name not in FILES_TO_REMOVE_LIST:
+            master_content_zip.write(f'{feature_branch_content_dir}/{file_name}')
+    master_content_zip.close()
```
ronykoz, line 145: close this one as well feature_branch_content_dir

Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+from google.cloud import storage
+import google.auth
+import warnings
+# disable insecure warnings
+requests.packages.urllib3.disable_warnings()
+
+
+ACCEPT_TYPE = "Accept: application/json"
+CONTENT_API_URI = "https://circleci.com/api/v1/project/demisto/content"
+FILES_TO_REMOVE_LIST = ['content-descriptor.json', 'doc-CommonServer.json', 'doc-howto.json']
```
ronykoz, line 14: @yaakovi you are better familiar with this then i am, anything else we should add here?

ronykoz: @barchen1 i think that maybe we should include the old `'doc-CommonServer.json', 'doc-howto.json'` too, as they will be needed for old versions

Utils/merge_content_new_zip.py
```diff
@@ -0,0 +1,170 @@
...
+    build_num = get_last_successful_build_num(feature_branch_name)
+    if not build_num:
+        print("Couldn't find successful build in this branch")
+    current_feature_content_zip_file_path = f'/content/{feature_branch_name}/{build_num}/0/content_new.zip'
+    extract_destination_path = f'{ARTIFACTS_PATH}/feature_content_new_zip'
+    new_feature_content_zip_file_path = download_zip_file_from_gc(current_feature_content_zip_file_path, extract_destination_path)
+    if new_feature_content_zip_file_path:
+        merge_zip_files(master_content_new_zip_path, new_feature_content_zip_file_path)
+    else:
+        print('error')
```
ronykoz, line 165: please add descriptive message

Utils/merge_content_new_zip.py
yaakovi: special files:
* `content-descriptor.json` take only the new one (need to think of a way to include the asset ID (build number) of the secondary zips
* `doc-CommonServer.json` - take only the new (no reason to support the old)
* `doc-howto.json` - take only the new one (no reason it will be changed in older versions)
* `reputations.json` - there should only be one, need to merge both of them (in case there is an update)

PR #7801, opened 2020-07-01
Upload pack - skipped not updated hash commit

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Description
In case that commit hash cannot be found, pack upload is skipped.

### DIFF
--- a/Tests/Marketplace/upload_packs.py
+++ b/Tests/Marketplace/upload_packs.py
@@ -480,7 +480,8 @@ def print_packs_summary(packs_list):
     successful_packs = [pack for pack in packs_list if pack.status == PackStatus.SUCCESS.name]
     skipped_packs = [pack for pack in packs_list if
                      pack.status == PackStatus.PACK_ALREADY_EXISTS.name
-                     or pack.status == PackStatus.PACK_IS_NOT_UPDATED_IN_RUNNING_BUILD.name]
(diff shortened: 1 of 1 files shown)


Tests/Marketplace/upload_packs.py
```diff
@@ -480,7 +480,8 @@ def print_packs_summary(packs_list):
     successful_packs = [pack for pack in packs_list if pack.status == PackStatus.SUCCESS.name]
     skipped_packs = [pack for pack in packs_list if
                      pack.status == PackStatus.PACK_ALREADY_EXISTS.name
-                     or pack.status == PackStatus.PACK_IS_NOT_UPDATED_IN_RUNNING_BUILD.name]
+                     or pack.status == PackStatus.PACK_IS_NOT_UPDATED_IN_RUNNING_BUILD.name
+                     or pack.status == PackStatus.FAILED_DETECTING_MODIFIED_FILES.name]
```
yuvalbenshalom, line 484: FAILED_DETECTING_MODIFIED_FILES happens only if there is now hash? and if so, what will update the hash eventually?

IkaDemisto: @yuvalbenshalom  It happens when not updated build is trying to search more updated build hash. The updated hash will remain. The build should not fail in this case

PR #7806, opened 2020-07-02
Improving log prints

hod-alpert:
When there are no integrations to lock- print nothing about locks


Tests/test_content.py
```diff
@@ -1192,11 +1192,12 @@ def safe_lock_integrations(test_timeout: int,
...
-        f'Attempting to lock integrations {integration_names}, with timeout {test_timeout}',
-        print,
-        thread_index,
-        include_timestamp=True)
+    if integration_names:
+        prints_manager.add_print_job(
+            f'Attempting to lock integrations {integration_names}, with timeout {test_timeout}',
+            print,
+            thread_index,
+            include_timestamp=True)
```
Itay4, line 1200: maybe add here an `else` and print that no integration were found to lock

Itay4:
build fails on non related test
merging

PR #7817, opened 2020-07-02
Feed office365: added the tags parameter

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22895

## Minimum version of Demisto
- [ ] 5.5.0

   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
+++ b/Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
@@ -180,13 +180,13 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
                     indicator_mapping_fields["office365required"] = item.get('required')
                 if item.get('notes'):
                     indicator_mapping_fields["description"] = item.get('notes')
+                indicator_mapping_fields['tags'] = client.tags
 
                 indicators.append({
                     'value': value,
                     'type': type_,
-                    "tags": client.tags,
                     'rawJSON': raw_data,
-                    'fieget_indicators_commandlds': indicator_mapping_fields
+                    'fields': indicator_mapping_fields
                 })
 
     return indicators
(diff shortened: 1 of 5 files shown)


Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -37,21 +37,23 @@ def build_urls_dict(regions_list: list, services_list: list, unique_id) -> List[
...
+class Client:
     """
     Client to use in the Office 365 Feed integration. Overrides BaseClient.
     Office 365 IP address and URL web service announcement:
     https://docs.microsoft.com/en-us/office365/enterprise/managing-office-365-endpoints?redirectSourcePath=%252fen-us%252farticle%252fmanaging-office-365-endpoints-99cab9d4-ef59-4207-9f2b-3728eb46bf9a#webservice
     https://techcommunity.microsoft.com/t5/Office-365-Blog/Announcing-Office-365-endpoint-categories-and-Office-365-IP/ba-p/177638
     """
 
-    def __init__(self, urls_list: list, insecure: bool = False):
+    def __init__(self, urls_list: list, insecure: bool = False, tags: Optional[list] = None):
```
Itay4, line 48: lets stay consistent with `Dict` and `List` instead of `dict` and `list`

Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -180,10 +182,11 @@ def fetch_indicators(client: Client, indicator_type_lower: str, limit: int = -1)
...
                 indicators.append({
-                    "value": value,
-                    "type": type_,
-                    "rawJSON": raw_data,
-                    "fields": indicator_mapping_fields
+                    'value': value,
+                    'type': type_,
+                    "tags": client.tags,
+                    'rawJSON': raw_data,
+                    'fieget_indicators_commandlds': indicator_mapping_fields
```
Itay4, line 189: what ?

Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
jochman: But they are not the same things.
dict means it returns a dict while Dict requires to fill the structure of the dict itself.
`var: dict`
vs
`var: Dict[str, str]`

And it's just not clear to do things like:
`var: List[Any]`, it's confusing

PR #7844, opened 2020-07-05
Lastline v2 - Fixed the detailed description

bakatzir:
## Status
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/26135

   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/Lastline/Integrations/Lastline_v2/Lastline_v2_description.md
+++ b/Packs/Lastline/Integrations/Lastline_v2/Lastline_v2_description.md
@@ -1,8 +1,8 @@
 Provides threat analysts and incident response teams with the advanced
-  malware isolation and inspection environment, needed to safely execute advanced
-  malware samples and understand their behavior. 
-API key and API token are generate by lastline and given at the subscription time.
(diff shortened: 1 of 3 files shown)


Packs/Lastline/Integrations/Lastline_v2/Lastline_v2_description.md
```diff
@@ -1,8 +1,8 @@
...
-  malware samples and understand their behavior. 
-API key and API token are generate by lastline and given at the subscription time.
+malware isolation and inspection environment, needed to safely execute advanced
+malware samples and understand their behavior. 
+API key and API token are generate by Lastline and given at the subscription time.
 
-###How to use `threshold` parameter.
+### How to use the `threshold` parameter.
 by default `threshold` is set to 70. it means that for every file with severity over 70
 the Dbot score will be high. you can change it as you want.
```
amshamah419, line 8: Let's fix this as well.

```suggestion
By default, the `threshold` parameter is set to 70. This means that for every file with a severity of over 70
the Dbot score will be set to *high*. This parameter is configurable.
```

PR #7873, opened 2020-07-06
Instance test fix

ShahafBenYakir:
## Description
Increased TTL of instance test to 240 minutes


yaakovi:
> have you updated the "environment context" in circle configuration as well?
> this update will only fix the manual trigger

updated it for you

PR #7876, opened 2020-07-06
Panos threat (#7746)

bakatzir:
* Added basic requests_mock fixture

* Updated to add command and unit testing

* Further improved yml files and fixed lint issues

* Added basic requests_mock fixture

* Updated to add command and unit testing

* Further improved yml files and fixed lint issues

* Added release notes

* Updated to add outputs and slightly better results calls

* Rectified return val for inner function

* Update Packs/PAN-OS/Integrations/Panorama/Panorama_test.py

* Update Packs/PAN-OS/Integrations/Panorama/Panorama_test.py

* Add release notes and fix whitespace issues

Co-authored-by: Bar Katzir <37335599+bakatzir@users.noreply.github.com>

## Status
- [x] Ready

   - [x] No

internal contrib


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4718,6 +4718,73 @@ def panorama_add_static_route_command():
...
+    target = str(demisto.args()['target']) if 'target' in demisto.args() else None
+    result = panorama_get_predefined_threats_list(target)
+    demisto.results(fileResult('predefined-threats.json', json.dumps(result['response']['result']).encode('utf-8')))
+
+
+def panorama_block_vulnerability():
+    """
+    Ovverride a vulnerability signature such that it is in block mode
+    """
+    threatid = demisto.args()['threat_id']
```
yuvalbenshalom, line 4769:
```suggestion
    threatid = demisto.args().get('threat_id')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4718,6 +4718,73 @@ def panorama_add_static_route_command():
...
+    vulnerability_profile = demisto.args()['vulnerability_profile']
```
yuvalbenshalom, line 4770:
```suggestion
    vulnerability_profile = demisto.args().get('vulnerability_profile')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4718,6 +4718,73 @@ def panorama_add_static_route_command():
...
+    result = http_request(
+        URL,
+        'GET',
+        params=params
+    )
+    return result
+
+
+def panorama_get_predefined_threats_list_command():
+    target = str(demisto.args()['target']) if 'target' in demisto.args() else None
```
yuvalbenshalom, line 4760: why not just use get? dose it make a difference to put here None and not ''?
```suggestion
    target = str(demisto.args().get('target',''))
```

bakatzir: Yes. it needs to be a None if not existing and this is the convention already in the integration code.

bakatzir: example:
Screen Shot 2020-07-07 at 9 30 29

bakatzir:
fails on non related `PAN-OS Create Or Edit Rule Test (Mock Disabled)` which fails on nightly.
force merging.

PR #7883, opened 2020-07-06
Old branch script reputations add on

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/24383

## Description
adding reputations.json support for feature branch script


Utils/old_content_branch.py
```diff
@@ -190,6 +206,21 @@ def edit_all_packs(new_to_version):
...
+def edit_reputations_json(new_to_version):
+    print("Updating reputations.json\n")
+    rep_json_path = "Packs/NonSupported/IndicatorTypes/reputations.json"
+    with open(rep_json_path, 'r') as f:
+        rep_content = ujson.load(f)
+
+    for reputation in rep_content.get('reputations', []):
+        if parse_version(reputation.get('toVersion', "99.99.99")) > parse_version(new_to_version) > \
+                parse_version(reputation.get("fromVersion", "0.0.0")):
+            reputation['toVersion'] = new_to_version
```
ronykoz, line 218: are we adding the fromversion of this in the content creator? i don't think it's needed here @yaakovi WDYT?

yaakovi: you can ignore the `fromversion`.
this can be relevant only for `toversion` lower than 4.0.0

ronykoz:
@barchen1 can you take care of the last comment?

Utils/old_content_branch.py
barchen1: done

PR #7897, opened 2020-07-07
Small Tweaks to Mocking Process

avidan-H:
## Status
- [x] Ready

## Description
1. Changed the `run_and_record` method in `test_content.py` to only try and clean a mock file if the recording was successful - otherwise, there is no point.
2. Updated the `start` method of `MITMProxy` class to copy over a pre-existing `problematic_keys.json` (if there is one for that test) to the temporary directory in which the mock recording and log files for a given test are written to. The idea here is that `problematic_keys.json` files should only ever get added to, never freshly overwritten. That way, if someone has manually added problematic keys for a test (presumably keys that `timestamp_replacer.py` does not successfully detect itself), those keys will not get overwritten in a case where for example, the test playbook has been updated with additional commands and therefore is rerecorded the first time it runs after the update.

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No

### DIFF
--- a/Tests/mock_server.py
+++ b/Tests/mock_server.py
@@ -364,11 +364,13 @@ class MITMProxy:
             self.repo_folder, get_folder_path(playbook_id), 'problematic_keys.json'
         )
         current_problem_keys_filepath = os.path.join(path, get_folder_path(playbook_id), 'problematic_keys.json')
-        # copy the `problematic_keys.json` for the test to current temporary directory if it exists that way
-        # previously recorded or manually added keys will only be added upon and not wiped with an overwrite
-        silence_output(
-            self.ami.check_call, ['mv', repo_problem_keys_filepath, current_problem_keys_filepath], stdout='null'
-        )
+
+        # when recording, copy the `problematic_keys.json` for the test to current temporary directory if it exists
+        # that way previously recorded or manually added keys will only be added upon and not wiped with an overwrite
+        if record:
+            silence_output(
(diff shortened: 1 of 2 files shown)


Tests/mock_server.py
```diff
@@ -364,6 +364,9 @@ def start(self, playbook_id, path=None, record=False, thread_index=0, prints_man
             self.repo_folder, get_folder_path(playbook_id), 'problematic_keys.json'
         )
         current_problem_keys_filepath = os.path.join(path, get_folder_path(playbook_id), 'problematic_keys.json')
+        silence_output(
```
bakatzir, line 367: let's add the explanation in a comment, as it is not easy to interpret from code flow.

bakatzir: This way the `problematic_keys.json` will not be empty in the pwned v2 case that we saw, so we will not reach the error. right?

avidan-H: We wouldn't reach it for two reasons.
A. It wouldn't be empty as you stated.
B. Now we only call `proxy.clean_mock_file` if `run_test_logic` succeeds - so in this particular case, with my update, `proxy.clean_mock_file` wouldn't be called at all.

avidan-H:
Not sure why, but a CircleCI build isn't being triggered automatically. I'm going to trigger it manually.

PR #7898, opened 2020-07-07
fix associate indicator with url indicator and Update Threat Connect description

esharf:
## Status
- [ ] In Progress
- [x] Ready

fixes: https://github.com/demisto/etc/issues/25846

the bug was that the URL should be a decoded URL.
so the API request for `http://test.com` should be `/v2/indicators/urls/http%3A%2F%2Ftest.com/groups/<group_type>/<group_id>`


kirbles19:
Looks fine.

PR #7900, opened 2020-07-07
SDP Close command fix and Polling PB

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Description
Fixing a problem in the close command and adding a polling playbook.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/ServiceDeskPlus/Integrations/ServiceDeskPlus/ServiceDeskPlus.py
```diff
@@ -673,18 +673,21 @@ def close_request_command(client: Client, args: dict) -> Tuple[str, dict, Any]:
...
     """
     request_id = args.get('request_id')
-    params = {
+    closure_info: Dict[str, Any] = {
+        "requester_ack_resolution": args.get('requester_ack_resolution', 'false'),
+        "requester_ack_comments": args.get('requester_ack_comments', ''),
+        "closure_comments": args.get('closure_comments', ''),
+    }
+    if args.get('closure_code'):
+        closure_info["closure_code"] = {'name': args.get('closure_code')}
```
yuvalbenshalom, line 682: will this work if close code is empty? do we know if it is mandatory or not?

darkushin: It isn't mandatory. In case the user didn't choose any of the listed options it is an empty string and it shouldn't be added to the query params because it causes it to fail. The 'if' checks for the case that the user didn't choose any option and it won't be True for an empty string.

kirbles19:
@yaron-libman See playbook.
@yuvalbenshalom integration is reviewed, looked good.

PR #7912, opened 2020-07-08
First template of the new code structure

ronykoz:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
relates to https://github.com/demisto/etc/issues/24043

## Description
This is the initial blueprint of how i think we should move forward with the issue, please note that this is **super minimal and still work in progress**, i don't want to move further without your approval.

No need to make comments on code convention I'm aware i have some problems there, i just want to get an approval for the general concept first

The outputs should be in a different yml file generated by the `demisto-sdk json-to-outputs`

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -15,6 +15,7 @@ import sys
 import time
 import traceback
 from random import randint
+from typing import List
 import xml.etree.cElementTree as ET
 from collections import OrderedDict
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4029,3 +4035,157 @@ def update_integration_context(context, object_keys=None, sync=True):
 
 class DemistoException(Exception):
     pass
+
+
+class IndicatorsObject:
+    def __init__(self, indicators: list):
+        self.indicators: list = indicators
```
Itay4, line 4042: this is python3 syntax

PR #7917, opened 2020-07-08
PAN-OS - panorama-show-location-ip - Get the location of an IP address.

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24651

## Minimum version of Demisto
- [x] 3.5.0

   - [x] No

## Must have
- [x] Tests - tested both on fw and panorama
- [x] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -5333,8 +5333,8 @@ def panorama_show_location_ip_command():
     ip_address = demisto.args().get('ip_address')
     result = panorama_show_location_ip(ip_address)
 
-    if result['response']['@status'] != 'success':
-        raise Exception(f'Failed to successfully show the location of the specified ip {ip_address}.')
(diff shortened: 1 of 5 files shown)


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5310,6 +5310,57 @@ def panorama_device_reboot_command():
...
+
+
+def panorama_show_location_ip_command():
+    """
+    Check location of a specified ip address
+    """
+    ip_address = demisto.args().get('ip_address')
+    result = panorama_show_location_ip(ip_address)
+
+    if result['response']['@status'] != 'success':
```
yuvalbenshalom, line 5336: maybe cache that @status is there before?

bakatzir:
fails on: `PAN-OS Create Or Edit Rule Test` which fails on nightly.

amshamah419:
@bakatzir Docs look good

bakatzir:
force merging due to the non related failure.

PR #7921, opened 2020-07-08
Tanium Threat Response fix

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Description
Fix for Tanium Threat Response http request (revealed in test failure) - added 30 secs timeout and retries for tasks

   - [ ] Yes
       - Further details:
   - [x] No


Packs/TaniumThreatResponse/Integrations/TaniumThreatResponse/TaniumThreatResponse.py
```diff
@@ -14,6 +14,7 @@
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 ''' GLOBALS/PARAMS '''
+REQUEST_TIMEOUT = 60
```
ShahafBenYakir, line 17: 60 seconds is a lot. The timeout is 10s? Let's try it with 20s, it could be very painful to customers.

Packs/TaniumThreatResponse/ReleaseNotes/1_0_2.md
```diff
@@ -0,0 +1,5 @@
+<!--
+#### Integrations
+##### Tanium Threat Response
+test playbook fix
+-->
```
ShahafBenYakir, line 5: If we will add this timeout exceeds we will need to add this here.

PR #7922, opened 2020-07-08
[Fix] Prisma Access issue when fetching user

DeanArbel:
## Status
- [x] In Hold - (waiting customer validation)

## Related Issues
fixes: https://github.com/demisto/etc/issues/24498

## Description
Fixed an issue in `prisma-access-active-users` where the command would fail from time to time.

   - [x] No


Packs/PrismaAccess/ReleaseNotes/1_0_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Prisma Access
+Fixed an issue where ssh commands would fail from time to time.
```
amshamah419, line 4:
```suggestion
Fixed an issue where SSH commands would fail periodically.
```

Itay4:
@DeanArbel what is the status here?

DeanArbel:
This solution did not work, so we'll be implementing this command using another method.
This PR can be deleted.

PR #7926, opened 2020-07-08
[RiskIQ-127] RiskIQ Security Intelligence Services Feed Integration

crestdatasystems:
#### new Integration
 - A PassiveTotal with Security Intelligence Services Feed provides you with newly observed Domain, Host, Malware, Phishing, and Scam Blacklist with options for Daily and Hourly ingestion available.
 - Support indicators with custom indicator fields.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

## Demisto Partner?
- [x] The title must be in the following format: **[YOUR_PARTNER_ID] short description**


hod-alpert:
Hey, Great work!!

According to the description:
```
Existing clients can use their API keys in order to enable the functionality within XSOAR. If API keys are unknown or have been misplaced, clients can contact their account representative directly or send an email to **support@riskiq.com**.
```

However, I could not find a way to use API key in this integration, am I missing something?

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
```diff
@@ -0,0 +1,540 @@
...
+    :param indicator: Indicator dictionary.
+    :return: Dict of fields.
+    """
+    fields: Dict[str, Any] = {
+        'service': 'Passive Total',
+    }
+    try:
+        if feed_type in ['host', 'domain']:
+            if indicator.get('Timestamp'):
+                fields['firstseenbysource'] = datetime.fromtimestamp(int(indicator['Timestamp'])).isoformat() + 'Z'
```
hod-alpert, line 330: Can you show an indicator data example? how do you know that this date is in UTC?

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
```diff
@@ -0,0 +1,540 @@
...
+
+    # Validate the provided limit
+    validate_limit(limit)
+
+    # Validate the provided feed type.
+    validate_feeds([feed_type])
+
+    # Retrieving indicators from fetch indicators command.
+    indicators_list: List[Dict[str, Any]] = []
+    for indicators in fetch_indicators_command(client, [feed_type], limit=limit, search=search):
```
hod-alpert, line 477: The `get-indicators` command should be independent and given the same parameters- return the same results.

I cannot test it, but I believe that's not the case here if you're using the ** fetch_indicators_command** which starts to get the indicators after the `cached_key`.

So If you call `fetch_indicators_command` before a fetch was made or after a fetch was made, the last cache will be different and the resulting indicator as well, am I wight?

crestdatasystems: The current approach is to fetch the last uploaded S3 bucket object for an indicator type and show limited (at max 1000) indicators. 
The cache is only used to limit the search for bucket objects, whether the cache is there or not, always the last uploaded bucket object will only be fetched. Our understanding from docs was that this command is to see limited/sample indicators in the war-room.

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
crestdatasystems: These timestamps are in epoch format in data. To be sure that parsed time is in UTC, we are changing to "datetime.utcfromtimestamp"

crestdatasystems:
> Hey @crestdatasystems , Great work!!
> 
> According to the description:
> 
> ```
> Existing clients can use their API keys in order to enable the functionality within XSOAR. If API keys are unknown or have been misplaced, clients can contact their account representative directly or send an email to **support@riskiq.com**.
> ```
> 
> However, I could not find a way to use API key in this integration, am I missing something?
> Does these buckets belong to RiskIQ? how can I get access to them?

Hi @, Thanks for the review. Yes, they belong to RiskIQ, and the customers will have to get API keys (S3 access/secret keys) from RiskIQ, if they don't already have.

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
hod-alpert: I'd rather make this an aware datetime object and this why you can remove the 'Z' and completely trust the `isoformat` method.

Try: 
```
datetime.fromtimestamp(int(indicator['Timestamp']), ,datetime.timezone.utc).isoformat()
```
and don't forget to import the `timezone` from datetime (i.e `from datetime import datetime, timezone`)

hod-alpert: Same goes to all date parsing

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
hod-alpert: So if I'll fetch indicators twice while there is no change in the buckets, will both executions return the same output?

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
crestdatasystems: Thanks @hod-alpert, we will make changes to all date-time parsing to have aware Datetime object as you suggested.

Packs/SecurityIntelligenceServicesFeed/Integrations/SecurityIntelligenceServicesFeed/SecurityIntelligenceServicesFeed.py
crestdatasystems: Yes, It will produce same result.

crestdatasystems:
Hello @hod-alpert, we are planning to make changes in Datetime parsing as you suggested. So, do we need to make another PR for that?
and while testing this datetime parsing changes we encountered another small change that also needs to be included.

hod-alpert:
Thanks @crestdatasystems You can open another PR on the same branch `demisto:contrib/crestdatasystems_contrib/crestdatasys_riskiq_sis_feed_integration`
I'll test it too and see what else is missing

crestdatasystems:
Hello @hod-alpert, this is the new PR for timezone related changes as you suggested (https://github.com/demisto/content/pull/8065).

hod-alpert:
Great work @crestdatasystems !

One last thing I noticed while checking it:

If the query time is too long (7 days for example) the fetch execution takes too long (30 min) and raises a timeout exception from the server side, this because it tries to fetch all indicators in one time (and there can be quite plenty of those)

We need a way to make it return indicators to XSOAR in batches. For example, make it return indicators from one file at a time and then promote the file cache key.

Since it runs all the time it just might take more time to sync all indicators, but in the long run once it's synced it'll just update the new ones.

Can you take care of that? or you want me to do it?

If you can take can you also rebase on master?

PR #7934, opened 2020-07-09
Enable MS Graph Listener/Mail to use htmlBody arg

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25820

## Description
When automated messages are sent and the body type is supposed to be HTML, the message body will default to use the argument `body` which is not an HTML formatted body. 

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


amshamah419:
@yuvalbenshalom It's currently failing on Gmail quota issues

PR #7909, opened 2020-07-07
Fixed DT syntax error

aburt-demisto:
## Status
- [] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/26310

## Description
DT Syntax modified in NetOps playbook that is causing failure of status ID captures.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

PR #7938, opened 2020-07-09
Fixed DT syntax error

liorblob:
Follow up for https://github.com/demisto/content/pull/7909
fixes: demisto/etc#26310


Packs/PAN-OS/ReleaseNotes/1_4_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Playbooks
+- __NetOps - Firewall Version and Content Upgrade__
```
amshamah419, line 3: Should be like the following:
`##### NetOps - Firewall Version and Content Upgrade`

Packs/PAN-OS/ReleaseNotes/1_4_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Playbooks
+- __NetOps - Firewall Version and Content Upgrade__ 
+  - Fixed DT syntax issues.
```
amshamah419, line 4: Remove the leading space.

PR #7837, opened 2020-07-03
Update sane doc

agamm:
## Status
- [X] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
Related to the following issues:

demisto/etc#23510
demisto/etc#18250
PRs:

demisto/sane-doc-reports#72
demisto/sane-doc-reports#68
demisto/sane-doc-reports#69
demisto/sane-doc-reports#70
demisto/sane-doc-reports#71
https://github.com/demisto/dockerfiles/pull/2169

## Description
Fixes several sane-doc-reports bug and adds features:
- Fixes the svg image support for images
- Adds the ability to add customer logos to the headers
- Fixes the empty table behavior (no error now).

## Screenshots
image
image
image

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?
- [ ] The title must be in the following format: **[YOUR_PARTNER_ID] short description**

PR #7944, opened 2020-07-09
Update the releasenotes accordingly

agamm:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
https://github.com/demisto/content/pull/7837

## Description
Adds missing release notes for the sane-pdf-reports.
Relating to: https://github.com/demisto/etc/issues/25972


yuvalbenshalom:
@agamm actually, on a second thought, since only the current version pack RN can be updated, and this one is trying o fix an old one, this will not have effect in the bucket.
I think we can close this one.

PR #7945, opened 2020-07-10
fix slack notifications for nightly

hod-alpert:
## Description
Added slack notifications to server 6_0 (instead of tests 5_5)
Added nightly context to unittests job


Itay4:
build fails on non related tests
merging

PR #7963, opened 2020-07-12
DeHashed email command

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26210

## Description
Adding the !email command to the DeHashed integration.

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/DeHashed/Integrations/DeHashed/DeHashed.py
```diff
@@ -225,6 +236,47 @@ def dehashed_search_command(client: Client, args: Dict[str, str]) -> tuple:
...
+                {
+                    'Indicator': email_address[0],
+                    'Type': 'email',
+                    'Vendor': INTEGRATION_CONTEXT_BRAND,
+                    'Score': 0
+                }
+        }
+        return "No matching results found", context, None
+    else:
+        default_dbot_score_email = 2 if client.email_dbot_score == 'SUSPICIOUS' else 3
```
yuvalbenshalom, line 265: you need to check here if the exact mail is found, because we are using `contained`.
for example:
if you search for alice@gmail.com, you may find that prefix_alice@gmail.com will be found. but it is not the same mail.

please add a check here that the exact mail is in the results and please add a unit test for that.
return 0 if not.

kirbles19:
@darkushin Done.

PR #7965, opened 2020-07-12
No need to test integrations for the third time before running test p…

hod-alpert:
…laybook

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/23899


hod-alpert:
It's also defaults to True in `__create_integration_instance` and it's configured in `content-test-conf` if anyone wants to overrun it

PR #7970, opened 2020-07-12
[Fix] ElasticSearch API login issue (api key)

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26374

## Description
Improved API key handling - added a fallback that ensures API key is used when selected by user.

   - [x] No


Packs/Elasticsearch/Integrations/Elasticsearch_v2/Elasticsearch_v2_test.py
```diff
@@ -424,17 +424,6 @@ def test_elasticsearch_builder_called_with_username_password(mocker):
     assert es_mock.call_args[1].get('api_key') is None
 
 
-@patch("Elasticsearch_v2.API_KEY_ID", "demisto")
```
DeanArbel, line 427: Removed due to a change in the way the file is loaded - this causes issues testing this specific behavior, and as such the test isn't helpful.

DeanArbel, line 427: This test became redundant due to the change in the way the integration sets up API key

DeanArbel:
@amshamah419 The release notes script validator asked me add RN for ElasticsearchFeed in Packs/Elasticsearch/ReleaseNotes/1_0_2.md, which is unrelated to ElasticsearchFeed, and the demisto-sdk update-release-notes added it successfully. fyi.

amshamah419:
The validation is probably looking for the integration name in the file paths. If you pull from master, it might clear it.

PR #7994, opened 2020-07-14
SplunkPy fetch limit fix

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26385

## Description
Update the fetch limit parameter in case this field is left empty in the instance configurations.

   - [ ] Yes
       - Further details:
   - [x] No


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -21,7 +21,7 @@
 params = demisto.params()
 SPLUNK_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S"
 VERIFY_CERTIFICATE = not bool(params.get('unsecure'))
-FETCH_LIMIT = int(params.get('fetch_limit', 50))
+FETCH_LIMIT = int(params.get('fetch_limit')) if params.get('fetch_limit') else 50
```
guykeller, line 24: This does not change the functionality

Packs/SplunkPy/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "SplunkPy",
     "description": "Run queries on Splunk servers.",
     "support": "xsoar",
-    "currentVersion": "1.0.3",
+    "currentVersion": "1.1.0",
```
guykeller, line 5: This isn't a very significant change, isn't this a revision?

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
darkushin: @guykeller If the fetch_limit param is left empty in the instance configurations then an empty string is passed as the parameter. When doing params.get('fetch_limit', 50) where fetch_limit = "" the value is set to "" and not 50, because the default replacement to 50 would happen if it was None and not an empty string. On the other hand, checking if "" is False, hence causing FETCH_LIMIT to be 50.

kirbles19:
@darkushin Done

PR #8005, opened 2020-07-14
install relevant server vesrions

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26438

## Description
Currently server versions 4.1.0 to 5.0.0 are installed only if there is a relevant test that can run in them.

This PR changes them to be installed even if there is no test that can run in them as long as the modified files in the branche's commits can run in this version.

If no info could be extracted about the files- will run all versions.

In addition:
* Refactored the usage of id_set.json, conf.json and amu_builds.json so that they will be opened and read only once, and be saved as consts.
* Replaced prints with logging

### DIFF
--- a/Tests/scripts/collect_tests_and_content_packs.py
+++ b/Tests/scripts/collect_tests_and_content_packs.py
@@ -12,6 +12,7 @@ import argparse
 import coloredlogs
 import logging
 from distutils.version import LooseVersion
+from copy import deepcopy
 from typing import Dict, Tuple
 from Tests.Marketplace.marketplace_services import IGNORED_FILES
 import demisto_sdk.commands.common.tools as tools
@@ -21,10 +22,9 @@ coloredlogs.install(level=logging.DEBUG, fmt='[%(asctime)s] - [%(threadName)s] -
 
 
 class TestConf(object):
-    __test__ = False  # pytest will not try to run it just because it has Test prefix
+    __test__ = False  # required because otherwise pytest will try to run it as it has Test prefix
 
-    def __init__(self, conf):
-        #  (dict) -> None
+    def __init__(self, conf: dict) -> None:
 
         self._conf = conf
 
@@ -301,8 +301,8 @@ def collect_tests_and_content_packs(
         catched_scripts,
         catched_playbooks,
         tests_set,
-        id_set=ID_SET,
-        conf=CONF
+        id_set=deepcopy(ID_SET),
+        conf=deepcopy(CONF)
 ):
     """Collect tests for the affected script_ids,playbook_ids,integration_ids.
 
@@ -463,7 +463,7 @@ def check_if_fetch_incidents_is_tested(missing_ids, integration_ids, id_set, con
     return missing_ids, tests_set
 
 
-def find_tests_and_content_packs_for_modified_files(modified_files, conf=CONF, id_set=ID_SET):
+def find_tests_and_content_packs_for_modified_files(modified_files, conf=deepcopy(CONF), id_set=deepcopy(ID_SET)):
     script_names = set([])
     playbook_names = set([])
     integration_ids = set([])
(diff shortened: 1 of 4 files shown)


Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1184,19 +1160,53 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
...
         'Demisto PreGA': True,
         'Demisto Marketplace': True,
-        'Demisto two before GA': is_any_test_runnable(test_ids=tests, server_version=two_before_ga, conf=conf,
-                                                      id_set=id_set),
-        'Demisto one before GA': is_any_test_runnable(test_ids=tests, server_version=one_before_ga, conf=conf,
-                                                      id_set=id_set),
-        'Demisto GA': is_any_test_runnable(test_ids=tests, server_version=ga, conf=conf, id_set=id_set),
+        'Demisto two before GA': is_runnable_in_server_version(from_version, two_before_ga, to_version),
+        'Demisto one before GA': is_runnable_in_server_version(from_version, one_before_ga, to_version),
+        'Demisto GA': is_runnable_in_server_version(from_version, ga, to_version),
```
ronykoz, line 1207: doesn't it mean we don't start these instances if no test is runnable?

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1184,19 +1160,53 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
...
+    Args:
+        all_modified_files_paths: All modified files
+        id_set: the content of the id.set_json
+
+    Returns:
+        The string representation of the lowest from version and highest to version
+    """
+    max_to_version = LooseVersion('0.0.0')
+    min_from_version = LooseVersion('99.99.99')
+    for _, artifacts in id_set.items():
```
Itay4, line 1176:
```suggestion
    for artifacts in id_set.values():
```

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1184,19 +1160,53 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
...
+        for artifact_dict in artifacts:
+            for _, artifact_details in artifact_dict.items():
```
Itay4, line 1178:
```suggestion
            for artifact_details in artifact_dict.values():
```

Tests/scripts/collect_tests_and_content_packs.py
hod-alpert: No, sometimes files are modified and there are no test playbook that can verify them.

In that case we just want to install the relevant version and see that nothing is broken even if there are no tests to run in it.

I'm not checking the from version and to version of the tests, but of all the modified files.

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1184,19 +1160,53 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, two_be
...
+                    to_version = artifact_details.get('toversion')
+                    if from_version:
+                        min_from_version = min(min_from_version, LooseVersion(from_version))
+                    if to_version:
+                        max_to_version = max(max_to_version, LooseVersion(to_version))
+    if max_to_version.vstring == '0.0.0':
+        max_to_version = LooseVersion('99.99.99')
+    if min_from_version.vstring == '99.99.99':
+        min_from_version = LooseVersion('0.0.0')
+    logging.debug(f'modified files are {all_modified_files_paths}')
```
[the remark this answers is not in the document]
hod-alpert: I think it's important to show what are the changed files it takes into consideration, and what are the resulting selected tests and packs.

I think it could help reveal issues with this process if there are some, and that's why it's under debug mode.

PR #8017, opened 2020-07-14
Build Fixes & Tweaks

avidan-H:
## Status
- [x] Ready

## Description
Running the following command,
`./Utils/trigger_content_build_with_time_to_live.sh content_branch_name <CircleCiToken> 420 contrib_name:contrib_branch_name Packs/New_Pack/`
is supposed to trigger a build that will take the contribution and will run a regular content build using the contributor's code/additions. Currently, the build does not actually use the contributor's code/additions for testing on server instances. This PR implements proper handling in our build for this usecase. See this [build](https://app.circleci.com/pipelines/github/demisto/content/19257/workflows/f7a00841-2b11-4a62-a002-7a1410eb84fb) (which was triggered by `./Utils/trigger_content_build_with_time_to_live.sh code42_contrib <circle-token-redacted> 420 code42:download-file Packs/Code42/`) for an example where the contributor's code was not properly fetched for testing on server instances.

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,11 +71,13 @@ references:
   attach_workspace: &attach_workspace
     attach_workspace:
       at: *workspace_root
+
   add_ssh_keys: &add_ssh_keys
     add_ssh_keys:
       fingerprints:
         - "02:df:a5:6a:53:9a:f5:5d:bd:a6:fc:b2:db:9b:c9:47" # disable-secrets-detection
         - "f5:25:6a:e5:ac:4b:84:fb:60:54:14:82:f1:e9:6c:f9" # disable-secrets-detection
+
   prepare_environment: &prepare_environment
     run:
       name: Prepare Environment
@@ -143,9 +145,11 @@ references:
         unzip -q "$CIRCLE_ARTIFACTS/content_test.zip" -d "test_content_dir"
         zip -j "$CIRCLE_ARTIFACTS/all_content.zip" test_content_dir/*
         rm -r test_content_dir
+
   restore_cache: &restore_cache
     restore_cache:
       key: venv-{{ checksum "dev-requirements-py2.txt" }}-{{ checksum "dev-requirements-py3.txt" }}-{{ checksum ".circleci/build-requirements.txt" }}-{{ checksum "package-lock.json" }}
+
   destroy_instances: &destroy_instances
     run:
       name: Destroy Instances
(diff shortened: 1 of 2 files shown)


avidan-H:
I'm adding the `docs-approved` label since there is no documentation to review.

.circleci/config.yml
```diff
@@ -180,12 +184,13 @@ references:
       paths:
         - project
 
-  Secrets: &Secrets
+  secrets: &secrets
```
Itay4, line 187: why did you lower case it? i think it is better shown in the step as `Secrets`

.circleci/config.yml
```diff
@@ -273,21 +284,41 @@ references:
...
     run:
       name: Update Tests step
       when: always
       command: |
         python3 ./Tests/scripts/update_conf_json.py
         cp "./Tests/conf.json" "$CIRCLE_ARTIFACTS/conf.json"
+
+  get_contribution_pack: &get_contribution_pack
+    when:
+      condition: << pipeline.parameters.contrib_branch >>
```
Itay4, line 305: should we use `CONTRIB_BRANCH` here ?

Tests/scripts/prepare_content_packs_for_testing.sh
```diff
@@ -37,10 +37,6 @@ echo "Updating modified content packs in the bucket ..."
 
 if [ ! -n "${NIGHTLY}" ]; then
     CONTENT_PACKS_TO_INSTALL_FILE="./Tests/content_packs_to_install.txt"
-
-  if [ -n "${CONTRIB_BRANCH}" ]; then
```
yaakovi, line 41: why did you remove this?
it was added because the `Collect tests and packs to install` did not recognize it as a changed pack (as this might be a new pack and not part of the repo)

.circleci/config.yml
avidan-H: image
In the step itself, it is capitalized - I only changed the alias to be lowercase in keeping with the format of all of the other aliases.

.circleci/config.yml
avidan-H: We set this env variable in the `environment` alias step - if a job doesn't use this alias for setting the environment, then the variable will not be set. I think it's preferable to use the pipeline parameter since that should always be present.

Tests/scripts/prepare_content_packs_for_testing.sh
avidan-H: Cause this would mistakenly add the contributor's pack name to the `content_packs_to_install.txt` file even when the contributor's pack wasn't pulled (which was the case in our build). The proper way to do this is to pull in the contributor's changes and have those changes detected in the `collect_test_list_and_content_packs` step the same way we do for all of our content.

PR #8029, opened 2020-07-15
Update index in case pack is missing as subfolder 

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Related Issues
related to: https://github.com/demisto/etc/issues/26536

## Description
Update index in case it is missing pack folder for any reason. Partial solution for original issue. 
The race condition between builds will be solved by adding commit hash and in separate PR.

### DIFF
--- a/Tests/Marketplace/marketplace_services.py
+++ b/Tests/Marketplace/marketplace_services.py
@@ -140,6 +140,7 @@ class PackStatus(enum.Enum):
     FAILED_REMOVING_PACK_SKIPPED_FOLDERS = "Failed to remove pack hidden and skipped folders"
     FAILED_RELEASE_NOTES = "Failed to generate changelog.json"
     FAILED_DETECTING_MODIFIED_FILES = "Failed in detecting modified files of the pack"
+    FAILED_SEARCHING_PACK_IN_INDEX = "Failed in searching pack folder in index"
 
 
 class Pack(object):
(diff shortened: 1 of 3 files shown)


IkaDemisto:
@guykeller FYI

Tests/Marketplace/upload_packs.py
```diff
@@ -675,8 +676,14 @@ def main():
             pack.cleanup()
             continue
 
-        # in case that pack already exist at cloud storage path, skipped further steps
-        if skipped_pack_uploading:
+        task_status, exists_in_index = pack.check_if_exists_in_index(index_folder_path)
+        if not task_status:
+            pack.status = PackStatus.FAILED_SEARCHING_PACK_IN_INDEX.name
```
yaakovi, line 681: why is that not inside the pack method?

Tests/Marketplace/marketplace_services.py
```diff
@@ -1239,6 +1240,31 @@ def _search_for_images(self, target_folder):
...
+        task_status, exists_in_index = False, False
+
+        try:
+            if not os.path.exists(index_folder_path):
+                print_error(f"{GCPConfig.INDEX_NAME} does not exists.")
+                return task_status, exists_in_index
+
+            exists_in_index = os.path.exists(os.path.join(index_folder_path, self._pack_name))
+            task_status = True
+        except Exception as e:
```
yaakovi, line 1263: what exception can come here? looks redundant...

Tests/Marketplace/upload_packs.py
IkaDemisto: What do you mean? It is pack method

Tests/Marketplace/marketplace_services.py
IkaDemisto: Nothing, just for same convention

Tests/Marketplace/upload_packs.py
yaakovi: why not update the status inside the `check_if_exists_in_index` function?

PR #8033, opened 2020-07-15
Upload packs - check if index is already updated

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Related Issues
fixes: https://github.com/demisto/etc/issues/26536

## Description
Commit hash of running build will be added to `index.json` file under `commit` field. 
In case that running build cannot find commit hash from index -> build is not updated, should be skipped.
Additionally if no packs files were added/modified/deleted between current commit and one from index, then upload packs is skipped.


Tests/Marketplace/upload_packs.py
```diff
@@ -471,6 +474,52 @@ def get_recent_commits_data(content_repo):
...
+        return
+
+    skipping_build_task_message = "Skipping Upload Packs To Marketplace Storage Step."
+
+    try:
+        with open(os.path.join(index_folder_path, f"{GCPConfig.INDEX_NAME}.json")) as index_file:
+            index_json = json.load(index_file)
+
+        index_commit_hash = index_json.get('commit', remote_previous_commit_hash)
+        index_commit = content_repo.commit(index_commit_hash)
```
glicht, line 500: What happens if the commit is not found? Are you handling this case?

Tests/Marketplace/upload_packs.py
```diff
@@ -471,6 +474,52 @@ def get_recent_commits_data(content_repo):
...
+        current_commit = content_repo.commit(current_commit_hash)
+
+        if current_commit.committed_datetime <= index_commit.committed_datetime:
+            print_warning("Index is already updated.")
```
glicht, line 504: Add here the hash and dates. So it will be easier to understand and debug.

Tests/Marketplace/upload_packs.py
IkaDemisto: Yes, in line 516. `ValueError` is raised when no hash found.

glicht: Would rather you specifically catch this error here, as ValueError is a generic error that something else may raise and then we would want to error out if something else threw this..

IkaDemisto:
> @IkaDemisto I see that you are doing: `index_blob.upload_from_filename(index_zip_path)`. you should be using `if_generation_match` as an arg where you retrieve the generation number from the downloaded index. That way you only update the index if it hasn't change between the time you've downloaded it and the time your are uploading. Take a look here for an example how this is used: https://github.com/demisto/content/blob/master/Tests/test_content.py#L1327

@glicht Done

Tests/Marketplace/upload_packs.py
```diff
@@ -78,25 +78,26 @@ def download_and_extract_index(storage_bucket, extract_destination_path):
...
         os.mkdir(extract_destination_path)
 
     if not index_blob.exists():
         os.mkdir(index_folder_path)
-        return index_folder_path, index_blob
+        return index_folder_path, index_blob, index_generation
 
-    # index zip should never be cached in the memory, should be updated version
-    index_blob.cache_control = "no-cache,max-age=0"
     index_blob.download_to_filename(download_index_path)
```
glicht, line 98: I think this should be done after `reload` and use `if_generation_match` to make sure the version we download is the generation number we saved.

IkaDemisto: @glicht Switched line after `reload` but I don't see that `download_to_filename` function has `if_generation_match` in `google-cloud-storage` library version that we use (although in official docs there is such an argument).
Screen Shot 2020-07-19 at 0 00 56

glicht: I think you are looking at old code. v1.29.0 has it. See here: https://github.com/googleapis/python-storage/blob/v1.29.0/google/cloud/storage/blob.py#L956

PR #8034, opened 2020-07-15
removing setup remote docker from config.yml

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Description
Removing setup remote docker from all jobs but unittests since they are flaky and not needed.

* Added the `Not running this job` step because each job must have a least one step that runs in it in any case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -320,7 +320,7 @@ references:
 jobs:
   Setup Environment:
     <<: *container_config
-    resource_class: medium+
+    resource_class: medium
     <<: *environment
     steps:
       - checkout
@@ -337,7 +337,7 @@ jobs:
 
   Run Unit Testing And Lint:
     <<: *container_config
-    resource_class: medium+
+    resource_class: medium
     <<: *environment
     steps:
       - *attach_workspace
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -316,15 +316,14 @@ references:
                 git fetch $USER $BRANCH
                 git checkout $USER/$BRANCH $PACK
                 exit 0
-    
+
 jobs:
   Setup Environment:
     <<: *container_config
     resource_class: medium+
```
glicht, line 323: Shouldn't this be `medium`?

.circleci/config.yml
```diff
@@ -738,7 +787,20 @@ jobs:
...
     <<: *environment
     steps:
-      - setup_remote_docker
+      - when:
+          condition:
+            or:
+              - << pipeline.parameters.instance_tests >>
+              - << pipeline.parameters.contrib_branch >>
+              - << pipeline.parameters.demisto_sdk_nightly >>
+              - << pipeline.parameters.nightly >>
```
glicht, line 796: I think 6.0 should run in nightly. Not sure about non_ami_run. @yaakovi fyi

hod-alpert: My bad, you're right

hod-alpert: non_ami_run runs in 5_0 currently, fixing it too

hod-alpert:
Non ami build example: https://app.circleci.com/pipelines/github/demisto/content/19589/workflows/5a02d056-9778-487c-aa7a-e8b4a6ba1899

PR #8040, opened 2020-07-16
Infocyte - unmock

Itay4:
unmocking the integration test until further invetigation


yaakovi:
better to link with an issue

PR #8074, opened 2020-07-19
NumberOfPhishingAttemptPerUser - Changed the default query to search only for incident from the last 30 days

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26688#

## Screenshots
aside from ut, tested in cs.demisto.works for sanity:
Screen Shot 2020-07-19 at 11 31 28

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/ReleaseNotes/1_2_8.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### NumberOfPhishingAttemptPerUser
+- Changed the default query to search only for incident from the last 30 days.
```
amshamah419, line 4:
```suggestion
- Changed the default query to search only for incidents from the last 30 days.
```

bakatzir:
fails on non related tests from commonscripts: `WordTokenizeTest`, `TestWordFileToIOC`
which their scripts are in deprecated content. checking regardless if we still need these tests.

bakatzir:
@ShahafBenYakir I've moved tests of deprecated content in the last commit to noncircletests, care to have a look to make sure i'm not missing anything?

PR #8075, opened 2020-07-19
AwakeSecurity: Update logo (#8061)

jochman:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

Updated to the new AwakeSecurity logo

PR #8090, opened 2020-07-19
Revert "AwakeSecurity: Update logo (#8061)"

Itay4:
Reverts demisto/content#8075

fails validation as image size is larger than 10kb


Itay4:
build fails on non relevant validation (rn for non supported)
merging

PR #8100, opened 2020-07-20
Update awakesecurity picture

jochman:
Now features a valid size.


Itay4:
@jochman is this still relevant ?

PR #8152, opened 2020-07-22
ReadPDFFilev2 Handle mixed encoding

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26760

## Description
The issue is a bit nuanced. When `pdfinfo` is called, it by default returns a UTF-8 encoded string. However, the metadata which is extracted contains a UTF-16 LE BOM character (`\xff\xfe`) which when interpreted by python as a UTF-8 string, causes it to crash with the following error:
```
The script failed read PDF file due to an error: 'utf-8' codec can't decode byte 0xfe in position 124: invalid start byte
```
This PR proposes a solution where we receive a byte string from the subprocess as opposed to UTF-8 text and instead gracefully handle invalid characters by replacing them with UTF-8 legal invalid characters.

The returned byte string contains the following which is the cause of the issue:
```
Producer:       REDACTED ActiveReports  ™  for .NET\nCreationDate:   \xfe\xff\n
```

The result is a war room readable response and valid context.


Packs/CommonScripts/Scripts/ReadPDFFileV2/ReadPDFFileV2.py
```diff
@@ -149,7 +150,7 @@ def get_pdf_htmls_content(pdf_path, output_folder):
     for file_name in html_file_names:
         with open(file_name, 'rb') as f:
             for line in f:
-                html_content += line.decode('utf-8')
+                html_content += str(line)
```
yuvalbenshalom, line 153: is this related?

amshamah419: Sadly yes

yuvalbenshalom: @amshamah419 but here you did not change the way you read the file. so how dose this change helps?

amshamah419: Because the PDF body also contains broken encoding (not sure why). I sat on this for hours since I originally thought there was no fundamental difference between `str()` and .decode('utf-8)` 

`str()` when called without any arguments is the same as `object.__str__` which does not check for Python validity and will return a string object whether it is valid or not since the fallback is to return it as `repr()`

`.encode()` by default assumes the bytes to be utf-8 and handles errors strictly. So in this case, it fails because the mixed encoding isn't valid.

amshamah419: Worth noting that if we use `.decode('utf-8', 'replace')` it will work also.

amshamah419:
Checking with the customer to see if we can use the PDF they sent to debug. If not, then trying to find one will be tough.

Packs/CommonScripts/Scripts/ReadPDFFileV2/ReadPDFFileV2.py
yuvalbenshalom: Nice

yaakovi:
> Checking with the customer to see if we can use the PDF they sent to debug. If not, then trying to find one will be tough.

not mandatory, explain that it will go to a public repo and will be accessible to everyone.
they most probably will not approve of it

PR #8157, opened 2020-07-22
move print statement to actual name of the instance being configured …

avidan-H:
## Status
- [x] Ready

## Description
When configuring an integration instance via `__create_integration_instance`, the _actual_ name given to the instance will be shown in the print message.

## Minimum version of Demisto
- [x] 4.5.0
- [x] 5.0.0
- [x] 5.5.0

   - [x] No


avidan-H:
Adding the `docs-approved` label since there is no documentation that needs to be reviewed.

PR #8167, opened 2020-07-22
Gmail bugfix send mail

unparalleled-js:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://dfircommunity.slack.com/archives/CMWN08YB1/p1595281292043000

## Description
Fixes a bug where in Gmail integration where `send-mail` command would send both body and htmlBody

## Screenshots

Before:
Screen Shot 2020-07-28 at 12 35 41

After:
Screen Shot 2020-07-28 at 12 47 43

## Minimum version of Demisto
- [ ] 4.5.0
- [x] 5.0.0
- [ ] 5.5.0

   - [x] Yes
       - Further details: Fixes bug that maybe some depended on existing (but unlikely?)
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 
There is lack of tests all around when it comes to this integration.

## Demisto Partner?
I am a partner but am not contributing to our partner integration


avidan-H:
@unparalleled-js To better understand the issue you are trying to address, can you add a screenshot to the PR description showing the bug, and a second screenshot of the same data-collection email after using your suggested fix?

unparalleled-js:
@avidan-H I just attached two images as requested, one for Before and one for After.
Notice that in Before, the information is repeated twice. This is because it is attached to both `body` and `htmlBody`.  After only attaching the message once, we get the desired result. My understanding is that `send-mail` should either use `body` or `htmlBody`, but not both.

avidan-H:
@Itay4 This would technically break backward-compatibility. Can we move forward with this anyhow to fix this behavior?

unparalleled-js:
> @Itay4 This would technically break backward-compatibility. Can we move forward with this anyhow to fix this behavior?

Perhaps it is debatable whether this actually breaks backwards compat. If it does indeed fix a bug, then it does not break the intent.

anara123:
I think the reason it appears twice is a bug which cause by two times adding the message to messages.
image
@Itay4 I think the PR fixes the issue properly

anara123:
@avidan-H you seem to be right. Then one message should be `part2 = MIMEText(html, "html")`

unparalleled-js:
Looks like you all got it from here?

avidan-H:
@unparalleled-js yeah, thanks for submitting this issue and fix.
@anara123 I'm going to close this PR and open a new one with the alternate fix that was discussed here.

PR #8170, opened 2020-07-23
Set default classifiers and mappers

Itay4:
## Status
- [x] Ready

## Description
Setting the right key for default classifiers and mappers in integrations


Itay4:
build fails on non relevant docker image version and non related tests
merging

PR #8146, opened 2020-07-21
Contributing js transformer If-Then-Else.

tundisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
N/A

## Description
Contributes a new transformer for simple if-then-else logic.  This can potentially reduce the number of tasks required for a given playbook.

## Screenshots
if-then-else1

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [x] 5.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation 

## Demisto Partner?
N/A

PR #8180, opened 2020-07-23
If then else transformer

liorblob:
Original PR: https://github.com/demisto/content/pull/8146


Tests/conf.json
```diff
@@ -2848,6 +2848,10 @@
             "integrations": "Blueliv ThreatCompass",
             "playbookID": "Blueliv_ThreatCompass_test",
             "fromversion": "5.0.0"
+        },
+        {
+            "playbookID": "IfThenElse-Test",
+            "fromversion": "5.5.0"
```
Itay4, line 2854: why ?

liorblob: it's a test playbook... why not?

Itay4: why not run it on all versions?

PR #8241, opened 2020-07-28
Fix failure to parse p7m files by ParseEmailFiles

yuvalbenshalom:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26525

## Description
P7m files fail because in the customers example, the nested SMIME mail is actually the real one, and the one outside is the wrapper.

test is not included, but I ran it locally.(because of sensitive data)


Packs/CommonScripts/Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3594,10 +3591,10 @@ def handle_eml(file_path, b64=False, file_name=None, parse_only_headers=False, m
                 text = get_utf_string(part.get_payload(decode=True), 'TEXT')
         email_data = None
         # if we are parsing a signed attachment there can be one of two options:
-        # 1. it is a wrapper and we can ignore the outer "email"
-        # 2. it is not a wrapper and will not get into recursion, therefore we need the second condition
+        # 1. it is 'multipart/signed' so it is probably a wrapper and we can ignore the outer "email"
+        # 2. if it is 'multipart/signed' but has 'to' address so it is actually a real mail.
         if 'multipart/signed' not in eml.get_content_type()\
-                or ('multipart/signed' in eml.get_content_type() and IS_NESTED_EML is False):
+                or ('multipart/signed' in eml.get_content_type() and extract_address_eml(eml, 'to')):
```
yuvalbenshalom, line 3597: @anara123 and @yaakovi  please take a look.
In the S/MIME example that caused the bug from the customer(that I used and tested locally) the real mail was actually the nested one. so I figured a better way to check if it Is the warper or not is to check if it has a `to` address.

please tell me what do you think.

yaakovi: not sure i understand the scenario:
getting a phishing mail, sending him as an attachment to mailbox `x`.
when downloading the mail (wrapper) from `x` and parsing it, wouldn't both the mails will have a `to` field?
wrapper -> `x` mailbox
actual mail -> my mailbox

yuvalbenshalom: as you can see in the issue, the case is an .eml file of the signed smime mail. the way it is parsed here is as an nested mail where one is the "sigend" part and one is the actual mail.

in the prevues implementation it expected that the signed part, with has no `to` or `subject` or whatever, is in the nested part, so it is not being parsed as mail.(and we have aUT example with it)

in my case, the signing comes before, and the nested part has the actual mail. so I figured, it is better to just check if it has `to` field and drop the nesting check.

this is passing all of the other smime UT examples that we have.

yaakovi (approved):

PR #8255, opened 2020-07-28
Fixed titlecase metadata fields

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Related Issues
fixes: https://github.com/demisto/etc/issues/26102

## Description
Titlecase metadata fields only when input word is all lower case.


yaakovi:
failure is unrelated to the change

PR #8260, opened 2020-07-29
RSA NetWitness Security Analytics - add readme

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26969

   - [x] No

## Must have
- [x] Tests - no
- [x] Documentation


bakatzir:
@anara123 Should I also deprecate it? Or is stating in the docs the incompatibility with NW v11  sufficient?

Packs/RsaNetwitnessSecurityAnalytics/Integrations/README.md
```diff
@@ -0,0 +1,433 @@
+RSA Security Analytics is a distributed and modular system that enables highly flexible deployment architectures that scale with the needs of the organization. Security Analytics allows administrators to collect two types of data from the network infrastructure, packet data and log data.
+This integration was developed using an unofficial API, and is not compatible with RSA NetWitness v11 and onwards.
```
yuvalbenshalom, line 2: @bakatzir should we direct the user to use v11.1 in that case?

maybe we should also add it in the description so it will be visible in first glance?

bakatzir: done

PR #8309, opened 2020-08-01
xdr_iocs - updated tags param

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/22895

## Description
updated tags param

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation 

## Demisto Partner?


Packs/CortexXDR/Integrations/XDR_iocs/XDR_iocs.py
```diff
@@ -422,7 +422,7 @@ def main():
     params = demisto.params()
     Client.severity = params.get('severity', '').upper()
     Client.query = params.get('query', Client.query)
-    Client.tag = params.get('tag', Client.tag)
+    Client.tag = params.get('tag', Client.tag) if params.get('tag', Client.tag) else params.get('feedTags', Client.tag)
```
yaakovi, line 425: since `Client.tag` is not falsy, this will result in always using `tag` and not `feedTags`.

image the next scenario:
1) setting up the integration (`tags` is set)
2) upgrading pack / content
3) switching tags value (now `feedTags` is set)
in the current state, step 3 will not have any effect.

esharf, line 425: `params.get('feedTags', params.get('tag', Client.tag))`
@orhovy i think you should use something like this.

orhovy: I changed it to what @esharf  suggested, this will have an effect in my understanding.

orhovy:
> You should add a UT that mocks the demisto.params() so we can explain why keeping line 425 is important and how it helps us not break BC.
> (1 time with `tag` mocked, 2nd with `feedTags`)

I added 3 types , one without tags param which will set to default value, one with tag param and one with feedTags

PR #8315, opened 2020-08-02
demistomock - convert docstring to google style

Itay4:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/26399

## Description
Converting docstring style to google style as it looks better when rendering docs

### DIFF
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -388,18 +388,19 @@ def params():
     """(Integration only)
     Retrieves the integration parameters object
 
-    :return: Integrations parameters object
-    :rtype: ``dict``
+    Returns:
+      dict: Integrations parameters object
+
     """
     return {}
 
 
 def args():
-    """
-    Retrieves a command / script arguments object
+    """Retrieves a command / script arguments object
 
-    :return: Arguments object
-    :rtype: ``dict``
+    Returns:
+      dict: Arguments object
+
     """
     return {}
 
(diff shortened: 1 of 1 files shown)


Tests/demistomock/demistomock.py
```diff
@@ -388,18 +388,19 @@ def params():
...
     return {}
 
 
 def args():
-    """
-    Retrieves a command / script arguments object
+    """Retrieves a command / script arguments object
+
+    Returns:
+      dict: Arguments object
```
ronykoz, line 402:
```suggestion
      dict. Arguments object
```
here and everywhere else

Tests/demistomock/demistomock.py
```diff
@@ -408,37 +409,35 @@ def command():
...
 
-    :type msg: ``str``
-    :param msg: The message to be logged
+    Args:
+      msg (str): The message to be logged
 
-    :return: No data returned
-    :rtype: ``None``
+    Returns:
+      None: No data returned
```
ronykoz, line 426: no need to add that when no data returned

Itay4: need it so it will be clear that nothing is returned

Tests/demistomock/demistomock.py
Itay4: both are ok actually

glicht:
@Itay4 flake8 styling is failing

PR #8332, opened 2020-08-03
ArcSight ESM v2 - Fixed an issue where long *eventId* were not proces…

bakatzir:
…sed correctly in the ***as-get-security-events*** command

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26425

## Screenshots
in the referenced issue

   - [x] maybe - changing outputs in the yml from int to str and from unknown to str. @yaakovi WDYT?

## Must have
- [x] Tests

### DIFF
--- a/Packs/ArcSightESM/Integrations/ArcSightESMv2/ArcSightESMv2_test.py
+++ b/Packs/ArcSightESM/Integrations/ArcSightESMv2/ArcSightESMv2_test.py
@@ -108,7 +108,7 @@ def test_use_rest(mocker, use_rest, cmd_name, expected_rest_endpoint):
 
 
 def test_decode_arcsight_output_event_ids():
-    """Unit test
+    """Unit test - When output to the incident context integers, demisto can round them if they are bigger than 2^32
     Given
     - a long eventId, baseEventIds
     When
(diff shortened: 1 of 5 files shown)


Packs/ArcSightESM/Integrations/ArcSightESMv2/ArcSightESMv2_test.py
```diff
@@ -106,3 +105,22 @@ def test_use_rest(mocker, use_rest, cmd_name, expected_rest_endpoint):
         arcsight_cmd()
         last_request = m.last_request
         assert last_request.url == expected_rest_endpoint
+
+
+def test_decode_arcsight_output_event_ids():
```
yuvalbenshalom, line 110: not sure that this UT is relevant, as the problem manifest itself on the server side.

bakatzir: I also think it it is only a partial coverage, but I think it might be useful to understand why we cast the ids.
I'll add a more thorough explanation in the test.

Packs/ArcSightESM/ReleaseNotes/1_0_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### ArcSight ESM v2
+- Fixed an issue where long *eventId* were not processed correctly in the ***as-get-security-events*** command.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where a long *eventId* is not processed correctly in the ***as-get-security-events*** command.
```

yaakovi:
>  maybe - changing outputs in the yml from int to str and from unknown to str. @yaakovi WDYT?
since no one will ever use this field (ID) as an actual number, this is fine.

PR #8335, opened 2020-08-03
Use unrar for unzipping .rar files

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27110

## Description
- Added a new docker image so we can maintain it better. Added unrar and p7zip to an alpine image called `demisto/unzip`
- Separate processes are used depending on the file extension. if `.rar` is in the file extension, we use unrar


Packs/CommonScripts/Scripts/UnzipFile/UnzipFile.yml
```diff
@@ -36,15 +36,15 @@ outputs:
...
 - file
 timeout: '0'
 type: python
-dockerimage: demisto/7zip
+dockerimage: devdemisto/unzip:1.0.0.10195
 runonce: false
 tests:
 - ZipFile-Test
 - UnzipFile-Test
-subtype: python2
```
bakatzir, line 50: i'm not sure you can change the subtype of the automation. if it is a must, we might need to do a v2.

Packs/CommonScripts/Scripts/UnzipFile/UnzipFile.yml
```diff
@@ -36,15 +36,15 @@ outputs:
...
 script: '-'
+subtype: python3
 system: false
 tags:
 - Utility
 - file
 timeout: '0'
 type: python
-dockerimage: demisto/7zip
+dockerimage: devdemisto/unzip:1.0.0.10195
```
bakatzir, line 46: a reminder to change after merging this: https://github.com/demisto/dockerfiles/pull/2295

Packs/CommonScripts/Scripts/UnzipFile/UnzipFile.yml
amshamah419: I don't think it's a must. Just figured we would bring it to py3. I can gen the docker image as a python2 image without any issues. Question is if we want to do `toVersion` --> `fromVersion` and just quietly replace as a v2. @yuvalbenshalom what are your thoughts?

yuvalbenshalom: because this is a simple one, it is ok just to test it with regular zip, 7zip and password, and just add `dockerimage45`. no need for toversion/fromversion.

PR #8336, opened 2020-08-03
SearchIncidentsV2 - Fixed the *query* argument to not support input a…

bakatzir:
…s array

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27136

## Minimum version of Demisto
- [x] 5.0.0


bakatzir:
Syslog fails
`Failed: playbook: Test Syslog with integration(s): Syslog Sender,syslog failed`
non related.

PR #8383, opened 2020-08-05
Mocks Logging

avidan-H:
## Status
- [x] Ready

## Description
Adds a `run_tests.log` file to the artifacts for each server instance `Run Tests` circleci job, to which `mock_server.py` logs extra debugging statements. The outputs of `stdout` and `stderr` of the `mitmproxy` process are logged there as well.

### DIFF
--- a/Tests/mock_server.py
+++ b/Tests/mock_server.py
@@ -191,6 +191,10 @@ class MITMProxy:
         self.empty_files = []
         self.rerecorded_tests = []
 
+        self.logs_file = open(os.getenv('TESTS_LOGS'), 'a') if os.getenv('TESTS_LOGS') else open(
+            '/home/circleci/project/artifacts/run_tests.log', 'a')
+        self.print_to_log = lambda *objects: print(*objects, file=self.logs_file)
+
         silence_output(self.ami.call, ['mkdir', '-p', tmp_folder], stderr='null')
 
     def configure_proxy_in_demisto(self, demisto_api_key, server, proxy=''):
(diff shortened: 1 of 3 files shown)


avidan-H:
Adding the `docs-approved` label since there is no documentation to review here.

Tests/test_content.py
```diff
@@ -899,6 +899,7 @@ def execute_testing(tests_settings, server_ip, mockable_tests_names, unmockable_
                 updating_mocks_msg = "Pushing new/updated mock files to mock git repo."
                 prints_manager.add_print_job(updating_mocks_msg, print, thread_index)
                 ami.upload_mock_files(build_name, build_number)
+                prints_manager.execute_thread_prints(thread_index)
```
yaakovi, line 902: why not execute it directly under the finally? why specifically for AMI on the master branch?

Tests/mock_server.py
```diff
@@ -191,6 +191,10 @@ def __init__(self, public_ip,
         self.empty_files = []
         self.rerecorded_tests = []
 
+        self.logs_file = open(os.getenv('TESTS_LOGS'), 'a') if os.getenv('TESTS_LOGS') else open(
+            '/home/circleci/project/artifacts/run_tests.log', 'a')
+        self.print_to_log = lambda *objects: print(*objects, file=self.logs_file)
```
yaakovi, line 196: consider also adding `flush=True`

avidan-H: good idea.

Tests/test_content.py
avidan-H: not sure I understand this comment.

yaakovi: why is that in the condition? why not run at any time in the `finally` block

PR #8384, opened 2020-08-05
Auto map support custom fields

erezh31:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26980

## Description
Improve logic -when alias name equal find the best match to the incident field
Support user fields (till now only system fields)

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


erezh31:
No need for docs

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
```diff
@@ -526,8 +533,25 @@ def get_alias_index(field_name, alias):
     return SIEM_FIELDS[field_name]['aliases'].index(alias)  # type: ignore
 
 
-def get_most_relevant(field_name, field_mappings):
-    candidates = sorted(field_mappings, key=lambda x: get_alias_index(field_name, x[1]))
+def get_most_relevant_json_field(field_name, json_field_to_alias):
+    if not json_field_to_alias:
```
lperryp, line 537: I think the condition here needs to be `if len(json_field_to_alias)>0 `

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
```diff
@@ -526,8 +533,25 @@ def get_alias_index(field_name, alias):
...
+        return
```
lperryp, line 538: Also here check how it affects `match_for_incidents` that you might return None as a matching (perhaps filter none values at `match_for_incidents`)

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
```diff
@@ -526,8 +533,25 @@ def get_alias_index(field_name, alias):
...
+    # calculate jaccard score for each alias, and get the candidates with max score
+    scores = {}
+    for json_field, alias in json_field_to_alias.items():
+        scores[json_field] = jaccard_similarity_for_string_terms(json_field, alias)
+    scores = {k: v for k, v in scores.items() if v == max(scores.values())}
+
+    # calculate jaccard score for each field, and get the candidates with max score
+    for json_field, alias in json_field_to_alias.items():
+        scores[json_field] = jaccard_similarity_for_string_terms(json_field, field_name)
+    scores = {k: v for k, v in scores.items() if v == max(scores.values())}
```
lperryp, line 549: In case ` jaccard_similarity_for_string_terms(json_field, alias)` is high, it's possible that you'll return a mapping between json_field to its alias?

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
```diff
@@ -526,8 +533,25 @@ def get_alias_index(field_name, alias):
...
+
+    # for candidates with the same score with the least alias index
+    candidates = sorted(list(scores.keys()),
+                        key=lambda json_field: get_alias_index(field_name, json_field_to_alias[json_field]))
```
lperryp, line 553: Is it still going to work if `json_field` is not in `SIEM_FIELDS[field_name]`? might cause an exception?

erezh31: No, it will suppose to be there cuz probably it still match on some alias
anyway index return -1 and it's just for sorting so no error

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
erezh31: json field doesn't have alias inciden field it's the one with alias
so the match is json field with incident field alias

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
erezh31: added a condition to verify non null there

Packs/Base/Scripts/DBotSuggestClassifierMapping/DBotSuggestClassifierMapping.py
erezh31: you mean len(json_field_to_alias) == 0, suppose to be the same but changed

PR #8391, opened 2020-08-06
Revert "Mocks Logging"

avidan-H:
Reverts demisto/content#8383


avidan-H:
Adding `docs-approved` label since there is no documentation to review.

PR #8395, opened 2020-08-06
CrowdStrike Streaming v2 - add a note about one client can connect to a stream at a time

Itay4:
## Status
- [x] Ready

## Description
Added a note that only one api client can connect to the stream


kirbles19:
@Itay4 Done.

PR #8406, opened 2020-08-07
SplunkPy fetch bugfix - Added Intervals

amshamah419:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26598
fixes: https://github.com/demisto/etc/issues/27264

## Description
There currently exists an issue with splunk where an incident may be generated, but not indexed within our normal 1-min fetch window. This version adds the ability to set a longer fetch interval if needed to prevent incidents from being missed.


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -479,7 +486,12 @@ def fetch_incidents(service):
     earliest_fetch_time_fieldname = dem_params.get("earliest_fetch_time_fieldname", "index_earliest")
     latest_fetch_time_fieldname = dem_params.get("latest_fetch_time_fieldname", "index_latest")
 
-    kwargs_oneshot = {earliest_fetch_time_fieldname: last_run,
+    # Handle last run time logic
+    lastrun_datetime = datetime.strptime(last_run, SPLUNK_TIME_FORMAT)
+    last_run_time_windowed_dt = lastrun_datetime - timedelta(minutes=LAST_RUN_WINDOW)
```
amshamah419, line 491: Need better naming here

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -179,6 +180,9 @@ def severity_to_level(severity):
 
 
 def notable_to_incident(event):
+    inc_uuid = None
+    if demisto.get(event, '_bkt'):
+        inc_uuid = event['_bkt']
```
yuvalbenshalom, line 185: I think we need to make an info, when uuid dose not exist.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
```diff
@@ -105,6 +105,13 @@ configuration:
   name: use_requests_handler
   required: false
   type: 8
+- defaultvalue: '1'
+  display: The interval in minutes to fetch incidents. (e.g. 5 would fetch every five
```
yuvalbenshalom, line 109: we will need better explanation to this interval as it go backwards in 5 minutes.

Itay4:
@amshamah419 can we close this?

amshamah419:
@Itay4 The issue is still ongoing with some customers. Until we have a more concrete solution, I prefer to keep this open so we can continue to track.

Itay4:
@amshamah419 do we still need this open ?

amshamah419:
@Itay4 No, we can go ahead and kill it.

PR #8426, opened 2020-08-09
unskipping cdl

hod-alpert:
Unskipping CDL 
Related to https://github.com/demisto/etc/issues/24346


hod-alpert:
Encountered some issues with the dev instance now, will review them and unskip once all is done

PR #8433, opened 2020-08-10
Added download count to pack metadata

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Related Issues
fixes: https://github.com/demisto/etc/issues/27129

## Description
Added pack download count to pack metadata using big query and processing data frame using pandas library in place.

### DIFF
--- a/dev-requirements-py3.txt
+++ b/dev-requirements-py3.txt
@@ -20,3 +20,4 @@ circleci==1.2.2
 pylint==2.5.3
 coloredlogs==14.0
 pandas==1.1.0
+pyarrow==1.0.0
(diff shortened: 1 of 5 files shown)


dev-requirements-py3.txt
```diff
@@ -19,3 +19,5 @@ git+https://github.com/coyo8/parinx.git@6493798ceba8089345d970f71be4a896eb6b081d
 circleci==1.2.2
 pylint==2.5.3
 coloredlogs==14.0
+pandas==1.1.0
+pyarrow==1.0.0
```
yaakovi, line 23: why do you need this package?

IkaDemisto: `to_dataframe` function requires it:
```
                This method requires the ``pyarrow`` and
                ``google-cloud-bigquery-storage`` libraries.
```

PR #8439, opened 2020-08-10
Zip marketplace-dist instead of marketplace-ci-build

orlichter1:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27020

## Description
Upload to artifacts content from marketplace-dist bucket instead of `marketplace-ci-build` bucket.

## Screenshots

## Minimum version of Demisto
- [ ] 4.5.0
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


.circleci/config.yml
```diff
@@ -486,7 +486,7 @@ jobs:
                   GCS_PATH=$(mktemp)
                   ZIP_FOLDER=$(mktemp -d)
                   echo $GCS_MARKET_KEY > $GCS_PATH
-                  python3 ./Tests/Marketplace/zip_packs.py -b 'marketplace-ci-build' -z $ZIP_FOLDER -a $CIRCLE_ARTIFACTS -s $GCS_PATH -n $CIRCLE_BUILD_NUM -br $CIRCLE_BRANCH
+                  python3 ./Tests/Marketplace/zip_packs.py -b 'marketplace-dist' -z $ZIP_FOLDER -a $CIRCLE_ARTIFACTS -s $GCS_PATH
```
IkaDemisto, line 489: I think that this part should run only on master branch, it has no impact on feature branches. 
And additionally be part of different and separate workflow and not `Create Instances` together with `Upload Packs To Marketplace Storage` step. Part of it defined here: https://github.com/demisto/etc/issues/26708
@yaakovi wdyt?

orlichter1: This step takes all of 16 seconds. I do agree that it should be moved to the `Packs To Marketplace Storage` step when we get to that task but I'm not sure if necessary to limit it to certain branches (master, releases...) as it's very very fast. 
I'll make a few more tests to be sure it keeps on being fast.
@IkaDemisto @yaakovi

IkaDemisto: @orlichter1 It is not related to performance. Packs are uploaded only on master branch to `marketplace-dist`. So it is needless to download packs from bucket on non master branch as it is same data each time. Only on master branch it will be different.

orlichter1: I understand. I could add a `if...fi` block in the shell script (or add a circle if) to do this. 
Just need an OK from @yaakovi. 
If this happens what should be the condition? If `$CIRCLE_BRANCH` is `master` of course but we'll also need this for releases. How do release branches look?

orlichter1:
@IkaDemisto Except the comment about when to run this - looks ok?

IkaDemisto:
> @IkaDemisto Except the comment about when to run this - looks ok?

Yes, looks good.

.circleci/config.yml
yaakovi: run this step only for master and release branches

orlichter1:
@yaakovi Should we merge this now? :)

PR #8450, opened 2020-08-10
Microsoft Graph User - Added the ***msgraph-user-change-password*** command

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27201

## Description
This command requiers a special permission. added to the desc and will be added to the readme.
we have an issue with registered demisto app users of it, i'll explain tomorrow.

## Screenshots
Screen Shot 2020-08-10 at 21 50 48

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests - no as we get 204 from the api, and since the API module resides in another file
- [x] Documentation - soon

### DIFF
--- a/Packs/MicrosoftGraphUser/Integrations/MicrosoftGraphUser/MicrosoftGraphUser.yml
+++ b/Packs/MicrosoftGraphUser/Integrations/MicrosoftGraphUser/MicrosoftGraphUser.yml
@@ -167,7 +167,7 @@ script:
       isArray: false
       name: password
       required: true
-      secret: true
+      secret: false
     - default: false
       description: ' The user principal name, for example: foo@test.com. '
       isArray: false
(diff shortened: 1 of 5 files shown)


Packs/MicrosoftGraphUser/Integrations/MicrosoftGraphUser/MicrosoftGraphUser.yml
```diff
@@ -429,23 +423,69 @@ script:
...
+    - default: false
+      description: User ID or userPrincipalName to update properties for.
+      isArray: false
+      name: user
+      required: true
+      secret: false
+    - default: false
+      description: The new password
+      isArray: false
+      name: password
```
yuvalbenshalom, line 450: I think this argument should be sensitive

PR #8475, opened 2020-08-12
PassiveTotal - limit MD of enrichment cmds

bakatzir:
## Status
- [x] ~~In Progress - pending customer verification~~ Ready

## Print screen:
We have no instance, but an example to the hard limit:
Screen Shot 2020-08-16 at 10 42 12

## Related Issues
fixes: https://github.com/demisto/etc/issues/27394

   - [x] No

### DIFF
--- a/Packs/PassiveTotal/Integrations/integration-PassiveTotal.yml
+++ b/Packs/PassiveTotal/Integrations/integration-PassiveTotal.yml
@@ -17,29 +17,30 @@ configuration:
   required: false
 - display: Username/API Key
   name: Username
-  defaultvalue: ""
   type: 0
   required: true
 - display: API Secret
   name: APIKey
-  defaultvalue: ""
   type: 4
   required: true
 - display: Use system proxy settings
   name: proxy
-  defaultvalue: ""
   type: 8
   required: false
 - display: Trust any certificate (not secure)
   name: insecure
-  defaultvalue: ""
   type: 8
   required: false
 - display: Tags to signify this is a malicious domain
   name: tags
-  defaultvalue: "malware,blacklist,phishing,typo-squatting"
+  defaultvalue: malware,blacklist,phishing,typo-squatting
   type: 0
   required: true
+- display: Whether the enrichment commands return verbose response or not.
+  name: verboseEnrichment
+  defaultvalue: "false"
+  type: 8
+  required: false
 script:
   script: |-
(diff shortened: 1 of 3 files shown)


Packs/PassiveTotal/Integrations/integration-PassiveTotal.yml
```diff
@@ -249,7 +252,12 @@ script:
                     };
                 }
             }
-            var fullResult = [{Type: entryTypes.note, Contents: jsonRes, ContentsFormat: formats.json, HumanReadable: md, EntryContext: context}];
+            if (verboseEnrichment === false) {
+                var fullResult = [{Type: entryTypes.note, Contents: context, ContentsFormat: formats.json, HumanReadable: md, EntryContext: context}];
```
anara123, line 256: The `context` here contains stuff like DBotScore and Domain. I don't see why we should return it in `Contents`
What exactly is returned in non verboseEnrichment?

bakatzir: Changed the logic to show 20 keys in the md and in the raw-response.

Packs/PassiveTotal/Integrations/integration-PassiveTotal.yml
```diff
@@ -223,6 +217,17 @@ script:
                     }
                 }
             }
+            jsonKeys = Object.keys(jsonRes);
+            if (jsonKeys.length > 20) {
+                limit = 20;
+            } else {
+                limit = jsonKeys.length;
+            }
+            var jsonMd = {};
```
yuvalbenshalom, line 226: just to make sure, we only need to filter the MD? the context itself is not packed?
dose this make sense?

bakatzir: The related domains do not go into the context.

Packs/PassiveTotal/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### PassiveTotal
+- Added a limit to the number of related domain showed in the war room for enrichment commands.
```
amshamah419, line 4:
```suggestion
- Added a limit to the number of related domains showed in the war room for enrichment commands.
```

PR #8492, opened 2020-08-13
Evaluate ml docker version bump

teizenman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26849


Packs/ML/TestPlaybooks/script-CreateIncidentsForEvaluateMLModllAtProduction.yml
```diff
@@ -74,6 +74,6 @@ outputs:
 scripttarget: 0
 subtype: python2
 runonce: false
-dockerimage: demisto/python:2.7.18.9326
+dockerimage: demisto/python:2.7.18.10627
 runas: DBotWeakRole
-fromversion: 5.0.0
+fromversion: 4.5.0
```
reutshal, line 79: Why?

teizenman: The playbook is using this script, and it is `fromversion:4.5.0` and this script is not used anywhere but in the playbook so I adjusted it to be the same.

PR #8537, opened 2020-08-17
Added tlp support to API modules

teizenman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/26162


avidan-H:
Hey, @teizenman please don't merge yet - I have some comments. I'll submit soon.

Packs/ApiModules/Scripts/HTTPFeedApiModule/HTTPFeedApiModule.py
```diff
@@ -269,6 +272,7 @@ def get_indicator_fields(line, url, feed_tags: list, client: Client):
     :param url: The feed URL
     :param client: The client
     :param feed_tags: The indicator tags.
+    :param tlp_color: Traffic Light Protocol color.
```
avidan-H, line 275: You haven't passed it as a function argument here so I'm not sure why you've listed it as a parameter in your docstring - you've implemented `tlp_color` as an instance variable of the `Client` class above.

Packs/ApiModules/Scripts/HTTPFeedApiModule/HTTPFeedApiModule.py
```diff
@@ -432,6 +437,7 @@ def feed_main(feed_name, params=None, prefix=''):
             args['feed_name'] = feed_name
             if feed_tags:
                 args['feedTags'] = feed_tags
+            args['tlp_color'] = params.get('tlp_color')
```
avidan-H, line 440: Your code has `tlp_color` as an instance variable of the `Client` class, yet here you seem to be incorporating it into the `args` dictionary that is passed to command functions. Why?

Packs/ApiModules/Scripts/JSONFeedApiModule/JSONFeedApiModule.py
```diff
@@ -125,7 +129,12 @@ def fetch_indicators_command(client: Client, indicator_type: str, feedTags: list
                 if not current_indicator_type:
                     continue
 
-                indicator = {'value': indicator_value, 'type': current_indicator_type, 'fields': {'tags': feedTags}}
+                indicator = {'value': indicator_value, 'type': current_indicator_type,
+                             'fields': {
+                                 'tags': feedTags,
+                                 'trafficlightprotocol': client.tlp_color
+                             }
+                             }
```
avidan-H, line 137:
```suggestion
                indicator = {
                	'value': indicator_value,
                	'type': current_indicator_type,
                    'fields': {
                    	'tags': feedTags,
                    	'trafficlightprotocol': client.tlp_color
                    }
                }
```
You have strange formatting here.

avidan-H:
@teizenman I think it also might make sense to include the updated generic feed integration that corresponds with each of the updated `ApiModules`.

teizenman:
@avidan-H 
`You've set the default TLP value to RED - how come, and are we sure that's the desired default value?`
That is currently a placeholder since I wanted to have you review the process of adding the TLP at first.

teizenman:
TAXII2 test is currently failing, unrelated to the changes
Im on it

teizenman:
@avidan-H Can you go over it now that we concluded the open issues?

Packs/ApiModules/Scripts/HTTPFeedApiModule/HTTPFeedApiModule.py
```diff
@@ -325,6 +328,7 @@ def get_indicator_fields(line, url, feed_tags: list, client: Client):
         attributes['value'] = value = extracted_indicator
         attributes['type'] = feed_config.get('indicator_type', client.indicator_type)
         attributes['tags'] = feed_tags
+        attributes['trafficlightprotocol'] = client.tlp_color
```
avidan-H, line 331: When this function is called via `fetch_indicators_command`, there exists the potential for the  `tlp` color not to get set. I think you need to adjust the `custom_fields_creator` method to handle that case.

Packs/ApiModules/Scripts/HTTPFeedApiModule/HTTPFeedApiModule_test.py
```diff
@@ -221,6 +221,7 @@ def test_feed_main_fetch_indicators(mocker, requests_mock):
             'asndrop_country': 'US',
             'asndrop_org': 'LAKSH CYBERSECURITY AND DEFENSE LLC',
             'tags': tags.split(','),
+            'trafficlightprotocol': None,
```
avidan-H, line 224: I see that you've updated the unit test because now you are populating the tlp field even when its value is `None` (i.e. the user didn't set a value in the integration instance configuration) - can you check if this impacts what you see when viewing an indicator in the UI? If it is shown there, then I think this should be changed so that we only populate the field when it isn't `None`.

Packs/ApiModules/Scripts/JSONFeedApiModule/JSONFeedApiModule.py
```diff
@@ -125,7 +129,14 @@ def fetch_indicators_command(client: Client, indicator_type: str, feedTags: list
                 if not current_indicator_type:
                     continue
 
-                indicator = {'value': indicator_value, 'type': current_indicator_type, 'fields': {'tags': feedTags}}
+                indicator = {
+                    'value': indicator_value,
+                    'type': current_indicator_type,
+                    'fields': {
+                        'tags': feedTags,
+                        'trafficlightprotocol': client.tlp_color
```
avidan-H, line 137: By setting it here, wouldn't the value be overridden by a tlp value that came in from the indicator feed source (assuming a case in which that data was provided as part of the feed)? I thought we said that the value set during instance configuration should take precedence. If that is the case, you would need to conditionally set this field after the mapping bit on lines 146-149.

avidan-H (approved):
Something I noticed across the board is that because you don't conditionally set the `trafficlightprotocol` field on the indicators and since the default value of the `tlp_color` in the client is `None`, that in the case this field was mapped to from the source feed's data, that you will be overriding that value with `None` which we should not be doing. Please add a unit test for this case to make sure you've handled it appropriately in each of the API modules.

avidan-H (changes requested):
I meant to press `Request changes`, not `Approve` - whoopsies. ¯\_(ツ)_/¯

avidan-H:
@teizenman did you do any manual testing and verify the results are as expected in the UI when executing the `fetch`/`get` `_indicators_command`?

avidan-H (approved):

PR #8586, opened 2020-08-19
Fixed input bug in Detonate File - ThreatGrid under Detonate File - Generic

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27101
fixes: https://github.com/demisto/etc/issues/27597

## Description
- Fixed a bug where the input field called "private" in **PLAYBOOK** `Detonate File - ThreatGrid` did not populate correctly into the inputs of `Detonate File - ThreatGrid` as a **SUBPLAYBOOK** inside `Detonate File - Generic`.
- Improved description of the "private" input.

No - because we do not update playbook inputs that were customized when updating content, then any customer who set `Detonate File - ThreatGrid` to private=false, will not get our change that resets it back to default (which is private=true).


idovandijk:
@yaron-libman please review https://github.com/demisto/content/blob/94fed9232f83afadadfeb4b6a6b1e65bca8ae7f7/Packs/ThreatGrid/Playbooks/playbook-Detonate_File_-_ThreatGrid.yml
lines 517-518 + changelogs. Thanks!

yaron-libman:
reviewed and approved.

michalgold (approved):

PR #8592, opened 2020-08-19
CyberArk PAS - added documentation + small update

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
- Added documentation for the integration.
- Updated the the cyberark-pas-credentials-verify command to use the new API.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No


Packs/CyberArkPAS/Integrations/CyberArkPAS/CyberArkPAS.py
```diff
@@ -503,7 +503,7 @@ def verify_credentials(self,
         """
         This function uses the V1 CyberArk PAS api, currently there is no matching function in V2
         """
-        url_suffix = f"/PasswordVault/WebServices/PIMServices.svc/Accounts/{account_id}/VerifyCredentials"
+        url_suffix = f"/PasswordVault/API/Accounts/{account_id}/Verify"
```
bakatzir, line 506: why this change? are we testing it?

Packs/CyberArkPAS/Integrations/CyberArkPAS/README.md
```diff
@@ -0,0 +1,1804 @@
...
+| isFetch | Fetch incidents | False |
+| max_fetch | Max fetch | False |
+| fetch_time | First fetch timestamp \(&lt;number&gt; &lt;time unit&gt;, e.g., 12 hours, 7 days, 3 months, 1 year\) | False |
+| score | CyberArk PAS score \(0.0\-100.0\) | False |
+| incidentType | Incident type | False |
+| insecure | Trust any certificate \(not secure\) | False |
+| proxy | Use system proxy settings | False |
+
+4. Click **Test** to validate the URLs, token, and connection.
+## Commands
```
bakatzir, line 22: Before the commands, is there anything a customer should know regarding the setup/permissions needed?

ohaim1008: not that I know of, our instance is no the "traditional" one.

Packs/CyberArkPAS/Integrations/CyberArkPAS/CyberArkPAS.py
ohaim1008: I run the old api command and got an error, it looks like it is not supported anymore...
I believe that is some point it will happen to all the old ones too (but the other commands that are using the old api don't have a new version yet but this one has).

bakatzir: legit. please add in the readme for all the commands that use the v1(delete/add safe and if there are any others) a note so the customers will now that it is bound to change. you can consult with andrew/shirley to make sure it is written correctly.

bakatzir (approved):

ohaim1008:
@amshamah419 Hi Andrew, I've added to couple commands the note "Uses the V1 API".
The others are using a new api so those commands might change in the future, any thoughts of how should I write it?
Or is it ok as it is.

amshamah419:
You could use something like `Uses the V1 of the API and may change in the future.` @ShirleyDenkberg what do you think?

This is just going in the README correct? If so we don't really have a limit to how much we can write and it may be best to let the customer know that because this command uses an older version of the API, it may be changed in the future and could affect backward compatibility.

bakatzir:
Correct. 
we might `deprecate` these commands in the not very far future and update them with the new API endpoint.
but we wanted to give a heads up, as orel foresaw it coming.

ohaim1008:
@amshamah419 sounds good!
Yes only for the read me. I hardly believe the command will change dramatically and that it will affect the customers, but it should be mentioned either way.

PR #8689, opened 2020-08-26
Fix DateStringToISOFormat Unit Test

avidan-H:
## Status
- [x] Ready

## Description
I was using `capfd.disable()` to suppress `stderr` that I was outputting expectedly in certain cases since `stderr` would fail the unit test. This was bad practice since this could swallow/mask other (non-expected) errors. I replaced the use of `capfd.disable()` with simply patching `demisto.error()`.

   - [x] No

## Must have
- [x] Tests


avidan-H:
I am adding the `docs-approved` label since there is no documentation that needs review.

glicht (approved):

PR #8690, opened 2020-08-26
demistoUrls better docstring format

bakatzir:
does not looks so good on pandev
Screen Shot 2020-08-27 at 0 28 53

Let me know what you think.

### DIFF
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -758,14 +758,15 @@ def setContext(contextPath, value):
 
 def demistoUrls():
     """Retrieves the following URLs related to the incident.
-    You can use these URLs when you send notifications outside the system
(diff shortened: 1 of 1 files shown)


Tests/demistomock/demistomock.py
```diff
@@ -757,15 +757,15 @@ def setContext(contextPath, value):
...
+    """Retrieves the following URLs related to the incident.
+    You can use these URLs when you send notifications outside the system
+    over email or slack and would like to include a link to the incident.
         - evidenceBoard
         - investigation
         - relatedIncidents
         - server
         - warRoom
         - workPlan
-        You can use these URLs when you send notifications outside the system over email or slack
```
Itay4, line 767: its a bit weird the way it looks now
you can just dedent this section and it will show it was meant

Itay4 (approved):

PR #8717, opened 2020-08-30
fix SetAndHandleEmpty script

esharf:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27926

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/SetAndHandleEmpty/SetAndHandleEmpty.py
```diff
@@ -1,16 +1,35 @@
...
-        context_entry = {args.get('key'): value}
+        human_readable = f'Key {key} set'
+        context_entry = {key: value}
     else:
-        context_entry = {}
         human_readable = 'value is None'
+        context_entry = {}
+
+    if args.get('append') == 'false' and context_entry:
+        demisto.executeCommand('DeleteContext', {'key': key})
```
yaakovi, line 31: why `Delete`?

esharf: Because when append is false we should override the old key.

PR #8718, opened 2020-08-30
Infocyte - move test to noncircle test

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26445


ShahafBenYakir (approved):
Does not appear in conf.json? if it is let's remove it from there.

PR #8738, opened 2020-08-31
ArcSight ESM v2 - Changed the hard limit of each fetch to be 300 instead of 50. Recommended limit is 50

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/18660

   - [x] No

## Must have
- [x] Documentation


bakatzir:
@kirbles19 I've changed the hard limit on the fetch incidents from 50 to 200, but want to add a warning that in case of slowness/crashes - try lowering the max fetch param. can you review the readme? or let me know if it should be someplace else.

yuvalbenshalom (approved):

PR #8753, opened 2020-08-31
Adding general query check for ES v2 test module

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
Adding general query check for ES v2 test module


Packs/Elasticsearch/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -430,6 +447,10 @@ def test_func():
         except ValueError as e:
             return_error("Inserted time format is incorrect.\n" + str(e) + '\n' + TIME_FIELD + ' fetched: ' + hit_date)
 
+    else:
+        # check that we can reach any indexes in the supplied server URL
+        test_general_query(es)
```
Itay4, line 452: why are we doing it only in case fetch is not set ?

roysagi: If fetch is set then we are checking if we can approach the specific index for fetch - this serves the same purpose as this check so I only test this if the fetch is not set

Packs/Elasticsearch/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -289,6 +289,23 @@ def test_general_query(es):
...
+    """Test executing query to all available indexes.
+
+    Args:
+        es(Elasticsearch): an Elasticsearch object to which we run the test.
+    """
+    try:
+        query = QueryString(query='*')
+        search = Search(using=es, index='*').query(query)[0:1]
+        response = search.execute().to_dict()
+        _, total_results = get_total_results(response)
```
Itay4, line 302:
```suggestion
        get_total_results(response)
```

Packs/Elasticsearch/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -289,6 +289,23 @@ def test_general_query(es):
...
+
+    except NotFoundError as e:
+        return_error("Failed executing general search command - please check the Server URL and port number "
+                     "and the supplied credentials.\nError message: {}.".format(str(e).split(',')[2][2:-1]))
```
Itay4, line 306: whats with the error message here? seems like its prone to bugs to index error

roysagi: Honestly, I do not recall how the error looks like but this format has been in the integration since its inception a year ago with no issues to it. I can change it to simply print the `str(e)` if you think it is a better move.

Itay4: yea lets do that

Itay4 (approved):

roysagi:
@amshamah419 - mind checking my RN here please?

amshamah419 (commented):
RNs look good.

PR #8820, opened 2020-09-04
VXStream - Filter only supported files

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26767

## Description
Added a configurable regex to filter out png and dtd files.

## Screenshots
image

image

   - [x] Maybe
       - Further details: Users who relied on the playbook to fail on unsupported file types will not fail.
   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation


Itay4:
i'd prefer we will be consistent with all other detonation playbooks, and have a regex for supported file types, and vice-versa

in general, this product supports the same as falcon X does, so you can just take task 10 from https://github.com/demisto/content/blob/master/Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_X.yml#L255

we can also add the supported file types to the playbook description as done here https://github.com/demisto/content/blob/master/Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_X.yml#L6

see supported file types in https://www.hybrid-analysis.com/faq#what-files-can-falcon-sandbox-analyze

Packs/CrowdStrikeFalconSandbox/TestPlaybooks/playbook-FilterOutRegex-Test.yml
```diff
@@ -0,0 +1,160 @@
...
+      type: playbook
+      iscommand: false
+      brand: ""
+    nexttasks:
+      '#none#':
+      - "2"
+    scriptarguments:
+      EnvironmentID:
+        simple: "100"
+      FIlterOutRegex:
```
Itay4, line 54: please update test playbook

jochman:
Closing to handle in the integration.

PR #8830, opened 2020-09-06
fixing error in parsing oproxy response

hod-alpert:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24346
## Description
Failed with the following error 
```
   
   File "<string>", line 897, in <module>
   File "<string>", line 856, in main
   File "<string>", line 47, in __init__
   File "<string>", line 64, in _get_access_token
   File "<string>", line 90, in _oproxy_authorize
 ValueError: invalid literal for int() with base 10: ''
```

When executing this line `expires_in = int(oproxy_response.get(EXPIRES_IN, MINUTES_60))`
It seems like a case where the `EXPIRES_IN` key was in the dict but with empty string as value.
The solution is to add a fallback to 0 in case the value was empty string


Itay4 (approved):

kirbles19:
@hod-alpert Done

PR #8834, opened 2020-09-06
Iot no images

ShahafBenYakir:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
This is the same PR as this without the docs images (currently fails our validations)
https://github.com/demisto/content/pull/8808


ShahafBenYakir:
Force merging - this is a copy of an external PR that was approved by me but for reasons of SDK issues had to be copied to another PR

PR #8835, opened 2020-09-06
move upload packs to google step to before the tpb stage in 6.0 master build

bakatzir:
## Status
- [x] Ready


ShahafBenYakir:
Didn't see where we are checking that content was installed before uploading to MP, maybe I missed it?

PR #8866, opened 2020-09-07
demisto sdk validate - Changes in content

ohaim1008:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Demisto-sdk will check fetch params - changing content to fit the new changes.
The changes are: the params "first_fetch" and "max_fetch" have to appear in case the integration has a fetch command.
In the "max_fetch" param there has to be a default value.
The content changes in this pr is renaming the current params in the integration, in case one of the params is missing the IN126 will be added to the pack-ignore file.

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


bakatzir (changes requested):
you can not change the param `name` in the integration yml without supporting the old integration param name in the code.
the reason is that when the custoemr updates the content pack, his. code is being updated, but if he already has an instance configured, the instance params names(and params in general) will not change.
This is one of the reasons BTW that you can not add a mandatory param to an existing integration.

Packs/Akamai_SIEM/Integrations/Akamai_SIEM/Akamai_SIEM.py
```diff
@@ -373,8 +373,8 @@ def main():
     try:
         if command == 'fetch-incidents':
             incidents, new_last_run = fetch_incidents_command(client,
-                                                              fetch_time=params.get('fetchTime'),
-                                                              fetch_limit=params.get('fetchLimit'),
+                                                              fetch_time=params.get('first_fetch'),
+                                                              fetch_limit=params.get('max_fetch'),
```
bakatzir, line 377:
```suggestion
                                                              fetch_time=params.get('FetchTime, first_fetch'),
                                                              fetch_limit=params.get('fetchLimit, max_fetch'),
```
+ a comment why the next dev should not cahnge this line(due to BC)
this will be have to be done to all the .py files in this PR.

ohaim1008:
We have decided to use `ignore` instead of changing the integration.
Will open a new pr for that.

PR #8868, opened 2020-09-08
CommonServerPython - Added the timeline object to the **CommandResult…

bakatzir:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/24788

   - [x] No

## Must have
- [x] Tests - ~~will follow.~~
- [x] Documentation https://github.com/demisto/content-docs/pull/292


yuvalbenshalom (approved):

yaakovi (approved):

bakatzir:
I think I've forgot to regard the `value` and the fact that the indicators can contain their own category and message.
The UT made it clearer for me.
please re-review.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2575,6 +2575,65 @@ def to_context(self):
...
+            if isinstance(indicator, dict):
+                if 'Value' in indicator.keys():
+                    timeline['Value'] = indicator['Value']
+
+                if 'Category' in indicator.keys():
+                    timeline['Category'] = indicator['Category']
+                elif category:
+                    timeline['Category'] = category
+                else:
+                    timeline['Category'] = 'Integration Update'
```
anara123, line 2614: We can also add a condition to check if this integration or automation, and update the category accordingly

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2575,6 +2575,65 @@ def to_context(self):
...
+    :param category: indicator category.
+    :type message: ``str``
+
+    :param message: indicator message.
+    :return: a list of indicators
+
+    :return: None
+    :rtype: ``None``
+    """
+    def __init__(self, indicators=None, category=None, message=None):
```
anara123, line 2595: the interface here is not clear. because indicators could contain strings or dicts.
I would suggest CommandResults to have `List[IndicatorsTimeline]`
`indicators` will contain only list of strings. indicators and message should be mandatory params.
If some user want to have message per indicator, he will create list of IndicatorTimeline objects (rare case).

bakatzir:
@anara123 more comments?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2641,6 +2692,7 @@ def to_context(self):
         else:
             human_readable = None  # type: ignore[assignment]
         raw_response = None  # type: ignore[assignment]
+        indicators_timeline = IndicatorsTimeline().indicators_timeline  # type: ignore[assignment]
```
anara123, line 2695: why you need this line? remove it and use directly the `self.indicators_timeline.indicators_timeline`

anara123 (approved):
see my comment

PR #8922, opened 2020-09-12
Set delete context to d bot role

yaakovi:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27980

## Description
Set script role as DBot

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_2_41.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_2_41.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### DeleteContext
+- Set script role as DBot.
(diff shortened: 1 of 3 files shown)


dorsha (approved):

aviadl:
I think there is also a bug in the delete all flow
Where it doesn’t show an error if it fails ....

yaakovi:
> I think there is also a bug in the delete all flow Where it doesn’t show an error if it fails ....

@aviadl 
in case of an error, it will print (as a regular entry the following message:
`Context cleared with the following errors: <list of errors>`

should this be replaced with an error entry? (it will be considered as a BC break)

kirbles19:
@yaakovi Done.

PR #8958, opened 2020-09-15
ThreatQ Integration 1.0.6

robert-phe:
robert.asztalos@3pillarglobal.com

## Status
- [x] In Progress
- [ ] Ready

## Description
- Removed threatq-search-by-name command
- Fix default parameters for the following commands:
    - threatq-search-by-name
    - threatq-upload-file
    - threatq-get-all-indicators
    - threatq-get-all-events
    - threatq-get-all-adversaries

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
+++ b/Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
@@ -640,7 +640,7 @@ def aggregate_search_results(indicators, default_indicator_type, generic_context
 def search_by_name_command():
     args = demisto.args()
     name = args.get('name')
-    limit = args.get('limit') or '10'
+    limit = args.get('limit', '10')
 
     if limit and isinstance(limit, str) and not limit.isdigit():
         return_error('limit argument must be an integer.')
@@ -1044,7 +1044,7 @@ def upload_file_command():
     args = demisto.args()
     entry_id = args.get('entry_id')
     title = args.get('title')
-    malware_safety_lock = args.get('malware_safety_lock') or 'off'
+    malware_safety_lock = args.get('malware_safety_lock', 'off')
     file_category = args.get('file_category')
 
     file_info = demisto.getFilePath(entry_id)
@@ -1103,8 +1103,8 @@ def download_file_command():
 
 def get_all_objs_command(obj_type):
     args = demisto.args()
-    page = int(args.get('page')) or 0
-    limit = int(args.get('limit')) or 50
+    page = int(args.get('page', 0))
+    limit = int(args.get('limit', 50))
     if limit > 200:
         limit = 200
 
--- a/Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
+++ b/Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
@@ -617,6 +617,25 @@ def test_module():
         demisto.results('ok')
 
 
+def get_indicator_type_id(indicator_name: str) -> str:
+    indicator_types_res = tq_request(
+        method='GET',
+        url_suffix='/indicator/types',
+        retrieve_entire_response=True
+    )
+    try:
+        indicator_types = indicator_types_res.json().get('data')
+    except ValueError:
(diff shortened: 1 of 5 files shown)


Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
```diff
@@ -1186,8 +1103,8 @@ def download_file_command():
 
 def get_all_objs_command(obj_type):
     args = demisto.args()
-    page = int(args.get('page'))
-    limit = int(args.get('limit'))
+    page = int(args.get('page')) or 0
+    limit = int(args.get('limit')) or 50
```
reutshal, line 1107:
```suggestion
    page = int(args.get('page', 0))
    limit = int(args.get('limit'), 50)
```

Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
```diff
@@ -1127,7 +1044,7 @@ def upload_file_command():
     args = demisto.args()
     entry_id = args.get('entry_id')
     title = args.get('title')
-    malware_safety_lock = args.get('malware_safety_lock')
+    malware_safety_lock = args.get('malware_safety_lock') or 'off'
```
reutshal, line 1047:
```suggestion
    malware_safety_lock = args.get('malware_safety_lock', 'off')
```

Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
```diff
@@ -656,74 +637,10 @@ def aggregate_search_results(indicators, default_indicator_type, generic_context
...
-    )
-
-    return_outputs(readable, entry_context, search_results)
-
-
 def search_by_name_command():
     args = demisto.args()
     name = args.get('name')
-    limit = args.get('limit')
+    limit = args.get('limit') or '10'
```
reutshal, line 643:
```suggestion
    limit = args.get('limit', '10')
```

reutshal (commented):
Hi, thanks for contributing this!
Overall looks good, few things - 
* I am not clear why did you remove the command?
This could cause BC breaking for customers using this.
* I'm suggesting adding the default values to the arguments in the yml files as well, that way the customers will be aware of the default values.

Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
robert-phe: 2bd88f9

Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
robert-phe: 2bd88f9

Packs/ThreatQ/Integrations/ThreatQ_v2/ThreatQ_v2.py
robert-phe: 2bd88f9

robert-phe:
Hello,

- The command was removed because it uses the TQ deprecated endpoint "/search/advanced".
- As far as I saw the default values are added to the yaml and displayed in the UI, but it seems that the args.get(<param>) returns None even if the default is set in the yaml

reutshal:
> Hello,
> 
> * The command was removed because it uses the TQ deprecated endpoint "/search/advanced".
> * As far as I saw the default values are added to the yaml and displayed in the UI, but it seems that the args.get(-param-) returns None even if the default is set in the yaml

In that case, sounds like the command should be deprecated rather than removed from the integration.
You can easily do that by adding: `deprecated: true` under this command in the yml file and then customers can still use it but they will know it is deprecated and will not be supported.

robert-phe:
As discussed, re-added the code for `threatq-advanced-search` and added the `deprecated: true` in the yaml file.
Note: The `threatq-advanced-search` is still removed from the readme file.
9badb2e

reutshal (approved):

PR #8968, opened 2020-09-15
demistomock - added the incident() method to the file as it was added to python looper

reutshal:
## Status
- [] In Progress
- [X] Ready

## Description
added the incident() method to the file as it was added to python looper


Tests/demistomock/demistomock.py
```diff
@@ -741,6 +741,20 @@ def incidents(incidents=None):
         )
 
 
+def incident(incident=None):
```
Itay4, line 744: the func does not take any argument and behaves the same in both scripts and integrations - it returns the incident it is ran in

Itay4 (changes requested):

Tests/demistomock/demistomock.py
reutshal: Sorry, my bad

Tests/demistomock/demistomock.py
```diff
@@ -741,6 +741,16 @@ def incidents(incidents=None):
         )
 
 
+def incident():
+    """Retrieves the current incident
```
Itay4, line 745: @kirbles19 please review the docstring here

it will be presented in https://xsoar.pan.dev/docs/reference/articles/demisto-class

yaakovi (approved):

kirbles19:
Looks good.

Itay4 (approved):

PR #8975, opened 2020-09-15
[NONE] Credentials, engine, and enable/disable integration options

akusei:
## Status
- [ ] In Progress
- [x] Ready

Although this works, it may be a good idea to separate `enabled` and `engine` outside of the `params` node to avoid integration param name collisions but that would require changing the signature of method `__create_integration_instance`. If everyone is in agreement and no one minds the method signature changing, I can make the change.

## Related Issues
None that I know of

## Description
The test_integration script can be used to create/update an integration instance but in a very limited way. This PR addresses this issue and allows for configuring stored credentials, enabling/disabling instances, and selection of an xsoar-engine.

## Example AD integrations configured in conf.json
The following examples illustrate the new `engine` and `enabled` fields as well as shows how to use both stored credentials and username/password credentials

Example showing use of stored credentials
```
{
  "integrations": [
    {
      "name": "Active Directory Query v2",
      "params": {
        "engine": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
        "enabled": true,
        "integrationInstanceName": "instance_name_to_create_or_update",
        "base_dn": "dc=domain,dc=com",
        "credentials": {
          "credential": "stored_credential_id",
          "passwordChanged": true
        },
        "ntlm": false,
        "page_size": "500",
        "port": "636",
        "secure_connection": "SSL",
        "server_ip": "ad.server.tld",
        "unsecure": false
      }
    }
  ]
}
```

Example conf.json using username/password
```
{
  "integrations": [
    {
      "name": "Active Directory Query v2",
      "params": {
        "engine": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
        "enabled": true,
        "integrationInstanceName": "instance_name_to_create_or_update",
        "base_dn": "dc=domain,dc=com",
        "credentials": {
	  "identifier": "username",
	  "password": "password",
          "passwordChanged": true,
        },
        "ntlm": false,
        "page_size": "500",
        "port": "636",
        "secure_connection": "SSL",
        "server_ip": "ad.server.tld",
        "unsecure": false
      }
    }
  ]
}
```
## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

This was tested with 6.0.0 but it should work with anything the test_integrations.py script works with

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Tests/test_integration.py
+++ b/Tests/test_integration.py
@@ -506,14 +506,18 @@ def __create_integration_instance(client, integration_name, integration_instance
                     'validate "Test": {})'.format(integration_name, instance_name, validate_test)
     prints_manager.add_print_job(start_message, print, thread_index)
 
+    # allow enabling/disabling and choosing an xsoar engine by UUID
+    engine = integration_params.get('engine', '')
+    enabled = integration_params.get('enabled', 'true')
+
     # define module instance
     module_instance = {
         'brand': configuration['name'],
         'category': configuration['category'],
         'configuration': configuration,
         'data': [],
-        'enabled': "true",
-        'engine': '',
+        'enabled': str(enabled),
+        'engine': engine,
         'id': '',
         'isIntegrationScript': is_byoi,
         'name': instance_name,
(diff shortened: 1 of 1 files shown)


ShahafBenYakir:
Hi @akusei why do you want to make these changes? What is that you are trying to achieve?

ShahafBenYakir:
@akusei ?

akusei:
This was added for our use case. We were given a script by Bharathwaj Tiruputkuzhi that was missing a lot of functionality in terms of automating the management of integration instances. He also pointed us to this script and this method in particular for creating an instance but again it lacks basic features that you'd expect from a function that allows creation of an integration. So in response we added a very small code change that adds the ability to enable/disable an integration on creation, choose an engine, and use stored credentials. Without those features, it was useless to us.

Since then we've had time to work with the UI API, this content repo and the script Bhharathwaj provided and found it just lacking in general so we ended up scrapping the use of this repo and the script he gave us in favor of one we wrote from scratch which works much better for our environment. If you'd like to close this without merge, I'm fine with that otherwise maybe this small change could be useful for someone else struggling with the 0 documentation API and limited functionality of this method.

ShahafBenYakir:
Closing this as this should not be part of a content pack

PR #8989, opened 2020-09-16
GetIndicatorDBotScore - Get vendors from sourceBrands

adi88d:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28481

## Description
Use the array `sourceBrands` to gets indicator vendor names from indicator data.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/GetIndicatorDBotScore/GetIndicatorDBotScore.py
```diff
@@ -30,16 +30,26 @@
...
+                "Vendor": source,
+                "Score": score
+            })
+    else:
+        ec["DBotScore"].append({
+            "Indicator": indicator,
+            "Type": indicator_type,
+            "Vendor": '',
+            "Score": score
+        })
```
yaakovi, line 51: why would it happen?

adi88d: Im not sure why but its possible situation, i had some in my local env:
image

kirbles19:
@adi88d Done.

yaakovi (approved):

Packs/CommonScripts/Scripts/GetIndicatorDBotScore/GetIndicatorDBotScore.py
Nitzperetz, line 51: @anara123 , the new script is good, but still, seems that when there are not sources, we still add empty DBotScore - why? Still needed?

anara123: Yes, the indicator can have no sources and still exists in indicators page. We want to return that indicator too.

Nitzperetz: and we have to return it on the DBotScore?
It could also be wrong, DBot didn't set it, it could have been set manually.


If you think it could be changed in content let me know. Otherwise we would add the extra prop.

anara123: Either way we need to add new prop, plan to implement an integration that will query XSOAR indicators database and return DBotScore as any other threat intel integration.

PR #9011, opened 2020-09-18
[Documentation] Expanded demistomock docstring to emphasise incidents keys

DeanArbel:
## Status
- [x] Ready

## Description
Added docstring detailing the fields that are expected to be in incidents when calling `demisto.createIncidents`, with emphasis on `type` - as it's not required by the server, but should be provided.

### DIFF
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -863,7 +863,7 @@ def createIncidents(incidents, lastRun=None, userID=None):
         List of incident objects to create, with the following required keys, and some common optional keys
             - name (required) - str
             - type (required) - if not provided will create unclassified incident) - str
-            - labels (required) - list of {"type": _, "value": _} objects
+            - labels (optonal) - list of {"type": _, "value": _} objects
             - rawJSON (required) - str
             - occurred (optional) - str
             - details (optional) - str
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -859,7 +859,12 @@ def createIncidents(incidents, lastRun=None, userID=None):
     Creates incident in long running execution
 
     Args:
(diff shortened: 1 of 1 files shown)


Tests/demistomock/demistomock.py
```diff
@@ -859,7 +859,15 @@ def createIncidents(incidents, lastRun=None, userID=None):
     Creates incident in long running execution
 
     Args:
-      incidents (list): List of incident objects to create
+      incidents (list):
+        List of incident objects to create, with the following required keys, and some common optional keys
+            - name (required) - str
+            - type (required) - if not provided will create unclassified incident) - str
```
yaakovi, line 865: so not required?

Tests/demistomock/demistomock.py
```diff
@@ -859,7 +859,15 @@ def createIncidents(incidents, lastRun=None, userID=None):
...
+            - labels (required) - list of {"type": _, "value": _} objects
```
yaakovi, line 866: are you sure labels are required?

Tests/demistomock/demistomock.py
DeanArbel: it's not required by the server, but we don't want to create incidents that are unclassified, so it's "required" 🙂

Tests/demistomock/demistomock.py
DeanArbel: You're right, they are not required. fixing via suggestion:
```suggestion
            - labels (optonal) - list of {"type": _, "value": _} objects
```

Tests/demistomock/demistomock.py
DeanArbel:
```suggestion
            - type (required - if not provided will create unclassified incident) - str
```

yaakovi (approved):

PR #9012, opened 2020-09-18
Hackaton - Tidy project

GalRabin:
## Status
- [x] In Progress
- [ ] Ready

## Description
Only the First draft.

## Must have
- [ ] Tests
- [ ] Documentation


hod-alpert (changes requested):
Looks good!!
a few comments:
1. In both `tidy-python` and `tidy-golang` we have a parameter `globals` which can be given as array of globals, are we sure this is possible? AFAIK there can be only one global configured at a time.
2. In both `tidy-python` and `tidy-golang` since we're talking about limited amount of versions I think I would make it PRE-DEFINED with all relevant versions (for both globals and versions)
3. What happened to the profiles and users we discussed? Aren't they supposed to be there too?
4. Installing go, python and cloning git repos are not enough IMO, this is not the time-consuming part and I think we should be able to do more then just that to be valuable. For example
In the internal demisto ansible you have also installed bashrc to zshrc with global variables and useful aliases and also some git config was there. I think this can be the game changer or at least part of it.

GalRabin:
# Tasks:
## @GalRabin 
- [ ] Python
- [ ] GoLang
- [ ] Git clones
- [ ] Github ssh key

## @hod-alpert 
- [ ] Env variables, bash_profile, zshrc
- [ ] ZSH

- [ ] Home brew

hod-alpert:
Merged to tidy/master

PR #9117, opened 2020-09-24
WhereFieldsEquals - optional stringify

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28613

## Description
exposed the stringify as an optional choice.


Packs/CommonPlaybooks/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
```diff
@@ -255,6 +255,9 @@ tasks:
               getField:
                 value:
                   simple: Address
+              stringify:
+                value:
+                  simple: "false"
```
Itay4, line 260: this might be problematic as users count it to return as string atm
perhaps make it a playbook input?

yaakovi: it's used as a transformer to the email address input of `ad-get-user` (and before the `unique` transformer).
in the current state, the `unique` doesn't work because it doesn't get a list so it passes it as is.
next, the command gets a string representation of a list. it doesn't support arrays and doesn't work as expected (see opened bug).

with this fix, the `unique` works and the command gets an actual list so it is triggered multiple times, once per input.

Itay4 (approved):

roysagi:
@yaakovi - Docs look good no chanages made.

PR #9153, opened 2020-09-29
Create dedicated client before applying proxy settings for mock tests

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: Internal

## Description
Create a new client before applying proxy settings


Tests/mock_server.py
```diff
@@ -194,12 +195,18 @@ def __init__(self, public_ip,
...
-    def configure_proxy_in_demisto(client, proxy=''):
-        system_conf_response = demisto_client.generic_request_func(
-            self=client,
-            path='/system/config',
-            method='GET'
-        )
+    def configure_proxy_in_demisto(username, password, server, prints_manager, thread_index, proxy=''):
+        client = demisto_client.configure(base_url=server, username=username,
+                                          password=password, verify_ssl=False)
+        try:
```
yaakovi, line 201: no need for the try-except, in case this fails we want to stop.
otherwise, the rest of the tests will run without proxy

Tests/mock_server.py
```diff
@@ -215,9 +222,12 @@ def configure_proxy_in_demisto(client, proxy=''):
             'data': system_conf,
             'version': -1
         }
-        response = demisto_client.generic_request_func(self=client, path='/system/config',
-                                                       method='POST', body=data)
-        # client.api_client.pool.close()
+        try:
```
yaakovi, line 225: no need for the try-except, in case this fails we want to stop.
otherwise, the rest of the tests will run without proxy

yaakovi (approved):

Tests/mock_server.py
amshamah419: I need to catch the error to prevent the headers from displaying. Would executing a `sys.exit()` work?

yaakovi: we are catching the exception at a higher level, handle it there

PR #9158, opened 2020-09-29
Fix for DomainSquatting task in Email Address Enrichment 2.1 (WhereFieldEquals)

idovandijk:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - Can't update pack due to SDK bug with "--all" parameter as well as specific pack which indicates that I'm behind master https://github.com/demisto/etc/issues/28871

## Related Issues
fixes: https://github.com/demisto/etc/issues/27290

## Description
Fixed the DomainSquatting task which uses `WhereFieldEquals` after the transformer was added the `Stringify` parameter. Changed the argument to false to allow multiple results in the transformer.

No


ronykoz (approved):

ronykoz:
@idovandijk let's add RN and merge this

idovandijk:
> @idovandijk let's add RN and merge this

I'm getting errors when running update-release-notes both when giving the pack path and the pack name, with or without the `-u minor` argument
image

Additionally I see that this is also open demisto/etc#28871 so I can't see how to use the SDK to merge this

ronykoz:
Can you use `Email Address Enrichment - Generic v2.1` and create the RN manually for now?
@roysagi @Itay4 let's prioritize this bug

idovandijk:
@kirbles19 can you please review the following RN? 
"Fixed an issue where the emails were not checked for domain-squatting due to a problem with the WhereFieldEquals transformer."

PR #9163, opened 2020-09-29
Common Server Python - import errors not caught in python 2

orlichter1:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/28646

## Description
In python 2 import typing raised an error even though it was handled. Fixed it and removed a code that caused scripts to fail. The code was up to 4.5

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


orlichter1:
@Itay4 - @yaakovi asked me to tag you :)

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -333,24 +337,6 @@ def auto_detect_indicator_type(indicator_value):
     return None
 
 
-# ===== Fix fetching credentials from vault instances =====
```
Itay4, line 336: @liorblob is it save to remove this ?

liorblob: I don't think so... not sure it was handled on server side

orlichter1: @liorblob From what @yaakovi and I saw this was up to 4.5. Isn't that the case?

liorblob: what do you mean by "up to 4.5"?

orlichter1: This code was added in CommonServerPython4.1 as can be [seen here](https://github.com/demisto/content/pull/3361/files).

liorblob: That file doesn't exist anymore... so if you remove that it will not work for versions under 4.5

orlichter1: @Itay4 are we safe on that front because of the feature branch?

yaakovi: @orlichter1 yes, 4.5 and below will still get this code via the feature branch.
@liorblob as long as this is fixed for version 5.0.0 on the server side - we can remove it from here

liorblob: should be ok then

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -29,6 +29,10 @@
     from urllib3.util import Retry
     from typing import Optional, List, Any
 except Exception:
+    if sys.version_info[0] < 3:
+        # in python 2 an exception in the imports might still be raised even though it is caught.
+        # for more info see https://cosmicpercolator.com/2016/01/13/exception-leaks-in-python-2-and-3/
+        sys.exc_clear()
     pass
```
yaakovi, line 36: remove the `pass`

yaakovi (approved):

orlichter1:
@roysagi Could you have a quick look at the release notes?

PR #9250, opened 2020-10-01
Azure sentinel enhancement - `azure-sentinel-get-entity-by-id`

ChanochShayner:
In case you are a technological partner:

If you are not a technological partner (i.e. PANW employee, customer or individual developer), please delete this section.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26936

## Description
Update the command "azure-sentinel-get-entity-by-id" to retrieve more information about the entity.

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -405,19 +405,21 @@ def incident_add_comment_command(client, args):
 
 def get_entity_by_id_command(client, args):
     entity_id = args.get('entity_id')
-    url_suffix = f'entities/{entity_id}'
-
-    result = client.http_request('GET', url_suffix, is_get_entity_cmd=True)
+    expansion_id = "98b974fd-cc64-48b8-9bd0-3a209f5b944b"
```
guykeller, line 408: why is this hard coded?

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -405,19 +405,21 @@ def incident_add_comment_command(client, args):
...
+    url_suffix = f'entities/{entity_id}/expand'
+    data = {
+        "expansionId": expansion_id
+    }
 
-    flattened_result = flatten_entity_attributes(result)
+    result = client.http_request('POST', url_suffix, is_get_entity_cmd=True, data=data)
```
guykeller, line 414: Shouldn't you combine the new field with the previous values? 
Are you currently just throwing the full object into the context?

Packs/AzureSentinel/ReleaseNotes/1_0_3.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Azure Sentinel (Beta)
+- Update the command "azure-sentinel-get-entity-by-id" to retrieve more information about the entity.
```
guykeller, line 4:
```suggestion
- Updated the command ***azure-sentinel-get-entity-by-id*** to retrieve more information about the entity.
```

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
guykeller: Moreover, why didn't we modify the yml as well?

guykeller (changes requested):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
ChanochShayner: Take a quick look at this - https://techcommunity.microsoft.com/t5/azure-sentinel/get-entities-for-a-sentinel-incidient-by-api/m-p/1422643

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
ChanochShayner: We do not have events in the system to check what is coming back.
That's why I'm not entirely sure of my answers.
I think the current implementation also has the information that was in the previous implementation and more information.

guykeller (approved):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
guykeller: Why did you resolve this? @ChanochShayner 
@dantavori Would love your advice here

guykeller (changes requested):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
dantavori: @ChanochShayner this might help:
* current command's outputs - https://github.com/Azure/azure-rest-api-specs/blob/master/specification/securityinsights/resource-manager/Microsoft.SecurityInsights/preview/2019-01-01-preview/examples/entities/GetSecurityAlertEntityById.json
* `/expand` outputs - https://github.com/Azure/azure-rest-api-specs/blob/master/specification/securityinsights/resource-manager/Microsoft.SecurityInsights/preview/2019-01-01-preview/examples/entities/expand/PostExpandEntity.json

note that the `/expand` example it's an *account* entity (and not a security alert), but if the structure is the same for every entity type then it looks like the two responses above have the same structure (id, name, type, kind and properties) and what differentiates them is the "properties" value (I guess each one has different attributes). So maybe the right thing to do is to combine the properties from the two responses.

However, this is just an advice and the best thing to do is to try the endpoint with data. Did you manage to get something with @Arsenikr ?

guykeller (changes requested):

ChanochShayner:
@dantavori @guykeller There was a meeting with the customer this morning,
The part we added to the integration works well.
Ready for review.
Thanks.

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -97,6 +97,10 @@ def format_date(date):
     return dateparser.parse(date).strftime(DATE_FORMAT)
 
 
+def str_to_bool(string_var):
+    return True if string_var == 'true' else False
```
guykeller, line 101:
```suggestion
    return string_var.lower() == 'true'
```

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -97,6 +97,10 @@ def format_date(date):
     return dateparser.parse(date).strftime(DATE_FORMAT)
 
 
+def str_to_bool(string_var):
```
guykeller, line 100: `string_var` - not a good name

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -405,26 +409,54 @@ def incident_add_comment_command(client, args):
...
+            'AzureSentinel.Entity(val.ID === obj.ID)': flattened_result
+        }
+
+        return (
+            readable_output,
+            outputs,
+            result
+        )
+    else:
+        expansion_id = "98b974fd-cc64-48b8-9bd0-3a209f5b944b"
```
guykeller, line 436: Please add a comment explaining why this should be hard coded like this

guykeller (approved):
Approving with notes - please fix them before merging

PR #9263, opened 2020-10-01
Fix for Private Repo Pack Installation

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: Internal

## Description
The build was searching for packs to install by the display name. Private repo does not use display names for installations.


Tests/Marketplace/search_and_install_packs.py
```diff
@@ -269,17 +280,22 @@ def search_pack_and_its_dependencies(client, prints_manager, pack_id, packs_to_i
         lock (Lock): A lock object.
     """
     pack_data = []
-
     if pack_id not in packs_to_install:
         pack_display_name = get_pack_display_name(pack_id)
         if pack_display_name:
             pack_data = search_pack(client, prints_manager, pack_display_name, thread_index, lock)
+        if pack_data is None:
```
ronykoz, line 287: when will it happen?

ronykoz (approved):

Tests/Marketplace/search_and_install_packs.py
amshamah419: What I found is that if it's a new pack, or premium pack, and doesn't currently exist in the marketplace bucket, it won't find an `id` or `version`. This was a way of keeping the existing install logic while remedying that.

PR #9308, opened 2020-10-05
SNow v2 fix empty credentials error

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29126

## Description
Fixing an issue where an empty username/password in the instance configurations would result in a non-indicative error.

   - [ ] Yes
       - Further details:
   - [x] No


Packs/ServiceNow/ReleaseNotes/1_3_3.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### ServiceNow v2
+- Fixed a bug where empty username or password in the instance configurations would result in an error.
```
bakatzir, line 4:
```suggestion
- Fixed a bug issue empty username or password in the instance configuration resulted in a non informative error.
```

bakatzir (commented):
How does the error look now? attach a screenshot please.
also, add an rn regarding the dockerimage update.

darkushin:
@bakatzir 
> How does the error look now? attach a screenshot please.
> also, add an rn regarding the dockerimage update.

image
The error message is a bit long, but it indicates the correct reason for the error in the beginning: `User Not Authenticated.. Required to provide Auth information.`

bakatzir (approved):
legit. and clearer than before.

PR #9351, opened 2020-10-07
Microsoft Defender Advanced Threat Protection - rm try-except from test module

Itay4:
## Status
- [x] Ready

## Description
Removed `try-except` from the test module that swallowed the exception raised

   - [x] No


yaakovi (approved):
why was it there in the first place? how the error looks now? is it indicative?

kirbles19:
@Itay4 Done

Itay4:
> why was it there in the first place? how the error looks now? is it indicative?

not sure why it was like that
the error will look like in any other microsoft integration, and it is as indicative as microsoft returns
for example
image

PR #9404, opened 2020-10-11
added : except ValueError in the right place in order to show a relev…

abaumgarten:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29048

## Description
An except ValueError was added in order to show a relevant error when occured

## Minimum version of Demisto
- [x] 5.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
+++ b/Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
@@ -45,6 +45,8 @@ def http_request(method, url_suffix, params={}, data=None):
                 f'Error in API call to Sentinel One [{res.status_code}] - [{res.reason}] \n'
                 f'Error details: [{errors}]'
             )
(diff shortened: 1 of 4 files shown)


Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -45,6 +45,8 @@ def http_request(method, url_suffix, params={}, data=None):
                 f'Error in API call to Sentinel One [{res.status_code}] - [{res.reason}] \n'
                 f'Error details: [{errors}]'
             )
+        except ValueError as value_error:
+            raise value_error
```
guykeller, line 49: I think that they meant that no matter what the error is, don't just return the error but add details about it.
Do you understand what I mean?

guykeller (approved):

PR #9417, opened 2020-10-12
Added fetch incidents capability

spearmin10:
## Status
- [] In Progress
- [x] Ready

## Description
Added fetch incidents capability to the Symantec ATP integration

## Minimum version of Demisto
- [ ] 5.0.0
- [x] 5.5.0
- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
+++ b/Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
@@ -1,41 +1,67 @@
 commonfields:
   id: Symantec Advanced Threat Protection
   version: -1
+vcShouldKeepItemLegacyProdMachine: false
 name: Symantec Advanced Threat Protection
 display: Symantec Advanced Threat Protection
 category: Endpoint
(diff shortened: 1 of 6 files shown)


Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
```diff
@@ -261,14 +275,117 @@ script:
...
+            var n = now;
+            n.setTime(now.getTime() - 1 * DAY_IN_MILLIS);
+            startTime = n.toISOString();
+        }
+        var next = false;
+        var p = {verb: 'query'};
+        add(p, 'where', query);
+        addTime(p, 'start_time', startTime);
+        addTime(p, 'end_time', endTime);
+        if (limit) {
```
orlichter1, line 360: What happens if no limit is inserted as a parameter? This would mean no limit would be set? If so, this could potentially return hundreds of incidents to XSOAR. Please add a default limit of a 200 even if it isn't a parameter.

Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
```diff
@@ -261,14 +275,117 @@ script:
...
+            var res = doReq('POST', type, p);
+            if (res.obj.result && res.obj.result.length !== 0) {
+                if (type == 'incidents') {
+                    incidents = incidents.concat(getIncidentsFromIncidentsResult(res.obj.result));
+                } else {
+                    incidents = incidents.concat(getIncidentsFromEventsResult(res.obj.result));
+                }
+                next = 'next' in res.obj ? res.obj.next : null;
+            }
+        } while (next && limit && incidents.length < limit);
```
orlichter1, line 376: Not sure I understand the logic of the `do.. while` loop. In what scenario will we do the loop more than once? If there is a limit set then the number of results we will get back from the API call will be less or equals to the limit.
If it's less, then no `next` will return. If it's equal, then `next` will be returned but `incidents.length == limit` won't it?
What I suggest is keep the `next` parameter in the `setLastRun` and in the next fetch use it to fetch the next batch of incidents, do this until there's no `next` parameter then proceed as usual. We're using this logic in a few integrations like [Splunkpy](https://github.com/demisto/content/blob/e1309967af65fdca92ee0035487030092acbb8d4/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py#L505) if you want to have a look.

I'm happily available on the DFIR channel for any questions about this.

Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
```diff
@@ -1,42 +1,68 @@
...
+  required: false
+  options:
+  - incidents
+  - events
+  - incidentevents
+- display: Query string for fetch incidents
+  name: fetch_incidents_query
+  type: 0
+  required: false
+- display: Maximum number of events per fetch.
```
orlichter1, line 42: Please change to `Fetch Limit (Maximum 200)`

orlichter1 (changes requested):

Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
spearmin10: The default limit is defined in `fetch_incidents_limit` configuration (default is 20). I adjusted the style of `limit` according to other functions in the integration.

orlichter1: What I mean is if someone removes the `fetch_incidents_limit` completely, the `params.fetch_incidents_limit` will return an empty string (`""`) and will not enter the if statement. If I'm not mistaken this will cause the request to bring all the incidents without any limit. 
Screen Recording 2020-10-14 at 12 54 24

spearmin10:
Pull back

PR #9421, opened 2020-10-12
add node modules setup to .bootstrap

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Add node modules setup to bootstrap so README validation will work out of the box.

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.hooks/bootstrap
+++ b/.hooks/bootstrap
@@ -35,6 +35,10 @@ if [ "$1" == "-u" -o "$1" == "--update" ]; then
         echo "Updating python 3 dependencies ..."
         venv/bin/pip3 install -r dev-requirements-py3.txt
     fi
+    if [ -d "node_modules" ]; then
+        echo "Updating node modules"
+        npm install
+    fi
     echo "Done updating dependencies"
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


.hooks/bootstrap
```diff
@@ -130,3 +134,18 @@ else
     fi
 fi
 
+echo "======================="
+
+if [ -n "$NO_NPM" ]; then
+    echo "Skipping npm node modules setup as environment varialbe NO_NPM is set"
+else
+    if ! command -v npm >/dev/null 2>&1; then
```
anara123, line 142: you can add this code in update flow too, in case user already have hooks and he runs ./hooks/bootstrap -u

anara123 (approved):

.hooks/bootstrap
glicht: It's there. At line 38. We just check if the node_modules dir exists as an indication.

anara123: but if all the developers that didn't install node_modules and have bootstrap installed, when they will run `.hooks/bootstrap -u` , the process will try to find `node_modules` and will not find it.
It is not a big deal, developers can delete the .venv and install the hook again

glicht: It is better to do accordingly to the folder. So if someone choose to not install we don't do this every time. Devs can run nom install if needed.

PR #9474, opened 2020-10-14
Phishing preprocessing rule

lperryp:
## Description
- Updated dedup logic to close duplicate incidents and link it to the earliest duplicate incident (based on incident id logic)
- Added support for custom type field 
- Added preprocessing of URLS for improving deduplication

## Related Issues
fixes: https://github.com/demisto/etc/issues/29263

## Status
- [x] In Progress
- [ ] Ready

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


lperryp:
@kirbles19 - when this is reviewed, can we also go over the error messages in the script?

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -47,16 +50,22 @@ def get_existing_incidents(input_args):
...
     status_scope = input_args.get('statusScope', 'All')
+    query_components = []
+    if 'query' in get_incidents_args:
+        query_components.append(get_incidents_args['query'])
     if status_scope == 'ClosedOnly':
-        if 'query' in get_incidents_args:
-            get_incidents_args['query'] = '({}) and (status:Closed)'.format(get_incidents_args['query'])
-        else:
-            get_incidents_args['query'] = 'status:Closed'
+        query_components.append('status:Closed')
```
erezh31, line 57: I think closed need to be lower case

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -47,16 +50,22 @@ def get_existing_incidents(input_args):
...
     elif status_scope == 'NonClosedOnly':
-        if 'query' in get_incidents_args:
-            get_incidents_args['query'] = '({}) and (-status:Closed)'.format(get_incidents_args['query'])
-        else:
-            get_incidents_args['query'] = '-status:Closed'
+        query_components.append('-status:Closed')
```
erezh31, line 59: I think closed need to be lower case

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -47,16 +50,22 @@ def get_existing_incidents(input_args):
...
+    elif status_scope == 'All':
+        pass
+    else:
+        return_error('Unsupported statusScope: {}'.format(status_scope))
+    incident_type = input_args.get('incidentTypes', DEFAULT_ARGS['incidentTypes'])
```
erezh31, line 64: I don't think we need to force incident type,  it it's empty just skip this part in the query

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -47,16 +50,22 @@ def get_existing_incidents(input_args):
...
+    type_field = input_args.get('incidentTypeFieldName', 'type')
```
erezh31, line 65: I think we can also give up on this part, we have the extra query argument for cases which would like to have different field for type

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -131,6 +149,8 @@ def preprocess_incidents_df(existing_incidents):
 def filter_out_new_incident(existing_incidents_df, new_incident):
     same_id_mask = existing_incidents_df['id'] == new_incident['id']
     existing_incidents_df = existing_incidents_df[~same_id_mask]
+    earlier_incidents_mask = existing_incidents_df['id'].astype(int) < int(new_incident['id'])
```
erezh31, line 152: you need to consider that the ID of the incidents gonna change because we are moving to ES (elasticsearch).
Wrap with try-catch covenanting the id to int.
When you sort use created & id column - it will always return same incident as first.
if there is a few with the same time -it will just get the id with smaller id (if you success parse to int it will the most early ingested one)
when we move to ES ids not gonna be INT anymore... :(

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -204,23 +222,28 @@ def main():
...
+    demisto.results('found {} incidents by query'.format(len(existing_incidents)))
     if len(existing_incidents) == 0:
         create_new_incident()
         return
     new_incident = demisto.incidents()[0]
     new_incident_df = preprocess_incidents_df([new_incident])
+    if len(new_incident_df) == 0:  # len(new_incident_df)==0 means new incident is too short
+        create_new_incident()
+        return
     existing_incidents_df = preprocess_incidents_df(existing_incidents)
```
erezh31, line 234: I notice that the preprocess_incidents_df doesn't really support the case when the fields are empty - you should validate the fields exist in the current incident (throw error if not) and filter out incidents  where those field does not exists (for the text fields you can do a condition that at least one of them is not empty maybe)

erezh31 (changes requested):

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
lperryp: We want to return error if incident has no text fields (if it's run in the phishing investigation playbook)?
 Currently incidents with no text fields will have short length (1) and will be eliminated by their length

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -32,31 +35,37 @@
...
-            get_incidents_args['query'] = '-status:Closed'
+        query_components.append('-status:closed')
+    elif status_scope == 'All':
+        pass
+    else:
+        return_error('Unsupported statusScope: {}'.format(status_scope))
+    incident_type = input_args.get('incidentTypes')
+    if incident_type is not None:
+        type_field = input_args.get('incidentTypeFieldName', 'type')
+        type_query = '{}:{}'.format(type_field, incident_type)
```
erezh31, line 65: wrap the type value with quates

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -159,42 +193,51 @@ def find_duplicate_incidents(new_incident, existing_incidents_df):
         mask = (existing_incidents_df[FROM_FIELD] != '') & \
                (existing_incidents_df[FROM_FIELD] == new_incident[FROM_FIELD])
         existing_incidents_df = existing_incidents_df[mask]
-    existing_incidents_df.sort_values(by='similarity', ascending=False, inplace=True)
+    existing_incidents_df.sort_values(by='similarity', inplace=True, ascending=False)
```
erezh31, line 196: `try:
    existing_incidents_df['id'] = existing_incidents_df.id.apply(int)
except:
    pass
`

erezh31, line 196: Add sorting by inverse similarity & created & id

erezh31 (changes requested):

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.yml
```diff
@@ -1,15 +1,17 @@
 args:
 - default: false
   defaultValue: type
-  description: The name of the incident field where its type is stored. Default it
-    "type". Change this arguement only in case you use a custom field for specifying
-    incident type.
+  description: A comma-separatetd list of names of the incident fields where incident type
```
erezh31, line 4: here we don't need comma-sep only in the values

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -73,6 +73,17 @@ def get_existing_incidents(input_args):
     return incidents
 
 
+def generate_incident_type_query_component(type_fields_arg, type_values_arg):
+    type_fields = [x.strip() for x in type_fields_arg.split(',')]
```
erezh31, line 77: type should be single field

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
```diff
@@ -193,7 +204,14 @@ def find_duplicate_incidents(new_incident, existing_incidents_df):
         mask = (existing_incidents_df[FROM_FIELD] != '') & \
                (existing_incidents_df[FROM_FIELD] == new_incident[FROM_FIELD])
         existing_incidents_df = existing_incidents_df[mask]
-    existing_incidents_df.sort_values(by='similarity', inplace=True, ascending=False)
+    existing_incidents_df['distance'] = existing_incidents_df['similarity'].apply(lambda x: 1 - x)
```
erezh31, line 207: really nice!!

erezh31 (approved):

kirbles19:
@lperryp Done

Packs/Phishing/Scripts/PhishingDedupPreprocessingRule/PhishingDedupPreprocessingRule.py
lperryp: thanks!

PR #9484, opened 2020-10-14
[Enhancement] Cortex Data Lake - New Fetch Mode: firewall.file_data

DeanArbel:
## Status
- [x] In hold

## Related Issues
fixes: https://github.com/demisto/etc/issues/29102

## Description
- Added new parameter **Fetch Table** - fetch from different types of tables in **fetch-incidents**. Added a new 
    fetch table: **firewall.file_data**. 
- Added new parameter **Fetch Fields** - comma-separated list of fields that'll be fetched with every incident. e.g. 
    "pcap,session_id". Enter "*" for all possible fields.

## Screenshots
image

   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

@kirbles19 Please review:
- Packs/CortexDataLake/ReleaseNotes/1_2_6.md
- Packs/CortexDataLake/Integrations/CortexDataLake/README.md
- Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.yml


DeanArbel:
@kirbles19 Please review:
- Packs/CortexDataLake/ReleaseNotes/1_2_6.md
- Packs/CortexDataLake/Integrations/CortexDataLake/README.md
- Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.yml

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.yml
```diff
@@ -25,6 +25,16 @@ configuration:
...
   required: false
   type: 0
+- additionalinfo: Choose from what table incidents will be fetched from.
+  defaultvalue: firewall.threat
+  display: Fetch Table
+  hidden: false
+  name: fetch_table
+  options:
+  - firewall.threat
+  - firewall.file_data
```
bakatzir, line 35: do we want to support fetching from both tables ?

Packs/CortexDataLake/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "Cortex Data Lake",
     "description": "Palo Alto Networks Cortex Data Lake provides cloud-based, centralized log storage and aggregation for your on premise, virtual (private cloud and public cloud) firewalls, for Prisma Access, and for cloud-delivered services such as Cortex XDR",
     "support": "xsoar",
-    "currentVersion": "1.2.5",
+    "currentVersion": "1.2.6",
```
bakatzir, line 5: consider a minor change and not a revision upon a change/addition in the fetch mechanism.

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.yml
DeanArbel: Simultaneously? there was no decision about it, but it'll be a bit problematic as there is no way to know from what table the incident was fetched once it's fetched

bakatzir: You are right @DeanArbel, maybe in that case we will not allow fields selection to simplify the fetch.
@prasenshelar do we want to support fetch Simultaneously?
or will the user have to have 2 instances for that?

yuvalbenshalom (commented):
looks good, but I think the new files are not checked in the test module.
can we test if the fields exist in the schema in general?
I think this is useful if we use free text.

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake.yml
prasenshelar: @bakatzir we will have to decide this. We can ask users to create separate instances or just keep it in one.
The requirement is to be able to fetch entries from both tables and keep it simple as possible. 

If we can just add a parameter to the entries that says from which table the log was fetched from then that would be great.

yuvalbenshalom (approved):

PR #9501, opened 2020-10-15
fix proxy settings - several integrations

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28638

## Description
Adding handle proxy.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
   - [x] No

### DIFF
--- a/Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
+++ b/Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
@@ -6,6 +6,8 @@ import re
 import requests
 import json
 
+handle_proxy()
+
 RANGE_OPERATORS = ['in-range', 'is-between', 'not-in-range']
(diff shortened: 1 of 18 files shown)


Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -6,6 +6,8 @@
 import requests
 import json
 
+handle_proxy()
```
yaakovi, line 9: better to keep it in main.
initialize the `SESSION` global within the main as well.

yaakovi: after this fix, let's do a quick demo

yaakovi (approved):

PR #9507, opened 2020-10-15
IPv6 support added in search

rozie:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: N/A

## Description
Wanted to have IPv6 support on Urlscan search. Also chaned query type logic (no further checks are made if valid IP address is found).
Didn't test with Demisto versions other than 6.0.0

Please label PR with _hacktoberfest-accepted_ if you accept this PR.

## Screenshots
Screenshot 2020-10-15 at 15 16 01

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [x ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
+++ b/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
@@ -1,9 +1,9 @@
 import demistomock as demisto
 from CommonServerPython import *
 from CommonServerUserPython import *
-from typing import Dict
 
 '''IMPORTS'''
+from typing import Dict
 import time
 import requests
 import collections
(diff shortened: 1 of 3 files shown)


orlichter1 (changes requested):
It's a great idea to support IPv6 in the integration especially if URLScan holds information about that! Changing the type of python or the docker image would cause backward compatibility issues in the system so in order to overcome this and not create a whole new integration, please use `is_ip_valid` from CommonServerPython with `accept_v6_ips=True` this should work in python 2 and not require the docker or python change.

orlichter1 (approved):

PR #9508, opened 2020-10-15
Hod setup with user data

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Description
With [this](https://github.com/demisto/content-test-conf/pull/978) PR merged - there is no need in the `setup instance` step because it will be executed by the instance itself in the creation with the user-data.

Also removed the failure mechanism from the `wait until server ready` script since the implemented timeout does the job instead.


hod-alpert:
@glicht downloading cloud-init file.

Itay4 (approved):

PR #9538, opened 2020-10-18
Generic Webhook Integration

Itay4:
## Status
- [x] Ready

## Description
Generic webhook integration to run in long running container to create incidents from queried events.

## Minimum version of Demisto
- [x] 5.5.0

## Must have
- [x] Tests
- [x] Documentation


anara123:
@Itay4 is there an issue related? Remind me why do we need this integration? Builtin XSOAR webhook API is not enough?

glicht:
A few things I would add to make it more generic:
* Support https (as we do in EDL)
* Authentication move to support credentials so Basic auth is supported and if the user value has a special value of `_header:<header_name>` we use the specified header and password as the value. 
* if there is no `raw_json` use the whole payload as the raw_json value and provide default value for `name`. Think of the case that you want to support github webhook out of the box.

anara123:
@Itay4 lets add the option to save examples for mapping, like you did in CS Streaming

Packs/GenericWebhook/Integrations/GenericWebhook/GenericWebhook.py
```diff
@@ -0,0 +1,74 @@
...
+        except ValueError as e:
+            raise ValueError(f'Invalid listen port - {e}')
+        if demisto.command() == 'test-module':
+            return_results('ok')
+        elif demisto.command() == 'long-running-execution':
+            while True:
+                try:
+                    default_log_stream = StringIO()
+                    log_config = uvicorn.config.LOGGING_CONFIG
+                    log_config['handlers']['default']['stream'] = default_log_stream
```
glicht, line 59: I actually think we should enable the logging handler. It helps debugging all kinds of issues. You can try using: `DemistoHandler`. I guess you will need to add the following line:
```
log_config['handlers']['default']['class'] = 'DemistoHandler`
```
also I would enable the access_log.

glicht: Or try using `IntegrationLogger()` instead of StringIO.

Itay4: i initialize the app logger to be `DebugLogger()` so i think its redundant so set up the logging handler here as well, but i might miss something
lets discuss

ShirleyDenkberg:
@glicht Doc review done.

Packs/GenericWebhook/Integrations/GenericWebhook/GenericWebhook.py
```diff
@@ -0,0 +1,149 @@
...
+
+def fetch_samples() -> None:
+    """Extracts sample events stored in the integration context and returns them as incidents
+
+    Returns:
+        None: No data returned.
+    """
+    integration_context = get_integration_context()
+    sample_events = json.loads(integration_context.get('sample_events', '[]'))
+    incidents = [{'rawJSON': json.dumps(event)} for event in sample_events]
```
anara123, line 86: don't you miss 'name' for the incident?

Itay4: not needed
thats what it looks like:
image

anara123: but we should only show in the samples the rawJSON. THis is how pull from instance shows the samples in regular-fetch incidents

Itay4: i've changed it to return incident object, which includes the name, rawjson, type and occurred fields, as is without modifications - like we do in regular fetch

Itay4:
@kirbles19 @ShirleyDenkberg i've added a disclaimer about the usage of authentication, please review https://github.com/demisto/content/pull/9538/files#diff-b9f9cf77d8b42dbcf55f078a9ecbd681d6ff6c5eb00e8dfd52493fe1a8112fc7R9

kirbles19:
@Itay4 Done

Itay4:
@glicht @anara123 made requested changes, please review

glicht (approved):

Packs/GenericWebhook/Integrations/GenericWebhook/GenericWebhook.yml
```diff
@@ -0,0 +1,61 @@
...
+- display: Private Key (Required for HTTPS)
+  additionalinfo: Required for HTTPS, if not using server rerouting
+  name: key
+  required: false
+  type: 14
+- display: Incident type
+  name: incidentType
+  type: 13
+  required: false
+- additionalinfo: Because this is a push-based integration, it cannot fetch sample events in
```
glicht, line 37: Add a note to disable once done with mapping.

anara123 (approved):

PR #9598, opened 2020-10-21
Jira v2

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29648

## Description
Fixed an issue where the arguments projectKey and issueTypeId where were not specified as mandatory. That issue caused an error when creating a new issue in Jira. Added UT as well.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Jira/Integrations/JiraV2/JiraV2_test.py
```diff
@@ -9,12 +9,48 @@
...
+
+
+def test_create_issue_command_before_fix_mandatory_args_issue(mocker):
+    mocker.patch.object(demisto, 'args', return_value=integration_args_missing_mandatory)
+    from JiraV2 import create_issue_command
+    try:
+        # when there are missing arguments, an Exception is raised to the user
+        create_issue_command()
+    except Exception:
+        assert True
```
teizenman, line 51: You were right to test this case, but you are not getting what you desire that way.
With the current logic, this test won't fail even if `create_issue_command` is not throwing an error.
You should test it using `with pytest.raises...` to see that the `Exception` is thrown.

Packs/Jira/Integrations/JiraV2/JiraV2_test.py
```diff
@@ -9,12 +9,48 @@
...
+        "projects": [{'id': '1234', 'key': 'testKey', 'name': 'testName'}]
+    }
+    mocker.patch('JiraV2.jira_req', return_value=user_data)
+    mocker.patch.object(demisto, "results")
+    create_issue_command()
+    assert demisto.results.call_count == 1
+
+
+def test_create_issue_command_before_fix_mandatory_args_issue(mocker):
+    mocker.patch.object(demisto, 'args', return_value=integration_args_missing_mandatory)
```
teizenman, line 45: A better practice would be to test each of the args separately, to verify that even if one of them is missing an exception would be thrown.

teizenman (changes requested):
Good job with the UTs.

teizenman (approved):

PR #9607, opened 2020-10-22
ParseEmailFiles - payload filtering

teizenman:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29476

## Description
There was a need to filter strings out of the payload data before iterating over it.

## Screenshots
Using the two example files from the issue:
Before change:
image

After:
image


ronykoz:
@teizenman please add a UT for this

anara123:
@teizenman what is the result of the ParseEmailFiles after your fix? It returns just the attached email name?

teizenman:
> @teizenman please add a UT for this

Was waiting to see if I can get one of the files from the issue, and I can, so Im adding it as UT.

teizenman:
> @teizenman what is the result of the ParseEmailFiles after your fix? It returns just the attached email name?

The outcome is the same as before, but for some files, the `get_payload()` function is returning a list of `email.message.Message` objects, and `\n` strings that should be ignored.

anara123 (approved):

PR #9609, opened 2020-10-22
Unit tests for get_modified_files_for_testing function

jochman:
- [x] Ready
Added 100% coverage for get_modified_files_for_testing function function for future development.

image


Tests/scripts/utils/tests/test_get_modified_files_for_testing.py
```diff
@@ -0,0 +1,86 @@
...
+    Then: Validate the output contains or not the given files
+    """
+
+    def test_python_file(self, mocker):
+        diff_line = "M       Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.py"
+        yml_file = "Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.yml"
+        mocker.patch(
+            "Tests.scripts.collect_tests_and_content_packs.glob.glob",
+            return_value=[yml_file])
+        assert ([yml_file], [], [], False, [], set(), False, False) == get_modified_files_for_testing(diff_line)
```
Itay4, line 21: can we make it a bit more indicative what exactly we're asserting here?
its hard to understand what are all these `[]` and booleans

jochman: you're right - So as the function itself. I'll figure it out

Itay4 (approved):

PR #9610, opened 2020-10-22
Update docker image and fix bootstrap

GalRabin:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29637
fixes: https://github.com/demisto/etc/issues/29577

## Description
1. Update the docker image.\
2. Adding error when virtualenv isn't 16.7.7


dev-requirements-py3.txt
```diff
@@ -1,14 +1,14 @@
 flake8==3.8.3
 bandit==1.6.2
-demisto-py==2.0.11
+demisto-py>=2.0.11
```
glicht, line 3: All versions should be locked to avoid unexpected failures. Don't use '>='.
```suggestion
demisto-py==2.0.19
```

dev-requirements-py3.txt
```diff
@@ -1,14 +1,14 @@
...
 demisto-sdk==1.2.6
 pytest==6.0.2
 pytest-mock==3.3.1
 requests-mock==1.8.0
 ruamel.yaml==0.16.5
 pipenv>=2020.8.13
 autopep8==1.5.4
-mypy==0.782
+mypy==0.790
```
glicht, line 11: This change I think fails our `lint` as they added additional testing logic. I would do this in a separate PR to deal with all the fixes.

glicht (commented):
Please check the build why it is failing

.hooks/bootstrap
```diff
@@ -76,6 +76,12 @@ else
             echo "ERROR: virtualenv is missing. Please install using: pip install virtualenv" 1>&2
             exit 1
         fi
+        VIRTUALENV_VERSION=$(virtualenv --version)
+        if [[ $VIRTUALENV_VERSION != *"16.7.7"* ]]; then
```
glicht, line 80: I think any version of 16.7.* is good. 
```suggestion
        if [[ $VIRTUALENV_VERSION != *16.7.* ]]; then
```

glicht: Also this check needs to be done only if we are going to install PY2. So it should be done later on in the code and check what we concluded regarding PY2.

glicht (changes requested):

.hooks/bootstrap
GalRabin: I added verification

dev-requirements-py3.txt
GalRabin: Done

dev-requirements-py3.txt
GalRabin: Done

.hooks/bootstrap
```diff
@@ -76,6 +76,12 @@ else
             echo "ERROR: virtualenv is missing. Please install using: pip install virtualenv" 1>&2
             exit 1
         fi
+        VIRTUALENV_VERSION=$(virtualenv --version)
+        if [[ $VIRTUALENV_VERSION != *16.7.* && "${SETUP_PY2}^^" != "NO" ]]; then
```
yaakovi, line 80: why double `^`?

yaakovi (approved):
looks good to me

.hooks/bootstrap
glicht: I would just compare to "yes" as we do in the rest of the script:
```suggestion
        if [[ $VIRTUALENV_VERSION != *16.7.* && "${SETUP_PY2}" == "yes" ]]; then
```

glicht: Also need to move this whole code block after the logic of `SETUP_PY2:-auto`. line 90

glicht (changes requested):

.hooks/bootstrap
GalRabin: ^ - First letter upper case.
^^ - All letters upper case.
Typing proof.

GalRabin:
I moved it to line 90, The instances setup that occurred before related to demisto-py upgrade (Which I currently didn't apply).

.hooks/bootstrap
```diff
@@ -87,6 +87,14 @@ else
                 INCLUDE_PY2_NOTE=yes
             fi
         fi
+
+        VIRTUALENV_VERSION=$(virtualenv --version)
+        if [[ $VIRTUALENV_VERSION != *16.7.* && "$SETUP_PY2" != "yes" ]]; then
```
glicht, line 92: We should fail only if py2 == yes
```suggestion
        if [[ $VIRTUALENV_VERSION != *16.7.* && "$SETUP_PY2" == "yes" ]]; then
```

glicht (changes requested):

glicht (approved):

PR #9617, opened 2020-10-22
EWS O365 body logic

teizenman:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29622

## Description
According to the standard of `send-mail` when`htmlBody` is provided, the `body` arg is disregarded.
Tested as many integrations with `send-mail` that I could, and went over the logic of the rest and they look fine.


ronykoz (approved):
@anara123 what do you think, will it break BC?

anara123:
in this case body contains the extra stuff at the top of the mail that causes the issue?

teizenman:
@anara123 
Yes, this should be ignored as the rest of the send-mail commands.

anara123:
@teizenman ok, lets merge

PR #9647, opened 2020-10-26
Proofpoint TAP v2 - short interval bug

ChanochShayner:
…d interval*.

## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29795

## Description
Fixed an issue where fetch-incident failed on too short request interval.

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


ChanochShayner:
@jochman 
Should the integration be sent for testing to the customer to make sure the problem is resolved?

Packs/ProofpointTAP/Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
```diff
@@ -52,7 +52,7 @@ def get_fetch_times(last_fetch):
     elif isinstance(last_fetch, datetime):
         times.append(last_fetch.strftime(time_format))
     while now - last_fetch > timedelta(minutes=59):
-        last_fetch += timedelta(minutes=59)
+        last_fetch += timedelta(minutes=58)
```
ronykoz, line 55: it failed on too short and you are making it shorter?

ChanochShayner: It fails at a 30-second interval, 
and the change I made makes sure the shortest interval that can be is a minute.

ChanochShayner:
The test failed to check the number of intervals that will enter the list -
As in issue - a 10-second interval enters the list.
https://app.circleci.com/jobs/github/demisto/content/160332
image

Packs/ProofpointTAP/Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
jochman: you shouldn't change it as checks for the longest interval

jochman:
> @jochman
> Should the integration be sent for testing to the customer to make sure the problem is resolved?

You can test is with unit-tests only as the bug is well defined

Packs/ProofpointTAP/Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
jochman: or: add an informative comment on why it's 58 minutes.

jochman (changes requested):
Add a comment of how/why the interval changed to 58 minutes.

Packs/ProofpointTAP/Integrations/ProofpointTAP_v2/ProofpointTAP_v2.py
```diff
@@ -51,7 +51,7 @@ def get_fetch_times(last_fetch):
         last_fetch = datetime.strptime(last_fetch, time_format)
     elif isinstance(last_fetch, datetime):
         times.append(last_fetch.strftime(time_format))
-    while now - last_fetch > timedelta(minutes=59):
+    while now - last_fetch > timedelta(minutes=59) and now - last_fetch > timedelta(minutes=60):
```
guykeller, line 54: why add another condition? The first condition is included in the second.
Also, why does this solve the bug?

guykeller (changes requested):

ChanochShayner:
According to the issue the problem is not solved, it seems that the problem is not solved (even though the "get_fetch_times" function certainly does not return times if the differences are less than a minute).
@jochman do you have an idea what else can cause the problem?

ChanochShayner:
@jochman The customer reports that the error no longer occurs, but it happens to him from time to time (once or twice a day) the following error - 
image
Is this a known error in this integration?

jochman:
@ChanochShayner what with this pr?

PR #9652, opened 2020-10-26
Get indicators By Query ename

erezh31:

## Status
- [ ] In Progress
- [x] Ready

@Nitzperetz FYI


erezh31:
@roysagi small change - the name was too generic

Packs/Base/Scripts/GetIndicatorsByQuery/GetIndicatorsByQuery.yml
```diff
@@ -52,7 +52,7 @@ commonfields:
   id: GetIndicatorsByQuery
   version: -1
 enabled: true
-name: GetIndicatorsByQuery
+name: HashIndicatorsFields
```
roysagi, line 55: Id and name should be the same - this will fail validation.
If you will change the id, it will be a BC break and require force merge (which I can't do)... are we sure this is needed?

roysagi (changes requested):

Packs/Base/Scripts/GetIndicatorsByQuery/GetIndicatorsByQuery.yml
erezh31: I know it should be the same, but @Nitzperetz asked me to change the script name, because it's too generic.
@Itay4 what we usually do in this case?

erezh31:
we cannot really rename, just deprecate and create new automation, don't think it's worth it

PR #9657, opened 2020-10-26
Fixed pytest run on Tests and Utils

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29435

## Description
Refactored names to not be collected by Pytest, added import to globally work (slack import).
Running pytest on both Utils and Tests folder.

   - [x] No

## Must have
- [x] Tests

### DIFF
--- a/Tests/test_content.py
+++ b/Tests/test_content.py
@@ -24,6 +24,11 @@ import requests
 import demisto_client.demisto_api
 from demisto_client.demisto_api.rest import ApiException
 try:
+    """
+    Those dual-imports are required as Slack updated their sdk and it breaks BC.
+    `from slack import...` is for new slack and local runs.
+    `from slackclient import...` is for old slack and running in CircleCI (old slack in docker image)
+    """
     from slack import WebClient as SlackClient  # New slack
 except ModuleNotFoundError:
     from slackclient import SlackClient  # Old slack
--- a/Tests/test_content.py
+++ b/Tests/test_content.py
@@ -23,7 +23,10 @@ import urllib3
 import requests
 import demisto_client.demisto_api
 from demisto_client.demisto_api.rest import ApiException
-from slack import WebClient as SlackClient
+try:
+    from slack import WebClient as SlackClient  # New slack
+except ModuleNotFoundError:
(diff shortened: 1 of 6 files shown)


Tests/test_integration.py
```diff
@@ -818,6 +820,7 @@ def configure_proxy_unsecure(integration_params):
 # 3. wait for playbook to finish run
 # 4. if test pass - delete incident & instance
 # return playbook status
+@pytest.mark.skip(reason="Not a test")
 def test_integration(client, server_url, integrations, playbook_id, prints_manager, options=None, is_mock_run=False,
```
Itay4, line 824: why not rename the func like you did in the classes?

Tests/test_content.py
```diff
@@ -23,7 +23,10 @@
 import requests
 import demisto_client.demisto_api
 from demisto_client.demisto_api.rest import ApiException
-from slackclient import SlackClient
+try:
```
Itay4, line 26: why is this needed?

dev-requirements-py3.txt
```diff
@@ -20,4 +20,6 @@ circleci==1.2.2
 pylint==2.6.0
 coloredlogs==14.0
 pandas==1.1.0
-pyarrow==1.0.0
\ No newline at end of file
+pyarrow==1.0.0
+prettytable==1.0.1
```
Itay4, line 24: why are those needed? we have them in the build reqs

Tests/test_content.py
jochman: We have an old slack in our dockerimage.
On local envs mostly we will get the new slack.

dev-requirements-py3.txt
jochman: When the reqs are in the requirements, it'll assure everyone will get them

Tests/test_content.py
Itay4: please comment why this is needed

dev-requirements-py3.txt
Itay4: please remove

Itay4 (approved):

PR #9700, opened 2020-10-28
ServiceNow V2 OAuth

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23099

## Description
Adding support for OAuth2 for service now v2.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -11,6 +11,18 @@ configuration:
   name: credentials
   required: false
   type: 9
+- additionalinfo: If using OAuth, the client id of the application that was created.
```
glicht, line 14: Let's use credentials for this.  I think we can use the same credentials we use for username/password and add a checkbox "Use OAuth" and explain that when using OAuth enter into Username the cliient_id and into password the client secret.

glicht (changes requested):

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
darkushin: When using OAuth the username and password are still required for getting the first refresh token, so I think the checkbox won't be enough.

DeanArbel (commented):
Good job. See my comments.

Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule.py
```diff
@@ -0,0 +1,152 @@
...
+        }
+        try:
+            headers = {
+                'Content-Type': 'application/x-www-form-urlencoded'
+            }
+            res = self.http_request('POST', url_suffix=OAUTH_URL, data=data, headers=headers)
+            if 'error' in res:
+                return_error(
+                    f'Error occurred while creating an access token. Please check the Client ID, Client Secret '
+                    f'and that the given username and password are correct.\n{res}')
```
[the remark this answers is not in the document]
darkushin: The OAuth authorization flow in this integration is a bit different. After providing the client id and client secret in the instance configuration, the user is supposed to run once a login function and provide username and password we can create a refresh token. Therefore all 4 are needed and I think it should be `And`. Let me know if you agree.

kirbles19:
@darkushin Done.

Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule.py
```diff
@@ -0,0 +1,166 @@
...
+        else:
+            data = {'client_id': self.client_id,
+                    'client_secret': self.client_secret}
+
+            # Check if a refresh token exists. If not, raise an exception indicating to call the login function first.
+            if previous_token.get('refresh_token'):
+                data['refresh_token'] = previous_token.get('refresh_token')
+                data['grant_type'] = 'refresh_token'
+            else:
+                raise Exception('Could not create an access token. User might be not logged in. Try running the'
```
yuvalbenshalom, line 135: please add a unit test for this case.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -1908,9 +1941,61 @@ def test_module(client: Client, *_) -> Tuple[str, Dict[Any, Any], Dict[Any, Any]
...
             raise ValueError(f"The field [{client.incident_name}] does not exist in the ticket.")
 
+
+def test_module(client: Client, *_) -> Tuple[str, Dict[Any, Any], Dict[Any, Any], bool]:
+    """
+    Test the instance configurations when using basic authorization.
+    """
+    # Notify the user that test button can't be used when using OAuth 2.0:
+    if client.use_oauth:
+        return_error('Test button cannot be used when using OAuth 2.0. Please use the !servicenow-oauth-login command '
```
yuvalbenshalom, line 1951: please add a unit test for this case

yuvalbenshalom (changes requested):
looks good, just add some unit tests for the error cases that are related to the configuration of outh2/basic - I marked them in the comments.
BTW, you can do this without the instance.

Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule_test.py
```diff
@@ -0,0 +1,62 @@
+from ServiceNowApiModule import *
+import demistomock as demisto
+
```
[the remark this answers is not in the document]
darkushin: I added tests for this case in the ServiceNowv2_test.py file, as the error is raised from there so it will be a specific error for each integration and not a general one.

Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule.py
darkushin: Option (c) in the UT tests this case. I updated the documentation of the tests to clarify it.

yuvalbenshalom (approved):

yuvalbenshalom (approved):

PR #9722, opened 2020-10-29
[Enhancement] CrowdStrikeMalquery - reputation command to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description

   - [x] Yes

## Must have
- [x] Tests


yuvalbenshalom (approved):

Packs/CrowdStrikeMalquery/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### CrowdStrike Malquery
+**Breaking Change:** **file** command changed to return multiple entries (entry per indicator) instead of a single entry.
```
amshamah419, line 4:
```suggestion
**Breaking Change:** The ***file*** command was changed to return multiple entries (entry per indicator) instead of a single entry.
```

PR #9742, opened 2020-10-29
MS Graph Mail - Headers Missing Bug

ChanochShayner:
…mage.

## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29404

## Description
Fixed an issue where the mail headers is empty.
## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


ChanochShayner:
@Itay4 before you changed the timestamp filter in fetch incidents there was no '$' before the filter - 'select'
And for some reason it made headers come back from API empty -

With `$` - empty headers - 
image
image


Same mail without `$` - 
image
image

Itay4:
@ChanochShayner according to [microsoft docs](https://docs.microsoft.com/en-us/graph/query-parameters#select-parameter) it should be with `$`
what i think happens is that with `$` is actually beahves as intended and it outputs only the key fields, while the way it was before and now with the change in this PR, without `$`, it just ignores the `select` param
lets verify that with removing the `select` param and see what's returned and if it can be removed completely

ChanochShayner:
@Itay4 If you remove the 'select' filter completely, the headers return empty, really weird.
image
image

Itay4:
seems like a bug on graph api
lets add a comment in the code to clarify what we want to achieve by setting `select: *`

ChanochShayner:
> seems like a bug in graph api
> lets add a comment in the code to clarify what we want to achieve by setting `select: *`

@Itay4 Done

ronykoz (approved):

PR #9754, opened 2020-11-01
[Enhancement] OpenPhish v2 - reputation command to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description
* Update the reputation command **url** to return a different result for each indicator.

   - [x] Yes

## Must have
- [x] Tests

### DIFF
--- a/Packs/OpenPhish/ReleaseNotes/2_0_1.md
+++ b/Packs/OpenPhish/ReleaseNotes/2_0_1.md
@@ -1,4 +1,4 @@
 
 #### Integrations
 ##### New: OpenPhish v2
-- Updated the command **url** to return a different result for each indicator. 
+**Breaking Change** url command changed to return multiple entries (entry per indicator) instead of a single entry.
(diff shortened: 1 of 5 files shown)


Packs/OpenPhish/ReleaseNotes/2_0_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### New: OpenPhish v2
+**Breaking Change** url command changed to return multiple entries (entry per indicator) instead of a single entry.
```
amshamah419, line 4: Just make sure to update the docs (if there are any) to show the change.

DeanArbel: There is no way to display this change via the readme - it should mainly affect (if at all) custom scripts using `executeCommand`

yuvalbenshalom (approved):

PR #9757, opened 2020-11-01
[Enhancement] VirusTotal - reputation command to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description

   - [x] Yes

## Must have
- [x] Tests

### DIFF
--- a/Packs/VirusTotal/Integrations/integration-VirusTotal_5.5.yml
+++ b/Packs/VirusTotal/Integrations/integration-VirusTotal_5.5.yml
@@ -256,11 +256,12 @@ script:
         } else {
             r = [res.obj];
         }
-        var md = '';
-        ec["DBotScore(val.Indicator && val.Indicator === obj.Indicator && val.Vendor === obj.Vendor && val.Type === obj.Type)"] = [];
-        ec[outputPaths.file] = [];
+        var entryList = [];
         for (var i=0; i<r.length; i++) {
-            md += '## VirusTotal Hash Reputation for: ' + r[i].resource + '\n';
+            ec = {};
+            ec["DBotScore(val.Indicator && val.Indicator === obj.Indicator && val.Vendor === obj.Vendor && val.Type === obj.Type)"] = [];
+            ec[outputPaths.file] = [];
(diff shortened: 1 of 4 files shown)


Packs/VirusTotal/Integrations/integration-VirusTotal_5.5.yml
```diff
@@ -282,8 +284,15 @@ script:
...
-            ec["DBotScore(val.Indicator && val.Indicator === obj.Indicator && val.Vendor === obj.Vendor && val.Type === obj.Type)"].push({Indicator: hash, Type: 'file', Vendor: 'VirusTotal', Score: dbotScore});
+
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].md5, Type: 'hash', Vendor: 'VirusTotal', Score: dbotScore});
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].md5, Type: 'file', Vendor: 'VirusTotal', Score: dbotScore});
+
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].sha1, Type: 'hash', Vendor: 'VirusTotal', Score: dbotScore});
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].sha1, Type: 'file', Vendor: 'VirusTotal', Score: dbotScore});
+
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].sha256, Type: 'hash', Vendor: 'VirusTotal', Score: dbotScore});
+            ec[DBOTSCORE_KEY].push({Indicator: r[i].sha256, Type: 'file', Vendor: 'VirusTotal', Score: dbotScore});
```
yuvalbenshalom, line 295: aren't we pushing a list of indicators to the same entry?

yuvalbenshalom (commented):
see my comment

Packs/VirusTotal/Integrations/integration-VirusTotal_5.5.yml
DeanArbel: You're right, but it should be okay since it's only 3 indicators. It won't create a big entry, and they are aliases for the same indicator, see a screenshot from XSOAR:
image

yuvalbenshalom (approved):

PR #9763, opened 2020-11-01
[Enhancement] VirusTotal - reputation command to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description
* Update the reputation command  **ip** **domain** **hash** and **url** to return a different result for each indicator.

   - [x] Yes

## Must have
- [x] Tests

### DIFF
--- a/Packs/RecordedFuture/ReleaseNotes/1_0_2.md
+++ b/Packs/RecordedFuture/ReleaseNotes/1_0_2.md
@@ -1,0 +1,5 @@
+
+#### Integrations
+##### Recorded Future v2
+**Breaking Change** reputation commands **ip** **domain** **hash** and **url** changed to return multiple entries (entry per indicator) instead of a single entry.
+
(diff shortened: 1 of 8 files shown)


Packs/RecordedFuture/ReleaseNotes/1_0_2.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Recorded Future v2
+**Breaking Change** reputation commands **ip** **domain** **hash** and **url** changed to return multiple entries (entry per indicator) instead of a single entry.
```
amshamah419, line 4:
```suggestion
**Breaking Change** the reputation commands **ip**, **domain**, **hash**, and **url** were changed to return multiple entries (entry per indicator) instead of a single entry.
```

yuvalbenshalom (approved):

PR #9766, opened 2020-11-01
ExtractIndicatorsFromTextFile - added a file that will switch textwithniticators.txt file (testing file)

abaumgarten:
## Status
- [] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
related : https://github.com/demisto/etc/issues/30018

## Description
Added this file in order to switch the file textwithindicators.txt that used to test ExtractIndicatorsFromTextFile

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

PR #9776, opened 2020-11-02
ExtractIndicatorsFromTextFile - fixed an issue where txt file with P…

abaumgarten:
…ortuguese characters was not successfully parsed.

added Portuguese characters to the text that TPB is testing

## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30018
related: https://github.com/demisto/content/pull/9766

## Description
Fixed an issue where txt file with Portuguese characters was not successfully parsed.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
```diff
@@ -19,10 +19,10 @@
 with open(filePath, mode='r') as f:
     data = f.read(maxFileSize)
     try:
-        data = data.decode('unicode_escape').encode('utf-8')
+        data = data.decode('iso-8859-1', 'unicode_escape').encode('utf-8')
```
ronykoz, line 22: Seems that UTF-8 is a wider option, we can only add more options, not reduce them

Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
```diff
@@ -22,7 +22,7 @@
         data = data.decode('unicode_escape').encode('utf-8')
     # unicode_escape might throw UnicodeDecodeError for strings that contain \ char followed by ascii characters
     except UnicodeDecodeError:
-        data = data.encode('utf-8')
+        data = data.decode('iso-8859-1').encode('utf-8')
```
ronykoz, line 25: we should keep the `data.encode('utf-8')` as well I think, unless this code covers it

abaumgarten: When running TPB with this line 
data.encode('utf-8') , the task Extract Indicators From File - Generic v2 fails.
When return to this line:
data = data.decode('iso-8859-1').encode('utf-8')
it works good. so, I don't think that we should keep that line ..

ronykoz (approved):

PR #9808, opened 2020-11-04
[Enhancement] HelloWorld - reputation commands to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description

   - [x] Yes

## Must have
- [x] Tests

### DIFF
--- a/Packs/HelloWorld/ReleaseNotes/1_1_12.md
+++ b/Packs/HelloWorld/ReleaseNotes/1_1_12.md
@@ -1,4 +1,4 @@
 
 #### Integrations
 ##### HelloWorld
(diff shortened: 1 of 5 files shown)


anara123 (approved):

Packs/HelloWorld/ReleaseNotes/1_1_12.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### HelloWorld
+The  commands **ip** **domain** and **helloworld-scan-results** were changed to return multiple entries (entry per indicator) instead of a single entry.
```
amshamah419, line 4:
```suggestion
The **ip**, **domain**, and **helloworld-scan-results** commands were changed to return multiple entries (entry per indicator) instead of a single entry.
```

PR #9830, opened 2020-11-05
remove duplicate incident fields

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
removed duplicated files (fields are identical)
image
image

after installation, only one appears (which means there isn't any BC breakage):
image

@altmannyarden 
the following fields are still left as duplicates as they differ in the type association configuration:
* Packs/Compliance/IncidentFields/incidentfield-Media_Notification.json
* Packs/Compliance/IncidentFields/MediaNotificationStatus.json


Itay4 (approved):

altmannyarden:
I think you can remove the Packs/Compliance/IncidentFields/incidentfield-Media_Notification.json because it is not associated to any of the compliance packs (and the other incident field is associated)

yaakovi:
`incidentfield-Media_Notification.json` is associated with all incident types
`MediaNotificationStatus.json` is associated with two specific incident types

wouldn't it cause issues deleting the first one?

PR #9837, opened 2020-11-05
CrowdStrike Falcon Streaming - store 5 last sample events instead of 20

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30215

## Description
Storing now only the latest 5 events to avoid getting the cache too big

   - [x] No


kirbles19:
@Itay4 Done

PR #9866, opened 2020-11-08
[Enhancement] PassiveTotalV2 - reputation command to return a list of results

DeanArbel:
## Status
- [x] Ready

## Related Issues

## Description
* Update the commands pt-get-pdns-details, pt-whois-search and pt-get-components to return a different result for each indicator.

   - [x] Yes

## Must have
- [x] Tests


Packs/PassiveTotal/ReleaseNotes/2_0_5.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### PassiveTotal v2
+**Breaking Change** pt-get-pdns-details, pt-whois-search and pt-get-components commands changed to return multiple entries (entry per indicator) instead of a single entry.
```
amshamah419, line 4:
```suggestion
**Breaking Change:** ***pt-get-pdns-details***, ***pt-whois-search*** and ***pt-get-components*** commands were changed to return multiple entries (entry per indicator) instead of a single entry.
```

yuvalbenshalom (approved):

PR #9880, opened 2020-11-09
Fix for private packs not added to index

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: Internal

## Description
When we moved/split the private commands/functions from the public, the update_index_with_priced_packs function was ported over to the private build.

### DIFF
--- a/Tests/Marketplace/upload_packs.py
+++ b/Tests/Marketplace/upload_packs.py
@@ -931,8 +931,9 @@ def main():
     upload_core_packs_config(storage_bucket, build_number, index_folder_path)
 
     # finished iteration over content packs
-    upload_index_to_storage(index_folder_path, extract_destination_path, index_blob, build_number, private_packs,
-                            current_commit_hash, index_generation)
+    upload_index_to_storage(index_folder_path=index_folder_path, extract_destination_path=extract_destination_path,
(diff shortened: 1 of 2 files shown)


Tests/Marketplace/upload_packs.py
```diff
@@ -463,6 +463,99 @@ def get_recent_commits_data(content_repo: Any):
...
+    Returns:
+        list: priced packs from private bucket.
+
+    """
+    private_index_path = ""
+    private_packs = []
+
+    try:
+        (private_index_path, private_index_blob, _) = \
+            download_and_extract_index(private_storage_bucket,
```
guyfreund, line 487: This is not compatible with generic bucket, `download_and_extract_index` will use `GCPConfig.STORAGE_BASE_PATH` to construct the `index_storage_path`, that is true only if the bucket is the production bucket but for example for the build bucket this won't work. Can we support the generic case here?

amshamah419: @guyfreund We will open a separate PR for this change. The purpose of this PR is to revert a change so we can fix the index in the prod bucket.

yaakovi (changes requested):
don't forget to revert the config.yml changes

anara123 (approved):

PR #9900, opened 2020-11-10
CommonServerPython - remove context log

Itay4:
## Status
- [x] Ready

## Description
Removed log of context object to be stored in integration context as it might be huge and/or contain sensitive data

   - [x] No


kirbles19:
@Itay4 Done.

glicht (approved):

PR #9919, opened 2020-11-10
move helper functions to common

anara123:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Move helper functions to CommonServerPython
arg_to_int
arg_to_timestamp

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2735,6 +2750,99 @@ def __init__(self, indicators=None, category=None, message=None):
...
+        throws exception if ``True`` and argument provided is None
+
+    :return:
+        returns an ``int`` if arg can be converted
+        returns ``None`` if arg is ``None`` and required is set to ``False``
+        otherwise throws an Exception
+    :rtype: ``Optional[int]``
+    """
+
+    if not arg:
```
glicht, line 2781: Won't this do problems is arg==0 ?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2735,6 +2750,99 @@ def __init__(self, indicators=None, category=None, message=None):
...
+        if required is True:
+            raise ValueError('Missing "{}"'.format(arg_name))
+        return None
+    if isinstance(arg, str):
+        if arg.isdigit():
```
glicht, line 2786: Maybe want to use `isdecimal`?

glicht: Actually see that in py2 isdecimal is only for unicode. So ignore.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -2735,6 +2750,99 @@ def __init__(self, indicators=None, category=None, message=None):
...
+
+    This function is used to quickly validate an argument provided to XSOAR
+    via ``demisto.args()`` into an ``int`` type. It will throw a ValueError
+    if the input is invalid. If the input is None, it will throw a ValueError
+    if required is ``True``, or ``None`` if required is ``False.
+
+    :type arg: ``Any``
+    :param arg: argument to convert
+
+    :type arg_name: ``str``
```
glicht, line 2767: Why do we need `arg_name`? Just for printing the errors? It looks to me like something we can remove.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -28,6 +28,7 @@
     from requests.adapters import HTTPAdapter
     from urllib3.util import Retry
     from typing import Optional, List, Any
+    import dateparser
```
glicht, line 31: demisto/python:1.3-alpine doesn't have the dateparser package. Need to do a conditional import if we want to use this.

glicht:
we have failures in the build because of name conflicts. Maybe think of a different name or namespace these functions in to a a class with static functions.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
anara123: Yes, to print errors. The purpose of this code was to provide easy way for developers to assert and convert their args/params. 
Without returning arg name or param name, we lose the purpose.
But instead I can just print `aa is not a valid number` `2012-22-34 is not a valid date/time range` 
What do you think @glicht ?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
anara123: It is already in try/except

glicht: ok

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -28,6 +28,10 @@
     from requests.adapters import HTTPAdapter
     from urllib3.util import Retry
     from typing import Optional, List, Any
+
+    global dateparser
```
glicht, line 32: Why this?

anara123: will remove that. there was a unit test failing. Was trying something

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht: Sounds good. Let's remove the arg_name or give it an empty value by default so this function can be used as a generic function.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht: @anara123 see my comment

anara123: right, will change

glicht (approved):

PR #9924, opened 2020-11-11
[Test-Fix] Test - CrowdStrike Falcon

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30068

## Description
removed `cs-falcon-process-details` test because processes are being archived


yaakovi (approved):

yaakovi:
keep the command execution and only remove the context validation

DeanArbel:
> keep the command execution and only remove the context validation

That cannot be done, since the API returns an error 400 when looking for an "invalid" process.

PR #9925, opened 2020-11-11
pan-os Added the *profile-setting* argument to the ***panorama-create-rule***, ***panorama-edit-rule*** commands

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28444

## Minimum version of Demisto
- [x] 3.5.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -2001,6 +2001,12 @@ script:
       name: category
       required: false
       secret: false
+    - default: false
+      description: A profile setting group.
+      isArray: false
+      name: profile_setting
```
yuvalbenshalom, line 2007: is this like a free text input or a multi select in the product? if a multi select, how dose the user get the correct values to use?

bakatzir: free text, per the object he will have to create(manually from the UI)
as same as address groups objects, etc

yuvalbenshalom (approved):

kirbles19:
@bakatzir Done

PR #9943, opened 2020-11-12
Tanium Ask playbook - Added delete context for TaniumQuestion key

teizenman:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29525

## Description
The playbook queried the context at `Tanium.Question` and this has caused redundant requests to the API.

### DIFF
--- a/Packs/Tanium/Playbooks/playbook-Tanium_-_Ask_Question.yml
+++ b/Packs/Tanium/Playbooks/playbook-Tanium_-_Ask_Question.yml
@@ -1,8 +1,7 @@
 id: Tanium - Ask Question
 version: -1
-fromversion: "5.0.0"
 name: Tanium - Ask Question
-description: 'This playbook used generic polling to gets question result.'
+description: This playbook used generic polling to gets question result.
 starttaskid: "0"
 tasks:
   "0":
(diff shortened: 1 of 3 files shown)


Packs/Tanium/Playbooks/playbook-Tanium_-_Ask_Question.yml
```diff
@@ -163,14 +209,18 @@ inputs:
   value: {}
   required: false
   description: ""
+  playbookInputQuery:
 - key: parameters
   value: {}
   required: false
   description: ""
+  playbookInputQuery:
```
ronykoz, line 217: why do we have this twice?

ronykoz (approved):
so why are we deleting the context after the tanium.question?

Packs/Tanium/Playbooks/playbook-Tanium_-_Ask_Question.yml
teizenman: Its different keys.

teizenman:
> so why are we deleting the context after the tanium.question?

Its not after, this is just how the YML is arragned, the first task in no. 5 (`DeleteContext`) and then goes to no. 1 (`tn-ask-question`)

PR #9962, opened 2020-11-13
AlphaSOC NBA - Fix pagination bug

ioj:
## Status
- [ ] In Progress
- [x] Ready

## Description
This PR fixes a bug that prevented downloading a full set of new incidents when the number of incidents was very high (over a thousand). The integration didn't track the `more` field in the API response and was downloading only the first page of the results.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

fixes: https://github.com/demisto/etc/issues/30283

### DIFF
--- a/Packs/AlphaSOC_Network_Behavior_Analytics/Integrations/integration-AlphaSOC_Network_Behavior_Analytics.yml
+++ b/Packs/AlphaSOC_Network_Behavior_Analytics/Integrations/integration-AlphaSOC_Network_Behavior_Analytics.yml
@@ -155,6 +155,7 @@ script:
 
         return {
             follow: response.follow,
+            more: response.more,
             content: response.alerts,
             threats: response.threats
         }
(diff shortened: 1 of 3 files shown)


mchasepan:
@anara123 FYI this was the small change we talked about.

mchasepan:
FYI @bakatzir this is PartnerID: alphasoc-16

anara123:
@mchasepan OK.

bakatzir (changes requested):
This is an important fix.

while tending to the missing incidents and fetching them is important, it is also important to put a `limit` on it, so to not fetch thousands of incidents in a single fetch and clug Demisto: https://xsoar.pan.dev/docs/integrations/fetching-incidents#fetch-limit.
there are multiple examples in the repo: https://github.com/demisto/content/blob/master/Packs/ArcSightESM/Integrations/ArcSightESMv2/ArcSightESMv2.py#L408

after the break, you'll need to update the integration context with the appropriate time stamp accordingly, and the next fetch will continue from there.

I would recommend adding another param that will indicate from which date to start the fetch, assuming the API supports that.

Also, please fill the contribution form.

ioj:
After internal discussion, we decided to change the approach for this issue and opened https://github.com/demisto/content/pull/10019. The current version of the integration retrieves at most 1000 incidents which is above the limit you recommend in the documentation, so there's no point to handle pagination to retrieve even more.

The problem with high amount of incidents may arise when low severity incidents are downloaded by default. This is configurable, and can be mitigated by setting this option to a higher value. 

> after the break, you'll need to update the integration context with the appropriate time stamp accordingly, and the next fetch will continue from there.

Yep, it currently works like that, the integration [saves the `follow` marker returned by the API](https://github.com/demisto/content/blob/master/Packs/AlphaSOC_Network_Behavior_Analytics/Integrations/integration-AlphaSOC_Network_Behavior_Analytics.yml#L134-L144).

PR #9987, opened 2020-11-15
Added a break condition in case the installation failed on pack that …

hod-alpert:
…wasn't on the installation list

## Status
- [ ] In Progress
- [x] Ready

## Description
As seen in [this nightly build](https://app.circleci.com/pipelines/github/demisto/content/43378/workflows/2b1bf87f-bb05-474c-9c81-452aa28d8dac/jobs/187377) - There are two options to why it entered the endless loop.
1. The QRadar and CommonTypes packs are a mandatory dependencies for other packs in the list and that's why the installation failed with them even after they were removed.
2. The removal mechanism can only differ in one pack from the original packs to install that the method has received since the parameter `packs` is a set of packs made of `packs_to_install` - the malformed packs.

### DIFF
--- a/Tests/Marketplace/Tests/search_and_install_packs_test.py
+++ b/Tests/Marketplace/Tests/search_and_install_packs_test.py
@@ -247,8 +247,20 @@ def test_not_find_malformed_pack_id():
                                         'Reason: This is an error message without pack ID'):
         script.find_malformed_pack_id('This is an error message without pack ID')
 
+
 @timeout_decorator.timeout(3)
 def test_install_nightly_packs_endless_loop(mocker):
+    """
+    Given
+    - Packs to install with two packs that cannot be installed
+        (usually because their version does not exist in the bucket)
+    When
+    - Run install_nightly_packs method with those packs
+    Then
+    - Ensure the function does not enter an endless loop and that it gracefully removes the two damaged packs from the
+     installation list
+    """
+
     def generic_request_mock(self, path: str, method, body=None, accept=None, _request_timeout=None):
         requested_pack_ids = {pack['id'] for pack in body['packs']}
         for bad_integration in {'bad_integration1', 'bad_integration2'}:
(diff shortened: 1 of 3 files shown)


Itay4 (commented):
1. lets trigger a build which simulates the scenario we are handling here against these changes to verify it works as expected
2. please add unit test

hod-alpert:
@Itay4 Added unittests and verified it fails on current master (meaning it enters an endless loop)

[This](https://app.circleci.com/pipelines/github/demisto/content/43721/workflows/06d51a3b-44b3-420c-98af-4b699138bf98) is a nightly build that should pass with my changes only
And [this](https://app.circleci.com/pipelines/github/demisto/content/43719/workflows/8a6b7cfb-ac6e-4c49-8f4a-6f30e2c9e8af) is a nightly build that should show the pass after removing two different packs that are not existing in the production bucket.

Tests/Marketplace/Tests/search_and_install_packs_test.py
```diff
@@ -245,3 +246,25 @@ def test_not_find_malformed_pack_id():
     with pytest.raises(Exception, match='The request to install packs has failed. '
                                         'Reason: This is an error message without pack ID'):
         script.find_malformed_pack_id('This is an error message without pack ID')
+
+@timeout_decorator.timeout(3)
+def test_install_nightly_packs_endless_loop(mocker):
```
Itay4, line 251: please add test docstring

Tests/Marketplace/Tests/search_and_install_packs_test.py
```diff
@@ -245,3 +246,25 @@ def test_not_find_malformed_pack_id():
     with pytest.raises(Exception, match='The request to install packs has failed. '
                                         'Reason: This is an error message without pack ID'):
         script.find_malformed_pack_id('This is an error message without pack ID')
+
+@timeout_decorator.timeout(3)
```
Itay4, line 250: why is this needed?

hod-alpert: Because theoretically this can lead to an endless loop

Itay4 (approved):

PR #10005, opened 2020-11-16
remove always run when not needed

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28593

## Description
Some steps had `run: always` even tho they are dependent on the step before. Removed than

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -754,7 +754,6 @@ jobs:
             - run:
                 name: Run Tests - Server 5.0
                 shell: /bin/bash
-                when: always
                 command: |
                   if [ -z $INSTANCE_WAS_CREATED ];
                     then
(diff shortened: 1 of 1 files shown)


hod-alpert:
The always is needed here because not any failure in previous steps is a reason to fail the build and not run the tests

hod-alpert:
In server 5.0 and 5.5 there are env variable called `INSTANCE_WAS_CREATED`.

You expand them to 6.0 and server master and change it to `False` or to an empty string in case waiting for servers to be ready fails.

jochman:
The previous step is `waiting for server to be ready`. So if the previous step fails - run tests should not work anyway.

hod-alpert:
Remove the `always` from the content installation step too for 5.0 and 5.5

hod-alpert (approved):

PR #10008, opened 2020-11-16
Modify Timestamp Replacer for more Robust Detection

avidan-H:
## Status
- [x] Ready

## Description
Updates the `timestamp_replacement.py` script to use `dateparser`, whose `parse` method is more robust than that of the one currently in use from `dateutil.parser`. Need to wait to merge this PR until the AMIs are updated - see https://github.com/demisto/content-test-conf/pull/1035.

## Minimum version of Demisto
- [x] 5.0.0
- [x] 5.5.0
- [x] 6.0.0

   - [x] No

### DIFF
--- a/Tests/timestamp_replacer.py
+++ b/Tests/timestamp_replacer.py
@@ -268,11 +268,8 @@ class TimestampReplacer:
         for key, val in query_data:
             # don't bother trying to interpret an argument less than 4 characters as some type of timestamp
             if len(val) > 4:
-                try:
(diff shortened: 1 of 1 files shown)


hod-alpert (changes requested):
Let's remove all those try except clauses since they are no longer necessary

avidan-H:
I'm adding the `docs-approved` label since there is no documentation that requires a TW's review.

hod-alpert (approved):

PR #10020, opened 2020-11-17
FireEye HX - use get instead of direct key access in endpoint context

Itay4:
## Status
- [x] Ready

## Description
modified the endpoint context from dict direct key access to get to avoid key error exceptions

   - [x] No


Packs/FireEyeHX/Integrations/FireEyeHX/FireEyeHX.py
```diff
@@ -1033,13 +1033,13 @@ def get_agent_id(host_name):
...
-        'MACAddress': host['primary_mac'],
-        'OS': host['os']['platform'],
-        'OSVersion': host['os']['product_name']
+        'Hostname': host.get('hostname'),
+        'ID': host.get('_id'),
+        'IPAddress': host.get('primary_ip_address'),
+        'Domain': host.get('domain'),
+        'MACAddress': host.get('primary_mac'),
+        'OS': host.get('os', {}).get('platform'),
+        'OSVersion': host.get('os', {}).get('product_name')
```
yaakovi, line 1042: you can use `dict_safe_get` - your call

yaakovi (approved):

PR #10027, opened 2020-11-18
support for smime.p7m - ParseEmailFiles

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30674

## Description
Added support for smime.p7m email file.

## Screenshots
image

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests


Shellyber:
@Itay4 So the name of the file in this specific example is none. Do we want to change it? if so, what should we put instead? 
There is no subject in the payload

Itay4 (approved):

PR #10030, opened 2020-11-18
Cybereason - improve handling for auth failures

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30337

## Description
improved error handling for authorization failures which raised json decode errors

   - [x] No

## Must have
- [x] Tests

### DIFF
--- a/Packs/Cybereason/Integrations/Cybereason/Cybereason_test.py
+++ b/Packs/Cybereason/Integrations/Cybereason/Cybereason_test.py
@@ -260,3 +260,49 @@ def test_query_file(mocker):
     assert 'Cybereason file query results' in result[0]['HumanReadable']
     assert result[0]['EntryContext']['Cybereason.File(val.MD5 && val.MD5===obj.MD5 || val.SHA1 && '
                                      'val.SHA1===obj.SHA1)'][0]['Machine'] == 'desktop-p0m5vad'
+
+
+def test_login_failed(requests_mock, mocker):
+    """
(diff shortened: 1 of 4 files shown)


Packs/Cybereason/Integrations/Cybereason/Cybereason_test.py
```diff
@@ -260,3 +260,49 @@ def test_query_file(mocker):
...
+        'credentials': {
+            'identifier': 'username',
+            'password': 'password'
+        },
+        'proxy': True
+    })
+    mocker.patch.object(demisto, 'command', return_value='test-module')
+    return_error_mock = mocker.patch('Cybereason.return_error')
+    requests_mock.post('http://server/login.html', content=login_failed_html)
+    requests_mock.post('http://server/rest/visualsearch/query/simple', content=login_failed_html)
```
ShahafBenYakir, line 302: Why do we need this mock? after a failure in login() shouldn't an exception be raised?

Itay4: @ShahafBenYakir it is a bit misleading - the login will never actually fail, as it always return an HTML (and hence we can't really verify it), unlike the actual API requested which return an HTML only in case of failure

ShahafBenYakir (approved):

kirbles19:
@Itay4 Done

PR #10052, opened 2020-11-19
[infrastructure] Add auto handling for stale contribution branches

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30519

## Description
Add a job to delete stale branches (over 60 days since the last commit).


Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,70 @@
...
+    branch_names = []
+    for branch in repo.get_branches():
+        if branch.name.startswith('contrib/'):
+            prs_with_branch_as_base = repo.get_pulls(state='OPEN', base=branch.name)
+            if prs_with_branch_as_base.totalCount < 1:
+                # load all members
+                inspect.getmembers(branch.commit)
+
+                # check if HEAD commit is stale
+                if (last_modified := branch.commit.last_modified) and (last_commit_datetime := parse(last_modified)):
```
bakatzir, line 37: this one checks that the branch is stale indeed. but how do you check that it does not have a referenced PR?

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,70 @@
...
+        repo (Repository): The repository whose branches will be searched and listed
+
+    Returns:
+        (List[str]): List of branch names that are stale and have the "contrib/" prefix
+    """
+    now = datetime.now(timezone.min)
+    branch_names = []
+    for branch in repo.get_branches():
+        if branch.name.startswith('contrib/'):
+            prs_with_branch_as_base = repo.get_pulls(state='OPEN', base=branch.name)
```
yaakovi, line 31: make sure it is not used as the target (the internal PR)

Utils/github_workflow_scripts/close_stale_contrib_branches.py
yaakovi, line 37: using python3.8 - 🆒

.github/workflows/handle-stale-contributor-branch-daily.yml
```diff
@@ -0,0 +1,29 @@
+name: Check contributor packs
+on:
+  schedule:
+    # runs at minute 0 of the 0th and 0th hour every day.
```
yaakovi, line 4:
```suggestion
    # runs every day at midnight (UTC time).
```

.github/workflows/handle-stale-contributor-branch-daily.yml
```diff
@@ -0,0 +1,29 @@
...
+          python -m pip install --upgrade pip
+          pip install pipenv
+      - name: Sync Base Branch to Master
+        env:
+          CONTENTBOT_GH_ADMIN_TOKEN: ${{ secrets.CONTENTBOT_GH_ADMIN_TOKEN }}
+        run: |
+          echo "Handling stale contribution base branches (contrib/*)"
+          cd Utils/github_workflow_scripts
+          pipenv install --dev
+          pipenv run ./close_stale_contrib_branches.py
```
yaakovi, line 28: why do you need `cd`? any special reason?
```suggestion
          pipenv install --dev
          pipenv run Utils/github_workflow_scripts/close_stale_contrib_branches.py
```

.github/workflows/handle-stale-contributor-branch-daily.yml
```diff
@@ -0,0 +1,29 @@
+name: Check contributor packs
```
Itay4, line 1: update name

.github/workflows/handle-stale-contributor-branch-daily.yml
DeanArbel: no real reason

.github/workflows/handle-stale-contributor-branch-daily.yml
DeanArbel:
```suggestion
name: Handle stale contributor branches
```

.github/workflows/handle-stale-contributor-branch-daily.yml
avidan-H: If I'm not mistaken, `pipenv` determines which virtual environment to use based on the directory from which it is executed. So if you run it this way, I don't think it'll locate the virtual environment specified by the `Pipfile` and `Pipfile.lock` in `Utils/github_workflow_scripts`. I imagine there is an environment variable you could set to get around this, but otherwise, I don't think this will work as you expect.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,70 @@
...
+
+    stale_contrib_branches = list(set(get_stale_branch_names_with_contrib(content_repo)))
+    for branch_name in stale_contrib_branches:
+        try:
+            print(f'Deleting {branch_name}')
+            run_command(f"git push origin --delete {branch_name}", exit_on_error=False)
+        except RuntimeError as e:
+            # check delete was successful
+            if '[deleted]' not in str(e):
+                print_error(f"Deletion of {branch_name} encountered an issue: {str(e)}")
```
avidan-H, line 63: Why are we importing `demisto_sdk`? This seems to me like unnecessary overhead. I'm fairly certain that you'll be able to accomplish whatever it is you're trying to do using `pygithub`. I think you can just do the following:
```
branch_ref = repo.get_git_ref(<branch-ref>)
branch_ref.delete()
```

Utils/github_workflow_scripts/close_stale_contrib_branches.py
avidan-H: As @yaakovi said, you need to make sure there isn't an open internal PR for the branch.

avidan-H (changes requested):
You also need to update the `Pipfile` and `Pipfile.lock` - `dateparser` is not a package currently installed in the virtual environment.

avidan-H (commented):
Also, this is purely a preference, but I think the file name would be better as `.github/workflows/handle-stale-contributor-branches.yml` - it's a bit less wordy and we still get the essence of the purpose of the workflow without the `daily` suffix. Up to you though.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,77 @@
...
+    """Return the list of branches that have the prefix of "contrib/" without open pull requests
+    and that have not been updated for 2 months (stale)
+
+    Args:
+        repo (Repository): The repository whose branches will be searched and listed
+
+    Returns:
+        (List[str]): List of branch names that are stale and have the "contrib/" prefix
+    """
+    now = datetime.now(timezone.min)
```
avidan-H, line 29: What is this `timezone.min` that you are using?

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,77 @@
...
+    for branch in repo.get_branches():
+        # Make sure the branch is contrib
+        if branch.name.startswith('contrib/'):
+            prs_with_branch_as_base = repo.get_pulls(state='OPEN', base=branch.name)
+            prs_with_branch_as_head = repo.get_pulls(state='OPEN', head=f'{organization}:{branch.name})')
+
+            # Make sure there are no open prs pointing to/from the branch
+            if prs_with_branch_as_base.totalCount < 1 and prs_with_branch_as_head.totalCount < 1:
+                # load all members
+                inspect.getmembers(branch.commit)
```
avidan-H, line 41: This seems to me like something that you used during development and should be removed for production (also the import).

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,77 @@
...
+    debug_mode = len(sys.argv) >= 2 and 'debug' in sys.argv[1].casefold()
+    t = Terminal()
+    if debug_mode:
+        enable_console_debug_logging()
+    gh = Github(os.getenv('CONTENTBOT_GH_ADMIN_TOKEN'), verify=False)
+    organization = 'demisto'
+    repo = 'content'
+    content_repo = gh.get_repo(f'{organization}/{repo}')
+
+    stale_contrib_branches = list(set(get_stale_branch_names_with_contrib(content_repo)))
```
avidan-H, line 63: Why are you doing `list(set(`? Are there duplicate branches returned? Doesn't seem to me like there should be.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
```diff
@@ -0,0 +1,77 @@
...
+    for branch_name in stale_contrib_branches:
+        try:
+            print(f'Deleting {branch_name}')
+            branch_ref = content_repo.get_git_ref(os.path.join('heads', branch_name))
```
avidan-H, line 67: If you are using `os.path.join` because you are concerned with a branch name that starts with a forward slash - this does not solve that case (plus I'm pretty sure that git restricts ref names from having consecutive slashes so I don't think this would come up). I think you can just use an f-string here.

avidan-H (changes requested):

.github/workflows/handle-stale-contributor-branches.yml
```diff
@@ -0,0 +1,28 @@
...
+          python-version: '3.8'
+      - name: Install Python Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+      - name: Sync Base Branch to Master
+        env:
+          CONTENTBOT_GH_ADMIN_TOKEN: ${{ secrets.CONTENTBOT_GH_ADMIN_TOKEN }}
+        run: |
+          echo "Handling stale contribution base branches (contrib/*)"
```
avidan-H, line 25: As I said in a previous comment, you'll need to switch to the containing directory.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
DeanArbel: This is actually intentional - the members of `branch.commit` are supposed to be lazily loaded, but they are failing to load when calling the `branch.commit` directly - so this is a workaround to make it load its members.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
DeanArbel: it's 0 - it's basically to set the timezone to GMT.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
DeanArbel: There shouldn't be duplicates, but initial runs yielded duplicates. I'm not sure why that would happen, might be an issue in the branches generator. In either case, I added this code just in case.

.github/workflows/handle-stale-contributor-branches.yml
```diff
@@ -0,0 +1,29 @@
...
+          cd Utils/github_workflow_scripts
+          pipenv install --dev
+          pipenv run Utils/github_workflow_scripts/close_stale_contrib_branches.py
```
avidan-H, line 28:
```suggestion
          pipenv run ./close_stale_contrib_branches.py
```

.github/workflows/handle-stale-contributor-branches.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Install Python Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+      - name: Sync Base Branch to Master
```
avidan-H, line 21: The name of this step is incorrect.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
avidan-H: Maybe add that as a comment above it.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
avidan-H: I don't think there is a need for this. Instead of using `branch.commit.last_modified`, try using `branch.commit.commit.last_modified`. I think that should resolve that issue.

Utils/github_workflow_scripts/close_stale_contrib_branches.py
avidan-H: Do you still see that happen? Anyway, it's fine to leave it, but then I don't think there is a need to cast it back to a list. Should be ok to leave it as a set.

avidan-H (changes requested):

Utils/github_workflow_scripts/close_stale_contrib_branches.py
DeanArbel: Let me run some more tests to make sure. I'll remove the list cast - the list was used mainly for the debug print, but it shouldn't matter, as both are readable.

avidan-H (approved):

PR #10056, opened 2020-11-19
Added improvements to timestamp_replacer.py

hod-alpert:
## Related Issues
https://github.com/demisto/etc/issues/30765

## Description
* Implemented logging with format `[%(asctime)s] - [%(funcName)s] - %(message)s'`
* Implemented a method that safely parses a date object, and in case of failure - it logs the problematic value and continue.

### DIFF
--- a/Tests/timestamp_replacer.py
+++ b/Tests/timestamp_replacer.py
@@ -1,4 +1,5 @@
 import json
+from ast import literal_eval
 import functools
 import urllib
 from collections import OrderedDict
@@ -11,10 +12,14 @@ from mitmproxy.script import concurrent
 from mitmproxy.addons.serverplayback import ServerPlayback
 from time import ctime
 from dateparser import parse
+import logging
+
+logging.basicConfig(level=logging.DEBUG,
+                    format='[%(asctime)s] - [%(funcName)s] - %(message)s')
 
 
 def record_concurrently(replaying: bool = False):
-    '''
+    """
     A decorator to return a decorator that just executes the function it decorates normally if 'replaying' is true,
     (AKA mitmdump is executing in server-replay mode or reading in a mock file and cleaning it and saving the cleaned
     mock to a new file), otherwise pass the 'concurrent' decorator so that when mitmdump is executing in recording
@@ -26,15 +31,17 @@ def record_concurrently(replaying: bool = False):
 
     Returns:
         (function): decorator
-    '''
-    print(f'replaying={replaying}')
+    """
+    logging.info(f'replaying={replaying}')
     if replaying:
         def nonconcurrent_decorator(func):
             @functools.wraps(func)
             def passthrough_wrapper(*args, **kwargs):
                 value = func(*args, **kwargs)
                 return value
+
             return passthrough_wrapper
+
         return nonconcurrent_decorator
     else:
         return concurrent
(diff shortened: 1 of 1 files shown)


Tests/timestamp_replacer.py
```diff
@@ -189,39 +196,44 @@ def clean_multipart_form(self, req: HTTPRequest) -> None:
...
             else:
                 content = ''
+            logging.info(f'cleaning json body: content={content}')
             json_data = content.startswith('{')
             if json_data:
-                content = json.loads(content, object_pairs_hook=OrderedDict)
-                self.modify_json_body(req, content)
+                try:
+                    content = OrderedDict(literal_eval(content))
+                    # content = json.loads(content, object_pairs_hook=OrderedDict)
```
avidan-H, line 215:
```suggestion
```

Tests/timestamp_replacer.py
```diff
@@ -236,95 +248,100 @@ def modify_json_body(self, req: HTTPRequest, json_body: dict) -> None:
...
             json_data = content.startswith('{')
             if json_data:
-                content = json.loads(content, object_pairs_hook=OrderedDict)
-                json_keys = self.determine_problematic_keys(content)
-                self.json_keys.update(json_keys)
+                try:
+                    content = OrderedDict(literal_eval(content))
+                    json_keys = self.determine_problematic_keys(content)
+                    self.json_keys.update(json_keys)
+                except Exception:
```
avidan-H, line 331: Let's handle the particular exception that `literal_eval` raises when it can't parse the argument it receives. Otherwise, let's log the exception itself and the value of `content` instead of assuming it was an exception from `ast.literal_eval`. This comment is also applicable to the `clean_json_body` function.

Tests/timestamp_replacer.py
```diff
@@ -236,95 +248,100 @@ def modify_json_body(self, req: HTTPRequest, json_body: dict) -> None:
...
 
         Args:
             content (dict): The json request body to iterate through and find problematic timestamp data.
 
         Returns:
             List[str]: A list of keys (in dot notation, e.g. 'query.filter.time' is an example of what could be one
                 problematic key) whose values are potentially timestamp data.
-        '''
+        """
+
```
avidan-H, line 344:
```suggestion
```
Why this line? Let's remove it.

avidan-H (changes requested):

Tests/timestamp_replacer.py
hod-alpert: When you do logging.exception it will also log the exception trace back and message.
All I'm doing is adding the content it was attempted on in the beggining so we can debug.
I think we would want to see that exception traceback if it was for literal_eval or for determine_problematic_keys

avidan-H: I agree that we should output the value of content - my issue is that you are catching all exceptions, and then your exception message could be misleading because it assumes that it is an exception raised by `literal_eval`, regardless of what actually raised the exception. My initial suggestion wasn't that you shouldn't log `content` when handling the exception raised by `literal_eval`. We should always log it. I think you should either implement my initial suggestion or you should change the message you log so that it doesn't assume the exception stems from `literal_eval`.

hod-alpert: @avidan-H Changed the message.

Although just to make sure we understand it - you will see the trace, so even if the message was misleading - the trace will show you excactly what kind of exception it was and why it was raised.

avidan-H (approved):

PR #10063, opened 2020-11-20
Use aware datetime object in unittests

hod-alpert:
# using aware datetime object in unittests

The method `calc_run_at_time` return aware date.
The unittest should do the same or it will generate different date.


Itay4 (approved):

Itay4:
build fails on expected content installation
merging

PR #10070, opened 2020-11-20
fixed 6.0 'wait until server ready' step in upload packs flow

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
wait until server 6.0 is ready instead of server master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1182,7 +1182,12 @@ jobs:
       - *attach_workspace
       - *add_ssh_keys
       - *prepare_environment
-      - *wait_until_server_ready
+      - run:
+          name: Wait until server ready
+          shell: /bin/bash
+          when: always
+          command: |
+            python3 ./Tests/scripts/wait_until_server_ready.py "Demisto 6.0"
       - run:
           name: Install Packs
           shell: /bin/bash


guyfreund (approved):
I don't think we will be able to merge this before the scheduled run, so I'll trigger it after merging

.circleci/config.yml
```diff
@@ -1182,7 +1182,12 @@ jobs:
       - *attach_workspace
       - *add_ssh_keys
       - *prepare_environment
-      - *wait_until_server_ready
+      - run:
+          name: Wait until server ready
+          shell: /bin/bash
+          when: always
+          command: |
+            python3 ./Tests/scripts/wait_until_server_ready.py "Demisto 6.0"
```
guyfreund, line 1190:
```suggestion
            python3 ./Tests/scripts/wait_until_server_ready.py "$INSTNACE_ROLE"
```

guyfreund: I think we should refactor the `Install Packs in Demisto Marketplace - Master` job too

guyfreund (changes requested):

.circleci/config.yml
ShahafBenYakir: @yaakovi Is the AMI name of server master is `Demisto Marketplace`? If it is no need to sit `Install Packs in Demisto Marketplace - Master`

ShahafBenYakir (approved):

.circleci/config.yml
yaakovi: as @ShahafBenYakir said, no need to modify the second job since the step is using the correct role

PR #9799, opened 2020-11-03
MicrosoftGraphListener - Add self deployed microsoft graph listener

abaumgarten:
## Status
- [] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/28041

## Description
Fixed an issue where self deployed option was not supported.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

PR #10072, opened 2020-11-21
Microsoft Graph Mail Single User - fix detailed description

Itay4:
## Status
- [x] Ready

## Description
Fixed detailed description of Microsoft Graph Mail Single user which was wrongly modified in https://github.com/demisto/content/pull/9799/files#diff-d78f8c79150f97d3142e4170ce617372437b1a34bef40f72963fd5ac2bd980a1L1


ShahafBenYakir (commented):
Can you give a bit of background here? why are we removing this permission explanation?

yaakovi (approved):

PR #10074, opened 2020-11-21
Upload flow fixes

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
Two issues here:
1. Prepare Content didn’t copy the content/id_set.json file because it was configured to copy only files under content/packs. The upload of id_set.json was handled by the upload_packs.py script only when there was something to upload, so in cases we don’t upload, the id_set.json  file wasn’t copied.
2. We need to make sure that if we don’t upload in prepare content, we don’t upload in the last job too..

   - [ ] Yes
       - Further details:
   - [x] No

@hod-alpert FYI


Tests/Marketplace/copy_and_upload_packs.py
```diff
@@ -247,6 +247,20 @@ def verify_copy(successful_packs, pc_successful_packs_dict):
     assert not not_uploaded and not mistakenly_uploaded, error_str
 
 
+def check_if_need_to_upload(pc_successful_packs_dict, pc_failed_packs_dict):
```
ShahafBenYakir, line 250: add type hints

Tests/Marketplace/copy_and_upload_packs.py
```diff
@@ -247,6 +247,20 @@ def verify_copy(successful_packs, pc_successful_packs_dict):
...
+    """ If the two dicts are empty then no upload was done in Prepare Content step, so we need to skip uploading
+
+    Args:
+        pc_successful_packs_dict: The successful packs dict
+        pc_failed_packs_dict: The failed packs dict
+
+    """
+    if not pc_successful_packs_dict and not pc_failed_packs_dict:
```
ShahafBenYakir, line 258: Why do we want to continue the upload if there were packs that failed to upload in previous step?

ShahafBenYakir (commented):
Can we test this in some way? at least the prepare_content_packs_for_testing part?

Tests/Marketplace/copy_and_upload_packs.py
guyfreund: Because we will handle the failed packs in the packs loop, this is by design

guyfreund:
> Can we test this in some way? at least the prepare_content_packs_for_testing part?

Successful upload flow: https://app.circleci.com/pipelines/github/demisto/content/45624/workflows/19e40b11-4d37-424b-8bdf-102460486941

ShahafBenYakir (approved):

PR #10093, opened 2020-11-23
CircleCI config - set docker heap env vars

Itay4:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/30292

## Description
setting the docker heap env vars as suggested in https://discuss.circleci.com/t/using-official-docker-images/13196


glicht:
@Itay4 I don't think this is related. The setting is for Cassandra's JVM heap.

PR #10101, opened 2020-11-23
Add transform_results arg to cdl-query-logs

jtschichold:
## Status
- [ ] In Progress
- [x] Ready

## Description
Current implementation of `cdl-query-logs` command automatically maps results from CDL API into a context common to all Log types. See [here](https://github.com/demisto/content/blob/master/Packs/CortexDataLake/Integrations/CortexDataLake/README.md#context-example) for an example.
This mapping is fine as long as no special CDL query functions are used that could modify the results fields. When the results field are modified via the query, the context mapping does not collect fields falling outside common mapping.

Example:

The query: 
```
SELECT source_ip.value AS src_ip,log_source_id,vsys,COUNT(*) AS count FROM `firewall.traffic` WHERE dest_ip.value="54.240.251.124" AND source_user is NULL AND action.id=0 AND log_time > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 30 DAY) AND protocol.value=LOWER("TCP") AND dest_port=443 GROUP BY log_source_id, vsys, src_ip
```

Returns:
```
	{
		"count": 3,
		"log_source_id": "007055000130601",
		"src_ip": "10.0.5.25",
		"vsys": "vsys1"
	}
```

But when mapped to common context all the fields are `null` and custom fields are dropped:
```
	{
		"App": null,
		"TimeGenerated": null,
		"SourceIP": null,
		"DestinationLocation": null,
		"FileSHA256": null,
		"FileName": null,
		"RuleMatched": null,
		"LogSourceName": null,
		"NatDestination": null,
		"NatDestinationPort": null,
		"CharacteristicOfApp": null,
		"SourceLocation": null,
		"DestinationIP": null,
		"Action": null,
		"IsNat": null,
		"Protocol": null,
		"NatSource": null,
		"AppCategory": null
	}
```

### Changes
This PR adds a new bool argument to `cdl-query-logs` called `transform_results` to control the mapping of results to the common context:
- if set to `Yes` (default) the current behaviour is applied (map to common context)
- if set to `No`, mapping is skipped and unaltered CDL records are returned in `CDL.Logging`, a behaviour similar to `splunk-search`.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake_test.py
+++ b/Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake_test.py
@@ -133,3 +133,71 @@ def test_get_table_name():
     assert get_table_name(query) == 'firewall.threat'
     query = 'Wrongly formmated query'
     assert get_table_name(query) == 'Unrecognized table name'
+
+
+def test_query_logs_command_transform_results():
+    from CortexDataLake import query_logs_command
+    cdl_record = {
+        'action': {'value': 'allow'},
+        'app': 'web-browsing',
+        'protocol': {'value': 'tcp'},
+        'dest_ip': {'value': '198.51.100.20'},
+        'rule_matched': 'Allow-All',
+        'characteristics_of_app': [
+            "able-to-transfer-file",
+            "has-known-vulnerability",
+            "tunnel-other-application",
+            "prone-to-misuse",
+            "is-saas"
+        ],
+        'log_source_name': 'ngfw1',
+        'is_nat': False,
+        'nat_dest_port': 80,
+        'nat_dest': {'value': '198.51.100.20'},
+        'nat_source': {'value': '203.0.113.1'},
+        'source_ip': {'value': '203.0.113.1'},
+        'app_category': 'networking',
+        'source_location': '203.0.113.0-203.0.113.255',
+        'dest_location': '198.51.100.0-198.51.255',
+        'count': 100,
+    }
+
+    class MockClient():
+        def query_loggings(self, query):
+            return [cdl_record], []
+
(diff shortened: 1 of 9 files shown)


DeanArbel (changes requested):
Good work! this addition has great value.
Please see my comments below.

Packs/CortexDataLake/Integrations/CortexDataLake/CortexDataLake_test.py
```diff
@@ -133,3 +133,71 @@ def test_get_table_name():
...
+        'SourceLocation': '203.0.113.0-203.0.113.255',
+        'DestinationLocation': '198.51.100.0-198.51.255',
+        'FileSHA256': None,
+        'FileName': None,
+        'TimeGenerated': None
+    }]}
+
+    # test with transform_results options, should transform to common context
+    _, results_xform, _ = query_logs_command(
+        {'limit': '1', 'query': 'SELECT * FROM `firewall.traffic`', 'transform_results': 'No'},
```
DeanArbel, line 200: Per my suggestion to change the argument to a predefined: `true/false`.
```suggestion
        {'limit': '1', 'query': 'SELECT * FROM `firewall.traffic`', 'transform_results': 'false'},
```

jtschichold:
Thanks @DeanArbel ! Followed your suggestions, could you check?

DeanArbel (approved):
Great job!
The next step is a technical writer review in our internal repo. I'll handle this from here.

PR #10131, opened 2020-11-25
IAM - Tags and Use-cases change

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

Changing the name of the ILM use-case and adding approved tags,
currently influencing the premium build

### DIFF
--- a/Packs/Workday/pack_metadata.json
+++ b/Packs/Workday/pack_metadata.json
@@ -9,7 +9,7 @@
     "created": "2020-05-24T10:19:27Z",
     "categories": [
         "IT Services",
-        "Identity Lifecycle Management"
+        "Identity and Access Management"
     ],
     "tags": [],
     "useCases": [],
--- a/Packs/Workday/pack_metadata.json
+++ b/Packs/Workday/pack_metadata.json
@@ -9,7 +9,7 @@
     "created": "2020-05-24T10:19:27Z",
     "categories": [
         "IT Services",
-        "Identity and Access Management"
+        "Identity Lifecycle Management"
     ],
     "tags": [],
     "useCases": [],
--- a/Tests/Marketplace/approved_tags.json
+++ b/Tests/Marketplace/approved_tags.json
@@ -31,9 +31,6 @@
 		"Spam",
 		"TIM",
 		"Threat Intelligence",
-		"VPN",
-		"Provisioning",
-		"Employee",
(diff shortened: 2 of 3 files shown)


Tests/Marketplace/approved_usecases.json
```diff
@@ -15,7 +15,7 @@
 		"HIPAA Breach Notification",
 		"Health Check",
 		"Hunting",
-		"Identity And Access Management",
+		"Identity Lifecycle Management",
```
anara123, line 18: don't delete it, there are other packs that use "Identity And Access Management"

Tests/Marketplace/approved_tags.json
```diff
@@ -31,6 +31,9 @@
 		"Spam",
 		"TIM",
 		"Threat Intelligence",
-		"VPN"
+		"VPN",
+		"Provisioning",
+		"Employee",
+		"ILM"
```
anara123, line 37: no need to add these new tags. we need to remove them from the IAM pack itself

Packs/Workday/pack_metadata.json
```diff
@@ -9,7 +9,7 @@
     "created": "2020-05-24T10:19:27Z",
     "categories": [
         "IT Services",
-        "Identity and Access Management"
+        "Identity Lifecycle Management"
```
anara123, line 12:
```suggestion
        "Identity And Access Management"
```

anara123 (changes requested):

Tests/Marketplace/approved_tags.json
ohaim1008: we don't want them anymore? not even ILM?

Tests/Marketplace/approved_usecases.json
ohaim1008: ok, so I'll just change it to small "a" from some reason it's capital here but small anywhere else

Tests/Marketplace/approved_tags.json
anara123: we don't. We have IAM , it is enough

anara123 (approved):

Tests/Marketplace/approved_tags.json
anara123, line 37:
```suggestion
		"VPN"
```

PR #10135, opened 2020-11-25
Google Cloud Functions - handle no funcs found

Itay4:
## Status
- [x] Ready

## Description
handle the case in which no functions are returned to avoid key error

   - [x] No


orlichter1 (approved):
Do we want to add a small unittest too for this?

PR #10164, opened 2020-11-29
killing prints manager

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29261

## Description
This PR is about removing the ParallelPrintsManager and the different print methods from the build's steps (`print_error`, `print_warning` etc.)

In addition:
- Moved all the docker limit logs to debug file
- Moved most of the proxy prints to the debug file and added the stdout of the process.
- Since the test playbook step in private builds has one thread only - replaced the PrintsManager with logging (instead of loggingManager) 
- added logging to the slack notifier step
- Replaced as many `.format` strings with f-string

## Open design questions:
1. what logs should be visible on the console and what logs should appear in the file only:
    - Should the locking logs appear?
    - Should the proxy logs appear?

## When reviewing:
Open the `Run Tests.log` file from artifacts to see the difference between the two

Successful nightly build link - https://app.circleci.com/pipelines/github/demisto/content/46984/workflows/d1173e83-dbbd-4a93-ac86-6353c418adb8/jobs/201386
Successful instance testing build link - https://app.circleci.com/pipelines/github/demisto/content/47455/workflows/66684f44-bd75-422f-833d-f1fa395e8a38/jobs/203070
Successful private build link - https://github.com/demisto/content-private/runs/1469317268

### DIFF
--- a/Tests/configure_and_test_integration_instances.py
+++ b/Tests/configure_and_test_integration_instances.py
@@ -22,8 +22,8 @@ from Tests.scripts.utils.log_util import install_logging
 from paramiko.client import SSHClient, AutoAddPolicy
 import demisto_client
 from ruamel import yaml
-from demisto_sdk.commands.common.tools import print_error, print_warning, print_color, LOG_COLORS, run_threads_list, \
-    run_command, get_yaml, str2bool, format_version, find_type
+from demisto_sdk.commands.common.tools import run_threads_list, run_command, get_yaml,\
+    str2bool, format_version, find_type
 from demisto_sdk.commands.common.constants import FileType
 from Tests.test_integration import __get_integration_config, __test_integration_instance, \
     __disable_integrations_instances
@@ -436,7 +436,7 @@ def change_placeholders_to_values(placeholders_map, config_item):
     return json.loads(item_as_string)
 
 
-def set_integration_params(integrations, secret_params, instance_names, placeholders_map):
+def set_integration_params(integrations, secret_params, instance_names, placeholders_map, logging_module=logging):
     """
     For each integration object, fill in the parameter values needed to configure an instance from
     the secret_params taken from our secret configuration file. Because there may be a number of
@@ -459,6 +459,7 @@ def set_integration_params(integrations, secret_params, instance_names, placehol
             configuration values.
         placeholders_map: (dict)
              Dict that holds the real values to be replaced for each placeholder.
+        logging_module (Union[ParallelLoggingManager,logging]): The logging module to use
 
     Returns:
         (bool): True if integrations params were filled with secret configuration values, otherwise false
@@ -485,9 +486,9 @@ def set_integration_params(integrations, secret_params, instance_names, placehol
                                                for optional_integration in integration_params]
                     failed_match_instance_msg = 'There are {} instances of {}, please select one of them by using' \
                                                 ' the instance_name argument in conf.json. The options are:\n{}'
-                    print_error(failed_match_instance_msg.format(len(integration_params),
-                                                                 integration['name'],
-                                                                 '\n'.join(optional_instance_names)))
+                    logging_module.error(failed_match_instance_msg.format(len(integration_params),
+                                                                          integration['name'],
+                                                                          '\n'.join(optional_instance_names)))
                     return False
 
             integration['params'] = matched_integration_params.get('params', {})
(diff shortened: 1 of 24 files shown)


Itay4:
@hod-alpert @ShahafBenYakir @yaakovi could there be additional build scenarios that we should test against other than the 3 mentioned?

Itay4:
@hod-alpert about the open questions, the integration locking should not appear in the console
regarding the proxy, im pretty sure none should appear in the console - @avidan-H wdyt?

Tests/configure_and_test_integration_instances.py
```diff
@@ -810,75 +811,6 @@ def report_tests_status(preupdate_fails, postupdate_fails, preupdate_success, po
     return testing_status
 
 
-def set_marketplace_gcp_bucket_for_build(client, prints_manager, branch_name, ci_build_number, is_nightly, is_private):
```
Itay4, line 813: how come these functions can be removed?

ShahafBenYakir:
It's a bit loaded, but since you changed the slack notifier all of our flows are relevant. So missing are: SDK nightly, Demisto SDK Nightly - Run Against Cortex XSOAR, upload and force upload.

Tests/configure_and_test_integration_instances.py
hod-alpert:
```
def set_marketplace_url(servers, branch_name, ci_build_number):
    url_suffix = f'{branch_name}/{ci_build_number}'
    config_path = 'marketplace.bootstrap.bypass.url'
    config = {config_path: f'https://storage.googleapis.com/marketplace-ci-build/content/builds/{url_suffix}'}
    for server in servers:
        server.add_server_configuration(config, 'failed to configure marketplace custom url ', True)
    logging.success('Updated marketplace url and restarted servers')
    logging.info('sleeping for 60 seconds')
    sleep(60)
```
This is the one we use now it appears

Tests/mock_server.py
```diff
@@ -171,19 +171,20 @@ class MITMProxy:
...
-                 repo_folder=MOCKS_GIT_PATH, tmp_folder=MOCKS_TMP_PATH, debug=False):
+    def __init__(self,
+                 public_ip,
+                 logging_manager,
+                 repo_folder=MOCKS_GIT_PATH, tmp_folder=MOCKS_TMP_PATH):
         self.public_ip = public_ip
         self.current_folder = self.repo_folder = repo_folder
         self.tmp_folder = tmp_folder
-        self.debug = debug
+        self.logging_manager = logging_manager
```
Itay4, line 187: add the attribute to the docstring

Tests/mock_server.py
```diff
@@ -223,6 +223,7 @@ def configure_proxy_in_demisto(username, password, server, proxy=''):
         }
         response = demisto_client.generic_request_func(self=client, path='/system/config',
                                                        method='POST', body=data)
+        self.logging_manager.debug(f'Server configurations before proxy server configurations:\n{pformat(response)}')
```
Itay4, line 226:
```suggestion
        self.logging_manager.debug(f'Server configurations after proxy server configurations:\n{pformat(response)}')
```

Tests/mock_server.py
```diff
@@ -223,6 +223,7 @@ def configure_proxy_in_demisto(username, password, server, proxy=''):
         }
         response = demisto_client.generic_request_func(self=client, path='/system/config',
```
Itay4, line 224: note `response` contains more than just the server conf so the log below might be a bit misleading

hod-alpert:
@ShahafBenYakir The slack notifier will not actually send notifications unless it's running from master branch
The changes there are pretty simple, are you sure we must run all of those build steps just for that?

Tests/mock_server.py
```diff
@@ -298,75 +295,62 @@ def clean_mock_file(self, playbook_id, path=None, thread_index=0, prints_manager
             command = 'mitmdump -ns ~/timestamp_replacer.py '
             log_file = os.path.join(path, get_log_file_path(playbook_id, record=True))
             # Handle proxy log output
-            debug_opt = f' >>{log_file} 2>&1' if not self.debug else ''
+            debug_opt = f' | sudo tee -a {log_file}'
```
Itay4, line 298: @avidan-H please review the changes in the mock server as well

Tests/mock_server.py
```diff
@@ -427,36 +413,29 @@ def start(self, playbook_id, path=None, record=False, thread_index=0, prints_man
...
-            prints_manager.add_print_job(f'lsof_cmd_output={lsof_cmd_output}', print, thread_index)
-        except CalledProcessError as e:
-            cleaning_err_msg = 'No process listening on port 9997'
-            prints_manager.add_print_job(cleaning_err_msg, print_error, thread_index)
-            err_msg = f'command `{command}` exited with return code [{e.returncode}]'
-            err_msg += f'{err_msg} and the output of "{e.output}"' if e.output else err_msg
-            prints_manager.add_print_job(err_msg, print_error, thread_index)
-
-    def stop(self, thread_index=0, prints_manager=None):
+            self.logging_manager.debug(f'lsof_cmd_output={lsof_cmd_output}')
```
Itay4, line 425: i think we're in python 3.8 here, so you can just do:
```python
self.logging_manager.debug(f'{lsof_cmd_output=}')
```

Tests/scripts/slack_notifier.py
```diff
@@ -321,6 +322,7 @@ def slack_notifier(build_url, slack_token, test_type, env_results_file_name=None
 
 
 def main():
+    install_logging('Slack Notifier.log')
```
Itay4, line 325: i think its better to avoid whitespaces in file names

Tests/private_build/run_content_tests_private.py
```diff
@@ -429,8 +382,7 @@ def manage_tests(tests_settings: SettingsTester):
 
 
 def main():
-    install_simple_logging()
-    print("Time is: {}\n\n\n".format(datetime.datetime.now()))
+    install_logging('Run Tests.log')
```
Itay4, line 385: i think its better to avoid whitespaces in file names

ShahafBenYakir:
@hod-alpert I don't think that we have to run these tests

Tests/scripts/utils/log_util.py
```diff
@@ -187,8 +189,8 @@ def _add_logger(self, thread_name: str):
         log_queue = queue.Queue(-1)
         queue_handler = QueueHandler(log_queue)
         queue_handler.setLevel(logging.DEBUG)
-
         logger = logging.getLogger(thread_name)
+        logger.propagate = False
```
Itay4, line 193: why is this change?

Tests/test_content.py
```diff
@@ -43,8 +40,9 @@
 from Tests.test_integration import Docker, check_integration, disable_all_integrations
 from Tests.test_dependencies import get_used_integrations, get_tests_allocation_for_threads
 from demisto_sdk.commands.common.constants import FILTER_CONF, PB_Status
-from demisto_sdk.commands.common.tools import print_color, print_error, print_warning, \
-    LOG_COLORS, str2bool
+from demisto_sdk.commands.common.tools import str2bool
+
+logging_manager: ParallelLoggingManager = None
```
Itay4, line 45: when are we using logging module and when are we using logging manager?

Tests/scripts/utils/log_util.py
hod-alpert: Because without it the log message is passed to the root logger as well which logs it to stdout too, and we get duplicated logs each time.

Tests/test_content.py
hod-alpert: There are common methods that are being used from other steps as well.
These steps have logging module installed only.

This way those steps will still print in real time while the test content step will send the logs to the logs manager to handle

Itay4 (approved):

Tests/test_content.py
```diff
@@ -720,19 +640,17 @@ def run_test_scenario(tests_queue, tests_settings, t, proxy, default_test_timeou
...
-                                                                                                  test_to_version)
-        prints_manager.add_print_job(warning_message, print_warning, thread_index)
-        prints_manager.add_print_job(f'------ Test {test_message} end ------\n', print, thread_index,
-                                     include_timestamp=True)
+        logging_manager.info(f'------ Test {test_message} start ------')
+
+        warning_message = \
+            f'Test {test_message} ignored due to version mismatch (test versions: {test_from_version}-{test_to_version})'
+        logging_manager.warning(warning_message)
+        logging_manager.info(f'------ Test {test_message} end ------\n')
```
avidan-H, line 648: Why do we print a test start message and test end message for tests we're skipping? Can't we just print the ignore message?

hod-alpert: Sometimes you have a point

Tests/test_content.py
```diff
@@ -1510,16 +1360,9 @@ def lock_expired(lock_file: storage.Blob, lock_timeout: str) -> bool:
 
 
 def main():
-    install_simple_logging()
-    print("Time is: {}\n\n\n".format(datetime.datetime.now()))
+    global logging_manager
+    logging_manager = ParallelLoggingManager('Run Tests.log')
```
avidan-H, line 1364:
```suggestion
    logging_manager = ParallelLoggingManager('Run_Tests.log')
```

Tests/test_content.py
```diff
@@ -841,12 +759,15 @@ def get_json_file(path):
...
+                    server_ip,
+                    mockable_tests_names,
+                    unmockable_tests_names,
+                    tests_data_keeper,
+                    is_ami=True):
     server = SERVER_URL.format(server_ip)
     server_numeric_version = tests_settings.serverNumericVersion
-    start_message = "Executing tests with the server {} - and the server ip {}".format(server, server_ip)
-    prints_manager.add_print_job(start_message, print, thread_index)
+    logging_manager.info(f'Executing tests with the server {server} - and the server ip {server_ip}', real_time=True)
```
avidan-H, line 770: Remind me why it is that you set `real_time` in some places? Why specifically here do we want it? This function is sometimes executed in threads.

Tests/test_content.py
```diff
@@ -913,58 +833,48 @@ def execute_testing(tests_settings, server_ip, mockable_tests_names, unmockable_
...
-        if exc.__class__ == ApiException:
-            error_message = exc.body
-        else:
-            error_message = f'~~ Thread {thread_index + 1} failed ~~\n{str(exc)}\n{traceback.format_exc()}'
-        prints_manager.add_print_job(error_message, print_error, thread_index)
-        prints_manager.execute_thread_prints(thread_index)
-        failed_playbooks.append(f'~~ Thread {thread_index + 1} failed ~~')
+    except Exception:
+        logging_manager.exception('~~ Thread failed ~~')
+        logging_manager.execute_logs()
```
avidan-H, line 866: Do we need this `execute_logs()` here and in the `try` clause if you are invoking it in the `finally` clause on line 870?

avidan-H (commented):
See my questions and comments 👇.

Tests/test_content.py
hod-alpert: Because these are standalone logs with no context and I think it's better to show them on real time.

I see no damage in only these come up in the middle of a test playbook logs (something that will rarely happen)

PR #10168, opened 2020-11-29
Updated email body when sending update to the reporting user in Phishing - Core

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31037

## Description
Updated the email body for the email that gets sent to the user when a conclusion is drawn regarding benign and malicious emails.

## Minimum version of Demisto
5.0.0

No


idovandijk:
@michalgold PR is ready

michalgold (approved):

PR #10170, opened 2020-11-29
Add json loads in case literal_eval fails

hod-alpert:
## Status
- [x] In Progress
- [ ] Ready

## Description
Added an attempt to json.loads content in which literal_eval failed on.

## Screenshots
image

image

### DIFF
--- a/Tests/timestamp_replacer.py
+++ b/Tests/timestamp_replacer.py
@@ -213,8 +213,15 @@ class TimestampReplacer:
                 try:
                     content = OrderedDict(literal_eval(content))
                     self.modify_json_body(req, content)
+                    return
                 except Exception:
                     logging.exception(f'failed to run literal_eval on content {content}')
+                try:
+                    logging.info('parsing the request body with "literal_eval" failed - trying with "json.loads"')
+                    content = json.loads(content, object_pairs_hook=OrderedDict)
+                    self.modify_json_body(req, content)
+                except Exception:
+                    logging.exception(f'failed to run json.loads on content {content}')
 
     def modify_json_body(self, req: HTTPRequest, json_body: dict) -> None:
         """Modify the json body of a request by replacing any timestamp data with constant data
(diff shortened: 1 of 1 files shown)


Tests/timestamp_replacer.py
```diff
@@ -213,8 +213,14 @@ def clean_json_body(self, req: HTTPRequest) -> None:
...
                     content = OrderedDict(literal_eval(content))
                     self.modify_json_body(req, content)
+                    return
                 except Exception:
                     logging.exception(f'failed to run literal_eval on content {content}')
+                try:
+                    content = json.loads(content, object_pairs_hook=OrderedDict)
+                    self.modify_json_body(req, content)
+                except Exception:
+                    logging.exception(f'failed to run json.loads on content {content}')
```
avidan-H, line 223: If you nest this `try`-`except` block under the first exception handler, then there is no need for the `return` statement.

Tests/timestamp_replacer.py
```diff
@@ -213,8 +213,14 @@ def clean_json_body(self, req: HTTPRequest) -> None:
                 try:
                     content = OrderedDict(literal_eval(content))
                     self.modify_json_body(req, content)
+                    return
                 except Exception:
                     logging.exception(f'failed to run literal_eval on content {content}')
+                try:
+                    content = json.loads(content, object_pairs_hook=OrderedDict)
```
avidan-H, line 220: Should we perhaps add a log statement here - something to the effect of `'parsing the request body with "literal_ast" failed - trying with "json.loads"'`? What do you think?

Tests/timestamp_replacer.py
```diff
@@ -327,8 +333,15 @@ def handle_json_body(self, req: HTTPRequest) -> None:
                     content = OrderedDict(literal_eval(content))
                     json_keys = self.determine_problematic_keys(content)
                     self.json_keys.update(json_keys)
+                    return
```
avidan-H, line 336: Same comments as above.

avidan-H (changes requested):

Tests/timestamp_replacer.py
hod-alpert: Correct but that's weird.
I personally don't like nested code
let me know how important it is for you, if it is I'll change it

avidan-H (approved):

PR #10184, opened 2020-11-30
added createIndicators to the demistomock

jochman:
- [x] Ready


Tests/demistomock/demistomock.py
```diff
@@ -1040,3 +1040,7 @@ def mapObject(obj, mapper, mapper_type):
 
 def getModules():
     return {}
+
+
+def createIndicators(indicators):
```
Itay4, line 1045: we already have it..

PR #10186, opened 2020-11-30
Mail Listener v2 - fetch emails from the day before the last fetch

Itay4:
## Status
- [x] Ready

## Description
the IMAP client uses the `SINCE` operator which searches disregarding time and timezone see https://imapclient.readthedocs.io/en/2.1.0/api.html#imapclient.IMAPClient.search and https://tools.ietf.org/html/rfc3501 (page 51)
this means that there could be a case where emails wont be fetched (depends on the timezone)

the suggested fix is always to look one day back and by that bypass any timezone issue

   - [x] No

## Must have
- [x] Tests

### DIFF
--- a/Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
+++ b/Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
@@ -246,7 +246,7 @@ def fetch_mails(client: IMAPClient,
             continue
         # The search query filters emails by day, not by exact date
         email_message_object = Email(message_bytes, include_raw_body, save_file, mail_id)
-        if not first_fetch_time or email_message_object.date > first_fetch_time:
+        if email_message_object.date > first_fetch_time:
             mails_fetched.append(email_message_object)
     return mails_fetched, messages
 
@@ -291,7 +291,7 @@ def generate_search_query(latest_created_time: Optional[datetime],
         # Removing Parenthesis and quotes
         messages_query = messages_query.strip('()').replace('"', '')
     # Creating a list of the OR query words
-    messages_query_list = messages_query.split() + ['SINCE', latest_created_time]  # type: ignore[list-item]
+    messages_query_list = messages_query.split() + ['SINCE', latest_created_time - timedelta(days=1)]  # type: ignore[list-item]
     return messages_query_list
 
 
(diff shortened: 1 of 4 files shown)


Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
```diff
@@ -246,7 +246,7 @@ def fetch_mails(client: IMAPClient,
             continue
         # The search query filters emails by day, not by exact date
         email_message_object = Email(message_bytes, include_raw_body, save_file, mail_id)
-        if not first_fetch_time or email_message_object.date > first_fetch_time:
+        if email_message_object.date > first_fetch_time:
```
ShahafBenYakir, line 249: `first_fetch_time` can be none why removing the condition?

Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
```diff
@@ -291,7 +291,7 @@ def generate_search_query(latest_created_time: Optional[datetime],
         # Removing Parenthesis and quotes
         messages_query = messages_query.strip('()').replace('"', '')
     # Creating a list of the OR query words
-    messages_query_list = messages_query.split() + ['SINCE', latest_created_time]  # type: ignore[list-item]
+    messages_query_list = messages_query.split() + ['SINCE', latest_created_time - timedelta(days=1)]  # type: ignore[list-item]
```
ShahafBenYakir, line 294: latest_created_time can be none here too no? make is it a bit problematic to do `-` operation

ShahafBenYakir (commented):
If we will add a day to every first fetch time, we can't ever reset the first fetch to be something like an hour ago or 12 hours ago. Why not just explaining in the documentation that the first fetch is calculated without TZ so users can understand that and calculate on their own?

Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
hod-alpert: how can it be None?

Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
hod-alpert: Same here

hod-alpert:
@ShahafBenYakir The problem we're having is that the mail search resolution is that of a day, you cannot query mails from 5 minutes ago for example only the logs from today or from yesterday

hod-alpert (approved):
@Itay4 Maybe we should show the search resolution in the documentation of the first fetch parameter?
Make it more visible that the arguments should not be 5 minutes ago or something like that.

Itay4:
closing this pr, will open a new one with a more robust solution

PR #10191, opened 2020-11-30
Checkpoint Firewall v2 - pass the host id to the request properly in the delete-host cmd

Itay4:
## Status
- [x] Ready

## Description
fixed an issue where only the second char in the host id would be passed to the request func

   - [x] No

## Must have
- [x] Tests


ShahafBenYakir (approved):

Itay4:
build fails on docker image validation
merging

PR #10198, opened 2020-12-01
In case content-test-data is already exist - reset it and pull from m…

hod-alpert:
…aster

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/30511

## Description
In case content-test-data already exists on server - will reset it (in case there was uncommited changes) and pull from master

### DIFF
--- a/Tests/scripts/clone_mocks.sh
+++ b/Tests/scripts/clone_mocks.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
-ssh-keyscan github.com >> ~/.ssh/known_hosts
-
 if [[ ! -d "content-test-data" ]]; then
+    ssh-keyscan github.com >> ~/.ssh/known_hosts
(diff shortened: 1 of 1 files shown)


Tests/scripts/clone_mocks.sh
```diff
@@ -5,4 +5,6 @@ ssh-keyscan github.com >> ~/.ssh/known_hosts
 
 if [[ ! -d "content-test-data" ]]; then
     git clone git@github.com:demisto/content-test-data.git
-fi
\ No newline at end of file
+  else
+    cd content-test-data && git reset --hard && git pull -r
```
avidan-H, line 9: Why `git reset --hard`? Shouldn't the repo "be clean" when this is executed?

hod-alpert: It should, but in case we want to start from failed it won't be

avidan-H (approved):

PR #10228, opened 2020-12-02
Anomali match rebrand and DGA tag addition

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30617

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No - did not change command names and outputs, only integration diaply name and PB name

## Must have
- [x] Tests
- [x] Documentation


bakatzir:
@yuvalbenshalom Also added the `DGA` tag:
Screen Shot 2020-12-08 at 11 08 20

bakatzir:
I've added the tag only in case the domain is suspicious or malicious, per the req from Arseny.

yuvalbenshalom (approved):

PR #10229, opened 2020-12-02
Added no_output_timeout: 5h to missing steps

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Description
added missing param no_output_timeout: 5h


hod-alpert (approved):

glicht:
@ShahafBenYakir @hod-alpert is this really needed? I am concerned that if something gets stuck then the build stays hanging for 5h.

ShahafBenYakir:
@glicht 
1. From what I recall we never had such behavior of stuck build.
2. We have it in all other run tests steps (this was accidentally left out). 
3. Until now we had a manual mechanism that used to print dummy message so we wouldn't get such error: 
Server 5_0 (207642) - demisto:content 2020-12-02 11-39-24

PR #10233, opened 2020-12-02
Jira v2 - fix upload file to work properly with oauth1

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31183

## Description
fixed the upload file command to work with OAuth1 config

   - [x] No


Packs/Jira/Integrations/JiraV2/JiraV2.py
```diff
@@ -554,24 +563,17 @@ def issue_upload_command(issue_id, upload, attachment_name=None):
...
+        },
         files={'file': (attachment_name or file_name, file_bytes)},
-        auth=(USERNAME, API_TOKEN or PASSWORD),
-        verify=USE_SSL
+        resp_type='json'
     )
 
-    if not res.ok:
-        return_error(f'Failed to execute request, status code:{res.status_code}\nBody: {res.text}'
-                     + "\nMake sure file name doesn't contain any special characters" if res.status_code == 500 else "")
```
ShahafBenYakir, line 571: You don't want to add this error for 500 cases?

Itay4: good catch, added it

ShahafBenYakir (approved):

yaakovi (approved):
[PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)

PR #10261, opened 2020-12-04
ThreatGrid Detonate URL - add ThreatGrid prefix to the integration outputs

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31277

   - [x] No


ShahafBenYakir (approved):
What about detonate file?

Itay4:
> What about detonate file?

good catch, added

PR #10268, opened 2020-12-05
Symantec Blue Coat - open file by path in the upload-file command

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31303

   - [x] No


yaakovi (approved):
need to make sure the file is uploaded to the product with a valid name and not our internal (hash-based) name

PR #10275, opened 2020-12-06
Microsoft API Module - use set_to_integration_context_with_retries

Itay4:
## Status
- [x] Ready

## Description
Modify the MS API module to use `set_to_integration_context_with_retries` instead of `setIntegrationContext`
This is needed for the cases we want to write to the integration context immediately (using `sync=True`), for example when using load balancing groups when one engine writes to the integration context (on token expiry) and another one still uses the expired one

   - [x] No


yaakovi (approved):

glicht:
I think you also need to change `demisto.getIntegrationContext()` to `get_integration_context` for things to work with sync.

Itay4:
will open a new PR to do this change gradually

PR #10285, opened 2020-12-06
Application fields for ILM (IAM) pack

idovandijk:
## Status
Ready

## Related Issues
Related to: https://github.com/demisto/etc/issues/30043

## Description
Added the Application Name and Application ID fields to CommonTypes. They will be used in the ILM pack for now.

## Minimum version of Demisto
5.0.0

No


michalgold (approved):

idovandijk:
@yaron-libman since we are releasing these fields as common fields I believe we should not include the name Okta in the release notes, can I revert your last change?

PR #10289, opened 2020-12-07
Collect tests only in relevant files

abaumgarten:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31064

## Description
We should not collect tests based on changes for the metadata file, only for changes in relevant files like py/yml/etc...
Delete the part of the code that added all Pack's tests after a change in metadata file.

## Screenshots
**Before** my fix in collect tests, the build looks like this: (with a minor change in CommonScripts/AddKeyToList.py):
https://app.circleci.com/pipelines/github/demisto/content/50016/workflows/44eca69c-16b2-44cf-b109-94132b1e1d61/jobs/212942
Screen Shot 2020-12-07 at 13 00 56

**After** my fix in collect tests , the build looks like this: (with a minor change in CommonScripts/AddKeyToList.py):
https://app.circleci.com/pipelines/github/demisto/content/50017/workflows/5f5dfe24-f6ae-4857-9a37-d829ff0bc83a/jobs/212945
Screen Shot 2020-12-07 at 13 01 51

So, seems like only relevant tests were collected, and not all Pack’s tests.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Tests/scripts/collect_tests_and_content_packs.py
+++ b/Tests/scripts/collect_tests_and_content_packs.py
@@ -1071,10 +1071,6 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, minimu
     if modified_files_with_relevant_tests:
         tests, packs_to_install = find_tests_and_content_packs_for_modified_files(modified_files_with_relevant_tests,
                                                                                   conf, id_set)
-    # for pack in modified_metadata_list:
-    #     pack_tests = get_tests_for_pack(tools.pack_name_to_path(pack))
-    #     packs_to_install.add(pack)
-    #     tests = tests.union(pack_tests)
 
     # Adding a unique test for a json file.
     if is_reputations_json:
--- a/Tests/scripts/collect_tests_and_content_packs.py
+++ b/Tests/scripts/collect_tests_and_content_packs.py
@@ -1071,10 +1071,6 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, minimu
     if modified_files_with_relevant_tests:
         tests, packs_to_install = find_tests_and_content_packs_for_modified_files(modified_files_with_relevant_tests,
                                                                                   conf, id_set)
-    for pack in modified_metadata_list:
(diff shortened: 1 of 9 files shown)


ShahafBenYakir (commented):
looks good. 
Before merging, need to discuss on removing the second code part that adds the template text "Found a unitest for..." it is probably irrelevant

abaumgarten:
> looks good.
> Before merging, need to discuss on removing the second code part that adds the template text "Found a unitest for..." it is probably irrelevant

@ShahafBenYakir offered to delete this code part 
image
@guykeller , what do you think?

ShahafBenYakir:
@abaumgarten 
My suggestion corresponds to the option of running the default tests. (when deleting this piece of code it will not add the dummy text "Found a unitest for..." which will cause the default test PB to run)

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1071,10 +1071,6 @@ def get_test_list_and_content_packs_to_install(files_string, branch_name, minimu
     if modified_files_with_relevant_tests:
         tests, packs_to_install = find_tests_and_content_packs_for_modified_files(modified_files_with_relevant_tests,
                                                                                   conf, id_set)
-    for pack in modified_metadata_list:
-        pack_tests = get_tests_for_pack(tools.pack_name_to_path(pack))
-        packs_to_install.add(pack)
-        tests = tests.union(pack_tests)
```
guykeller, line 1077: Could you explain why this is the needed change?

abaumgarten: Sure. In this code, we are running over all the metadata files of the pack that was changed and add all tests of each pack to 'tests'. This is exactly what we don't want to add anymore to 'tests'. We want only to include tests of files that were changed. The tests that we do want to include are already in 'tests' 
Screen Shot 2020-12-07 at 18 34 08

abaumgarten:
opening a new PR , this one is broken

PR #10292, opened 2020-12-07
Unmocking integrations that uses files

hod-alpert:
Moving integrations to be unmocked.


Itay4 (approved):
@hod-alpert please open a task to set specific playbooks as unmockable

PR #10297, opened 2020-12-07
Mt performance pack test

yaakovi:
## Status
- [x] In Progress
- [ ] Ready


IkaDemisto:
Related to: https://github.com/demisto/etc/issues/31389

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1151,9 +1156,22 @@ def get_from_version_and_to_version_bounderies(all_modified_files_paths: set, id
...
+        pack_metadata_path = os.path.join(tools.pack_name_to_path(metadata), PACKS_PACK_META_FILE_NAME)
+        pack_metadata = get_pack_metadata(pack_metadata_path)
+        from_version = pack_metadata.get('serverMinVersion')
+        to_version = pack_metadata.get('serverMaxVersion')
+        if from_version:
+            min_from_version = min(min_from_version, LooseVersion(from_version))
+            max_from_version = max(max_from_version, LooseVersion(from_version))
+        if to_version:
+            max_to_version = max(max_to_version, LooseVersion(to_version))
+
```
hod-alpert, line 1174: Lets make this a different method so we can add unittests for it

hod-alpert: Especially the case where we have the beta word in the value

yaakovi: 1) no reason to split to another function, I will add unit tests to it in the new PR.
2) `beta` version is handled by `LooseVersion` but I will add a test for it too.

PR #10298, opened 2020-12-07
Validating requirements files are updated

hod-alpert:
Validating requirements files are updated.
Used the same validation we currently have for the config.yml to avoid a situation where the sdk version has changed with some validations but builds that are not updated with it can still pass and then potentially fail master builds.

fixes: https://github.com/demisto/etc/issues/29249

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -557,26 +557,16 @@ jobs:
           command: |
             python3 ./Tests/scripts/circleci_spell_checker.py $CIRCLE_BRANCH
       - run:
-          name: Check if CircleCI's config file is up to date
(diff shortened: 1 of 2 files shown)


Itay4:
@hod-alpert how did we test this?

hod-alpert:
@Itay4 For now only commit workflow.

Later today I will trigger:
- [ ] nightly build
- [ ] instance testing build
- [ ] sdk nightly build
- [ ] Upload packs build

Anything else?

The only check that needs to be done IMO is validating that the right server version instance is created.

Itay4 (approved):

hod-alpert:
Manually

PR #10305, opened 2020-12-07
Feed azure

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31259

## Description
Most of the indicators appear twice or more under different categories. The added fix filters the resulted list to contain the object with the most data for each indicator.


Packs/FeedAzure/Integrations/FeedAzure/README.md
```diff
@@ -0,0 +1,60 @@
+Azure.CloudIPs Feed Integration.
+This integration was integrated and tested with version xx of AzureFeed
```
roysagi, line 2: @ShahafBenYakir -  you need to add a version for Azure here

roysagi (changes requested):

Packs/FeedAzure/Integrations/FeedAzure/README.md
ShahafBenYakir: Removing this line as there is no API for this feed (we only download a file from the [MS azure page](https://www.microsoft.com/en-us/download/confirmation.aspx?id=56519))

Itay4 (approved):

PR #10306, opened 2020-12-07
CrowdStrike Streaming v2 - handle 404 status code

Itay4:
## Status
- [x] Ready

Fixes https://github.com/demisto/etc/issues/29952

## Description
on stream refresh, we might encounter a 404 in case the stream is no longer alive
in that case we should stop trying refresh and try to discover a new stream

   - [x] No


glicht (approved):

Itay4:
@glicht i've added the socket read timeout, can you please review again?

anara123 (approved):

glicht (approved):

PR #10307, opened 2020-12-07
DateStringToISOFormat - handle no tz

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31391

   - [x] No

## Must have
- [x] Tests


ShahafBenYakir (approved):
I think this could break BC (if someone performed some string operation on this output for example). Why not adding a boolean arg to decide if to add TZ or not?

Itay4:
@ShahafBenYakir added the arg
@roysagi modified the rn, please review

ShahafBenYakir (approved):

PR #10350, opened 2020-12-10
Emptied the empty tags

Shakedp:
## Status
- [x] Ready

## Related Issues
closes https://github.com/demisto/etc/issues/31473

## Description
No more empty tags to choose from


amit-davidson:
@MarkShleifer FYI
For existing customers, the fields will still be there. No one complained since version 5.5, so it's ok to leave them in my opinion. It shouldn't change much since they already have the package. I would just test it to make sure it doesn't impact anything. 
For new customers, it should install them a fresh one with those spaces so it's good for them as well.

DeanArbel (approved):
Please bump the pack version (revision), and add release notes with the changes.
You can do both that by running `demisto-sdk update-release-notes` and editing the release notes file that'll be generated.

amit-davidson (approved):
Just make sure that upgrading an existing environment with this change behaves correctly

PR #10353, opened 2020-12-10
Splunk - handle notable raw fields where key equals value

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31548

## Description
if a notable event raw field was provided in the `extractFields` integration parameter wrapped in double quotes, and the field does not exist in the event, then splunk returns it anyway with the value equals the key
added iteration over the event keys to check for that, and replace value with empty string in that case

   - [x] No

### DIFF
--- a/Packs/SplunkPy/ReleaseNotes/1_2_6.md
+++ b/Packs/SplunkPy/ReleaseNotes/1_2_6.md
@@ -1,4 +1,4 @@
 
 #### Integrations
 ##### SplunkPy
-- Fixed an issue where a notable event raw field that was provided wrapped in double quotes was returned with the key as the value in the incident metadata if it did not exist in the notable event.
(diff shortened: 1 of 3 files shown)


ShahafBenYakir (approved):
Splonk

Packs/SplunkPy/ReleaseNotes/1_2_6.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### SplunkPy
+- Fixed an issue where a notable event raw field that was provided wrapped in double quotes was returned with the key as the value in the incident metadata if it did not exist in the notable event.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where providing a raw field which is wrapped in double quotes in a notable event, would return with the key set as the value in the incident metadata if it did not exist.
```

PR #10385, opened 2020-12-14
Added slack capabilities

ShahafBenYakir:
## Status
- [x] In Progress
- [ ] Ready

## Description
Added slack notifications for contribution PRs. 

## Screenshots
Slack | Slackbot | Global | 50 new items 2020-12-17 15-37-36


Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+            client (WebClient): Slack web-client object.
+            message_blocks (list): List fo blocks representing the message blocks.
+            pr (PullRequest): object that represents the pull request.
+
+        Returns:
+            (list): List containing a dictionary which represents the message title
+    """
+    client.chat_postMessage(
+        channel="WHCL130LE",
+        # channel=SLACK_CHANNEL_TO_SEND_PR_TO,
```
ShahafBenYakir, line 176: Will revert this after we will finish the PR (don't want to spam the global channel before)

avidan-H (commented):
Since there is no issue correlated to this PR, can you please elaborate in the PR description exactly what the goals are for this PR? More details will better enable me to review this PR. For example,
1. Where do you want this slack message to appear? Is it intended as a direct message to the reviewer who has been assigned or to alert all the users in a specific channel or to alert a specific user (presumably the one who has been assigned to review the contribution) in a channel using the `@` mechanism available in slack?
2. What information do you need/want to include in the notification?

ShahafBenYakir:
@avidan-H The idea behind this feature is to help everyone involved (contribution team/BD team/Management) to be aware of new PRs open. The message will be sent to the `#contribution-reviews` channel.

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+            key (str): pack related key (example pack version)
+            value (str): pack related value (example 1.0.0)
+
+        Returns:
+            (dict): section entry for the slack block-kit
+    """
+    return ({
+        "type": "section",
+        "text": create_slack_markdown(f'```{key}: {value}\n```')
+    })
```
avidan-H, line 77: You can collapse these into one function that accepts a `type` argument of either `fields` or `text`, and the value to assign it.

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+        Returns:
+            (list): List containing a dictionary which represents the message title
+    """
+    header = [{
+        "type": "header",
+        "text": {
+            "type": "plain_text",
+            "text": f"{pr.title}",
+            "emoji": True
+        }
```
avidan-H, line 158:
```suggestion
        "text": create_slack_markdown(f"<{pr.html_url}|{pr.title}>")
```
Let's just do this and either remove line 183 or make it the same value as what we have here.

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+            message_blocks (list): List fo blocks representing the message blocks.
+            pr (PullRequest): object that represents the pull request.
+
+        Returns:
+            (list): List containing a dictionary which represents the message title
+    """
+    client.chat_postMessage(
+        channel="WHCL130LE",
+        # channel=SLACK_CHANNEL_TO_SEND_PR_TO,
+        url="https://google.coms",
```
avidan-H, line 177: What is this?

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+        attachments=[
+            {
+                "color": GREEN_COLOR,
+                "blocks": message_blocks
+            }],
```
avidan-H, line 182: Why are the blocks being passed as `attachments` and not `blocks`?

avidan-H (changes requested):
Also, don't forget to update the permissions of the script if you haven't already.

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: Attachments can contain blocks (as can be seen here).

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: 1. Using `create_slack_markdown` to create the text fo the header would not work (it needs `"type": "plain_text"` and not ` "type": "mrkdwn"` as being create in `create_slack_markdown`)
2. If you will look at the image I attached you will see that the header is different from the text created in line 183 (We have both because a header cannot contain hyperlink). Maybe it is worth removing the header but I think it adds good structure to the message.

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: It is possible, tho, in my opinion, this change will only save a few lines of code but will make this part of code a bit less clear.

ShahafBenYakir:
Thanks for the feedback man, what do you mean by update the permissions?

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,220 @@
...
+        "type": "mrkdwn",
+        "text": text
+    }
+
+
+def create_slack_fields(text_fields: List) -> dict:
+    """Create slack block-kit section entry with fields key
+
+        Args:
+            text_fields (List): String to appear in the entry
```
avidan-H, line 54:
```suggestion
            text_fields (List): list of key-value tuples
```

avidan-H (approved):
Tov, do you.

avidan-H (changes requested):
You need change the channel you are sending to before you merge though.

Utils/github_workflow_scripts/send_slack_message.py
```diff
@@ -0,0 +1,219 @@
...
+                "blocks": message_blocks
+            }],
+        text=f"<{pr.html_url}|*New Contribution:* {pr.title}>")
+
+
+def main():
+    t = Terminal()
+    payload_str = get_env_var('EVENT_PAYLOAD')
+    if not payload_str:
+        raise ValueError('EVENT_PAYLOAD env variable not set or empty')
```
avidan-H, line 190: `get_env_var` raises an `EnvVariableError` if you don't pass a default value as a second argument so there is no need to explicitly raise here.

Utils/github_workflow_scripts/send_slack_message.py
avidan-H: You never answered as to why this URL is here. And also, "coms"?

avidan-H (changes requested):

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: True. Fixed.

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: Was for testing, I thought I removed it in my CR fixes but apparently, I did not... Fixed now

Utils/github_workflow_scripts/send_slack_message.py
ShahafBenYakir: True. Fixed.

avidan-H (approved):
Merge from `master`. Your build is failing.

PR #10391, opened 2020-12-14
logger decorator - log func return value in debug-mode

Itay4:
## Status
- [x] Ready

## Description
log func return value in debug-mode when using `logger` decorator


ShahafBenYakir (approved):
Didn't we say we want to stop working with the logger decorator?

anara123:
I think we need to improve it. Not necessarily in this PR.
log decorator can print to demisto.debug in general and only if debug-mode is enabled

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1172,7 +1172,10 @@ def logger(func):
 
     def func_wrapper(*args, **kwargs):
         LOG('calling {}({})'.format(func.__name__, formatAllArgs(args, kwargs)))
-        return func(*args, **kwargs)
+        ret_val = func(*args, **kwargs)
+        if is_debug_mode():
+            demisto.debug('Return value: {}'.format(str(ret_val)))
```
yaakovi, line 1177: I would add the function name as well to avoid confusion.
```suggestion
            demisto.debug('Return value [{}]: {}'.format(func.__name__, str(ret_val)))
```

yaakovi (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht, line 1177: Shouldn't we use `LOG` here? Maybe also if in debug_mode add an immediate flush.

Itay4: @glicht doesnt the flush occur in the `DebugLogger` anyway?

anara123 (approved):

PR #10392, opened 2020-12-14
LogRhythmRest - safe get of content type header

Itay4:
## Status
- [x] Ready

## Description
safe get of content-type header

   - [x] No

## Issues:
fixes: https://github.com/demisto/etc/issues/31677


ShahafBenYakir (approved):
🦪

PR #10398, opened 2020-12-15
In case of non nightly build - logs can be executed in real time

hod-alpert:
## Description
The threads in `test_content.py` file are only used in nightly build.
That means that in regular commit - we can log in real time.


yaakovi (approved):
🆒

PR #10412, opened 2020-12-15
Demisto Lock - set integration context with sync

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31733

   - [x] No


aviadl (commented):
@Itay4 this feature does not work on javascript
You will need to re-write it in python
But we need to think about it in depth, since we might get a deadlock this way

PR #10417, opened 2020-12-16
Barracude indicator

bakatzir:
## Status
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/29016

   - [x] No - as in this case it is a single indicator anyhow.

### DIFF
--- a/Packs/Barracuda/ReleaseNotes/1_0_1.md
+++ b/Packs/Barracuda/ReleaseNotes/1_0_1.md
@@ -1,0 +1,7 @@
+
+#### Integrations
+##### Barracuda Reputation Block List (BRBL)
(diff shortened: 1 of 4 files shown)


Packs/Barracuda/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+##### Barracuda Reputation Block List (BRBL)
+- **Breaking Change** ip command changed to return multiple entries (entry per indicator) instead of a single entry.
```
DeanArbel, line 4: This is not a breaking change - the output is the same as it was before, and it doesn't return multiple entries.

DeanArbel (approved):
Nicely done, see my comments below.

PR #10420, opened 2020-12-16
spamcop ip cmd, indicaotr and outputs

bakatzir:
## Status
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/29016

   - [x] No - as actually it is returning now a single IOC anyhow.

### DIFF
--- a/Packs/Spamcop/ReleaseNotes/1_0_1.md
+++ b/Packs/Spamcop/ReleaseNotes/1_0_1.md
@@ -1,6 +1,6 @@
 
 #### Integrations
 ##### Spamcop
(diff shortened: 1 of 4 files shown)


Packs/Spamcop/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+##### Spamcop
+- **Breaking Change** ip command changed to return multiple entries (entry per indicator) instead of a single entry.
```
DeanArbel, line 4: This is not really a breaking change - there's no change to the output.

DeanArbel (approved):
Nice, see my comment below.

PR #10440, opened 2020-12-16
Sorting query data to get the same request

hod-alpert:
As can be seen from the attached photo - there are integration who can send the same request with different order in the params which will cause the proxy to miss the request
image

This PR should solve it.


avidan-H (approved):

avidan-H:
🍬 💯

PR #10444, opened 2020-12-16
Fixed Cortex XDR - Port Scan lateral movement bug

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31751

## Description
- Added a check for whether lateral movement alerts exist before saving their datetime using `SetAndHandleEmpty`, because even though it handles empty inputs, the task failed due to a transformer being used on empty values.
- Added an extra safety measure in the proceeding flow of the playbook to make sure that the lateral movement flow occurs only if the previously mentioned key EXISTS (and not just if empty)
- Added playbook readme which was missing beforehand

## Minimum version of Demisto
5.0.0

No


idovandijk:
@kirbles19 please prioritize this to the extent possible as this is for a PoC. Needs just RN . Thanks.

altmannyarden (approved):

PR #10450, opened 2020-12-16
Added Recommendation ID filtering to Compliance data.

alannix-lw:

## Status
- [x] Ready
~~In Hold - Depends on https://github.com/demisto/dockerfiles/pull/3416~~ @bakatzir - I merged it. Eli - You can review.

## Related Issues
fixes: link to the issue

## Description
This commit adds the ability to filter returned compliance data via command rather than in a playbook.

## Screenshots

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation

### DIFF
--- a/Packs/Lacework/Integrations/Lacework/Lacework.py
+++ b/Packs/Lacework/Integrations/Lacework/Lacework.py
@@ -65,7 +65,7 @@ def create_entry(title, data, ec, human_readable=None):
     }
 
 
-def format_compliance_data(compliance_data):
+def format_compliance_data(compliance_data, rec_id):
     """
     Simplify the output/contents for Compliance reports
     """
@@ -73,6 +73,13 @@ def format_compliance_data(compliance_data):
     if len(compliance_data['data']) > 0:
 
         compliance_data = compliance_data['data'][0]
+
+        # If the user wants to filter on a recommendation ID
+        if rec_id:
+            # Iterate through all recommendations, removing irrelevant ones
+            for recommendation in compliance_data["recommendations"][:]:
+                if rec_id != recommendation["REC_ID"]:
+                    compliance_data["recommendations"].remove(recommendation)
 
         # Build Human Readable Output
         readable_output = tableToMarkdown("Compliance Summary",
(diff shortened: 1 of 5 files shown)


Packs/Lacework/Integrations/Lacework/Lacework.py
```diff
@@ -74,6 +74,13 @@ def format_compliance_data(compliance_data):
 
         compliance_data = compliance_data['data'][0]
 
+        # If the user wants to filter on a recommendation ID
+        if rec_id:
+            # Iterate through all recommendations, removing irrelevant ones
+            for recommendation in compliance_data["recommendations"][:]:
+                if rec_id != recommendation["REC_ID"]:
+                    compliance_data["recommendations"].remove(recommendation)
+
```
esharf, line 83: not about it but it looks that `rec_id` can be easily converted to an array supporting parameter.
(not familiar with the use-case but looks like a quick win)

alannix-lw: Thanks for the suggestion, Eli!  The initial use case for this was to simply return one result for playbooks, but checking against a list of recommendations still allows for that use case, and makes the command more extensible.  I'll push a revision in just a few minutes.

Packs/Lacework/Integrations/Lacework/Lacework.py
```diff
@@ -121,14 +129,15 @@ def get_azure_compliance_assessment():
 
     tenant_id = demisto.args().get('tenant_id')
     subscription_id = demisto.args().get('subscription_id')
+    rec_id = demisto.args().get('rec_id')
```
esharf, line 132:
```suggestion
    rec_id = argToList(demisto.args().get('rec_id'))
```

esharf (changes requested):

Packs/Lacework/Integrations/Lacework/Lacework.py
alannix-lw: Logic updated to leverage the argToList function.

esharf (approved):

PR #10465, opened 2020-12-17
Added Workday termination report mock data

idovandijk:
## Status
Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/31369

## Description
Added additional employees to the termination mock. The employees have past termination dates / last day of work dates, and future ones. This will test the `ConvertDatetoUTC` flow in the `IAM - Check If User Should Be Terminated` playbook. The employees with past dates should be terminated, but the ones with a future date should not be terminated.

No

@ohaim1008 fyi


idovandijk:
Not needed anymore. Used directly inside the long running Workday integration.

PR #10479, opened 2020-12-17
Removing package-lock.json from being part of cache key

hod-alpert:
It seems like running `npm install .` changes the package-lock.json which makes the checksum of the file before running the `prepare Environment step be different from after.

That's making the CircleCi cache mechanism miss the cache key every time because it tries to restore different key then what it saves.

fixes: https://github.com/demisto/etc/issues/31882

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,7 +216,7 @@ references:
 
   restore_cache: &restore_cache
     restore_cache:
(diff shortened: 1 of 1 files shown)


glicht:
@hod-alpert after looking at this we should leave the the lock in the checksum. I will handle this in a separate PR. You can close this one.

PR #10482, opened 2020-12-18
npm install: use ci when in CI env

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31882

## Description
Use `npm ci` when in CI env to avoid modifying pack lock file.

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


glicht:
Run with cache restore can be seen at: https://app.circleci.com/pipelines/github/demisto/content/53390/workflows/1b1fca33-5941-4a55-93c1-452fd8ec900c/jobs/224849

hod-alpert:
@glicht 
Since running `npm ci` on current content fails with 
image

And updating the lock file solves it, perhaps it's worth updating it in this PR too?
Otherwise I think on cases of a miss - the build will fail on the prepare environment step

hod-alpert (approved):

hod-alpert:
Checked it on the build - it seems to be ok without changing the lock file too.

Don't know why it fails locally.

glicht:
@hod-alpert I think you are failing locally because of the node version you are using. You should use node version 12. I just update .nvmrc to match CI env. I suggest using `nvm` to manage the node versions being used in each package.

PR #10489, opened 2020-12-19
Debug Logger - add demisto formatter to rm sensitive url query params

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31917

## Description
set the demisto handler logger to be int logger

## Must have
- [x] Tests

### DIFF
--- a/Packs/Base/ReleaseNotes/1_5_3.md
+++ b/Packs/Base/ReleaseNotes/1_5_3.md
@@ -1,4 +1,4 @@
 
 #### Scripts
 ##### CommonServerPython
-- Improved log formatter to remove sensitive URL query parameters.
+- Improved the log formatter to remove sensitive URL query parameters.
--- a/Packs/Base/ReleaseNotes/1_5_3.md
+++ b/Packs/Base/ReleaseNotes/1_5_3.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Improved log formatter to remove sensitive URL query parameters.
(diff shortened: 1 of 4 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4837,8 +4847,7 @@ def __init__(self):
             self.http_client_print = getattr(http_client, 'print', None)  # save in case someone else patched it already
             setattr(http_client, 'print', self.int_logger.print_override)
         self.handler = DemistoHandler()
```
glicht, line 4849: There is no need for the DemistoFormatter. The fix should be that the handler should use the IntegrationLogger to write to the log and not directly to `demisto.debug()`. I think the following will fix the issue:
```suggestion
        self.handler = DemistoHandler(self.int_logger)
```

glicht (changes requested):
See my comment. Please also add a unit test.

glicht:
Also in DemistoHandler here: https://github.com/demisto/content/blob/5bf9794cebb6b0db21a02870baf8281ed171c63b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py#L4811 Need to change to use `call` and not `write`. So it should be:

```python
self.int_logger(msg)
```

yaakovi (approved):

glicht (approved):

Packs/Base/ReleaseNotes/1_5_3.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Improved log formatter to remove sensitive URL query parameters.
```
amshamah419, line 4:
```suggestion
- Improved the log formatter to remove sensitive URL query parameters.
```

amshamah419 (approved):

PR #10518, opened 2020-12-21
Reduce private artefacts size

guykeller:
## Status
- [x] In Progress
- [ ] Ready

## Description
Adding required changes for reducing the size of the content private build artefacts

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Tests/Marketplace/marketplace_services.py
```diff
@@ -737,17 +737,25 @@ def sign_pack(self, signature_string=None):
...
             os.chmod(os.path.join(extract_destination_path, 'encryptor'), stat.S_IXOTH)
-            current_working_dir = os.getcwd()
             os.chdir(extract_destination_path)
             output_file = zip_pack_path.replace("_not_encrypted.zip", ".zip")
             subprocess.call('chmod +x ./encryptor', shell=True)
             full_command = f'./encryptor ./{pack_name}_not_encrypted.zip {output_file} "' \
                            f'{encryption_key}"'
+
             subprocess.call(full_command, shell=True)
+            new_artefacts = os.path.join(current_working_dir, 'private_artifacts')
```
anara123, line 753: I would recommend avoid hard coded values, and pass the output files as part of the script args.
In this case I think it is better to pass the path to save the encrypted file and non encrypted file.
But isn't the non encrypted file existed before?

guykeller: The path is relative to the existing path, and I feel this script already has enough arguments as it is, and there would be little added value here - this folder is only used for the artifacts anyway.
How strongly do you believe this change is necessary?

guyfreund: If you decide to use a hard-coded value, I suggest maintaining it in a class/global-variable. This can help us maintain the use of this constant more easily. See `BucketUploadFlow` class in the same file for reference. What do you think?

Tests/Marketplace/marketplace_services.py
```diff
@@ -737,17 +737,25 @@ def sign_pack(self, signature_string=None):
         finally:
             return task_status
 
-    def encrypt_pack(self, zip_pack_path, pack_name, encryption_key, extract_destination_path):
+    @staticmethod
+    def encrypt_pack(zip_pack_path, pack_name, encryption_key, extract_destination_path):
```
guyfreund, line 741: Let's add docstring for this function

Tests/Marketplace/marketplace_services.py
```diff
@@ -737,17 +737,25 @@ def sign_pack(self, signature_string=None):
...
             subprocess.call(full_command, shell=True)
+            new_artefacts = os.path.join(current_working_dir, 'private_artifacts')
+            if os.path.exists(new_artefacts):
+                shutil.rmtree(new_artefacts)
+            os.mkdir(path=new_artefacts)
+            shutil.copy(zip_pack_path, os.path.join(new_artefacts, f'{pack_name}_not_encrypted.zip'))
+            shutil.copy(output_file, os.path.join(new_artefacts, f'{pack_name}.zip'))
             os.chdir(current_working_dir)
         except subprocess.CalledProcessError as error:
             print(f"Error while trying to encrypt pack. {error}")
```
guyfreund, line 761:
```suggestion
        except subprocess.CalledProcessError as error:
            print(f"Error while trying to encrypt pack. {error}")
        except shutil.Error as error:
            do_x
```
I think it is worth handling shutil errors too

guyfreund (changes requested):

anara123 (approved):

guyfreund (approved):

PR #10519, opened 2020-12-21
Add a playback run after record

hod-alpert:
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31887

## Description
After a successful record - run another playback run on this new 

### Successful playback logs:
image

### Failed playback logs:
image

## TODO
- [ ] Add a confluence page about how to solve failure cases and add it's link to the warning log

### DIFF
--- a/Tests/mock_server.py
+++ b/Tests/mock_server.py
@@ -177,13 +177,14 @@ class MITMProxy:
                  repo_folder=MOCKS_GIT_PATH,
                  tmp_folder=MOCKS_TMP_PATH,
                  ):
+        is_branch_master = branch_name == 'master'
         self.public_ip = public_ip
         self.current_folder = self.repo_folder = repo_folder
         self.tmp_folder = tmp_folder
         self.logging_module = logging_module
         self.build_number = build_number
         self.ami = AMIConnection(self.public_ip)
-        self.should_update_mock_repo = branch_name == 'master'
+        self.should_update_mock_repo = is_branch_master
         self.empty_files = []
         self.failed_tests_count = 0
         self.successful_tests_count = 0
--- a/Tests/test_content.py
+++ b/Tests/test_content.py
@@ -383,10 +383,21 @@ def mock_run(conf_json_test_details, tests_queue, tests_settings, c, demisto_use
     # Mock recording - no mock file or playback failure.
     c = demisto_client.configure(base_url=c.api_client.configuration.host,
                                  api_key=c.api_client.configuration.api_key, verify_ssl=False)
-    run_and_record(conf_json_test_details, tests_queue, tests_settings, c, demisto_user,
-                   demisto_pass, proxy, failed_playbooks, integrations, playbook_id,
-                   succeed_playbooks, test_message, test_options, slack, circle_ci,
-                   build_number, server_url, build_name)
+    succeed = run_and_record(conf_json_test_details, tests_queue, tests_settings, c, demisto_user,
+                             demisto_pass, proxy, failed_playbooks, integrations, playbook_id,
+                             succeed_playbooks, test_message, test_options, slack, circle_ci,
+                             build_number, server_url, build_name)
+    if succeed:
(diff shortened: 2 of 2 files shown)


Tests/test_content.py
```diff
@@ -383,10 +383,20 @@ def mock_run(conf_json_test_details, tests_queue, tests_settings, c, demisto_use
...
+    if succeed and proxy.should_validate_playback:
+        logging_manager.info('Running another playback attempt on the new record to verify that the record is valid')
+        with run_with_mock(proxy, playbook_id) as result_holder:
+            status, _ = check_integration(c, server_url, demisto_user, demisto_pass, integrations,
+                                          playbook_id, logging_manager, test_options,
+                                          is_mock_run=True)
+            result_holder[RESULT] = status == PB_Status.COMPLETED
+        if status != PB_Status.COMPLETED:
+            logging_manager.warning(
+                'Playback on newly created record has failed, see confluence page on that matter')  # TODO add link
```
avidan-H, line 399: I don't know man - I don't think it's enough to just log a warning and provide a link to how to resolve the mock-related issues. If a build is green, the developer will just ignore the warning. If it were up to me and a re-recorded playback fails - that's a failure if you ask me and the build should be red. Because otherwise, what that means for us (you and me) practically speaking is that _we_ will have another test playbook and/or integration that we now need to investigate! We should stop the developer from merging it to `master` to begin with since it will just create work for us down the line. The developer should be forced to resolve the mock issues related to the integration they are working on as they are working on it.

avidan-H: oh, wait. Does it already fail the build? I just saw the [screenshot](https://user-images.githubusercontent.com/41257953/102794645-5722ba00-43b4-11eb-91b2-b0d30fb8c2af.png) you attached in the PR description.

hod-alpert:
Currently I'm not failing the build on that issue.

I think it might me a good idea to first add it as a warning just to see how that affects things and then in about a week fail the build on it.

What do you think @Itay4 @anara123 ?

Itay4:
@hod-alpert @avidan-H lets print a warning as a start and make sure this doesnt cause a lot of noise and let people know to keep an eye on it
after we gain some experience with it, we will make it fail the build

Tests/mock_server.py
```diff
@@ -177,13 +177,15 @@ def __init__(self,
...
+        is_branch_master = branch_name == 'master'
         self.public_ip = public_ip
         self.current_folder = self.repo_folder = repo_folder
         self.tmp_folder = tmp_folder
         self.logging_module = logging_module
         self.build_number = build_number
         self.ami = AMIConnection(self.public_ip)
-        self.should_update_mock_repo = branch_name == 'master'
+        self.should_update_mock_repo = is_branch_master
+        self.should_validate_playback = not is_branch_master
```
avidan-H, line 188:
```suggestion
```

Tests/test_content.py
```diff
@@ -383,10 +383,21 @@ def mock_run(conf_json_test_details, tests_queue, tests_settings, c, demisto_use
...
                                  api_key=c.api_client.configuration.api_key, verify_ssl=False)
-    run_and_record(conf_json_test_details, tests_queue, tests_settings, c, demisto_user,
-                   demisto_pass, proxy, failed_playbooks, integrations, playbook_id,
-                   succeed_playbooks, test_message, test_options, slack, circle_ci,
-                   build_number, server_url, build_name)
+    succeed = run_and_record(conf_json_test_details, tests_queue, tests_settings, c, demisto_user,
+                             demisto_pass, proxy, failed_playbooks, integrations, playbook_id,
+                             succeed_playbooks, test_message, test_options, slack, circle_ci,
+                             build_number, server_url, build_name)
+    if succeed and proxy.should_validate_playback:
```
avidan-H, line 390:
```suggestion
    if succeed:
```
I think we should always validate a new recording - not only when on non-master branches.

avidan-H (approved):

PR #10525, opened 2020-12-21
Rapid Breach Response pack - Expanded IOC coverage

altmannyarden:
## Status
- [ ] In Progress
- [x] Ready

## Description
Expanded IOC coverage for SolarStorm and SUNBURST:
1. Added git sources
2. Added CVE support 
3. Added IOC STIX file
4. XDR related incident search.

## Minimum version of Demisto
- [ ] 5.0.0
- [x] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/CommonPlaybooks/Playbooks/playbook-CVE_Enrichment_-_Generic_v2.yml
+++ b/Packs/CommonPlaybooks/Playbooks/playbook-CVE_Enrichment_-_Generic_v2.yml
@@ -34,15 +34,18 @@ tasks:
     note: false
     timertriggers: []
     ignoreworker: false
-  '1':
-    id: '1'
-    taskid: 319e7be3-5939-4436-8602-ba05fc455e4c
+    skipunavailable: false
+    quietmode: 0
+  "1":
+    id: "1"
+    taskid: 4cfa2898-da16-4d61-839a-cc66e5492541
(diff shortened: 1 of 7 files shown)


idovandijk (approved):
Approved after the following fixes:

* Related hashes playbook input - change to related IOCs. Use ExtractIndicators with `Inline` to extract them instead of manually creating indicators for the hashes.
* CVE Exposure - RiskSense playbook should output generic context (Endpoint)
* Kenna - get hostname from asset ID should also return generic context (Endpoint)
* Search Endpoints by CVE - Generic should output endpoint instead of using global context (no reason to prefer global - adds noise)
* Add unique to Search Endpoints By Hash - Generic V2
* Isolate Endpoint - Generic can get endpoint name, doesn't need ID. XDR playbook uses the name to get the ID.
* "Set compromised endpoints to context" task - need to get the endpoints from the similar incidents and not from the current incidents
* Add unique transformer to panorama query log tasks
* "Set compromised endpoints to context" - set to Endpoint.Hostname or something and not Endpoint
* Change "Is Defined" checks to "Is Not Empty"
* Add "merge"="true" to createNewIndicator tasks

and after seeing it run successfully on 5.5

PR #10544, opened 2020-12-22
Microsoft Graph User - rm tenant_id assign as refresh token

Itay4:
## Status
- [x] Ready

## Description
we use the `tenant_id` in the token retrieval URL and if we pass the refresh token when trying to refresh the access token the request fails

   - [x] No


Packs/MicrosoftGraphUser/ReleaseNotes/1_3_6.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Microsoft Graph User
+- Fixed an issue where using self deployed AAD app failed to refresh access tokens.
```
bakatzir, line 4:
```suggestion
- Fixed an issue where using self deployed app failed to refresh access tokens.
```

bakatzir (approved):

PR #10546, opened 2020-12-22
Azure Log Analytics - test api.loganalytics.io in the test module

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32058

   - [x] No


ShahafBenYakir (approved):
This is not a bug. It's a dev-bug/feature but well....

PR #10589, opened 2020-12-24
Skipping infoblox

hod-alpert:
Skipping infoblox
related issue: https://github.com/demisto/etc/issues/25651
Seems to fail for instance stability issues

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -3369,7 +3369,7 @@
         "_comment2": "~~~ UNSTABLE ~~~",
         "Tenable.sc": "unstable instance",
         "ThreatConnect v2": "unstable instance",
-
+        "Infoblox": "Unstable instance, issue 25651",
         "_comment3": "~~~ QUOTA ISSUES ~~~",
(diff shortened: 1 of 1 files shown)


ShahafBenYakir:
@hod-alpert Were we able to connect? If not let's involve @Arsenikr who has contact with Krishna from Infoblox

Tests/conf.json
```diff
@@ -3369,7 +3369,7 @@
         "_comment2": "~~~ UNSTABLE ~~~",
         "Tenable.sc": "unstable instance",
         "ThreatConnect v2": "unstable instance",
-
+        "Infoblox": "Unstable instance",
```
ShahafBenYakir, line 3372: Let's create an issue to keep discussing the instance issue and mention it here.

hod-alpert:
I was able to connect but the weird thing is the failed request wasn't actually failing, it created the policy - but the connection was closed from the server side for some reason, and after couple of more requests the server refused all connections for a while.

ShahafBenYakir (approved):

PR #10600, opened 2020-12-26
Common Server Python - debug-mode curl

Itay4:
## Status
- [x] Ready

## Description
parse the http log into curl query and add it to the debug-mode log

## Must have
- [x] Tests
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1120,6 +1121,32 @@ def write(self, msg):
                 self.messages.append(text)
             else:
                 demisto.info(text)
+                if is_debug_mode() and text.startswith('send:'):
```
yaakovi, line 1124: better to have it in a helper function - will make it easier to test and document

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1104,6 +1105,44 @@ def print_log(self, verbose=False):
...
+            curl = 'curl -X {} {} {}'.format(method, url, curl_headers)
+            if demisto.params().get('proxy'):
+                proxy_address = os.environ.get('https_proxy')
+                if proxy_address:
+                    curl += '--proxy {} '.format(proxy_address)
+            else:
+                curl += '--noproxy '
+            if demisto.params().get('insecure'):
+                curl += '-k '
+            self.curl.append(curl)
```
glicht, line 1144: Why not just print it out to the log? Why do we need to store this? It may get big and also it is better to have it next to when the call happened.

Itay4: @anara123 suggested the same and i think it will bring noise as it will break the flow of the log where i can see the what was sent and what returned 
also, if there is more than one call, i'd prefer to see all the curls one after the other so it would be easier to read them

glicht: Ok. We can try this way.

glicht (approved):

anara123 (approved):

PR #10625, opened 2020-12-27
fixing bug with premium non paid packs

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Description
Due to flawed logic, the case of a non-paid premium pack (a premium pack, that should be kept private, but has a price of 0) was not properly supported.
Thus, when HelloWorldPremium was turned into such a pack, it disappeared from the index.
This issue is now fixed.


Tests/Marketplace/marketplace_services.py
```diff
@@ -1056,9 +1056,21 @@ def prepare_release_notes(self, index_folder_path, build_number):
         task_status = False
         not_updated_build = False
 
+        print('doing plain ls')
```
anara123, line 1059: is this ready for review?

Tests/Marketplace/marketplace_services.py
```diff
@@ -572,7 +572,7 @@ def _parse_pack_metadata(user_metadata, pack_content_items, pack_id, integration
         pack_metadata['certification'] = Pack._get_certification(support_type=pack_metadata['support'],
                                                                  certification=user_metadata.get('certification'))
         pack_metadata['price'] = convert_price(pack_id=pack_id, price_value_input=user_metadata.get('price'))
-        if pack_metadata['price'] > 0:
+        if 'vendorId' in user_metadata:
```
anara123, line 575: `pack_metadata['price'] >= 0` or `'price' in pack_metadata`

guykeller: Both conditions you wrote will always evaluate to `True`.
Isn't the one I wrote correct?

Tests/Marketplace/marketplace_services.py
guykeller: No, currently debugging how to use this solution to fix the index

Tests/Marketplace/marketplace_services.py
anara123: Only private packs will have price, no? so it is ok to return True for `pack_metadata['price'] >= 0`
What is the exactly problem then? But your condition is also fine

guykeller: So I prefer to stick with my condition if it's alright with you
We don't validate the fact that `price:0` isn't set in public packs at the moment, and it seems more absolutely correct anyway

yuvalbenshalom (approved):

PR #10629, opened 2020-12-27
Loggin in to docker in the server

hod-alpert:
Login -in to docker to in the server to avoid docker rate limit violation

fixes: https://github.com/demisto/etc/issues/32162

### DIFF
--- a/Tests/scripts/wait_until_server_ready.py
+++ b/Tests/scripts/wait_until_server_ready.py
@@ -1,16 +1,17 @@
 """Wait for server to be ready for tests"""
 import json
-import sys
-import time
-import re
-from subprocess import check_output
-from time import sleep
-import requests
-import urllib3.util
-from Tests.scripts.utils.log_util import install_logging
 import logging
 import os
+import re
+import sys
+import time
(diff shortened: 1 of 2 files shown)


Tests/scripts/wait_until_server_ready.py
```diff
@@ -58,6 +59,17 @@ def download_cloud_init_logs_from_server(ip: str) -> None:
...
 
+def docker_login(ip: str) -> None:
+    """
+    Login-in to docker on the server to avoid docker rate limit quota violation
+    Args:
+        ip: The ip of the server that should be logged in
+    """
+    docker_username = os.environ.get('DOCKERHUB_USER')
+    docker_password = os.environ.get('DOCKERHUB_PASSWORD')
+    check_output(f'echo {docker_password} | ssh -o StrictHostKeyChecking=no ec2-user@{ip} '
```
glicht, line 70: This doesn't work. The command will simply `echo` everything (including the `|`). You should remove the echo and just pass: `input=docker_password` as an arg.

glicht (approved):

PR #10638, opened 2020-12-28
Adding test_content.v2 infrastructure

hod-alpert:
# Test content v2:

## Description:
This PR is about the new design and implementation of the test_content_v2 code.
This Barely changes the existing code but instead the current code logic was copied

## Design:
The new design is more object-oriented and has the following classes:
#### Conf:
Represents the contents of the conf.json file from content repo. Has no public methods.
#### TestConfiguration:
Represent a test configuration in conf.json file fromcontent repo. Has no public methods.
#### SecretConf:
Represent the contents of conf.json file from content-test-conf repo. Has no public methods.
#### IntegrationConfiguration:
Represents an integration configuration from conf.json file from content-test-repo. Has no public methods.
#### BuildContext:
Holds most of the data needed for the build. Has no public methods.
#### TestResults:
All data about the tests results is collected into this object and this object is responsible for printing and writing the results to files.
has the following public methods:
- add_proxy_related_test_data
- create_result_files
- print_test_summary
#### Integration:
Represents an integration at run time, holds the integration's configuration, instance and docker image.
Has the following public methods:
- create_integration_instance
- delete_integration_instance
- disable_integration_instance
- get_docker_images
- test_integration_instance
#### TestPlaybook:
Represents a Playbook at runtime, holdes the playbook's configuration, integrations, and this instance should know if it's a mockable playbook or not.
Has the following public methods:
- configure_integrations
- create_incident
- delete_incident
- delete_integration_instances
- disable_integrations
- should_test_run
#### TestContext:
A context for a single test, manager the incident and the docker tests for the playbooks.
Has a single public method:
- execute_test
#### TestsExecution:
This class is in charge of running all needed tests on a given server.
Has a single public method:
- execute_tests

## Design Diagram:
image
image

 
## Added features:
At most this implementation does not introduces any new features, most of the low level logic is copied, however there are a few logical changes:
- The tests filter is done in the beginning and only selected tests are being executed (unlike the current flow where all tests are executed, and along the way they are being filtered out)
- Removed the clustering of the tests for the nightly threads, using the locking mechanism - In the new flow the BuildContext has queues for the mockable and umockable tests and all threads will try to fetch tests from the queue and execute them.
- Some of the results are printed in **prettytable** format and only failed and successful will be showen on the consule, the rest will be printed in debug mode.


hod-alpert:
Open issues:
- [ ] Demisto client has a leak since it saves it's last response which prevent it from being deleted even when the calling method frame is closed.
- [ ] Searching for integration config return all integrations  configured in the server, last I checked in nightly - it was 586 integrations including the images and the python code - it was 34,000,000 chars. When 10 threads do it together it can cause a rate limit violation

PR #10223, opened 2020-12-02
Mail Listener v2 - add UID to IMAP search criteria

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31157

## Description
- added UID to IMAP search criteria, in order to avoid cases where the last fetch timestamp was not bumped in some scenarios and by that led to emails not being fetched
- after the first fetch, fetch will be by uid
- modified the first fetch to 1 hour
- fixed an issue where messages that were not ingested as incidents would get deleted from the mail box (if the param was enabled)

## Does it break backward compatibility?
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

PR #10650, opened 2020-12-29
Mail Listener v2 - revert removal of last_fetch from last run

Itay4:
## Status
- [x] Ready

## Description
revert removal of the `last_fetch` attribute from the last run object which stored the last fetch timestamp done in https://github.com/demisto/content/pull/10223
using it only for the first fetch after upgrading (when the last uid is `1`) and after that will use only the uid as criterion to fetch from

   - [x] No

### DIFF
--- a/Packs/MailListener/ReleaseNotes/1_0_2.md
+++ b/Packs/MailListener/ReleaseNotes/1_0_2.md
@@ -1,0 +1,4 @@
+
+#### Integrations
+##### Mail Listener v2
+- Reverted removal of the last fetch timestamp stored in the fetch incidents flow.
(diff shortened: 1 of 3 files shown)


Packs/MailListener/ReleaseNotes/1_0_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Mail Listener v2
+- Reverted removal of the last fetch timestamp stored in the fetch incidents flow.
```
amshamah419, line 4:
```suggestion
- Reverted a change made in `1.0.1` which would remove the last fetch timestamp stored in the fetch incidents flow.
```

glicht (approved):

PR #10668, opened 2020-12-29
[tests] Removed VxStream from skipped

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23795

## Description
Removed **VxStream** from skipped and from **Detonate File - Generic Test** requirements.


ShahafBenYakir (approved):
Oh now I see you did exactly what I suggested which is to remove it from some of our tests. Thanks!

PR #10674, opened 2020-12-30
Misp v2 argument fix

moishce:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/32208


ShahafBenYakir (approved):

moishce:
Need force merge because the validation requires updating the docker image
And the integration does not work with the updated docker image

PR #10694, opened 2020-12-31
O365 Security And Compliance - send device_code instead of code

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32274

## Description
according to https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code#authenticating-the-user, `device_code` should be sent in the request body

   - [x] No


GalRabin (changes requested):
I just validated that the flow working as expected with our account:
Screen Shot 2021-01-03 at 0 40 17

In addition I validated it with post collection which published by microsoft this is the correct flow.

As you allready suggested we should schedule a call with the customer to investigate this unexpected behaviour.

PR #10699, opened 2020-12-31
Maltiverse rm arguments

moishce:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/32208

### DIFF
--- a/Packs/Maltiverse/ReleaseNotes/1_0_1.md
+++ b/Packs/Maltiverse/ReleaseNotes/1_0_1.md
@@ -1,0 +1,5 @@
+
+#### Integrations
+##### Maltiverse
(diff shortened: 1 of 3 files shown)


Packs/Maltiverse/ReleaseNotes/1_0_1.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Maltiverse
+- ***breaking changes***: removed the *fullResponse* argument from all commands
```
ShahafBenYakir, line 4:
```suggestion
- Removed the *fullResponse* argument from all commands
```
I think this is not a breaking change

ShahafBenYakir:
`fullResponse` argument was not used in the code at all and thus is deleted. This will not break BC. Force merging

PR #10706, opened 2021-01-01
Common Server Python - add xml support and ignore all hosts for proxy to curl query

Itay4:
## Status
- [x] Ready

## Description
 - added support for XML body
 - fixed an issue where the `--noproxy` option took the `-k` option as input. set the input to `*` to ignore proxy on all hosts

   - [x] No

## Must have
- [x] Tests


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1131,8 +1131,9 @@ def build_curl(self, text):
         """
         http_methods = ['GET', 'POST', 'PUT', 'DELETE', 'PATCH']
         data = text.split("send: b'")[1]
-        if data.startswith('{'):
+        if data[0] in {'{', '<'}:
```
yaakovi, line 1134: can data be empty? if so, then:
```suggestion
        if data and data[0] in {'{', '<'}:
```

yaakovi (approved):

glicht (approved):

PR #10733, opened 2021-01-04
Running with podman configurations

hod-alpert:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29360

## Description
Added support to running the build with RHEL instances including:
- Separate server configuration
- login and logout to podman in rootless mode requires the login and logout commands to be executed in the user home directory
- Handling the `Docker Hardening test playbook` and validating that a podman instance will not be the one to run it.

## TBD:
- [x] Handling the `Docker Hardening test playbook` and validating that a podman instance will not be the one to run it.

A successful nightly build with all RHEL instances - https://app.circleci.com/pipelines/github/demisto/content/57756/workflows/55cf235b-9e33-401c-90d3-dd7e2122565e/jobs/241723


.circleci/config.yml
```diff
@@ -522,7 +522,7 @@ jobs:
     <<: *environment
     steps:
       - checkout
-      - *restore_cache
+      # - *restore_cache
```
glicht, line 525: you shouldn't need to do this when you use the git has in requirements as you did.

glicht (approved):
Don't forget to change restore_cache stuff and demisto_sdk version

PR #10759, opened 2021-01-06
Install Pack Fix

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32397

## Description
Fixing an issue where failure in install packs didn't cause the build to fail.
Additionally, changed the flow of upload-packs to support the 'serverMinVersion' field, i.e. if a pack requires a minimum server version greater than the server that is currently tested, don't try to install this pack on this server.

   - [ ] Yes
       - Further details:
   - [x] No


Tests/Marketplace/search_and_install_packs.py
```diff
@@ -332,11 +332,14 @@ def install_packs(client: demisto_client,
             result_object = ast.literal_eval(response_data)
             message = result_object.get('message', '')
             raise Exception(f'Failed to install packs - with status code {status_code}\n{message}')
-    except Exception:
-        logging.exception('The request to install packs has failed.')
+    except Exception as e:
+        logging.exception(f'The request to install packs has failed. Additional info: {str(e)}')
```
guyfreund, line 336:
```suggestion
        logging.exception(f'The request to install packs has failed. Additional info: {e.args[0]}')
```
Do we want to change it?

guyfreund (commented):
Looks good, please attach a successful upload-flow run

Tests/Marketplace/search_and_install_packs.py
darkushin: I think we will leave it that way as I saw some odd behavior when using `e.args[0]`.

darkushin:
> Looks good, please attach a successful upload-flow run before merging

Successful upload-flow run: https://app.circleci.com/pipelines/github/demisto/content/58812/workflows/923eec47-04b0-4020-8d18-b32dede922fc

Also, attaching screenshots from server 6_0 and server 6_1 that were tested, showing that in server 6_1 one additional pack that has `serverMinVersion: 6.1.0` was installed compared to server 6_0. In the logs of the attached build, we can assert that the one extra pack is `Multi-Tenant Performance` which indeed has `serverMinVersion: 6.1.0`.

Server6_0
Server6_1

yuvalbenshalom (approved):

guyfreund (approved):

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -441,12 +444,13 @@ def install_all_content_packs_for_nightly(client: demisto_client, host: str, ser
     install_packs(client, host, all_packs, is_nightly=True)
 
 
-def install_all_content_packs(client: demisto_client, host: str):
+def install_all_content_packs(client: demisto_client, host: str, server_version=None):
```
guyfreund, line 447:
```suggestion
def install_all_content_packs(client: demisto_client, host: str, server_version):
```
As the Upload-Flow is the only one using this function, let's keep this arg as mandatory

If you think its not mandatory, then a verification that isn't empty should be done before line 467

guyfreund (changes requested):

guyfreund (approved):

PR #10767, opened 2021-01-06
fixing legacy key handling

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32381

## Description
To this point, all packs in the marketplace were marked as `legacy` in the index.
This will no longer be the case for future versions.


guykeller:
Closing because a more surgical solution will be needed here.

PR #10788, opened 2021-01-07
Pan-os bug fix

reutshal:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32466

## Description
Added the correct xpath to the command 
Fixed another bug caused by incorrect table handling

   - [ ] Yes
       - Further details:
   - [X] No


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5544,152 +5544,158 @@ def get_security_profiles_command(security_profile: str = None):
...
             antivirus_rules = prettify_profiles_rules(rules)
-            content = [{
+            virus_content = [{
                 'Name': profiles['@name'],
                 'Rules': antivirus_rules
             }]
 
-        human_readable += tableToMarkdown('Antivirus Profiles', content)
-        context.update({"Panorama.Antivirus(val.Name == obj.Name)": content})
+        human_readable += tableToMarkdown('Antivirus Profiles', virus_content, headers=['Name', 'Decoder'])
```
yaakovi, line 5589: why `virus_content` have either `Rules` or 'Decoder'? why the table only have `Decoder`?

@bakatzir can you give your 2 cents here about the expected behavior?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5718,13 +5723,21 @@ def apply_security_profile(pre_post: str, rule_name: str, profile_type: str, pro
...
+        return_results(f'The profile {profile_name} has been applied to the rule {rule_name}')
+
+    else:  # firewall instance
+        firewall_xpath = f"{XPATH_RULEBASE}rulebase/security/rules/entry[@name='{rule_name}']/profile-setting/"\
+                         f"profiles/{profile_type}"
 
-    apply_security_profile(pre_post, rule_name, profile_type, profile_name)
-    return_results(f'The profile {profile_name} has been applied to the rule {rule_name}')
+        apply_security_profile(firewall_xpath, profile_name)
+        return_results(f'The profile {profile_name} has been applied to the rule {rule_name}')
```
yaakovi, line 5740: instead of duplicating those lines, just declare a single `xpath` variable and move them after the condition block.
```suggestion
    apply_security_profile(firewall_xpath, profile_name)
    return_results(f'The profile {profile_name} has been applied to the rule {rule_name}')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5718,13 +5723,21 @@ def apply_security_profile(pre_post: str, rule_name: str, profile_type: str, pro
...
             raise Exception('Please provide the pre_post argument when applying profiles to rules in '
                             'Panorama instance.')
+        panorama_xpath = f"{XPATH_RULEBASE}{pre_post}/security/rules/entry[@name='{rule_name}']/profile-setting/"\
+                         f"profiles/{profile_type}",
+        apply_security_profile(panorama_xpath, profile_name)
+        return_results(f'The profile {profile_name} has been applied to the rule {rule_name}')
+
+    else:  # firewall instance
+        firewall_xpath = f"{XPATH_RULEBASE}rulebase/security/rules/entry[@name='{rule_name}']/profile-setting/"\
+                         f"profiles/{profile_type}"
```
yaakovi, line 5737: are you sure that in the firewall case the `rulebase` is constant?
```suggestion
        xpath = f"{XPATH_RULEBASE}{pre_post}/security/rules/entry[@name='{rule_name}']/profile-setting/"\
                f"profiles/{profile_type}"

    else:  # firewall instance
        xpath = f"{XPATH_RULEBASE}rulebase/security/rules/entry[@name='{rule_name}']/profile-setting/"\
                f"profiles/{profile_type}"
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
bakatzir: there can be application exceptions and theart id exceptions, but not always. so i would add them with a Null=True.
sent an image on slack.

yaakovi (approved):

PR #10791, opened 2021-01-08
Fix security and compliance parse results

GalRabin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32589

## Description
Command: o365-sc-get-search-action

Error: Unable to get full results from search action (Only first result returned).

Error src: Code - line 215 , When parsing all results which return as string the assumptions was that comma differentiate each item. however comma could be contained inside sender/reciever/subject.

Solution: Better regex handling.

## Minimum version of Demisto
- [ ] 5.0.0
- [x] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Customer validation.


yaakovi (approved):
update your branch

GalRabin:
Updating - Customer approved proposed solution.

PR #10850, opened 2021-01-12
Using sdk commit hash

hod-alpert:
Using sdk commit hash without mitmdump output logs.


Itay4:
@hod-alpert please make sure @yuvalbenshalom @bakatzir and @roysagi are in the loop here

bakatzir (approved):

PR #10868, opened 2021-01-13
SyslogSender - Updated incorrect params and args naming

reutshal:
- [ ] In Progress
- [X] Ready

## Related Issues
relates: https://github.com/demisto/etc/issues/32208

## Description
- Updated the `init_manager` method to use the `priority` parameter since there is no `logging_level` parameter and I believe this is the same intention.
- Updated the display name of the priority parameter from `Log priority to send` to `Log level to send` since that's the way it's written in the README and I think it's a clearer display name.
- Updated both `syslog_send_notification` and `syslog_send` methods to use the `level` argument rather than the `log_level` argument since it doesn't exist in the yml.

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


liorblob:
please add a unit test to validate messages are sent with correct level/priority

reutshal:
@liorblob Added UT

liorblob (approved):

yaakovi (approved):

PR #10870, opened 2021-01-13
XFE v2 args fix

reutshal:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
relates: https://github.com/demisto/etc/issues/32208

## Description

- Added the threshold argument as the first option argument in the IP, URL, and domain commands
- Removed the **long** argument from the following commands as 'raw_response option can be used instead:
    - ***file***
    - ***domain*** 
    - ***url***
    - ***ip***
- Upgraded the Docker image to demisto/python3:3.9.1.14969.

   - [ ] Yes
   - [X] No
          Not breaking but I removed the 'long' argument

## Must have
- [ ] Tests
- [ ] Documentation


Packs/XForceExchange/ReleaseNotes/1_0_6.md
```diff
@@ -0,0 +1,12 @@
+
+#### Integrations
+##### IBM X-Force Exchange v2
+- The **threshold** argument in the ***ip*** command now works as expected.
+- The **threshold** argument in the ***url*** command now works as expected.
+- The **threshold** argument in the ***domain*** command now works as expected.
```
bakatzir, line 6:
```suggestion
- Fixed an issue where the **threshold** argument in the ***ip*** command was taking the value of the parameter instead.
- Fixed an issue where the **threshold** argument in the ***url*** command was taking the value of the parameter instead.
- Fixed an issue where the **threshold** argument in the ***domain*** command was taking the value of the parameter instead.
```

bakatzir (approved):

yaakovi (approved):

PR #10880, opened 2021-01-14
TAXII Feed - added result decoding

reutshal:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32773

## Description
This solution was suggested by the cs and it's fixing the customers issue + working on our side too.
Since I couldn't reproduce the bug, I still want to verify some things in the solution with the cs.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/FeedTAXII/Integrations/FeedTAXII/FeedTAXII.py
```diff
@@ -729,14 +729,18 @@ def _poll_collection(self, poll_service, begin, end):
...
         )
+        result.raw.decode_content = True
 
         while True:
             result_part_number = None
             result_id = None
             more = None
             tag_stack = collections.deque()  # type: ignore
+            xml_parser = etree.iterparse(result.raw, events=('start', 'end'), recover=True, dtd_validation=False,
+                                         load_dtd=False, encoding='utf-8')
```
DeanArbel, line 740: Why did you remove this from the try block? 
`load_dtd` and `dtd_validation` are False by default, please remove (see: https://lxml.de/api/lxml.etree.iterparse-class.html).
Why did you add `encoding` for `utf-8`? What happens when you don't provide encoding?

reutshal: As I said, the cs provided this solution and I asked her about those exact additional args :)
I will update here when I have more information.
I will move the line inside the try block again

yaakovi (commented):
do we have a recording of a bad reply? can we add a test for it?

reutshal:
> do we have a recording of a bad reply? can we add a test for it?

No, according to the ticket the customer is a governmental entity and they cannot share details about it or any examples,
And we will probably won't be able to reproduce

reutshal:
Added only the decode_content line since the customer replied:
image

yaakovi (approved):

PR #6423, opened 2020-04-07
Fix for WildFire upload command

teizenman:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/17534

## Description
Changed the way the file is being sent, now uses the `file.read()` has the data being sent.

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [X] No

PR #10897, opened 2021-01-16
WF upload-file rm .read

bakatzir:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/32772

## Description
relates also to: https://github.com/demisto/content/pull/6423
@teizenman can you share which files were failed to upload that made you add the `.read`?
with it uploading `.js` files is not possible, without it is ok.


bakatzir:
I was able to upload all the files from the previous referenced issue successfully, ready for reveiw.

teizenman (approved):

PR #10898, opened 2021-01-16
add test js file

bakatzir:
relates to: https://github.com/demisto/content/pull/10897

### DIFF
--- a/TestData/test.js
+++ b/TestData/test.js
@@ -1,0 +1,6 @@
+function test(a, b,) {
+	
+	var a = 3;
+	var b = 4;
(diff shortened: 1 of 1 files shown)


yaakovi (approved):

yaakovi:
better to add it as a test file in the relevant pack folder (so we can tell which test uses it).
also, if this is for a test playbook usage, then you can just create it during the test instead of downloading it (as far as I can tell, all you care about it the file extension)

bakatzir:
You are correct. will create in the TPB.

PR #10921, opened 2021-01-17
Qradar-Added another API call to the test module

reutshal:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32820

## Description
Added another endpoint to the test module in case of bad permissions/lack of offenses in the instance

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


DeanArbel (approved):
This solution works.

bakatzir (commented):
I would suggested more in the fetch functionality in the test-module, if you are already touching this. parsing the fetch params and validating them is always a good idea.

bakatzir (approved):

PR #10936, opened 2021-01-18
[Contribution Management] Removed automatic assignment of technical writer

DeanArbel:
## Status
- [x] Ready

## Description
Removed automatic assignment of @kirbles19 when creating internal PRs from external contribution PRs.

### DIFF
--- a/Utils/github_workflow_scripts/create_internal_pr.py
+++ b/Utils/github_workflow_scripts/create_internal_pr.py
@@ -19,7 +19,7 @@ def main():
     1. Creates new PR.
         A) Uses body of merged external PR as the body of the new PR.
         B) Uses base branch of merged external PR as head branch of the new PR to master.
-        C) Adds 'docs-approved' label if it was on the merged external PR, otherwise assigns 'kirbles19' as a reviewer.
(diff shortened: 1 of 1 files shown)


avidan-H:
How come we want to remove the assignment of a technical writer if the external PR was merged without being `docs-approved`? The internal PR will still require the `docs-approved` label in order to merge it.

avidan-H:
@DeanArbel if you are so inclined, you can add a condition to assign a technical writer if the external PR wasn't `docs-approved` and the pack's support level is `xsoar`. Might be more hassle than it's worth though, so I'll leave it up to you.

avidan-H (approved):

DeanArbel:
> @DeanArbel if you are so inclined, you can add a condition to assign a technical writer if the external PR wasn't `docs-approved` and the pack's support level is `xsoar`. Might be more hassle than it's worth though, so I'll leave it up to you.

I believe the contribution flow (which can change again in the future) will be easier to manage without adding this automation. There are cases where we wouldn't necessarily request a technical writer review for an `xsoar` pack, so it makes sense to have the reviewer use discretion for every pack.

PR #10942, opened 2021-01-18
fix qradar missing events

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32795

## Description
added a retry method to events fetch

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

   - [ ] Yes
   - [x] No


Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -805,12 +811,18 @@ def perform_offense_events_enrichment(
     events_query = {"headers": "", "query_expression": query_expression}
     print_debug_msg(f'Starting events fetch for offense {offense["id"]}.', client.lock)
     try:
-        query_status, search_id = try_create_search_with_retry(
-            client, events_query, offense
-        )
-        offense["events"] = try_poll_offense_events_with_retry(
-            client, offense["id"], query_status, search_id
-        )
+        for i in range(MAX_FETCH_EVENT_RETIRES):
```
yaakovi, line 814: in case the functions raise an exception, do we want to keep trying?
if so, then the `try-except` block should be inside the loop.
```python
for i in range(MAX_FETCH_EVENT_RETIRES):
    try:
        ...
    except Exception as e:
        ...
```

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -805,12 +811,18 @@ def perform_offense_events_enrichment(
...
+            query_status, search_id = try_create_search_with_retry(
+                client, events_query, offense
+            )
+            offense["events"] = try_poll_offense_events_with_retry(
+                client, offense["id"], query_status, search_id
+            )
```
yaakovi, line 820: why do we need to loop here for retries? looking at the name of the functions they should handle it on their own. add a comment explaining that

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -868,6 +880,8 @@ def try_poll_offense_events_with_retry(
             print_debug_msg(f"Error while fetching offense {offense_id} events, search_id: {search_id}. "
                             f"Error details: {str(e)}")
             failures += 1
+            if failures < max_retries:
```
yaakovi, line 883: should this be added in `try_create_search_with_retry` as well?

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
esharf: they retrying in case of error, not in case of missing events.

yaakovi (approved):

Packs/QRadar/ReleaseNotes/1_2_10.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### IBM QRadar v2
+- Fixed an issue where ***fetch-incidents*** missed events when fetching an offence with unindexed events.
```
yaakovi, line 4:
```suggestion
- Fixed an issue where ***fetch-incidents*** missed events when fetching an offense with unindexed events.
```

PR #10952, opened 2021-01-19
Support Deprecated Packs in Build

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31146

## Description
Fix a build-bug where the build would fail when deprecating a pack, i.e. adding `"hidden": true` to the pack_metadata.json.


darkushin:
Following [this comment](https://github.com/demisto/etc/issues/31146#issuecomment-751184003) in the issue I ran the following tests:
1) Before making any changes, I deprecated the `ServiceDeskPlus` pack (i.e. added `"hidden": true` to the pack_metadata.json) and checked the build output. As seen in the screenshot below, the pack was collected in the **Collect Test List And Content Packs** step in the **Create Instances** job.
image ([build](https://app.circleci.com/pipelines/github/demisto/content/61716/workflows/03b92b3f-b48b-4816-9373-a741a566d844/jobs/255975))

  After adding the required changes, we can see in the screenshot below that the `ServiceDeskPlus` pack was not collected and will not be installed.
  image([build](https://app.circleci.com/pipelines/github/demisto/content/62197/workflows/8404b2b8-74ea-4c68-8eb5-fe318f12ed54/jobs/257963))
  
  The purpose of this test was to check that deprecated packs are not being installed and that their tests are not collected.

2) Deprecating a pack that is a dependency pack of another pack: I deprecated the pack `Compliance` and made changes to the `SANS` pack which depends on the `Compliance` pack. As seen in the screenshot below, in the **Collect Test List And Content Packs** step in the **Create Instances** job only the `SANS` pack was collected and not the `Compliance` pack (as desired).
image
When reaching the **Server 6_0** job in the build, the job fails in the **Install Content And Configure Integrations On Server **, as it requires an installation of a deprecated pack. The build failure error can be seen below.
image ([build](https://app.circleci.com/pipelines/github/demisto/content/62185/workflows/6d778c93-a16c-4daf-85f3-8378b5312df3/jobs/257910))

Tests/scripts/utils/content_packs_util.py
```diff
@@ -35,10 +35,28 @@ def is_pack_xsoar_supported(pack_path: str) -> bool:
...
+
+    Args:
+        pack_path (str): The pack path
+
+    Returns:
+        True if the pack is deprecated, False otherwise
+    """
+    pack_metadata_path = os.path.join(pack_path, PACKS_PACK_META_FILE_NAME)
+    if not os.path.isfile(pack_metadata_path):
+        return True
```
darkushin, line 50: @yuvalbenshalom @ShahafBenYakir please let me know what you think about this. It checks whether the given path to the pack_metadata file is valid (similar to the check in the function `is_pack_xsoar_supported` above), and in case it isn't it returns True which will cause the pack NOT to be collected. Is this the behavior we want or maybe the opposite (collect the pack to the installed packs)?

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -372,7 +374,18 @@ def search_pack_and_its_dependencies(client: demisto_client,
 
         current_packs_to_install = [pack_data]
         if dependencies:
-            current_packs_to_install.extend(dependencies)
+            # Check that the dependencies don't include a deprecated pack:
+            for dependency in dependencies:
+                pack_path = os.path.join(PACKS_DIR, dependency.get('id'))
+                if is_pack_deprecated(pack_path):
+                    logging.exception(f'Pack {pack_id} depends on pack {dependency.get("id")} which is a deprecated '
```
hod-alpert, line 381:
```suggestion
                    logging.critical(f'Pack {pack_id} depends on pack {dependency.get("id")} which is a deprecated '
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -372,7 +374,18 @@ def search_pack_and_its_dependencies(client: demisto_client,
...
+                                      f'pack.')
+                    lock.acquire()
```
hod-alpert, line 383: If `SUCCESS_FLAG` is initialized as `True` and the threads can only set it to `False` there is no need to lock it since even in a race condition where two threads will try to set it to False at the same time - the result will be False in the end.

Tests/scripts/utils/content_packs_util.py
```diff
@@ -49,4 +67,4 @@ def should_test_content_pack(pack_name: str) -> bool:
     if not pack_name:
         return False
     pack_path = os.path.join(PACKS_DIR, pack_name)
-    return pack_name not in SKIPPED_PACKS and is_pack_xsoar_supported(pack_path)
+    return pack_name not in SKIPPED_PACKS and is_pack_xsoar_supported(pack_path) and not is_pack_deprecated(pack_path)
```
ShahafBenYakir, line 70:
```suggestion
    if  pack_name in SKIPPED_PACKS:
        return False, 'Pack is either the "NonSupported" Pack or the "DeprecatedContent" Pack.
    if not is_pack_xsoar_supported(pack_path):
        return False, 'Pack is not XSOAR supported".
    if is_pack_deprecated(pack_path)
        return False, 'Pack is Deprecated'
    Return True, _
```

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -997,7 +997,7 @@ def remove_tests_for_non_supported_packs(tests: set, id_set: dict) -> set:
         if content_pack_name_list:
             id_set_test_playbook_pack_name = content_pack_name_list[0]
 
-            # We don't want to test playbooks from Non-certified partners.
+            # We don't want to test playbooks from Non-certified partners or from deprecated packs.
```
ShahafBenYakir, line 1000:
```suggestion
            # We don't want to test playbooks that are not XSOAR supported, belong to the ignored packs or from deprecated packs.
```

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -997,7 +997,7 @@ def remove_tests_for_non_supported_packs(tests: set, id_set: dict) -> set:
...
             if not should_test_content_pack(id_set_test_playbook_pack_name):
                 tests_that_should_not_be_tested.add(test)
```
ShahafBenYakir, line 1002:
```suggestion
            should_test, reason = should_test_content_pack(id_set_test_playbook_pack_name)
            if not should_test
                tests_that_should_not_be_tested.add(f'{test}: {reason}')
```

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -1107,7 +1108,8 @@ def get_test_list_and_content_packs_to_install(files_string,
     packs_to_install = packs_to_install.union(packs_of_collected_tests)
 
     # All filtering out of packs should be done here
-    packs_to_install = {pack_to_install for pack_to_install in packs_to_install if pack_to_install not in IGNORED_FILES}
+    packs_to_install = {pack_to_install for pack_to_install in packs_to_install if pack_to_install not in IGNORED_FILES
+                        and should_test_content_pack(pack_to_install)}
```
ShahafBenYakir, line 1112:
```suggestion
                        and should_test_content_pack(pack_to_install)[0]}
```

ShahafBenYakir (commented):
Nice! 

https://github.com/demisto/content/blob/9818bbee1880f00b0616533bd397a9216e39e1bc/Tests/scripts/collect_tests_and_content_packs.py#L983-L985
 Let's change this also to: 
```
Filters out test playbooks, which belong to one of the following packs: 
- Non XSOAR supported packs
- DeprecatedContent, NonSupported Packs
- A deprecated pack.
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -372,7 +374,16 @@ def search_pack_and_its_dependencies(client: demisto_client,
 
         current_packs_to_install = [pack_data]
         if dependencies:
-            current_packs_to_install.extend(dependencies)
+            # Check that the dependencies don't include a deprecated pack:
+            for dependency in dependencies:
+                pack_path = os.path.join(PACKS_DIR, dependency.get('id'))
+                if is_pack_deprecated(pack_path):
```
anara123, line 380: why we don't pass pack_data to `is_pack_deprecated`, isn't the data contains the metadata ?
passing the path, means opening and loading files each time this function is called just to get single field

darkushin: @anara123 The `pack_data` returned from the `search_pack()` function includes only ID and version fields. It doesn't include the metadata of the pack.
image

hod-alpert: @anara123, @darkushin  we could add somewhere a file with all deprecated \ hidden packs and cache it so that all threads can just read it.

Maybe adding this info to the id_set where the id is the pack display name, any thoughs?

ShahafBenYakir: To check support we do the same thing as well. We can add a function that will create some kind of a cache but what will we save? time? I am not sure that building this cache upfront for all packs will be faster than performing the file reading operation a couple of times for each pack.

yuvalbenshalom: I think we can make this an enhancement If we will see that it extends the build time.

hod-alpert: Agreed, but let's check it before we merge to know better what to expect.

I would like to compare a build that installs say 20 packs that have non-trivial dependencies and see how much that adds to the build and then lets decide how much we want to prioritize it.

@bakatzir have just reduced the sdk build time by 8 minutes only by removing the repeating reading and parsing of the id_set file.

hod-alpert: BTW I see we are not using this flow in the nightly, is this on porpuse? @darkushin

yuvalbenshalom (approved):

Tests/Marketplace/search_and_install_packs.py
yuvalbenshalom: @hod-alpert 
as long as we are installing all of the packs in the bucket in the nightly, why should we filter?

ShahafBenYakir: @hod-alpert We do filter in nightly:
https://github.com/demisto/content/pull/10952/files#diff-25b80cc78a0efaee708db1ce39fb154ac3358eae6361e3e336922d8a77599863R1250

darkushin: @ShahafBenYakir we are filtering, but we are not checking for every pack that all its dependencies are not deprecated. So, for example, if pack A is deprecated and pack B depends on pack A, we will filter pack A but pack B will still be installed in nightly. Is it ok?

ShahafBenYakir: I think Pack B will not be installed. The operation will fail as it misses its dependency. no @hod-alpert?

darkushin: Yes, the entire build will fail and it will be written in the logs that pack B misses its dependency pack A. What I meant is that in nightly we will collect pack B and try to install it (but fail), whereas in a normal build we won't try to install pack B in the first place. I think it is ok this way.

hod-alpert: In that case I'm pretty sure the nightly will try to install it, fail and then try again without the problematic pack, it;s a feature @Shellyber implemented.

PR #10956, opened 2021-01-19
Hod/unmock playbook

hod-alpert:
- making playbooks unmockable
- Making 'ThreatCrowd - Test' run as single branch.


Itay4 (approved):
if we mark the tests as unmockable, should we remove the integrations from the unmockable list?

hod-alpert:
> if we mark the tests as unmockable, should we remove the integrations from the unmockable list?

We can, in this case it's not necessary because the integrations are mockable.

Itay4:
> > if we mark the tests as unmockable, should we remove the integrations from the unmockable list?
> 
> We can, in this case it's not necessary because the integrations are mockable.

i thought the integrations were set as unmockable, please ignore my comment

PR #10977, opened 2021-01-20
adding threshold to lastline client

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/32208

## Description
added the logic for *threshold* param.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
   - [x] No

### DIFF
--- a/Packs/Lastline/Integrations/Lastline_v2/Lastline_v2.py
+++ b/Packs/Lastline/Integrations/Lastline_v2/Lastline_v2.py
@@ -312,8 +312,8 @@ def main():
         'key': params.get('api_key'),
         'api_token': params.get('api_token')
     }
-    threshold = int(params.get('threshold', 70))
     api_params.update(demisto.args())
+    threshold = int(params.get('threshold', 70))
 
     if not credentials or not credentials.get('identifier') or not credentials.get('password'):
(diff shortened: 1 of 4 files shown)


reutshal (commented):
You also have a threshold argument for some of the command, why is it not in use?
I believe the order should be:
arg_threshold -> param_threshold -> default (70)

Packs/Lastline/Integrations/Lastline_v2/Lastline_v2.py
```diff
@@ -312,6 +313,7 @@ def main():
         'api_token': params.get('api_token')
     }
     api_params.update(demisto.args())
+    threshold = int(params.get('threshold', 70))
```
reutshal, line 316: This threshold is still taken from the params isn't it?
I suggest:
threshold = int(api_params.get('threshold, params.get('threshold', 70)))

esharf: Screen Shot 2021-01-20 at 11 33 20

PR #11022, opened 2021-01-22
Single branching 'AutoFocus V2 test' for proxy porpuses

hod-alpert:
## Related Issues
fixes: https://github.com/demisto/etc/issues/26464

## Description
Single branching the playbook

Build is failing on an actual playbook issue which is not related to the proxy and is not affected by this PR

### Before: 
image

### After:
image


DeanArbel (approved):
nice

PR #11032, opened 2021-01-23
Servicenow v2 - added delimiter argument

reutshal:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32909

## Description
Added a delimiter argument to all the relevant commands 
Added UT

@bakatzir @amshamah419 can you please also go over the fields_delimiter argument description in the yml?
Thanks!


Packs/ServiceNow/ReleaseNotes/2_1_7.md
```diff
@@ -0,0 +1,9 @@
+
+#### Integrations
+##### ServiceNow v2
+- Added the *fields_delimiter* arguments to the following commands:
```
bakatzir, line 4:
```suggestion
- Added the *fields_delimiter* argument to the following commands:
```

bakatzir (commented):
Do we know if we should allow any delimiter that we want? Should not it be more of a closed set?  @liorblob WDYT?

reutshal:
> Do we know if we should allow any delimiter that we want? Should not it be more of a closed set? @liorblob WDYT?

I think that most customers will not use it. 
But we do have a customer that one of their custom_fields is of the value of HTML and can contain the semicolon character inside.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -2336,9 +2376,6 @@ script:
   isfetch: true
   longRunning: false
   longRunningPort: false
-  ismappable: true
-  isremotesyncin: true
-  isremotesyncout: true
```
yaakovi, line 2341: why did you remove those?

yaakovi (approved):

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -191,6 +191,14 @@ script:
       name: additional_fields
       required: false
       secret: false
+    - default: false
+      defaultValue: ;
+      description: The delimiter character to use to separate any fields list argument
```
bakatzir, line 196: WDYT about this? if ok please tend to in all of the occurrences.
```suggestion
      description: The delimiter character to use to separate all of the fields provided in the custom_fields argument. Default is ";".
```

reutshal: The problem is, not only the 'custom_fields' argument uses the split method. There are some other arguments that are separated with semicolons. So I think that if the customer has changed this argument it should apply to all relevant arguments.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
reutshal: Adding back

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
amshamah419: > WDYT about this? if ok please tend to in all of the occurrences.
> 
> ```suggestion
> 
>       description: The delimiter character to use to separate all of the fields provided in the custom_fields argument. Default is ";".
> 
> ```

This sounds good. Let's apply this.

liorblob:
> Do we know if we should allow any delimiter that we want? Should not it be more of a closed set? @liorblob WDYT?

don't see a reason to limit this

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
amshamah419: > The problem is, not only the 'custom_fields' argument uses the split method. There are some other arguments that are separated with semicolons. So I think that if the customer has changed this argument it should apply to all relevant arguments.

I'm inclined to agree here. If anything we could add the delimiter argument to all pertinent commands with the default set to `;`, but give the customer the option to override it.

PR #11042, opened 2021-01-24
Jira-Create-Issue-Bug-Fix

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32986

## Description
Due to changes in the API due to GDPR - replaced the assignee argument to refer to the user's account ID rather than its name

### DIFF
--- a/Packs/Jira/Integrations/JiraV2/JiraV2.py
+++ b/Packs/Jira/Integrations/JiraV2/JiraV2.py
@@ -417,7 +417,7 @@ def get_issue_fields(issue_creating=False, **issue_args):
     if issue_args.get('assignee'):
         if not issue['fields'].get('assignee'):
             issue['fields']['assignee'] = {}
-        issue['fields']['assignee']['name'] = issue_args['assignee']
+        issue['fields']['assignee']['id'] = issue_args['assignee']
 
     if issue_args.get('reporter'):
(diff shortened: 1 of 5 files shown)


Packs/Jira/Integrations/JiraV2/JiraV2.py
```diff
@@ -417,7 +417,7 @@ def get_issue_fields(issue_creating=False, **issue_args):
     if issue_args.get('assignee'):
         if not issue['fields'].get('assignee'):
             issue['fields']['assignee'] = {}
-        issue['fields']['assignee']['name'] = issue_args['assignee']
+        issue['fields']['assignee']['id'] = issue_args['assignee']
```
yaakovi, line 420: better to keep the argument as an account name and in the code convert it to account ID.

yaakovi (changes requested):

Packs/Jira/Integrations/JiraV2/JiraV2.py
Bargenish: The endpoint in the API doc was marked as `EXPERIMENTAL` so I didn't want to depend on it.
Screen Shot 2021-01-24 at 17 00 09

yaakovi (approved):

PR #11051, opened 2021-01-24
Upload will change metadata in every upload

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32349
https://github.com/demisto/etc/issues/32287

## Description
When uploading packs we wish to update the metadata in the index in every upload.

## Screenshots

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Tests/Marketplace/upload_packs.py
+++ b/Tests/Marketplace/upload_packs.py
@@ -937,12 +937,6 @@ def main():
             pack.cleanup()
             continue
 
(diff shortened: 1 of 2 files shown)


ohaim1008:
I tried couple of options but this one looks like the best option.
It was checked in several scenarios - such as adding a tag to all packs in the index file but no showing them as changed packs in the slack message.

guyfreund (approved):
very elegant solution!

PR #11057, opened 2021-01-25
Supporting circle's 'commands' feature

hod-alpert:
## Related issue:
https://github.com/demisto/etc/issues/32500

## Description:
Modifying config.yml in order to implement the ssh tunneling more easily.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,16 @@
 version: 2.1
+commands:
+  set-instance-role-env-variable:
+    description: "Sets INSTANCE_ROLE env variable into $BASH_ENV file"
+    parameters:
+      instance_role:
+        type: string
+    steps:
+      - run:
+          name: Set INSTANCE_ROLE env variable
+          command: |
+            echo 'export INSTANCE_ROLE="<< parameters.instance_role >>"' >> $BASH_ENV
+            source $BASH_ENV
 parameters:
   non_ami_run:
     type: string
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -934,36 +896,10 @@ jobs:
...
-          shell: /bin/bash
-          when: always
-          command: |
-            if [ -z $INSTANCE_WAS_CREATED ];
-              then
-                echo "Skipping - instance was not created"
-                exit 0
-            fi
-            ./Tests/scripts/install_content_and_test_integrations.sh "$INSTANCE_ROLE"
+      - set-instance-role-env-variable
```
yaakovi, line 899: I think it is better to be implicit and pass the value here as well.

.circleci/config.yml
```diff
@@ -1075,6 +1011,9 @@ jobs:
       - *attach_workspace
       - *add_ssh_keys
       - *prepare_environment
+      - set-instance-role-env-variable:
+          instance_role: Server 5.5
+      - *wait_until_server_ready
```
yaakovi, line 1016: why did you keep the original `Wait until server ready`?

.circleci/config.yml
```diff
@@ -508,12 +521,34 @@ references:
...
+        python3 ./Tests/scripts/wait_until_server_ready.py "$INSTANCE_ROLE"
+
+  check_if_instance_was_created: &check_if_instance_was_created
+    run:
+      name: Check If Instance Was Created
+      when: always
+      command: |
+        export TEMP=$(cat ./Tests/filter_envs.json | jq ".\"$INSTANCE_ROLE\"")
+        echo "$INSTANCE_ROLE filter=$TEMP"
+        if [[ "$TEMP" == "true" || -n "${NIGHTLY}" ]];
```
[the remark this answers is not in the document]
hod-alpert: It's not relevant, but it's not harmful since in nightly the other machines jobs will not be triggered so this covers all cases well.

yaakovi (approved):

PR #11068, opened 2021-01-25
SplunkPy - redundant default args

bakatzir:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/33076
related: https://github.com/demisto/etc/issues/32855

## Description
see in referenced issue

   - [x] No


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.yml
```diff
@@ -459,7 +459,7 @@ script:
       description: List of collections.
       type: String
   - arguments:
-    - default: true
+    - default: false
```
Itay4, line 462: why are we setting that one to `false` as well? what is the behavior currently (before this change)? what happens if we pass a default arg?

bakatzir: this one is by mistake.

bakatzir, line 462:
```suggestion
    - default: true
```

yaakovi (approved):

PR #11072, opened 2021-01-25
CrowdStrike Falcon - pass missing args in retry when encountering 403

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32072

## Description
in case 403 occurred, not all args (`files` in this case) were passed on the second call to `http_request`

   - [x] No


yaakovi (changes requested):
rebase your branch please

yaakovi (approved):

PR #11100, opened 2021-01-26
EDL: save to integration context only required ioc data

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31888

## Description
save to context only the required two keys per indicator instead of the whole object.


Itay4:
@yaakovi can you please also change `demisto.getIntegrationContext()` to `get_integration_context()` ?

yaakovi:
> @yaakovi can you please also change `demisto.getIntegrationContext()` to `get_integration_context()` ?

Done

DeanArbel (approved):
Nice

PR #11109, opened 2021-01-27
Earlier Upload-Flow trigger

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
- Triggered both uploads to 11 (AM & PM) to prevent race conditions with nightly.
- Fixed an issue in Slack notifier where failed pack status was the name of the enum and not its value.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1318,7 +1318,7 @@ workflows:
   bucket_upload:
     triggers:
       - schedule:
-          # should trigger every day at 10 AM & 10 PM UTC (12:00 PM & 00:00 AM Israel Time)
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1319,7 +1319,7 @@ workflows:
     triggers:
       - schedule:
           # should trigger every day at 10 AM & 10 PM UTC (12:00 PM & 00:00 AM Israel Time)
```
Itay4, line 1321: please update comment as well

Itay4 (approved):

PR #11128, opened 2021-01-27
Workday events generator update

Shellyber:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33155

## Description
* Added `test-module` command
* Added command for every event type
* Added more information in the detailed description + added default port number
* Added entries to the war room and error handling
* Updated documentation

## Screenshots
Successful event:
image

Unsuccessful event:
image

## Minimum version of Demisto
- [x] 6.0.0

   - [x] No


Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+                "Last_Name": last_name,
+                "Job_Function": "Product Management Function",
+                "State": "California",
+                "Exec_Admin_Flag": "N",
+                "Preferred_Name": "Chester McBadbat",
+                "Regular_Employee_Flag": "Y",
+                "Preferred_Name_-_Last_Name": "McBadbat",
+                "Cost_Center_Code": "651116",
+                "Location": "Office - USA - CA - Headquarters"
+            }
```
dantavori, line 914: let's take this entry and put it as a constant outside the function (no need for the outer "Report_Entry") - for readability. then just modify and append it to the integration context.

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+    integration_context = get_integration_context()
+    reports = TERMINATE_USER_REPORT['Report_Entry']
+    for report in reports:
+        email = report.get('Email_Address')
+        if email == user_email:
+            users = [i for i in integration_context['Report_Entry'] if not (i['Email_Address'] == email)]
+            users.append(report)
+            integration_context['Report_Entry'] = users
+            set_integration_context(integration_context)
+            return
```
dantavori, line 937: could you please explain what is the logic here?

dantavori (commented):
I think that if we want to leverage this integration, we should support the following kinds of events' generation:
- One is by using all of our hard-coded reports (including the deleted NEW_HIRE_REPORT) as we currently do - a simple method to generate mock data for starters.
- The second one is a command we can call `workday-test-set-report-entry` that adds an entry (or updates an existing one) to the current report in the integration context, according to the arguments you pass (could be for example by a comma-separated list of indicatorField:value elements). it gives the users full control of the report, the type of events they want to generate and it could be very useful for them to understand which event is generated when.
for example, assuming we have the active employee `cchin@test.com`, we can do `!workday-test-set-report-entry email=cchin@test.com fields_to_update="Last_Hire_Date:01/01/2019"` and he will be terminated.
- Another command we can do is `workday-test-update-all` that will collectively update a specific field for all entries, to test multiple incidents at once. for example: `!workday-test-update-all field=Management_Level_1 value="Shelly Tzohar"`.

Edit: we can also implement some playbooks that will generate examples and check that the outcomes are as expected. But this is next level I guess.

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
-            if email == user_email:
-                integration_context['Report_Entry'].append(report)
-                set_integration_context(integration_context)
+def get_new_hire_reports(user_email, name):
+    integration_context = get_integration_context()
+    if not name or name == '':
+        return_error('Please insert the user full name (first name and last name).')
+    try:
+        first_name = name.split()[0]
+        last_name = name.split()[1]
```
anara123, line 864: if there are separate first name and last name, then we better have arg for each instead of splitting name arg.
replace name with `first_name` and `last_name` args

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+    for report in reports:
+        email = report.get('Email_Address')
+        if email == user_email:
+            users = [i for i in integration_context['Report_Entry'] if not (i['Email_Address'] == email)]
+            users.append(report)
+            integration_context['Report_Entry'] = users
+            set_integration_context(integration_context)
+            return
+        else:
+            return_error(f'The user "{user_email}" does not exist. Please choose one of the following: '
```
anara123, line 939: if you have 100 users, and the user you are looking for is at place 50, then the loop will fail at the first user, because  it will go to else branch

anara123 (changes requested):

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
-                users = [i for i in integration_context['Report_Entry'] if not (i['Email_Address'] == email)]
-                users.append(report)
-                integration_context['Report_Entry'] = users
-                set_integration_context(integration_context)
+    integration_context = get_integration_context()
+    reports = UPDATE_REPORT['Report_Entry']
+    for report in reports:
+        email = report.get('Email_Address')
+        if email == user_email:
+            users = [i for i in integration_context['Report_Entry'] if not (i['Email_Address'] == email)]
```
anara123, line 950: the logic here is not very clear to me. When we want to generate update report, we should get an existing report and update some of it's fields.
I would recommend to add to this command few args like `title` and `city`. Also add `field_name` and `field_value` for update of any field

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
-        integration_context = get_integration_context()
-        reports = NEW_HIRE_REPORT['Report_Entry']
-        for report in reports:
-            email = report.get('Email_Address')
-            if email == user_email:
-                integration_context['Report_Entry'].append(report)
-                set_integration_context(integration_context)
+def get_new_hire_reports(user_email, name):
+    integration_context = get_integration_context()
+    if not name or name == '':
```
Itay4, line 860:
```suggestion
    if not name:
```

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+        return_error('Please insert the user full name (first name and last name).')
```
Itay4, line 861: error message is a bit unclear, insert where?

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+    try:
+        first_name = name.split()[0]
+        last_name = name.split()[1]
+    except Exception:
```
Itay4, line 865: not sure i understand try-except here, if you wish to check that both first and last names were provided then just add a condition for that
also, dont ignore the exception, but at least log it

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
+        return_error('Missing first or last name.')
+    report_entry = {
+        "Report_Entry": [
```
Itay4, line 868: why did we move it from a const to be a func var?

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -970,81 +827,158 @@
...
-                set_integration_context(integration_context)
+def get_new_hire_reports(user_email, name):
+    integration_context = get_integration_context()
+    if not name or name == '':
+        return_error('Please insert the user full name (first name and last name).')
+    try:
+        first_name = name.split()[0]
+        last_name = name.split()[1]
+    except Exception:
+        return_error('Missing first or last name.')
```
Itay4, line 866: avoid using `return_error` in the code body, but raise exception instead

Shellyber:
Validate step fails due to: https://github.com/demisto/etc/issues/33991

Packs/Workday/ReleaseNotes/1_0_11.md
```diff
@@ -0,0 +1,9 @@
+
+#### Integrations
+##### Workday IAM Event Generator (Beta)
+- Fixed an issue where entries was not returned to the war room
+- Added new 4 commands:
```
dantavori, line 5:
```suggestion
- Fixed an issue where entries were not returned to the war room
- Added 4 new commands: 
```

dantavori (approved):

Packs/Workday/ReleaseNotes/1_0_11.md
```diff
@@ -0,0 +1,9 @@
+
+#### Integrations
+##### Workday IAM Event Generator (Beta)
+- Fixed an issue where entries were not returned to the war room
+- Added 4 new commands:
```
yaakovi, line 5:
```suggestion
- Added 4 commands: 
```

Packs/Workday/ReleaseNotes/1_0_11.md
```diff
@@ -0,0 +1,9 @@
+
+#### Integrations
+##### Workday IAM Event Generator (Beta)
+- Fixed an issue where entries were not returned to the war room
```
yaakovi, line 4:
```suggestion
- Fixed an issue where entries were not returned to the war room.
```

Packs/Workday/ReleaseNotes/1_0_11.md
```diff
@@ -0,0 +1,9 @@
...
+- Added 4 new commands: 
+    * *workday-generate-hire-event*
+    * *workday-generate-update-event*
+    * *workday-generate-rehire-event*
+    * *workday-generate-terminate-event*
```
yaakovi:
```suggestion
- Fixed an issue where entries were not returned to the war room.
- Added 4 commands: 
  - ***workday-generate-hire-event***
  - ***workday-generate-update-event***
  - ***workday-generate-rehire-event***
  - ***workday-generate-terminate-event***
```

yaakovi:
```suggestion
- Fixed an issue where entries were not returned to the war room.
- Added 4 commands: 
  - ***workday-generate-hire-event***
  - ***workday-generate-update-event***
  - ***workday-generate-rehire-event***
  - ***workday-generate-terminate-event***
```

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -478,585 +481,215 @@
...
+    elif demisto.command() == 'workday-generate-update-event':
+        get_update_report()
 
+    elif demisto.command() == 'workday-generate-rehire-event':
+        get_rehire_report()
 
-elif demisto.command() == 'generate-event':
-    generate_event()
+    elif demisto.command() == 'workday-generate-terminate-event':
+        get_terminate_report()
```
anara123, line 687: replace the function names to be `generate_*` and not `get_*` , because we generate terminate report

anara123: same for other functions

Packs/Workday/Integrations/WorkdayIAMEventsGenerator/WorkdayIAMEventsGenerator.py
```diff
@@ -478,585 +481,215 @@
...
+
+def get_terminate_report():
+    user_email = demisto.args().get('user_email')
+    integration_context = get_integration_context()
+    existing_email_addressees = []
+    now = datetime.now()
+    current_date = now.strftime("%m/%d/%Y")
+    for report in integration_context['Report_Entry']:
+        email_address = report.get('Email_Address')
+        existing_email_addressees.append(email_address)
```
anara123, line 590: instead of running twice over the list, I recommend to check if `email_address==user_email` then set `user_report = report`
This way you don't need to keep `existing_email_addressees`

anara123 (approved):
see my comments

PR #11145, opened 2021-01-28
MSGraphGroupsTPBFix

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27890

## Description
Added a verification step before adding a member to a group

## Screenshots
Microsoft_Graph_Groups_-_Test_Thu_Jan_28_2021


bakatzir (approved):
If you do that (branches) it means it can not be mocked IIRC.

Bargenish:
@bakatzir I actually removed the branches and single branched it.
The condition does not count for a branch since the 2 requests would never run simultaneously

bakatzir:
Good to learn. thx.

PR #11146, opened 2021-01-28
demistomock: add internalHttpRequest with doc string

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
internalHttpRequest 

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -1048,7 +1048,7 @@ def get_incidents():
 
 def internalHttpRequest(method, uri, body=None):
     """Run an internal HTTP request to the XSOAR server. The request runs with the permissions of the 
-    executing user, when a command is being executed manually (such as via the warroom or when browsing a widget).
+    executing user, when a command is being executed manually (such as via the War Room or when browsing a widget).
     When run via a playbook, will run with a readonly user with limited permissions isolated to the current incident only.
     Available for both Integrations and Scripts starting from Server version 6.1.
 
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -1047,7 +1047,7 @@ def get_incidents():
 
 
 def internalHttpRequest(method, uri, body=None):
-    """Run an internal HTTP request to the XSOAR server. The request runs with the permissions of the 
+    """Run an internal HTTP request to the XSOAR server. The request runs with the permissions of the
     executing user, when a command is being executed manually (such as via the War Room or when browsing a widget).
     When run via a playbook, will run with a readonly user with limited permissions isolated to the current incident only.
     Available for both Integrations and Scripts starting from Server version 6.1.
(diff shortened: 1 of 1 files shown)


Tests/demistomock/demistomock.py
```diff
@@ -1044,3 +1044,49 @@ def getModules():
 
 def get_incidents():
     return {}
+
+
+def internalHttpRequest(method, uri, body=None):
+    """Run an internal HTTP request to the XSOAR server. The request runs with the permissions of the 
+    executing user, when a command is being executed manually (such as via the warroom or when browsing a widget).
```
Itay4, line 1051:
```suggestion
    executing user, when a command is being executed manually (such as via the War Room or when browsing a widget).
```

Tests/demistomock/demistomock.py
```diff
@@ -1044,3 +1044,49 @@ def getModules():
...
+    When run via a playbook, will run with a readonly user with limited permissions isolated to the current incident only.
+    Available for both Integrations and Scripts starting from Server version 6.1.
+
+    Args:
+        method (str): HTTP method such as: GET or POST
+        uri (str): Server uri to request. For example: "/contentpacks/marketplace/HelloWorld".
+        body (str, optional): Optional body for a POST request. Defaults to None.
```
Itay4, line 1058:
```suggestion
        body Optional[str]: Optional body for a POST request. Defaults to None.
```

Tests/demistomock/demistomock.py
```diff
@@ -1044,3 +1044,49 @@ def getModules():
...
+
+    Returns:
+        dict: dict cotainnig the following fields:
+        * statusCode (int): HTTP status code such as 200
```
Itay4, line 1062: dont think this will render well, my guess is that it should be indented like the `incidents` arg of `createIncidents`

Tests/demistomock/demistomock.py
Itay4, line 1058: for all other functions, the format for default value is `(Default value = None)` so i think better to change that or change others to stay consistent

Tests/demistomock/demistomock.py
```diff
@@ -1044,3 +1044,49 @@ def getModules():
 
 def get_incidents():
     return {}
+
+
+def internalHttpRequest(method, uri, body=None):
+    """Run an internal HTTP request to the XSOAR server. The request runs with the permissions of the 
+    executing user, when a command is being executed manually (such as via the warroom or when browsing a widget).
+    When run via a playbook, will run with a readonly user with limited permissions isolated to the current incident only.
+    Available for both Integrations and Scripts starting from Server version 6.1.
```
Itay4, line 1053: dont think need to mention it is available for both, as for other functions we note if its for only integrations/scripts, and if its available in both we just dont comment that

glicht: Worth mentioning as this is a special one.

Tests/demistomock/demistomock.py
glicht: It renders ok: 
image

Itay4 (approved):

PR #11176, opened 2021-01-31
Added file retrieval from Cortex XDR to the generic playbook

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32979
fixes: https://github.com/demisto/etc/issues/33379

## Description
* Added the `Cortex XDR - Retrieve File Playbook` to the `Get File Sample From Path - Generic V2` playbook. Added the needed inputs to the `Get File Sample From Path - Generic V2` and the `Retrieve File from Endpoint - Generic V2` playbook.
* Added skip if unavailable to the playbook for both XDR and Carbon Black

## Screenshots
Get_File_Sample_From_Path_-_Generic_V2

## Minimum version of Demisto
5.0.0

No


idovandijk:
Will be handled by @EliorKedar

PR #11185, opened 2021-01-31
CommonServerPython + ApiModules: Fixed insecure error message

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33205

## Description
don't suggest checking insecure param when already checked.


Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule.py
```diff
@@ -74,7 +74,7 @@ def http_request(self, method, url_suffix, full_url=None, headers=None, json_dat
                                     f'\n{res}')
 
         except Exception as e:
-            if 'SSL Certificate Verification Failed' in e.args[0]:
+            if self._verify and 'SSL Certificate Verification Failed' in e.args[0]:
```
Itay4, line 77: shouldnt this be `if not self._verify` ?

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -2452,6 +2452,12 @@ def test_http_request_ssl_error(self, requests_mock):
         with raises(DemistoException, match="SSL Certificate Verification Failed"):
             self.client._http_request('get', 'event', resp_type='response')
 
+    def test_http_request_ssl_error_insecure(cls, requests_mock):
+        requests_mock.get('http://example.com/api/v2/event', exc=requests.exceptions.SSLError('test ssl'))
+        client = cls.BaseClient('http://example.com/api/v2/', ok_codes=(200, 201), verify=False)
+        with raises(requests.exceptions.SSLError, match="test ssl"):
+            client._http_request('get', 'event', resp_type='response')
```
Itay4, line 2459: can we check here that the string "try selectin 'Trust any certificate'...' is not in the exception message?

Packs/ApiModules/Scripts/ServiceNowApiModule/ServiceNowApiModule.py
yaakovi: no, if the verify is true, it means that the checkbox is disabled (insecure is "not verify").

Itay4 (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
yaakovi: updated the match string, now it will test for an exact match.

PR #11196, opened 2021-02-01
fixing vulture error in ML

roysagi:
Fixing vulture error.


lperryp (approved):

bakatzir (approved):

bakatzir:
@lperryp We are merging this per the slack conversation as it is an infra blocker for MP pack upload. we can revise tomorrow if it generates any difficulties.

PR #11220, opened 2021-02-02
Update conf fix

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://app.circleci.com/pipelines/github/demisto/content/65138/workflows/44d51dd0-72eb-4aa2-ae46-65a28c8e23e0/jobs/268954

## Description
Looks like the timeout was reached and update_conf_json_fails.
Increased timeout, added try-except, and loggings.


Tests/scripts/update_conf_json.py
```diff
@@ -127,9 +128,16 @@ def main():
...
-                                          args=(pack_name, existing_test_playbooks), timeout=20)
-            future_object.add_done_callback(update_new_conf_json)
+                                          args=(pack_name, existing_test_playbooks), timeout=30)
+            try:
+                future_object.add_done_callback(update_new_conf_json)
+                logging.debug(f'Successfully added pack: {pack_name} test  to conf.json')
+
+            except Exception as err:
+                logging.error(f'Failed adding pack: {pack_name} tests:\n{err}')
+                sys.exit(1)
```
hod-alpert, line 140: Catching the exception here will not help you, you should add it to the `update_new_conf_json` method instead

hod-alpert: If what you're trying to achieve is to fail the build - add `raise` to line 124

Tests/scripts/update_conf_json.py
```diff
@@ -127,9 +128,16 @@ def main():
...
             future_object = pool.schedule(generate_pack_tests_configuration,
-                                          args=(pack_name, existing_test_playbooks), timeout=20)
-            future_object.add_done_callback(update_new_conf_json)
+                                          args=(pack_name, existing_test_playbooks), timeout=30)
+            try:
+                future_object.add_done_callback(update_new_conf_json)
+                logging.debug(f'Successfully added pack: {pack_name} test  to conf.json')
+
+            except Exception as err:
+                logging.error(f'Failed adding pack: {pack_name} tests:\n{err}')
```
hod-alpert, line 139:
```suggestion
            except Exception:
                logging.exception(f'Failed adding pack: {pack_name} tests')
```

Tests/scripts/update_conf_json.py
```diff
@@ -114,14 +114,15 @@ def generate_pack_tests_configuration(pack_name, existing_test_playbooks):
     return pack_integrations, pack_test_playbooks, pack_name
 
 
-def update_new_conf_json(future):
+def update_new_conf_json(future, pack_name):
```
hod-alpert, line 117:
```suggestion
def update_new_conf_json(future):
```
This won't work either, The process pull calls this method and it calls it only with the future as argument

Tests/scripts/update_conf_json.py
```diff
@@ -114,14 +114,15 @@ def generate_pack_tests_configuration(pack_name, existing_test_playbooks):
...
     try:
         pack_integrations, pack_test_playbooks, pack_name = future.result()  # blocks until results ready
         if pack_test_playbooks:
             NEW_CONF_JSON_OBJECT.extend(calc_conf_json_object(pack_integrations, pack_test_playbooks))
-    except Exception:
-        logging.exception('Failed to collect pack test configurations')
 
+    except Exception:
+        logging.exception(f'Failed to collect pack test configurations for pack: {pack_name}')
+        sys.exit(1)
```
hod-alpert, line 125:
```suggestion
        raise
```

Tests/scripts/update_conf_json.py
ShahafBenYakir: Can't we get the argument from the process?

ShahafBenYakir: Don't we have a way to get the pack name?

Tests/scripts/update_conf_json.py
```diff
@@ -118,18 +118,21 @@ def update_new_conf_json(future):
...
+
     except Exception:
         logging.exception('Failed to collect pack test configurations')
-
+        raise
 
 def main():
     install_logging('Update_Tests_step.log', include_process_name=True)
     existing_test_playbooks = load_test_data_from_conf_json()
     with ProcessPool(max_workers=os.cpu_count(), max_tasks=100) as pool:
```
hod-alpert, line 129: Now that we know about this issue in circle - https://github.com/demisto/etc/issues/33244

Lets reduce the number of cpu's to something between 2 to 4.

Tests/scripts/update_conf_json.py
hod-alpert: Not here
But if its something other then timeout - it will raise exception in the processing itself and it will be more verbose

hod-alpert (approved):

PR #11231, opened 2021-02-02
XDR - Fixed an issue when the user is none

reutshal:

## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33306

## Description
Added a check that the users value is not None 

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CortexXDR/Integrations/CortexXDRIR/CortexXDRIR.py
```diff
@@ -1228,11 +1228,13 @@ def create_account_context(endpoints):
     for endpoint in endpoints:
         domain = endpoint.get('domain')
         if domain:
-            for user in endpoint.get('users', []):
-                account_context.append({
-                    'Username': user,
-                    'Domain': domain,
-                })
+            users = endpoint.get('users', [])  # in case the value of 'users' is None
+            if users and isinstance(users, list):
```
yaakovi, line 1232: can `users` be anything other than a list (or None)? if not, then no need to check `isinstance(users, list)`

yaakovi (commented):
add a unit test

Packs/CortexXDR/Integrations/CortexXDRIR/CortexXDRIR.py
reutshal: That's the thing, I'm not sure and the documentation does not help so I wanted to avoid any further errors like the previous one. From what I've seen, even one value there is present inside a list.

yaakovi (approved):

PR #11289, opened 2021-02-06
SetIfEmpty: Fix set if empty non ascii chars

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33445

## Description
improved handling of non-ASCII chars.
tested with playbook as the error is within the demisto class and cannot be tested via unit tests.

### DIFF
--- a/Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
+++ b/Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
@@ -10,7 +10,13 @@ def get_value_to_set(args):
     if value is None or (apply_if_empty and is_value_empty(value)):
         value = args.get('defaultValue')
 
-    return encode_string_results(value)
+    if isinstance(value, list):
+        for i, item in enumerate(value):
+            value[i] = encode_string_results(item)
+
+        return value
+    else:
+        return encode_string_results(value)
 
 
 def is_value_empty(value):
--- a/Packs/CommonScripts/ReleaseNotes/1_3_18.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_3_18.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### SetIfEmpty
+- Fixed an issue where the script failed on non-ASCII data.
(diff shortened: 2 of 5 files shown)


Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
```diff
@@ -10,7 +10,13 @@ def get_value_to_set(args):
     if value is None or (apply_if_empty and is_value_empty(value)):
         value = args.get('defaultValue')
 
-    return encode_string_results(value)
+    if isinstance(value, list):
+        for i, item in enumerate(value):
+            value[i] = encode_string_results(item)
```
reutshal, line 15:
```suggestion
        for key, item in enumerate(value):
            value[key] = encode_string_results(item)
```

Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
```diff
@@ -10,7 +10,13 @@ def get_value_to_set(args):
...
+
+        return value
+    else:
+        return encode_string_results(value)
```
reutshal, line 19: Can the value be a dict? will the encoding work on it?

reutshal (commented):
Can you attach an image of the playbook?

Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
yaakovi: I was able to reproduce it only with a list of strings

Packs/CommonScripts/Scripts/SetIfEmpty/SetIfEmpty.py
yaakovi: since `value` is a list, `i` stands for index.

yaakovi:
> Can you attach an image of the playbook?

SetIfEmpty_-_non-ascii_chars_-_Test_Sat_Feb_06_2021

reutshal (approved):

PR #11297, opened 2021-02-07
SymantecATP-FetchIncidentFix

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32727

## Description
Fixed an issue where the current timestamp was set as the last run timestamp instead of the greatest incident timestamp.
*The customer used this version of the integration for several days to test it and confirmed it.*


Packs/Symantec_Advanced_Threat_Protection/Integrations/integration-Symantec_Advanced_Threat_Protection.yml
```diff
@@ -384,25 +396,29 @@ script:
...
             }
         }
         next = 'next' in res.obj ? res.obj.next : null;
-        
+
         if (next) {
             setLastRun({ time: startTime, next: next, end_time: endTime });
         } else {
-            setLastRun({ time: endTime });
+            setLastRun({ time: currentFetch ? currentFetch : startTime});
```
yaakovi, line 421: are you sure it should be `startTime`?

Bargenish: if no incident was fetched then I don't want the last run to be updated to `now` but to remain the same (as we do in other integrations)

yaakovi (approved):

PR #11299, opened 2021-02-07
Don't search for server integrations in id_set

hod-alpert:
Ignoring server integrations in the tests and packs collection


ShahafBenYakir:
As we talked, this should be solved on the SDK right?

hod-alpert:
> As we talked, this should be solved on the SDK right?
Yes, and still we should not fail on it if it reproduces IMO.
If you think we should fail on it - then LMK and I'll improve the exception raised to be more informative.

ShahafBenYakir:
I actually think it should fail... This could be an issue not only for sever integrations and we would want to know that this integration does not appear in the id_set.
Let's fail it with an indicative message with the name of the integration

ShahafBenYakir (approved):

PR #11319, opened 2021-02-08
Fix: Avoid HTML template formatting if HTML body is not used

0xThiebaut:
## Status
- [ ] In Progress
- [X] Ready
- [ ] In Hold

## Description
By default the `htmlBody` defaults to `None` resulting in a formatting exception should a template be used. This commit introduces an additionnal check to ensure `htmlBody` is not `None`.

   - [ ] Yes
   - [X] No


ChanochShayner:
I'm closing this PR and I'll do the fix here - https://github.com/demisto/content/pull/11333

PR #11331, opened 2021-02-08
CSP GetRemoteDataResponse Bugfix

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
Not every mirrored object uses the `id` field to store its unique identifier, e.g. Splunk uses the `event_id` field.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -5894,7 +5894,6 @@ class GetRemoteDataResponse:
         :rtype: ``list``
         """
         if self.mirrored_object:
-            demisto.info('Updating object {}'.format(self.mirrored_object["id"]))
             return [self.mirrored_object] + self.entries
 
 
--- a/Packs/Base/ReleaseNotes/1_7_14.md
+++ b/Packs/Base/ReleaseNotes/1_7_14.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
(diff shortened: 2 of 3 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -5891,7 +5891,6 @@ def extract_for_local(self):
         :rtype: ``list``
         """
         if self.mirrored_object:
-            demisto.info('Updating object {}'.format(self.mirrored_object["id"]))
```
ShahafBenYakir, line 5894: Do we have any other "constant" field (maybe title/name) we can use for logging? Seems useful to have this printed to logs

Packs/Base/ReleaseNotes/1_7_11.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Maintenance and stability enhancements.
```
ShahafBenYakir, line 4: Let's give better notes here

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
guyfreund: No, as each object being mirrored is different, depends on the product.

ShahafBenYakir (approved):

PR #11349, opened 2021-02-09
Workday IAM fetch fix

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Description
Replaced incidents' names from givenname + surname to email address. This should fix an issue where incidents were created with the name "None None" because the givenname + surname were not available in the report. The email is required (and unique per user) so it won't happen in this case.


Shellyber (approved):

Packs/Workday/ReleaseNotes/1_0_9.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Workday IAM
+- Maintenance and stability enhancements.
```
Shellyber, line 4: Maybe better to write an explanation about the bug fix?

PR #11353, opened 2021-02-09
fixing the DT

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33358

## Description
Fixed an issue with DT that cause some commands to override the previous context.

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
   - [x] No


Packs/EWS/Integrations/EWSO365/EWSO365.py
```diff
@@ -86,11 +86,13 @@
 MAILBOX = "mailbox"
 MAILBOX_ID = "mailboxId"
 FOLDER_ID = "id"
+TARGET_MAILBOX = 'receivedBy'
 
 # context paths
-CONTEXT_UPDATE_EWS_ITEM = "EWS.Items(val.{0} === obj.{0} || (val.{1} && obj.{1} && val.{1} === obj.{1}))".format(
-    ITEM_ID, MESSAGE_ID
-)
+CONTEXT_UPDATE_EWS_ITEM = f"EWS.Items((val.{ITEM_ID} === obj.{ITEM_ID} || " \
```
yaakovi, line 92: why do we use `or` between the item ID and the message ID conditions? can the item ID be missing?

yaakovi (commented):
were you able to reproduce?
how did you test?

Packs/EWS/Integrations/EWSO365/EWSO365.py
esharf: not sure why but I didn’t want to remove it.

esharf:
@yaakovi I didn’t try it but it sounds like the solution so I sent it to the customer testing.

esharf:
@yaakovi the provide solution is working for me, i think we can merge it.

yaakovi (approved):

PR #11360, opened 2021-02-10
Slack v2 Improved argument description

reutshal:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33477

## Description
- Improved the 'message' argument description to prevent further issues.
- Added the raw response to the 'Contents' entry in the send-notification command.
- 
   - [ ] Yes
       - Further details:
   - [X] No

### DIFF
--- a/Packs/Slack/Integrations/Slack/Slack.yml
+++ b/Packs/Slack/Integrations/Slack/Slack.yml
@@ -223,7 +223,7 @@ script:
     name: mirror-investigation
   - arguments:
     - default: true
-      description: The message content.
+      description: "The message content. When mentioning another slack user, make sure to do so in the following format: <@user_name>."
       isArray: false
       name: message
       required: false
--- a/Packs/Slack/Integrations/Slack/README.md
+++ b/Packs/Slack/Integrations/Slack/README.md
@@ -162,7 +162,7 @@ Investigation mirrored successfully, channel: my-mirror
   <tbody>
     <tr>
       <td>message</td>
-      <td>The message content.</td>
+      <td>The message content. When mentioning another slack user, make sure to do so in the following format: <@user_name>.</td>
       <td>Optional</td>
     </tr>
     <tr>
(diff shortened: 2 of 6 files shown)


reutshal:
@roysagi @amshamah419 whoever takes this can please take a quick look at the description I updated? Thanks 🙏

Packs/Slack/Integrations/Slack/Slack.yml
```diff
@@ -223,7 +223,7 @@ script:
     name: mirror-investigation
   - arguments:
     - default: true
-      description: The message content.
+      description: "The message content. When mentioning another slack user, make sure to do so in the following format: <@user_name>."
```
liorblob, line 226: Not sure about putting this in the argument description, the integration doc might be more appropriate. 
Also, this can also apply for tagging channels, e.g. #general etc

reutshal: I can add it there in addition, but when a customer is using the command I think that the argument description is more relevant than the readme, I'm afraid the customer will miss it if it is present only in the readme

liorblob: yeah in any case it should be added to the doc. and don't forget to add an example for a channel too :)

roysagi:
@reutshal - Please ping me when you are done here and I can check the docs.

reutshal:
@roysagi Done, added the description to the README also

liorblob (approved):

yaakovi (approved):

jplonie:
This update is breaking slack send-notification for me. It eventually returns with 

Error from SlackV2 is : Script failed to run: 
Timeout Error: Docker code script failed due to timeout, consider changing timeout value for this automation, 
Error: Failed to decode (loop) data from docker code script: unexpected EOF
This may happen when you don't return an explicit entry (or entries) from a script.
for entry structure please review the 'Script Helper' (2605) (2604) (2603)

PR #11369, opened 2021-02-10
Adding search rank to pack index metadata

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31107

## Description
Generating rank search for packs and adding it to the metadata in the index.
Added to the upload flow.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests


ohaim1008:
You can see the test bucket build and index links in the issue.

Tests/Marketplace/marketplace_services.py
```diff
@@ -555,6 +555,50 @@ def _get_certification(support_type, certification=None):
...
+    @staticmethod
+    def _get_search_rank(tags, certification, content_items):
+        """ Returns pack search rank.
+
+        The initial value is 0
+        In case the pack has the tag Featured, its search rank will increase by 10
+        In case the pack was released in the last 30 days, its search rank will increase by 10
+        In case the pack is certified, its search rank will increase by 10
+        In case all the pack's integration are deprecated and there is at least 1 integration in the pack,
+        the pack's search rank will decrease by 10
```
guyfreund, line 567:
```suggestion
        the pack's search rank will decrease by 50
```

guyfreund (changes requested):
Looking good, Few notes:
- Please add a UT for the case of a deprecated pack
- Just to be sure, `searchRank` is calculated from scratch each time? it's not changing dynamically over time?
- What packs have been tested? (Can't see this from the attached run/index)

Tests/Marketplace/marketplace_services.py
ohaim1008: * Sure
* Yes, currently that's the logic, might change in the not so long future, furthermore, it depends on the `New` tag that is checked and added in each upload so it make sense for now.
* All, you can see that the field was added to all the packs metadata files.

ohaim1008:
Added the changes, I'll run another test on the test bucket before merging.

Tests/Marketplace/Tests/marketplace_services_test.py
```diff
@@ -168,6 +171,43 @@ def test_new_tag_removed(self, dummy_pack_metadata, dummy_pack):
                                                           is_feed_pack=False)
 
         assert parsed_metadata['tags'] == ["tag number one", "Tag number two", 'Use Case']
+        assert parsed_metadata['searchRank'] == 10
+
+    def test_deprecated_pack_search_rank(self, dummy_pack_metadata, dummy_pack):
+        """
+        Given a deprecated pack
+        Then: calculate the search rank
```
bakatzir, line 179:
```suggestion
        Given: a  pack
        When: All the integrations in it are deprecated.
        Then: calculate the search rank
```

Tests/Marketplace/Tests/marketplace_services_test.py
```diff
@@ -168,6 +171,43 @@ def test_new_tag_removed(self, dummy_pack_metadata, dummy_pack):
...
+        }
+        parsed_metadata = dummy_pack._parse_pack_metadata(user_metadata=dummy_pack_metadata,
+                                                          pack_content_items=content_items,
+                                                          pack_id='test_pack_id', integration_images=[],
+                                                          author_image="", dependencies_data={},
+                                                          server_min_version="5.5.0", build_number="dummy_build_number",
+                                                          commit_hash="dummy_commit", downloads_count=10,
+                                                          is_feed_pack=False)
+
+        assert parsed_metadata['searchRank'] == -40
```
bakatzir, line 210: why -40? is the `pack` new?

bakatzir (commented):
Let's add a test case for the most common scenario of adding a new integration to a pack(v2) and deprecating the v1 integration) e.g: writing SNOWv2 adding it to SNOW pack add deprecating SNOW(v1).
score should be 10 If I am not mistaken.

Tests/Marketplace/Tests/marketplace_services_test.py
ohaim1008: it's a certified one

ohaim1008:
New integration in an existed pack won't change the searchRank score, only new packs are getting +10, I will change the documentation to clarify the tests.

ohaim1008:
Tested again - no change from the last one. PR is ready.

bakatzir (commented):
Looks good from my side. @yuvalbenshalom Please review as well.

bakatzir (approved):

guyfreund (approved):

PR #11376, opened 2021-02-10
Ignoring docker image validation in 5.5 and 5.0

hod-alpert:
Avoid validation of docker images during content installation.


Itay4 (approved):

glicht:
@hod-alpert Looks good. We may also get problems with 6.0+ versions. But let's wait and see. If so, we can do this for all versions and only enable on nightly.

PR #11417, opened 2021-02-14
FortiSIEM login error debug

bakatzir:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/33750

## Description
more detailed login error

   - [x] No


Packs/FortiSIEM/Integrations/FortiSIEM/FortiSIEM.py
```diff
@@ -76,9 +76,12 @@ def login():
     response = session.get(login_url, verify=VERIFY_SSL)
 
     # get the VIEW_STATE from the xml returned in the UI login page.
+    resp_text = response.text.encode('utf-8')
```
yaakovi, line 79: I think now is a good opportunity to mock the response for future uses.
make sure it's secret-free before posting it.

yaakovi: also, if the login page was changed - this will have to be re-written as this request is not part of the REST API.

PR #11446, opened 2021-02-16
git related changes to be entered upon sdk 1.3.0 release

roysagi:
git related changes to be entered upon sdk 1.3.0 release


yaakovi (commented):
looks good.
it's pending SDK release so not approving yet

PR #10922, opened 2021-01-17
EDL: Optimize collapse IPs option

yaakovi:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/32832

## Description
optimized the code using the `netaddr` package (already used by the code)

PR #11449, opened 2021-02-16
ExportIndicators: filtered null indicators

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33616

## Description
* filtered null indicators: [here](https://github.com/demisto/content/pull/11449/files#diff-13f432b912da701f9003e938395ad349623c6deef2329fb2f5a4d2e559347b76R352-R353)
* applied updates from EDL improve performance: https://github.com/demisto/content/pull/10922


bakatzir (approved):
were you able to test the changes with the customer? If we do not have UT that reproduces, I advise doing that prior to merging.
https://github.com/demisto/content/pull/10922#pullrequestreview-570078032

yaakovi:
> were you able to test the changes with the customer? If we do not have UT that reproduces, I advise doing that prior to merging.
> [#10922 (review)](https://github.com/demisto/content/pull/10922#pullrequestreview-570078032)

the code optimization was tested as part of the EDL effort (using an environment with many indicators)
the second is pretty straight forward and also copied from EDL

PR #11460, opened 2021-02-16
Add logs

ShahafBenYakir:
Related to this: https://github.com/demisto/content/pull/11456/checks?check_run_id=1911871900


avidan-H (changes requested):
@ShahafBenYakir I think we should move this script to `Utils/github_workflow_scripts` and refactor it to use the `pygithub` package. It would trim down the script, and unify the behavior with our other github `actions` scripts. If you just want to add logs for now, we should add `response.text` to the output wherever else the script exits on a bad response as well, not only the place where it happened this time.

ShahafBenYakir:
@avidan-H Let's open an issue for the refactoring

avidan-H (approved):

PR #11488, opened 2021-02-18
remock playbooks with oproxy test

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
I have added a new server configuration for `NO_PROXY` env so we can use now direct access to oproxy even if we run with proxy (e.g. `HTTP_PROXY` set or something like this).
we still need to decide from where we should take the list of hosts but this is a POC.
there is a part of oproxy dependents that still not working with proxy but i think that this is related to https://github.com/demisto/etc/issues/33496 and will be solved after the sdk release.


Tests/configure_and_test_integration_instances.py
```diff
@@ -37,11 +37,18 @@
 
 MARKET_PLACE_MACHINES = ('master',)
 SKIPPED_PACKS = ['NonSupported', 'ApiModules']
+NO_PROXY = ','.join([
+    'oproxy.demisto.ninja',
+    'oproxy-dev.demisto.ninja',
+    # 'login.microsoftonline.com'
```
yaakovi, line 43: remove if not relevant

Tests/configure_and_test_integration_instances.py
```diff
@@ -37,11 +37,18 @@
...
+])
+NO_PROXY_FOR_5_0_0 = {'python.pass.extra.keys': f'--env##no_proxy={NO_PROXY}'}  # noqa: E501
```
yaakovi, line 45: change name if relevant for 5.5 too

Tests/configure_and_test_integration_instances.py
```diff
@@ -37,11 +37,18 @@
...
 DOCKER_HARDENING_CONFIGURATION = {
     'docker.cpu.limit': '1.0',
     'docker.run.internal.asuser': 'true',
     'limit.docker.cpu': 'true',
-    'python.pass.extra.keys': '--memory=1g##--memory-swap=-1##--pids-limit=256##--ulimit=nofile=1024:8192'
+    'python.pass.extra.keys': f'--memory=1g##--memory-swap=-1##--pids-limit=256##--ulimit=nofile=1024:8192##--env##no_proxy={NO_PROXY}',    # noqa: E501
+    'powershell.pass.extra.keys': f'--env##no_proxy={NO_PROXY}'
```
yaakovi, line 51:
```suggestion
    'powershell.pass.extra.keys': f'--env##no_proxy={NO_PROXY}',
```

yaakovi (approved):
nice work

PR #11493, opened 2021-02-18
EWS-365 - Fixed an issue of access denied error

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33906

## Description
Fixed an issue where `source_mailbox` was not set to the default of the argument `traget_mailbox`

   - [ ] Yes
       - Further details:
   - [x] No


DeanArbel (approved):
Make sure this works for the customer.
Perhaps it'd be a good idea to explain via a comment that the client's default_target_mailbox` is the `authorization` source for the instance (otherwise it's hard to understand the change).

abaumgarten:
customer confirmed it works.

Packs/EWS/ReleaseNotes/1_7_9.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### EWS O365
+- Fixed an issue where the command **ews-move-item-between-mailboxes** failed due to access denied error.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where the ***ews-move-item-between-mailboxes*** command would fail due to an access denied error.
```

PR #11505, opened 2021-02-21
remock playbooks with streaming files

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#33496


yaakovi (approved):
monitor those integrations in the next few days and see if they run with the mocks or recording each time.

esharf:
I will handle the nightly tests in https://github.com/demisto/etc/issues/34079

PR #11514, opened 2021-02-21
CheckDockerImageAvailableTest - modify test to work against xsoar-registry.pan.dev

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30117

## Description
modify the test to work against xsoar-registry.pan.dev due to docker rate limit


glicht (approved):
Not a big fan of this change, but I guess we have this covered from the unit test side (for now).

PR #11550, opened 2021-02-23
TAXII 2 add incrementalFeed parameter

DeanArbel:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/24029

## Description
Added an `incrementalFeed` parameter to TAXII 2, since it's incremental feed.

   - [x] No

## Screenshots
image
image

### DIFF
--- a/Packs/FeedTAXII/Integrations/FeedTAXII2/FeedTAXII2.py
+++ b/Packs/FeedTAXII/Integrations/FeedTAXII2/FeedTAXII2.py
@@ -227,7 +227,7 @@ def main():
 
     initial_interval = params.get("initial_interval")
     fetch_full_feed = params.get("fetch_full_feed") or False
-    is_incremental_feed = params.get('feedIncremental')
+    is_incremental_feed = params.get('feedIncremental') or False
     limit = try_parse_integer(params.get("limit") or -1)
     limit_per_request = try_parse_integer(params.get("limit_per_request"))
 
--- a/Packs/FeedTAXII/Integrations/FeedTAXII2/FeedTAXII2.py
+++ b/Packs/FeedTAXII/Integrations/FeedTAXII2/FeedTAXII2.py
@@ -29,6 +29,12 @@ def try_parse_integer(
     return res
 
 
+def assert_incremental_feed_params(fetch_full_feed, is_incremental_feed):
+    if fetch_full_feed == is_incremental_feed:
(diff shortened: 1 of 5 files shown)


teizenman (approved):

glicht:
@DeanArbel I think this param needs to be configurable and go together with `fetch_full_feed`. So when `fetch_full_feed` is true incremental should be set to false.

Packs/FeedTAXII/ReleaseNotes/1_0_8.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### TAXII 2 Feed
+- Maintenance and stability enhancements.
```
glicht, line 4: We should change this. My suggestion:

```suggestion
- Added the *Incremental Feed* configuration parameter. This param indicates that the feed pulls only indicators that are new or modified. Set this param to true when *Full Feed Fetch* is disabled.
```

Packs/FeedTAXII/Integrations/FeedTAXII2/FeedTAXII2.py
```diff
@@ -221,12 +227,16 @@ def main():
 
     initial_interval = params.get("initial_interval")
     fetch_full_feed = params.get("fetch_full_feed") or False
+    is_incremental_feed = params.get('feedIncremental')
```
glicht, line 230: There is a case that in update, this will return None and then the assert will fail. See: https://github.com/demisto/etc/issues/30318

```suggestion
    is_incremental_feed = params.get('feedIncremental') or False
```

glicht (approved):
See my comments

Packs/FeedTAXII/ReleaseNotes/1_0_8.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### TAXII 2 Feed
+- Added the *Incremental Feed* configuration parameter. This param indicates that the feed pulls only indicators that are new or modified. Set this param to true when *Full Feed Fetch* is disabled.
```
yaakovi, line 4:
```suggestion
- Added the *Incremental Feed* configuration parameter. This parameter indicates that the feed pulls only indicators that are new or modified. Set this parameter to true when *Full Feed Fetch* is disabled.
```

yaakovi (approved):

PR #11553, opened 2021-02-24
Update conf.json

esharf:


Tests/conf.json
```diff
@@ -3446,7 +3446,7 @@
         "Mail Listener v2": "Integration has no proxy checkbox",
         "Cortex XDR - IOC": "'Cortex XDR - IOC - Test' is using also the fetch indicators which is not working in proxy mode",
         "SecurityAndCompliance": "Integration doesn't support proxy",
-        "Cherwell": "Submits a file - tests that send files shouldn't be mocked",
+        "Cherwell": "Submits a file - tests that send files shouldn't be mocked. this problem was fixed but the test is not running any more because the integration is skkiped",
```
yaakovi, line 3449:
```suggestion
        "Cherwell": "Submits a file - tests that send files shouldn't be mocked. this problem was fixed but the test is not running anymore because the integration is skipped",
```

Tests/conf.json
```diff
@@ -3436,7 +3436,7 @@
         "urlscan.io": "Uses data that comes in the headers",
         "CloudConvert": "has a command that uploads a file (!cloudconvert-upload)",
         "Symantec Messaging Gateway": "Test playbook uses a random string",
-        "Cylance Protect": "Test playbook (get_file_sample_by_hash_-_cylance_protect_-_test) downloads a file",
+        "Cylance Protect": "Test playbook (get_file_sample_by_hash_-_cylance_protect_-_test) downloads a file. this problem was fixed but the test is not running any more because the integration is deprecated",
```
yaakovi, line 3439: in that case, remove it from the list (no reason to specify something that is not running)

yaakovi, line 3439: in that case, remove it from the list (no reason to specify something that is not running)
```suggestion
```

yaakovi (approved):

yaakovi:
remove both `Cylance Protect` tests from the test list

PR #11565, opened 2021-02-24
Marketplace: Remove Contrib Suffix

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33761

## Description
Added a function to remove the contribution details from the integration display name.
**This modification takes place *after* the upload step and affects only the integration top-level entry in the parsed metadata file.**

## Screenshots
After changes:

Screen Shot 2021-02-24 at 17 31 49

### DIFF
--- a/Tests/Marketplace/marketplace_services.py
+++ b/Tests/Marketplace/marketplace_services.py
@@ -1811,7 +1811,7 @@ class Pack(object):
             return task_status, exists_in_index
 
     @staticmethod
-    def remove_contrib_suffix_from_name(display_name: str):
+    def remove_contrib_suffix_from_name(display_name: str) -> str:
         """ Removes the contribution details suffix from the integration's display name
         Args:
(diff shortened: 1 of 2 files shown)


Tests/Marketplace/marketplace_services.py
```diff
@@ -1810,6 +1810,24 @@ def check_if_exists_in_index(self, index_folder_path):
         finally:
             return task_status, exists_in_index
 
+    @staticmethod
+    def remove_contrib_suffix_from_name(display_name: str):
```
Itay4, line 1814:
```suggestion
    def remove_contrib_suffix_from_name(display_name: str) -> str:
```

yaakovi (approved):
this will require re-uploading of all relevant packs so the metadata files will be updated.

PR #11571, opened 2021-02-24
Active Directory Fixed an issue in iam-create-user

Shellyber:
## Status
- [ ] In Progress
- [x] Ready

## Description
Removed duplicate `ou` field that was passed to `iam-create-user` command.

## Minimum version of Demisto
- [x] 6.0.0

   - [x] No


Packs/Active_Directory_Query/ReleaseNotes/1_1_6.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Active Directory Query v2
+- Fixed an issue where ``iam-create-user`` command failed because of a duplicate field.
```
yaakovi, line 4:
```suggestion
- Fixed an issue where the ***iam-create-user*** command failed because of a duplicate field.
```

Packs/Active_Directory_Query/Integrations/Active_Directory_Query/Active_Directory_Query.py
```diff
@@ -773,6 +773,7 @@ def create_user_iam(default_base_dn, args, mapper_out, disabled_users_group_cn):
         else:
             user_dn = generate_dn_and_remove_from_user_profile(ad_user)
             object_classes = ["top", "person", "organizationalPerson", "user"]
+            ad_user.pop('ou')
```
yaakovi, line 776: how can you be sure there is more than one?
what happens if there is only one/none?
```suggestion
            ad_user.pop('ou', None)
```

yaakovi (commented):
please add a unit test simulating this scenario

Packs/Active_Directory_Query/Integrations/Active_Directory_Query/Active_Directory_Query.py
Shellyber: So In `ad_user` we already have the `ou` field which is populated to the `user_dn`. So we removed it from the `ad_user` when creating the user in AD.
There will always be one 'ou'. Its mandatory and exists in the mapper. it will fail the creation if there is None

Shellyber:
@yaakovi There is a problem with creating a test for this. There are no unittests and no mocks for the `Connection` class and the `conn.add` functionality. The code is written in a way that it's very hard to mock things.
As I explained before, the `ou` will never be empty in this case, since it has been used in the mapping step. 
If the mapper is wrong, the IAM playbook will fail with that reason

yaakovi (approved):

PR #11606, opened 2021-03-01
Add vendor id private repo

JasBeilin:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/27617

## Description
changed vendor id to partner id and added vendor id

   - [x] No


guykeller (approved):
Looks good IMO, @guyfreund can you please verify we're not breaking anything for you?

guyfreund (changes requested):
Looking good 👏 
@JasBeilin please attach a successful test-upload run and a screenshot of the public `index.json`

guykeller:
@guyfreund The test upload flow can only be run after we merge to private.
Would you like us to merge it with the branch ref, so the changes will be updated in `marketplace-dist-private` and then show such a run?

guyfreund (approved):

PR #11651, opened 2021-03-04
SplunkPy Basic Auth

amshamah419:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34169

## Description
Authentication fails behind a load balancer since the connection is authenticated for specific servers. Added the basic=true flag to ensure auth with every request.

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -1086,7 +1086,8 @@ def main():
         'app': demisto.params().get('app', '-'),
         'username': demisto.params()['authentication']['identifier'],
         'password': demisto.params()['authentication']['password'],
-        'verify': VERIFY_CERTIFICATE
+        'verify': VERIFY_CERTIFICATE,
+        'basic': True
     }
 
     if use_requests_handler:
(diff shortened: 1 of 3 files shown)


glicht (changes requested):
Need to take also care of: `splunk_edit_notable_event_command` which does direct http calls.

ShahafBenYakir:
@guyfreund FYI

amshamah419:
@glicht If this works for the customer we can implement it. This is a known issue for Splunk cloud users due to the load balancer not passing the auth session to the different servers. This is more or less a patch solution until something official from Splunk comes out. If this doesn't work, it would require a redesign of the auth session and we _really_ would need to evaluate if we want to do that.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -1086,7 +1086,8 @@ def main():
         'app': demisto.params().get('app', '-'),
         'username': demisto.params()['authentication']['identifier'],
         'password': demisto.params()['authentication']['password'],
-        'verify': VERIFY_CERTIFICATE
+        'verify': VERIFY_CERTIFICATE,
+        'basic': True
```
guyfreund, line 1090: If you set it to basic at all cases we wouldn't be able to use `auth token`:



In what cases do we need a basic auth?

glicht:
@amshamah419 I think we can close this one. Seems that auth token fixed the issue.

PR #11653, opened 2021-03-04
EWS - Security&Compliance - Search Action Limit Fix

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34416

## Description
Fixed an issue where using the `limit` parameter with value -1 in the `O365-sc-get-search-action` command resulted in an error.

   - [ ] Yes
       - Further details:
   - [x] No


darkushin:
@GalRabin @yaakovi As can be seen in the screenshot, when the user asks to export the results file to the war room, we create a file containing all the results even if the user has set a limit. Is this on purpose or should I change it to the limit set by the user?
Same applies also to the `GetSearchActionCommand` function.
image

yuvalbenshalom (commented):
looks ok for me
@yaakovi please take a look too

GalRabin:
> @GalRabin @yaakovi As can be seen in the screenshot, when the user asks to export the results file to the war room, we create a file containing all the results even if the user has set a limit. Is this on purpose or should I change it to the limit set by the user?
> Same applies also to the `GetSearchActionCommand` function.
> image

Its not in purpose its a bug , You right.
Thanks!

GalRabin (commented):
Its not in purpose its a bug , You right.
Thanks!

yaakovi (approved):
validate is failing on an SDK issue, @yuvalbenshalom force-merge when ready.

PR #11677, opened 2021-03-07
Jira V2 - fixed command jira-get-id-by-attribute to return only active users

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34485
fixes: https://github.com/demisto/etc/issues/34676
related: https://github.com/demisto/dockerfiles/pull/4434

## Description
- Fixed an issue where the command **jira-get-id-by-attribute** returned a list of active and inactive users. Now, the new endpoint returns a list of only active users.
The customer's issue probably caused by a long list of inactive users and then the specific active user was not found. 
Updated TBP to use UserName instead of Email address because when using Auth1 email address returns no results.

the new endpoint docs: 
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-user-search/#api-rest-api-3-user-picker-get

## Minimum version of Demisto
- [x] 5.0.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Jira/Integrations/JiraV2/JiraV2.py
```diff
@@ -197,14 +199,14 @@ def search_user(query: str, max_results: str = '50'):
...
-    url = f"rest/api/latest/users/search?query={query}&maxResults={max_results}"
+    url = f"rest/api/latest/user/search?query={query}&maxResults={max_results}"
     res = jira_req('GET', url, resp_type='json')
     return res
 
 
 def get_account_id_from_attribute(attribute: str, max_results: str = '50') -> Union[CommandResults, str]:
     """
-    https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-user-search/#api-rest-api-3-user-search-get
+    https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-user-search/#api-rest-api-3-user-picker-get
```
Bargenish, line 209: no need to change. you're still using the user/search endpoint and not the picker one.

Bargenish (approved):

Packs/Jira/ReleaseNotes/1_2_13.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Atlassian Jira v2
+- Fixed an issue where the command **jira-get-id-by-attribute** returned a list of active and inactive users. Now, the new endpoint returns a list of only active users.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where the **jira-get-id-by-attribute** command would return a list of active and inactive users. Now, the new endpoint returns only a list of active users.
```

yuvalbenshalom (changes requested):
@abaumgarten see my comment here https://github.com/demisto/etc/issues/34485#issuecomment-792263955

I do not understand what is the difference between the endpoints

abaumgarten:
> @abaumgarten see my comment here [demisto/etc#34485 (comment)](https://github.com/demisto/etc/issues/34485#issuecomment-792263955)
> 
> I do not understand what is the difference between the endpoints

The old one:
image

The new one:
image

Please see my comment in the issue

yuvalbenshalom (approved):

PR #11679, opened 2021-03-07
Microsoft Graph Mail - get message as attachment

tallieber:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34339

## Description
get-attachment command returned only file types, added item type (message)
## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation


Packs/MicrosoftGraphMail/Integrations/MicrosoftGraphMail/MicrosoftGraphMail.py
```diff
@@ -1150,14 +1152,46 @@ def delete_mail_command(client: MsGraphClient, args):
...
+
+def create_attachment(raw_response, user_id):
+    attachment_type = raw_response.get('@odata.type', '')
+    # Documentation about the different attachment types
+    # https://docs.microsoft.com/en-us/graph/api/attachment-get?view=graph-rest-1.0&tabs=http
+    if 'itemAttachment' in attachment_type:
+        return_results(item_result_creator(raw_response, user_id))
+    if 'fileAttachment' in attachment_type:
+        return demisto.results(file_result_creator(raw_response))
+    return {}
```
yuvalbenshalom, line 1185: please add a debug log here that there is an unsupported attachment type and it was not added to the incident. and print the type in the log.

yuvalbenshalom (changes requested):
looks good. please add a tests for this case and checkout the build failures.
see my comment.

Packs/MicrosoftGraphMail/Integrations/MicrosoftGraphMail/MicrosoftGraphMail.py
```diff
@@ -1150,14 +1152,48 @@ def delete_mail_command(client: MsGraphClient, args):
...
+    if 'message' in item_type:
+        message_id = raw_response.get('id')
+        item['id'] = message_id
+        mail_context = build_mail_object(item, user_id=user_id, get_body=True)
+        human_readable = tableToMarkdown(
+            f'Attachment ID {message_id} \n **message details:**',
+            mail_context,
+            headers=['ID', 'Subject', 'SendTime', 'Sender', 'From', 'HasAttachments', 'Body']
+        )
+        return CommandResults(outputs_prefix='MSGraphMail',
```
yuvalbenshalom, line 1167: I second glance, note that you should not call CommandResults from the internal function, but rather from the command function, to make this more general business logic function.
also, it is better to create the human readable out there if possible.

yuvalbenshalom (changes requested):
for completion please take a look at `_get_email_attachments` that is being used in fetch incidents. I just took a look and I am not sure if this flow should be merged or is it a completely diffrent flow.

see my comment.

Packs/MicrosoftGraphMail/Integrations/MicrosoftGraphMail/MicrosoftGraphMail.py
tallieber: It is not completely different, it try to do the same but returning eml file type. From my testing it does the same in fetch incident and in get-attachment, maybe we want to get the option to the customer to get attachment by eml type

tallieber:
@yuvalbenshalom @bakatzir Yuval raised an important point about the fetch-incident command that also execute get-attchment for each message he fetches. I tested it to find the difference between the get-attachment executed from fetch incidents and the get-attachment command that I added him the ability to return message type attachments:

1. File type attachment- there is no difference, both return war room entry with the file
2. Item type attachment- in both cases we refer only to message type attachment : 
- **get-attachment executed from fetch incident** return eml file (the url suffix they use to get the details for the eml file is relevant only to message type)
- **get-attachment command** return message object (MSGraphMail) with context entry. I used url suffix that return details about all of the item types (contact and event)

I opened an enhancement issue to cope with event and contact attachment types, I think we also need to open an issue to create uniform behavior for this two use cases- get attachment from fetch-incident command and get-attachment-command. 
For now do you want me to merge it like this? Or to change it and return eml file? (Lets remember that the customer who opened the bug said this fix was good for him)

yuvalbenshalom (approved):

Packs/MicrosoftGraphMail/ReleaseNotes/1_0_17.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Microsoft Graph Mail
+- **msgraph-mail-get-attachment** command did not handle attachments from "message" type.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where the ***msgraph-mail-get-attachment*** command would not handle attachments that were from a `message` type.
```

PR #11720, opened 2021-03-09
Add 'controller' parameter support to MISPv2 integration searches

ag-michael:

## Status
- [] In Progress
- [x] Ready

## Description
This PR adds the 'controller' parameter in the search() api calls to PyMISP under the MISPv2 integration.

Right now, due to lack of this controller parameter XSOAR ends up performing all searches under the 'events' controller where searching for an attribute value results in entire events being pulled from MISP instead of the matching attribute. This has the side-effect of causing memory consumption issues when large events have one more more matching attributes resulting in hundreds of megabytes (or more) of events being pulled into XSOAR incidents. 

## Screenshots

## Minimum version of Demisto
- [ x] 5.0.0
- [ x] 5.5.0
- [ x] 6.0.0

   - [ x] Yes
       - Further details: Any work previously done by users that assumed entire events being pulled,parsed and processed in XSOAR might encounter issues, that said as far as I can tell this PR only results in reduction of response size from MISP.
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/MISP/Integrations/MISP_V2/MISP_V2.py
+++ b/Packs/MISP/Integrations/MISP_V2/MISP_V2.py
@@ -394,7 +394,7 @@ def check_file(file_hash):
         return_error('Invalid hash length, enter file hash of format MD5, SHA-1 or SHA-256')
 
     # misp_response will remain the raw output of misp
-    misp_response = MISP.search(value=file_hash)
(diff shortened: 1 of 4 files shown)


teizenman:
Since this is a breaking change to the current functionality, it will be considered to be added in another manner.

PR #11747, opened 2021-03-10
Protect landing-page sections file

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
We need to protect our landing-page sections file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,9 +1,9 @@
-# Marketplace related
+# Marketplace Related
 /Tests/Marketplace/approved_tags.json @demisto/content-leaders @michalgold
 /Tests/Marketplace/approved_usecases.json @demisto/content-leaders @michalgold
 /Tests/Marketplace/landingPage_sections.json @demisto/content-leaders @michalgold
 
-# Marketplace and Upload flow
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,6 +1,7 @@
 # Marketplace related
 /Tests/Marketplace/approved_tags.json @demisto/content-leaders @michalgold
 /Tests/Marketplace/approved_usecases.json @demisto/content-leaders @michalgold
+/Tests/Marketplace/landingPage_sections.json @demisto/content-leaders @michalgold
```
ShahafBenYakir, line 4: What is this file I don't see in the repo

ShahafBenYakir, line 4: @Bargenish Can we change the naming to stick with the current convention?

Bargenish: @ShahafBenYakir  it's a bit different case than the approved tags and usecases so I'm not sure it's right here. but we don't have any limitations to change it.

ShahafBenYakir: I meant the file name, it is landingPage_sections and should be landing_page_sections

PR #11762, opened 2021-03-11
Sample Incident Generator (Deprecated) - deprecate

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://panw-global.slack.com/archives/C011E5T5K17/p1615462034207600

   - [x] No


Itay4:
dont think this should be deprecated, lets discuss

ShahafBenYakir:
What was decided here?

bakatzir:
> What was decided here?

I think we can deprecate, see:
https://github.com/demisto/etc/issues/34673#issuecomment-796857457

bakatzir:
Decide not to deprecate for now, per the issue. closing.

PR #11767, opened 2021-03-11
Fireeye key fix

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34313

## Description
Was missing the submission key in the returned result

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/fireeye/Integrations/integration-fireeye.yml
```diff
@@ -381,7 +382,7 @@ script:
     sendRequest(commandDictionary.logout.url, commandDictionary.logout.method, token);
 
 
-    if (typeof result === 'object' && !result['submission_Key']) {
+    if (typeof result === 'object' && !result['submission_Key'] && command != 'fe-submit-status') {
```
bakatzir, line 385: the purpose here is to update the submisison key upon a submission, correct? if so, isn;t it better to:
```suggestion
    if (typeof result === 'object' && !result['submission_Key'] && (command == 'fe-submit' || command == 'fe-submit-url')) {
```

Packs/fireeye/Integrations/integration-fireeye.yml
```diff
@@ -362,6 +362,7 @@ script:
                 if (contentType && contentType.indexOf('application/xml') !== -1) {
                   status_result = JSON.parse(x2j(response.Body));
                 }
+                status_result['submission_Key'] = submissionKeyList[i];
```
bakatzir, line 365: need to handle the `fe-submit-url-status` also in these case, I believe it is lacking.

ShahafBenYakir: I don't think so. As I mentioned above, the fix here is due to a change in behavior I previously did to `fe-submit-status` **alone** (to handle multiple submissions at once). So unless we have a way to reproduce this same issue we had with `fe-submit-status` when trying to submit multiple ids in the `fe-submit-url-status` this is not needed. (due to the same if statement you commented above )

Packs/fireeye/Integrations/integration-fireeye.yml
ShahafBenYakir: The purpose here is to update the submission key wherever it missing. The integration worked just fine for all commands before the change I did for the fe-submit-status which broke this behavior for the command (since the returned result is a list instead of dict). Since I only changed (and broke) the `fe-submit-status` command this fix is only relevant for this command I am not sure on what other cases this statement is needed (I can imagine but I don't think it is smart to assume here)

bakatzir (approved):

ShahafBenYakir:
@roysagi @amshamah419 Can you have a quick look?

Packs/fireeye/ReleaseNotes/1_0_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### FireEye (AX Series)
+- Fixed an issue with the ***fe-submit-status*** command where the submission key was missing from the result.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where the submission key was missing from the result of the ***fe-submit-status*** command.
```

PR #11830, opened 2021-03-16
RF Feed - fixed fetch indicators json exception

abaumgarten:
## Status
- [x] In Progress
- [] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34776

## Description
Fixed an issue where **fetch indicators** failed due to missing EvidenceDetails value in the received indicator.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/FeedRecordedFuture/Integrations/FeedRecordedFuture/FeedRecordedFuture.py
```diff
@@ -344,12 +345,13 @@ def fetch_indicators_command(client, indicator_type, limit: Optional[int] = None
                     raw_json['score'] = score = client.calculate_indicator_score(risk)
                     raw_json['Criticality Label'] = calculate_recorded_future_criticality_label(risk)
                 lower_case_evidence_details_keys = []
-                evidence_details = json.loads(item.get('EvidenceDetails', '{}')).get('EvidenceDetails', [])
-                if evidence_details:
-                    raw_json['EvidenceDetails'] = evidence_details
-                    for rule in evidence_details:
-                        rule = dict((key.lower(), value) for key, value in rule.items())
-                        lower_case_evidence_details_keys.append(rule)
+                if item.get('EvidenceDetails', '{}'):
```
guykeller, line 348: `{}` is not False, are you sure this fix works? Can you add a build that failed before and passes now, with a UT for example?

abaumgarten: The issue is when we have for example item = {'EvidenceDetails' : None}
and then doing json.loads on None will raise an exception. 
I have added a UT for this case :) 
When I revert my fix the UT fails.

guykeller: Sounds good.
Just one comment - currently, we're getting evidence details twice.

Let's get it once, save it in a variable and use it on both line 348 and line 349

guykeller (approved):
See my only comment

Packs/FeedRecordedFuture/ReleaseNotes/1_0_8.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Recorded Future RiskList Feed
+- Fixed an issue where **fetch indicators** failed due to missing EvidenceDetails value in the received indicator.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where ***fetch-indicators*** would fail when the received indicator was missing the *EvidenceDetails* value.
```

PR #11843, opened 2021-03-17
Humio/fix time range error on empty intitial fetch

AlexanderBrandborg:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: no issue created

## Description
The fetch_incidents function didn't handle the case of the initial fetch being empty. In this case the saved query start would be set to a value of 1, which would break further fetches. In this case we now default the query start to be the the timestamp of the last attempted fetch.

This has been tested manually.

## Screenshots
---

## Minimum version of Demisto
With this small change, I don't believe the minimum version has changed, but I can't seem to find info on what the minimum was previously given as.
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Humio/Integrations/Humio/Humio.py
```diff
@@ -307,7 +307,12 @@ def fetch_incidents(client, headers):
             ts = int(result.get("@timestamp", backup_ts))
             if ts > max_ts:
                 max_ts = ts
-        max_ts += 1
+
+        # Ensures that max_ts gets a reasonable value if no events were returned on first run
+        if(not response_data):
```
teizenman, line 312:
```suggestion
        if not response_data: 
```

teizenman (commented):
Please fix the lint issues and we can go forward with this fix

Packs/Humio/Integrations/Humio/Humio.py
AlexanderBrandborg: > Please fix the lint issues and we can go forward with this fix

New commit should fix linting issues

teizenman (approved):

PR #11859, opened 2021-03-18
MS Teams MGMT - add sleep after adding user to cloned team

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33410

## Description
in the test, after we add a member to the cloned team, we list the team members and then make that user the team owner
the issue was that the member was not listed in the team members because it could take some time for the member to be added, so added a sleep to wait a bit


ShahafBenYakir (commented):
Add tests

ShahafBenYakir (approved):

PR #11867, opened 2021-03-18
Adding validation for landingPage_sections.json file

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/34750

## Description
- Validating the file schema and that it contains valid json.
- Validating that all packs that receive a Tag via this file are valid (using latest index.zip file).
- Don't trigger private build when modifying the file.
- Do not run Server tests for 5.5 and 5.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -777,6 +777,29 @@ jobs:
                     exit 1
                   fi
       - run:
+          name: Validate landingPageSections.json
+          when: always
+          command: |
+            if [[ $CIRCLE_BRANCH =~ pull/[0-9]+ ]]
+              then
+                echo "Skipping - not running in contribution build"
+                exit 0
+            fi
+            GCS_PATH=$(mktemp)
+            echo "$GCS_MARKET_KEY" > "$GCS_PATH"
+            gcloud auth activate-service-account --key-file="$GCS_PATH" > auth.out 2>&1
+            echo "successfully activated google cloud service account"
+
+            # Download index.zip
+            INDEX_PATH=$(mktemp)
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -776,6 +776,29 @@ jobs:
...
+          when: always
+          command: |
+            if [[ $CIRCLE_BRANCH =~ pull/[0-9]+ ]]
+              then
+                echo "Skipping - not running in contribution build"
+                exit 0
+            fi
+            GCS_PATH=$(mktemp)
+            echo "$GCS_MARKET_KEY" > "$GCS_PATH"
+            gcloud auth activate-service-account --key-file="$GCS_PATH" > auth.out 2>&1
```
glicht, line 790: there is no need to authenticate to download the index. the index file is public and then you can also run this on a fork. I actually expect to receive PRs from forks (marketing).

.circleci/config.yml
```diff
@@ -776,6 +776,29 @@ jobs:
...
+
+            # Download index.zip
+            INDEX_PATH=$(mktemp)
+            gsutil cp gs://marketplace-dist/content/packs/index.zip $INDEX_PATH
+            echo "successfully downloaded index.zip into $INDEX_PATH"
+            gcloud auth revoke --all &> /dev/null
+            rm $GCS_PATH
+
+            UNZIP_PATH=$(mktemp -d)
+            unzip $INDEX_PATH -d $UNZIP_PATH
```
glicht, line 801: shouldn't you call validate_landing_page_sections.py  after the unzip?

glicht (approved):

PR #11887, opened 2021-03-21
CommandResult did not return dbot where no outputs_prefix or outputs_key_field supplied

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Description
When creating a CommandResult object with not outputs_prefix or outputs_keys_field, there was an overwrite of the `outputs` object instead of an update.
If the `outputs` passed in is a dict, it will now work as designed.
A list would over-write the outputs.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/ReleaseNotes/1_7_28.md
+++ b/Packs/Base/ReleaseNotes/1_7_28.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+Fixed an issue in CommandResult.to_context function that resulted no DBotScore in entry context when passing an IP indicator.
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -4371,7 +4371,11 @@ class CommandResults:
                 outputs_key = '{}'.format(self.outputs_prefix)
                 outputs[outputs_key] = self.outputs
             else:
-                outputs = self.outputs  # type: ignore[assignment]
+                if outputs and isinstance(self.outputs, dict):
+                        outputs.update(self.outputs)
+                else:
+                    outputs = self.outputs  # WILL OVERRIDE SCORES IF OUTPUTS IS LIST
+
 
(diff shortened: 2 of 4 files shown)


Packs/Base/ReleaseNotes/1_7_28.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+Fixed an issue in CommandResult.to_context function that resulted no DBotScore in entry context when passing an indicator.
```
amshamah419, line 4:
```suggestion
Fixed an issue in the **CommandResult.to_context** function that would result in no DBotScore being created in the entry context while passing an indicator.
```

Itay4:
looks ok to me
@anara123 do you mind having a look as well?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4371,7 +4371,11 @@ def to_context(self):
                 outputs_key = '{}'.format(self.outputs_prefix)
                 outputs[outputs_key] = self.outputs
             else:
```
anara123, line 4373: This code will also handle an edge case if outputs provided as non dict type (string, int, lists), and without outputs_prefix - because EntryContext must be dict
```suggestion
            elif not isinstance(self.outputs, dict) and not self.outputs_prefix:
                raise ValueError('outputs_prefix is missing')
            else:
```

anara123 (commented):
see my comment. and add unit test to my case

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
jochman: moved it to the `__init__` as it's a prerequisite. We might consider making a function that asserts all those changes.

anara123 (approved):

PR #11913, opened 2021-03-22
added Use Case tag to approved tags

merit-maita:
## Status
- [ ] In Progress
- [x] Ready

## Description
added Use Case to approved tags file.

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Tests/Marketplace/approved_tags.json
+++ b/Tests/Marketplace/approved_tags.json
@@ -36,6 +36,7 @@
 		"Courses of Action",
 		"Unit 42",
 		"Best Practices",
-		"New"
+		"New",
+		"Use Case"
 	]
 }


Tests/Marketplace/approved_tags.json
```diff
@@ -36,6 +36,7 @@
 		"Courses of Action",
 		"Unit 42",
 		"Best Practices",
-		"New"
+		"New",
```
glicht, line 39: I think you can remove New tag. It is auto added as needed and shouldn't be added to packs. @yaakovi FYI

yaakovi (approved):

PR #11923, opened 2021-03-22
Gmail no inbox failure

orhovy:
Added try and catch when trying to get mailboxes of an account that has no mailboxes
When one of the gmail users mailbox is disabled the command failed with an exception that `Mail service not enabled`.
This can be reproduced using this steps: https://support.google.com/a/answer/57919?hl=en

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34977

## Description

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -1125,7 +1126,13 @@ def search(user_id, subject='', _from='', to='', before='', after='', filename='
         'v1',
         ['https://www.googleapis.com/auth/gmail.readonly'],
         command_args['userId'])
-    result = service.users().messages().list(**command_args).execute()
+    try:
+        result = service.users().messages().list(**command_args).execute()
+    except HTTPError as e:
```
bakatzir, line 1131: I think that this way we would miss all of the mesasges.

bakatzir: I would suggest try getting the messages per user, and skip the users without a mail service.

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -1125,7 +1126,13 @@ def search(user_id, subject='', _from='', to='', before='', after='', filename='
...
+        if "Mail service not enabled" in str(e):
+            result = {}
+        else:
+            raise Exception(e)
```
yaakovi, line 1135:
```suggestion
            raise
```

yaakovi (commented):
looks good to me, were you able to test it?

Packs/Gmail/Integrations/Gmail/Gmail.py
yaakovi: this function searches in a specific mailbox, there is an iteration on all mailboxes in the calling function.

orhovy:
Tested it and saw the fix.

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -1125,7 +1125,13 @@ def search(user_id, subject='', _from='', to='', before='', after='', filename='
         'v1',
         ['https://www.googleapis.com/auth/gmail.readonly'],
         command_args['userId'])
-    result = service.users().messages().list(**command_args).execute()
+    try:
+        result = service.users().messages().list(**command_args).execute()
```
bakatzir, line 1129: let's log the disabled mailbox details(if we have them).

bakatzir (approved):

Packs/Gmail/Integrations/Gmail/Gmail.py
orhovy: We dont have it ... the api do not mention that on the response:
HttpError: <HttpError 400 when requesting https://gmail.googleapis.com/gmail/v1/users/gsuiteroot%40wealthsimple.com/messages?q=from%3Aykornyk%40gmail.com++subject%3ATEST&alt=json&maxResults=100&includeSpamTrash=true returned "Mail service not enabled". Details: "Mail service not enabled">.
As the command returns all mailboxes we do not which one returns an error.

PR #11927, opened 2021-03-22
Updating index and metadata of packs with changed tags

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34750

## Description
Added a method `get_packs_from_landing_page`, that parses the diff of the landing page sections file into modified packs.
These will be added to the packs that should be installed and in the `Prepare Content Packs For Testing` it's metadata will be updated with the newest and the packs will be installed on the server instances.

In addition -  modified the `get_modified_files_for_testing` method to return a class instead of tuple to make refactoring of that method a bit easier (no need to update all the places this method is called if there is a need to edit its output.

Most of this PR is about modifying the unittests to support this change.

### DIFF
--- a/Tests/scripts/collect_tests_and_content_packs.py
+++ b/Tests/scripts/collect_tests_and_content_packs.py
@@ -14,6 +14,7 @@ from distutils.version import LooseVersion
 from typing import Dict, Tuple, Union, Optional
 
 import demisto_sdk.commands.common.tools as tools
+from Tests.scripts.utils.collect_helpers import LANDING_PAGE_SECTIONS_JSON_PATH
 from demisto_sdk.commands.common.constants import *  # noqa: E402
 
 from Tests.scripts.utils import collect_helpers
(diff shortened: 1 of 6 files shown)


Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -793,6 +794,26 @@ def extract_matching_object_from_id_set(obj_id, obj_set, server_version='0'):
...
+    Parses the changes made by the current branch to the landing page sections file into modified packs.
+    These modified packs will have the metadata files changed.
+    Args:
+        branch_name: The name of the branch to compare changes from origin/master
+
+    Returns:
+        A set of packs that needs to be updated.
+    """
+    with open(LANDING_PAGE_SECTIONS_JSON_PATH, 'r') as file:
+        landing_page_sections_content = json.load(file)
```
Itay4, line 808: lets change this to something like `landing_page_sections_content = json.load(file).keys()` so we wont have to call `.keys()` in every loop iteration

Itay4 (approved):
looks good

ShahafBenYakir (approved):

PR #11938, opened 2021-03-24
Netskope - Fixed an issue where long Alert IDs where rounded down

bakatzir:
## Status
- [x] In Hold - customer verification

## Related Issues
fixes: https://github.com/demisto/etc/issues/34887

## Description
casting the long numbers in the json to string

   - [x] Yes - it casts long numbers.

## Must have
- [x] Tests -  no, no instance.


bakatzir:
@yaakovi @yuvalbenshalom Please see the suggested sol and the handling in the ref issue before I send this for verification to the customer.

Packs/Netskope/Integrations/integration-Netskope.yml
```diff
@@ -84,6 +84,11 @@ script:
         }
         var body;
         try {
+            // if fetching incidents, stringify long numbers to not round down long ids
```
yaakovi, line 87: add an example of what you are trying to catch in the regex

Packs/Netskope/Integrations/integration-Netskope.yml
```diff
@@ -1,8 +1,8 @@
 commonfields:
-  id: Netskope
+  id: Netskope_cast_long_integers
   version: -1
-name: Netskope
-display: Netskope
+name: Netskope_cast_long_integers
+display: Netskope_cast_long_integers
```
guykeller, line 5: please revert

yuvalbenshalom, line 5: fix naming

Packs/Netskope/Integrations/integration-Netskope.yml
```diff
@@ -84,6 +84,12 @@ script:
         }
         var body;
         try {
+            // if fetching incidents, stringify long numbers to not round down long ids, e.g:
+            // "dlp_incident_id":3747385551915191779 --> "dlp_incident_id":"3747385551915191779"
+            if (stringifyData && stringifyData.toLowerCase() == 'true') {
+                stringifyBody = result.Body.replace(/([\[:])?(\d{11,})([,\}\]])/g, "$1\"$2\"$3");
+                return JSON.parse(stringifyBody).data
```
guykeller, line 91: Why do you return the data, and not the body itself like in the other return statement?
Also, I suggest having only one return statement, unless you feel there is a reason for more than one

Packs/Netskope/Integrations/integration-Netskope.yml
```diff
@@ -84,6 +84,12 @@ script:
         }
         var body;
         try {
+            // if fetching incidents, stringify long numbers to not round down long ids, e.g:
+            // "dlp_incident_id":3747385551915191779 --> "dlp_incident_id":"3747385551915191779"
+            if (stringifyData && stringifyData.toLowerCase() == 'true') {
```
guykeller, line 89: Why is `stringifyData` not a bool?

yuvalbenshalom (changes requested):
look good just fix the integraton id/name

Packs/Netskope/Integrations/integration-Netskope.yml
bakatzir: this is the return for the fetch incident only which stringify the results to handle the long numbers.
I know it is not the prettiest, but given the fact that we have no instance and every change requires a validation, I'm ok with this.

guykeller (approved):

yuvalbenshalom (approved):

PR #11978, opened 2021-03-29
Update name of package for newer version

jonathan-s:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/RecordedFuture/Integrations/RecordedFuture/RecordedFuture.yml
+++ b/Packs/RecordedFuture/Integrations/RecordedFuture/RecordedFuture.yml
@@ -2,7 +2,7 @@ commonfields:
   id: Recorded Future v2
   version: -1
 name: Recorded Future v2
-display: Recorded Future v2
+display: Recorded Future v2.1
 category: Data Enrichment & Threat Intelligence
 image:
 description: Unique threat intel technology that automatically serves up relevant


teizenman:
@jonathan-s 
Why would you want to change the display name of the integration?
This can cause confusion among users. It is recommended to use the same name.

jonathan-s:
Hi @teizenman, 
That was a request I got from one of my colleagues at Recorded Future where I work to reflect that the version has in fact been upgraded, rather than just saying that this is version 2 of the plugin. 

I'm also noting that in the release notes it still says that it's version 1. Perhaps it would be better to reflect the change there. For context, we recently merged a PR where there were quite a lot of changes going in. So it would be good to let users know that there has been an update by signaling the version number change which is the intention behind this PR.

teizenman:
This is what the marketplace is for, users can see the version they are using, they will have to download and upgrade the current version in order to have it running.

This is what it will look like for user that did not yet upgrade to the newest version:
image

As you can see - they will get the notification to upgrade, and what has chagned.

Packs/RecordedFuture/Integrations/RecordedFuture/RecordedFuture.yml
```diff
@@ -2,7 +2,7 @@ commonfields:
   id: Recorded Future v2
   version: -1
 name: Recorded Future v2
-display: Recorded Future v2
+display: Recorded Future v2.1
```
ShahafBenYakir, line 5: @jonathan-s why have we decided to do this change? The `v2` refers to the difference between the previous (XSOAR supported) pack to this one. I think the addition `.1` could confuse customers with the real pack version (which shouldn't appear in the display name).

jonathan-s:
I've discussed with my colleague and he's agreed to leave it as is.

PR #11990, opened 2021-03-30
SearchIncidentsV2 - fixed encoding error

abaumgarten:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35244

## Description
Fixed an issue where the given incident id got from context data and was not got there after the stringify option.
The code checks if the type of the incident id is valid (int or str). In other cases, an indicative error returns.  

## Minimum version of Demisto
- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No


abaumgarten:
I have sent the script with new changes to the customer, waiting for his response. Locally my changes solved the issue.

anara123 (commented):
`id` arg is of type int (when it passed from the context). I would suggest to validate it differently. Instead of encoding and decoding, just add a check if it is a valid number/digit.

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2_test.py
```diff
@@ -41,3 +41,19 @@ def test_is_valid_args(_input, expected_output):
...
+
+    When:
+        - running the script as a playbook task
+
+    Then:
+        - validating that the incident is is a valid input
+
+    """
+    output = is_valid_args({'id': 123})
+    assert output is True
```
guykeller, line 59: Isn't this more readable and Pythonic IYO?
```suggestion
    is_valid_id = is_valid_args({'id': 123})
    assert is_valid_id
```

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
```diff
@@ -28,6 +28,11 @@ def is_valid_args(args: Dict):
             except UnicodeDecodeError as ex:
                 error_msg.append(f'Error while parsing the argument: "{_key}" '
                                  f'\nError:\n- "{str(ex)}"')
+            except TypeError:
+                try:
+                    _ = json.dumps(value)
```
guykeller, line 33: Not sure I understand why this is the fix.
Can you elaborate, please?

abaumgarten:
> `id` arg is of type int (when it passed from the context). I would suggest to validate it differently. Instead of encoding and decoding, just add a check if it is a valid number/digit.

I thought about doing it that way but then I realized that I don't know what all the other types can be (for example `'name', 'status', 'notstatus', 'reason', 'level', 'owner', 'type', 'query'` ), so I chose the safest way (with json dumps). If you think that the way you suggested is better I can change it of course.

guykeller (approved):

Packs/CommonScripts/ReleaseNotes/1_3_31.md
```diff
@@ -0,0 +1,5 @@
+
+#### Scripts
+##### SearchIncidentsV2
+- Fixed an issue where the value of an incident id from type int raised a TypeError exception.
```
amshamah419, line 4:
```suggestion
- Fixed an issue where the value of an incident ID was an integer and would raise a TypeError exception.
```

Packs/CommonScripts/ReleaseNotes/1_3_31.md
```diff
@@ -0,0 +1,5 @@
...
+- Updated the Docker image to demisto/python3:3.9.2.17957.
```
amshamah419, line 5:
```suggestion
- Updated the Docker image to: *demisto/python3:3.9.2.17957*.
```

PR #12027, opened 2021-04-01
CommonServerPython - add skip update to return error in get remote data

Itay4:
## Status
- [x] Ready

## Description
added the string `skip update` to the error message returned from `return_error` when running the `get-modified-remote-data` command

   - [x] No

## Must have
- [x] Tests


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4659,6 +4660,9 @@ def return_error(message, error='', outputs=None):
     if not isinstance(message, str):
         message = message.encode('utf8') if hasattr(message, 'encode') else str(message)
 
+    if is_command and demisto.command() == 'get-modified-remote-data' and not isinstance(error, NotImplementedError):
+        message += 'skip update'
```
yaakovi, line 4664: add a space (or a new line) before the word `skip`
in your test, the resulting message is `Test Errorskip update`

yaakovi (changes requested):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
Itay4: done

yaakovi (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4659,6 +4660,9 @@ def return_error(message, error='', outputs=None):
     if not isinstance(message, str):
         message = message.encode('utf8') if hasattr(message, 'encode') else str(message)
 
+    if is_command and demisto.command() == 'get-modified-remote-data' and not isinstance(error, NotImplementedError):
+        message += ' skip update'
```
glicht, line 4664: I would format it differently so it is clear that 'skip update' is not from the error message.
```suggestion
        message = 'skip update. error: ' + message
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4659,6 +4660,9 @@ def return_error(message, error='', outputs=None):
     if not isinstance(message, str):
         message = message.encode('utf8') if hasattr(message, 'encode') else str(message)
 
+    if is_command and demisto.command() == 'get-modified-remote-data' and not isinstance(error, NotImplementedError):
```
glicht, line 4663: `error` can be omitted (and usually is). In that case we should use sys.exc_info() to get the last exception and check that is in not `NotImplementedError`

glicht (changes requested):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
Itay4: done

glicht (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht: excellent

PR #12067, opened 2021-04-05
Symatnec DLP - Possible solution, more logs

teizenman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30286

## Description
Removed the condition for the last run values updates.
Added more logs.

## Minimum version of Demisto
- [X] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/SymantecDLP/Integrations/SymantecDLP/SymantecDLP.py
+++ b/Packs/SymantecDLP/Integrations/SymantecDLP/SymantecDLP.py
@@ -647,11 +647,15 @@ def fetch_incidents(client: Client, fetch_time: str, fetch_limit: int, last_run:
     :return: A list of Demisto incidents
     """
     # We use parse to get out time in datetime format and not iso, that's what Symantec DLP is expecting to get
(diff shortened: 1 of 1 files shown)


Packs/SymantecDLP/Integrations/SymantecDLP/SymantecDLP.py
```diff
@@ -692,9 +698,12 @@ def fetch_incidents(client: Client, fetch_time: str, fetch_limit: int, last_run:
                 incident['attachment'] = attachments
 
             incidents.append(incident)
-            if incident_id == incidents_ids[-1]:
-                last_incident_time = incident_creation_time
-                last_incident_id = incident_id
+
+            last_incident_time = incident_creation_time
```
glicht, line 702: It could be the incidents are not sorted. So I would save the new time only if it is larger that the last one saved.

PR #12070, opened 2021-04-05
skip packs linting due to lint bug

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Description
Skipping packs linting in upload-flow until SDK lint bug is fixed.
TODO: Revert this PR when the bug is fixed.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -402,9 +402,10 @@ references:
       when: always
       no_output_timeout: 5h
       command: |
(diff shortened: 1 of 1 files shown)


hod-alpert (changes requested):
A build has just passed this stage, let's wait before merging that

guyfreund:
> A build has just passed this stage, let's wait before merging that

@hod-alpert This implies that we need to force upload to production - which is not optimal, better to skip this for now and revert when we know the cause of the issue (as the issue is not deterministic)

guyfreund:
Closing for the moment... will re-open if the issue re-occurres.

PR #12080, opened 2021-04-05
Infinipoint - bump docker image

Itay4:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/33818

   - [x] No


bakatzir (commented):
If we do not have an actual instance to test on, we should contact the contrib, no?

Itay4:
> If we do not have an actual instance to test on, we should contact the contrib, no?

we do have

anara123 (approved):

PR #12200, opened 2021-04-13
Upload - store index.json in the bucket and ci artifacts

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26368
fixes: https://github.com/demisto/etc/issues/33617

## Description
store `index.json` in the bucket and in the CI artifacts


guyfreund (commented):
Looking good, but we also need to refactor the upload job (copy and upload script), after that we can trigger a testing upload flow

glicht (approved):

guyfreund (changes requested):
I also think we should upload the `index.json` to the root of the `content/Packs` and not to the index root as it will be compressed along with the Index.

Itay4:
triggered upload https://app.circleci.com/pipelines/github/demisto/content/79528/workflows/e5e2baa4-4c74-4626-8ce3-acfa5a480d08

guyfreund (approved):
Nice

PR #12205, opened 2021-04-13
Writing create-instances job

hod-alpert:
## Description:
This is the create instances job.
It contains pretty much the same flow as in circle with the following changes:
- The paths of the files that needs to be used in later jobs is changed to the artifacts folder.
- `GCS_ARTIFACTS_KEY` and `GCS_ARTIFACTS_KEY` env variables now contains the **path** to the files containing the service account and not the content of the json file, this simplifies things when authenticating to GCP.
- `CIRCLE_BUILD_NUM` will. be replaced by `CI_BUILD_ID`
- `CIRCLE_WORKFLOW_ID` will be replaced by `CI_PIPELINE_ID`
- `CIRCLE_BRANCH` will be replaced by `CI_COMMIT_BRANCH`
- The `$CIRCLE_NODE_INDEX` does not exist in gitlab which brings up the question of whether to maintain the circle upload path: `content/$CIRCLE_BRANCH/$CIRCLE_BUILD_NUM/$CIRCLE_NODE_INDEX` (where CIRCLE_NODE_INDEX = 0) or just remove it (which could be a bit simpler but break BC) @glicht @guyfreund 
- Since the build's image runs on python 3.9 which does not support Enum extension - a small fix for the sdk has been implemented as well [here](https://github.com/demisto/demisto-sdk/pull/1140).
## Related issues:
fixes: https://github.com/demisto/etc/issues/35417

## Build example:
https://code-stage.pan.run/xsoar/content/-/jobs/11263021

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -85,3 +85,80 @@ stages:
   - create-instances
   - run-instances
   - upload-to-marketplace
+
+create-instances:
+  stage: create-instances
+  script:
+    - |
+      demisto-sdk create-id-set -o ./Tests/id_set.json
+      cp ./Tests/id_set.json $ARTIFACTS_FOLDER
+    - python3 Utils/release_notes_generator.py $CONTENT_VERSION $CI_COMMIT_SHA $CI_BUILD_ID --output $ARTIFACTS_FOLDER/packs-release-notes.md --github-token $GITHUB_TOKEN
+    - ./Documentation/commonServerDocs.sh
+    -  demisto-sdk create-content-artifacts -a $ARTIFACTS_FOLDER --cpus 8 --content_version $CONTENT_VERSION
+    - |
+      gcloud auth activate-service-account --key-file="$GCS_ARTIFACTS_KEY" > auth.out 2>&1
+      echo "successfully activated google cloud service account"
+    - successful_feature_branch_build=$(gsutil ls "gs://xsoar-ci-artifacts/content/$FEATURE_BRANCH_NAME/*" | tail -n 1 | grep -o -E "content/$FEATURE_BRANCH_NAME/[0-9]*")
+    - echo $successful_feature_branch_build
+    - python3 Utils/merge_content_new_zip.py -f $FEATURE_BRANCH_NAME -b $successful_feature_branch_build
+    - zip -j $ARTIFACTS_FOLDER/uploadable_packs.zip $ARTIFACTS_FOLDER/uploadable_packs/* || ((($? > 0)) && echo "failed to zip the uploadable packs, ignoring the failure" && exit 0)
+    - rm -rf $ARTIFACTS_FOLDER/uploadable_packs
+    - python3 ./Tests/scripts/update_conf_json.py
+    - cp "./Tests/conf.json" "$ARTIFACTS_FOLDER/conf.json"
+    - |
+      if [ -n "${INSTANCE_TESTS}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          [ -n "${NIGHTLY}" ] && IS_NIGHTLY=true || IS_NIGHTLY=false
+          python3 ./Tests/scripts/collect_tests_and_content_packs.py -n $IS_NIGHTLY
+      fi
+    - python3 ./Tests/Marketplace/packs_dependencies.py -i ./Tests/id_set.json -o $ARTIFACTS_FOLDER/packs_dependencies.json
+    - |
+      if [[ -n "${CONTRIB_BRANCH}" || -n "${DEMISTO_SDK_NIGHTLY}" ]] ; then
+        echo "Skipping Preparing Content Packs For Testing since nightly instances uses production bucket"
+      else
+        ./Tests/scripts/prepare_content_packs_for_testing.sh "$GCS_MARKET_BUCKET"
+      fi
+    - |
+      if [[ -n "${DEMISTO_SDK_NIGHTLY}" ]] || [[ $CI_COMMIT_BRANCH != master ]] && [[ $CI_COMMIT_BRANCH != 20\.* ]] && [[ $CI_COMMIT_BRANCH != 21\.* ]]; then
+        echo "Skipping packs download to artifact on non master or release branch"
+      else
+        ZIP_FOLDER=$(mktemp -d)
+        python3 ./Tests/Marketplace/zip_packs.py -b 'marketplace-dist' -z $ZIP_FOLDER -a $ARTIFACTS_FOLDER -s $GCS_MARKET_KEY
+      fi
+    - |
+      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          ./Tests/scripts/upload_artifacts.sh
+      fi
+    - |
+      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          [ -n "${TIME_TO_LIVE}" ] && TTL=${TIME_TO_LIVE} || TTL=300
+          if [ -n "${NIGHTLY}" ] ;
+            then
+              export IFRA_ENV_TYPE=Nightly # disable-secrets-detection
+          elif [ -n "${INSTANCE_TESTS}" ] ;
+            then
+              export IFRA_ENV_TYPE="Server 5.5" # disable-secrets-detection
+          elif [[ -n "${CONTRIB_BRANCH}" || -n "${DEMISTO_SDK_NIGHTLY}" ]] ;
+            then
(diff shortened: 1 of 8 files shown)


.gitlab-ci.yml
```diff
@@ -85,3 +85,80 @@ stages:
...
+      fi
+    - python3 ./Tests/Marketplace/packs_dependencies.py -i ./Tests/id_set.json -o $ARTIFACTS_FOLDER/packs_dependencies.json
+    - |
+      if [[ -n "${CONTRIB_BRANCH}" || -n "${DEMISTO_SDK_NIGHTLY}" ]] ; then
+        echo "Skipping Preparing Content Packs For Testing since nightly instances uses production bucket"
+      else
+        ./Tests/scripts/prepare_content_packs_for_testing.sh "$GCS_MARKET_BUCKET"
+      fi
+    - |
+      if [[ -n "${DEMISTO_SDK_NIGHTLY}" ]] || [[ $CI_COMMIT_BRANCH != master ]] && [[ $CI_COMMIT_BRANCH != 20\.* ]] && [[ $CI_COMMIT_BRANCH != 21\.* ]]; then
```
DinaMeylakh, line 124: what do the 20\.* 21\.* commit branches mean?

.gitlab-ci.yml
```diff
@@ -85,3 +85,80 @@ stages:
...
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          ./Tests/scripts/upload_artifacts.sh
+      fi
+    - |
+      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          [ -n "${TIME_TO_LIVE}" ] && TTL=${TIME_TO_LIVE} || TTL=300
```
DinaMeylakh, line 142:
```suggestion
    - |
      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
        then
          echo "Skipping - not running in INSTANCE_TESTS build"
        else
          ./Tests/scripts/upload_artifacts.sh
          [ -n "${TIME_TO_LIVE}" ] && TTL=${TIME_TO_LIVE} || TTL=300
```

.gitlab-ci.yml
```diff
@@ -85,3 +85,80 @@ stages:
...
+    - echo $successful_feature_branch_build
+    - python3 Utils/merge_content_new_zip.py -f $FEATURE_BRANCH_NAME -b $successful_feature_branch_build
+    - zip -j $ARTIFACTS_FOLDER/uploadable_packs.zip $ARTIFACTS_FOLDER/uploadable_packs/* || ((($? > 0)) && echo "failed to zip the uploadable packs, ignoring the failure" && exit 0)
+    - rm -rf $ARTIFACTS_FOLDER/uploadable_packs
+    - python3 ./Tests/scripts/update_conf_json.py
+    - cp "./Tests/conf.json" "$ARTIFACTS_FOLDER/conf.json"
+    - |
+      if [ -n "${INSTANCE_TESTS}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
```
DinaMeylakh, line 111:
```suggestion
          echo "Skipping - not collecting tests and content packs for INSTANCE_TESTS build"
```

DinaMeylakh (commented):
Please add echos with titles for each step, it will help the debugging and looking for the same step in circle when things need interrogating

Tests/scripts/prepare_content_packs_for_testing.sh
```diff
@@ -3,28 +3,26 @@
...
-KF=$(mktemp)
-echo "$GCS_MARKET_KEY" > "$KF"
-gcloud auth activate-service-account --key-file="$KF" > auth.out 2>&1
+gcloud auth activate-service-account --key-file="$GCS_MARKET_KEY" > auth.out 2>&1
 echo "Auth loaded successfully."
 
 # ====== BUILD CONFIGURATION ======
 GCS_BUILD_BUCKET="marketplace-ci-build"
-BUILD_BUCKET_PATH="content/builds/$CIRCLE_BRANCH/$CIRCLE_BUILD_NUM"
+BUILD_BUCKET_PATH="content/builds/$CI_COMMIT_BRANCH/$CI_BUILD_ID"
```
guyfreund, line 25: Make sure we use the same path for `copy_and_upload_packs.py` in the Upload job, otherwise the upload will fail.

.gitlab-ci.yml
hod-alpert: No idea actually, taking them from the config.yml as is.

Itay4: these are content release branches

hod-alpert:
> Please add echos with titles for each step, it will help the debugging and looking for the same step in circle when things need interrogating, other than that looks good.

Not sure it's needed when we have the colors set like this.
seems pretty much self explanatory don't you think?
image

DinaMeylakh:
> > Please add echos with titles for each step, it will help the debugging and looking for the same step in circle when things need interrogating, other than that looks good.
> 
> Not sure it's needed when we have the colors set like this.
> seems pretty much self explanatory don't you think?
> image

I think that because of the collapsed lines it will still be tricky, plus I think the titles will help the average developer understand to what step to compare to in circle in case of something failing.

.gitlab-ci.yml
```diff
@@ -0,0 +1,164 @@
...
+    - |
+      if [[ -f "$BASH_ENV" ]]; then
+        source "$BASH_ENV"
+      fi
+    - echo "=== Creating new clean logs folder ==="
+    - rm -rf $ARTIFACTS_FOLDER/logs
+    - mkdir -p $ARTIFACTS_FOLDER/logs
+    - echo "=== Granting execute permissions on files ==="
+    - chmod +x ./Tests/scripts/*
+    - chmod +x ./Tests/lastest_server_build_scripts/*
```
yaakovi, line 25: I think we can stop supporting this folder (the flow doesn't work anymore).

need to document the alternative option (create a temporary AMI)

.gitlab-ci.yml
```diff
@@ -0,0 +1,164 @@
...
+    - python3 --version
+    - node --version
+    - npm --version
+    - demisto-sdk --version
+
+
+variables:
+  CONTENT_VERSION: "21.3.2"
+  SERVER_VERSION: "6.0.0"
+  GIT_SHA1: "6fee942518c6b2275a6cb61c1824574171cf58f9" # guardrails-disable-line disable-secrets-detection
```
yaakovi, line 68: those were moved to `.circleci/content_release_vars.sh`

.gitlab-ci.yml
```diff
@@ -0,0 +1,164 @@
...
+    - ./Documentation/commonServerDocs.sh
+    -  demisto-sdk create-content-artifacts -a $ARTIFACTS_FOLDER --cpus 8 --content_version $CONTENT_VERSION
+    - |
+      gcloud auth activate-service-account --key-file="$GCS_ARTIFACTS_KEY" > auth.out 2>&1
+      echo "successfully activated google cloud service account"
+    - successful_feature_branch_build=$(gsutil ls "gs://xsoar-ci-artifacts/content/$FEATURE_BRANCH_NAME/*" | tail -n 1 | grep -o -E "content/$FEATURE_BRANCH_NAME/[0-9]*")
+    - echo $successful_feature_branch_build
+    - python3 Utils/merge_content_new_zip.py -f $FEATURE_BRANCH_NAME -b $successful_feature_branch_build
+    - zip -j $ARTIFACTS_FOLDER/uploadable_packs.zip $ARTIFACTS_FOLDER/uploadable_packs/* || ((($? > 0)) && echo "failed to zip the uploadable packs, ignoring the failure" && exit 0)
+    - rm -rf $ARTIFACTS_FOLDER/uploadable_packs
```
yaakovi, line 105: why aren't those executed as the same step? they belong to the same logic unit

.gitlab-ci.yml
```diff
@@ -0,0 +1,164 @@
...
+      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
+        else
+          ./Tests/scripts/upload_artifacts.sh
+      fi
+    - |
+      if [ -n "${DEMISTO_SDK_NIGHTLY}" ];
+        then
+          echo "Skipping - not running in INSTANCE_TESTS build"
```
yaakovi, line 140: ?

.gitlab-ci.yml
hod-alpert: Its more verbose to run commands one by one IMO.

.gitlab-ci.yml
hod-alpert: Maintaining the same logic as in circle but since in gitlab all the jobs run from a single process I cannot use `exit` command.

hod-alpert:
All changes of this PR has been fixed.

This PR has been merged into gitlab-master branch and the changes can be reviewed at https://github.com/demisto/content/pull/12205

PR #12255, opened 2021-04-19
VirusTotal - Fixed an issue where the *Source Reliability* integratio…

bakatzir:
…n parameter was not properly set.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35934

## Description
screenshots in the issue.
also bullet proofed the usage of `params.integrationReliability`

## Minimum version of Demisto
- [x] 5.5.0

   - [x] No


yaakovi (approved):
looks ok, if we want to add those RN (they will be merged with the last version, might look confusing)

PR #12271, opened 2021-04-20
Moving FileType extention from content repo to the sdk

hod-alpert:
Moving FileType extention from content repo to the sdk in order to support python 3.9

## Related PRs:
https://github.com/demisto/demisto-sdk/pull/1219


Itay4:
this is ok because we import `constants` which include `FileType` ?

hod-alpert:
> this is ok because we import `constants` which include `FileType` ?

Exactly.
And the Enum extension is not supported in python 3.9

Itay4 (approved):

PR #12363, opened 2021-04-26
Kafka v2 - add max_bytes_per_message parameter

reutshal:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35726

## Description
Added the "max_bytes_per_message" as an integration parameter so that the customer could control the value and prevent timeout errors according to the issue mentioned. 

   - [ ] Yes
       - Further details:
   - [X] No


teizenman (approved):

reutshal:
@roysagi @amshamah419 Please take a look at the RN and at the new integration parameter 🙏

Packs/Kafka/Integrations/Kafka_V2/Kafka_V2.yml
```diff
@@ -53,6 +55,14 @@ configuration:
   name: incidentType
   required: false
   type: 13
+- additionalinfo: |-
+    The max number of bytes of messages to attempt to fetch with each fetch request. Should be in multiples of 1024. If the fetching is taking a long time, you should consider increasing this value. Default is '1048576'.
```
roysagi, line 59:
```suggestion
    The max number of message bytes to retrieve in each attempted fetch request. Should be in multiples of 1024. If the fetching process is taking a long time, you should consider increasing this value. Default is '1048576'.
```

PR #12373, opened 2021-04-26
Qradar long running test module bug

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36113

## Description
The test module was failing when the `limit` was `None` and the `longRunning` was `True`, due to the fact that the query in the test module was still sending the word 'limit' in the query yet with no value. after consulting with Dean, we do not want to enable that so added a suitable error message instead.

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


DeanArbel (changes requested):
I'm not sure we want to allow this feature. Please schedule a meeting with me and @David-BMS

Noy-Maimon:
@roysagi Hi, can you please check the error message in the .py file, the `additionalinfo` field in the .yml file, and the RN file? Thanks :)

DeanArbel (approved):
Nice, please make sure QRadar v3 handles this case as well.

reutshal (approved):

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -715,6 +715,9 @@ def test_module(client: QRadarClient):
     params = demisto.params()
     is_long_running = params.get("longRunning")
     if is_long_running:
+        if not params.get('events_limit'):
+            raise DemistoException('Unlimited long running instance fetch is not supported, please limit your fetch'
+                                   ' using the "Max number of events per incident" parameter.')
```
roysagi, line 720:
```suggestion
                                   ' by using the "Max number of events per incident" parameter.')
```

PR #12388, opened 2021-04-27
Fixing server keys issue

hod-alpert:
In case there is an integration with server keys - those server keys needs to be added to the current configurations, not overwrite them.


abaumgarten (approved):
Looks like this fix solved the issue (my build passes now- I added @hod-alpert's fix to my PR in order to test it).
https://app.circleci.com/pipelines/github/demisto/content/81647/workflows/c12a93e3-ac51-48f4-a5a9-9d3e25504bbc/jobs/337699

PR #12404, opened 2021-04-27
json feed add indicator_type to rawjson + test improvement

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
* Add an option that the `_indicator_type` is added to the rawjson so it is easy to use mapping of the auto detect value.
* Check in jsonfeed **Test** that *Indicator Type* and *Auto detect indicator type* are NOT both set.
* Fix an issue with `limit`

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/FeedJSON/Integrations/FeedJSON/FeedJSON.py
```diff
@@ -12,9 +12,14 @@ def main():
...
             'extractor': params.get('extractor'),
             'indicator': params.get('indicator', 'indicator'),
+            'rawjson_include_indicator_type': params.get('rawjson_include_indicator_type'),
         }
     }
 
+    if params.get('auto_detect_type') and params.get('indicator_type'):
+        return_error('Indicator Type should not be set if Auto Detect Indicator Type is checked.'
+                     ' Either use Auto Detect or set manually the Indicator Type.')
+
```
teizenman, line 22: We can make it more general by adding it to: https://github.com/demisto/content/blob/master/Packs/ApiModules/Scripts/JSONFeedApiModule/JSONFeedApiModule.py#L252

WDYT?

glicht: I didn't want to introduce a breaking change across all feeds that use this module. I would rather leave it in json feed. Other feeds usually set this anyway in a hard coded way.

teizenman (approved):

Packs/FeedJSON/Integrations/FeedJSON/FeedJSON.py
glicht: anyway also doing a commit to only do this validation on test-module command.

glicht:
@amshamah419 can you take a look at the JSON Feed release notes: https://github.com/demisto/content/pull/12404/files#diff-8f5c8742e64c355cd5201a1d735e41e4386c2bb5d0caab3236797397eaab419eR1 .

Packs/FeedJSON/ReleaseNotes/1_1_3.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### JSON Feed
+- **Breaking Change**: The integration now validates that not both configuration params: *Indicator Type* and *Auto detect indicator type* are set. You must set only one of the configuration params. Previously, if both were set, *Auto detect indicator type* was used and *Indicator Type* was silently ignored.
```
amshamah419, line 4:
```suggestion
- **Breaking Change**: The integration now validates that both configuration params: *Indicator Type* and *Auto detect indicator type* are not set. You may set only one of the configuration params at a time. Previously, if both were set, *Auto detect indicator type* was used and *Indicator Type* was silently ignored.
```

PR #12411, opened 2021-04-28
Added option to EDL Intgeration for appending a slash at the end of URLs

jtschichold:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
*No issue*

## Description
This PR adds an option to the Palo Alto Networks EDL integration to append a slash at the end of hostname only URLs.

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/EDL/Integrations/EDL/EDL.py
+++ b/Packs/EDL/Integrations/EDL/EDL.py
@@ -58,6 +58,7 @@ class RequestArguments:
                  limit: int = 10000,
                  offset: int = 0,
                  url_port_stripping: bool = False,
+                 url_add_slash: bool = False,
                  drop_invalids: bool = False,
                  collapse_ips: str = DONT_COLLAPSE):
 
@@ -65,6 +66,7 @@ class RequestArguments:
         self.limit = limit
         self.offset = offset
         self.url_port_stripping = url_port_stripping
+        self.url_add_slash = url_add_slash
         self.drop_invalids = drop_invalids
         self.collapse_ips = collapse_ips
 
@@ -79,6 +81,9 @@ class RequestArguments:
             return True
 
         elif self.url_port_stripping != last_update_data.get('url_port_stripping'):
+            return True
+
+        elif self.url_add_slash != last_update_data.get('url_add_slash'):
             return True
 
         elif self.collapse_ips != last_update_data.get('collapse_ips'):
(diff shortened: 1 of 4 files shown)


jtschichold:
While this PR does not break backwards compatibility inside XSOAR, in some scenarios it could break the communication with PAN-OS: if the slash is added at the end of an EDLs and the EDL is used in a PAN-OS Anti-Spyware profile, the trailing slash would make the domain invalid and PAN-OS will automatically ignore it.
Basically when the "Add Slash to URLs" option is used, the resulting EDL can only be used as an EDL for URLs.

anara123 (approved):

Packs/EDL/Integrations/EDL/EDL.py
```diff
@@ -330,6 +335,9 @@ def create_values_for_returned_dict(iocs: list, request_args: RequestArguments)
                 # invalid tokens in indicator- if drop_invalids is set - ignore the indicator
                 if request_args.drop_invalids:
                     continue
+            # if the indicator is a URL without a slash, we add a slash at the end
+            if request_args.url_add_slash and '/' not in indicator:
```
DeanArbel, line 339: @jtschichold this will affect all indicator types that are not `IP/CIDR` is this the intended result?
If this is meant for a specific type of indicator (such as URL) you can use the `ioc_type`. Otherwise, please ignore my comment.

DeanArbel (approved):
Look good, please review my comment.

Packs/EDL/Integrations/EDL/EDL.py
jtschichold: The url_add_slash option should be applied with the same logic as url_port_stripping option.
The existing code is applying url_port_stripping option to all the IOC that are not IPs/CIDRs.
In theory both url_add_slash should only be applied to URL, Domain and DomainGlob.
Please also check my comment on the PR about possible issue with existing EDL. Maybe the add_slash option should be set to false by default?

DeanArbel: @jtschichold I feel like the previous behaviour wasn't intended, seeing as it refers to URLs in the docs (@hod-alpert might know better). I won't ask to change the existing behaviour, but I think that going forward we should enforce a stricter check for `indicator type transformations`.
Please set it to `default: false` as you mentioned, since this might break BC.

@anara123 fyi

Packs/EDL/Integrations/EDL/EDL.yml
```diff
@@ -63,6 +63,13 @@ configuration:
   name: url_port_stripping
   required: false
   type: 8
+- additionalinfo: If selected, a slash will be appended to URLs composed only of an hostname.
+    For example, 'www.example.com' would become 'www.example.com/'
+  defaultvalue: 'false'
```
jtschichold, line 68: @DeanArbel default is now "false". This should work, right?

DeanArbel: yeah, this should be alright. The default`False` should handle `None` cases for instances saved before the change

ytauman:
hi @jtschichold - can you please elaborate on why this enhancement is needed? what is the use case?
it would be great if this could be done in a designated github issue :)

jtschichold:
Closing this. Will reopen if needed.

PR #12417, opened 2021-04-28
Running with --network=host on RHEL instances

hod-alpert:
Adding `--network=host` on podman containers to solve the latest podman issue with the proxy.

Build example: https://app.circleci.com/pipelines/github/demisto/content/82083/workflows/c2837755-91c8-4258-a83b-84b6fdc17e3e/jobs/339270 (Follow `Thread-10` in the logs)

### DIFF
--- a/Tests/configure_and_test_integration_instances.py
+++ b/Tests/configure_and_test_integration_instances.py
@@ -41,7 +41,7 @@ NO_PROXY = ','.join([
     'oproxy.demisto.ninja',
     'oproxy-dev.demisto.ninja',
 ])
-NO_PROXY_CONFIG = {'python.pass.extra.keys': f'--env##no_proxy={NO_PROXY}'}  # noqa: E501
+NO_PROXY_CONFIG = {'python.pass.extra.keys': f'--env##no_proxy={NO_PROXY}##--network=host'}  # noqa: E501
 DOCKER_HARDENING_CONFIGURATION = {
     'docker.cpu.limit': '1.0',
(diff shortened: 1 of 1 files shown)


Tests/configure_and_test_integration_instances.py
```diff
@@ -41,7 +41,7 @@
     'oproxy.demisto.ninja',
     'oproxy-dev.demisto.ninja',
 ])
-NO_PROXY_CONFIG = {'python.pass.extra.keys': f'--env##no_proxy={NO_PROXY}'}  # noqa: E501
+NO_PROXY_CONFIG = {'python.pass.extra.keys': f'--env##no_proxy={NO_PROXY}##--network=host'}  # noqa: E501
```
Itay4, line 44: if i understand correctly, this also affects instances that run with docker
shouldnt we apply it only to instances with podman?

glicht: It should be only with podman

glicht:
after some research let's use following config: `--network=slirp4netns:cidr=192.168.0.0/16`

hod-alpert:
@glicht here's a new build with the new configuration https://app.circleci.com/pipelines/github/demisto/content/82158/workflows/5b48fa04-c7e5-4fb4-a906-75447bf2f118/jobs/339553.

The host ip was `10.0.11.62`, but at the very least we can tell nothing is broken, I think we can merge this.

Itay4 (approved):

glicht (approved):

PR #12419, opened 2021-04-28
return results failed to receive a result entry in a list

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Description
If return results would have received a list containing a CommandResult object and a dict representing a demisto entry it would have failed as it was trying to run `.to_context()` function on the dict. Instead, the new solution calls the `return_results` function again for each non string/dict item, and for the regular str/dict items it calls `demisto.results`, such that if one of them is a `demisto.results` dict it will be handled correctly.

## Test Failure
https://app.circleci.com/pipelines/github/demisto/content/82455/workflows/f8fbc9f8-ba85-415a-9d4b-ddf9da37d7b6/jobs/340977


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -5033,36 +5033,38 @@ def return_results(results):
         demisto.results(None)
         return
 
-    if results and isinstance(results, list) and len(results) > 0 and isinstance(results[0], CommandResults):
+    elif results and isinstance(results, list) and len(results) > 0:
```
yaakovi, line 5036: if the first two conditions are True then the third must be True.
```suggestion
    elif results and isinstance(results, list):
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -5033,36 +5033,38 @@ def return_results(results):
...
+        result_list = []
         for result in results:
-            demisto.results(result.to_context())
-        return
+            if isinstance(result, (dict, str)):
```
yaakovi, line 5039: why can't we run `return_results` for them too? performance-wise?
if so, we should include `int` as well (transformer scripts)

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
ShahafBenYakir: ma.
which first? if results is an empty list you want to call demisto.results on it (and this is what it won't get into this if)

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
ShahafBenYakir: No because of BC. Until now if you called `return_results([entry1, entry2])`, demisto.results would've been called once with a list argument, if you will call `return_results` on each item you will be calling demisto.results for each entry (twice).

yaakovi (approved):

PR #12467, opened 2021-05-02
test version

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36408

## Description
we don't need to separate the list just to extend if it’s necessary.

- [ ] 5.5.0
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
   - [x] No

## Must have
- [ ] Tests

### DIFF
--- a/Packs/ExpanseV2/Integrations/ExpanseV2/ExpanseV2.py
+++ b/Packs/ExpanseV2/Integrations/ExpanseV2/ExpanseV2.py
@@ -990,7 +990,7 @@ def get_issues_command(client: Client, args: Dict[str, Any]) -> CommandResults:
     total_results, max_page_size = calculate_limits(args.get('limit', None))
 
     provider = ','.join(argToList(args.get('provider')))
-    business_units = ','.join(argToList(args.get('business_unit')))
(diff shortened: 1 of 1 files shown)


Packs/ExpanseV2/Integrations/ExpanseV2/ExpanseV2.py
```diff
@@ -990,7 +990,7 @@ def get_issues_command(client: Client, args: Dict[str, Any]) -> CommandResults:
     total_results, max_page_size = calculate_limits(args.get('limit', None))
 
     provider = ','.join(argToList(args.get('provider')))
-    business_units = ','.join(argToList(args.get('business_unit')))
+    business_units = ','.join(argToList(args.get('business_unit'), separator='__not_required_here__'))
```
yaakovi, line 993: this is not a good solution and breaks BC.

yaakovi, line 993: how will it behave in the case of a list input that one element contains a comma?

PR #12527, opened 2021-05-04
Supporting zipping content packs from GCS on master branch

hod-alpert:
In the new env - GCS keys are files, not strings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -819,11 +819,8 @@ jobs:
                     exit 0
                   fi
 
-                  GCS_PATH=$(mktemp)
                   ZIP_FOLDER=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


avidan-H (commented):
How will this work? The `GCS_MARKET_KEY` being a file variable is only in GitLab I thought.

hod-alpert:
> How will this work? The `GCS_MARKET_KEY` being a file variable is only in GitLab I thought.

To enable the same behaviour I added this file https://github.com/demisto/content/blob/hod/gitlab_support_fix/.circleci/gitlab-ci-env-variables.sh

avidan-H (approved):

yaakovi (approved):

PR #12541, opened 2021-05-04
Check point test failed bug

MosheEichler:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36572

## Description
add a try/except to the test-module to avoid error in case there is not json, As it seems in the logs in case the status code = 500 no json is returned

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CheckpointFirewall/ReleaseNotes/2_0_9.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### CheckPoint Firewall v2
+- Fixed an issue where the ***checkpoint-show-objects*** command failed.
```
reutshal, line 4: Please be more specific in the RN - mention that the error from the API was not handled correctly. Also mention the version API fix

reutshal (changes requested):

Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
```diff
@@ -1776,13 +1776,19 @@ def test_module(base_url: str, sid: str, verify_certificate) -> str:
...
+        response = requests.post(base_url + 'show-api-versions',
+                                 headers={'Content-Type': 'application/json', 'X-chkp-sid': sid},
+                                 verify=verify_certificate, json={})
+        reason = ''
+        if response.json().get('message') == "Missing header: [X-chkp-sid]":
+            reason = '\nWrong credentials! Please check the username and password you entered and try' \
+                     ' again.\n'
+    except Exception as e:
+        if '500' in str(e):  # for status code 500
+            return 'Server Error: make sure Server URL and Server Port is correctly set'
```
reutshal, line 1789:
```suggestion
            return 'Server Error: make sure Server URL and Server Port are correctly set'
```

Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
```diff
@@ -1796,6 +1802,8 @@ def main():
     password = params.get('username', {}).get('password')
 
     server = params['server']
+    if server[-1] == "/":
```
reutshal, line 1805: Remind my why did we add this?

MosheEichler: If the user enters in the URL field something like this **www.example.com/**, Later in the code we connect the port to the url so it's will be like this **www.example.com/:443/web_api** instead of **www.example.com:443/web_api**
Screen Shot 2021-06-10 at 16 03 37

reutshal (approved):

PR #12629, opened 2021-05-10
URL apostrophes bug fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36751

## Description
Added support for apostrophes when extracting URL.


teizenman (changes requested):
Add a test to the playbook

reutshal:
> Add a test to the playbook

@dansterenson There is a test-playbook that checks the indicators extraction, called - "Indicators-reputation-Test"
Please add an example of the URL and make sure we extract it as expected.

Packs/Base/TestPlaybooks/playbook-Indicators-reputation-.json-Test.yml
```diff
@@ -1017,6 +1017,14 @@ tasks:
...
             value:
               simple: http://www.foo.bar/taz?())
+      - - operator: notContainsGeneral
+          left:
+            value:
+              simple: ${URL.Data}
+            iscontext: true
+          right:
+            value:
+              simple: https://Test.com/this'that
```
teizenman, line 1027: It's not the same as the input, is that on purpose?

reutshal (approved):

Packs/Base/TestPlaybooks/playbook-Indicators-reputation-.json-Test.yml
dansterenson: Yes, when I upload the updated playbook and execute it, I get:

Before the changes (URL doesn't support apostrophes):
image

After the changes (URL does support apostrophes):
image

teizenman (approved):

PR #12651, opened 2021-05-11
GetIndicatorDBotScore: fixed the ignored vendor score

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36877

- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
   - [x] No

## Must have
- [x] Tests


Packs/CommonScripts/Scripts/GetIndicatorDBotScore/GetIndicatorDBotScore.py
```diff
@@ -26,12 +26,15 @@ def get_dbot_score_data(indicator, indicator_type, source, score):
...
     indicator_type = entry["indicator_type"]
     indicator_type = INDICATOR_TYPES.get(indicator_type, indicator_type).lower()
-    sources = entry.get('sourceBrands', [])
-    sources = sources if sources else [None]
-    for source in sources:
+    sources = entry.get('moduleToFeedMap', {})
+    if entry.get('manualScore'):
+        sources[entry.get('setBy')] = {}
+    elif not sources:
+        sources[None] = {}
```
yaakovi, line 33: why? isn't that creates a dummy indicator?

esharf: @yaakovi no.
Because we have the indicator in the query and the score in the entry (here it’s just a part of the entry).
you can an example for this in https://github.com/demisto/content/blob/0805e4a58b64f65d15be59cbd28b2f4f8bd8ae80/Packs/CommonScripts/Scripts/GetIndicatorDBotScore/GetIndicatorDBotScore_test.py#L39-#L63

yaakovi (approved):

PR #12663, opened 2021-05-11
fixing validation

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Description
Not validating that `HelloWorldPremium` is on the premium server packs page. 
It isn't there, by design (for now) since we reduced its price to 0.


DinaMeylakh (approved):
Will it still be pushed to index as premium? why not ask server to show the HelloWorldPremium in the Premium filter instead?

guykeller:
@DinaMeylakh We did discuss this option, I can explain in private

PR #12683, opened 2021-05-12
Cylance context paths bug

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36643

## Description
The context outputs under `File` had 2 errors:
1. `Sha256` instead of `SHA256` - fix: the capitalised version was added under `File`
2.` DbotScore.Indicator` had the file's name instead of the hash - fix: the hash is now under `DbotScore.Indicator` (and the file name was removed)

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [x] Yes
       - Further details: only in cases a customer used `DbotScore.Indicator` for the file's name (which is not the usecase for that field).
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
```diff
@@ -1347,6 +1352,47 @@ def fetch_incidents():
     demisto.setLastRun({'time': current_run.isoformat().split('.')[0]})
 
 
+def add_capitalized_hash(threats_context):
```
reutshal, line 1354: Please improve the name of this method, it is not very self-explanatory.
I would suggest something like add_capitalized_hash_to_context.

Also, I think you can handle both cases in one method, see my suggestion

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
```diff
@@ -1347,6 +1352,47 @@ def fetch_incidents():
...
+
+    Returns:
+        threats_context(dict): dicts of context outputs for the threat of interest, containing
+        the key and value 'Sha256' (and possibly Md5) as well as the key and value 'SHA256' (and possible MD5).
+    """
+    if threat_context.get('Sha256'):
+        threat_context['SHA256'] = threat_context.get('Sha256')
+    if threat_context.get('Md5'):
+        threat_context['MD5'] = threat_context.get('Md5')
+    return threat_context
```
reutshal, line 1392:
```suggestion
    if not isinstance(threats_context, list):
        threats_context = [threats_context]
       
    for context_item in threats_context:
        if context_item.get('Sha256'):
            context_item['SHA256'] = context_item.get('Sha256')
        if context_item.get('Md5'):
            context_item['MD5'] = context_item.get('Md5')

    return threats_context
```

Make sure to handle the case where the context is empty / None if possible

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
```diff
@@ -265,6 +265,9 @@ script:
     - contextPath: File.SHA256
       description: The SHA256 hash of the threat.
       type: string
+    - contextPath: File.MD5
+      description: The SHA256 hash of the threat.
+      type: string
```
reutshal, line 270: Shouldn't new context paths need to be added to all commandsd using the new method? Is it already there?

reutshal (changes requested):

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
Noy-Maimon: They are already there. The thing is that there are no File.Sha256 for any of the commands (the CamleCase version). I wasnt sure if if should add them, what do you think?

Noy-Maimon: They are already there. The thing is that there are no File.Sha256 for any of the commands (the CamleCase version). I wasnt sure if if should add them, what do you think?

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
Noy-Maimon: Done

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.yml
reutshal: Add them, if the output paths don't match the real context then it's another bug that should be solved.

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
```diff
@@ -1347,6 +1351,30 @@ def fetch_incidents():
...
+
+    Args:
+        threats_context(list): list of dicts of context outputs for the threats of interest, each containing
+        the key 'Sha256' (and possibly (Md5)).
+
+    Returns:
+        threats_context(list): list of dicts of context outputs for the threats of interest, each containing
+        the key and value 'Sha256' (and possibly Md5) as well as the key and value 'SHA256' (and possible MD5).
+    """
+    if not isinstance(threats_context, list):
```
reutshal, line 1366: Can there be a situation where the context is None or an empty dict? I would add a check for that

reutshal (approved):

Packs/Cylance_Protect/Integrations/Cylance_Protect_v2/Cylance_Protect_v2.py
Noy-Maimon: Done

reutshal (changes requested):
Waiting on the yml outputs

reutshal (approved):

PR #12702, opened 2021-05-13
Change ExampleJSScript to work against reachable domain

asafshen:
`http://www.demisto.com` seems to be moved permanently (status 301)

changed the example JS script to work against `http://paloaltonetworks.com`

fixes https://github.com/demisto/etc/issues/36993


Packs/CommonScripts/Scripts/script-ExampleJSScript.yml
```diff
@@ -65,7 +65,7 @@ script: >-
 
   // Util function http
 
-  var res = http('http://www.demisto.com');
+  var res = http('http://www.paloaltonetworks.com');
```
Itay4, line 68:
```suggestion
  var res = http('https://www.paloaltonetworks.com');
```

Itay4 (approved):

Packs/CommonScripts/Scripts/script-ExampleJSScript.yml
asafshen: @Itay4 I rather keep it `http` in order to not depend on the site certificates

yaakovi:
@asafshen
please also replace the `Demisto server` on line 6 to `Cortex XSOAR server`

PR #12707, opened 2021-05-13
Add enforceWarninglist to search functions

stevengoossensB:
## Status
- [ ] In Progress
- [x] Ready

## Description
Adds the possibility to exclude attributes on the warninglist from searches

- [ x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ x ] No

## Must have
- [ ] Tests
- [ ] Documentation


stevengoossensB:
Added release notes and did version bump, as required

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
```diff
@@ -1990,6 +2001,17 @@ script:
       - 'false'
       required: false
       secret: false
+    - auto: PREDEFINED
+      default: true
```
yaakovi, line 2005:
```suggestion
      default: false
```

Packs/MISP/Integrations/MISP_V2/MISP_V2.py
```diff
@@ -960,7 +960,8 @@ def search(post_to_warroom: bool = True) -> Tuple[dict, Any]:
         'last',
         'eventid',
         'uuid',
-        'to_ids'
+        'to_ids',
+        'enforceWarninglist'
```
yaakovi, line 964:
```suggestion
        'enforceWarninglist',
```

Packs/MISP/Integrations/MISP_V2/MISP_V2.py
```diff
@@ -1030,7 +1034,8 @@ def search_attributes() -> Tuple[dict, Any]:
         'uuid',
         'to_ids',
         'last',
-        'include_decay_score'
+        'include_decay_score',
+        'enforceWarninglist'
```
yaakovi, line 1038:
```suggestion
        'enforceWarninglist',
```

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
```diff
@@ -135,6 +135,17 @@ script:
...
```
yaakovi, line 139:
```suggestion
      default: false
```

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
```diff
@@ -135,6 +135,17 @@ script:
...
+      defaultValue: 'true'
```
yaakovi, line 140: what is the default behavior when this is not specified in the request?
shouldn't this be false?

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
```diff
@@ -1990,6 +2001,17 @@ script:
...
```
yaakovi, line 2006: what is the default behavior when this is not specified in the request?
shouldn't this be false?

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
stevengoossensB: Hi, the default behaviour is 'false', meaning it will also search IOCs that are on warninglists etc. I've put it here to true, since I believe this is the more prudent behaviour, not causing false positive because of bad IOCs (common domain names, shared hosting IPs,...)

yaakovi: sounds reasonable.
however, from a BC point of view - we should avoid changing the current behavior and keep the new default behavior similar to the original one.

please remove the `defaultValue` key.

Packs/MISP/Integrations/MISP_V2/MISP_V2.yml
yaakovi: same here

yaakovi (approved):

PR #12709, opened 2021-05-13
Expanse   handle incident   add else to condition

tomer-pan:
## Status
- [] In Progress
- [x] Ready

## Description
An "else" path was added to a conditional task.

- [ ] 5.5.0
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No


tomer-pan:
This was pushed in #13155

PR #12712, opened 2021-05-13
Trend Micro DDA: attempted to fix bad request

yaakovi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35030

## Description
attempted to fix bad request by removing the 'Host' header when empty

## Notes
* not converting to package as we don't have an instance
* not adding tests as we don't have an instance


Itay4 (approved):

yaakovi:
copied release notes from other RNs, adding `docs-approved` label

PR #12718, opened 2021-05-15
TAXII Server - fix returned server url address

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37028

## Description
the assumptions of setting of the server URL address, which was returned in the response, were wrong and caused the wrong URL to be returned

this pull requests fixes both the URL returned by parsing the request:
1. setting the URL scheme from the request env var set by wsgi
2. setting the URL endpoint `/instance/execute` if it is used
3. 
   - [x] No

## Must have
- [x] Tests


liorblob (commented):
No changes required for
`def get_poll_response(self, taxii_message: PollRequest) -> Response:`
?

Packs/TAXIIServer/Integrations/TAXIIServer/TAXIIServer.py
```diff
@@ -281,15 +288,23 @@ def yield_response() -> Generator:
...
+        base_url = f'{self.url_scheme}://{self.host}'
+        if request_headers and '/instance/execute' in request_headers.get('X-Request-URI', ''):
+            # if the server rerouting is used, then the X-Request-URI header is added to the request by the server
+            # and we should use the /instance/execute endpoint in the address
+            calling_context = get_calling_context()
+            instance_name = calling_context.get('IntegrationInstance', '')
+            return os.path.join(base_url, 'instance', 'execute', instance_name)
         else:
-            return self.host
+            return f'{base_url}:{self.port}'
```
glicht, line 307: I may be wrong here but I think when not going through the server, we set the protocol to `https`.

glicht:
Can we create a playbook test that does requests to the long running port both via direct port and via server https and validate that we get the correct values?

Packs/TAXIIServer/Integrations/TAXIIServer/TAXIIServer.py
Itay4: not sure i follow
if we're not using the server rerouting, then the scheme can be either http or https - depends if a certificate was set in the integration instance
basically it can also be https if they're using reverse proxy to direct the traffic the https traffic to the http integration web app
am i missing something?

glicht (approved):

Packs/TAXIIServer/Integrations/TAXIIServer/TAXIIServer.py
```diff
@@ -281,15 +288,24 @@ def yield_response() -> Generator:
...
             The service URL according to the protocol.
         """
-        if self.http_server:
-            return f'{self.host}:{self.port}'
+        if request_headers and '/instance/execute' in request_headers.get('X-Request-URI', ''):
+            # if the server rerouting is used, then the X-Request-URI header is added to the request by the server
+            # and we should use the /instance/execute endpoint in the address
+            self.url_scheme = 'https'
+            calling_context = get_calling_context()
+            instance_name = calling_context.get('IntegrationInstance', '')
```
glicht, line 304: How do we deal here with an instance name that has spaces or special chars that require url encoding?

glicht (commented):
Please also add the `TIM` tag in pack_metadata

Itay4:
@glicht added the URL param

Packs/TAXIIServer/Integrations/TAXIIServer/TAXIIServer.py
glicht: @Itay4 I think we need to URL encode here?

glicht (approved):

Packs/TAXIIServer/Integrations/TAXIIServer/TAXIIServer.py
Itay4: yea added it in the line below

glicht (approved):

PR #12726, opened 2021-05-18
Service now v2 get tickets return empty bug

MosheEichler:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36981

## Description
add a condition to check if the response value of key ***opened_by*** is a dict, in case we use the ***sysparm_exclude_reference_link=true*** parameter there is no references and the value of key ***opened_by*** is not a dict

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -208,14 +208,14 @@ def create_ticket_context(data: dict, additional_fields: list = None) -> Any:
 
     # These fields refer to records in the database, the value is their system ID.
     closed_by = data.get('closed_by')
-    if closed_by:
+    if closed_by and type(closed_by) == dict:
         context['ResolvedBy'] = closed_by.get('value', '')
     opened_by = data.get('opened_by')
-    if opened_by:
+    if opened_by and type(opened_by) == dict:
```
reutshal, line 214: What about all the else cases here?

reutshal (changes requested):

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
MosheEichler: I fixed it
I attached 2 pictures to understand the difference when using the ***sysparm_exclude_reference_link=true*** parameter
Screen Shot 2021-05-18 at 17 24 09
Screen Shot 2021-05-18 at 17 25 14

reutshal (approved):

PR #12770, opened 2021-05-20
Zip content packs step optimization

PA-Rweins:
- Changed copy of content packs to using the gsutil tool.
- Created a separate function to find the zip files downloaded from the storage bucket.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29253

## Description
There was a need to optimise the time of the zip content packs step. It was optimised by using the gsutil tool.

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


PA-Rweins:
## ETA
- Code the solution itself - 1.5 days
- Tests - 0.5 days
- cr + fixes 1 day

PA-Rweins:
@guyfreund 
I tested it using the specific storage bucket you provided me.
How should it be tested for a general build using a general bucket?

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,109 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    if gcp_path == BUILD_GCP_PATH:
+        src_path = os.path.join('gs://', storage_bucket_name, gcp_path, branch_name, circle_build, 'content', 'packs')
+    else:
+        src_path = os.path.join('gs://', storage_bucket_name, gcp_path, branch_name, circle_build)
+
+    if not branch_name or not circle_build:
+        src_path = src_path.replace('/builds/content', '')
+
+    try:
+        process = subprocess.Popen(["Tests/scripts/cp_gcp_dir.sh", src_path, destination_path])
```
glicht, line 194: Use check_output which is a higher level api: https://docs.python.org/3/library/subprocess.html#subprocess.check_output

Tests/scripts/cp_gcp_dir.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+gsutil -m cp -r "$1" "$2"
```
glicht, line 2: Seem like a waste to have a file for a one line exec. You can simply call subprocess to execute this command with the specified args.

glicht (changes requested):

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+        sys.exit(1)
+
+
+def copy_to_other_dir(zipped_packs):
+    """
+    Copies zip files if needed
+    Args:
+        zipped_packs: A list of dictionaries containing pack name as key and it's latest zip path as value
+    """
+    if os.path.exists('/home/runner/work/content-private/content-private/content/artifacts/'):
```
anara123, line 211: we should not use hard coded path.
The best practice is to define an input for the whole script, set this path as default path `'/home/runner/work/content-private/content-private/content/artifacts/'`, and path this path and other pathes from function to function.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
-            else:
-                logging.warning(f'Did not find a pack to download with the prefix: {pack_prefix}')
+    if gcp_path == BUILD_GCP_PATH:
+        zip_path = os.path.join(zip_path, 'packs')  # directory of the packs
+    else:
+        zip_path = os.path.join(zip_path, circle_build)  # directory of the packs
+
+    dir_entries = os.listdir(zip_path)
+    print(f'zip path is: {zip_path}')
+    print(f"List of packs in repo is {', '.join(dir_entries)}")
```
anara123, line 153: either remove prints or use logging module

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    try:
+        process = subprocess.Popen(["Tests/scripts/cp_gcp_dir.sh", src_path, destination_path])
+        out, err = process.communicate()
+        if out:
+            print("out" + out.decode('utf-8'))
+        if err:
+            print("err" + err.decode('utf-8'))
+    except Exception as e:
+        logging.critical(f"Failed to run cp_gcp_dir.sh, Error:{e}")
+        sys.exit(1)
```
teizenman, line 202: This function is surrounded by `try-except` in the main function, maybe you should consider changing the logic here to handle all exceptions in the main.

Tests/Marketplace/zip_packs.py
```diff
@@ -244,12 +288,20 @@ def main():
     zipped_packs = []
     success = True
     try:
-        zipped_packs = download_packs_from_gcp(storage_bucket, gcp_path, zip_path, circle_build, branch_name)
+        download_packs_from_gcp(storage_bucket_name, gcp_path, zip_path, circle_build, branch_name)
     except Exception:
         logging.exception('Failed downloading packs')
         success = False
```
teizenman, line 294: Does it make sense to go on with the logic when you were not able to download from GCP?

Tests/Marketplace/zip_packs.py
```diff
@@ -244,12 +288,20 @@ def main():
...
 
-    if remove_test_playbooks:
+    try:
+        zipped_packs = get_zipped_packs_names(zip_path, gcp_path, circle_build)
+    except Exception:
+        logging.exception('No zip files were found')
+        success = False
```
teizenman, line 300: Same question as above.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+        if out:
+            print("out" + out.decode('utf-8'))
+        if err:
+            print("err" + err.decode('utf-8'))
+    except Exception as e:
+        logging.critical(f"Failed to run cp_gcp_dir.sh, Error:{e}")
+        sys.exit(1)
+
+
+def copy_to_other_dir(zipped_packs):
```
guyfreund, line 205:
```suggestion
def copy_zipped_packs_to_artifacts(zipped_packs):
```
Better to give more informational naming here

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    Args:
+        storage_bucket_name: The name of the GCP bucket to download from.
+        gcp_path: The path of the packs in the GCP bucket.
+        destination_path: The path to download the packs to.
+        branch_name: The branch name of the build.
+        circle_build: The number of the circle ci build.
+    """
+
+    if gcp_path == BUILD_GCP_PATH:
+        src_path = os.path.join('gs://', storage_bucket_name, gcp_path, branch_name, circle_build, 'content', 'packs')
```
guyfreund, line 186: Please assign the script with `STORAGE_BASE_PATH` var, so we can control the calling from outside without needing to differentiate between buckets. See `upload_packs.py` as an example.

Tests/Marketplace/zip_packs.py
```diff
@@ -229,10 +277,6 @@ def main():
             os.mkdir(zip_path)
         artifacts_path = '/home/runner/work/content-private/content-private/content/artifacts'
```
guyfreund, line 278: Please expose a script var for artifacts folder.

Tests/Marketplace/zip_packs.py
```diff
@@ -198,9 +247,9 @@ def get_latest_pack_zip_from_blob(pack, blobs):
         blob: The zip blob of the pack with the latest version.
     """
     blob = None
-    blobs = [b for b in blobs if os.path.splitext(os.path.basename(b.name))[0] == pack and b.name.endswith('.zip')]
+    blobs = [b for b in blobs if os.path.splitext(os.path.basename(b))[0] == pack and b.endswith('.zip')]
     if blobs:
-        blobs = sorted(blobs, key=lambda b: LooseVersion(os.path.basename(os.path.dirname(b.name))), reverse=True)
+        blobs = sorted(blobs, key=lambda b: LooseVersion(os.path.basename(os.path.dirname(b))), reverse=True)
```
teizenman, line 252: You're going over the list twice, you can do all the logic with one pass.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+            latest_zip = get_latest_pack_zip_from_blob(entry, pack_files)
+            if not latest_zip:
+                logging.warning(f'Failed to get the zip of the pack {entry} from GCP')
+                continue
+            print(f"Found latest zip of {entry}, which is {latest_zip}")
+            zipped_packs.append({Path(latest_zip).stem: latest_zip})
 
     if not zipped_packs:
         logging.critical('Did not find any pack to download from GCP.')
         sys.exit(1)
```
teizenman, line 170: Same as above.

Tests/Marketplace/zip_packs.py
```diff
@@ -244,12 +288,20 @@ def main():
...
         success = False
 
-    if remove_test_playbooks:
+    try:
+        zipped_packs = get_zipped_packs_names(zip_path, gcp_path, circle_build)
+    except Exception:
+        logging.exception('No zip files were found')
+        success = False
+
+    copy_to_other_dir(zipped_packs)
```
teizenman, line 302:
```suggestion
    if private_build:
        copy_to_other_dir(zipped_packs)
```

Tests/Marketplace/zip_packs.py
```diff
@@ -244,12 +288,20 @@ def main():
...
+
+    if zipped_packs and remove_test_playbooks:
```
teizenman, line 304: Currently, if `zipped_packs` is empty the script has already ended.

teizenman (changes requested):
Change all prints to logging.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    """
+    if os.path.exists('/home/runner/work/content-private/content-private/content/artifacts/'):
+        for zip_pack in zipped_packs:
+            for name, path in zip_pack.items():
+                logging.info(f"Copying pack from {path} to /home/runner/work/content-private/"
+                             f"content-private/content/artifacts/packs/{name}.zip")
+                shutil.copy(path,
+                            f'/home/runner/work/content-private/content-private/content/artifacts/'
+                            f'packs/{name}.zip')
+
```
guyfreund, line 219: Please handle `Shutil.Error`

guyfreund (changes requested):

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed, thanks.

Tests/scripts/cp_gcp_dir.sh
PA-Rweins: I originally used it as part of the python code but was told to move it to a bash script. So should I put it as part of the python code anyway?

Tests/Marketplace/zip_packs.py
PA-Rweins: I moved the logic of copying to artifacts from another function. It was already used as a hard coded path before my changes.
This path is the only path that will be used in this function. Maybe I should just change it to a constant value in the beginning of this python file and use it here?

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,110 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    if not branch_name or not circle_build:
+        src_path = src_path.replace('/builds/content', '')
+
+    try:
+        process = subprocess.Popen(["Tests/scripts/cp_gcp_dir.sh", src_path, destination_path])
+        out, err = process.communicate()
+        if out:
+            print("out" + out.decode('utf-8'))
+        if err:
+            print("err" + err.decode('utf-8'))
```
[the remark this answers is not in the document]
PA-Rweins: Changed to check_output as @glicht suggested and from prints to logging

Tests/Marketplace/zip_packs.py
PA-Rweins: you're right, removed it.

Tests/Marketplace/zip_packs.py
PA-Rweins: It seems unnecessary to me as well. It was added by @liorblob.
Is there a reason you've added it and didn't just end the entire process?

liorblob: I guess it's for the `cleanup()` at the end

Tests/Marketplace/zip_packs.py
PA-Rweins: This function logic was added in an earlier version, I just removed the attribute access to match the changes I made. Should I refactor this code anyway?

Tests/Marketplace/zip_packs.py
PA-Rweins: I guess you're referring to the try-except I used in another function? 
If yes, then I can simply remove the logging message and keep the sys.exit, it will still raise an exception that will be caught by the main function. WDYT?

Tests/Marketplace/zip_packs.py
PA-Rweins: In the original code this copy code happened without checking for the private_build. Are you sure it's needed?

Tests/Marketplace/zip_packs.py
PA-Rweins: As mentioned by @liorblob above. If zipped packs is empty we still need to call the cleanup function in the end of the script.

Tests/Marketplace/zip_packs.py
PA-Rweins: Thanks, changed it.

Tests/Marketplace/zip_packs.py
PA-Rweins: Ok I added it for the first case.
What about the other case when gcp_path != BUILD_GCP_PATH.
We probably also need to change the config.yml so if gcp_path != BUILD_GCP_PATH it will create the correct path right?

guyfreund: yes

Tests/Marketplace/zip_packs.py
PA-Rweins: Do you mean a variable that can be set from the step call in config.yml?
Because it seems that there's already one, you can see it in line 251.

guyfreund: yes, no need to hard-code the artifacts path, just pass it to the script

Tests/Marketplace/zip_packs.py
PA-Rweins: Sure, added handling.

Tests/Marketplace/zip_packs.py
teizenman: I believe so since you're trying to optimize the functionality of the step.

Tests/Marketplace/zip_packs.py
teizenman: Note the `except Exception` will not catch `SystemExit`.
Reference: https://docs.python.org/3/library/exceptions.html#base-classes

Tests/Marketplace/zip_packs.py
teizenman: I think it's needed to increase the readability of the code, it will help the next developer to understand it better.

Tests/Marketplace/zip_packs.py
teizenman: In the `get_zipped_packs_names` function you have the following code:
`    if not zipped_packs:
        sys.exit(1)`
So it's not going to get to line 304 in this case.

Tests/Marketplace/zip_packs.py
PA-Rweins: Ok I deleted this hard-coded value and kept the definition where it gets the value from the config.yml. The $CIRCLE_ARTIFACTS should contain the right value either when it's private build or not?

Tests/Marketplace/zip_packs.py
PA-Rweins: Added it.

Tests/Marketplace/zip_packs.py
PA-Rweins: Sure, changed it.

Tests/scripts/cp_gcp_dir.sh
guyfreund: It can be done from the config yml

Tests/Marketplace/zip_packs.py
PA-Rweins: I read the code again and I can't see where it goes over the list twice. It seems that first it creates a list of only zip files and then it sorts that list. 
I'd be happy if you could please explain where you see an extra traversal on the list.

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed.

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed.

Tests/Marketplace/zip_packs.py
PA-Rweins: You're right. I changed it now so there's only sys.exit(1) after cleanup was called and only if success is false

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed it.

Tests/scripts/cp_gcp_dir.sh
PA-Rweins: Yes it can be done from the config yml, however it means moving some logic from python to config yml. What is our preference?

teizenman (commented):
Looking good from my side.

Tests/scripts/cp_gcp_dir.sh
glicht: You can do this in Python with one line 
```python
subprocess.check_output(['gsutil', '-m', 'cp', '-r, src, dest])
```
Best to remove this file and have python call subprocess

PA-Rweins: Sure will do so.

Tests/Marketplace/zip_packs.py
avidan-H: Please implement @anara123's recommendation.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,84 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+def download_packs_from_gcp(storage_bucket_name, dest_path):
+    """
+    Copies all content in the gcp path to destination_path
+    Args:
+        storage_bucket_name: The name of the GCP bucket to download from.
+        dest_path: The path to download the packs to.
+    """
+    src_path = os.path.join('gs://', storage_bucket_name, GCPConfig.STORAGE_BASE_PATH)
+
+    process = subprocess.check_output(['gsutil', '-m', 'cp', '-r', src_path, dest_path])
```
avidan-H, line 174: Unless the service account has already been used to authenticate to Cloud Storage in a previous step in the build process... this won't work. We shouldn't make that assumption. We should leave passing the service account as a script parameter and perform authentication in the event that it is necessary.

Tests/Marketplace/zip_packs.py
```diff
@@ -212,12 +239,13 @@ def main():
...
-    service_account = option.service_account
-    circle_build = option.circle_build
-    branch_name = option.branch_name
-    gcp_path = option.gcp_path
+    storage_base_path = option.storage_base_path
     remove_test_playbooks = option.remove_test_playbooks
     private_build = option.private
+
+    if storage_base_path:
+        GCPConfig.STORAGE_BASE_PATH = storage_base_path
```
avidan-H, line 247: `STORAGE_BASE_PATH` is both an attribute of an import as well as being a constant, as indicated by it being capitalized. Please do _not_ assign to this variable.

Tests/Marketplace/zip_packs.py
```diff
@@ -1,20 +1,19 @@
...
-from Tests.Marketplace.marketplace_services import init_storage_client
 from Tests.Marketplace.marketplace_constants import IGNORED_FILES, PACKS_FULL_PATH
 from Tests.scripts.utils.log_util import install_logging
 from demisto_sdk.commands.common.tools import LooseVersion, str2bool
+from Tests.Marketplace.marketplace_constants import GCPConfig
+import subprocess
+from pathlib import Path
 
 ARTIFACT_NAME = 'content_marketplace_packs.zip'
 MAX_THREADS = 4
```
avidan-H, line 16: You can delete this global variable since you've removed the use of threads.
```suggestion
```

Tests/Marketplace/zip_packs.py
```diff
@@ -34,6 +33,8 @@ def option_handler():
     parser.add_argument('-b', '--bucket_name', help="Storage bucket name", required=True)
     parser.add_argument('-br', '--branch_name', help="Name of the branch", required=False)
     parser.add_argument('-n', '--circle_build', help="Number of the circle build", required=False)
```
avidan-H, line 35: You've removed the use of a couple of the script's arguments in the `main` function but didn't remove them from the `options_handler`. Please ensure these are synced.

avidan-H (changes requested):
Did you test these changes in GitLab? If not, then please do that as well. Once you've done that please attach the relevant builds to the PR (either in the description or as a new comment) so we can view the results.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,84 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+    Copies all content in the gcp path to destination_path
+    Args:
+        storage_bucket_name: The name of the GCP bucket to download from.
+        dest_path: The path to download the packs to.
+    """
+    src_path = os.path.join('gs://', storage_bucket_name, GCPConfig.STORAGE_BASE_PATH)
+
+    process = subprocess.check_output(['gsutil', '-m', 'cp', '-r', src_path, dest_path])
+    if process:
+        logging.info(f"cp_gcp_dir.sh output: {process}")
```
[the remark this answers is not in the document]
PA-Rweins: Fixed.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,61 +114,84 @@ def remove_test_playbooks_from_signatures(path, filenames):
...
+            for name, path in zip_pack.items():
+                logging.info(f"Copying pack from {path} to /home/runner/work/content-private/"
+                             f"content-private/content/artifacts/packs/{name}.zip")
+                shutil.copy(path,
+                            f'/home/runner/work/content-private/content-private/content/artifacts/'
+                            f'packs/{name}.zip')
+
+
 def executor_submit(executor, download_path, blob):
     executor.submit(blob.download_to_filename, download_path)
```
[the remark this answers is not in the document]
PA-Rweins: Fixed

Tests/Marketplace/zip_packs.py
PA-Rweins: > I don't see that you addressed @guyfreund's comment.

You can see at line 270 that I added a try except surrounding this function call to handle it's errors. Due to decision made by me and @teizenman to handle all excetpions in main.

Tests/Marketplace/zip_packs.py
```diff
@@ -195,15 +218,19 @@ def get_latest_pack_zip_from_blob(pack, blobs):
...
     """
-    blob = None
-    blobs = [b for b in blobs if os.path.splitext(os.path.basename(b.name))[0] == pack and b.name.endswith('.zip')]
-    if blobs:
-        blobs = sorted(blobs, key=lambda b: LooseVersion(os.path.basename(os.path.dirname(b.name))), reverse=True)
-        blob = blobs[0]
+    latest_blob_name = None
+    latest_blob_version = None
+    for current_blob in blobs:
+        current_pack_name = os.path.splitext(os.path.basename(current_blob))[0]
```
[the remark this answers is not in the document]
PA-Rweins: No, I actually changed the original function which used an actual blob object. It's actually just the name of a file in the pack. I will change the terminology of the function and update the documentation. Thanks.

Tests/Marketplace/zip_packs.py
```diff
@@ -195,15 +218,19 @@ def get_latest_pack_zip_from_blob(pack, blobs):
...
+        if current_pack_name == pack and current_blob.endswith('.zip'):
+            current_pack_zip_version = LooseVersion(os.path.basename(os.path.dirname(current_blob)))
```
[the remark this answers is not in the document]
PA-Rweins: Same as above

Tests/Marketplace/zip_packs.py
PA-Rweins: I'm was not aware of the service account. 
However, if it has already been authenticated in a previous step why should we authenticate again? What is the scenario where we'll need to re-authenticate? Is this step will ever be executed independently?

Tests/Marketplace/zip_packs.py
PA-Rweins: I actually copied this behaviour, as suggested by @guyfreund, from copy_and_upload_packs.py file.

Do you have another solution? Can I use this practice as it is already being used in other files?
 
Screen Shot 2021-06-15 at 14 23 29

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed.

Tests/Marketplace/zip_packs.py
PA-Rweins: Fixed.

Tests/Marketplace/zip_packs.py
PA-Rweins: > Please implement @anara123's recommendation.

Have you seen my suggestion?
This function is used only when it's a private build, and this path is specifically for the case of a private build.
Is it needed to expose this argument as an input if this flow happens just for a specific use case anyway? (private build)

Tests/Marketplace/zip_packs.py
avidan-H: This is bad practice and I would avoid doing this if there is a viable alternative.

Tests/Marketplace/zip_packs.py
```diff
@@ -212,12 +229,13 @@ def main():
     storage_bucket_name = option.bucket_name
     zip_path = option.zip_path
     artifacts_path = option.artifacts_path
```
avidan-H, line 231: I don't see where you are using this script argument. You'll want to use it since the artifacts path is different for GitLab and CircleCI.

PA-Rweins: It's used in lines 286 and 288

Tests/Marketplace/zip_packs.py
PA-Rweins: Sure, fixed it.

PA-Rweins:
> @PA-Rweins please also follow up regarding my comment [here](https://github.com/demisto/content/pull/12770#pullrequestreview-683709191).

https://code.pan.run/xsoar/content/-/jobs/6085731

You can view the results of the step in gitlab under upload-to-marketplace.

.gitlab/ci/bucket-upload.yml
```diff
@@ -148,6 +148,21 @@ upload-packs-to-marketplace:
         python3 ./Tests/Marketplace/copy_and_upload_packs.py -a $PACK_ARTIFACTS -e $EXTRACT_FOLDER -pb "$GCS_MARKET_BUCKET" -bb "$GCS_BUILD_BUCKET" -s $GCS_MARKET_KEY -n $CI_PIPELINE_ID -c $CI_COMMIT_BRANCH -pbp "$STORAGE_BASE_PATH"
       fi
     - section_end "Upload Packs To Marketplace Storage"
+    - section_start "Zip Content Packs From GCS"
+    - |
+      if [ -n "${BUCKET_UPLOAD}" ]; then
+        if [[ $CI_COMMIT_BRANCH != master ]] && [[ $CI_COMMIT_BRANCH != 21\.* ]] && [[ $CI_COMMIT_BRANCH != 22\.* ]]; then
```
avidan-H, line 154: Let's change the conditions to conform to the others - let's switch the conditions to be positive. So you'll need to rearrange the sections accordingly.

.gitlab/ci/bucket-upload.yml
```diff
@@ -148,6 +148,21 @@ upload-packs-to-marketplace:
         python3 ./Tests/Marketplace/copy_and_upload_packs.py -a $PACK_ARTIFACTS -e $EXTRACT_FOLDER -pb "$GCS_MARKET_BUCKET" -bb "$GCS_BUILD_BUCKET" -s $GCS_MARKET_KEY -n $CI_PIPELINE_ID -c $CI_COMMIT_BRANCH -pbp "$STORAGE_BASE_PATH"
       fi
     - section_end "Upload Packs To Marketplace Storage"
+    - section_start "Zip Content Packs From GCS"
+    - |
+      if [ -n "${BUCKET_UPLOAD}" ]; then
```
avidan-H, line 153: No need for this condition since this job only executes if `BUCKET_UPLOAD` is set to `true` anyways.

.gitlab/ci/bucket-upload.yml
```diff
@@ -148,6 +148,21 @@ upload-packs-to-marketplace:
...
+        if [[ $CI_COMMIT_BRANCH != master ]] && [[ $CI_COMMIT_BRANCH != 21\.* ]] && [[ $CI_COMMIT_BRANCH != 22\.* ]]; then
+          echo "Skipping packs download to artifact on non master or release branch"
+          exit 0
```
avidan-H, line 156: Let's not manually exit here - I want the remaining commands in the job to run, even if it's only to say that they are skipping steps.

Tests/Marketplace/Tests/zip_packs_test.py
```diff
@@ -102,10 +102,11 @@ def test_copy_zipped_packs_to_artifacts(self, mocker):
         """
         import shutil
         zipped_packs = [{'Slack': 'content/packs/Slack/1.0.1/Slack.zip'}]
+        artifatcs_path = 'dummy_path'
         mocker.patch.object(shutil, 'copy', side_effect=None)
         mocker.patch('os.path.exists', return_value=True)
 
-        copy_zipped_packs_to_artifacts(zipped_packs)
+        copy_zipped_packs_to_artifacts(zipped_packs, artifatcs_path)
```
avidan-H, line 109: You have a typo in the variable name - it should be `artifacts_path`.

avidan-H (changes requested):

Tests/Marketplace/Tests/zip_packs_test.py
PA-Rweins: thanks, fixed it.

.gitlab/ci/bucket-upload.yml
PA-Rweins: Ok, removed it.

.gitlab/ci/bucket-upload.yml
PA-Rweins: Fixed it.

.gitlab/ci/bucket-upload.yml
PA-Rweins: I changed the conditions to be positive, but what do you mean by rearranging the sections? what sections?

.gitlab/ci/on-push.yml
```diff
@@ -78,16 +78,6 @@ create-instances:
...
-    - section_start "Zip Marketplace Packs"
-    - |
-      if [[ $CI_COMMIT_BRANCH != master ]] && [[ $CI_COMMIT_BRANCH != 21\.* ]] && [[ $CI_COMMIT_BRANCH != 22\.* ]]; then
-        echo "Skipping packs download to artifact on branch that is not master or a release branch"
-        echo "CI_COMMIT_BRANCH=$CI_COMMIT_BRANCH"
-      else
-        ZIP_FOLDER=$(mktemp -d)
-        python3 ./Tests/Marketplace/zip_packs.py -b 'marketplace-dist' -z $ZIP_FOLDER -a $ARTIFACTS_FOLDER -s $GCS_MARKET_KEY
-      fi
-    - section_end "Zip Marketplace Packs"
```
avidan-H, line 90: Can leave this here and just use the condition of `if [[ -n "${BUCKET_UPLOAD}" ]]; then` for when to this should run since the `create-instances` job in the `bucket-upload` flow inherits from this job. Therefore there is no need to add this step in the `upload-packs-to-marketplace` job.

Tests/Marketplace/zip_packs.py
```diff
@@ -114,63 +109,80 @@ def remove_test_playbooks_from_signatures(path, filenames):
         logging.warning(f'Could not find signatures in the pack {os.path.basename(os.path.dirname(path))}')
 
 
-def download_packs_from_gcp(storage_bucket, gcp_path, destination_path, circle_build, branch_name):
+def get_pack_files(entry_path):
```
avidan-H, line 112: It's hard for me to imagine that we don't have this functionality implemented elsewhere that you can import.

avidan-H (changes requested):

.circleci/config.yml
```diff
@@ -1101,6 +1090,22 @@ jobs:
...
 
                   python3 ./Tests/Marketplace/copy_and_upload_packs.py -a $PACK_ARTIFACTS -e $EXTRACT_FOLDER -pb "$GCS_MARKET_BUCKET" -bb "$GCS_BUILD_BUCKET" -s $GCS_MARKET_KEY -n $CI_PIPELINE_ID -c $CIRCLE_BRANCH -pbp "$STORAGE_BASE_PATH"
+            - run:
+                name: Zip Content Packs From GCS
+                command: |
+                  if [ -n "${BUCKET_UPLOAD}" ]; then
+                    if [[ $CIRCLE_BRANCH != master ]] && [[ $CIRCLE_BRANCH != 20\.* ]] && [[ $CIRCLE_BRANCH != 21\.* ]]; then
+                      echo "Skipping packs download to artifact on non master or release branch"
+                      exit 0
+                    fi
```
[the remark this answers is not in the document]
PA-Rweins: Why? I added the bucket upload because we want it to happen only during upload flow and the second condition for making sure it's master

.gitlab/ci/on-push.yml
PA-Rweins: So I should remove the entire _Zip Content Packs From GCS_ from `bucket-upload.yml` ?

Tests/Marketplace/zip_packs.py
PA-Rweins: Ok, where should I look for it or how is it called?

.circleci/config.yml
```diff
@@ -1101,6 +1090,22 @@ jobs:
                   fi
 
                   python3 ./Tests/Marketplace/copy_and_upload_packs.py -a $PACK_ARTIFACTS -e $EXTRACT_FOLDER -pb "$GCS_MARKET_BUCKET" -bb "$GCS_BUILD_BUCKET" -s $GCS_MARKET_KEY -n $CI_PIPELINE_ID -c $CIRCLE_BRANCH -pbp "$STORAGE_BASE_PATH"
+            - run:
+                name: Zip Content Packs From GCS
+                command: |
+                  if [ -n "${BUCKET_UPLOAD}" ]; then
```
guyfreund, line 1096:
```suggestion
```
no need for that, this stage is being performed on upload-flow only

.gitlab/ci/bucket-upload.yml
```diff
@@ -148,6 +148,19 @@ upload-packs-to-marketplace:
         python3 ./Tests/Marketplace/copy_and_upload_packs.py -a $PACK_ARTIFACTS -e $EXTRACT_FOLDER -pb "$GCS_MARKET_BUCKET" -bb "$GCS_BUILD_BUCKET" -s $GCS_MARKET_KEY -n $CI_PIPELINE_ID -c $CI_COMMIT_BRANCH -pbp "$STORAGE_BASE_PATH"
       fi
     - section_end "Upload Packs To Marketplace Storage"
+    - section_start "Zip Content Packs From GCS"
+    - |
+      if [ -n "${BUCKET_UPLOAD}" ]; then
+        if [[ $CI_COMMIT_BRANCH == master ]] || [[ $CI_COMMIT_BRANCH == 21\.* ]] || [[ $CI_COMMIT_BRANCH == 22\.* ]]; then
```
guyfreund, line 154: why do we need to check for `BUCKET_UPLOAD`?

.gitlab/ci/bucket-upload.yml
```diff
@@ -148,6 +148,19 @@ upload-packs-to-marketplace:
...
+          if [[ $GCS_MARKET_BUCKET != $GCS_PRODUCTION_BUCKET ]]; then
+            STORAGE_BASE_PATH="upload-flow/builds/$CI_COMMIT_BRANCH/$CI_PIPELINE_ID/content/packs"
+          fi
```
guyfreund, line 157: why do we do this on non-production runs?

guyfreund (changes requested):
Not sure why we download from the build bucket instead of production bucket, as we will run as part of the upload.

.circleci/config.yml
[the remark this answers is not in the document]
PA-Rweins: Wil remove and will add conditions from 1103-1105 in config.yml. after upload step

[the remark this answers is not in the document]
PA-Rweins: fixed

yaakovi (approved):

PR #12797, opened 2021-05-23
Edl 500 error bug

MosheEichler:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/36876

## Description
change in the ***is_demisto_version_ge*** function to compare the build_number as two numbers instead of two strings
## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -1203,6 +1203,26 @@ def test_is_demisto_version_build_ge(mocker):
...
+    """
+    mocker.patch.object(
+        demisto,
+        'demistoVersion',
+        return_value={
+            'version': '6.1.0',
+            'buildNumber': '999999'
+        }
+    )
+    assert not is_demisto_version_ge('6.1.0', build_number='1000000')
```
reutshal, line 1223: Why not add to the previous UT an assertion like this:

assert not is_demisto_version_ge('6.0.0', '400000')

reutshal (changes requested):
please see my comment regarding the test

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
MosheEichler: I remove this function and I added to the previous function, this case ***is_demisto_version_ge('6.0.0', '6')***, That in strings comparison it's bigger than build_number '500000' but in integers comparison it's lower

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -1201,6 +1201,7 @@ def test_is_demisto_version_build_ge(mocker):
     assert not is_demisto_version_ge('6.0.0', '50001')
     assert not is_demisto_version_ge('6.1.0', '49999')
     assert not is_demisto_version_ge('5.5.0', '50001')
+    assert is_demisto_version_ge('6.0.0', '6')
```
reutshal, line 1204: Please move to line 1201 to keep consistency.
I would also add a comment explaining this assertion, something like - 'Added with the fix of __link to the issue__ '

reutshal (approved):

PR #12815, opened 2021-05-24
Remove DEMISTO_SDK_NIGHTLY env var

avidan-H:
## Status
- [x] Ready

## Description
The `dmst-content-team` slack channel received this [message](https://panw-global.slack.com/archives/G011E63JXPB/p1621805925116800) about the `SDK Nightly Unit Tests`. But as you can see in the GitLab [pipeline](https://code.pan.run/xsoar/content/-/pipelines/1170205) it links to, this was actually a scheduled run for the `NIGHTLY` build (check the build parameters section of the [create-instances job](https://code.pan.run/xsoar/content/-/jobs/5705693). I removed the env variable from the `demisto_sdk_nightly` jobs that was causing the `Demisto SDK Nightly`-related jobs to run on every scheduled run (nightly, instance-testing, etc.). Now the `DEMISTO_SDK_NIGHTLY` environment variable isn't set by the `Demisto SDK Nightly`-related job configurations but rather only by it being passed via the scheduled trigger.

## Screenshots
Can see in the attached screenshot that the `Demisto SDK Nightly`-related jobs ran in the pipeline in addition to the `create-instances` and `server_master` jobs that ran for the scheduled nightly (which we don't want to happen).
image

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/sdk-nightly.yml
+++ b/.gitlab/ci/sdk-nightly.yml
@@ -4,12 +4,14 @@
 
 
 .upload-entities-to-cortex-xsoar: &upload-entities-to-cortex-xsoar
+  - section_start "Upload Entities to Cortex XSOAR" --collapsed
   - demisto-sdk upload -i Packs/HelloWorld/Integrations/ --insecure
   - demisto-sdk upload -i Packs/HelloWorld/TestPlaybooks/playbook-HelloWorld-Test.yml --insecure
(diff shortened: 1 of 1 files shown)


avidan-H:

Itay4 (approved):

PR #12855, opened 2021-05-25
GraphQL - add the fetch_schema_from_transport param

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37390

   - [x] No

## Must have
- [x] Documentation


Packs/GraphQL/Integrations/GraphQL/GraphQL.py
```diff
@@ -70,7 +70,10 @@ def main() -> None:
 
         transport = RequestsHTTPTransport(**request_params)
         handle_proxy()
-        client = Client(transport=transport, fetch_schema_from_transport=True)
+        client = Client(
+            transport=transport,
+            fetch_schema_from_transport=params.get('fetch_schema_from_transport', True),
```
glicht, line 75: you can get here None after upgrade as the server will put there a null after upgrade. See here: https://github.com/demisto/etc/issues/37415

Itay4: added handling for it

Packs/GraphQL/Integrations/GraphQL/README.md
```diff
@@ -131,3 +132,5 @@ Execute a mutation request to the GraphQL server.
 >|---|
 >| reaction: {"content": "HOORAY"}<br/>subject: {"id": "MDU6SXNzdWUyMzEzOTE1NTE="} |
 
+## Troubleshooting
+  - If you are encountering the error `GraphQLError: Cannot query field`, uncheck the **Fetch the schema from the transport** integration parameter.
```
glicht, line 136:
```suggestion
  - If you are encountering the error `GraphQLError: Cannot query field`, you may be failing because of a schema validation error. Uncheck the **Fetch the schema from the transport** integration parameter to disable the schema validation.
```

glicht (approved):
Let's just also test in the test PB this param set to false.

PR #12866, opened 2021-05-26
Various GitLab Tweaks & Improvements

avidan-H:
## Status
- [x] Ready

## Description
Includes small tweaks and fixes across various GitLab things, including -
- Updates the upload flow's test trigger script's usage documentation
- Stores artifacts of jobs in the upload flow _even_ when they fail
- Assign command execution code to an `EXIT_CODE` variable so that the job's `script` configuration continues. For example, the `failed_tests.txt` would still be created even if tests fail during test execution (e.g. for the nightly build when the run tests step almost always fails because there is _some_ test playbook that inevitably fails)
- Deduplicate code in upload flow
- Update `demisto-py` dependency to the latest version (`3.0.1`) which accounts for if the `DEMISTO_CONNECTION_POOL_MAXSIZE` variable is set in the environment
- Move GitLab main config file under the `.gitlab/ci/` directory to be together with the rest of the GitLab configuration files

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Itay4 (approved):

avidan-H:
Adding the `docs-approved` label as there isn't any documentation that requires review by a technical writer.

Itay4:
as expected, gitlab pipeline is not running due to changes in this pr
merging

PR #12870, opened 2021-05-26
fix bug where is_demisto_version_ge failed on dev env

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37433

## Must Have:
- [x] Tests


Itay4 (approved):

glicht (commented):
I noticed that the logic regarding comparing build number is flawed. We should only compare builld number if the server versions are equal. @Itay4 FYI

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6268,9 +6268,17 @@ def is_demisto_version_ge(version, build_number=''):
...
+        if server_version.get('version') > version:
+            return True
+        elif server_version.get('version') == version:
+            if build_number:
+                return int(server_version.get('buildNumber')) >= int(build_number)
+            return True  # No build number
+        else:
+            return False
+    except ValueError:
+        # dev editions are not comparable
```
glicht, line 6280: Let's add an info log here. It should only happen on dev, so if we do reach this flow it is worth logging.

PR #12877, opened 2021-05-26
JWT New Integration

ayman-m:

## Status
- [x] In Progress
- [ ] Ready

## Description
JSON Web Token (JWT) is a compact, URL-safe means of representing claims to be transferred between two parties. This Integration can be used to Generate New JWT Tokens, Encode and Decode Existing Ones.

- [ ] 5.5.0
- [x ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ x] No

## Must have
- [ ] Tests
- [x ] Documentation


ayman-m:
Hi @adi88d, just wanted to check if you had a chance to review this PR, the build failure is because a random JWT token used as an example of the command output, nothing secret in this token

Packs/JWT/Integrations/JWT/JWT.py
```diff
@@ -0,0 +1,188 @@
...
+    payload = jwt.encode(claims, secret_key, algorithm='HS256')
+
+    return jti, payload
+
+
+''' COMMAND FUNCTIONS '''
+
+
+def test_module():
+    return 'ok'
```
adi88d, line 73: Why test_module is not checking anything?

Packs/JWT/Integrations/JWT/JWT.py
```diff
@@ -0,0 +1,188 @@
...
+        outputs=result,
+    )
+
+
+''' MAIN FUNCTION '''
+
+
+def main():
+
+    base_url = urljoin(demisto.params()['url'])
```
adi88d, line 154:
```suggestion
    base_url = demisto.params().get('url')
```

Packs/JWT/Integrations/JWT/JWT.yml
```diff
@@ -0,0 +1,243 @@
+category: Authentication
+commonfields:
+  id: JWT
+  version: -1
+configuration:
+- additionalinfo: The authentication server URL
```
adi88d, line 6:
```suggestion
- additionalinfo: The authentication server URL for example: https://example.com/auth/v2/token
```

maybe its better to add more information about the url

Packs/JWT/Integrations/JWT/JWT.yml
```diff
@@ -0,0 +1,243 @@
...
+  display: Your server URL
+  name: url
+  required: true
+  type: 0
+- additionalinfo: The secret key to sign the authentication token.
+  display: Secret Key
+  hidden: false
+  name: key
+  required: true
+  type: 0
```
adi88d, line 17:
```suggestion
  type: 4
```
sensitive parameters should be encrypted

Packs/JWT/Integrations/JWT/JWT.py
```diff
@@ -0,0 +1,188 @@
...
+    if sub:
+        claims['sub'] = sub
+    if scp:
+        claims['iss'] = scp
+    if scp:
+        claims['nbf'] = nbf
+    if additional_claims:
+        claims.update(json.loads(additional_claims))
+
+    payload = jwt.encode(claims, secret_key, algorithm='HS256')
```
adi88d, line 64:
```suggestion
    payload = jwt.encode(claims, secret_key, algorithm='HS256')
```
In my opinion its better to pass the algorithm as command arg instead of using hardcoded value of `HS256`
like `ES256 ` see the [documentation](https://pyjwt.readthedocs.io/en/latest/api.html)
but this is not a must

adi88d (changes requested):
Nice!

please remove the secrets from the files like access token and authentication token
and add `https://jwt.io` to `.secrets-ignore` file

Packs/JWT/Integrations/JWT/JWT.py
ayman-m: JWT has different set of claims (jti,iss,aud, ... and unlimited number of additional claims) that largely vary with different authentication services, some services that are to be used with this integration to generate JWT tokens may expect claims in the authentication payload that are not defined in the integration parameters which would result in a test failure because of missing claims, I can create additional parameters for all the RFC claims and have one parameter that accept additional (non RFC claims) as a JSON object, in that case we can have some testing login since we have all the claims needed for the test . will push another commit to reflect that

Packs/JWT/Integrations/JWT/JWT.yml
ayman-m: It's any JWT authentication service URL, not specific to any product, that why I kept it generic

Packs/JWT/Integrations/JWT/JWT.yml
ayman-m: noted, will fix that

ayman-m:
Thanks! will update .secrets-ignore and remove the tokens

Packs/JWT/Integrations/JWT/JWT.py
adi88d: nice!
please update the README file too with the new params and arguments you added(Scope,Issued At, Algorithm...)

Packs/JWT/Integrations/JWT/JWT.py
adi88d: Got it
I think in that case we can just returns 'ok' in the test, because now when user wants to configure the integration he need to populate non mandatory params like Issuer and Subject in order to pass the test.

adi88d (approved):

PR #12884, opened 2021-05-27
Splunk raw labels fix

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35176

## Description
Incident labels were not properly processed when the word message was in the raw data.
Introduced a fix for it.

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -1391,6 +1391,8 @@ def rawToDict(raw):
             key_val_arr = raw.split(",")
             for key_val in key_val_arr:
                 single_key_val = key_val.split(":", 1)
+                if len(single_key_val) <= 1:
+                    single_key_val = key_val.split("=", 1)
                 if len(single_key_val) > 1:
                     val = single_key_val[1]
                     key = single_key_val[0].strip()
--- a/Packs/SplunkPy/ReleaseNotes/2_1_2.md
+++ b/Packs/SplunkPy/ReleaseNotes/2_1_2.md
@@ -1,4 +1,4 @@
 
 #### Integrations
 ##### SplunkPy
-- Fixing the way labels are parsed from notable events' raw data.
+- Fixed the way labels are parsed from the raw data of a notable event.
--- a/Packs/SplunkPy/ReleaseNotes/2_1_2.md
+++ b/Packs/SplunkPy/ReleaseNotes/2_1_2.md
@@ -1,0 +1,4 @@
+
+#### Integrations
+##### SplunkPy
+- Fixing the way labels are parsed from notable events' raw data.
--- a/Packs/SplunkPy/pack_metadata.json
+++ b/Packs/SplunkPy/pack_metadata.json
@@ -2,7 +2,7 @@
     "name": "Splunk",
     "description": "Run queries on Splunk servers.",
     "support": "xsoar",
-    "currentVersion": "2.1.1",
+    "currentVersion": "2.1.2",
     "author": "Cortex XSOAR",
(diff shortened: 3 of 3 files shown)


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -1391,6 +1391,8 @@ def rawToDict(raw):
             key_val_arr = raw.split(",")
             for key_val in key_val_arr:
                 single_key_val = key_val.split(":", 1)
+                if len(single_key_val) <= 1:
+                    single_key_val = key_val.split("=", 1)
                 if len(single_key_val) > 1:
```
esharf, line 1396:
```suggestion
               else:
```

esharf (approved):

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
guykeller: This is not a valid change - notice that in the `if` we change the value of `single_key_val`

Packs/SplunkPy/ReleaseNotes/2_1_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### SplunkPy
+- Fixing the way labels are parsed from notable events' raw data.
```
amshamah419, line 4:
```suggestion
- Fixing the way labels are parsed from the raw data of a notable event.
```

alujones:
As the person who originally reported this problem, I don't see this as a complete, or correct, fix.
Why does the word "message" anywhere in the _raw trigger this parsing route? Note that "message" could be in the name of an asset, or in a string describing an incident parameter, or any of a number of different text strings that should not be given meaning in the context of parsing this.
What you have in this code looks like it's a python embodiment of the following pseudocode:

1. Try to parse raw as JSON. If good, end.
2. If it's not JSON, but the word "message" is anywhere:
2.1. Try to parse it again as JSON, but manually instead of using json.loads, this time.
2.2. That didn't work? Try and parse as "key=value", kind of.
3. Try and parse as "key=value".

Now, I don't understand why the word "message" is some kind of trigger here - maybe you have another customer with a need for this.

Here are some example strings that will show you how this will fail:
```
>>> rawToDict("key1=\"filter1\"")
{'key1': 'filter1'}
>>> rawToDict("key1=\"filter1\",key2=\"filter2\"")
{'key1': 'filter1', 'key2': 'filter2'}
>>> rawToDict("key1=\"filter1\",key2=\"filter2 message for you\"")
{'key1': 'filter1', 'key2': 'filter2 message for you'}
>>> rawToDict("key1=\"filt:er1\",key2=\"filter2 message for you\"")
{'key1=filt': 'er1', 'key2': 'filter2 message for you'}
>>> rawToDict("key1=\"filt:er1\",key2=\"filter2 mesage for you\"")
{'key1': 'filt:er1', 'key2': 'filter2 mesage for you'}
```
Numbering the calls 1 to 5, here's an annotated list of what's happening:
1. This is just showing that a single key assignment works.
2. An assignment of two keys works.
3. If I put the word "message" in, we're taking the "message" path, even though the word is in a text value. But it still works.
4. Here, it breaks - if a value contains a colon, that will be preferred over the equals sign. Because of the word "message".
5. The same input, with the word "message" changed to "mesage" (with one S), now works as we expect it to.

I've asked this question of the support team, but I don't think they passed it on:
Why does the "if 'message' in raw:" block of code exist in the first place? It seems that this could be completely removed.
At the moment, I'm not generating any inputs that cause this to fail, which is why I've allowed tech support to close this incident. I just don't think the code is robust or correct, and the fact that I can quickly provide you with an example of an innocuous input that triggers a parsing failure shows that.

PR #12919, opened 2021-05-30
Slack notify as child pipeline

avidan-H:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Changes the slack notification system to run in a child pipeline. We previously had set it up to run in the `after_script` section of various jobs of the pipelines of which we are interested in the run status. There were a number of problems with that.
- The `after_script` section ran in a clean bash environment which required us to rerun all of the setup just so we could run the slack notification script. This resulted in us duplicating the overhead of any setup for every job in a pipeline.
- Having the slack notifier run in the `after_script` section of a job of the pipeline it was to report on would result in incorrect reporting. This is because when checking the pipeline status, since the slack notifier was from a _running_ job in the pipeline the status of the pipeline would be `running` instead of `failure` or `success`. This is also why it wouldn't really help to run the slack notification from its own job within the pipeline either.
- If for some reason a preceding job would fail, causing subsequent jobs to get skipped, this could cause the job in which the slack notifier was set to run in the `after_script` section of not to run at all if it happened to be one of those skipped jobs.

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


.gitlab/ci/on-push.yml
```diff
@@ -84,8 +84,8 @@ create-instances:
     - .push-rule
   needs: ["create-instances"]
   stage: run-instances
-  variables:
-    DESTROY_INSTANCES: "true"
+  # variables:
+  #   DESTROY_INSTANCES: "true"
```
Itay4, line 88:
```suggestion
```

Itay4 (approved):

avidan-H:
I'm adding the `docs-approved` label since there isn't any documentation requiring review in this PR.

PR #12937, opened 2021-05-31
Fix condition to only apply during bucket upload flow

avidan-H:
## Status
- [x] Ready

## Related Issues
Can see an example of the validations not running in this `demisto-sdk` nightly GitLab run (where they should have) - https://code.pan.run/xsoar/content/-/jobs/5807340#queue-placeholder/containers/0

## Description
The `.run-validation` configuration which job configurations in various workflows inherit from contains an incorrect condition causing validations to not run. The condition prevented the validations from running during the current overlap period where we have the bucket upload flow in GitLab uploading to a test bucket.

## Screenshots
image
> example of the validations not running in the `Demisto SDK Nightly` flow in GitLab (where they should have run).

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [x] ~~Documentation~~ N/A

### DIFF
--- a/.gitlab/ci/global.yml
+++ b/.gitlab/ci/global.yml
@@ -208,7 +208,7 @@
     - section_end "Update Tests"
     - section_start "Validate Files and Yaml"
     - |
-      if [[ "$(echo "$GCS_MARKET_BUCKET" | tr '[:upper:]' '[:lower:]')" != "marketplace-dist" ]]; then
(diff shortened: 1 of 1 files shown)


avidan-H:
Adding the `docs-approved` label since there is not any documentation that requires review.

avidan-H:
@roysagi fyi - fix for what you pointed out to me yesterday.

Itay4 (approved):

PR #12042, opened 2021-04-03
added fetch delta as a integration parameter

orhovy:
Added the fetch delta as a param

## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/35117
https://github.com/demisto/etc/issues/35375

## Description
Added fetch delta

## Screenshots
Paste here any images that will help the reviewer

## Minimum version of Demisto
- [ ] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

PR #13020, opened 2021-06-04
Proofpoint Threat Response - set first fetch param default value by or operator

Itay4:
## Status
- [x] Ready

fixes: https://github.com/demisto/etc/issues/37757

## Description
in https://github.com/demisto/content/pull/12042 we added the first fetch param
integration instances that were configured before that change now fail with:
```
   File "<string>", line 19, in <module>
   File "<string>", line 6314, in parse_date_range
 AttributeError: 'NoneType' object has no attribute 'strip'
```
thats because the param exist in the params dict, but with value of `None` 
setting the default by using the `or` operator

setting default to `12 hours` according to the default in the yml file

   - [x] No


bakatzir (approved):
please add the desc of default value in the integration param in the yml

Itay4:
> please add the desc of default value in the integration param in the yml

add it where?

orhovy (approved):

yaakovi (approved):

PR #13021, opened 2021-06-04
CommonServerPython - added service_account to sensitive params

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37747

   - [x] No

### DIFF
--- a/Packs/Base/ReleaseNotes/1_12_2.md
+++ b/Packs/Base/ReleaseNotes/1_12_2.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Improved secrets mechanism detection of log prints.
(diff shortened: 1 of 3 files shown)


djmasters:
In the long term is there any way we could look to change this to check if the parameter type is listed as encrypted? I can foresee us having to curate (and document) the sensitive params list otherwise.

glicht:
@djmasters we can continue the discussion on etc

glicht (approved):

Packs/Base/ReleaseNotes/1_12_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Improved secrets mechanism detection of log prints.
```
bakatzir, line 4:
```suggestion
- Improved secrets mechanism detection and obfuscation of log prints.
```

PR #13046, opened 2021-06-06
CommonServer: remove None values

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Remove None values from params to avoid issues as seen at: https://github.com/demisto/etc/issues/37415

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1177,6 +1177,32 @@ def empty(x):
...
+    res = []
+    for key in list_of_keys:
+        if data[key] is None:
+            res.append(key)
+            del data[key]
+    return res
+
+
+if (not os.getenv('COMMON_SERVER_NO_AUTO_PARAMS_REMOVE_NULLS')) and hasattr(demisto, 'params'):
+    res = remove_none_elements(demisto.params())
```
anara123, line 1201: this can break BC, no?
If the integration has code like `demisto.params()['status']`. So instead of returning None, it will fail on KeyError.

glicht: Will try a different approach.

glicht: New approach. See update

anara123 (approved):

PR #13079, opened 2021-06-08
new test for Cortex XDR - IOC integration

esharf:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25598

## Description
Probably the problem will not appear in case that proxy will be used.

### DIFF
--- a/Packs/CortexXDR/TestPlaybooks/Coretx_XDR_-_IOC_-_Test_without_fetch.yml
+++ b/Packs/CortexXDR/TestPlaybooks/Coretx_XDR_-_IOC_-_Test_without_fetch.yml
@@ -1,0 +1,408 @@
+id: Cortex XDR - IOC - Test without fetch
+version: -1
+name: Cortex XDR - IOC - Test without fetch
+starttaskid: "0"
+tasks:
+  "0":
+    id: "0"
+    taskid: b290fe9a-86bc-411f-8227-c45dcea1fcd6
+    type: start
+    task:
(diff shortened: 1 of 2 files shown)


Packs/CortexXDR/.pack-ignore
```diff
@@ -26,4 +26,7 @@ ignore=PB110
 ignore=PB110
 
 [file:PaloAltoNetworks_Cortex_XDR_Incident_Sync.yml]
-ignore=PB110
\ No newline at end of file
+ignore=PB110
+
+[file:XDR_iocs_-_Test.yml]
```
yaakovi, line 31: why are we keeping it? you can delete the old test

esharf: Because if we will have a way to run `fetch-indicator` (e.g. like fetch-incidents).
this test will be really useful.

yaakovi: in that case:
* open a dev-task for it
* remove it from the ignore
* add it to the skipped tests and link to the issue

Packs/CortexXDR/.pack-ignore
```diff
@@ -28,6 +28,7 @@ ignore=PB110
 [file:PaloAltoNetworks_Cortex_XDR_Incident_Sync.yml]
 ignore=PB110
 
+<<<<<<< HEAD
```
yaakovi, line 31:
```suggestion
```

yaakovi (approved):

PR #13116, opened 2021-06-09
Adding nginx support for export indicators

roysagi:

## Status
- [x] In Progress
- [x] Ready
- [ ] In Hold - waiting for customer conformation

## Related Issues
fixes: https://github.com/demisto/etc/issues/37619

## Description
Adding nginx support for export indicators.

## Screenshots
@DeanArbel and I have tested the performance of the integration by running the integration on a rich env with a limit of 20K indicators.

 - First run - with engine took ~50 sec on the code and the rest took about 5 sec
 
image

- Second run - everything comes from cache took less then a sec all around

image


glicht:
@roysagi let's add the unit tests regarding nginx. They will probably be the same as EDL.

adambaumeister:
is it nessecary to lift NGINX entirely into this integration?

It seems to me the perf issue here is caused because the entire result of the indicator query is stored in the integration context. 

https://github.com/demisto/content/blob/04585b29dd550682e7945907cedfc7eb7be49b84/Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py#L311-L329

But this is not required. The only time `current_iocs` is actually used is by `get_ioc_values_str_from_context` which simply returns the raw values anyway. 

I suspect if you change this to cache only the values instead of the indicator data this problem will resolve or at least improve dramatically without the addition of NGINX.

glicht:
@adambaumeister you are correct. We shouldn't store in the Context.

DeanArbel:
@adambaumeister you are correct, and the use of the `integration_context` can be removed following the implementation NGINX, however, the integration will still need to use the full indicators data, as there are formats such as CSV and JSON that return this full data back.

adambaumeister:
I did some testing on this. By removing the IOCs from the integration context this completely fixes the issue. Below testing uses 4 curl queries to the ExportIndicators service.

**Before fixing; indicator query returns 16k entries**
&nbsp; | time_namelookup | time_appconnect | time_pretransfer | time_redirect | time_starttransfer | time_total
-- | -- | -- | -- | -- | -- | --
1st | 0.004s | 0.000s | 0.004s | 0 | 142.008s | 142.008
2nd | 0.004s | 0.000s | 0.004s | 0 | 27.106s | 27.107
3rd | 0.004s | 0.000s | 0.004s | 0 | 48.044s | 48.045
4th | 0.004s | 0.000s | 0.004s | 0 | 27.081s | 27.082

**Post fix; same query, same server, same number of indicators**
&nbsp; | time_namelookup | time_appconnect | time_pretransfer | time_redirect | time_starttransfer | time_total
-- | -- | -- | -- | -- | -- | --
1st | 0.004s | 0.000s | 0.004s | 0 | 13.128s | 13.129
2nd | 0.004s | 0.000s | 0.004s | 0 | 0.068s | 0.069
3rd | 0.004s | 0.000s | 0.004s | 0 | 0.036s | 0.036
4th | 0.004s | 0.000s | 0.004s | 0 | 0.036s | 0.036


Now, you are correct, that we need the IOC data for JSON parsing. However, we don't need it every time. We only need it when the indicator query changes. In this way the current code "double dips" - the IOCs are stored in the context, AND the string json result is ALSO stored (in `dmst_export_iocs_values`). However - we can simply re-query the database in the instance the query changes, as this is insignificant and ephermeral (the IOC data is loaded into memory once, then discarded, until it changes again).

Take a look at PoC code here: https://github.com/adambaumeister/content/blob/ExportIndicatorsPerf/Packs/ExportIndicators/Integrations/ExportIndicators/ExportIndicators.py

Let me know if you want any more information.

roysagi:
@adambaumeister - thanks for the input. 
I've looked at the EDL integration and have consulted with @DeanArbel about it. We came to a conclusion that the integration context caching is only really required for on-demand mode run (the same is done in EDL) so I've changed the caching to only run through the `eis-update` command.

Tested it on the env and it seems to work with all queries and list sizes as intended.
Will now split the NGINX to an API module and this PR should be good to go.

roysagi:
Using pre existing documentation - adding docs approved.

Packs/ApiModules/Scripts/NGINXApiModule/NGINXApiModule.py
```diff
@@ -0,0 +1,305 @@
+
```
DeanArbel, line 1: @roysagi can you add it to EIS and EDL and make sure they work ok with the API module?

roysagi: I have added this to EIS but I think we should add it to EDL in a different PR in order to keep this PR slim for the bug fix.

DeanArbel (approved):
Looks good

PR #13134, opened 2021-06-10
Tldextract update

barchen1:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37003

## Description
In tldextract 3.0 they did a breaking change. See: https://github.com/john-kurkowski/tldextract/blob/master/CHANGELOG.md#300rc1-2020-10-12
We are using this library in FetchIndicatorsFromFile script  so I updated it to work with the latest tldextract

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/FetchIndicatorsFromFile/ReleaseNotes/1_0_3.md
```diff
@@ -0,0 +1,7 @@
+
+#### Scripts
+##### FetchIndicatorsFromFile
+- Maintenance and stability enhancements.
```
guykeller, line 4: Let's change this, should be avoided as much as possible @barchen1 @bakatzir

guykeller (approved):
Let's add the complete reasoning (as you've explained to me) for this change to the description of this PR, so it will be trackable

PR #13137, opened 2021-06-10
GitLab Flows Tweaks and Fixes

avidan-H:
## Status
- [x] Ready

## Fixes
- https://github.com/demisto/etc/issues/37789

## Description
Implements the following fixes and tweaks:
- Removes the job rules condition from the slack-notification child pipeline workflow. This is because any job in the parent pipeline that triggers the slack notification child pipeline should have the proper rule conditions for running. Therefore we can remove the rule conditions from the child.
- Adds a `when: always` field to rules of jobs that should run even when previous jobs failed. Used in the various `<trigger-slack-notification>` jobs and the jobs whose artifacts the trigger jobs depend upon. This way, these jobs won't be skipped if previous jobs in the pipeline fail (which should prevent the slack notification to the `dmst-content-team` channel from not occurring).
- Explicitly pass the `$SLACK_CHANNEL` environment variable to the `slack-notification` child pipeline from the parent pipeline's trigger job. This is useful when passing an alternate value for the environment variable when triggering a pipeline manually (either via the UI or API) - this way the value in the child pipeline is not the default but rather the value that was passed to the parent from the trigger.
- Removes the `run-validations` job from the `on-push` flow.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/on-push.yml
+++ b/.gitlab/ci/on-push.yml
@@ -21,19 +21,6 @@ trigger-private-build:
     - python3 Utils/get_private_build_status.py --github-token $GITHUB_TOKEN
 
 
-run-validations:
-  extends:
-    - .run-validations
-    - .push-rule
-
-
-# Uncomment when docker in docker will work
-# run-unittests-and-lint:
-#   extends:
-#     - .run-unittests-and-lint
-#     - .push-rule
-
-
 create-instances:
   extends:
     - .default-job-settings
@@ -160,6 +147,7 @@ server_master:
   rules:
     - if: '$CI_PIPELINE_SOURCE =~ /^(push)$/ && $CI_COMMIT_BRANCH !~ /^[0-9]{2}\.[0-9]{1,2}\.[0-9]$/'
     - if: '$NIGHTLY'
+      when: always
   variables:
     INSTANCE_ROLE: "Server Master"
 
@@ -174,3 +162,6 @@ slack-notify-nightly-build:
     PIPELINE_TO_QUERY: $CI_PIPELINE_ID
     WORKFLOW: 'Content Nightly'
     JOB_NAME: 'server_master'
+    # Passes the environment variable from the parent pipeline to the child which can be useful for cases
+    # when triggering pipeline with alternate env variable value passed in the API call
+    SLACK_CHANNEL: '$SLACK_CHANNEL'
--- a/.gitlab/ci/sdk-nightly.yml
+++ b/.gitlab/ci/sdk-nightly.yml
@@ -1,6 +1,12 @@
 .sdk-nightly-schedule-rule:
   rules:
-    - if: '$CI_PIPELINE_SOURCE =~ /^(schedule)$/ && $DEMISTO_SDK_NIGHTLY == "true"'
+    - if: '$CI_PIPELINE_SOURCE =~ /^(schedule|trigger)$/ && $DEMISTO_SDK_NIGHTLY == "true"'
+
+# used for jobs which we want to run in a pipeline even when previous jobs in the pipeline fail e.g. slack notification
+.sdk-nightly-schedule-rule-always:
+  rules:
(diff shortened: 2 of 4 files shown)


.gitlab/ci/sdk-nightly.yml
```diff
@@ -93,13 +100,9 @@ demisto-sdk-nightly:create-instance:
 demisto-sdk-nightly:run-commands-against-instance:
   extends:
     - .default-job-settings
-    - .sdk-nightly-schedule-rule
+    - .sdk-nightly-schedule-rule-always
   variables:
-    DESTROY_INSTANCES: "true"
     INSTANCE_ROLE: "Server Master"
-  needs: ["demisto-sdk-nightly:create-instance"]
```
DinaMeylakh, line 100: why is this not needed anymore?

.gitlab/ci/sdk-nightly.yml
```diff
@@ -1,6 +1,12 @@
 .sdk-nightly-schedule-rule:
   rules:
-    - if: '$CI_PIPELINE_SOURCE =~ /^(schedule)$/ && $DEMISTO_SDK_NIGHTLY == "true"'
+    - if: '$CI_PIPELINE_SOURCE =~ /^(schedule|trigger)$/ && $DEMISTO_SDK_NIGHTLY == "true"'
+
+# used for jobs which we want to run in a pipeline even when previous jobs in the pipeline fail e.g. slack notification
+.sdk-nightly-schedule-rule-always:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE =~ /^(schedule|trigger)$/ && $DEMISTO_SDK_NIGHTLY == "true"'
+      when: always
```
DinaMeylakh, line 9: Can you extend the .sdk-nightly-schedule rule with always? might save a line of code duplication

.gitlab/ci/on-push.yml
```diff
@@ -21,19 +21,6 @@ trigger-private-build:
     - python3 Utils/get_private_build_status.py --github-token $GITHUB_TOKEN
 
 
-run-validations:
-  extends:
-    - .run-validations
-    - .push-rule
```
DinaMeylakh, line 27: why are we deleting this?

.gitlab/ci/sdk-nightly.yml
avidan-H: By default all jobs of a previous stage need to finish before a job in a subsequent stage start.

.gitlab/ci/sdk-nightly.yml
avidan-H: We don't want all `sdk-nightly`-related jobs to always run. Only the ones that slack-notification depend upon need to always execute regardless of the run-status of previous jobs in the pipeline.

.gitlab/ci/on-push.yml
avidan-H: I think we are going to leave `validations` and `unittests and lint` in CircleCI for all workflows for now - scheduled or not.

avidan-H:
Adding the `docs-approved` label since there isn't any documentation that requires review.

.gitlab/ci/sdk-nightly.yml
DinaMeylakh: yeah I understand, I thought about something like this
```suggestion
# used for jobs which we want to run in a pipeline even when previous jobs in the pipeline fail e.g. slack notification
.sdk-nightly-schedule-rule-always:
  extends: .sdk-nightly-schedule-rule
  when: always 
```
but I'm not sure this would work with well other rules (applying the always to the specfic rule rather than all of them), maybe theres a syntax that does do the trick that you know of

avidan-H: Ah - my bad, I misunderstood your comment. I don't think the `extends` syntax would work properly/as you expect as it's included in your suggestion. You could probably successfully use the `!reference` syntax but... ¯\_(ツ)_/¯

DinaMeylakh (approved):

PR #13198, opened 2021-06-14
SailPoint IdentityIQ correct yaml outputs

DeanArbel:
## Status
- [x] Ready

## Description
- Corrected the documented integration outputs to match the actual outputs.
- Added proper error handling for failed auth

   - [x] Yes
       - Further details: the actual output didn't change, but it'll be recognized as failed due to changes in the yaml.


bakatzir (approved):
looks good. even though the validation points out this is a BC break, in fact it is not, as the integrations outputs always went to the path `IdentityIQ`.
so while it will require a force-merge, no need to update in the RN as breaking change.

PR #13209, opened 2021-06-14
Fix passing SLACK_CHANNEL env var value

avidan-H:
## Status
- [x] Ready

## Description
Our scheduled pipelines passed the string literal `$SLACK_CHANNEL` instead of the environment variable's value. See this [pipeline](https://code.pan.run/xsoar/content/-/pipelines/1250578) for an example.

## Additional Changes
Change the `upload-packs-to-marketplace` job to always run during the bucket-upload flow even if previous jobs failed.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Itay4 (approved):

avidan-H:
Adding the `docs-approved` label since there isn't any documentation requiring review.

PR #13265, opened 2021-06-16
Fix release notes creation in sdk-nightly & on push flows

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://code.pan.run/xsoar/content/-/jobs/6074172

## Description
In the job artifacts of the link above, the [packs-release-notes.md file](https://code.pan.run/xsoar/content/-/jobs/6074172/artifacts/file/artifacts/packs-release-notes.md) which was created (and which we use when creating a content release) didn't detect and add the release notes since the last release. This is because we were passing the current branch's ref instead of the `$GIT_SHA1` env variable stored value from `.circleci/content_release_vars.sh` which contains the sha ref of the last release.

## Additional Changes
Added the creation of the release notes file and content descriptor file to the `on-push` flow (to only execute on master or release branches).

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/on-push.yml
+++ b/.gitlab/ci/on-push.yml
@@ -71,9 +71,20 @@ create-instances:
         ./Tests/scripts/prepare_content_packs_for_testing.sh "$GCS_MARKET_BUCKET"
       fi
     - section_end "Prepare Content Packs for Testing"
+    - section_start "Create Content Descriptor"
+    - |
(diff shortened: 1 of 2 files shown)


avidan-H:
@moishce FYI

Itay4 (approved):

avidan-H:

PR #13314, opened 2021-06-21
Fix copy of content-descriptor.json to artifacts

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://code.pan.run/xsoar/content/-/jobs/6159487

## Description
Reverting putting the execution of `Utils/release_notes_generator.py` under a condition checking whether we were on the `master` or a release branch. That script also generates the `content-descriptor.json` file which is used in updating the content on an xsoar instance.

## Screenshots
image

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No - it fixes the bc that I broke 🙍🏼‍♂️ in this [PR](https://github.com/demisto/content/pull/13265)

## Must have
- [ ] Tests
- [ ] Documentation


Itay4 (approved):

avidan-H:

Itay4 (approved):

DinaMeylakh (approved):
I think the echo can be removed now but it's not important

PR #13333, opened 2021-06-22
rebranding - apex -> apex one

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38256

   - [x] No


ShahafBenYakir (commented):
I think we also need to fix this in description.md, README.md and strings/titels in py file.

DeanArbel (approved):
Looks good

PR #13345, opened 2021-06-22
SMBv2: Bugs fixes

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37970
fixes: https://github.com/demisto/etc/issues/38165

## Description

- Adding the client_guid to the clientConfig instance only if has a value as a possible workaround for a customer
>Customer updated this workaround worked for him.

- Added a monkeypatch to set the `datetime_value` return value of the `_parse_value` in the `DateTimeField` class to datetime.today() on OverflowError exception.
- Opened https://github.com/jborean93/smbprotocol/issues/114 in the smbprotocol project.


yaakovi (approved):
can we add a unit test for `_parse_value`?

PR #13418, opened 2021-06-27
Some Improvements to GitLab Flows

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
- Adds a `fan-in` stage
- Adds a `fan-in` job to the following flows
	- nightly
	- bucket-upload
	- sdk-nightly
- Changes the bucket-upload flow so that the `upload-packs-to-marketplace` job will not be run if previous jobs failed
- Adds retry to the slack-notify job
- Makes the slack notification trigger job's status dependent on the child pipeline's status, i.e. it will fail if the slack-notify job fails

## Screenshots
Should prevent such cases as in the attached screenshot where the packs were uploaded despite installation of those packs on a server instance failing
image

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/.gitlab-ci.yml
+++ b/.gitlab/ci/.gitlab-ci.yml
@@ -12,6 +12,7 @@ stages:
   - create-instances
   - run-instances
   - upload-to-marketplace
+  - fan-in  # concentrate pipeline artifacts to single job before triggering child slack pipeline
 
 
 variables:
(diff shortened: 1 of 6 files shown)


Itay4 (approved):

avidan-H:

PR #13438, opened 2021-06-28
Demisto Class - add note about custom fields returned from the incident method

Itay4:
## Status
- [x] Ready

## Description
clarify how custom fields are returned from the `demisto.incident()` func

### DIFF
--- a/Tests/demistomock/demistomock.py
+++ b/Tests/demistomock/demistomock.py
@@ -773,7 +773,8 @@ def incidents(incidents=None):
 
 def incident():
     """Retrieves the current incident and all its fields (e.g. name, type).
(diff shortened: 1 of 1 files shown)


Tests/demistomock/demistomock.py
```diff
@@ -772,7 +772,8 @@ def incidents(incidents=None):
 
 
 def incident():
-    """Retrieves the current incident
+    """Retrieves the current incident and all its fields (e.g. name, type).
+    The incident custom fields will be populated under as an array under the CustomFields attribute.
```
glicht, line 776: Should be `dict` and also can add an example: `demisto.incident()['CustomFields'].get('filename')`

glicht (approved):

PR #13644, opened 2021-07-12
Pan os fetch long running

bakatzir:
## Status
- [x] In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/16695

- [x] 5.0.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.yml
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.yml
@@ -48,6 +48,15 @@ configuration:
   name: additional_malicious
   required: false
   type: 12
+- display: Incident Type
+  name: incident_type
+  required: false
+  type: 13
+- display: Query to fetch logs
+  name: query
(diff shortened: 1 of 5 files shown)


Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -48,6 +48,31 @@ configuration:
   name: additional_malicious
   required: false
   type: 12
+- display: Incident Type
```
Itay4, line 51: arent you missing the `longRunning` parameter here?

bakatzir: It is in line 4694
https://github.com/demisto/content/pull/13644/files#diff-f261a9104c2b46a147c9f66fbd3d1c8018f43e76a5d82cd0feb75294ab9e9056R4694

PR #13649, opened 2021-07-12
Copy notes to incident bug

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38861

## Description
Fixed an issue where the automation failed to copy the notes.

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/CopyNotesToIncident/CopyNotesToIncident.py
```diff
@@ -27,8 +29,19 @@ def copy_notes_to_target_incident(args: Dict[str, Any]) -> CommandResults:
...
-            md = f'## {len(note_entries)} notes copied'
+            result = []
+            retries = 3
+            sleep_time = 1
+            for i in range(retries):
+                result = demisto.executeCommand("addEntries", {"id": target_incident, "entries": note_entries})
+                if result:
+                    break
+                sleep(sleep_time)
+                sleep_time += 1
```
DeanArbel, line 40: use exponential backoff, i.e. `sleep_time *= 1`. Also, add a `debug` log to indicate that it failed

Packs/CommonScripts/Scripts/CopyNotesToIncident/CopyNotesToIncident.py
```diff
@@ -27,8 +29,19 @@ def copy_notes_to_target_incident(args: Dict[str, Any]) -> CommandResults:
...
+            if result:
```
DeanArbel, line 41: make sure to use `isError` as well, e.g.
```suggestion
            if result and not isError(result):
```

DeanArbel (commented):
@ChanochShayner Nice work.
Do note we also have a `retry` in our `playbook tasks`, though this solution does offer a good UX for CLI usage.
If we go forward with this implementation, please consider adding a `retry` arg. @guykeller wdyt?

Packs/CommonScripts/Scripts/CopyNotesToIncident/CopyNotesToIncident.py
ChanochShayner: removed sleep

Packs/CommonScripts/Scripts/CopyNotesToIncident/CopyNotesToIncident.py
ChanochShayner: done

ChanochShayner:
I was able to reproduce the problem, and removing the ID & Version from the entry resolves the issue.
https://github.com/demisto/etc/issues/38861#issuecomment-878189819

Packs/CommonScripts/Scripts/CopyNotesToIncident/CopyNotesToIncident.py
```diff
@@ -8,6 +8,11 @@
 ''' STANDALONE FUNCTION '''
 
 
+def remove_id_and_version_from_entry(entry):
+    del entry['ID']
+    del entry['Version']
```
DeanArbel, line 13: use `pop()` with `None` if the keys don't exist
```suggestion
    entry.pop('ID', None)
    entry.pop('Version', None)
```

DeanArbel (approved):
Good job. see my comment below.

guykeller (approved):

PR #13661, opened 2021-07-12
Copy the updated conf to Tests/conf.json before running tests

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: the nightlys testing different amount of tests

## Description
For GitLab, because the updated `conf.json` doesn't persist at `Tests/conf.json` between jobs, we need to copy the updated `conf.json` file from the artifacts directory to `Tests/conf.json` path before the `run_tests.sh` step since that script is configured to look there. Otherwise, the ordinary `conf.json` is used from when the repo is cloned at the start of the job. Resulting

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] ~~Tests~~ N/A
- [x] ~~Documentation~~ N/A


avidan-H:
Adding the `docs-approved` label since there is no documentation requiring the review of a technical writer.

Itay4 (approved):

PR #13662, opened 2021-07-12
Adding multiple fixes for CS HealthCheck pack

tsachzim:
Adding few fixes to scripts that analyzing large incidents

## Description
Fixing multiple issues


tsachzim:
please add 'Common Widgets' as a dependency to the pack.

Packs/HealthCheck/Scripts/HealthCheckAnalyzeLargeInvestigations/HealthCheckAnalyzeLargeInvestigations.py
```diff
@@ -35,32 +35,38 @@ def format_dict_keys(data: List[Dict[str, Any]]) -> List:
...
+    incidentsbiggerthan1mb = []
+    incidentsbiggerthan10mb = []
+    incidentswithmorethan500entries = []
+    for incident in res['data']:
+        if incident['AmountOfEntries'] >= 500:
+            incidentswithmorethan500entries.append(incident)
+        if round(incident['Size(MB)']) >= 10:
+            incidentsbiggerthan10mb.append(incident)
+        else:
+            incidentsbiggerthan1mb.append(incident)
```
yaakovi: this means that only incidents larger than 1MB **and** smaller than 10MB will be considered as over 1MB.
for example, incident with size 15MB will only be counted once (in the over 10MB list).

is that ok?

tsachzim: correct.

PR #13665, opened 2021-07-13
removing unnecessary section in test data

roysagi:
removing unnecessary section in test data - 
first run with change in code just to check that lint passes


Itay4 (approved):

Itay4:
build fails on non related instance issue
merging

PR #13673, opened 2021-07-13
Add V2 of SlackAsk to support SlackV3

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: Internal requirement for SlackV3 usage

## Description
- Added `slackInstance` argument so a customer can specify a specific instance to send the message with.
- Bumped the `using-brand` to SlackV3

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


amshamah419:
Approving own docs since the documentation has not changed in any considerable way.

Packs/Slack/Scripts/SlackAskV2/SlackAskV2.py
```diff
@@ -0,0 +1,137 @@
...
+                                   DATE_FORMAT)
+    default_response = demisto.get(demisto.args(), 'defaultResponse')
+
+    entitlement_string = entitlement + '@' + demisto.investigation()['id']
+    if demisto.get(demisto.args(), 'task'):
+        entitlement_string += '|' + demisto.get(demisto.args(), 'task')
+
+    args = {
+        'ignoreAddURL': 'true',
+        'using-brand': 'SlackV3'
```
liorblob, line 80: @itay4 @amshamah419 do we want the `using-brand`?

amshamah419: I think since we are using `send-notification` we should just to prevent other integrations from firing. I'm open to other ideas though since I don't like the idea of bumping the script for any slack version change.

Packs/Slack/ReleaseNotes/2_0_1.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### New: SlackAskV2
+- Sends a message (question) to either user (in a direct message) or to a channel. The message includes predefined reply options. The response can also close a task (might be conditional) in a playbook. (Available from Cortex XSOAR 5.5.0).
```
yuvalbenshalom, line 4: lets add also that this was added in order to support slack v3

Packs/Slack/Scripts/SlackAskV2/README.md
```diff
@@ -0,0 +1,102 @@
+<p>
+  Sends a message (question) to either user (in a direct message) or to a channel. The message includes predefined reply options. The response can also close a task (might be conditional) in a playbook.
```
yuvalbenshalom, line 2: lets add the comment that this should be used with slack V3

Packs/Slack/Scripts/SlackAskV2/README.md
```diff
@@ -0,0 +1,102 @@
+<p>
```
yuvalbenshalom, line 1: BTW, why is this on HTML and nor markdown?

yuvalbenshalom (changes requested):
see my comments, 
and please add here the addition for slackAsk(V1) that will catch the error when there is no instance and will propose that if a SlackV3 is enabled, the user should use the SlackAskV2.

Packs/Slack/Scripts/SlackAskV2/README.md
amshamah419: I used the sdk command `cp -r ~/SlackAsk/. ~/SlackAskV2`. I suppose I can use `generate-docs` instead.

yuvalbenshalom (approved):

Itay4:
wouldn't it be better to add a brand arg to the existing script?

amshamah419:
@Itay4 My concern with that is for us to not break BC with SlackV2, V2 would have to be the default. Then customers moving to V3 would have to specify that they want V3 and it may not be very user friendly.

I personally would like to see a way for us to see what scripts are installed and then handle the logic internally based on what's enabled, but I'm not sure if that's even on the roadmap.

I'm 100% open to other ideas though.

PR #13708, opened 2021-07-15
Hash incident fields test remove from skip

ChanochShayner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26850

## Description
remove from skip.

- [x] No

## Must have
- [ ] Tests
- [ ] Documentation


erezh31 (approved):

Packs/ML/TestPlaybooks/script-CreateIncidentsForEvaluateMLModllAtProduction.yml
```diff
@@ -84,4 +84,4 @@ subtype: python2
 runonce: false
 dockerimage: demisto/python:2.7.18.10627
 runas: DBotWeakRole
-fromversion: 5.5.0
```
yuvalbenshalom, line 87: why did you change the from version from 5.5.0 to 5.0.0?

ChanochShayner: This script was created only for the TPB and is from version 5.5 while the TPB is from version 5.0 so the TPB always fails in version 5 because it fails to install this script.

Packs/ML/Scripts/HashIncidentsFields/HashIncidentsFields.py
```diff
@@ -41,12 +41,13 @@ def is_key_match_fields_to_hash(key, fields_to_hash):
 
 def hash_incident_labels(incident_labels, fields_to_hash):
     labels = []
-    for label in incident_labels:
-        if is_key_match_fields_to_hash(label.get('type'), fields_to_hash):
-            value = label.get('value') or ''
-            if value:
-                label['value'] = hash_value(value)
-        labels.append(label)
+    if isinstance(incident_labels, list):
```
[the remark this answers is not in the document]
ChanochShayner: it could be None when there is no labels

guykeller (approved):

PR #13767, opened 2021-07-19
Run validation and unit tests and lint even against test bucket

avidan-H:
## Status
- [x] Ready

## Description
So that we see `run-unittests-and-lint-upload-flow` and `run-validations-upload-flow` jobs actually run the relevant `demisto-sdk` commands in the GitLab Upload Flow (even though we are running there against a test bucket) unlike what you see in the attached screenshots and what happens in the bucket upload flow [here](https://code.pan.run/xsoar/content/-/pipelines/1397925). So we can verify that it is working properly.

## Screenshots
image
image

- [ ] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/global.yml
+++ b/.gitlab/ci/global.yml
@@ -154,25 +154,21 @@
     - section_end "Test Infrastructure"
     - section_start "Run Unit Testing and Lint"
     - |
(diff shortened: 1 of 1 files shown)


avidan-H:
Adding the `docs-approved` label since there isn't any documentation requiring review by a technical writer.

Itay4 (approved):

PR #13782, opened 2021-07-20
update AF query

bakatzir:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/39321

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


yuvalbenshalom:
@bakatzir Is it possible to add any test here?

bakatzir:
> @bakatzir Is it possible to add any test here?

this script does not execute the call, only prepares it.
I will add a test that the call args are properly formatted.

DeanArbel (approved):
Nice

PR #13798, opened 2021-07-21
[Elasticsearch Feed] Fix lastTime timestamp in wrong format

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39361

## Description
Changed the timestamp to a `13 digit` timestamp (mili seconds) rather than `16 digits` (microseconds).
   - [x] No


DeanArbel:
Tested on a client's env.

guykeller (approved):

PR #13848, opened 2021-07-26
Incidents test playbook

daryakoval:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes:https://github.com/demisto/etc/issues/34585

## Description
Fetch Incidents test playbook. 

## Screenshots

- [ ] 5.5.0
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


daryakoval:
## Fetch Incidents design:
---

Same as fetch indicators, but we will check if no incidents were fetched, just skip fields check and proceed to health check. Else if there are Incidents, we will check if all fields exist.

image
 
First let's delete context and `get incident field to search` using demisto rest api, in order to get searchfield, sourcebrand fields.

After that we should get instance_name using integration name. Using REST API, and new script I wrote: `GetInstanceName`.

Then `Set incidents search query` just set context key: searchQuery, value : `sourceInstances:${Instances.instanceName}*` .

Now we can start checking if feed already fetched incidents using: `VerifyEnoughInsidents` and `Check If Indicators exists`. `VerifyEnoughInsidents` - new script I wrote, it is a wrapper for search indicators, in order to use genericPolling. If fetch didn't happened yet, we will use GenericPolling: 
image

image

Now we use `SearchIncidentsV2` to get all incidents. 
Then I wrote script in order to check if all fields that we asked are mapped in all incidents that were fetched. It happens using script `VerifyContextFieldsList` at step named: `Check If Fields Exists`. 

**Here it is different from Fetch indicators playbook**: if no incidents were fetched, just skip fields check and proceed to health check. 

After all this, we can check Integration Health using script `VerifyIntegrationHealth` that I wrote too, it checks if there were some errors during the fetch. There is one more GenericPolling, since integration health check can be done only after fetch is done. ("Pulled X Indicators" message appears near the instance). I am polling until fetch is done - the integration appears in the returned by rest API dict, and check if there were some errors. 
image

dantavori (approved):

richardbluestone:
The `VerifyContextFieldsList` automation was not included. What does it do?

What does `search field` do? I don't understand the description: Comma separated list of fields to confirm that exists in al indicators.
Are we searching for indicator fields that exist in XSOAR that match the relevant criteria?

richardbluestone:
Reviewed and updated.

PR #13860, opened 2021-07-27
Create indicator relationship bug

ChanochShayner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39519

## Description
Handle STIX prefix in CreateIndicatorRelationship arguments.

- [x] 6.2.0

- [x] Yes


bakatzir (approved):
please inline comment the UT for the future dev

bakatzir:
fails on a few servers only due to the gitlab outage. force merging.

PR #13865, opened 2021-07-27
[Fix] ParseEmailFile MIME encoded subject decode fallback

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38707

## Description
Add a fallback in case MIME decoding fails on the subject (for example 2 utf-8 encoded in the same message) - use default decoding flow.

   - [x] No

## Must have
- [x] Tests


Packs/CommonScripts/ReleaseNotes/1_4_10.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### ParseEmailFiles
+- Improved parsing of MIME encoded subjects.
```
Itay4, line 4:
```suggestion
- Improved decoding of MIME encoded words.
```

Itay4 (approved):

PR #13897, opened 2021-07-29
[QRadar v2] Add Requests timeout advanced parameter

DeanArbel:
## Status
- [x] Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/39254

## Description
Added an advanced parameter to control the requests timeout used in QRadar v2.

   - [x] No


Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -19,6 +19,7 @@
 urllib3.disable_warnings()
 
 """ ADVANCED GLOBAL PARAMETERS """
+REQUEST_TIMEOUT = 20                # number of seconds to wait for a request result
```
anara123, line 22: 20 seconds is enough?

anara123 (approved):

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
DeanArbel: should be, but I increased it to 30 just in case. It's customizable through the advanced params in any case.

PR #13899, opened 2021-07-29
O365 - Security And Compliance - Search And Delete is enabled

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39666

## Description
A task to check if an instance is enable was added (took from `O365 - Security And Compliance - Search`).

## Screenshots
image

   - [ ] Yes
       - Further details:
   - [x] No


abaumgarten:
@idovandijk FYI

yuvalbenshalom:
@abaumgarten did you check also the case that it is not enabled?

abaumgarten:
> @abaumgarten did you check also the case that it is not enabled?

Yes, indeed

yuvalbenshalom (approved):

PR #13973, opened 2021-08-04
Fix a git diff cmd

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://code.pan.run/xsoar/content/-/jobs/6865777 (hopefully)

## Description
Implements a couple of modifications.
- Simplifies the `git` commands run in `collect_tests_and_content_packs.py` to retrieve the `git diff` between the previous and current commits.
- Improves the GitLab nightly trigger script's usability.
- Set the `fetch` command to explicitly fetch from the `origin` remote.

Linked [here](https://code.pan.run/xsoar/content/-/pipelines/1468557) is the bucket upload flow triggered against this branch.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] ~~Tests~~ N/A
- [x] ~~Documentation~~ N/A

### DIFF
--- a/.gitlab/ci/global.yml
+++ b/.gitlab/ci/global.yml
@@ -36,9 +36,11 @@
 
 .clone_and_export_variables: &clone_and_export_variables
   - source .gitlab/helper_functions.sh
+  - section_start "Git - Job Start Actions" --collapsed
   - git fetch --unshallow origin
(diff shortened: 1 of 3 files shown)


avidan-H:
Adding the `docs-approved` label since there isn't any documentation requiring review by a technical writer.

Itay4 (approved):

PR #13994, opened 2021-08-04
TwitterSOARx Integration Addition

cbrake1:

## Status
- [ ] In Progress
- [x] Ready

## Description
The addition of a new integration called "TwitterSOARx" and a sample playbook for it to the Marketplace. The integration allows for twitter searches to be performed in XSOAR.

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/TwitterSOARx/TestPlaybooks/playbook-Remove_Excluded_Accounts.yml
```diff
@@ -0,0 +1,141 @@
...
+    quietmode: 0
+  "5":
+    id: "5"
+    taskid: 025432d6-af62-49d4-8282-e4a5b3cccfce
+    type: regular
+    task:
+      id: 025432d6-af62-49d4-8282-e4a5b3cccfce
+      version: -1
+      name: Remove Excluded Users
+      scriptName: TwitterRemoveExcludedAccounts
```
Itay4, line 102: what is this script?

Packs/TwitterSOARx/pack_metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "TwitterSOARx",
```
Itay4, line 2:
```suggestion
    "name": "Twitter",
```

Itay4 (commented):
thanks for the contribution !
general comments:
1. please rename `TwitterSOARx` to `Twitter` in the filenames as well
2. not sure what you meant to do with those doc files, but we cant show them in the ui or in the docs site. i recommend adding all the info in the readme
3. please remove the usage of `print` in the code

Packs/TwitterSOARx/TestPlaybooks/playbook-Remove_Excluded_Accounts.yml
cbrake1: This script is a sample playbook that returns results of a Twitter search (from the twitter-get-users command) and filters out accounts that are in a given List in XSOAR. This allows a user to search for users and filter accounts that are known to be harmless to them so they do not have to see accounts that they have added to a list for any given reason (such as having investigated the account, determining it is harmless and wanting to not see it again)

Itay4:
@cbrake1 please note the build still fails on a few of my comments above and due to your changes
let me know if you need my help with that

cbrake1:
Sorry for the delay, I am still awaiting a download for the Community Edition of XSOAR. I've completed the changes suggested in your comments and resolved several issues in the Unit Testing. I am hoping to be able to use the Community Edition of XSOAR to test my integration as I make changes to pass the Pylint and Mypy tests. If I do not receive a download by tomorrow, I will finish addressing the errors in those two tests, though I would not be able to test the slight changes in XSOAR just yet. 

I do have one question, the Mypy returns an error stating:
 "Module has no attribute "parse"  [attr-defined]
                q = urllib.parse.quote(' ') + demisto.args().get('q')[1:]"
Though the code works does exactly what it is intended to do, and I do intend to call urllib.parse.quote, which is documented here: https://docs.python.org/3/library/urllib.parse.html
I am not sure how to go about this error as I cannot remove the '.parse' from 'urllib.parse.quote'. I can work around this by removing the need for urllib, but it makes the code more difficult to understand. I only need urllib.parse.quote to change any '#' characters to '%23' when encoded in a URL. I can do this manually in the code to work around the error with urllib.parse, but it does make it a bit more difficult for a fresh pair of eyes to understand the code. 

Should I carry out that work around, or is there anything else I can do to address the error stating urllib has no attribute 'parse'?

Itay4:
@cbrake1 you can ignore this with the following `# type: ignore[attr-defined]`
see https://mypy.readthedocs.io/en/stable/error_codes.html

cbrake1:
> @cbrake1 you can ignore this with the following `# type: ignore[attr-defined]`
> see https://mypy.readthedocs.io/en/stable/error_codes.html

I read the documentation and believe I have added the appropriate code to ignore the attr-defined errors with urllib, but I still am running into attr-defined errors in the MyPy errors.

I am also running into name-defined errors with the 'client' expected arg-type. Based off of use in sample code in the coding conventions documentation and other content packs, I believe that I am using the 'client' arg type correctly. What can I do to resolve these errors, or are there any additional resources I can view? Thank you for your time and help.

Itay4:
@cbrake1 i believe you can just change `Client.` to `self.` in those cases 
let me know if that doesnt work for you

cbrake1:
> @cbrake1 i believe you can just change `Client.` to `self.` in those cases
> let me know if that doesnt work for you

I'm still having some troubles in the main function with the self argument as I have to include a value in the functions to run the Twitter search commands (lines 222, 225, and 231), but 'self' is treated as an undefined variable, and 'Client' does not work either, though it is expecting the "Client" arg-type, according to Mypy.

Itay4:
there was missing an initialization of the `Client` object so i've added it and also made a few more changes - please have a look at my last commits
also, i've notived you're missing the `proxy` and `insecure` integration parameters, can you please add them? you can see https://xsoar.pan.dev/docs/integrations/code-conventions for more details

Itay4:
@cbrake1 probably better to add the parameters both to the YML file and the code, so if you have issues testing these additions i suggest we dont do them at this time
are there any other changes you wish to do or can we merge this?

cbrake1:
> @cbrake1 probably better to add the parameters both to the YML file and the code, so if you have issues testing these additions i suggest we dont do them at this time
> are there any other changes you wish to do or can we merge this?

If I can add the `proxy` and `insecure` additions at a later time when I have the ability to test them, then we can merge. I am still awaiting an opportunity to get an instance of the XSOAR community edition so until then I cannot test changes. If we can merge without those two additions and I'll add them in a later update, then I'd say it's ready to be merged.

Itay4 (approved):

PR #14012, opened 2021-08-05
[EDL] Add CIDR to Collpased IPs

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40018

## Description
Added support for CIDR indicators when using the *Should Collapse IPs* parameter.

   - [x] No

## Must have
- [x] Tests


bakatzir (commented):
there is also an IPv6CIDR while we are at it.
also - I suggest splitting the UT to different cases, as it is hard to understand the expected result for CIDR, at least IMO.

DeanArbel:
> there is also an IPv6CIDR while we are at it.
> also - I suggest splitting the UT to different cases, as it is hard to understand the expected result for CIDR, at least IMO.

IPV6CIDR was dropped, since there's an issue with using its ranges (as the range can be greater than python max int).

I'd say these tests could use a rewrite in general.

yuvalbenshalom (approved):

bakatzir (approved):

PR #14023, opened 2021-08-05
Changes API call for incident for recursive fetch from Gurucul

gra-xsoar:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/Gurucul/ReleaseNotes/1_1_2.md
```diff
@@ -0,0 +1,5 @@
+#### Integrations
+##### Gurucul-GRA
+
+- **Bug Fixes**:
+  - **Breaking Change**: Fixed an issue with fetch incidents.
```
Shellyber, line 5: @gra-xsoar Why the breaking changes?

gra-xsoar: its incident fetch change that's why we wrote: "Breaking change"?
if not required then we will update the release file

Shellyber:
@gra-xsoar Can you please explain what is the change? What is the difference between then and the fix now? 
Also please update the docker image of the integration to `3.9.6.22912` and add this update to the release notes as well

gra-xsoar:
1. Prior we provide a call to our API per minute which by default set in Demisto and the same way we implemented in our API but client-side default time was set as 60 min and for the same reason our code is not synced up with demisto data
so for the quick fix, we removed our per minute API call and now we gave the call for the whole day, and for avoiding duplicates we are given them as pre-process rule.in the next release, we will fix this issue which is independent of the duration.

2. I updated the docker image as 3.9.6.22912 in integration.
3. Removed breaking change

Shellyber (approved):

PR #14056, opened 2021-08-08
GitHub get path data command

tomneeman151293:

## Status
- [ ] In Progress
- [x] Ready

## Description
For Update Docker Image Auto, added this command to retrieve data of given path. This will be used to get SHA of files for docker image update playbook tasks.
this command was needed because currently docker image playbook is implemented retrieving SHA of files using un-authenticated http calls, which has a small quota. This command solves this by performing authenticated call (using integration params) so quota will not be reached (authenticated has 1k + calls per hour allowed)

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/GitHub/Integrations/GitHub/GitHub.py
```diff
@@ -1834,6 +1834,56 @@ def fetch_incidents_command_rec(start_time, last_time, page=1):
...
+    repo: str = args.get('repository') or REPOSITORY
+    organization: str = args.get('organization') or USER
+    branch_name: Optional[str] = args.get('branch_name')
+
+    url_suffix = f'/repos/{organization}/{repo}/contents/{relative_path}'
+    url_suffix = f'{url_suffix}?ref={branch_name}' if branch_name else url_suffix
+
+    headers = {
+        'Authorization': "Bearer " + TOKEN,
+        'Accept': 'application/vnd.github.VERSION.object',
```
Shellyber, line 1855: Why do we add this? is this needed? 
```suggestion
        'Accept': 'application/vnd.github.VERSION.object',
```

Packs/GitHub/Integrations/GitHub/GitHub.py
```diff
@@ -1834,6 +1834,56 @@ def fetch_incidents_command_rec(start_time, last_time, page=1):
     return incidents, last_time
 
 
+def get_path_data():
```
Shellyber, line 1837: Shouldn't be `get_file_data`?

Shellyber (commented):
Looking good! 👍 
Please see my comments

Packs/GitHub/Integrations/GitHub/GitHub.py
tomneeman151293: For consistency, object returns the expected. I am afraid in future they will change their API to do something else if Accept was not given.

Packs/GitHub/Integrations/GitHub/GitHub.py
tomneeman151293: It works for path as well. Not only file so I included it

tomneeman151293: See the README example!

Shellyber (approved):

PR #14066, opened 2021-08-09
[QRadar v2] Improve Handling for Empty Fetch Windows

DeanArbel:
## Status
- [x] In Progress

## Related Issues
related: https://github.com/demisto/etc/issues/39254

## Description
Add expanding search window (number of offenses to pull * number of tries) when trying to find the first fetchable offense.


Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -910,7 +910,7 @@ def fetch_raw_offenses(client: QRadarClient, offense_id, user_query):
     """
     # try to adjust start_offense_id to user_query start offense id
     try:
-        if isinstance(user_query, str) and "id>" in user_query:
```
tomneeman151293, line 913: I suggest you to use the regex I have made in QRadar v3, see ID_QUERY_REGEX. WDYT?

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -956,10 +955,17 @@ def seek_fetchable_offenses(client: QRadarClient, start_offense_id, user_query):
...
+                if tries % 10 == 0:
+                    last_run = get_integration_context(SYNC_CONTEXT)
+                    last_run["id"] = end_offense_id
+                    set_integration_context(last_run, SYNC_CONTEXT)
             else:
                 latest_offense_fnd = True
-    return raw_offenses, fetch_query
+    if isinstance(raw_offenses, list):
+        raw_offenses.reverse()
+    return raw_offenses[:client.offenses_per_fetch], fetch_query
```
tomneeman151293, line 968: Is the slicing really needed? doesn't query limit you to client.offenses_per_fetch anyway?

tomneeman151293 (approved):
Awesome. Gave you suggestions if you feel like changing but no real comment

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
DeanArbel: It does, but since the window is increasing, it may increase to above the wanted size, so we have to do another slice.

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
DeanArbel: Great suggestion

Packs/QRadar/Integrations/QRadar_v2/QRadar_v2.py
```diff
@@ -956,10 +955,17 @@ def seek_fetchable_offenses(client: QRadarClient, start_offense_id, user_query):
                     )
                 lim_id = lim_offense[0]["id"]  # if there's no id, raise exception
             if lim_id >= end_offense_id:  # increment the search until we reach limit
-                start_offense_id += client.offenses_per_fetch
+                start_offense_id += (client.offenses_per_fetch * tries)
+                tries += 1
+                if tries % 10 == 0:
+                    last_run = get_integration_context(SYNC_CONTEXT)
+                    last_run["id"] = end_offense_id
```
tomneeman151293, line 962: I think you might miss incidents this way.
if offense corresponding to end_offense_id was not fetched (was sliced in line 968), the next time this function will be called will miss it

DeanArbel: It's ok, because then this code will be unreachable (it only runs when there's no fetch), so it's impossible to make it miss offenses, as none were fetched in this timeframe.

PR #13904, opened 2021-07-30
Pan os url filter

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39671

## Screenshots
Screen Shot 2021-07-30 at 9 08 24

## Minimum version of Cortex XSOAR
- [x] 5.0.0

## Does it break backward compatibility?
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

PR #14070, opened 2021-08-09
Palo Alto Networks PAN-OS- Fixed an issue where the creation of URL Filtering object was not successful in 8.X PAN-OS versions

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40173
also fixes: https://github.com/demisto/etc/issues/27501

root cause: https://github.com/demisto/content/pull/13904/files
prev issue: https://github.com/demisto/etc/issues/39671

## Description
the prev pr fixed 9.x and above, but broke 8.x.

- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests - I will check that panorama test PB runs successfully on both 8.1 and 9.0 instances.
- [x] Documentation


yuvalbenshalom (approved):
looks good, but maybe its possible to cover the addition of action tag in UT?

bakatzir:
> looks good, but maybe its possible to cover the addition of action tag in UT?

I've taken the liberty to add UT for 8.x and 9.x url_filter and custom_url_category creation.
I believe we should force-merge due to the lab outage.

bakatzir:
fails on 
```
[2021-08-15 09:16:25] - [MainThread] - [ERROR] - Failed Tests: 
							 - Panorama Best Practise - Test (Mock Disabled)
```

all other tests are passing. force merging.

PR #14075, opened 2021-08-09
Add gke runner tag for non-dind related jobs

avidan-H:
## Status
- [x] Ready

## Description
Adds the `gke` tag for non docker-in-docker ("dind") related jobs. This applies for all jobs except for the Run Lint & Unit Tests jobs which run `dind` service containers. The `gke` tagged runners are more performant according to the engineering-productivity team.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] ~~Tests~~ N/A
- [x] ~~Documentation~~ N/A

### DIFF
--- a/.gitlab/ci/bucket-upload.yml
+++ b/.gitlab/ci/bucket-upload.yml
@@ -66,6 +66,8 @@ create-instances-upload-flow:
 
 
 .install_packs_in_server:
+  tags:
+    - gke
   needs: ["create-instances-upload-flow"]
   stage: run-instances
   artifacts:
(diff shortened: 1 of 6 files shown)


avidan-H:

avidan-H:
We can ignore the failure `ci/circleci: Run Validations` failure. The validations don't know how to deal with the GitLab-related YAML files since we use a GitLab feature where the syntax is `!reference` which is apparently an invalid YAML syntax (but valid for GitLab). Once the changes are in `master` this error isn't raised since we only check changed files.

avidan-H:
Though it looks like a problem that needs to be handled generally speaking since the `configure_and_test_integration_instances.py` fails on the same problematic file in all of the `server <version>` GitLab jobs.

avidan-H:
See successful build [here](https://code.pan.run/xsoar/content/-/pipelines/1495350) when I changed the build to install the latest `demisto-sdk`, to see that it truly passes. Reverting it now. Will need to be force merged.

Itay4:
build failed due to issue on sdk validate that was fixed
merging

PR #14079, opened 2021-08-09
Add non-contrib stale branch deletion job

avidan-H:
## Status
- [x] Ready

## Description
Adds a scheduled job to delete stale 'non-contrib' branches. Creates a PR titled after the branch that will be deleted (and closes it immediately and deletes the branch) so that worst-case scenario we can restore the branch from the PR in GitHub if necessary.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] ~~Tests~~ N/A
- [x] ~~Documentation~~ N/A


avidan-H:
Adding the `docs-approved` label since there is no documentation requiring review.

avidan-H:
The GitLab build passed. For some reason the GitHub status check for it didn't update, my guess would be because of the degraded GitHub performance that was occurring.

Itay4:
status was not updated due to issue on gitlab
merging

PR #14119, opened 2021-08-10
BaseClient _http_req: `response` resp_type, assign res to DemistoException 

dorschw:
## Status
- [ ] In Progress
- [x] Ready

## Related issues
fixes: https://github.com/demisto/etc/issues/40244
fixes: https://github.com/demisto/etc/issues/40245

## Description
1. On a specific case in `_http_request()`, the `res` field of `DemistoException` was not filled, it now is. 
2. added `resp_type`=`response` to return the raw `Requests.response` object. it is returned as default, but it may be better to expect a specific value rather than a catchall. (UT already test with `resp_type=response`)

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


DeanArbel (approved):

Packs/Base/ReleaseNotes/1_13_15.md
```diff
@@ -0,0 +1,4 @@
+#### Scripts
+##### CommonServerPython
+- On a specific case in `_http_request()`, the `res` field of `DemistoException` was not filled, it now is.
```
amshamah419, line 3:
```suggestion
- Fixed an issue where when given a specific case in the `_http_request()` method, the `res` field of `DemistoException` would not be filled.
```

PR #14128, opened 2021-08-10
Checkpoint playbooks bug

guykeller:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40208

## Description
Changed the brand being searched for to the V2 one.
Not a BC break since V1 is deprecated, thus no longer has to be supported.
Tested locally.


bakatzir (commented):
I think the operator should be `in`, not `isEqualString`, so the PB would support both the deprecated v1, and v2.
changing it to not allow v1 would break bc to customer using a deprecated, but not EOL integration.

Packs/CheckpointFirewall/ReleaseNotes/2_0_12.md
```diff
@@ -0,0 +1,8 @@
+
+#### Playbooks
+##### Checkpoint - Block IP - Custom Block Rule
+- Fixed an issue where the playbook would require a Checkpoint Firewall V1 instance, rather than V2.
+##### Checkpoint - Publish&Install configuration
+- Fixed an issue where the playbook would require a Checkpoint Firewall V1 instance, rather than V2.
+##### Checkpoint - Block URL
+- Fixed an issue where the playbook would require a Checkpoint Firewall V1 instance, rather than V2.
```
bakatzir, line 8:
```suggestion
- Fixed an issue where the playbook would run only with a Checkpoint Firewall V1 instance.
##### Checkpoint - Publish&Install configuration
- Fixed an issue where the playbook would run only with a Checkpoint Firewall V1 instance.
##### Checkpoint - Block URL
- Fixed an issue where the playbook would run only with a Checkpoint Firewall V1 instance.
```

bakatzir (approved):

PR #14130, opened 2021-08-11
Migrate bucket upload workflow to GitLab

avidan-H:
## Status
- [x] Ready

## Description
- Removes upload dev rules that would have jobs not run when `MARKET_GCS_BUCKET` equaled `marketplace-dist-dev`
- Removes assignment of `MARKET_GCS_BUCKET` environment variable to `marketplace-dist-dev` in various job configurations
- Changes the default value of `MARKET_GCS_BUCKET` to the production value (AKA `marketplace-dist`)
- Removes the scheduled bucket-upload workflow from the CircleCI configuration

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] Maybe

## Must have
- [x] ~~Tests~~ N/A
- [x] ~~Documentation~~ N/A


avidan-H:
I'm adding the `docs-approved` label since there isn't any documentation for review.

avidan-H:
@DinaMeylakh @guyfreund I would appreciate it if you could take a look at this PR seeing as the two of you are far more familiar with this flow than myself - I easily could be missing/forgetting something here.

guyfreund (commented):
Looks good to me. Do we plan to remove all additional infra on `config.yml` once we see that `GitLab` is stable enough?

Itay4:
> Looks good to me. Do we plan to remove all additional infra on `config.yml` once we see that `GitLab` is stable enough?

correct

.circleci/config.yml
```diff
@@ -1610,18 +1610,6 @@ workflows:
...
-          # should trigger every day at 9 AM & 9 PM UTC (11:00 PM & 11:00 AM Israel Time)
-          cron: "0 9,21 * * *"
-          filters:
-            branches:
-              only:
-                - master
-    jobs:
-      *bucket_upload_jobs
-
   bucket_upload_trigger:
```
DinaMeylakh, line 1613: Do we want to keep this?

DinaMeylakh (changes requested):
allow Utils/gitlab_triggers/trigger_upload_flow_build.sh to use marketplace-dist also

.circleci/config.yml
avidan-H: I think this falls under the category of all the remaining infrastructure we would need to still remove pending seeing that the flow operates in `GitLab` stably as @guyfreund mentioned in his comment https://github.com/demisto/content/pull/14130#pullrequestreview-727090083.

DinaMeylakh (approved):

guyfreund (approved):

guyfreund:
good luck to us all

Noy-Maimon (approved):

Itay4:
@Noy-Maimon @guyfreund merging this - fyi

PR #14176, opened 2021-08-12
GitLab Job to Delete Branches Already Deleted from Github

avidan-H:
## Status
- [x] Ready

## Description
Our GitLab `content` repository doesn't reflect when branches are deleted in the Github `content` repository from which it is mirrored. This leads to the branches in the GitLab repo ever-increasing and eventually in turn, causing repo cloning failures at the start of GitLab jobs because the repo's size is so great and the request takes so much time that almost always results in connectivity interruptions and failures. Adds a GitLab job that executes a script to delete "excess" branches in GitLab - AKA branches that have already been deleted from the Github `content` repo but still exist in GitLab. I will add a scheduled pipeline that will run once a day to trigger this job once this is merged.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.gitlab/ci/.gitlab-ci.yml
+++ b/.gitlab/ci/.gitlab-ci.yml
@@ -41,3 +41,4 @@ include:
   - local: .gitlab/ci/bucket-upload.yml
   - local: .gitlab/ci/instance-test.yml
   - local: .gitlab/ci/sdk-nightly.yml
+  - local: .gitlab/ci/miscellaneous.yml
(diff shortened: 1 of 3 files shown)


Itay4 (approved):

avidan-H:
Adding the `docs-approved` label since there is no documentation requiring review.

PR #14299, opened 2021-08-15
LogRhythm  - handle number_of_days arg

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40358

   - [x] No

## Must have
- [ ] Tests
did not add test as there seems to be an issue with our instance to run this api


Itay4:
TDD: see test failure in https://code.pan.run/xsoar/content/-/jobs/7074407

yaakovi (approved):

PR #14431, opened 2021-08-19
SentinelOne V2 

devang-metron:
## Status
- [ ] In Progress
- [x] Ready

## Description
1. Fixed argument issue in following commands:
    - sentinelone-reactivate-site
    - sentinelone-create-white-list-item
2. Updated docker image version

- [x] 5.0.0
- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
+++ b/Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
@@ -216,9 +216,15 @@ class Client(BaseClient):
         response = self._http_request(method='GET', url_suffix=endpoint_url)
         return response.get('data', {})
 
-    def reactivate_site_request(self, site_id):
+    def reactivate_site_request(self, site_id, expiration, unlimited):
         endpoint_url = f'sites/{site_id}/reactivate'
-        response = self._http_request(method='PUT', url_suffix=endpoint_url)
+        payload = {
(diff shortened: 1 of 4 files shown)


guykeller:
@devang-metron 
Hi there! thanks for contributing.

Some notes:
1. I'd love to set a short talk and discuss these changes - what would be the best way to reach you?
2. It seems that the changes you've made are breaking backwards compatibility, is that on purpose? can you explain why it's necessary?

guykeller (changes requested):
It seems that the changes you've made are breaking backwards compatibility, is that on purpose? can you explain why it's necessary?

devang-metron:
@guykeller

> I'd love to set a short talk and discuss these changes - what would be the best way to reach you?

I am available for a call in mornings EST time (7am - noon EST) and at devang(Demisto DFIR community).

> It seems that the changes you've made are breaking backwards compatibility, is that on purpose? can you explain why it's necessary?

Yes the changes are done on purpose.
For `sentinelone-reactivate-site`:
  - we need to send data which will include expiration or the unlimited flag.

For `sentinelone-create-white-list-item`:
  - we need to send both group_ids and site_ids. otherwise api call fails.

guykeller (approved):

PR #14460, opened 2021-08-22
Feed no update tag

adi88d:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40918
https://github.com/demisto/etc/issues/38056
https://github.com/demisto/etc/issues/24110

## Related Pull requests
https://github.com/demisto/content/pull/13940

## Description
- Improved implementation of fetch indicators using ETAG header:
1. On the first request we save the `ETag` and `Last-Modified` values from the response headers in the integration LastRun.
2. Next time we want to fetch indicators and we have the`ETag` and `Last-Modified` values, we need to do the request and set the `If-None-Match` header with the  `ETag` value and the `If-Modified-Since` header with the  `Last-Modified` value.
3. If the request status code is 304 - that means there is no new indicators to fetch.

- Use `createIndicators` method with `noUpdate` only when demisto version is higher than 6.5.0

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ApiModules/Scripts/HTTPFeedApiModule/HTTPFeedApiModule.py
```diff
@@ -280,34 +291,37 @@ def custom_fields_creator(self, attributes: dict):
...
     last_modified = response.headers.get('Last-Modified')
 
-    set_integration_context({'last_modified': last_modified, 'etag': etag})
+    context[url] = {'last_modified': last_modified, 'etag': etag}
+    set_integration_context(context)
 
-    if old_etag and old_etag != etag:
-        demisto.debug('New indicators fetched - the ETag value has been updated,'
-                      ' createIndicators will be executed with noUpdate=False.')
+    if response.status_code == 304:
```
glicht, line 316: This logic should be done before saving integration context. When 304 is returned server may respond  with empty etag.

glicht (changes requested):
Missing use of `If-Modified-Since` when `Last-Modified` is used. see: https://httpwg.org/specs/rfc7232.html#header.if-modified-since

glicht (changes requested):
@yaakovi the whole `noupdate` logic depends on the server version. I don't see where this is taken into account.

glicht:
@yaakovi @adi88d what is the status here?

adi88d:
@glicht @yaakovi The pR is ready and waiting for review
I invited you for a demo on Wednesday 20.10.21

tomneeman151293:
Need to see why tests of BitcoinAbuse weren't collected for this PR. as they're failing since then

PR #14462, opened 2021-08-22
Autofocus url support and error suppression 

merit-maita:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35793
fixes: https://github.com/demisto/etc/issues/32007

## Description
added support for urls with non ascii charaters in the url command.

## Screenshots

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/AutoFocus/ReleaseNotes/1_3_7.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Palo Alto Networks AutoFocus v2
+- Added support for urls with non ascii characters in the *url* command.
```
yaakovi, line 4:
```suggestion
- Added support for URLs with non-ASCII characters in the ***url*** command.
```

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -1118,6 +1118,17 @@ def resolve_ip_address(ip):
     return None
 
 
+def convert_url_to_ascii_character(url_name):
+    regex_groups = re.search('([a-zA-Z\W]*)([^a-zA-Z\W]*)([a-zA-Z\W]*)', url_name)
```
yaakovi, line 1122: what if there is more than one special character (for example, `www.MüMünchen.com`)?

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2_test.py
```diff
@@ -398,3 +398,16 @@ def test_create_relationships_list():
         assert relation.get('entityA') == 'Test'
         assert relation.get('entityB') == expected_name_entity_b[i]
         assert relation.get('entityBType') == expected_entity_b_types[i]
+
+
+URLS_LIST = [
+    ("www.München.com", "www.Mxn--tdanchen.com"),
+    ("www.example.com", "www.example.com")
```
yaakovi, line 405:
```suggestion
    ("www.example.com", "www.example.com"),
```

yaakovi (approved):

bakatzir (commented):
I would add the JAP example in the referenced issue to the UT, and not just German `ü`, to make sure that it tends to the issue that was raised.

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
```diff
@@ -337,8 +339,8 @@ def parse_response(resp, err_operation):
             return return_error(err_msg)
     # Unexpected errors (where no json object was received)
     except Exception as err:
-        err_msg = f'{err_operation}: {err}'
-        return return_error(err_msg)
+        demisto.results(f'{err_operation}: {err}')
+        sys.exit(0)
```
bakatzir, line 343: is this changed intentionally? if so please explain why

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.yml
```diff
@@ -30,6 +30,11 @@ configuration:
   name: proxy
   required: false
   type: 8
+- display: Suppress errors for non found indicators
+  name: handle_error
+  required: false
+  defaultvalue: true
```
bakatzir, line 36: by default this should be false to not break BC.

Packs/AutoFocus/ReleaseNotes/1_3_7.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Palo Alto Networks AutoFocus v2
+- Added support for URLs with non-ASCII characters in the ***url*** command.
+- Added the parameter **handle_error** to have an option to handle errors with 404, 409 codes.
```
bakatzir, line 5: please use the display name, as `handle_error` is not visible/helpful to customers.

bakatzir (changes requested):
please also remove null entries in the markdown table.

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
merit-maita: yes it is, this is one of the fixes we wanted to apply, since we dont want the costumer to get an error just a note that irt does not exist

bakatzir (approved):
nice work.
please make sure that the UT pass, and that the TPB of AFv2 runs successfully in this build, and not skipped.

Packs/AutoFocus/Integrations/AutofocusV2/AutofocusV2.py
anara123: sys.exit(0) is not a good solution. Instead I would raise a specific exception and in the main will catch it and ignore.

PR #14469, opened 2021-08-22
setGridField: Allow column names to have underscores

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39641

## Description
Column names were transformed to `flatcase` (CLI name), although the server allows (see `utils/strings.js/stripToColumnName`). This caused data of columns whose name include underscores to not be shown on the grid. Changing them to lower case (rather than converting to camelCase, which would remove underscores) allows using column names that include underscores.
 
- [x] 5.0.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_4_24.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_4_24.md
@@ -1,0 +1,5 @@
+#### Scripts
+
+##### SetGridField
+
+- Grid column names are allowed to include underscores (`_`). Underscores are no longer removed from provided column names. 
--- a/Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
+++ b/Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
@@ -28,7 +28,8 @@ def normalized_string(phrase: str) -> str:
 
 
 def normalized_column_name(phrase: str) -> str:
-    """ Normalize columns or Grid to connected word in lower-case.
+    """ Normalize columns or Grid to connected word in lowercase, to match the logic of stripToClumnName() from
+        the client's `strings.js` and the server logic.
 
     Args:
         phrase: Phrase to normalize.
@@ -295,7 +296,7 @@ def build_grid_command(grid_id: str, context_path: str, keys: List[str], columns
         raise DemistoException(f'The number of keys: {len(keys)} should match the number of columns: {len(columns)}.')
     # Get old Data
     old_table = get_current_table(grid_id=grid_id)
-    # Change columns to all lower case. Not using `normalize()` as underscores are allowed in columns names.
+    # Change columns to all lower case (underscores allowed).
     columns = [normalized_column_name(phrase) for phrase in columns]
     # Create new Table from the given context path.
     new_table: pd.DataFrame = build_grid(context_path=context_path,
--- a/Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
+++ b/Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
@@ -122,7 +122,7 @@ def get_current_table(grid_id: str) -> pd.DataFrame:
 
 
 @logger
-def validate_entry_context(context_path, entry_context: Any, keys: List[str], unpack_nested_elements: bool):
+def validate_entry_context(context_path: str, entry_context: Any, keys: List[str], unpack_nested_elements: bool):
     """ Validate entry context structure is valid, should be:
         - For unpack_nested_elements==False:
             1. List[Dict[str, str/bool/int/float]]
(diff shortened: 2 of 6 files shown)


Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
```diff
@@ -277,8 +277,8 @@ def build_grid_command(grid_id: str, context_path: str, keys: List[str], columns
         raise DemistoException(f'The number of keys: {len(keys)} should match the number of columns: {len(columns)}.')
     # Get old Data
     old_table = get_current_table(grid_id=grid_id)
-    # Normalize columns to match connected words.
-    columns = [normalized_string(phrase) for phrase in columns]
+    # Change columns to all lower case. Not using `normalize()` as underscores are allowed in columns names.
+    columns = [phrase.lower() for phrase in columns]
```
yaakovi, line 281: what about other special characters that need to be removed (like `$` or `%`)?

yaakovi (dismissed):

Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
dorschw: was just about to commit :)
By the way, the current `normalize` that is applied to keys does not remove them, but on the other hand none of them is an accepted char (unlike `_` for columns). Should I open an issue for it?

Packs/CommonScripts/Scripts/SetGridField/SetGridField_test.py
```diff
@@ -89,9 +89,13 @@ def test_build_grid(datadir, mocker, keys: list, columns: list, dt_response_json
     ).to_dict()
 
 
+should_be_removed_by_normalize_col = ' \n &$#%?!*;׳()🦦ץ' * 20 + "this should be truncated, more than 255 chars"
```
DeanArbel, line 92: Please add a test that contains valid non-alphanumeric keys such as +.

Packs/CommonScripts/ReleaseNotes/1_4_24.md
```diff
@@ -0,0 +1,5 @@
+#### Scripts
+
+##### SetGridField
+
+- Grid column names are allowed to include underscores (`_`). Underscores are no longer removed from provided column names.
```
DeanArbel, line 5:
```suggestion
#### Scripts

##### SetGridField
- **Breaking Change:** Added support for **columns** with underscores (`_`). Previously underscores in **columns** were removed automatically by the script.
```

Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
```diff
@@ -21,12 +21,30 @@ def normalized_string(phrase: str) -> str:
     Examples:
         >>> normalized_string("TestWord")
         "testword"
-        >>> normalized_string("hello_world")
-        "hello_world"
+        >>> normalized_string("helloworld")
```
DeanArbel, line 24: the input should still be `hello_world`.

Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
```diff
@@ -21,12 +21,30 @@ def normalized_string(phrase: str) -> str:
...
+    Returns:
+        str: Normalized phrase.
+
+    Examples:
+        >>> normalized_string("Test Word!@#$%^&*()-=+")
+        "testword"
+        >>> normalized_string("hello_world@")
+        "hello_world"
+    """
+    return re.sub(r'[^a-zA-Z\d_]', '', phrase[:255]).lower()
```
DeanArbel, line 45: Please add a note that this matches the logic of `stripToClumnName()` from the client's `strings.js`.

Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
```diff
@@ -277,8 +295,8 @@ def build_grid_command(grid_id: str, context_path: str, keys: List[str], columns
         raise DemistoException(f'The number of keys: {len(keys)} should match the number of columns: {len(columns)}.')
     # Get old Data
     old_table = get_current_table(grid_id=grid_id)
-    # Normalize columns to match connected words.
-    columns = [normalized_string(phrase) for phrase in columns]
+    # Change columns to all lower case. Not using `normalize()` as underscores are allowed in columns names.
```
DeanArbel, line 298: This comment is no longer true. Please correct it.

DeanArbel (changes requested):
Please make sure to test valid non-alphanumeric characters.

DeanArbel (approved):
Good work

Packs/CommonScripts/Scripts/SetGridField/SetGridField.py
yaakovi: should the slicing happen before the replacement?
currently, the final result may be less than 255 letters.

dorschw: yes, see [here](https://code.pan.run/xsoar/web-client/-/blob/master/src/util/strings.js#L461-466)

PR #14492, opened 2021-08-23
setGridField: undo column name truncation

dorschw:
## Status
- [x] Ready

## Description
Soon after merging #14469, we noticed that column names are only truncated by the web client when shown to user, but not when sent to the server. To not break column names longer than 255 chars, we decided to revert this truncation. 

- [x] 5.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


Packs/CommonScripts/ReleaseNotes/1_4_24.md
```diff
@@ -1,4 +0,0 @@
-#### Scripts
```
DeanArbel, line 1: Don't remove this, as it's still a valid release note.

Packs/CommonScripts/ReleaseNotes/1_4_25.md
```diff
@@ -0,0 +1,4 @@
+#### Scripts
+
+##### SetGridField
+- Column names are no longer truncated after 255 chars (added on very recent v1.4.24)
```
DeanArbel, line 4:
```suggestion
- Add back support for column names with more than 255 chars (support was removed in v1.4.24).
```

DeanArbel (approved):
Please undo the RN delete.

Packs/CommonScripts/ReleaseNotes/1_4_24.md
dorschw: unintentional, restoring.

PR #14508, opened 2021-08-24
Add markdown images support in sanePdfReport

davidbinyamin:
## Status
- [x] Ready

## Related Issues
Related: https://github.com/demisto/etc/issues/40210

## Related PRs
branch | PR
------ | ------
Sane-reports PR | https://github.com/demisto/sane-reports/pull/183
Server MR |  https://code.pan.run/xsoar/server/-/merge_requests/16194

## Description
Full solution is described in the related issue.
in `SanePdfReport.py`, i added the ability to listen to markdown images requests (from `reportServer`) and fetch the requested image from demisto.

Relevant server changes will be present only in 6.5 but this change is BC

   - [x] No


davidbinyamin:
@aviadl @dorsha FYI last piece of the puzzle

glicht (changes requested):
We should add logic to disable the image server if running on Server version below 6.5.

aviadl:
@glicht why not have 2 scripts with from -to version on them ?

davidbinyamin:
@glicht @aviadl what is the verdict regarding version control?

glicht:
> @glicht why not have 2 scripts with from -to version on them ?

It's an option. We can do that too. But from previous experience this introduces extra work as you need to fix things in 2 places when dealing with bugs. From what I see the logic changes are minimal (but I may be wrong), so I think 1 script with a version check will be easier to maintain. Anyway @davidbinyamin we can do a short call to discuss if needed.

davidbinyamin:
Adding the ver check is pretty straight forward, ill do it

glicht (approved):

PR #14513, opened 2021-08-24
Make some fields available to all incident types

nb-pdragoi:

## Status
- [x] In Progress
- [ ] Ready

## Description
Some fields are only associated with certain incident types, while I feel they are a little too general to be used only for those. This change makes them available to every type.
For Example, the "email", "email to" and "email from" fields. These might be used in an extremly wide range of cases (either manual ones or fetched by an integration) and its impossible to expect every XSOAR customer/user to: 

- add the incident type to the main repository
- add that incident type to the associatedTypes list
- wait for a review

Even if they did, the list will grow very large and some incident types might not even be useful for other customers, so it would be pointless to have them in the main repository.

Of course, these fields can be modified inside the XSOAR UI, and add the incident type in there. The problem with this is that the  field would get rewritten at the next pack update/reinstall and all the changes would be lost. In addition to this, even if the custom incident type is associated to the field, it is still shown as a system field and not exported as a custom content, making it difficult to version control internally.

One workaround would be the ability to have multiple fields inside xsoar with the same "displayname". So you would have, for example, "incidentfield_email" and "incidentfield_email_custom" as ids, but both of them would have the "name" equal to "email". This is a little harder to do, since it will require a pretty big platform change (for example, you should be able to tell the difference between which field is which in the field mapping section of a playbook task).

- [ ] 5.5.0
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Phishing/IncidentFields/incidentfield-Attachment_Count.json
+++ b/Packs/Phishing/IncidentFields/incidentfield-Attachment_Count.json
@@ -23,13 +23,7 @@
  "content": true,
  "group": 0,
  "hidden": false,
- "associatedTypes": [
-  "Phishing"
- ],
- "systemAssociatedTypes": [
-  "Phishing"
- ],
- "associatedToAll": false,
+ "associatedToAll": true,
  "unmapped": false,
  "unsearchable": false,
  "fromVersion": "5.0.0"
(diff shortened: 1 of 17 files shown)


abaumgarten:
@nb-pdragoi Thank you for your contribution and for your valid and the good point you have raised. After some conversations we had, we decided that this issue is going to be handled by a platform feature that will let the user the freedom to define and edit any display name field for incident fields. For the option to move incident fields between packs, it might cause us some other issues so we don't support it on the content side. Closing this PR, thanks again.

PR #14526, opened 2021-08-25
CommonServerPython - check if session exist before trying to close it

Itay4:
## Status
- [x] Ready

## Description
if an integration uses the `BaseClient` without actually using `_http_request` (such as ServiceNowV2 - no idea why) then the following would be found in the logs:
```
Traceback (most recent call last):
  File "<string>", line 7061, in __del__
AttributeError: 'Client' object has no attribute '_session'
```

its not an actual problem but better to avoid it by checking if there is a session to close 

   - [x] No


Packs/Base/ReleaseNotes/1_13_25.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### CommonServerPython
+- Fixed an issue where there was an attempt to close a session that was initialized.
```
yaakovi, line 4:
```suggestion
- Fixed an issue where there was an attempt to close a session that was not initialized.
```

yaakovi (approved):

PR #14529, opened 2021-08-25
Zscaler - handle returned URLs protocols

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40700

   - [x] No

## Must have
- [x] Tests


Packs/Zscaler/Integrations/Zscaler/Zscaler.py
```diff
@@ -389,8 +389,16 @@ def url_lookup(args):
...
     pre_table_data = []
+    urls_list = argToList(url)
     for data in raw_res:
         suspicious_categories = ['SUSPICIOUS_DESTINATION', 'SPYWARE_OR_ADWARE']
+        res_url = data.get('url')
+        for url in urls_list:
+            # since zscaler expects to recieve a URL without the protocol, we omit it in `lookup_request`
+            # in the response, the URL is returned as it was sent, so we add back the protocol by replacing
+            # the URL retruned with the one we got as an argument
+            if res_url in url:
```
yaakovi, line 400: maybe use `endswith`?
what if the URLs looks like the following:
* https://google.com
* https://mail.google.com

in both cases, we will end up with `https://google.com`

yaakovi (approved):

Itay4:
build fails on validation of skipped test due to missing instance
merging

PR #14533, opened 2021-08-25
SDP Create Request, fix POST request

bdanjoux:
This is a fix In the Service Desk Plus Pack.
This fix concerns the service-desk-plus-request-create Integration command.

The current POST query for request creation puts input_data in the URL parameters.
Although this is supported, it may create errors mentionning 'URI too long' if input data is too long.
This commit introduces the use of the POST request body instead of the URL parameters to pass the input_data


bdanjoux:
I made a mistake and will start again from scratch.

PR #14561, opened 2021-08-27
Symantec DLP - fix handling of fetch limit

Itay4:
## Status
- [x] Ready

## Related Issues
none yet

## Description
in case the fetch limit was set to less than the number of incidents returned from dlp, and the last run would not get updated

   - [x] No

## Must have
- [x] Tests


yaakovi (approved):

Itay4:
build fails on:
1. non updated docker image - not updating due to insufficient test coverage
2. skipped test due to missing instance

merging

PR #14578, opened 2021-08-29
fixing dups and attribute naming

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Description
fixing attribute errors

PR #14579, opened 2021-08-29
fix upload-flow bug in collect_content_items

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/content/pull/14578

## Description
As can be seen in the following [upload](https://code.pan.run/xsoar/content/-/jobs/7335091), packs are not failing but the pack processing is continued (`Screenshot 1`). This should not happen and we should see failure in that Pack method, as can be seen in the same build, packs are not listed as failed to upload. Bug reproduce is showed in `Screenshot 2`.

## Screenshot 1
image

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Tests/Marketplace/marketplace_services.py
+++ b/Tests/Marketplace/marketplace_services.py
@@ -1659,7 +1659,7 @@ class Pack(object):
             logging.exception(f"Failed collecting content items in {self._pack_name} pack")
         finally:
             self._content_items = content_items_result
-            return task_status, content_items_result
+            return task_status
 
     def load_user_metadata(self):
         """ Loads user defined metadata and stores part of it's data in defined properties fields.


Tests/Marketplace/marketplace_services.py
```diff
@@ -1659,7 +1659,7 @@ def collect_content_items(self):
             logging.exception(f"Failed collecting content items in {self._pack_name} pack")
         finally:
             self._content_items = content_items_result
-            return task_status, content_items_result
+            return task_status
```
yaakovi, line 1662: don't we need the other return value?
wouldn't it be better to fix the condition in the calling function?
```python
if not results[0]:
    print('error')

guyfreund: Nope, we set the value in the line before as an attribute of the Pack object and use it later on, therefore it is not needed.

Noy-Maimon (approved):

PR #14587, opened 2021-08-30
Update Threat Intel objects and their score

ChanochShayner:
## Status
- [x] Ready

## Description
Update Threat Intel objects and their score:
Threat Actor - Malicious, Infrastructure - Suspicious.

- [x] No


Bargenish (approved):
@ChanochShayner  no need to add

PR #14599, opened 2021-08-31
Cb fetch fix

JasBeilin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40844

## Description
Fixed fetch-incidents

- [x] 5.5.0

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Carbon_Black_Enterprise_Response/Integrations/CarbonBlackResponseV2/CarbonBlackResponseV2.py
```diff
@@ -728,49 +740,60 @@ def endpoint_command(client: Client, id: str = None, ip: str = None, hostname: s
...
+    if feedname:
+        query_params['feedname'] = feedname
+
     if status:
         for current_status in argToList(status):
             demisto.debug(f'{INTEGRATION_NAME} - Fetching incident from Server with status: {current_status}')
-            res = client.get_alerts(status=current_status, feedname=feedname)
+            query_params['status'] = current_status
+            # we create a new query containing params since we do not allow both query and params.
+            res = client.get_alerts(query=_create_query_string(query_params), limit=max_results)
```
DeanArbel, line 779: Why add it under the `query_params` instead of using the function args `feedname` and `status`?
Should they be removed from the function signature?

DeanArbel (approved):
Looks good. Please go over my question.

Packs/Carbon_Black_Enterprise_Response/Integrations/CarbonBlackResponseV2/CarbonBlackResponseV2.py
JasBeilin: as I stated in the comment (if it is not clear I would change it)- we create a new query containing params since we do not allow both query and params.
The get_alerts function is being used for other command as well, where it can get the feedname or status as parameters, or a query.
Since we do not allow both, It will fail if we provide in the fetch both query and other params, and since we must provides dates in the query and it is not a user-created query, we create a new query with both dates and the user params.

PR #14642, opened 2021-09-01
Pop3 fix ssl default value

tomer-pan:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40992

## Description
Changing the ssl checkbox default value from "'True'" to "true" - this resolves an issue with setting up a new instances.

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
+++ b/Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
@@ -51,7 +51,7 @@ script:
   script: '-'
   type: python
   subtype: python2
-  dockerimage: demisto/python:2.7.18.22912
+  dockerimage: demisto/python:2.7.18.24019
 beta: true
 tests:
 - MailListener-POP3 - Test
--- a/Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
+++ b/Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
@@ -20,7 +20,7 @@ configuration:
   name: password
   required: true
   type: 4
-- defaultvalue: 'True'
+- defaultvalue: true
   display: Use SSL connection
   name: ssl
   required: false
--- a/Packs/MailListener_-_POP3/ReleaseNotes/1_0_5.md
+++ b/Packs/MailListener_-_POP3/ReleaseNotes/1_0_5.md
@@ -1,4 +1,4 @@
 
 #### Integrations
 ##### MailListener - POP3 Beta
-- SSL Connection default value changed from **'True'** to **true**.
+- SSL connection default value changed from **'True'** to **true**.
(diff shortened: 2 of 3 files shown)


David-BMS (approved):

Packs/MailListener_-_POP3/ReleaseNotes/1_0_5.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### MailListener - POP3 Beta
+- SSL connection default value changed from **'True'** to **true**.
```
bakatzir, line 4:
```suggestion
- Fixed an issue where the *Use SSL connection* parameter default value was not set properly.
```

Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
```diff
@@ -20,7 +20,7 @@ configuration:
   name: password
   required: true
   type: 4
-- defaultvalue: 'True'
+- defaultvalue: true
```
bakatzir, line 23: I think that be default we set it to `false`. @Itay4 ?

Itay4: it should be `true` as we want users to use ssl by default, but this is a breaking change and should be mentioned in the release notes as such

bakatzir (approved):

Packs/MailListener_-_POP3/ReleaseNotes/1_0_5.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### MailListener - POP3 Beta
+-  **Breaking changes:** Update the **Use SSL connection** parameter default value to *true*.
```
bakatzir, line 4:
```suggestion
-  **Breaking changes:** Update the **Use SSL connection** parameter default value to *true*.
- Updated the Docker image to: *demisto/python:2.7.18.24019*.
```

Packs/MailListener_-_POP3/Integrations/MailListener_POP3/MailListener_POP3.yml
tomer-pan: @bakatzir @Itay4  I don't think this is a breaking change.
I have tested it on my local server - I have configured an instance without SSL, then changed the marketplace url to use the updated url (copied it from the gitlab pipeline), then updated the pop3 pack, refreshed and the existing instance didn't change. New instance is now working as expected though (which means the change works only on new instances). Should we revert the release notes to be without the "Breaking changes"?

tomer-pan: Removed it.

PR #14585, opened 2021-08-30
ParseEmailFiles -  roll back to multiple encoding part

tallieber:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: [40877](https://github.com/demisto/etc/issues/40877)

## Description
We have some issues with decode headers  like this: `This =?UTF-8?B?VGVzdMKu?= passes` so we added a function for checking mime encoding words and we expanded it to handle multiple encoding: ```
 =?iso-2022-jp?B?GyRCJWEhPCVrLSEkSHxxGyhC?= '
        '=?iso-2022-jp?B?GyRCRnxLXDhsSjg7eiQsST08KCQ1JGwkSiQkSjg7eiROJUYlOSVIGyhC?=
```. 
Since every header that contains multiple encoding words and previously could handle with `decode_header(s)`  will now  goes to the new code.
We will prefer to have a mistakes once a time in case like this: 
'This is test =?iso-2022-jp?B?GyRCJWEhPCVrLSEkSHxxGyhC?= '
        '=?iso-2022-jp?B?GyRCRnxLXDhsSjg7eiQsST08KCQ1JGwkSiQkSjg7eiROJUYlOSVIGyhC?='

I assume we should check it and handle it with the new repository. @moishce 

## Screenshots
Paste here any images that will help the reviewer

## Minimum version of Cortex XSOAR
- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation

PR #14754, opened 2021-09-05
ParseEmailFiles fails on demisto.debug line

tallieber:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [41085](https://github.com/demisto/etc/issues/41085)

## Description
We use demisto.debug to print an error where a string could not be decoded. the problematic strings contains characters that raised a server error: `Error from Scripts is : Script failed to run: 
Error: Failed to decode (loop) data from docker code script: invalid character '{' after object key:value pair
This may happen when you don't return an explicit entry (or entries) from a script.`
Will open an issue for this problem.


Packs/CommonScripts/Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3432,8 +3432,7 @@ def convert_to_unicode(s, is_msg_header=True):
                 try:
                     res += decoded_s.decode(encoding).encode('utf-8')
                 except UnicodeDecodeError:
-                    demisto.debug('Failed to decode encoded_string:'
-                                  'encoding: {}, encoded_text: {}'.format(encoding, decoded_s))
+                    demisto.debug('Failed to decode encoded_string with encoding {}'.format(encoding))
```
Itay4, line 3435: lets log here `decoded_s.decode(encoding, errors='replace').encode('utf-8')`

Itay4 (commented):
please add test to the test playbook that reproduces the issue

tallieber:
@Itay4 I added a test playbook, in the playbook we see that there is a header that stays encoded (`TestSegurança ?� =?UTF-8?Q?Informação`). 
The value we had prior  to 1_4_26 release is `Test =?UTF-8?Q?Seguran=C3=A7a=20?=da =?UTF-8?Q?Informa=C3=A7=C3=A3o?=`, not decode at all.
We revert most of the code on 1_4_28 ([PR](https://github.com/demisto/content/pull/14585)) and now will handle some cases with replace as in the first line.  
So for now as we said in the revert we will not handle this kind of encoded text and we should handle it properly when re-write this code. 
For the error which raised by this bug: [41085](https://github.com/demisto/etc/issues/41085), this PR is taking care of it, and the test playbook is relevant for it too.

Packs/CommonScripts/ReleaseNotes/1_4_33.md
```diff
@@ -0,0 +1,5 @@
+
+#### Scripts
+##### ParseEmailFiles
+- Fixed an issue where ParseEmailFiles fails while print error to log.
```
Itay4, line 4:
```suggestion
- Fixed an issue where the automation script failed to process MIME encoded-words.
```

TestData/test-subject-raised-error-on-log.eml
```diff
@@ -0,0 +1,18 @@
+Subject: Test
```
Itay4, line 1: this should be in the `test_data` subdir of the script dir

Packs/CommonScripts/TestPlaybooks/playbook-ParseEmailFiles-test.yml
```diff
@@ -804,12 +783,129 @@ tasks:
...
+    skipunavailable: false
+    quietmode: 0
+  "25":
+    id: "25"
+    taskid: 8e82f75f-87d0-48fc-8eed-6b829b328a57
+    type: regular
+    task:
+      id: 8e82f75f-87d0-48fc-8eed-6b829b328a57
+      version: -1
+      name: Subject with special chars
```
Itay4, line 821:
```suggestion
      name: Organization header with MIME encoded-word
```

Itay4 (approved):
please note my comments

PR #14755, opened 2021-09-05
updated docker image for gitlab

DinaMeylakh:
Updating gitlab to a new docker image

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc#35419

### Instance Tests
https://code.pan.run/xsoar/content/-/pipelines/1677489

### Bucket Upload
https://code.pan.run/xsoar/content/-/pipelines/1677747

### SDK Nightly
https://code.pan.run/xsoar/content/-/pipelines/1680866

### Nightly
https://code.pan.run/xsoar/content/-/pipelines/1692528
https://code.pan.run/xsoar/content/-/jobs/7720311
As compared to master's nightly:
https://code.pan.run/xsoar/content/-/jobs/7717475


.gitlab/ci/global.yml
```diff
@@ -96,16 +97,24 @@
     - rm -rf $ARTIFACTS_FOLDER/logs
     - mkdir -p $ARTIFACTS_FOLDER/logs
     - section_end "Creating new clean logs folder"
+    - source $NVM_DIR/nvm.sh
+    - nvm use default
+    - npm ci --cache .npm --prefer-offline
+    - npm install jsdoc-to-markdown@5.0.3 -g  # disable-secrets-detection
```
avidan-H, line 103: Can we make this its own section?

avidan-H: Also, it may be helpful/relevant to additionally log the output to a logs file. I also think it may be useful to add a condition to test and see if `nvm` is installed (something like `if test $(command -v nvm); then ...`) and if not, add informative messages etc. etc. if you know what I mean - do you agree?

avidan-H: I have a few questions about this section,
1. can't we have the image so that nvm is installed and sourced upon new shell instantiation (so we don't need the line `source $NVM_DIR/nvm.sh`)?
2. do you have it in the docker image so that `nvm` is configured so that `.nvrmc` files are respected as documented [here](https://github.com/nvm-sh/nvm#bash)? I ask cause it would make its usage more flexible for different repos.
3. why do we need to install `jsdoc-to-markdown` explicitly here? If it's something needed globally, should we not install it in the docker image already?

.hooks/bootstrap
```diff
@@ -118,7 +118,7 @@ else
                 exit 1
             fi
         fi
-        virtualenv -p $PY3CMD venv
+        virtualenv --system-site-packages -p $PY3CMD venv
```
avidan-H, line 121: Can you please explain this change?

avidan-H (commented):
Looks good for the most part - I made some suggestions, let me know what you think - if you agree or disagree. Otherwise, I was just looking to understand the why for some of your changes before approving - I commented with questions where this applies.

.gitlab/ci/global.yml
DinaMeylakh: 0. About your second comment: I think if we take into account a case where someone changes the image to an image that doesn't use nvm it will either fail (if we leave it like this) or output a message and continue (if we change it to what you suggested). Since the nvm is installed so other repos could use this image more easily, I would rather it fail and make the developer think about the changes. 

1. I actually think you are right and we don't need this line anymore. Will push without it.

2. If you're talking about the bash script being inside the .bashrc - then no. Since we only use one nvm version, I see no point in guessing what other repos will need when installing other versions, especially since nvm.sh provides various functions to deal with special needs like ```nvm_find_nvmrc()``` for example. 

3. We need jsdoc-to-markdown installed explicitly since we use the ```jsdoc2md``` CLI command. In order for a CLI command to be installed properly through a node module installation it needs the "-g" flag which means its installed globally regardless of the package.json file (which contains the packages that will be installed in the npm ci command). 
Since other repos using this image could use different versions of jsdoc-to-markdown and might have no need for the ```jsdoc2md``` CLI command, I think it's better if it is installed explicitly like this.

.hooks/bootstrap
DinaMeylakh: Basically we wanted to get rid of this line in our code
`venv/bin/pip3 install -r .circleci/build-requirements.txt >> $ARTIFACTS_FOLDER/logs/installations.log 2>&1`
and here I did that by installing the build-requirements python packages inside the python installed on the image. 
If we use virtualenv, the python won't be able to access the system installed python packages, only the ones installed on the venv. This flag lets us use both: if the python package is not installed in the virtualenv the python interpreter will look for packages installed outside of it next. The quote from the virtualenv documentation is:
`Give the virtual environment access to the system site-packages dir.`
https://virtualenv.pypa.io/en/latest/cli_interface.html?highlight=site-packages#system-site-packages

That said I think I might need to add a flag also when installing the packages. TBC

.gitlab/ci/global.yml
avidan-H: > 2
I don't follow your response. Let's resolve over slack.
> 3
Good point. 👍🏼

.hooks/bootstrap
avidan-H: 👍🏼

.gitlab/ci/global.yml
```diff
@@ -65,12 +64,11 @@
 .install_venv: &install_venv
   - section_start "Installing Virtualenv" --collapsed
   - |
-    if [ -f "./venv/bin/activate" ]; then
+    if [ -f "./venv/bin/activatemmm" ]; then
```
avidan-H, line 67: typo fix
```suggestion
    if [ -f "./venv/bin/activate" ]; then
```

avidan-H (approved):
Just commit the typo fix and you're good.

PR #14778, opened 2021-09-08
Fixed an issue where proxy settings haven't been given in BasicClient.

spearmin10:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
Unable to connect Unit42 server via a web proxy.

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/FeedUnit42/Integrations/FeedUnit42/FeedUnit42.py
+++ b/Packs/FeedUnit42/Integrations/FeedUnit42/FeedUnit42.py
@@ -35,7 +35,8 @@ class Client(BaseClient):
             api_key: unit42 API Key.
             verify: boolean, if *false* feed HTTPS server certificate is verified. Default: *false*
         """
-        super().__init__(base_url='https://stix2.unit42.org/taxii', verify=verify)
+        super().__init__(base_url='https://stix2.unit42.org/taxii', verify=verify,
+                         proxy=argToBoolean(demisto.params().get('proxy')))
         self._api_key = api_key
         self._proxies = handle_proxy()
         self.objects_data = {}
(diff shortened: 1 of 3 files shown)


ShahafBenYakir:
@spearmin10 Look at the current `__init__` function there's a [statement](https://github.com/demisto/content/blob/master/Packs/FeedUnit42/Integrations/FeedUnit42/FeedUnit42.py#L40):
```python
self._proxies = handle_proxy()
```
This should handle the proxy behavior including reading the value from the params. 
Do you think it is not working have you had proxy issues with this feed?

spearmin10:
> @spearmin10 Look at the current `__init__` function there's a [statement](https://github.com/demisto/content/blob/master/Packs/FeedUnit42/Integrations/FeedUnit42/FeedUnit42.py#L40):
> 
> ```python
> self._proxies = handle_proxy()
> ```
> 
> This should handle the proxy behavior including reading the value from the params.
> Do you think it is not working have you had proxy issues with this feed?

@ShahafBenYakir No. It doesn't work.
In `super().__init__(base_url='https://stix2.unit42.org/taxii', verify=verify)`, the `proxy` argument is given with `False` as default value. See: `def __init__(self, base_url, verify=True, proxy=False, ok_codes=tuple(), headers=None, auth=None):`

When passing `False` to `proxy` argument, `http_proxy` and `https_proxy` in the os.environment are removed in `super().__init__`, so `handle_proxy()` returns an emtpy value after the initialization.

ShahafBenYakir:
@spearmin10 You are right I missed that. Thanks 👍

tallieber (approved):

PR #14779, opened 2021-09-08
Cisco Umbrella Investigation - handle list of domains

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41151
fixes: https://github.com/demisto/etc/issues/24338

   - [x] No

## Must have
- [x] Tests


Itay4:
TDD: see test failure in https://code.pan.run/xsoar/content/-/jobs/7500021

ShahafBenYakir:
@Itay4 How come this worked from CLI and not from a playbook? I would expect the opposite (playbook will run the command once for each domain - which is supported, whereas the CLI receives the argument as a list and this is obviously not supported). 
Also, this doesn't look like a bug more like a feature I don't see the description or anything of that sort that says the command supports a list.

PR #14915, opened 2021-09-16
Allow attaching files that do not have an explicit file name in XSOAR

martinspielmann:
## Status
- [ ] In Progress
- [x] Ready

## Description
When providing embedded attachments to the `send-mail` command useing the `attachCIDs` parameter, sending the email fails when the attachment does not have a `uploaded_file_name`. 
See screenshot 1.
With this change, the file EntryID is used when no file name provided. This fixes the issue for embedded attachments.
For regular (not embedded) attachments everything works like before.

There are many situation when `uploaded_file_name` is empty (''), e.g., when the file entry is created by `!Base64ListToFile`.
See screenshot 2.

Looking at the `zip` function and comment a few lines above this change, it seems like you also stumbled over that problem at some point. However, you did not use the pre-filled `attach_name` here.
Using the `attach_name` pre-filled with the file EntryID if `uploaded_file_name` is empty fixes the issue.

## Screenshots
Screenshot 1: Before the change, when `provided_names` is false the attachment's name attribute is set to an empty string when `uploaded_file_name` is empty. This will cause error 400 in graph API. 
grafik

Screenshot 2: output of `demisto.getFilePath()` for a file created using `!Base64ListToFile`. name attribute is empty.
grafik

- [ ] 5.5.0
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/MicrosoftGraphMail/Integrations/MicrosoftGraphMail/MicrosoftGraphMail.py
+++ b/Packs/MicrosoftGraphMail/Integrations/MicrosoftGraphMail/MicrosoftGraphMail.py
@@ -381,7 +381,7 @@ class MsGraphClient:
                 '@odata.type': cls.FILE_ATTACHMENT,
                 'contentBytes': b64_encoded_data.decode('utf-8'),
                 'isInline': is_inline,
-                'name': attach_name if provided_names else uploaded_file_name,
+                'name': attach_name if provided_names or not uploaded_file_name else uploaded_file_name,
                 'size': file_size
             }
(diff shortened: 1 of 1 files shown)


barchen1:
It Will be fixed in https://github.com/demisto/content/pull/14939

barchen1:
Merged in https://github.com/demisto/content/pull/14939
closing.

PR #14916, opened 2021-09-16
[BugFix] StixParser Custom Fields

nericksen:

## Status
- [x] In Progress
- [ ] Ready

## Description
We would like to be able to search on the “Stix Package Id” and “Indicator ID” and populate the details to the layout, which are created when a STIX file is uploaded via the UI or API. I see in the !StixParser we extract this data from the Stix file and add it to the indicator custom fields 
```
entry["CustomFields"] = {"indicatorId": ind_id, "stixPackageId": pkg_id}
```
these are not OOTB indicator fields as far as I can tell.  If I create indicator fields with the same name the Threat Intel page shows them automatically populated, but they are not searchable and do not appear populated on the indicator layout.  Is this a bug? I don’t see how anyone would necessarily use these custom fields as it stands today at least for this use case.  As custom fields are referenced using all lowercase should the code be changed to the following and we add these as OOTB indicator fields? This would solve their particular issue but not sure if backwards compatibility is an issue here that would block consideration.

```
entry["CustomFields"] = {"indicatorid": ind_id, "stixpackageid": pkg_id}
```

## Screenshots
Screen Shot 2021-09-16 at 10 41 47 AM
Screen Shot 2021-09-16 at 10 41 02 AM
Screen Shot 2021-09-16 at 10 39 39 AM

- [ ] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

Not sure this may potentially break it if someone has created a custom scripts which reference the custom fields as they are formatted today.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/Scripts/StixParser/StixParser.py
+++ b/Packs/Base/Scripts/StixParser/StixParser.py
@@ -115,7 +115,7 @@ def create_indicator_entry(
     entry = dict()
     entry["indicator_type"] = indicator_type
     entry["value"] = value
-    entry["CustomFields"] = {"indicatorId": ind_id, "stixPackageId": pkg_id}
+    entry["CustomFields"] = {"indicatorid": ind_id, "stixpackageid": pkg_id}
     entry["source"] = source if source else ind_id.split("-")[0]
     entry["score"] = score
     # Times
--- a/Packs/Base/Scripts/StixParser/StixParser.py
+++ b/Packs/Base/Scripts/StixParser/StixParser.py
@@ -336,7 +336,7 @@ def stix2_to_demisto(stx_obj):
 
 def create_new_ioc(data, i, timestamp, pkg_id, ind_id):
     data.append({})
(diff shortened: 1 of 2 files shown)


nericksen:
Looks like `Packs/CommonScripts/Scripts/CreateIndicatorsFromSTIX/CreateIndicatorsFromSTIX.py` would also have to be updated if this is something we want to consider..

abaumgarten:
@yaakovi can you please approve this as well? I am not sure if it breaks BC or not

abaumgarten:
@nericksen thank you for your contribution. I think that the file you have changed has to be changed in another place as well.
In line 339 (master's version). Can you please edit this line as well? thanks. 
image (13)

nericksen:
Updated `StixParser.py` based on feedback above.  Also updated `Packs/CommonScripts/Scripts/CreateIndicatorsFromSTIX/CreateIndicatorsFromSTIX.py` since it calls `StixParser` command and attempts to access `stixPackageId` as well.

abaumgarten (approved):

PR #14940, opened 2021-09-19
IntSightUnitTestsFix

tomneeman151293:
possible fix to the instability unit test failures from IntSight


ShahafBenYakir (approved):
nice 👍

tomneeman151293:
@ShahafBenYakir I will say though that tests passed in build BEFORE the changes (but not locally). Tests now pass both locally and in build. Hopefully this is indeed a solution

PR #14984, opened 2021-09-22
Exclude protected branches

avidan-H:
## Status
- [x] Ready

## Description
Our maintenance script `delete_stale_non_contrib_branches.py` which deletes stale non-contribution branches will try to create time after time a posterity PR for stale branches which are protected (since it doesn't successfully delete the branch after creating the posterity PR and therefore the branch is collected as stale the next time the script is run). I've updated the script so that it does not consider protected branches.

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] Tests - N/A
- [x] Documentation

### DIFF
--- a/Utils/github_workflow_scripts/delete_stale_non_contrib_branches.py
+++ b/Utils/github_workflow_scripts/delete_stale_non_contrib_branches.py
@@ -17,7 +17,7 @@ print = timestamped_print
 
 def get_non_contributor_stale_branch_names(repo: Repository) -> List[str]:  # noqa: E999
     """Return the list of branches that do not have the prefix of "contrib/" without open pull requests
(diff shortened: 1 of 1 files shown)


Itay4 (approved):
we have another job to delete those, right?

avidan-H:
> we have another job to delete those, right?

to delete which?

avidan-H:

Itay4:
> 
> to delete which?

the protected branches

avidan-H:
> > to delete which?
> 
> the protected branches

No - operating under the impression that protected branches are protected for a valid reason and therefore should not be deleted.

PR #15066, opened 2021-09-27
Update CrowdStrikeFalcon.yml to fix capitalisation error

neilw73:

## Status
- Ready

## Related Issues
fixes: I couldn't find Issues section in Github, below is a full description.

## Description
It seems your command arguments text box at the xsoar playbook integration settings has auto-capitalisation and so for Crowdstrike you have this listed as the available options that become available:

Normal,containment_pending,contained,lift_containment_pending

However, the integration, like all others, uses lower-case only and so the python script will fail when you submit 'Normal' as it is only expecting 'normal'.

I expect this error is probably prevalent across many more integrations and properties.

## Screenshots
image

- 6.2.0

   - No

### DIFF
--- a/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
+++ b/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
@@ -83,12 +83,12 @@ script:
       secret: false
     - auto: PREDEFINED
       default: false
-      description: 'The status of the device. Possible values are: "Normal", "containment_pending",
+      description: 'The status of the device. Possible values are: "normal", "containment_pending",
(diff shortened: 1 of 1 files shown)


neilw73:
@tallieber, Sorry I just cannot get your demisto-sdk to install to create the documentation file requested by circleCI. It keep giving me missing files, which to me look like your paths for the creation of the sdk are just too long for Windows.

This change is trivial but breaks the integration and I suspect given the nature of the bug it might affect many more integration parameters too.

tallieber:
Thanks @neilw73. This is actually a bug, so we will take it from here! thank you

neilw73:
@tallieber If it helps, I've just found another one, in 'Palo Alto Networks PAN-OS' your default firewall name is 'Vsys1' and it should be 'vsys1', leaving as is means it fails with an obscure error.

PR #15081, opened 2021-09-29
Fix Gitlab Branch Deletion Scheduled Pipeline

avidan-H:
## Status
- [x] Ready

## Description
The scheduled pipeline to delete Gitlab branches that exist in the `content` project in Gitlab but do not exist in the GitHub repository (from which the repo is mirrored) was not executing as expected. Background on the issue is that the mirroring of the Github repo does not extend to branch deletion so once a branch is mirrored into Gitlab it is never subsequently deleted. The scheduled pipeline did not execute because the stage was set to `.pre`. I've updated it to use the `fan-in` stage - no stage was exactly appropriate for this cleanup pipeline so my choice was somewhat arbitrary. 

## Screenshots
The number of branches in the `content` project in Gitlab prior to manually triggering the pipeline
image

The number of branches in the `content` project in Gitlab subsequent to running the pipeline - note that the number of branches now matches the count which exists in GitHub (at the time of opening this PR)
image

- [x] 5.5.0
- [x] 6.0.0
- [x] 6.1.0
- [x] 6.2.0

   - [x] No

## Must have
- [x] ~~Tests~~ - N/A
- [x] ~~Documentation~~ - N/A

### DIFF
--- a/.gitlab/ci/miscellaneous.yml
+++ b/.gitlab/ci/miscellaneous.yml
@@ -2,7 +2,8 @@ delete-mismatched-branches:
   tags:
     - gke
   stage: .pre
-  extends: .default-job-settings
+  extends:
+    - .default-job-settings
   rules:
     - if: '$DELETE_MISMATCHED_BRANCHES'
   script:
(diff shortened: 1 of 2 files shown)


avidan-H:

yaakovi (approved):

PR #15144, opened 2021-10-03
adding PB110 ignore until fix in sdk

tomneeman151293:


yaakovi:
already handled here: https://github.com/demisto/content/pull/15141

PR #15157, opened 2021-10-03
Checkpoint FWv2: test connection on SID restore

dorschw:
## Status
- [ ] In Progress
- [x] Ready

fixes: https://github.com/demisto/etc/issues/39704

## Description
When restoring SID, test it to make sure it's valid. If it isn't valid, attempt to login.

   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
+++ b/Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
@@ -64,9 +64,9 @@ class Client(BaseClient):
                                           domain_arg: str = None):
 
         if sid_from_context := demisto.getIntegrationContext().get('cp_sid'):
-            demisto.debug(f"restore sid: found some session id ({sid_from_context}) in context ")
+            demisto.debug(f"restore sid: found some session id ({sid_from_context}) in context")
             if self.test_connection() == 'ok':
-                demisto.debug(f"restore sid: success, setting restored session id ({sid_from_context}) on Client ")
+                demisto.debug(f"restore sid: success, setting restored session id ({sid_from_context}) on Client")
                 self.sid = sid_from_context
                 return
             else:
--- a/Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
+++ b/Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
@@ -62,12 +62,18 @@ class Client(BaseClient):
 
     def restore_sid_from_context_or_login(self, username: str, password: str, session_timeout: int,
                                           domain_arg: str = None):
+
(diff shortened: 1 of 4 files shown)


Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2.py
```diff
@@ -62,12 +62,18 @@ def login(self, username: str, password: str, session_timeout: int, domain_arg:
...
-        else:
-            demisto.debug("restore sid: failed to restore, logging in")
-            self.login(username, password, session_timeout, domain_arg)
+            demisto.debug(f"restore sid: found some session id ({sid_from_context}) in context ")
+            if self.test_connection() == 'ok':
+                demisto.debug(f"restore sid: success, setting restored session id ({sid_from_context}) on Client ")
+                self.sid = sid_from_context
+                return
+            else:
+                demisto.debug(f"restore sid: test-connection with session id ({sid_from_context}) failed")
```
DeanArbel, line 73: I think we can mask most of the sid - we only intend to use it for debug purposes, so it's a good practice to not print the full sid.
Please remove the trailing space in line 67.

Packs/CheckpointFirewall/Integrations/CheckPointFirewallV2/CheckPointFirewallV2_test.py
```diff
@@ -103,6 +106,50 @@ def test_checkpoint_login_mechanism(mocker, session_id, integration_context, sho
...
+    mocker.patch.object(CheckPointFirewallV2.demisto, 'params', return_value={'username': {'identifier': 'user',
+                                                                                           'password': 'pass'},
+                                                                              'session_timeout': 600,
+                                                                              'domain_arg': 'domain',
+                                                                              'server': SERVER_URL,
+                                                                              'port': port})
+    mocked_address = f'{HTTPS_WWW_BASE_URL}:{port}/web_api'
+
+    import requests_mock
+    with requests_mock.Mocker() as request_mocker:
```
DeanArbel, line 141: you can use the `requests_mock` as a fixture, so you don't have to use the `with` statement.
Check other usages in the content repo.

Packs/CheckpointFirewall/ReleaseNotes/2_1_5.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### CheckPoint Firewall v2
+- Improved login mechanism: restoration of session id is followed by a connection test, and failure to connect triggers a login.
```
DeanArbel, line 4: This message isn't clear for the average user, as the word `restoration` wouldn't be understood unless you understand the full context of how the integration works.
I'd say that the keyword should be `saved`, so it'll be something along the line of
```
usage of previously saved session id...
```

DeanArbel (commented):
Very nice.
Please make sure to validate the changes before asking for another review.

PR #15188, opened 2021-10-05
Redlock 504 error fix & Adding pagination

ilaner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41718, https://github.com/demisto/etc/issues/24600, https://github.com/demisto/etc/issues/30314

## Description
Customers encountered an 504 when using this integration (`fetch-incidents`). Three bugs were spotted:

1. The url was changed to https://api.prismacloud.io/ (instead of https://api.redlock.io/.)
2. There's a bug that the the integration ignores the last fetch configuration and fetching the results from all the times.
3. pagination was not supported at all.

This PR fixes them, introducing API v2 and using limit argument for pagination
In addition, changed default last fetch time to 24 hours and added link to API reference in description

- [x] 5.5.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PrismaCloud/Integrations/RedLock/RedLock.py
```diff
@@ -525,17 +525,24 @@ def fetch_incidents():
...
+            'type': 'absolute',
+            'value': {
+                'startTime': last_incident_time,
+                'endTime': now
+            }
+        },
+        'filters':
+            [{'name': 'alert.status', 'operator': '=', 'value': 'open'}],
+        'sortBy': ['alertTime:asc'],
+        'limit': demisto.getParam('limit')}
```
DeanArbel, line 545: Please add a default value in case this is empty, and use max function with the default value.

Make sure to add to `test-module` a check that the user didn't provide a value greater than the default value.

Packs/PrismaCloud/Integrations/RedLock/RedLock.py
```diff
@@ -548,15 +555,21 @@ def fetch_incidents():
     demisto.info("Executing Prisma Cloud (RedLock) fetch_incidents with payload: {}".format(payload))
     response = req('POST', 'alert', payload, {'detailed': 'true'})
     incidents = []
+    new_fetches = []
     for alert in response:
+        alert_id = alert.get('id')
+        if alert_id in last_fetches:
+            continue
+        new_fetches.append(alert_id)
```
DeanArbel, line 563: Let's review this offline, I believe the solution might create an issue if you fetch the same payload twice in a row (if for example there are more than `limit` alerts with the same alert time - In that case, you'll delete the prev run `last_fetches` and refetch it).

DeanArbel (changes requested):

Packs/PrismaCloud/ReleaseNotes/1_9_3.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+##### Prisma Cloud (RedLock)
+- changed the url to https://api.prismacloud.io/
+- changed fetch incidents to support pagination
+- fixed bug in fetch incidents with last fetch time
```
bakatzir, line 6: bug is a forbidden word in rn. use only issue.
```suggestion
- Changed the default *Server URL* to: https://api.prismacloud.io/
- Added pagination to the fetch incidents.
- Fixed an issue where the fetch failed in case there was a large number of non fetched alerts.```

bakatzir:
fix the test as well: https://github.com/demisto/etc/issues/24600

PR #15232, opened 2021-10-07
pascalToSpace header transform

BEAdi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41754

## Description
Fixed the issue - now the function replaces tokens in place.
Previously, it replaced all occurrences of the token `t` with `t.title()`. When `t` is a substring of another token it will be treated as a new token for some cases (e.g. `is` was previously capitalized inside `Dis`, meaning `isDisabled` turned into `Is D Is Abled`).
Now, the replacement is done using `re.sub()`, which changes the token as it finds it. The capitalization for the first token is done before the changes (which means we can change the regex from `([A-Z]?[a-z]+)` to `([A-Z][a-z]+)`, but I decided not to change it in case it is used on other packs).

- [x] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests - added
- [ ] Documentation


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6324,10 +6324,9 @@ def pascalToSpace(s):
     if not isinstance(s, STRING_OBJ_TYPES):
         return s
 
-    tokens = pascalRegex.findall(s)
-    for t in tokens:
-        # double space to handle capital words like IP/URL/DNS that not included in the regex
-        s = s.replace(t, ' {} '.format(t.title()))
+    s = s[0].upper() + s[1:]
```
guykeller, line 6327: Why not use capitalize?

BEAdi: Because the string is given as one concatenated word and using `capitalize` will lead to lowering all chars other than the first one.
Anyway, I removed it and changed to use `title` inside `re.sub`, to catch another edge case.

PR #15241, opened 2021-10-08
MicrosoftDefenderAdvancedThreatProtection Integration updates

fahadysf:

## Status
- [ ] In Progress
- [X] Ready

## Related Issues
N/A

## Description
Adds a list of dictionaries to the Machine context item returned by get_machine_details. This data can be useful in identifying MAC Addresses associated with IPs on the machine and using the MAC address in playbook tasks. A use case for this is where incidents generated by MS ATP involve quarantining the machine in a NAC solution like Cisco ISE using the machine's MAC Address. If the MAC is already available from the machine data, external lookups or additional tasks to match the IP to the machine's MAC address wont be needed. (We faced this requirement at a customer).

## Screenshots
image

- [X] 5.5.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0

   - [ ] Yes
   - [X] No

## Must have
- [X] Tests
- [X] Documentation

### DIFF
--- a/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection_test.py
+++ b/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection_test.py
@@ -778,13 +778,19 @@ def test_print_ip_addresses():
 
 
 human_readable_result = '### Microsoft Defender ATP machine None details:\n'\
-                        '|ID|ComputerDNSName|OSPlatform|LastIPAddress|LastExternalIPAddress|HealthStatus|RiskScore|ExposureLevel|IPAddresses|\n'\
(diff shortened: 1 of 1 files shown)


iyeshaya:
Had a talk with @fahadysf 
We agreed that the IpAddresses addition should be more readable with less details, in addition he will give the user an option to retrieve mac address with a different function.

fahadysf:
I've updated the microsoft-atp-get-machine-details command to organize the info about IPAddresses based on interfaces (grouped by unique mac address). This leads to much cleaner and more intuitive context data with one entry per intereface instead of duplicated entries for different ipv4 or ipv6 addresses for the same interface.

image

Additionally the human readable output is also made more concise without losing any detail. Most endpoints would have less than 4 interfaces so the below output with the latest commit is on the higher end of maximum expected information density shown in the markdown table while still remaining very readable.

image

For retreiving the MAC address, a simple "set" task in the playbook which is run after the context is updated (i.e. after the microsoft-atp-get-machine-details has run successfully) can retrieve the MAC address associated with an IP Address on the machine. The below example shows how we could find the MAC Address associated with the "LastIPAddress" filed in the machine details (which is generally the IP actively used for outside communication from the endpoint. With this in mind I don't see a need for adding any functionality for looking up the MAC associated with an IP on a machine.

image

PR #15256, opened 2021-10-10
EditServerConfig - new script

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41777

## Description
Added a new script for updating a server configuration by key and value. 
Three options are available:
1. When action=update and `key` does not exist, the key will be added to the server config with the new value.
2. When action=update and `key` already exist, the key will be updated to the server config with the new value.
3. When action=remove and `key` already exist, the key will be removed from the server config.

## Screenshots

- [x] 6.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/EditServerConfig/EditServerConfig.py
```diff
@@ -0,0 +1,68 @@
...
+    execute_command(
+        "demisto-api-post",
+        {
+            "uri": SERVER_SYSTEM_CONFIG_PATH,
+            "body": {"data": server_config,
+                     "version": -1},
+        })
+
+
+def remove_server_config(key: str, server_config: dict):
```
guykeller, line 26:
```suggestion
def remove_key_from_server_config(key: str, server_config: dict):
```

Packs/CommonScripts/Scripts/EditServerConfig/EditServerConfig.py
```diff
@@ -0,0 +1,68 @@
...
+    action = args.get('action', "")
+    key = args.get('key', "")
+    value = args.get('value', "")
+
+    sys_conf = get_current_server_config()
+
+    if action == "update":
+        update_server_config(key, value, sys_conf)
+    elif action == "remove":
+        remove_server_config(key, sys_conf)
```
guykeller, line 50: What about an `add` action?

Packs/CommonScripts/Scripts/EditServerConfig/EditServerConfig.py
```diff
@@ -0,0 +1,68 @@
...
+    if key in server_config:
+        server_config.pop(key)
+        set_system_config(server_config)
+
+
+def update_server_config(key: str, value: str, server_config: dict):
+    if not value:
+        raise DemistoException(
+            "EditServerConfig Error: You must give a value when you want to update a server configuration.")
+    server_config[key] = value
```
guykeller, line 36: Should the updated values always be of type string?

Packs/CommonScripts/ReleaseNotes/1_4_44.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### New: EditServerConfig
+- Added a new script which updates the server configuration (available from XSOAR 6.0.0).
```
guykeller, line 4:
```suggestion
- Added a new script that allows updating or deleting the values of server configuration keys (available from XSOAR 6.0.0).
```

Packs/CommonScripts/Scripts/EditServerConfig/EditServerConfig.py
abaumgarten: Yes, the server does the converting from string to the correct type in the background. For example (I gave a string and the value set is an int):
The task:
image
Output:
image
The result in the config:
image

Packs/CommonScripts/Scripts/EditServerConfig/EditServerConfig.py
abaumgarten: image

PR #15268, opened 2021-10-11
Added array input support for gcs-download-file command.

ChanochShayner:
## Status
- [x] Ready

## Description
Added array input support for gcs-download-file command.
Configuration setup PB - Send the list correctly to the gcp-download-file command.

- [x] No

### DIFF
--- a/Packs/GoogleCloudStorage/TestPlaybooks/playbook-GCS_Object_Operations_Test.yml
+++ b/Packs/GoogleCloudStorage/TestPlaybooks/playbook-GCS_Object_Operations_Test.yml
@@ -930,3 +930,4 @@ view: |-
   }
 inputs: []
 outputs: []
+fromversion: 5.0.0


Packs/GoogleCloudStorage/Integrations/GoogleCloudStorage/GoogleCloudStorage.py
```diff
@@ -272,15 +272,18 @@ def gcs_list_bucket_objects(client, default_bucket, args):
...
-    saved_file_name = args.get('saved_file_name', '')
-    bucket_name = get_bucket_name(args, default_bucket)
-
-    bucket = client.get_bucket(bucket_name)
-    blob = storage.Blob(blob_name, bucket)
-    saved_file_name = download_blob(client, blob, saved_file_name)
-
-    demisto.results(file_result_existing_file(saved_file_name))
+    list_results = []
+    blob_name = argToList(['object_name'])
```
yuvalbenshalom, line 276: this is strange. `argToList(['object_name'])` should return the list `['object_name']` with no change.
I think you meant to do:
```suggestion
    blob_name = argToList(args.get('object_name'))
```

Packs/GoogleCloudStorage/Integrations/GoogleCloudStorage/GoogleCloudStorage.py
```diff
@@ -272,15 +272,18 @@ def gcs_list_bucket_objects(client, default_bucket, args):
...
+    for obj in blob_name:
+        saved_file_name = args.get('saved_file_name', '')
+        bucket_name = get_bucket_name(args, default_bucket)
+
+        bucket = client.get_bucket(bucket_name)
```
yuvalbenshalom, line 281: please move this before the loop

yuvalbenshalom (changes requested):
see my comments.
I am not sure that this fix really worked, so I am concerned that the test is not really testing it.
also, the argument object_name is array flag should be changed to true in command gcs-download-file

PR #15297, opened 2021-10-12
PAN-OS - Documentation and metadata improvements around push

bakatzir:
## Status
- [x] Ready

relates to: https://github.com/demisto/etc/issues/41980

   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.yml
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.yml
@@ -205,7 +205,7 @@ script:
       - 'false'
     - description: Push description.
       name: description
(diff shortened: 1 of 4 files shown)


Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -205,7 +205,7 @@ script:
       - 'false'
     - description: Push description.
       name: description
-    description: Pushes configuration from PAN-OS to the configured device group.
+    description: Pushes rules from PAN-OS to the configured device group. In order to push the configuration to Prisma Access managed tenants(single or multi tenancy), use the device group argument with the device group which is associated with the tenant ID.
```
amshamah419, line 208:
```suggestion
    description: Pushes rules from PAN-OS to the configured device group. In order to push the configuration to Prisma Access managed tenants (single or multi tenancy), use the device group argument with the device group which is associated with the tenant ID.
```

PR #15349, opened 2021-10-17
Install marketplace packs with "latest" version

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41119

## Description
Fixed an issue where the automation doesn't update the pack when the chosen version was `latest`.

## Screenshots
Before the fix:
image

After the fix:
image

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/ContentManagement/ReleaseNotes/1_0_6.md
+++ b/Packs/ContentManagement/ReleaseNotes/1_0_6.md
@@ -1,4 +1,4 @@
 
 #### Scripts
 ##### MarketplacePackInstaller
-- Fixed an issue where the automation doesn't update the pack when the chosen version was `latest`.
+- Fixed an issue where the automation did not update the desired pack when the chosen version was `latest`.


Packs/ContentManagement/Scripts/MarketplacePackInstaller/MarketplacePackInstaller.py
```diff
@@ -211,6 +211,9 @@ def is_pack_already_installed(self, pack_data: Dict[str, str]) -> bool:
             return False
 
         try:
+            if pack_data['version'] in ['latest', '*']:
+                return parse(self.get_latest_version_for_pack(pack_data['id'])) == self.installed_packs[pack_data['id']]
+
             return parse(pack_data['version']) <= self.installed_packs[pack_data['id']]
```
guykeller, line 217: Why is the upper one with `id` == `id` and the lower one with `version` <= `id`? can you explain what each of them contains?

PR #15354, opened 2021-10-17
XDR mirroring bug fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41457

## Description
In Demisto version 6.1.0 and above, the get_remote_data_command will only be automatically executed on changed incidents
returned from get_modified_remote_data_command so we want to perform extra-data request on those incidents and there is no need to check again if_incident_was_modified_in_xdr.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


dansterenson:
@bakatzir can you please go over the release notes?

Packs/CortexXDR/ReleaseNotes/4_1_11.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Palo Alto Networks Cortex XDR - Investigation and Response
+- Fixed an issue where the integration did not handle mirroring properly.
```
bakatzir, line 4: can you specify in which cases? so the customer will know if it is relevant for him to upgrade the pack.
```suggestion
- Fixed an issue where the integration did not handle mirroring properly.
```

dansterenson: There was a problem with the incidents update time, I am not sure we should get too specific, what do you think?

bakatzir: usually you want to be specific.

PR #15416, opened 2021-10-21
Microsoft Graph API - add no_content arg

Itay4:
## Status
- [x] Ready

## Description
- improved handling of apis which do not return content in the response such as https://docs.microsoft.com/en-us/graph/api/riskyuser-dismiss?view=graph-rest-1.0
- fixed the handling of the `populate_context` arg
- removed the default value of `odata` as it breaks some APIs from running

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/MicrosoftGraphAPI/Integrations/MicrosoftGraphAPI/MicrosoftGraphAPI_test.py
```diff
@@ -0,0 +1,74 @@
...
+        - Ensure outputs are as expected
+    """
+    applications_res = load_test_data('applications.json')
+    requests_mock.get(
+        'https://graph.microsoft.com/v1.0/applications?$top=10',
+        json=applications_res,
+    )
+    args = {
+        'resource': '/applications',
+        'odata': '$top=10',
```
yaakovi, line 46: since you removed the default value, what will get returned if you drop this? how many entries?

Packs/MicrosoftGraphAPI/Integrations/MicrosoftGraphAPI/MicrosoftGraphAPI_test.py
```diff
@@ -0,0 +1,74 @@
...
+        - Ensure the command does not fail on no content response
+    """
+    requests_mock.post(
+        'https://graph.microsoft.com/v1.0/identityProtection/riskyUsers/dismiss',
+        text='',
+    )
+    args = {
+        'http_method': 'POST',
+        'resource': '/identityProtection/riskyUsers/dismiss',
+        'odata': '',
```
yaakovi, line 70: remove to mimic the command run in demisto

yaakovi (approved):
# pragma: no errors

Packs/MicrosoftGraphAPI/Integrations/MicrosoftGraphAPI/MicrosoftGraphAPI_test.py
Itay4: as many as there are
it was the concern initially and thats why i set it as default, but it breaks some apis and causes bad ux
got another idea ?

yaakovi: maybe state it in the command/argument description?

Itay4: done, can you have a look please?

yaakovi (approved):

PR #15541, opened 2021-10-28
Misp return attribute feed hit

abaumgarten:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42650

## Description
As the feed data includes exactly 6 fields that are final (won't cause any performance issues) we should return the attributes' feed hit data as part of the search-events command. 
We want to return these attributes fields only if the user asked for 'include_feed_correlations' arg. Otherwise, the response to the Context data won't include the `Attribute` field at all.

## Screenshots
You can see here the Event's attribute list. Each attribute object includes:
- Attribute ID
- Attribute value
- Feed (is a list of attributes feeds. Each feed list includes 6 final fields)
 
image

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/MISP/Integrations/MISPV3/README.md
```diff
@@ -48,7 +48,7 @@ Search for events in MISP. This search command will return only information abou
 | uuid | The event UUID to be returned by the search. For example, 59523300-4be8-4fa6-8867-0037ac110002. | Optional | 
 | page | If a limit is set, sets the page to be returned. For example, page 3, limit 100 will return records 201-&gt;300. Default is 1. | Optional | 
 | limit | Limit the number of events returned. Default is 50. | Optional | 
-| include_feed_correlations | Whether to return the event related feeds. Possible values are: true, false. | Optional | 
+| include_feed_correlations | Whether to return the event related feeds. Possible values are: true, false. Note: Only if this argument set to "true" the response will include attributes feed hits values.| Optional |
```
guykeller, line 51:
```suggestion
| include_feed_correlations | Whether to return the event related feeds. Possible values are: true, false. Note: Only if this argument set to "true" the response will include attributes' feed hits values.| Optional | 
```

Packs/MISP/Integrations/MISPV3/MISPV3_test.py
```diff
@@ -548,7 +548,7 @@ def test_build_events_search_response(mocker):
     from MISPV3 import build_events_search_response, EVENT_FIELDS
     search_response = util_load_json("test_data/search_event_by_tag.json")
     search_expected_output = util_load_json("test_data/search_event_by_tag_outputs.json")
-    search_outputs = build_events_search_response(search_response)
+    search_outputs = build_events_search_response(search_response, {'include_feed_correlations': True})
```
guykeller, line 551: Let's add a UT for the False case

guykeller (changes requested):

Packs/MISP/Integrations/MISPV3/MISPV3.py
```diff
@@ -980,13 +983,30 @@ def build_events_search_response(response: Union[dict, requests.Response]) -> di
...
         events[i]['publish_timestamp'] = misp_convert_timestamp_to_date_string(events[i].get('publish_timestamp'))
 
     formatted_events = replace_keys_from_misp_to_context_data(events)  # type: ignore
     return formatted_events  # type: ignore
 
 
+def build_attribute_feed_hit(event: dict, demisto_args=None):
+    """ We want to have the Attribute data as part of the search-events results only if the user asked for
+    include_feed_correlations. Otherwise, we don't want to return attributes data at all."""
+    if demisto_args and argToBoolean(demisto_args.get('include_feed_correlations', False)):
```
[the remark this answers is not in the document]
guykeller: What am I missing with this suggestion?
```suggestion
    if argToBoolean(demisto_args.get('include_feed_correlations', False)):
```

guykeller (approved):
one comment, leaving to your judgement

PR #15573, opened 2021-10-31
EWS O365 Add Delegate Access Type

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42683

## Description
Added a new parameter **Access Type** - toggle between `Impersonation` and `Delegate`. Previously only `Impersonation` was supported.

   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/EWS/Integrations/EWSO365/EWSO365.py
```diff
@@ -192,7 +192,7 @@ def __init__(
         )
         self.folder_name = folder
         self.is_public_folder = is_public_folder
-        self.access_type = kwargs.get('access_type') or IMPERSONATION
+        self.access_type = kwargs.get('access_type', '').lower() or IMPERSONATION
```
Itay4, line 195: for instances that were configured prior updating this, will there be `None` as the value of that param ?

DeanArbel: Yeah

Packs/EWS/Integrations/EWSO365/EWSO365.py
```diff
@@ -192,7 +192,7 @@ def __init__(
         )
         self.folder_name = folder
         self.is_public_folder = is_public_folder
-        self.access_type = kwargs.get('access_type') or IMPERSONATION
+        self.access_type = kwargs.get('access_type', IMPERSONATION).lower()
```
Itay4, line 195: lets change to the following to be on the safe side:
```suggestion
        self.access_type = (kwargs.get('access_type', IMPERSONATION) or IMPERSONATION).lower()
```

Itay4 (approved):

PR #15581, opened 2021-11-01
Update Docker Image To demisto/genericsql

wissamg:
####Docker Image Updates
##GenericSQL Pack:
Generic SQL


wissamg:
The google sdk for this integration was updated by the dependabot and this break the integration, for now we(@Shellyber and @wissamg ) decided to revert the docker image changes and add the dependabot ignore flag for this integration as workaround for now till the integration rewrite happening

Shellyber (approved):

PR #15592, opened 2021-11-02
[Indicators Searcher] Fix IndicatorsSearcher Stop Condition

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42855
fixes: https://github.com/demisto/etc/issues/42925
fixes: https://github.com/demisto/etc/issues/42782

## Description
Fixed an issue where different integrations could get stuck in a search loop when searching for more indicators than are available in the system.

   - [x] No


glicht:
@DeanArbel please schedule a code review session

glicht (changes requested):
Let's do this one TDD. Add a test that fails and then do the fix.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -8022,33 +8013,31 @@ def total(self):
...
     @limit.setter
     def limit(self, value):
-        self._next_limit = self._original_limit = value
+        self._limit = value
 
-    def _is_search_done(self):
+    def is_search_done(self):
         """
         Checks one of these conditions:
         1. self.limit is set, and it's updated to be less or equal to zero
```
tkatzir, line 8025:
```suggestion
        1. self.limit is set, and it's updated to be less or equal to zero - return True
```

Mention the return value in each case.

tkatzir (changes requested):

tkatzir (approved):

DeanArbel:
Validations are failing because of:
* Workday_IAM - missing test playbook.
* GetIndicatorsByQuery - missing test playbook.

Gitlab build is failing because of:
* `TAXII Feed` is unstable (flaky results that are caused by HTTP connection issues)
* In some builds the pack installation is failing, causing the build to fail.

Packs/Base/ReleaseNotes/1_15_1.md
```diff
@@ -1,4 +1,4 @@
 
 #### Scripts
 ##### CommonServerPython
-- Maintenance and stability enhancements.
\ No newline at end of file
+- Maintenance and stability enhancements.
```
glicht, line 4: Why this change?

Packs/EDL/Integrations/EDL/EDL.py
```diff
@@ -131,17 +131,13 @@ def create_new_edl(request_args: RequestArguments) -> str:
         size=PAGE_SIZE,
         limit=limit
     )
-    iocs = []
+    iocs: List[dict] = []
     formatted_iocs: set = set()
-    while True:
-        current_limit = limit + (limit - len(formatted_iocs))
-        indicator_searcher.limit = current_limit
+    while not indicator_searcher.is_search_done():
```
DeanArbel, line 136: Put a hard limit here - no more than 10 loops

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -7985,37 +7985,29 @@ def __init__(self,
...
-        size = min(self._size, self._next_limit or self._size)
         res = self.search_indicators_by_version(from_date=self._from_date,
                                                 query=self._query,
-                                                size=size,
+                                                size=self._size,
                                                 to_date=self._to_date,
                                                 value=self._value)
         fetched_len = len(res.get('iocs') or [])
         if fetched_len == 0:
+            self.limit = 0
```
DeanArbel, line 8007: replace limit with count - i.e. instead of lowering the limit, we'll increase the count.

Packs/Base/ReleaseNotes/1_15_1.md
DeanArbel: It's an inner change of the way the `IndicatorsSearcher` class works, but I'll change it to be something more infomercial.

DeanArbel: probably got picked up when merging from master. Reverting

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -8028,33 +8016,34 @@ def total(self):
...
-        if self._search_is_done:
-            return True
-
-        reached_limit = isinstance(self._next_limit, int) and self._next_limit <= 0
+        reached_limit = self.limit is not None and self.limit <= self._total_iocs_fetched
         if reached_limit:
+            demisto.debug("IndicatorsSearcher has reached its limit: {}".format(self.limit))
+            # update limit to match _total_iocs_fetched value
+            if self._total_iocs_fetched > self.limit:
+                self.limit = self._total_iocs_fetched
```
DeanArbel, line 8037: @glicht Please notice this line.
We update the `self.limit` with `self._total_iocs_fetched` to handle cases where it's greater than the limit - this will allow updates to limit to affect further searches that might be done while updating limit (otherwise `limit` can be lower than `_total_iocs_fetched` even after increasing its value by the calling integration).

glicht (approved):

PR #15605, opened 2021-11-03
Mail Listener Fetch fails on unicode issue

JasBeilin:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42462

## Description
For some reason, incident data is fetched as string from the API instead of bytes-string, therefore the character '\u' was recognized as special character when parsing using the mailparser module. 
We added casting to make sure it is always bytes.


Packs/MailListener/Integrations/MailListenerV2/MailListenerV2_test.py
```diff
@@ -138,3 +138,23 @@ def test_generate_labels():
...
+                                                (br'C:\User\u', br'C:\User\u')])
+def test_convert_to_bytes(mocker, src_data, expected):
+    """
+    Given:
+        A byte string representing response from API
+    When:
+        1. The string returns as string
+        2. The string returns as bytes
+    Then:
+        Fails if the casting to bytes fails.
```
yuvalbenshalom, line 153: but it does not fail, right?
it is not clear what's bing tested here.

Packs/MailListener/Integrations/MailListenerV2/MailListenerV2.py
```diff
@@ -381,6 +387,13 @@ def get_email_as_eml(client: IMAPClient, message_id: int) -> dict:
     return mail_file[0] if mail_file else {}
 
 
+def _convert_to_bytes(data) -> bytes:
+    demisto.debug("Converting data to bytes.")
+    bytes_data = bytes(data)
+    demisto.debug("Converted data successfully.")
```
yuvalbenshalom, line 393: I don't think we need this line. if the convention will fail, we will surly know anyway.

Packs/MailListener/ReleaseNotes/1_0_6.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Mail Listener v2
+- Fixed an issue in **fetch incidents** command where some data fetched from the API could not be parsed.
```
yuvalbenshalom, line 4:
```suggestion
- Fixed an issue where **fetch incidents** command failed to parse specific mail messages.
```

Packs/MailListener/Integrations/MailListenerV2/MailListenerV2_test.py
```diff
@@ -138,3 +138,23 @@ def test_generate_labels():
...
+
+@pytest.mark.parametrize('src_data, expected', [(r"C:\User\u".encode('utf-8'), br'C:\User\u'),
+                                                (br'C:\User\u', br'C:\User\u')])
+def test_convert_to_bytes(mocker, src_data, expected):
+    """
+    Given:
+        A byte string representing response from API
+    When:
+        1. The string returns as string
+        2. The string returns as bytes
```
yuvalbenshalom, line 151:
```suggestion
        the response is converted to bytes
```
maybe?

yuvalbenshalom (changes requested):
see my comments

yuvalbenshalom (approved):

PR #15615, opened 2021-11-03
Splunk Pre Release - handle offset bug

tallieber:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [https://github.com/demisto/etc/issues/42218](https://github.com/demisto/etc/issues/42218)

## Description
When we are getting number of events that is greater than the limit we will use the offset parameter to indicates to the next run which results to return from.  Event that was indexed between the first run and the second run and his occurrence time is included on the query’s time range - will change the order of the events. We could not use the offset parameter that relies on the idea that the order of results is not changing.
1. Remove the offset parameter. Now we will go over all the returned event and search for new events we have not seen. 
2. We will have a dict holds all event we saw and returned as incident and their occurrence time. 
3. We will remove events from that dict if their occurrence is no longer included on the query time range. 

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease.py
```diff
@@ -307,30 +310,25 @@ def fetch_notables(service, cache_object=None, enrich_notables=False):
...
-            'offset': 0,
-            'found_incidents_ids': last_run_fetched_ids
+            'found_incidents_ids': last_run_fetched_ids,
+            'batch_size': default_batch_size
         }
     else:
+        if len(reader) == batch_size:
+            batch_size += default_batch_size
+        latest_incident_fetched_time = get_latest_incident_time(incidents)
+        next_run = get_next_start_time(latest_incident_fetched_time, now, were_new_incidents_found=True)
```
guykeller, line 325: There may be no new incidents.
We need to change the logic here to make sure we handle the cases differently

guykeller (changes requested):

Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease.py
tallieber: handled in the last commit

Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease.py
```diff
@@ -307,31 +312,29 @@ def fetch_notables(service, cache_object=None, enrich_notables=False):
...
-        new_last_run = {
-            'time': next_run,
-            'offset': 0,
-            'found_incidents_ids': last_run_fetched_ids
-        }
+        batch_size = default_batch_size
+        extensive_log('[SplunkPyPreRelease] SplunkPyPreRelease - Next run time with no incidents found: {}. '
+                      'Batch size: {}'.format(next_run, batch_size))
+
+    elif len(incidents) == 0 and reader_length == batch_size:  # All events returned from the query was duplicates
```
guykeller, line 321:
```suggestion
    elif len(incidents) == 0:  # All events returned from the query was duplicates
```

guykeller (commented):
As far as I'm concerned, let's send to customers

Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease.py
```diff
@@ -249,46 +248,52 @@ def fetch_notables(service, cache_object=None, enrich_notables=False):
...
+    extensive_log('[SplunkPyPreRelease] SplunkPyPreRelease batch size is : {}'.format(batch_size))
+
+    kwargs_oneshot = build_fetch_kwargs(dem_params, batch_size, occured_start_time, now)
     fetch_query = build_fetch_query(dem_params)
 
     oneshotsearch_results = service.jobs.oneshot(fetch_query, **kwargs_oneshot)  # type: ignore
-    reader = results.ResultsReader(oneshotsearch_results)
-
+    reader = list(results.ResultsReader(oneshotsearch_results))
+    reader_length = len(reader)
```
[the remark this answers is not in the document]
tallieber: Unfortunately did not work. this is one of the ways to count

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy_description.md
```diff
@@ -1,4 +1,7 @@
 ## SplunkPy
+We strongly recommend you to use **Splunk Pre Release** integration which contains improvements regarding the fetch mechanism.
```
guykeller, line 2:
```suggestion
We strongly recommend using **Splunk Pre Release** integration which contains improvements regarding the fetch mechanism.   
Specifically, if you experience any issues regarding fetch logic, including (but not limited to) missing incidents, using the pre release version is the first step you should take.
```

Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease_description.md
```diff
@@ -8,9 +8,39 @@ To use Splunk token authentication, enter the text: *_token* in the **Username**
 To create an authentication token, go to [Splunk create authentication tokens](https://docs.splunk.com/Documentation/SplunkCloud/8.1.2101/Security/CreateAuthTokens).
 ***
 
-### Fetching notable events.
+# Splunk Enterprise Security Users
+
+## Fetching notable events.
 The integration allows for fetching Splunk notable events using a default query. The query can be changed and modified to support different Splunk use cases. (See [Existing users](#existing-users)).
 
+We are highly recommend you to read [Ingest Incidents from a SIEM Using Splunk article](https://docs.paloaltonetworks.com/cortex/cortex-xsoar/6-2/cortex-xsoar-tutorials/tutorials/splunk-use-case.html) before starting using this integration.
```
guykeller, line 16:
```suggestion
We highly recommend reading the [Ingest Incidents from a SIEM Using Splunk article](https://docs.paloaltonetworks.com/cortex/cortex-xsoar/6-2/cortex-xsoar-tutorials/tutorials/splunk-use-case.html) before starting using this integration.
```

Packs/SplunkPyPreRelease/Integrations/SplunkPyPreRelease/SplunkPyPreRelease_description.md
```diff
@@ -8,9 +8,39 @@ To use Splunk token authentication, enter the text: *_token* in the **Username**
...
+Outgoing miroriring is recommended for Cortex XSOAR version 6.2 and above. If you enable mirroring, you need to add the timezone of the splunk server (in minutes). For example, if using GMT and the timezone is GMT +3 hours, set the timezone to +180. For UTC, set the timezone to 0. Set this only if the Splunk server is different than the Cortex XSOAR server. This is relevant only for fetching notable events.
+13. Select Close Mirrored XSOAR Incident and Close Mirrored Splunk Notable Event, so when closing in one environment, it closes in the other.
+14. In the Enrichment Types field, select Asset, Drilldown and Identity.
+This enrichment provides additional information about Assets, drilldown, and identities that are related to the notable events you ingest.
+15. Fetch backwards window - this backward window is for cases there is a gap between the event occurrence time and the event index time on the server.
+The backward window should be the average time in minutes between events occurence time and index time. This value will be included on the search query time range in order to not miss events that was index on delay.
```
guykeller, line 38: Would love to hear your thoughts and feedback
```suggestion
15. Fetch backwards window - this backward window is for cases there is a gap between the event occurrence time and the event index time on the server.
In Splunk, there is often a delay between the time an incident is created (the event's "occurrence time") and the time it is actually searchable in Splunk and visible in the index (the event's "index time").
This delay can be caused by an inefficient Splunk architecture, causing higher event indexing latency. However, it can also be "by design", e.g. if some endpoints / machines that generate Splunk events are usually offline.
Another point to note is that Splunk's searches are based on the occurrence time behind the scenes. Meaning, Splunk itself uses occurrence time as its determining factor for bucket division and search.
Therefore - we can't use index time as our primary search key without making the searches inefficient.
The backwards window is a way for you to configure the longest delay you would like to support.
This parameter determines the size of the occurrence time "sliding window" we will support in our queries. For example, if set for 2 hours, we will always search for events that occurred up to 2 hours ago (and will of course ignore duplicates).
However, there is obviously a price - the larger the window, the longer it will take for fetch queries to complete.
The best value to set depends on the delays that you see in your system (consult with your Splunk expert / master), the number of events in your system and other network properties.
Please use this parameter with careful consideration.
```

guykeller (changes requested):

PR #15666, opened 2021-11-08
Giphy detailed description

dorschw:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/43282

## Description
updated the developer portal url, where users are directed to get an api key.

No

## Must have
- [ ] Documentation


DeanArbel (approved):
Make sure to add release notes.

PR #15683, opened 2021-11-09
Add dbotrole to AssignToNextShiftOOO

JasBeilin:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/43258

## Description
Added DbotRole as this script uses setting attribute to other incidents, after considering the fact that it will run as part of playbook on incidents came from SearchIncidentsV2 running on limited user.


JasBeilin:
Decided to handle within the code instead.

PR #15711, opened 2021-11-10
Fixing the splunk rawToDict parsing

agamm:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/43188

## Description
A rehaul for the `rawToDict` function, using somewhat more complex parsing for the query taking into account nested quotes.

The basic idea is that when there are nested quotes we can check for even number of quotes and and ending symbol (`,`). 
First we clean the raw string from unused non-key-value pairs and then we wrap each key-value pair without quotes with quotes.
We then iterate over the raw string using 2 indexes (one left and one right) we advance them each time we add a key value pair and append the result into the groups (final result of the function).

## Screenshots
before:
image

after:
image

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


tallieber:
@agamm I think we needed to fix code also on SplunkPyPreRelease

guykeller (approved):
Please add this new logic to the pre release version as well.
Other than that, looks good

tallieber (approved):

PR #15811, opened 2021-11-16
[EWS O365] Change Access Type param to be Single Select

DeanArbel:
## Status
- [x] Ready

## Description
Changed **Access Type** parameter to be *Single Select* instead of *Multi Select*.

## Screenshot
image

   - [x] No


Packs/EWS/ReleaseNotes/1_10_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### EWS O365
+- Changed **Access Type** parameter to be *Single Select* instead of *Multi Select*.
```
Itay4, line 4:
```suggestion
- Changed the **Access Type** parameter type to be *Single Select* instead of *Multi Select*.
```

Itay4 (approved):

PR #15643, opened 2021-11-07
Q radar additional tests

tomneeman151293:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/43045
fixes: https://github.com/demisto/etc/issues/43850
fixes: https://github.com/demisto/etc/issues/43832

## Description
Due to past regressions regarding integration context, adding tests to make sure it will be protected from later changes.

PR #15826, opened 2021-11-17
Refactor QRadar_v3 Long Running

tkatzir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
Related: https://github.com/demisto/etc/issues/43713

## Description
Refactor the Long Running command, to make it more stable.

   - [ ] Yes
       - Further details:
   - [x] No


tomneeman151293:
@tkatzir I am soon adding some changes to the QRadar python file, make sure to merge after this pr is merged so you will avoid difficult conflicts later on 
https://github.com/demisto/content/pull/15643

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
```diff
@@ -1348,10 +1347,18 @@ def create_search_with_retry(client: Client, fetch_mode: str, offense: Dict, eve
         f'SELECT {event_columns} FROM events WHERE INOFFENSE({offense_id}) {additional_where} limit {events_limit} '
         f'START {offense_start_time}'
     )
+    print_debug_msg(f'Trying to get events for offense ID: {offense_id}, '
+                    f'query_expression: {query_expression}.')
```
DeanArbel, line 1351: Note that the `query_expression` contains a lot of data under `{events_columns}` while staying very static. I think a better approach is to print a truncated version with just the following data:
1. `{offense_start_time}` to show the only dynamic field.
2. `{additional_where}` to show the user query that's being applied.

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
```diff
@@ -1348,10 +1347,18 @@ def create_search_with_retry(client: Client, fetch_mode: str, offense: Dict, eve
...
     num_of_failures = 0
     while num_of_failures <= max_retries:
         try:
-            return client.search_create(query_expression=query_expression)
+            print_debug_msg(f'Creating search for offense ID: {offense_id}, '
+                            f'query_expression: {query_expression}.')
+            ret_value = client.search_create(query_expression=query_expression)
+            print_debug_msg(f'Created search for offense ID: {offense_id}, '
+                            f'query_expression: {query_expression}, '
```
DeanArbel, line 1359: same comment as above.

DeanArbel (approved):
Nice work.
Please consider truncating the events logs.

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
```diff
@@ -1388,11 +1396,20 @@ def poll_offense_events_with_retry(client: Client, search_id: str, offense_id: i
...
     while num_of_failures <= max_retries:
         try:
+            print_debug_msg(f"Getting search status for {search_id}")
             search_status_response = client.search_status_get(search_id)
+            print_debug_msg(f"Got search status for {search_id}")
             query_status = search_status_response.get('status')
             # failures are relevant only when consecutive
             num_of_failures = 0
-            if query_status in TERMINATING_SEARCH_STATUSES:
+            print_debug_msg(f'Search query_status: {query_status}')
```
tomneeman151293, line 1405: I am a little afraid those many debug messages in the futures flow will really hurt performance
Imagine a customer defining to fetch more than THREADS COUNT 
because support_multithreaded is enabled, for every debug message only one will write each time
WDYT? @tkatzir @DeanArbel

tkatzir: Why should support_multithreaded have an effect on log messages?
Isn't the logger multi threaded compatible?

The issue is that QRadar isn't "reliable" (for lack of a better term...) and we want to know about each call to it, especially during long_running.

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
tkatzir: Accepted. Changed to: offense_id, offense_start_time, additional_where, events_limit.

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
tkatzir: Accepted. Changed to: offense_id, offense_start_time, additional_where, events_limit, ret_value.

Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py
tomneeman151293: @Itay4 How does the logging mechanism happen? does it uses the Demisto Do or has a separate flow that the demisto.info/debug/error etc won't be affected by the lock?

tkatzir: From what I can tell, it's:
QRadar_v3.print_debug_msg
demisto.debug

Since we `support_multithreading`, then we should be safe.

Itay4: all the log functions call `__do`, but not sure we should concerned with performance here - lets discuss

PR #15847, opened 2021-11-18
feat: Add VirusTotal Livehunt feed

danipv:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Add VirusTotal Livehunt as a Cortex feed to the pack.

## Screenshots

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/VirusTotal/Integrations/FeedLivehunt/FeedLivehunt.py
```diff
@@ -0,0 +1,266 @@
...
+            filter_: Optional[str] = None
+    ) -> dict:
+        """Retrieve VT Hunting Livehunt notifications files.
+        """
+        return self._http_request(
+            'GET',
+            'intelligence/hunting_notification_files',
+            params=assign_params(
+                filter=filter_,
+                limit=max(limit, 40),
```
dansterenson, line 61: Seems like the feed is always pulling the same 20 indicators because of the limit argument in you http call. We want to get the updated indicators each fetch, it can be done by getting the new updated one from the API (if the API supports it), or we can get all indicators from the API and then Cortex XSOAR will filter the new indicators. You can take a look at FeedOffice365 for example. In each fetch they are getting all indicators, the limit is (-1) so it does not effect them.

dansterenson (changes requested):

Packs/VirusTotal/Integrations/FeedLivehunt/FeedLivehunt.py
danipv: Yep, the API returns always the last indicators found in VT hunting, regardless whether they have been consumed before or not, unfortunately VT API doesn't support this. But, is this a problem? As the fetch ignore indicators that already exist in Cortex should't be a mayor problem, isn't it?

Packs/VirusTotal/Integrations/FeedLivehunt/FeedLivehunt.py
```diff
@@ -0,0 +1,315 @@
...
+                result.append({
+                    'data': indicator,
+                    'type': 'file',
+                    'FeedURL': self._base_url
+                })
+        except ValueError as err:
+            demisto.debug(str(err))
+            raise ValueError(f'Could not parse returned data as indicator. \n\nError message: {err}')
+
+        self.set_last_run()
```
dansterenson, line 71: If the fetch_command arg is 'False' (test_module) for example, do we still want to set the last run?

dansterenson (changes requested):
Nice job! please see my comment.

Packs/VirusTotal/Integrations/FeedLivehunt/FeedLivehunt.py
danipv: You are right, set last run isn't necessary when `fetch_command` is false. Good catch! Fixed.

dansterenson (approved):

PR #15858, opened 2021-11-18
blacklist and whitelist changes to md and yml files

tschanfeld:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
[fixes: link to the issue](https://github.com/demisto/etc/issues/42124)

## Description
Goal is to replace blacklist/black list with block list and whitelist/white list with allow list, WITHOUT breaking anything.
Two concerns - a) that I could have changed text that will break something b) that we can end up with a confusing mix of blacklist and block list, whitelist and allow list, all in the same files, since some items did not look safe to change. I edited only md and yml files.


Packs/AbuseDB/Integrations/AbuseDB/AbuseDB.yml
```diff
@@ -244,7 +244,7 @@ script:
     name: abuseipdb-get-blacklist
     outputs:
     - contextPath: AbuseIPDB.Blacklist
-      description: A list of blacklisted IP addresses.
+      description: A list of IP addresses on a block list.
```
yaakovi, line 247: are we ok with mixed terminology in this case?
the path is called `blacklist` which we cannot change while the description is using `block list`.

tschanfeld:
@ShahafBenYakir what is the status with this?

ShahafBenYakir:
@tschanfeld In progress

ShahafBenYakir:
Changes were only done in READMEs so Validation fails on unrelated issues

PR #15887, opened 2021-11-21
Disable MITRE ATTACK Indicator Type

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Disabled MITRE ATTACK Indicator type as it uses the same regex pattern as the new Attack Pattern type,
And we recommended the users use the v2 Pack.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/FeedMitreAttack/IndicatorTypes/reputation-mitreAttck.json
+++ b/Packs/FeedMitreAttack/IndicatorTypes/reputation-mitreAttck.json
@@ -13,7 +13,7 @@
     "enhancementScriptNames": [],
     "system": false,
     "locked": false,
-    "disabled": false,
+    "disabled": true,
     "file": false,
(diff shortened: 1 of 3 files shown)


ytauman:
is this needed, considering this isn't a pack that's installed by default (and if a customer really does insist on installing it - it will appear as 'not working'?

altmannyarden:
I had this type (MITRE ATTACK) and the Attack pattern indicator type didn't worked for me and I couldn't understand why. So I think it's very important to fix @ytauman

ytauman:
@altmannyarden that's because you have both.
we already deprecated the MITRE v1 pack - so if the customer insists on downloading it (and not the v2) - i want it to work for them.

PR #15890, opened 2021-11-21
SlackV3 - Fix for channel id timeout issue for large enterprise workspaces

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42709
fixes: https://github.com/demisto/etc/issues/42837
fixes: https://github.com/demisto/etc/issues/44344

## Description
This version fixes an issue where finding channel IDs on large enterprise workspaces would cause a timeout due to rate limiting.

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -836,6 +837,8 @@ async def slack_loop():
             await asyncio.sleep(float("inf"))
         except Exception as e:
             await handle_listen_error(f"An error occurred {str(e)}")
+        finally:
+            await client.disconnect()
```
amshamah419, line 841: This is purely to gracefully close the socket connection on failure.

Packs/Slack/Integrations/SlackV3/SlackV3_test.py
```diff
@@ -3920,3 +3922,227 @@ def test_pin_message_invalid_thread_id(mocker):
...
+    SlackV3.init_globals()
+    SlackV3.CHANNEL_FETCH_INTERVAL = "3 seconds"
+    time.sleep(1)
+    returned_time = await SlackV3.fetch_channels_iterable(last_update=last_update)
+
+    # Assert
+    assert returned_time.replace(microsecond=0) >= (last_update + timedelta(seconds=8)).replace(microsecond=0)
+
+
+def test_handle_date_interval_minutes():
```
tkatzir, line 4096: Maybe all these test_handle_date_interval_* can be one parameterized test?

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1249,109 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
     return ''
 
 
+def handle_date_interval(fetch_interval: str, last_update_time):
```
tkatzir, line 1251: This looks like a generic utility method. Maybe it can be in CommonServerPython? Be made static?
Also - missing documentation.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1249,109 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
...
+    range_split = fetch_interval.strip().split(' ')
+    number = 0
+    next_update_time = None
+    if len(range_split) != 2:
+        return_error('date_range must be "number date_range_unit", examples: (2 hours, 4 minutes,6 months, 1 day, '
```
tkatzir, line 1256:
```suggestion
        return_error('date_range must be "number date_range_unit", examples: (2 hours, 4 minutes, 6 months, 1 day, '
```

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1249,109 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
...
+        'second', 'seconds',
+        'minute', 'minutes',
+        'hour', 'hours',
+        'day', 'days',
+        'month', 'months',
+        'year', 'years',
+        ]:
+        return_error('The unit of date_range is invalid. Must be minutes, hours, days, months or years.')
+
+    if 'second' in unit:
```
tkatzir, line 1268: Wouldn't `unit.startswith('minute')` be more efficient?

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1249,109 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
     return ''
 
 
+def handle_date_interval(fetch_interval: str, last_update_time):
+    range_split = fetch_interval.strip().split(' ')
```
tkatzir, line 1252: What if there are several spaces? e.g. '2 &nbsp; &nbsp; &nbsp; &nbsp; hours'

tkatzir (changes requested):

Packs/Slack/Integrations/SlackV3/SlackV3.py
amshamah419: I'm open to moving it to common server, but is there a need? We have one that is similar, but looks back not forward. @yuvalbenshalom wdyt?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -1328,6 +1328,51 @@ def stringUnEscape(st):
...
+        :type fetch_interval: ``str``
+        :param fetch_interval: A string containing a date range e.g. (2 Days)
+
+        :type last_update_time: ``datetime``
+        :param last_update_time: The datetime object that the date range will be incremented on.
+
+        :return: A datetime object for the next interval.
+        :rtype: ``datetime``
+    """
+    range_split = " ".join(fetch_interval.split()).strip().split(' ')
```
tkatzir, line 1345: Would this work? One strip and one split (instead of join, strip, and 2 splits)
```
fetch_interval_split = fetch_interval.strip().split()
number = fetch_interval_split[0]
unit = fetch_interval_split[-1].lower()
```

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2050,6 +2120,64 @@ def pin_message():
         return_error(f"{slack_error}")
 
 
+def fetch_channels():
```
yuvalbenshalom, line 2123: I am not sure that it makes sense that we are not using the same core function to fetch channels names. seems to me like fetch_channels_iterable and fetch_channels should call the same function, and the update of the context should be outside.

yuvalbenshalom (commented):
lets have a session about this implementation.

Packs/Slack/Integrations/SlackV3/SlackV3.py
yuvalbenshalom: this should probably be merged with get_conversation_by_name as an internal function that will be called to search all or to get one channel name.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -819,23 +820,32 @@ def exception(message):
...
+                try:
+                    await client.disconnect()
+                except Exception as e:
+                    demisto.debug(f"Failed to close client. - {e}")
+            exception_await_seconds = 1
         except Exception as e:
-            await handle_listen_error(f"An error occurred {str(e)}")
+            demisto.debug(f"Exception in long running loop, waiting {exception_await_seconds} - {e}")
+            await asyncio.sleep(exception_await_seconds)
+            exception_await_seconds *= exception_await_seconds
```
amshamah419, line 848: This is the fix for the ghost. I'd rather get this out sooner than later due to it's impact to customers.

amshamah419: This doesn't play nicely with the other changes. I'll need to re-work what we do here. Disregard for now.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1246,79 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
...
+                                                            exclude_archived=True, cursor=cursor)
+            except SlackApiError as api_error:
+                demisto.debug(f'Got rate limit error (sync). Body is: {str(res)}\n{api_error}')
+                response = api_error.response
+                headers = response.headers  # type: ignore
+                if 'Retry-After' in headers:
+                    retry_after = int(headers['Retry-After'])
+                    total_try_time += retry_after
+                    if total_try_time < MAX_LIMIT_TIME:
+                        time.sleep(retry_after)
```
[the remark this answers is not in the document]
yuvalbenshalom: shouldn't this be an async sleep?

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1243,6 +1246,79 @@ async def check_and_handle_entitlement(text: str, user: dict, thread_id: str) ->
...
+                        continue
+                else:
+                    if total_try_time < MAX_LIMIT_TIME:
+                        time.sleep(5)
```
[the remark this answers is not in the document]
yuvalbenshalom: shouldn't this be an async sleep?

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2050,6 +2085,119 @@ def pin_message():
...
+                            time.sleep(retry_after)
+                            continue
+                    else:
+                        if total_try_time < MAX_LIMIT_TIME:
+                            time.sleep(5)
+                            continue
+                        else:
+                            raise
+                break
+    return updated_channel_list
```
yuvalbenshalom, line 2192: I think that if this function gets a conversation name that does not exist it will crush as `updated_channel_list` does not exist. please add a unit test for this first. (has conversation name and reaching the end of the cursor.

yuvalbenshalom (changes requested):
see my comments.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -598,6 +599,7 @@ async def long_running_loop():
         try:
             check_for_mirrors()
             check_for_unanswered_questions()
+            await asyncio.sleep(5)
```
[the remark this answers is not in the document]
amshamah419: It's an async sleep to free up the cpu to perform other tasks. Since it loops continually, if there is no sleep, then it will occupy the process and block the other coroutines from executing.

Packs/Slack/Integrations/SlackV3/SlackV3.py
amshamah419: So it didn't crash, but it did return an empty list. I fixed the code to return an error instead. We had a unit test which covered this, but the assertion was that the returned conversation was `None`, but an empty list also returns as a `None` type.

Covered in this test - test_get_conversation_by_name_not_exists

amshamah419:
@yuvalbenshalom Just pushed changes addressing the CR.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2141,6 +2201,50 @@ def save_channels_to_context(updated_channel_list):
...
+
+    Args:
+        updated_users_list: A list of users
+    """
+    # Save conversations to cache
+    integration_context = get_integration_context(SYNC_CONTEXT)
+    users = integration_context.get('users')
+    if users:
+        users = json.loads(users)
+        users.extend(updated_users_list)
```
yuvalbenshalom, line 2242: this can create duplicates in the list. we need to rethink if users should be a dict, or you should have on function that adds one user and another function that replaces all the users after you downloaded the complete new list. (maybe for replacing you dont even need a separated function)

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2210,6 +2314,80 @@ def fetch_channels_command():
     demisto.results("Successfully updated channels to the Integration Context")
 
 
+def fetch_users_from_api(user_name: str = ''):
+    """
+    Fetches users based on their name
+
+    Args:
+        user_name: Optional - The human readable form of a user's name.
```
yuvalbenshalom, line 2322: This is not the correct explanation of the argument function.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2210,6 +2314,80 @@ def fetch_channels_command():
     demisto.results("Successfully updated channels to the Integration Context")
 
 
+def fetch_users_from_api(user_name: str = ''):
+    """
+    Fetches users based on their name
```
yuvalbenshalom, line 2319: rather, fetches all users in workspace or a specific user by name. no?

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2210,6 +2314,80 @@ def fetch_channels_command():
...
+    demisto.debug('Fetching users')
+    cursor = None
+    updated_users_list: list = []
+
+    res = CLIENT.users_list(limit=PAGINATED_COUNT, cursor=cursor)
+
+    while True:
+        cursor = res.get('response_metadata', {}).get('next_cursor')
+        members = res['members'] if res and res.get('members') else []
+        if user_name:
```
yuvalbenshalom, line 2336: please add a log that state we are now fetching the specific {user}.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2210,6 +2314,80 @@ def fetch_channels_command():
...
+                        'email': member.get('email', '')
+                    }
+                }
+                updated_users_list.append(updated_member)
+        if not cursor:
+            demisto.info("Reached the end of looking for a user")
+            break
+        else:
+            total_try_time = 0
+            while True:
```
yuvalbenshalom, line 2359: `while total_try_time < MAX_LIMIT_TIME:`

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2210,6 +2314,80 @@ def fetch_channels_command():
...
+                try:
+                    res = CLIENT.users_list(limit=PAGINATED_COUNT, cursor=cursor)
+
+                except SlackApiError as api_error:
+                    demisto.debug(f'Got rate limit error (sync). Body is: {str(res)}\n{api_error}')
+                    response = api_error.response
+                    headers = response.headers  # type: ignore
+                    if 'Retry-After' in headers:
+                        retry_after = int(headers['Retry-After'])
+                        total_try_time += retry_after
```
yuvalbenshalom, line 2369: you should increment total_try_time before `if 'Retry-After' in headers` because now there is non increment inn case  'Retry-After' not in headers.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1328,6 +1303,91 @@ async def fetch_channels_async():
         await asyncio.sleep(5)
 
 
+async def fetch_users_iterable():
```
yuvalbenshalom, line 1306: please add doc string.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1328,6 +1303,91 @@ async def fetch_channels_async():
...
+                res = await ASYNC_CLIENT.users_list(limit=PAGINATED_COUNT, cursor=cursor)
+
+            except SlackApiError as api_error:
+                demisto.debug(f'Got rate limit error (sync). Body is: {str(res)}\n{api_error}.'
+                              f'Retries count is - {total_try_time} out of max retries limit: {MAX_LIMIT_TIME}')
+                response = api_error.response
+                headers = response.headers  # type: ignore
+                if 'Retry-After' in headers:
+                    retry_after = int(headers['Retry-After'])
+                    total_try_time += retry_after
```
yuvalbenshalom, line 1350: you should increment total_try_time before if 'Retry-After' in headers because now there is non increment inn case 'Retry-After' not in headers.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1328,6 +1303,91 @@ async def fetch_channels_async():
...
+                        raise
+            demisto.debug("Response received from Slack users.list api. No retry is necessary.")
+            break
+
+    # Save users to cache
+    demisto.debug(f"updated_users_list has a length of {len(updated_users_list)}. Updating the integration context.")
+    users = integration_context.get('users')
+    if users:
+        users = json.loads(users)
+        users.extend(updated_users_list)
```
yuvalbenshalom, line 1370: we have duplicates problem here. we should consider if to merge, replace, or start working with dict.
This is probably also relevant to conversations.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1328,6 +1303,91 @@ async def fetch_channels_async():
...
+                'profile': {
+                    'email': member.get('email', '')
+                }
+            }
+            updated_users_list.append(updated_member)
+        if not cursor:
+            demisto.info("Finished updating users list")
+            break
+        total_try_time = 0
+        while total_try_time < MAX_LIMIT_TIME:
```
yuvalbenshalom, line 1339: I still think this can be an internal function and I have an idea how, but you can fix the other comments and about this we should talk separately.

Packs/Slack/Integrations/SlackV3/SlackV3_test.py
```diff
@@ -4092,3 +4110,151 @@ async def test_fetch_channels_paginated_async(mocker):
 
     # Assert
     assert returned_time.replace(microsecond=0) >= (last_update + timedelta(seconds=8)).replace(microsecond=0)
+
+
```
yuvalbenshalom, line 4114: general comment, we need to consider a test that checks that user list is merged correctly.(after the fix of the `.extend(updated_users)`)

yuvalbenshalom (changes requested):
see my comments. we will need to talk about some, to decide what to do(noted in them), but most of them can be fixed before the call.

Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2050,6 +2154,240 @@ def pin_message():
...
+                try:
+                    res = CLIENT.conversations_list(types='private_channel,public_channel', exclude_archived=True,
+                                                    cursor=cursor)
+                except SlackApiError as api_error:
+                    demisto.debug(f'Got rate limit error (sync). Body is: {str(res)}\n{api_error}')
+                    response = api_error.response
+                    headers = response.headers  # type: ignore
+                    if 'Retry-After' in headers:
+                        retry_after = int(headers['Retry-After'])
+                        total_try_time += retry_after
```
yuvalbenshalom, line 2294: need to move outside the if

yuvalbenshalom (commented):
added more comment for channels, from what I found about users.

Packs/Slack/Integrations/SlackV3/README.md
```diff
@@ -29,6 +29,8 @@ Please refer to the video tutorial [found here](https://live.paloaltonetworks.co
     | `filtered_tags` | Comma-separated list of tags by which to filter the messages sent from XSOAR. Only supported in Cortex XSOAR V6.1 and above. | False |
     | `permitted_notifications` | Types of Notifications to send in the dedicated channel. | False |
     | `channel_fetch_interval` | The frequency with which to update the channel IDs. **Note:** For workspaces that contain large amounts of channels, it is advised to periodically fetch the new channels. | False |
+    | `users_fetch_interval` | The frequency with which to update the user IDs. **Note:** For workspaces that contain large amounts of users, it is advised to periodically fetch the new users. | False |
+
```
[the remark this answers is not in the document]
amshamah419: I'm not sure. On the one hand, it would allow us to reduce the amount of params, but on the other hand it can allow customers to configure each independently.

PR #15969, opened 2021-11-24
Generic webhooks curl request headers fix in README.md

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/content-docs/issues/837

## Description
Fixed an issue where the described request in the README.md of the integration had a missing header. The given request results an error response, adding the header fixes that.


Noy-Maimon:
The change was only in the curl command, so adding the `docs-approved` tag.

reutshal (approved):

PR #15988, opened 2021-11-25
SplunkPy - convert value of labels to string

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/44066

## Description
In SplunkPy converting any value of incident label to str to aviod server exception when using json.Unmarshal


ilappe:
This is the solution recommended after discussion with @Itay4

yaakovi (commented):
please run the test without the fix to make sure the issue reproduces.
if it does, add RN and merge.

yaakovi (approved):

PR #15996, opened 2021-11-27
Darktrace - set token params to be creds

Itay4:
## Status
- [x] Ready

   - [x] No


anara123 (approved):

Itay4:
@anara123 ive hidden the deprecated params and the `Recommended` terminology from the new params

PR #15997, opened 2021-11-27
Jira v2 - set api token to be creds

Itay4:
## Status
- [x] Ready

   - [x] No


Packs/Jira/Integrations/JiraV2/JiraV2.yml
```diff
@@ -104,6 +105,11 @@ configuration:
   name: fetch_attachments
   required: false
   type: 8
+- display: API token (Deprecated)
```
bakatzir, line 108: should we set this param to `hidden: true`?
it will not harm the current integration instances, and will be ease new ones.

anara123: if we have hidden then we can

anara123 (approved):

Packs/Jira/Integrations/JiraV2/JiraV2.py
```diff
@@ -9,7 +9,7 @@
 
 ''' GLOBALS/PARAMS '''
 BASE_URL = demisto.getParam('url').rstrip('/') + '/'
-API_TOKEN = demisto.getParam('APItoken')
+API_TOKEN = demisto.getParam('APItoken') or (demisto.getParam('credentials') or {}).get('password')
```
anara123, line 12: add validation and raise exception if creds are missing

Packs/Jira/Integrations/JiraV2/JiraV2.yml
Itay4: good idea, thanks @bakatzir

Itay4:
@anara123 ive hidden the deprecated params and the Recommended terminology from the new params

bakatzir (approved):

PR #16006, opened 2021-11-28
ResolveShortenedURL: Add the ability to use requests instead of unshorten.me api

Bargenish:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/44080)

## Description
Added an argument to use requests instead of the unshorten.me API
Pending verification.

## Screenshots
The first command execution would only be appearing in case a connection establishment failed. other exceptions would be raised.
Screen Shot 2021-11-28 at 21 37 04


Bargenish:
Closing as the cs and custtomer were not responsive.
This is ready but tests. reopen if relevant

PR #16058, opened 2021-12-01
HashiCorp Vault - add concat username to cred param

Itay4:
## Status
- [x] Ready

## Description
added the `Concat username to credential object name` to handle the case there is more than one secret in the folder, which causes to fetch credentials under the same name and then integrations are unable to use them

   - [x] No

## Must have
- [x] Documentation


ShahafBenYakir:
Isn't the 'secret' variable secret? meaning wouldn’t it be a problem displaying it in the name?

Itay4:
> Isn't the 'secret' variable secret? meaning wouldn’t it be a problem displaying it in the name?

its just great naming by @liorblob - `secret` is actually the folder name

ShahafBenYakir (approved):

ShahafBenYakir (approved):

liorblob (approved):

PR #16089, opened 2021-12-02
Update ParseEmailFiles

nb-pdragoi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Include more information extracted from the email file.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


ShahafBenYakir:
@nb-pdragoi Please share some more details why is this change needed? What is the use-case?
FYI @moishce

nb-pdragoi:
@ShahafBenYakir When there are images embedded in the html body of the email, their source is related to their "content-id". Those same images are also seen as attachments. So, the only way to know which attachments are embedded in the email html is to also retrieve the content-id and then replace the source in the html body with the actual file path on the XSOAR server.

dansterenson:
@nb-pdragoi It seems that the following tests have failed due to some changes you did, please take a look and fix it.

image

PR #16169, opened 2021-12-06
PB name change in the Block IP Generic V3 playbook

ssokolovich:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [#44505](https://github.com/demisto/etc/issues/44505)

## Description
Changed a strange bug caused probably during the format action. 
Seems like the SDK changed the playbook to be with its ID instead with `playbookName` and `Name` variables.

## Screenshots
Wrong 🔴 :
image

Right ✅ :
image

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonPlaybooks/Playbooks/playbook-Block_IP_-_Generic_v3.yml
+++ b/Packs/CommonPlaybooks/Playbooks/playbook-Block_IP_-_Generic_v3.yml
@@ -3450,8 +3450,8 @@ inputs:
   value: {}
   required: false
   description: Insert a tag name with which indicators will get tagged. This tag can
-    be used later in the External Dynamic Lists integration by using the tag for
-    filtering IPs in the indicator query.
+    be used later in the External Dynamic Lists integration by using the tag for filtering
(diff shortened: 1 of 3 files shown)


ssokolovich:
Tested on a newly deployed server - seems to be fixed.
(Note - needed 2 refreshes to finally see the sub-playbook inputs after installing the "Cisco FirePower" pack 🤷  )

image
image

David-BMS (approved):
@ssokolovich maybe we need a validation to catch sub playbooks that have the default ID by regex. Please check with @idovandijk If this is a valid SDK issue.

idovandijk:
We need to understand how this happened - so far it looks like a bug with `format`. Not sure we need a validation for this, because theoretically format can destroy anything and we can't validate everything. But we do need to make sure that the SDK unit-tests cover such cases @tomneeman151293 @yaakovi

PR #16199, opened 2021-12-07
Update ParseEmailFiles

nb-pdragoi:

## Status
- [ ] In Progress
- [x] Ready

## Description
When there are images embedded in the html body of the email, their source is related to their "content-id". Those same images are also seen as attachments. So, the only way to know which attachments are embedded in the email html is to also retrieve the content-id and then replace the source in the html body with the actual file path on the XSOAR server.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/ParseEmailFiles/ParseEmailFiles.py
```diff
@@ -3719,7 +3746,14 @@ def handle_eml(file_path, b64=False, file_name=None, parse_only_headers=False, m
...
                 'HeadersMap': headers_map,
-                'Attachments': ','.join(attachment_names) if attachment_names else '',
+                'Attachments': [
+                    {
+                        "Name": attachment_names[i],
+                        "Content-ID": attachment_content_ids[i],
+                        "Content-Disposition": attachment_content_dispositions[i],
+                        "FilePath": attachment_internal_path[i]
+                    } for i in range(len(attachment_names))
+                ],
```
anara123, line 3756: this is breaking backward compatibility.
we can add this data under different prop name.
like `AttachmentsData`

nb-pdragoi: Hello @anara123 

This is correct. However, is there any other reason we want to keep `Attachments` the way it is? It contains exactly the same infor as `AttachmentNames` which we have already, only the names are not in a list, but join together. This can easily be done using a transformer anyway.

Moreover, `Attachments` wasn't even included in the yml file as an output, so I doubt anyone was ever using it instead of `AttachmentNames`.

Thanks

anara123: Correct, the reason we return `Attachments` is because of backward compatibility. This is an output from older versions. That is why we don't included it in the yml (in order to prevent from users to use it), but we do return it to the  context because of BC.
So please update your code.

dorschw (approved):
Well done

dorschw (approved):
Nice

PR #16345, opened 2021-12-13
Fix set phisihing details

JasBeilin:
## Status
- [x] In Progress
- [x] Ready

Changed 'SetByIncidentId' to 'set' command using ExecuteCommandAt since dbotrole is not being considered when wrapping a low permissions script.


DeanArbel (approved):
That's great.

PR #15669, opened 2021-11-08
Fixed race condition in phishing campaigns

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40205

## Description
Fixed an issue that caused multiple phishing campaigns to be created at the same time instead of just one. I did it by adding a lock mechanism using the Demisto Lock integration.

## Screenshots
Detect_ _Manage_Phishing_Campaigns

## Minimum version of Cortex XSOAR
6.0.0

## Does it break backward compatibility?
No

PR #16417, opened 2021-12-16
Fixed race condition in phishing campaigns and layout overhaul

idovandijk:
## Status
**Ready**
~**In Progress** - Merge after https://github.com/demisto/etc/issues/43640 and https://github.com/demisto/etc/issues/43860 and https://github.com/demisto/etc/issues/43586~

## Related Issues
fixes: https://github.com/demisto/etc/issues/40205
This PR is the same as https://github.com/demisto/content/pull/15669 which was reverted eventually

## Description
Fixed an issue that caused multiple phishing campaigns to be created at the same time instead of just one. I did it by adding a lock mechanism using the Demisto Lock integration.

## Screenshots
Detect_ _Manage_Phishing_Campaigns

6.0.0

No


idovandijk:
Test is failing now because of https://github.com/demisto/etc/issues/44859
I will make an adjustment

David-BMS (approved):

PR #16425, opened 2021-12-16
EDL fix for invalid IPs

PA-Rweins:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/44526)

## Description
- Before collapsing to either CIDR or an IP range each IP address or CIDR is validated.
- Invalid IPs/CIRDs will be shown as part of the list but will not be collapsed.

## Screenshots
Screen Shot 2021-12-16 at 17 43 56

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/EDL/Integrations/EDL/EDL.py
```diff
@@ -216,14 +217,51 @@ def ips_to_ranges(ips: Iterable, collapse_ips: str):
...
+    if collapse_ips == COLLAPSE_TO_RANGES:
+        ips_range_groups = IPSet(valid_ips).iter_ipranges()
+        collapsed_list = ip_groups_to_ranges(ips_range_groups)
     else:
-        cidrs = IPSet(ips).iter_cidrs()
-        return ip_groups_to_cidrs(cidrs)
+        cidrs = IPSet(valid_ips).iter_cidrs()
+        collapsed_list = ip_groups_to_cidrs(cidrs)
+
+    collapsed_list.update(invalid_ips)
```
Bargenish, line 236: Why do we still want to add the invalid ips?

Packs/EDL/Integrations/EDL/EDL.py
```diff
@@ -216,14 +217,51 @@ def ips_to_ranges(ips: Iterable, collapse_ips: str):
...
+    return collapsed_list
+
+
+def is_valid_ip(ip: str) -> bool:
```
Bargenish, line 240: let's check the structure and only have maximum, 2 checks per IP.
not sure how scalable it is.
@DeanArbel wdyt?

Bargenish (changes requested):

Packs/EDL/Integrations/EDL/EDL.py
DeanArbel: It's a valid usecase. Could be a valid indicator that was created with a wrong type, so it should be in the EDL, but not calculated under collapsible ips.

Packs/EDL/Integrations/EDL/EDL.py
DeanArbel: I agree.
I think it's better to drop the `generic` implementation of `ip_groups_to_ranges` and `ips_to_ranges` by passing an extra argument - the type of expected `IP` / `CIDR` - that way the code doesn't need to be made of 4 `try/except`.

PA-Rweins: Fixed

DeanArbel (approved):
Good job

Bargenish (approved):

PR #16426, opened 2021-12-16
MicrosoftGraphGroups - Set enc_key to support creds

yaakovpraisler:
## Status
- [x] Ready

   - [x] No


Packs/MicrosoftGraphGroups/Integrations/MicrosoftGraphGroups/MicrosoftGraphGroups.py
```diff
@@ -413,7 +413,7 @@ def main():
     base_url = params.get('url', '').rstrip('/') + '/v1.0/'
     tenant = params.get('tenant_id')
     auth_and_token_url = params.get('auth_id')
-    enc_key = params.get('enc_key')
+    enc_key = params.get('enc_key') or (params.get('credentials') or {}).get('password')
```
Itay4, line 416: need to verify `enc_key` was provided since both params are not mandatory

Packs/MicrosoftGraphGroups/ReleaseNotes/1_0_13.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Azure Active Directory Groups
+- Added the *Key* integration parameter to support credentials fetching object.
```
dantavori, line 4:
```suggestion
- Added the *Key* integration parameter to support credentials fetching object.
  ***Note:*** if an integration instance is already configured, no action is required.
```
@Itay4 correct me if I'm wrong, worth mentioning

dantavori (approved):

Packs/MicrosoftGraphGroups/ReleaseNotes/1_0_13.md
Itay4: i think its a bit misleading as it might suggest that without action the user can now use the credential object
i would leave it is without this addition

PR #16466, opened 2021-12-18
Use signal profiling dump only on non-Windows OS

tkatzir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/44894

## Description
There's no `signal` in Windows OS.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -27,8 +27,15 @@ from threading import Lock
 import demistomock as demisto
 import warnings
 
-import signal
-
+OS_LINUX = False
+OS_MAC = False
+OS_WINDOWS = False
+if sys.platform == "linux" or sys.platform == "linux2":
+    OS_LINUX = True
+elif sys.platform == "darwin":
+    OS_MAC = True
+elif sys.platform == "win32":
+    OS_WINDOWS = True
 
 class WarningsHandler(object):
     #    Wrapper to handle warnings. We use a class to cleanup after execution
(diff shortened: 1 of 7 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -27,7 +27,15 @@
 import demistomock as demisto
 import warnings
 
-import signal
+OS_LINUX = False
+OS_MAC = False
+OS_WINDOWS = False
+if sys.platform == "linux" or sys.platform == "linux2":
```
anara123, line 33: use `startswith` instead of `==` as mentioned here https://docs.python.org/3/library/sys.html#sys.platform

tkatzir: Done.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -8515,20 +8509,35 @@ def signal_handler_profiling_dump(_sig, _frame):
     LOG.print_log()
 
 
-def register_signal_handler_profiling_dump(signal_type=signal.SIGUSR1, profiling_dump_rows_limit=PROFILING_DUMP_ROWS_LIMIT):
+# DEPRECATED - use register_signal_handler_profiling_dump instead
+def register_signal_handler_threads_dump():
```
anara123, line 8513: remove this function

anara123 (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
tkatzir: Removed.

PR #16500, opened 2021-12-19
BaseClient - support setting baseclient http read timeout via env var

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37569

   - [x] No

## Must have
- [x] Tests


anara123 (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -7288,7 +7296,7 @@ def _http_request(self, method, url_suffix='', full_url=None, headers=None, auth
                     files=files,
                     headers=headers,
                     auth=auth,
-                    timeout=timeout,
+                    timeout=(10, timeout),
```
glicht, line 7299: I would leave this as it was instead of hardcoding the 10. It will lead to edge cases.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -7096,6 +7098,9 @@ def __init__(self, base_url, verify=True, proxy=False, ok_codes=tuple(), headers
             if not verify:
                 skip_cert_verification()
 
+            # removing trailing = char from env var value added by the server
+            self.timeout = float(os.environ.get('REQUESTS_TIMEOUT.' + get_integration_name(), '')[:-1] or os.environ.get('REQUESTS_TIMEOUT', '')[:-1] or self.REQUESTS_TIMEOUT)  # noqa: E501
```
glicht, line 7102: Add the timeout to the constructor too, so integrations that implement a param for controlling timeout can pass this in.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -7170,7 +7175,7 @@ def _implement_retry(self, retries=0,
                 pass
 
         def _http_request(self, method, url_suffix='', full_url=None, headers=None, auth=None, json_data=None,
-                          params=None, data=None, files=None, timeout=10, resp_type='json', ok_codes=None,
+                          params=None, data=None, files=None, timeout=REQUESTS_TIMEOUT, resp_type='json', ok_codes=None,
```
glicht, line 7178: Set to None by default. If it is None, then use self.timeout. That way integration can override the value for specific cases.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht: Small cleanup use `os.getenv` so it is consistent with our other code sections that lookup env vars.

glicht: Also, not clear why you have `[:-1]`? It is removing the last char

glicht (changes requested):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
glicht: We should also support scripts that may use the client. So maybe modify `get_integration_name` or create a new function that works with scripts. For scripts we need: `context.ScriptName`

glicht (approved):

Itay4:
lint fails on `context deadline exceeded` 

lint on base pack passed:
```
#########
 Summary 
#########
Packages: 26
Packages PASS: 26
```

PR #16642, opened 2021-12-26
Mcafee atd creds

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/12772

   - [x] No


Packs/McAfee_Advanced_Threat_Defense/Integrations/McAfee_Advanced_Threat_Defense/McAfee_Advanced_Threat_Defense.py
```diff
@@ -798,6 +798,8 @@ def main():
     handle_proxy()  # Remove proxy if not set to true in params
     global API_HEADERS
     API_HEADERS = get_headers()
+    if not USERNAME and not PASSWORD:
```
Itay4, line 801: i think we need to move it to the top as the params are used in line 41

Itay4 (approved):

PR #16653, opened 2021-12-27
Demisto Rest API: JSON parse body only if exist

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45136)

## Description
Try to parse the body as json only if is exist


Packs/DemistoRESTAPI/Integrations/DemistoRESTAPI/DemistoRESTAPI.js
```diff
@@ -120,7 +120,11 @@ switch (command) {
         sendRequest('GET','user');
         return 'ok';
     case 'demisto-api-post':
-        var body = JSON.parse(args.body);
+        if(args.body)
+            var body = JSON.parse(args.body);
+        else
+            logDebug('The body is empty.')
+
         return sendRequest('POST',args.uri, args.body);
```
ilappe, line 128: @yaakovi 
we parse the body but didn't send the body variable to the request 
is this by mistake?

yaakovi: don't think so, please test this behavior

yaakovi (commented):
please test this change (enhance the test playbook)

Packs/DemistoRESTAPI/Integrations/DemistoRESTAPI/DemistoRESTAPI.js
ilappe: checked 
it fail if we send the body var

yaakovi (approved):

PR #16682, opened 2021-12-28
Improved the Phishing - Core test

idovandijk:
## Status
**In Progress** - test won't pass until https://github.com/demisto/etc/issues/45133 is fixed. Also needs https://github.com/demisto/etc/issues/44605 which is a script to use here.

## Related Issues
Related to: https://github.com/demisto/etc/issues/44049
and https://github.com/demisto/etc/issues/41432

## Description
Improved the Phishing - Core test:
- Added a check for indicator extraction from DOCX file
- Added a check to ensure EML is parsed correctly and playbook sets email to, from, subject, body and headers correctly in the layout.

6.0.0

No


idovandijk:
@dansterenson We have an interesting case here. The build should fail because:
1. It uses a subplaybook that doesn't exist yet
2. The checks it performs should lead to a failure because of a `ParseEmailFiles` bug

However, this test is not a standalone test, meaning a wrapper playbook should be running it, so maybe because of that the test is not triggered at all and we don't get a failure. We can agree that this should fail though.
Can you let me know what check is missing here and what we can do and what issue I can open? And also can it be related to `Phishing - Core - Test - Actual Incident` not being in conf.json? I deleted it earlier from there because the `Incident Starter` is what should be running.

dansterenson:
@esharf FYI

PR #16710, opened 2021-12-29
Timeout ut bug

Noy-Maimon:


yaakovi:
actual fix is here: https://github.com/demisto/content/pull/16711

PR #16731, opened 2021-12-30
Phishing - Generic v3 - new test playbooks

idovandijk:
## Status
**In Progress:**
- Need the ~subplayook that waits for and completes tasks https://github.com/demisto/content/pull/16733~ WaitAndCompleteTask script https://github.com/demisto/etc/issues/44605
- Need event generator https://github.com/demisto/etc/issues/38628 and then need to update the test playbook to use it
- Need ability to change playbook inputs (to configure internal domains) https://github.com/demisto/etc/issues/44414

## Related Issues
Related to: https://github.com/demisto/etc/issues/44049
https://github.com/demisto/etc/issues/44364

## Description
Adding tests for `Phishing - Generic v3`.

6.1.0

No


idovandijk:
Status: Looks like `Get Original Email + Search & Delete - Test` is failing because the mail sender fails, which makes the test fail and reset the playbook inputs back to the defaults, which in turn makes it not retrieve the original email, which results in the failure to extract one URL. All other tasks are successful (except for delete which hasn't run because of the same issue):
image

The other test still has an issue - it seems that the incident is being investigated despite the test not started the investigation, so the test fails when trying to start investigation of an already investigated incident

idovandijk:
Held back by build blockers::
- EWS issue (possibly throttling) https://github.com/demisto/etc/issues/47302
- Classifier + mapper config https://github.com/demisto/etc/issues/47389

idovandijk:
Related PR to speed up future builds here: https://github.com/demisto/content/pull/18165

idovandijk:
Made all the necessary fixes (also updated checks after latest changes to URL regex). Now blocked by https://github.com/demisto/etc/issues/47302
image

idovandijk:
Still EWS issue https://github.com/demisto/etc/issues/47302
image

idovandijk:
Update - test is failing only on master, investigating a new issue with mapping (might be a server bug) https://github.com/demisto/etc/issues/47625

tomer-pan (approved):

PR #16497, opened 2021-12-19
added trigger_build_with_ttl.sh to gitlab triggers

esharf:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38241

## Description
Added the option to trigger a build with time to live in Gitlab.

PR #16811, opened 2022-01-03
phishing layout fix

tomer-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45365

## Description
After phishing v3 update, versions lower than 6.1 will get the new layout which will not show all of the new fields. Instead we should return the older layout for older versions and change the id of the new one for the newer version.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Phishing/IncidentTypes/New_6_1/incidenttype-Phishing_6.1.json
+++ b/Packs/Phishing/IncidentTypes/New_6_1/incidenttype-Phishing_6.1.json
@@ -21,7 +21,7 @@
  "reputationCalc": 2,
  "onChangeRepAlg": 2,
  "fromVersion": "6.1.0",
- "layout": "Phishing_v3",
+ "layout": "Phishing Incident v3",
  "extractSettings": {
  "mode": "Specific",
   "fieldCliNameToExtractSettings": {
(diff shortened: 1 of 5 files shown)


idovandijk (commented):
Files look good to me, but please also check in 6.0 and 6.1 servers that it actually works. You can use[ the new feature](https://github.com/demisto/content/pull/16497) to trigger builds and keep servers up for X minutes.
And need Yaron to approve RN

tomer-pan:
> Files look good to me, but please also check in 6.0 and 6.1 servers that it actually works. You can use[ the new feature](https://github.com/demisto/content/pull/16497) to trigger builds and keep servers up for X minutes. And need Yaron to approve RN

Once I will see that the build is passing I will test both 6.0 and 6.1 with the new url to verify it works as planned.

yaron-libman:
Doc review completed.

idovandijk (approved):
Assuming you checked on the servers

PR #16867, opened 2022-01-04
Fix Secret detections

yaakovi:
## Status
- [x] In Progress
- [ ] Ready

## Description
ignore certificate validation in secret detection request


abaumgarten (changes requested):
It won't solve the problem

PR #16869, opened 2022-01-04
SQS fetch_incidents will respect the fetch limit instead of pulling only up to 10

jrauen:

## Status
- [x] In Progress
- [ ] Ready

## Description
The current SQS script will only pull up to 10 messages from the SQS queue and then exit. There is also no check on whether the number of messages retrieved is less than the max_fetch parameter. This change puts the receive_message function call into a while loop so multiple calls to the queue will happen in batches of ten. This occurs until the number of messages received and put into the incidents list is equal to the max_fetch parameter. 

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


merit-maita (commented):
Hello
first of all thank you for your contribution.
the provided solution with max_fetch does not work as we expect it to in Demisto.
to make things clear, i'll explain what is max_fetch in demisto, and what does it do in the provided solution:
in demisto: max_fetch param is the **maximum**  number of incidents we'd want to pull at once, for example if max_fetch = 50 , once we send the request to fetch incidents, does not matter how many incidents we got back, we'd want to return in the end a maximum number of 50 incidents. if we got less than 50 incidents, then its okay, they all will be returned, but if we got more than 50 incidents, only 50 incidents will be returned in the end. (thats why we use the first_fetch param and lest_run, to keep up with what was returned in the previous call, so we wouldnt return it again, in the next 50 incidents for example).
in the provided solution: the max_fetch param is treated as the **exact amount** of incidents to be returned, you used a loop to keep on fetching incidents until we get the same amount of max_fetch, which's not correct.
On the other hand:
lets say we would want to implement the max_fetch param feature with demisto logic, the API itself allows to return only up to 10 messages (incidents) at once. [look here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message),  so the maximum amount of incidents we'll ever get is 10.
which means, the only way for max_fetch param implementation to be useful is if its given the value of 10 or less. otherwise it wont since we'll never catch more than 10 anyways.
so now the question is: is it important for you to catch incidents with max_fetch in such case (maximum 10)? please let me know what you think so we can come up with the right solution .
Thank you.

Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
```diff
@@ -147,33 +147,36 @@ def parse_incident_from_finding(message):
...
+                WaitTimeSeconds=5,
+            )
+
+            if "Messages" not in messages.keys():
+                if len(incidents) == 0:
+                    if demisto.command() == 'fetch-incidents':
+                        demisto.incidents([])
+                    return messages, incidents, receipt_handles
+                else:
+                    break
```
jrauen, line 169: Make sure to review the second commit from this morning. I did catch that scenario you describe where the number of retrievable incidents is less than the configured max_fetch parameter. If the queue ever runs out of messages then the response will not contain the `Messages` key and the break statement will exit the while loop.

jrauen:
I understand the desire to limit to one call of an API per execution but that just isn't practical with SQS' very low message limit. It effectively means the ingestion rate is capped at 10 incidents per minute which is insufficient in a high volume environment. 

If the concern is prolonged execution of the integration then I would suggest a hard cap for the max_fetch parameter (e.g., 100), which will override a user-defined parameter if it's in excess of the cap. And setting the default max_fetch to 10 from 50.

Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
merit-maita: you're right, it makes max_fetch works as it's expected to

merit-maita:
> I understand the desire to limit to one call of an API per execution but that just isn't practical with SQS' very low message limit. It effectively means the ingestion rate is capped at 10 incidents per minute which is insufficient in a high volume environment.
> 
> If the concern is prolonged execution of the integration then I would suggest a hard cap for the max_fetch parameter (e.g., 100), which will override a user-defined parameter if it's in excess of the cap. And setting the default max_fetch to 10 from 50.

i see the problem you're referring to , you're right , i'll discuss it with my TL tomorrow and i will update ASAP.

merit-maita:
Hello @jrauen 
the idea you provided above seems good, if you like you can go ahead and implement  it, or you can open an enhancement request.
thanks!

jrauen:
I've implemented the enhancement in the latest commit. The max_fetch param is now set to 10. Also added a check for if the parameter was set to less than 1 or greater than 100. If max_fetch is less than 10 the SQS API call will set the MaxNumberOfMessages parameter to the max_fetch value.

Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
```diff
@@ -203,6 +207,13 @@ def main():
     timeout = params.get('timeout')
     retries = params.get('retries') or 5
     aws_queue_url = params.get('queueUrl')
+    max_fetch_param = params.get('max_fetch', 10)
+    if max_fetch_param < 1:
+        max_fetch = 1
+    elif max_fetch_param > 100:
+        max_fetch = 100
+    else:
+        max_fetch = max_fetch_param
```
merit-maita, line 216:
```suggestion
    max_fetch = min(params.get('max_fetch', 10), 100)
```

merit-maita (changes requested):
can you please add a description of the param max_fetch in reamde file (state that default is 10). don't forget to add Release Notes and update the pack version in pack_metadata file.
Thanks!

Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
jrauen: The problem with this is there needs to be a hard cap of 1 on the `max_fetch` which this dropped. To do what you're suggesting here we would need:
```python
max_fetch = max(min(params.get('max_fetch', 10), 100), 1)
```
This prevents a `max_fetch` of zero or negative numbers which would cause issues in the SQS API call which now directly uses the parameter. But when put into a single line it's not very readable or clear as to what's going on.

jrauen:
> can you please add a description of the param max_fetch in reamde file (state that default is 10). don't forget to add Release Notes and update the pack version in pack_metadata file. Thanks!

done

Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
merit-maita: i get what you're trying to say, but it does not make any sense to pass a value lower than 1, i dont think anyone  would pass -1 for example as a max fetch, in case they did, then its fine to get an error related to it.
the edit i added is good.
thanks for the effort tho!

merit-maita:
please make sure to pass the checks.

merit-maita (approved):

PR #16886, opened 2022-01-05
Feed mitre attackv2 links fix

iyeshaya:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45399


Shellyber:
Docker image won't be updated because of the following: https://github.com/demisto/etc/issues/41040. (still having issues with the taxii2 docker image).
Need to force merge

@Itay4

PR #16983, opened 2022-01-10
added rule to run-unittests-and-lint step to run when it is not upload

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
The upload build had a duplicate `run-unittests-and-lint` step 
added a rule that verifies whether the build is upload or not, and triggers the on-push step only if not upload.


.gitlab/ci/on-push.yml
```diff
@@ -26,6 +26,8 @@ trigger-private-build:
 run-unittests-and-lint:
   extends:
     - .run-unittests-and-lint
+  rules:
+    - if: '$BUCKET_UPLOAD == "false"'
```
yaakovi, line 30: isn't the default value is empty (variable doesn't exist)?
looking at your last build, this job did not run in your regular flow.

yaakovi (changes requested):

.gitlab/ci/on-push.yml
yucohen: correct
changed it.
https://code.pan.run/xsoar/content/-/pipelines/2236902

PR #17050, opened 2022-01-12
Fixes archer-search-records bug

yaakovpraisler:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45630)

## Description
Fixes the Archer search record bug when searching with incident id.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


Packs/ArcherRSA/Integrations/ArcherV2/ArcherV2.py
```diff
@@ -470,22 +478,25 @@ def search_records(
         search_field_name = ''
         search_field_id = ''
         fields_mapping = level_data['mapping']
+        level_id = level_data['level']
```
dantavori, line 481: can you elaborate on this field? will it always appear in the level data?

dantavori (approved):
looks great

Packs/ArcherRSA/Integrations/ArcherV2/ArcherV2.py
yaakovpraisler: We add it in the code [here](https://github.com/demisto/content/blob/master/Packs/ArcherRSA/Integrations/ArcherV2/ArcherV2.py#L354)

yaakovpraisler:
Validation fails due to non-updated docker image. It was not updated as it fails the build.
@Itay4 - Need a force merge.

PR #17095, opened 2022-01-13
Phishing domainsquatting fix

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/43001

## Description
Fixed a bug that caused the domain-squatting check to produce incorrect results.
The bug has been there for a very long time, and happened because of the old change for `WhereFieldEquals` transformer where they changed the "stringify" argument to "true" by default.

## Screenshots
Before the fix:
image

After the fix:
image

- 5.0.0

No


tschanfeld:
looks good

tomer-pan (approved):

PR #17098, opened 2022-01-13
Phishing v3 - fixed an issue where the reporter email would sometimes be enriched twice

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45738

## Description
Fixed an issue where the reporter email address may sometimes be enriched twice.

Longer explanation:
In the email address enrichment playbook (under phishing), we used the non-sanitized field `ReporterAdderss` instead of `incident.reporteremailaddress` field in a filter. This caused the reporter email address to be enriched again if the customer is using a mail listener that sends the email from with the name, like `Ido Van Dijk <ivandijk@paloaltonetworkscom>`.

## Screenshots
image

image

6.1.0

No

### DIFF
--- a/Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
+++ b/Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
@@ -2763,8 +2763,7 @@ inputs:
     type using machine learning.
   playbookInputQuery:
 - key: GetOriginalEmail
-  value:
-    simple: "True"
+  value: {}
   required: false
   description: |-
     For forwarded emails. When "True", retrieves the original email in the thread.
--- a/Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
+++ b/Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
@@ -8,10 +8,10 @@ starttaskid: "0"
 tasks:
   "0":
     id: "0"
-    taskid: 664506b8-2e32-418e-8366-975615da1945
+    taskid: a70c1df8-6bfa-48bc-8f84-d4d09e9f8269
     type: start
     task:
-      id: 664506b8-2e32-418e-8366-975615da1945
+      id: a70c1df8-6bfa-48bc-8f84-d4d09e9f8269
       version: -1
       name: ""
       iscommand: false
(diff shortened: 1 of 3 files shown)


Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
```diff
@@ -1866,7 +1865,7 @@ tasks:
                 iscontext: true
               right:
                 value:
-                  simple: ReporterAddress
+                  simple: incident.reporteremailaddress
```
tomer-pan, line 1868: The field `incident.reporteremailaddress` will hold the value of `ReporterAddress` since it is being set in task 85. `ReporterAddress` is in fact holding the sanitized value (task 134 + 12).
I can see you moved this task (85) to the top of the playbook which seems logical, however, I don't see the difference with using either one of `ReporterAddress` and `incident.reporteremailaddress` since both will hold the same value.
`ReporterAddress` is in fact holding the sanitized value.

Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
```diff
@@ -2766,7 +2763,8 @@ inputs:
     type using machine learning.
   playbookInputQuery:
 - key: GetOriginalEmail
-  value: {}
+  value:
+    simple: "True"
```
tomer-pan, line 2767: We want it to be empty or "False" since current customers with no permissions will get an error otherwise.
If it's empty, they will get the default value from the "Process Email" playbook (which is "False" by default)

Packs/Phishing/ReleaseNotes/3_0_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Playbooks
+##### Phishing - Generic v3
+- %%UPDATE_RN%%
```
tomer-pan, line 4: Description is missing. Also, the versio should be changed in the pack metadata file.

idovandijk: lol for some reason the changes weren't pushed..
I <3 VS Code

Packs/Phishing/Playbooks/Phishing_-_Generic_v3.yml
idovandijk: oh you're actually right.. so this makes the PR pointless

idovandijk: I might introduce this (logical) change in another PR with another fix, but I'll close this PR

idovandijk:
Closing because the change is not needed (ReporterAddress is already sanitized)

PR #17234, opened 2022-01-21
McAfee ePO v2 integration fix for Issue Number #45294

wissamg:
id: #45294

## Status
- [X] Ready

## Related Issues
fixes: [45294](https://github.com/demisto/etc/issues/45294)

## Description
After the session with the customer we have identified that some keys in the return dict has renamed in version 5.10.0 
The code was changed to be generic and accept all keys as return from the ePO server.

 - [X] No

## Must have
- [X] Tests


Shellyber (approved):
Great work.
Please fix validation in the build - need to update the docker image.

ShahafBenYakir:
Force merging - docker image shouldn't be updated

PR #17245, opened 2022-01-23
replace npm install with npm link

esharf:
## Status
- [x] Ready

## Description
replace npm install with npm link.


.gitlab/ci/global.yml
```diff
@@ -96,7 +96,8 @@
 .install_node_modules: &install_node_modules
   - section_start "Installing node modules" --collapsed
   - source $NVM_DIR/nvm.sh
-  - nvm use default
+  - nvm install 14.14.0
```
Itay4, line 99: lets hold on with that change, and if we do make it, lets just change the value in `.nvmrc`

yaakovi (approved):

avidan-H (approved):

PR #17293, opened 2022-01-24
Move artifact folder

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
* Fixed hard-coded references of the artifact folder to use a CLI input or the environment variable.
* Exposed the artifacts folder as ARTIFACTS_FOLDER and set it in the appropriate steps to ARTIFACTS_FOLDER_XSOAR.


.gitlab/ci/bucket-upload.yml
```diff
@@ -156,7 +162,7 @@ upload-packs-to-marketplace:
         fi
 
         if [[ $GCS_MARKET_BUCKET != $GCS_PRODUCTION_BUCKET ]] && [[ -z $STORAGE_BASE_PATH ]]; then
-         STORAGE_BASE_PATH="upload-flow/builds/$CI_COMMIT_BRANCH/$CI_PIPELINE_ID/xsoar/content"
+         STORAGE_BASE_PATH="upload-flow/builds/$CI_COMMIT_BRANCH/$CI_PIPELINE_ID/content"
```
avidan-H, line 165: Do we want different bucket path directories depending on whether it's `xsoar` or `mpv2`?

.gitlab/ci/on-push.yml
```diff
@@ -160,10 +168,16 @@ create-instances:
       fi
     - !reference [.open-ssh-tunnel]
     - python3 ./Tests/scripts/wait_until_server_ready.py "$INSTANCE_ROLE"
+    - |
+      # workaround for the hard-coded value in the sdk
+      cp "$ARTIFACTS_FOLDER/env_results.json" "./artifacts/env_results.json"
+      cp "$ARTIFACTS_FOLDER/filter_file.txt" "./artifacts/filter_file.txt"
+
```
avidan-H, line 175: This will be removed after the `demisto-sdk` is updated?

.gitlab/ci/on-push.yml
```diff
@@ -160,10 +168,16 @@ create-instances:
...
     - ./Tests/scripts/install_content_and_test_integrations.sh "$INSTANCE_ROLE" || EXIT_CODE=$?
     - cp -f $ARTIFACTS_FOLDER/conf.json Tests/conf.json
     - echo Going to sleep for 60 seconds to allow server finish indexing
     - sleep 60
+
```
avidan-H, line 180: nice - readability for the win. 👓

.gitlab/ci/bucket-upload.yml
```diff
@@ -168,7 +174,7 @@ upload-packs-to-marketplace:
         STORAGE_BASE_PATH="content"
       fi
       if [[ $GCS_MARKET_BUCKET != $GCS_PRODUCTION_BUCKET ]] && [[ -z $STORAGE_BASE_PATH ]]; then
-        STORAGE_BASE_PATH="upload-flow/builds/$CI_COMMIT_BRANCH/$CI_PIPELINE_ID/xsoar/content"
+        STORAGE_BASE_PATH="upload-flow/builds/$CI_COMMIT_BRANCH/$CI_PIPELINE_ID/content"
       fi
 
       PACKS_SRC="gs://$GCS_MARKET_BUCKET/$STORAGE_BASE_PATH/packs"
```
avidan-H, line 180: We don't need to override the variables in the `force-pack-upload`?

Tests/scripts/collect_tests_and_content_packs.py
```diff
@@ -139,15 +139,16 @@ def get_test_playbooks_configured_with_integration(self, integration_id):
 ID_SET = {}
 CONF: TestConf = None  # type: ignore[assignment]
```
avidan-H, line 141: Can we maybe perform all the `os.path.join(ARTIFACTS_FOLDER, <file-name.ext>)` as assignments to variables before the `if-elif-else` block conditions? Nesting it inside the call to `open` impinges on the readability a bit IMO plus we are repeating the ops twice here, once in the conditional expression and again in the call to `open` the file.

avidan-H (changes requested):
Did you double-check that there weren't other jobs that we need to override the environment variables (`ARTIFACTS_FOLDER` etc.)?

yaakovi:
> Did you double-check that there weren't other jobs that we need to override the environment variables (`ARTIFACTS_FOLDER` etc.)?

I checked the:
* on-push flow: relevant for branch builds, instance-test, and nightly
* bucket-upload flow: relevant for the upload
* sdk-nightly flow: did not touch, will have the original path

did not modify the `lint & unit tests` and `validate` jobs - they will keep the original path.

.gitlab/ci/bucket-upload.yml
yaakovi: it uploads to different buckets so don't think we need this indication in the path as well (duplicate information)

.gitlab/ci/on-push.yml
yaakovi: yes

.gitlab/ci/bucket-upload.yml
yaakovi: we probably should, as far as I remember, there are a few problems with that flow at the moment, and I didn't want to go down the rabbit hole

avidan-H (approved):

esharf (approved):

PR #17294, opened 2022-01-24
Mcafee tie fix

PA-Rweins:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/44989)

## Description
McAfee get reputation might return an error, handled error so a better explanation will be returned.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/McAfee-TIE/Integrations/McAfee-TIE/McAfee-TIE.py
```diff
@@ -176,6 +176,14 @@ def test():
         client.disconnect()
 
 
+def safe_get_file_reputation(tie_client, hash_param):
```
yaakovi, line 179: the proper fix should be to return an `Unknown` score instead of raising an error

yaakovi (changes requested):

Packs/McAfee-TIE/Integrations/McAfee-TIE/McAfee-TIE.py
PA-Rweins: Umm but what if McAfee will return some unknown error that we've not encountered yet?
Maybe add demisto log with the error and return unknown score?

anara123: When indicator will not be found, the returned error from the API will be very specific. For example 400 error status or something like this.
You will have to check if it is that error, then return Unknown
Other exception, then fail the command

Packs/McAfee-TIE/Integrations/McAfee-TIE/McAfee-TIE.py
```diff
@@ -176,6 +176,15 @@ def test():
...
+    try:
+        res = tie_client.get_file_reputation(hash_param)
+    except Exception as e:
+        demisto.log("McAfee failed to get file reputation with error: " + str(e))
```
yaakovi, line 183:
```suggestion
        demisto.info("McAfee failed to get file reputation with error: " + str(e))
```

Packs/McAfee-TIE/ReleaseNotes/1_0_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### McAfee Threat Intelligence Exchange
+- Added informative message for ***file*** command error.
```
yaakovi, line 4:
```suggestion
- Fixed an error where the ***file*** command failed for hashes for which there was no analysis.
```

yaakovi (approved):

PR #17373, opened 2022-01-30
Add dependencies.zip upload to prod

DeanArbel:
## Status
- [x] Ready

## Description
Add dependencies.zip upload to prod


Tests/Marketplace/marketplace_services.py
```diff
@@ -1140,8 +1140,43 @@ def copy_and_upload_to_storage(self, production_bucket, build_bucket, successful
                     self._aggregation_str = agg_str
             logging.success(f"Uploaded {self._pack_name} pack to {prod_pack_zip_path} path.")
 
+        # handle dependenices zip upload when found in build bucket
+        self.copy_and_upload_dependencies_zip_to_storage(
```
Noy-Maimon, line 1144: This should be in a condition since some uploads will not have those zipped dependencies (such as marketplacev2)

Noy-Maimon (commented):
Looks good, lets verify with an upload test run using the `trigger_test_upload_flow.sh` script.
In order to use the env var for the zipped packs, you should add it to that script as an option (this should be done for any new upload variable anyway)

Noy-Maimon (approved):

PR #17398, opened 2022-02-01
esm_v2_eol_update

bakatzir:
## Status
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/46267

## Description
updating support to match McAfee EOL https://kc.mcafee.com/corporate/index?page=content&id=KB94822

   - [x] No

### DIFF
--- a/Packs/McAfee_ESM/Integrations/McAfee_ESM_v2/README.md
+++ b/Packs/McAfee_ESM/Integrations/McAfee_ESM_v2/README.md
@@ -1,7 +1,7 @@
 
 Run queries and receive alarms from Intel Security ESM. Support version 10 and above.
 This integration was integrated and tested with version 11.3 of McAfee ESM v2.
-previous versions have been [EOL](https://kc.mcafee.com/corporate/index?page=content&id=KB94822) by the vendor.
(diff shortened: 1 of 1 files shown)


Packs/McAfee_ESM/Integrations/McAfee_ESM_v2/README.md
```diff
@@ -1,6 +1,8 @@
 
 Run queries and receive alarms from Intel Security ESM. Support version 10 and above.
```
Shellyber, line 2:
```suggestion
Run queries and receive alarms from Intel Security ESM.
```

Shellyber (changes requested):
Nice.
We need to change also this part:
image

bakatzir:
> Nice. We need to change also the `version` in this part: image

Are you certain? some customers might still use the EOL versions of ESM, and we stated in the readme the version we tested on. They can put even the EOL version, considering this is their responsibility, no? WDYT?

Shellyber:
@bakatzir not sure we want to encourage/show the option to use the EOL versions. 
But if those versions still works..

Shellyber (approved):

PR #17430, opened 2022-02-02
Panos network operations

adambaumeister:
## Status
- [x] In Progress
- [ ] Ready

## Description
This is a large expansion to the existing Panorama integration to support a series of new commands that will underpin the neew network operations pack. 

The new code uses a different paradigm for connecting to PAN-OS resources by autodiscovering the topology through Panorama and allowing the user to select the devices to run any specific command on using a filter. This means that, for most customers, they will only ever need one instance of the integration running regardless of the size of their topology. 

To support this, instead of wrapping the basic PAN-OS API, we use the public and well maintained pan-os-python SDK which greatly simplifies the logic; https://github.com/PaloAltoNetworks/pan-os-python 

This PR comes without docs updated or a full test suite.

These changes are dependent on https://github.com/demisto/dockerfiles/pull/6906 as the docker image in use has changed to support the additional python package.

## Screenshots

- [X] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No
   - [X] Maybe; None of the existing commands in the pack have changed, but `main()` has and the way commands are routed to their functions has been updated.  

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from dataclasses import dataclass, fields
 from typing import Generator, Callable, Union
 
-import panos.errors
+import panos.errors  # type: ignore
 
 # -- This is a way to get around trimming commonserverpython on import
 try:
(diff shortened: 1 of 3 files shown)


GuyAfik:
Because Pan-OS is such a sensitive pack where a lot of customers use that pack. In order to prevent from any bugs to happen, me and @adambaumeister have decided to split this PR into smaller PRs (6 in total). Right now 36 commands have been implemented so each PR @adambaumeister will implement 6 commands.
That would make the review process much easier and prevent from unnecessary bugs to occur.

Thank you @adambaumeister for this massive effort that you made!

GuyAfik:
duplicate of https://github.com/demisto/content/pull/17689

PR #17506, opened 2022-02-06
Adding credentials set support demisto rest api

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Description
added credentials for the api-key key and relevant tests.

   - [ ] Yes
   - [X] No


Packs/DemistoRESTAPI/Integrations/DemistoRESTAPI/DemistoRESTAPI.yml
```diff
@@ -11,14 +11,20 @@ configuration:
   defaultvalue: https://127.0.0.1
   type: 0
   required: true
-- display: Demisto Server API Key
-  name: apikey
-  type: 4
-  required: true
+- displaypassword: Demisto Server API Key
+  name: api_key
```
ShahafBenYakir, line 15: More meaningful
```suggestion
  name: creds_apikey
```

Packs/DemistoRESTAPI/Integrations/DemistoRESTAPI/DemistoRESTAPI.js
```diff
@@ -15,13 +15,16 @@ sendMultipart = function (uri, entryID, body) {
         // do nothing, use the body as is in the request.
         logDebug('could not parse body as a JSON object, passing as is. body: ' + JSON.stringify(body));
     }
-
+    var key = [params.apikey? params.apikey : (params.api_key? params.api_key.password : '')];
```
ShahafBenYakir, line 18: Why do we do it this way and not like this?
```suggestion
    var key = [params.apikey? params.apikey :  params.api_key.password ];
```
Existing users will have `params.apikey`. New users will have `params.api_key`, such that doing `params.api_key.password` will return the API key or 'undefined', but will not error.

YuvHayun: I tried this, but if no api-key is provided then the user will get the following error:
'TypeError: Cannot read property 'password' of undefined at script'

ShahafBenYakir: How could this be? Doesn't `api_key` always exist? even if  `password`  was not filled?

YuvHayun: It exist, but if password isn't filled then api_key will hold null value and therefore trying to access .password will raise an error.

ShahafBenYakir: Roger

ShahafBenYakir (commented):
I think you are missing this https://github.com/demisto/content/pull/17506#discussion_r802565411

Packs/DemistoRESTAPI/Integrations/DemistoRESTAPI/DemistoRESTAPI.yml
YuvHayun: Done.

ShahafBenYakir (approved):

ShahafBenYakir:
@YuvHayun I already approved this 9 days ago. Fix the build errors and merge.

PR #17525, opened 2022-02-06
Fix bug in copy-from RemoteAccess v2

yaakovpraisler:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/46181)

## Description
Fixed an issue where the `copy-from` command failed to copy PDF files.

## Screenshots
Before the fix:
Screen Shot 2022-02-06 at 18 13 31

After the fix:
Screen Shot 2022-02-06 at 18 13 47

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/RemoteAccess/Integrations/RemoteAccessv2/RemoteAccessv2.py
```diff
@@ -44,7 +44,7 @@ def perform_copy_command(ssh_client: SSHClient, file_path: str, destination_path
                 with tempfile.TemporaryDirectory() as temp_dir:
                     scp_client.get(file_path, f'{temp_dir}/{destination_path}')
                     with open(f'{temp_dir}/{destination_path}', 'rb') as f:
-                        return f.read()
+                        return str(f.read())
```
dantavori, line 47: better to change the return type of the function to `byte`, I'm not sure if it will always behave correctly if we cast it to string.

dantavori (changes requested):

Packs/RemoteAccess/Integrations/RemoteAccessv2/RemoteAccessv2.py
yaakovpraisler: done

dantavori (approved):

PR #17526, opened 2022-02-06
Fixed issue with last run when there are no alerts

ShahafBenYakir:
## Status
- [ ] In Progress
- [ ] Ready
- [x] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/46423

## Description
FireEye API for alerts is inclusive (returns alerts that are on the edges of the search) therefore when we use the creation time of the last fetched alert as the start tie in the next search we will always get this alert. Now, if the system did not receive new events in the next 48 hours (like in the aforementioned bug) the only event that will be returned by the API is the same event we already fetched. Lastly, in our code, in such case we don't do anything special to push the next search time forward (we use the same even creation time), which result in no new incidents.

## Must have
- [ ] Tests
- [ ] Documentation


ShahafBenYakir:
Failing UT
https://app.circleci.com/pipelines/github/demisto/content/144059/workflows/efe1424f-30af-4f42-9c5e-f439b3594ea0/jobs/480450

ShahafBenYakir:
Per this comment from CS about FireEyeCM failing with a similar issue, create a fix for it as well. 
TDD UT: https://app.circleci.com/pipelines/github/demisto/content/144134/workflows/375d0240-a4a1-4a17-9d14-5eeb4681313c/jobs/480643

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM.py
```diff
@@ -344,15 +344,21 @@ def fetch_incidents(client: Client, last_run: dict, first_fetch: str, max_fetch:
...
 
     if not all_alerts:
-        demisto.info(f'{INTEGRATION_NAME} no alerts were fetched at: {str(last_run)}')
-        # as no alerts occurred till now, update last_run time accordingly
-        next_run['time'] = to_fe_datetime_converter('now')
+        demisto.info(f'{INTEGRATION_NAME} no alerts were fetched from FireEye server at: {str(next_run)}')
+        # as no alerts occurred in the window of 48 hours from the given start time, update last_run window to the next
+        # 48 hours. If it is later than now -10 minutes take the latter (to avoid missing events).
+        two_days_from_last_search = (dateparser.parse(next_run['time']) + timedelta(hours=48))
+        now_minus_ten_minutes = dateparser.parse('10 minutes').replace(tzinfo=two_days_from_last_search.tzinfo)
```
Itay4, line 351: i think `replace` here might be problematic, lets discuss

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM_test.py
```diff
@@ -345,15 +345,18 @@ def test_fetch_incidents_last_alert_ids(mocker):
     - mock the Client's get_alerts_request.
     Then
     - Validate that no incidents will be returned.
-    - Validate that the last_run is at the latest incident fetched
+    - Validate that the last_run is pushed in two days from the latest incident fetched
```
Itay4, line 348: do we have test where can check that `now - 10minutes` is taken?

Packs/ApiModules/Scripts/FireEyeApiModule/FireEyeApiModule.py
```diff
@@ -114,10 +114,11 @@ def _get_token(self) -> str:
...
     @logger
     def _generate_token(self) -> str:
-        resp = self._http_request(method='POST', url_suffix='auth/login', resp_type='response')
-        if resp.status_code != 200:
+        try:
+            resp = self._http_request(method='POST', url_suffix='auth/login', resp_type='response')
+        except DemistoException as er:
             raise DemistoException(
-                f'Token request failed with status code {resp.status_code}. message: {str(resp)}')
+                f'Token request failed. message: {str(er)}')
```
Itay4, line 121: cant we just cast the status code to str? anyway i think it would be useful to return it in the error message

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM.py
```diff
@@ -344,15 +344,21 @@ def fetch_incidents(client: Client, last_run: dict, first_fetch: str, max_fetch:
...
+        next_search = min(two_days_from_last_search, now_minus_ten_minutes)
+        next_run = {
+            'time': next_search.isoformat(),
+            'last_alert_ids': []
```
Itay4, line 355: why do we also set `last_alert_ids` to be empty list ?

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM.py
```diff
@@ -344,15 +344,21 @@ def fetch_incidents(client: Client, last_run: dict, first_fetch: str, max_fetch:
...
+        }
+        demisto.info(f'{INTEGRATION_NAME} setting next run to: {str(next_run)}')
         return next_run, []
 
     alerts = all_alerts[:max_fetch]
     last_alert_ids = last_run.get('last_alert_ids', [])
-    if last_alert_ids:
```
Itay4, line 354: why is this not needed?

Packs/ApiModules/Scripts/FireEyeApiModule/FireEyeApiModule.py
ShahafBenYakir: That is exactly the problem, once there's an error there is no resp object (so you can't do resp.status_code). In that case the err contains the code.

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM.py
ShahafBenYakir: because we never clean that list, and in that case it ok to clean it.

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM.py
ShahafBenYakir: Because we were loading and dumping for no reason.

Packs/FireEyeCM/Integrations/FireEyeCM/FireEyeCM_test.py
ShahafBenYakir: Added

Packs/FireEyeCM/ReleaseNotes/1_1_9.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### FireEye Central Management
+- Fixed an issue in the *fetch-incidents* command where alerts that were created 48 hours or more after the last alert, were not fetched.
```
Itay4, line 4:
```suggestion
- Fixed an issue in the fetch incidents process where alerts were not fetched if they that were created 48 hours or later than the last alert that was fetched.
```

Itay4 (approved):

PR #17574, opened 2022-02-09
Syslog remove test module

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46542

## Description
 - Removed the `test_module` command since it's not being used (the `test-module` calls `prepare_globals_and_create_server `)
 - Added support for incident type

- [x] 6.0.0

   - [x] No


Packs/Syslog/Integrations/Syslogv2/Syslogv2.py
```diff
@@ -265,12 +255,14 @@ def prepare_globals_and_create_server(port: int, message_regex: Optional[str], c
         message_regex (Optional[str]): Regex. Will create incident only if Syslog message matches this regex.
         certificate (Optional[str]): Certificate. For SSL connection.
         private_key (Optional[str]): Private key. For SSL connection.
+        incident_type: Optional[str]: The incident Type.
 
     Returns:
         (StreamServer): Server to listen to Syslog messages.
     """
-    global MESSAGE_REGEX
+    global MESSAGE_REGEX, INCIDENT_TYPE
```
Itay4, line 263: no need for incident type to be global, please just use it as local var

Shellyber: This is the original design of the integration (Tom implemented it like that and forgot to add)
Changing it will be more confusing and unnecessary work. I think we should keep it as is

Itay4: its a mistake, you dont reassign the global var anyway so no need to set it as such from the start

you can change that as well while you're here if you wish

Packs/Syslog/ReleaseNotes/2_0_3.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+##### Syslog v2
+- Maintenance and stability enhancements.
+- Fixed an issue were *incident_type* was not passed to **create_incident_from_syslog_message** command.
```
Itay4, line 5:
```suggestion
- Fixed an issue where the incident type parameter was ignored and all incidents were created as *Unclassified*.
```

Itay4 (approved):

PR #17580, opened 2022-02-09
add egg to demisto-sdk requirement

jochman:
[x] - Ready

Egg is critical to venvs like pipenv to determine the dependency name.


yaakovi (approved):
not sure if needed, we are releasing a new version tomorrow anyway

PR #17632, opened 2022-02-13
Pan os change commands names

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/46229

## Related PRs
There is an additional PR that should be merged only **after** this PR is merged as it depends on this PR.

The following PR will change all the ```panorama``` based prefix commands in to ```pan-os``` based prefix commands to playbooks that depend on the ```PAN-OS``` pack (any playbook that uses ```panorama``` prefix commands and has ```PAN-OS``` pack as its dependency).  
https://github.com/demisto/content/pull/17786 

## Description

1) Deprecate all the commands starting with ```panorama``` prefixes
2) Duplicate all the commands that their prefix start with ```panorama``` and change that prefix to ```pan-os```
3) Update the test-playbooks for both panorama/firewall instances to use the new command prefixes (```pan-os```).
4) Update all the playbooks in the ```PAN-OS``` pack to use the new command prefixes (```pan-os```)

## Screenshots

When trying to install a Pack that contains commands from ```PAN-OS``` pack, make sure that when installing a ```PAN-OS``` dependent pack that it will require the user to also install the ```PAN-OS``` pack so it won't break playbooks that rely on the panorama integration commands. 

In this example The ```change management``` pack depends on the ```PAN-OS``` pack because it has a playbook ```playbook-PAN-OS_create_or_edit_policy``` that uses pan-os pack commands. In the below screenshot it is possible to see that when trying to install the ```Change Management``` pack, it will require ```PAN-OS``` pack to get installed as well. 
image

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


GuyAfik:
@bakatzir / @Itay4, would you say in case we deprecate the commands starting with ```panorama``` prefix that i should update the test-playbook as well to use the ```pan-os``` based commands?

In my opinion Test-Playbook should be updated to align with the new command names, but I still wanna get your opinion.

bakatzir:
> @bakatzir / @Itay4, would you say in case we deprecate the commands starting with `panorama` prefix that i should update the test-playbook as well to use the `pan-os` based commands?
> 
> In my opinion Test-Playbook should be updated to align with the new command names, but I still wanna get your opinion.

I concur.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -957,47 +983,14 @@ script:
     - contextPath: URL.Category
       description: The URL category.
       type: String
-  - arguments:
-    - default: true
-      description: URL to check.
-      isArray: true
-      name: url
-      required: true
-    description: Gets a URL category from URL filtering.
```
Itay4: why was this removed?

Packs/PAN-OS/ReleaseNotes/1_9_0.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Palo Alto Networks PAN-OS
+- Deprecated the commands starting with **panorama** prefix.
+- Changed the commands that start with **panorama** prefix to use **pan-os** prefix instead.
```
Itay4, line 5:
```suggestion
- The commands with the **panorama** prefix are deprecated. Use the commands with the **pan-os** prefix instead.
```

Itay4:
@GuyAfik do we have playbooks/scripts that use the commands with the `panorama` prefix? if so, lets discuss

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
GuyAfik: it wasn't removed, just changed its location to align with the new commands since this is a reputation command that will not be deprecated. see here:

image

GuyAfik:
> @GuyAfik do we have playbooks/scripts that use the commands with the `panorama` prefix? if so, lets discuss

Yes, We have playbooks that still use the old panorama commands. We also have a script that uses the `panorama` prefix. Should those be updated as well?

Packs/PAN-OS/ReleaseNotes/1_9_0.md
```diff
@@ -0,0 +1,63 @@
+
+#### Integrations
+##### Palo Alto Networks PAN-OS
+- The commands with the **panorama** prefix are deprecated. Use the commands with the **pan-os** prefix instead.
+
+#### Playbooks
+##### NetOps - Firewall Version and Content Upgrade
+- The commands with the **panorama** prefix are deprecated. The playbook uses the commands with the **pan-os** prefix instead in its tasks.
```
Itay4, line 8: change all playbook release notes to Maintenance and stability enhancements.

Itay4 (approved):

PR #17642, opened 2022-02-13
ParseEmailFilesV2: Added support for the *nesting_level_to_return* argument

moishce:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45133)

## Description
Added feature: parse only nested file.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/ParseEmailFilesV2/ParseEmailFilesV2.py
```diff
@@ -124,6 +136,9 @@ def main():
         if isinstance(output, dict):
             output = [output]
 
+        elif nesting_level_to_return != 'All files':
```
ShahafBenYakir, line 139:
```suggestion
        elif output and nesting_level_to_return != 'All files':
```

Packs/CommonScripts/Scripts/ParseEmailFilesV2/ParseEmailFilesV2_test.py
```diff
@@ -417,3 +419,83 @@ def test_md_output_with_body_text():
...
+    """
+    Given:
+    - parsed email output, nesting_level_to_return param - All files.
+    - parsed email output, nesting_level_to_return param - Outer file.
+    - parsed email output, nesting_level_to_return param - Inner file.
+    When:
+    - Getting all nested emails.
+    - Getting only outer email file.
+    - Getting only inner email file.
+    Then: Validate that returned result as expected.
```
ShahafBenYakir, line 439: This is twisted the `When should be `calling the parse_nesting_level function and the `Then` should be the 3 different scenarios you are asserting.

Packs/CommonScripts/Scripts/ParseEmailFilesV2/ParseEmailFilesV2_test.py
```diff
@@ -417,3 +419,83 @@ def test_md_output_with_body_text():
...
+    """
+    Given:
+    - A eml file contains eml, nesting_level_to_return param - All files.
+    - A eml file contains eml, nesting_level_to_return param - Outer file.
+    - A eml file contains eml, nesting_level_to_return param - Inner file.
+    When:
+    - Getting all nested emails.
+    - Getting only outer email file.
+    - Getting only inner email file.
+    Then: Validate that returned result as expected.
```
ShahafBenYakir, line 457: Also here the given and when are twisted.
Also, the then should be more specific to each scenario for example: `Validating the that length of the response is x and the max size is y.`

Packs/CommonScripts/Scripts/ParseEmailFilesV2/README.md
```diff
@@ -18,6 +18,7 @@ This script is based on the [parse-emails XSOAR python package](https://github.c
 | entryid | Entry ID with the Email as a file in msg or eml format |
 | parse_only_headers | Will parse only the headers and return headers table |
 | max_depth | How many levels deep we should parse the attached emails \(e.g. email contains an emails contains an email\). Default depth level is 3. Minimum level is 1, if set to 1 the script will parse only the first level email |
+| nesting_level_to_return | In case of nested email files (for instance, an EML file inside an EML file), determines which of the email files to return as an output. |
```
ShahafBenYakir, line 21: Don't forget to update docs after fixing the CR

ShahafBenYakir (changes requested):
Need to allign the `max_depth` comment to: 
```In case of nested email files (for instance, an EML file inside an EML file), determines the maximal number of emails attached to parse. The minimum number is 1, if set to 1 the script will parse only the first level email.
```

ShahafBenYakir (approved):

PR #17658, opened 2022-02-14
Dependabot Config

Itay4:
## Status
- [x] Ready

## Description
add dependabot config to set what to run on (and mainly what not to run on - the pipfiles under `Packs`)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "/client"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "npm"
+    directory: "/client"
```
avidan-H, line 9: I'm confused. We don't have a `client` directory in `content`. The `package.json` and `package-lock.json` files sit under the root.

avidan-H (changes requested):
Do you know if the `pip` configuration will detect our requirements files if they aren't using the name `requirements.txt` (which none of ours do)?

.github/dependabot.yml
Itay4: good catch - fixed

avidan-H (approved):

avidan-H:
I must say... excellent commit messages.

PR #17676, opened 2022-02-15
Remove panos dependency

ShahafBenYakir:
Added the skip if unavailable flag to the panorama tasks.


Packs/EDL/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "Generic Export Indicators Service",
     "description": "Ongoing distribution of indicators from XSOAR to other products in the network, using an endpoint with a list of indicators that can be pulled by external vendors.\n",
```
yaakovi, line 3: why do we have `\n` here? 🤔

yaakovi (approved):

Packs/EDL/pack_metadata.json
ShahafBenYakir: Git blame.......

PR #17705, opened 2022-02-16
searchincidentsv2: limit number of events

dorschw:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46740

## Description
added argument to limit the number of event items returned in incidents. default is 0 = all.

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_6_36.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_6_36.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### SearchIncidentsV2
+- Added the `trim_events` argument, to limit the number of events returned from incidents.
--- a/Packs/CommonScripts/ReleaseNotes/1_6_36.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_6_36.md
@@ -1,4 +1,4 @@
 
 #### Scripts
 ##### SearchIncidentsV2
-- Added the `trim_events` argument, to limit the number of events returned from incidents.
+- Added the *trim_events* argument to limit the number of events returned from incidents.
--- a/Packs/CommonScripts/ReleaseNotes/1_6_36.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_6_36.md
@@ -1,4 +1,4 @@
 
 #### Scripts
-##### SearchIncidentsV2
-- Added the *trim_events* argument to limit the number of events returned from incidents.
+##### New: HttpV2
+- Sends a HTTP request with advanced capabilities (Available from Cortex XSOAR 6.5.0).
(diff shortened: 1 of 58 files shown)


Packs/CommonScripts/ReleaseNotes/1_6_36.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### SearchIncidentsV2
+- Added the `trim_events` argument, to limit the number of events returned from incidents.
```
yaakovi, line 4:
```suggestion
- Added the *trim_events* argument to limit the number of events returned from incidents.
```

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2_test.py
```diff
@@ -75,6 +76,7 @@ def test_is_incident_id_valid(id_value, expected_output):
         u'id': u'1',
         u'type': u'TypeA',
         u'name': u'Phishing',
+        u'events': ['event1', 'event2', 'event3']
```
yaakovi, line 79: [PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)
```suggestion
        u'events': ['event1', 'event2', 'event3'],
```

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.yml
```diff
@@ -105,6 +105,15 @@ args:
   name: page
   required: false
   secret: false
+- default: false
+  description: |-
+    The number of events to return from the alert JSON. The default is 0, which returns all events.
+    Note that the trimming is from the top of the list, regardless of event time or other properties.
```
yaakovi, line 111: `top of the list` is ambiguous.
```suggestion
    Note that the trimming is from the end of the list, regardless of event time or other properties.
```

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2_test.py
dorschw: it isn't actually used in the test, I'll remove it instead.

yuvalbenshalom (approved):

BenJoParadise (approved):

BenJoParadise (changes requested):
A few comments:
- the parameter is called `trimevents` not `trimevent` - my bad, i relayed it wrong the first time.
- the parameter should be optional, and should not be passed to getIncident if it was not entered.

BenJoParadise (approved):

PR #17706, opened 2022-02-16
Ansible update

ayman-m:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
I added a new command to query job events by Id.

- [ x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/AnsibleTower/Integrations/AnsibleTower/AnsibleTower.py
+++ b/Packs/AnsibleTower/Integrations/AnsibleTower/AnsibleTower.py
@@ -365,6 +365,30 @@ def list_job_events(client: Client, args: dict) -> CommandResults:
     )
 
 
+def list_job_events_by_id(client: Client, args: dict) -> CommandResults:
+    job_id = args.get('job_id')
+    url_suffix = f'jobs/{job_id}/job_events/'
+
+    response = client.api_request(method='GET', url_suffix=url_suffix, params=args)
+
+    results = response.get('results', [])
+    if not results:
+        return CommandResults(
+            readable_output=f"No results were found for the following arguments {str(args)}",
+            raw_response=response
+        )
+
+    context_data = results_output_data(results)
+    headers = get_headers(context_data)
+    return CommandResults(
+        outputs_prefix='AnsibleAWX.JobEvents',
+        outputs_key_field='id',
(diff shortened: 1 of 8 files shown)


ayman-m:
@bziser with regards to the  "Possible backwards compatibility break" validation error, I did change the command !ansible-tower-job-events-list removing the "id" argument as the id is not a parameter that can be used with the API endpoint "/api/v2/job_events" as per Ansible's documentation and my testing of this command, the command doesn't return any job event data if the "id" argument is.

The id parameter can be used with a different API endpoint "/api/v2/jobs/{job_id}/job_events/" that's why I added an additional command to use the correct API function.

bziser:
> @bziser with regards to the "Possible backwards compatibility break" validation error, I did change the command !ansible-tower-job-events-list removing the "id" argument as the id is not a parameter that can be used with the API endpoint "/api/v2/job_events" as per Ansible's documentation and my testing of this command, the command doesn't return any job event data if the "id" argument is.
> 
> The id parameter can be used with a different API endpoint "/api/v2/jobs/{job_id}/job_events/" that's why I added an additional command to use the correct API function.

As we discussed on slack, the better approach will be to keep the id argument and pass it to the url_suffix.

Packs/AnsibleTower/Integrations/AnsibleTower/AnsibleTower.py
```diff
@@ -365,6 +365,31 @@ def list_job_events(client: Client, args: dict) -> CommandResults:
     )
 
 
+def list_job_events_by_id(client: Client, args: dict) -> CommandResults:
```
bziser, line 368: Please add unit test for the new command.

bziser:
@ayman-m 
To maintain compatibility, I suggest keeping the old command name: `ansible-tower-job-events-list`.
Your new command should be named: `ansible-tower-job-events-list-by-job-id`.
Please update your pull request with these changes.
Thanks for contributing!

bziser (approved):

PR #17790, opened 2022-02-21
Add else option for 'rsa-nw-snapshots-list-for-host' command

ostolero:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes:https://github.com/demisto/etc/issues/45990

## Description
In the rsa-nw-snapshots-list-for-host command, we can't be certain results return.
Added an else option for this situation so the test will pass.


Packs/RSANetWitness_v11_1/TestPlaybooks/playbook-RSANetWitnessv115-Test.yml
```diff
@@ -1,19 +1,19 @@
 id: RSANetWitnessv115-Test
-version: -1
+version: 3
```
ShahafBenYakir, line 2:
```suggestion
version: -1
```

ShahafBenYakir (approved):

PR #17833, opened 2022-02-23
Exchange 2016 Compliance Search - [bug]

omerKarkKatz:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46900

## Description
The client had a problem with the engine got an OSwindows but that was not defined we changed it to an OSerror so it will work on all operation systems. will see the new error after we get a reply from the customer.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


omerKarkKatz:
The issue related has been closed

PR #17851, opened 2022-02-24
Fix linter time out

ilappe:
## Description
It's look like the section `Validate Files and Yaml` in `run-validations` step take a lot of time 
and fail a lot of because of timeout 

the fix reduce time of linters from 4 min to ~1:20

https://panw-global.slack.com/archives/G011E63JXPB/p1645690092271249


yaakovi (approved):
it shouldn't affect the step timeout but the performance improvement is very good.

PR #17998, opened 2022-03-07
Generic Export Indicators Service: added default value to the query

moishce:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/47082)

## Description
When we get the data from the integration_context we use assign_params and if the query is an empty string it was removed and then on the text request the query is None 

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/EDL/Integrations/EDL/EDL.py
```diff
@@ -95,7 +95,7 @@ class RequestArguments:
     FILTER_FIELDS_ON_FORMAT_JSON = "name,type"
 
     def __init__(self,
-                 query: str,
+                 query: str = '',
```
ShahafBenYakir, line 98: Why do we think this is the problem? we are setting by default empty string 
https://github.com/demisto/content/blob/aa9fe55047aca22fddf83a62ab5f13b4ad63d3c1/Packs/EDL/Integrations/EDL/EDL.py#L833

moishce: here is the problem when we are getting the data from the integration_context we use assign_params which removes the query if it is an empty string, and on creating with this data a new RequestArguments object the query is None (because there is not a default value for the query on the init of the class).
https://github.com/demisto/content/blob/aa9fe55047aca22fddf83a62ab5f13b4ad63d3c1/Packs/EDL/Integrations/EDL/EDL.py#L160-L162

ShahafBenYakir: I see thanks

Packs/EDL/ReleaseNotes/3_0_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Generic Export Indicators Service
+- Fixed an issue on run *!export-indicators-list-update* command without *query* argument brakes the nginx server.
```
ShahafBenYakir, line 4:
```suggestion
- Fixed an issue where the ***export-indicators-list-update*** command would fail when given an empty *query* argument.
```

ShahafBenYakir (approved):
Looks good.
see my comment on the RN

PR #18014, opened 2022-03-08
Added parentheses to the fetch incidents filter

MosheEichler:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47018

## Description

1. Added parentheses in the fetch incidents query filter to avoid miss handled the query filter in case the user use `or` operator in the query.
For example the user use this query `riskLevel eq 'medium' or riskLevel eq 'high'` which should be `(riskLevel eq 'medium' or riskLevel eq 'high') and detectedDateTime gt 2022-03-07T01:38:48.604Z` but in fact it will be `riskLevel eq 'medium' or (riskLevel eq 'high' and detectedDateTime gt 2022-03-07T01:38:48.604Z)` because `and` is of higher precedence than `or`.
For more info see: https://docs.microsoft.com/en-us/azure/search/search-query-odata-filter#operator-precedence-in-filters 
2. Save the last_fetch as a string (without converting to datetime object and then return to string) to avoid duplicates while converting (when converting a string to datetime and then return it to string few zeros was added by default in the end that causes duplicates because 00:00:00.123 is bigger then 00:00:00.123000)

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/AzureActiveDirectory/ReleaseNotes/1_2_4.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Azure Active Directory Identity Protection (Beta)
+- Updated the Docker image to: *demisto/crypto:1.0.0.27243*.
+- Fixed an issue with ***fetch-incidents*** which caused incident duplicates.
```
amshamah419, line 5:
```suggestion
- Fixed an issue with the ***fetch-incidents*** command which would cause duplicate incidents.
```

yaakovi (approved):

PR #18016, opened 2022-03-08
Fixed a bug that caused Get User Manager Details to not retrieve the manager details

idovandijk:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47111

## Description
Fixed a bug related to case sensitivity and to inaccurate filters which sometimes caused the playbook to fail retrieving the manager details.
Updated the playbook image.

## Screenshots
Active_Directory_-_Get_User_Manager_Details

- 5.0.0

No

## Must have
- [x] Tests
- [x] Documentation


Packs/Active_Directory_Query/Playbooks/Active_Directory_-_Get_User_Manager_Email_README.md
```diff
@@ -33,4 +33,4 @@ This playbook does not use any sub-playbooks.
```
altmannyarden, line 33: I saw in the playbook yml you used extandcontext to the manager details - we should change the playbook outputs accordingly (I think task 7)

idovandijk: The outputs are set later on in the outputs. The extend-context is only used to differentiate between the key that holds the user details and the key that holds the manager details (we get both of them from AD)

altmannyarden (approved):

idovandijk:
@jrosenpanw can u please review the RN? https://github.com/demisto/content/blob/21aaa0cf86216e807ff456c9882475adc54f1828/Packs/Active_Directory_Query/ReleaseNotes/1_4_7.md

idovandijk:
Related to https://github.com/demisto/etc/issues/47247
@YuvHayun is working on a fix

idovandijk:
Update from call with @esharf:
- Eli will look into the issue with validate which is failing on a supposed dependency on the Core Alert Fields (there is no such dependency here).
- Secrets fails, will need to force merge. Opened an issue: https://github.com/demisto/etc/issues/47326

.circleci/config.yml
```diff
@@ -194,6 +194,7 @@ references:
     run:
       name: Validate Files and Yaml
       when: always
+      no_output_timeout: 1h
```
yaakovi, line 197: This change was merged in a different PR, pull from master

idovandijk: ok

idovandijk:
@anara123 there's a bug with validate, there is no dependency on Core Alert Fields. Need force merge please

idovandijk:
@ShahafBenYakir can u help force merge?

ShahafBenYakir:
Yes

PR #16457, opened 2021-12-17
Avoid unwanted string replacement in SlackV3

daftops:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: link to the issue

## Description
We noticed that the function "handle_tags_in_message_sync" doesn’t properly replace user tags in message we send. 
See below.
An email address including the Slack username will be replaced by the Slack Username instead of keeping the full email adress.

## Screenshots
image

## Minimum version of Cortex XSOAR
- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation

PR #18017, opened 2022-03-08
Remove handle_tags_in_message_sync function

daftops:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
Looking at SlackV2 code the **[handle_tags_in_message_sync()](https://github.com/demisto/content/blob/master/Packs/Slack/Integrations/SlackV3/SlackV3.py#L1891)** function doesn’t exist. However regex used in this function were only applied in async message by the function **[clean_message()](https://github.com/demisto/content/blob/master/Packs/Slack/Integrations/Slack/Slack.py#L388)** in previous Slack integration. Previously I fixed an [issue](https://github.com/demisto/content/pull/16457) in this function because of the user replacement regex was not working properly. Now we noticed that URL formatting is not properly working. Indeed, as per [Slack API documentation](https://api.slack.com/reference/surfaces/formatting#linking-urls), in a markdown message an URL should be formatted as follow:
`<Google|https://google.com>`
Today if you try to send this message it won't applied the link format.

By removing the usage of this function for SlackV3 we expect to retrieve the same behavior as SlackV2>

## Description

## Screenshots

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


yucohen:
@amshamah419 WDYT?

Packs/Slack/ReleaseNotes/2_5_3.md
```diff
@@ -0,0 +1,3 @@
+#### Integrations
```
amshamah419, line 1: Remove

Packs/Slack/ReleaseNotes/2_5_4.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Slack v3
+- Remove handle_tags_in_message_sync() function because of backward compatibility
```
amshamah419, line 4:
```suggestion
- Fixed an issue in the ***send-notification*** command where the formatting for a tagged URL would be removed.
```

amshamah419 (changes requested):
To prevent possible regression, let's also add to the unit test for this command, a case where a message has the URL formatting.

Packs/Slack/ReleaseNotes/2_5_3.md
daftops: @amshamah419 why ? when I run _demisto-sdk validate -g_ or/and the one to create Release Note, it automatically add this one and i f you take a look to all previous Release Notes, they contain this format (see Release Notes 2_5_1, 2_5_0 etc)

yucohen (approved):
@daftops looks great! will handle the RN in the internal

amshamah419 (approved):

PR #18057, opened 2022-03-11
Added a script to join lists by a format.

JasBeilin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/47285)

## Description
Added a script to take 2 list and join them by index according to a given format.

   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/ZipStringsArrays/ZipStringsArrays.yml
```diff
@@ -0,0 +1,35 @@
...
+outputs:
+- contextPath: zipped_list
+  description: A list of joined values by the format given.
+  type: string
+scripttarget: 0
+subtype: python3
+runonce: false
+dockerimage: demisto/python3:3.10.1.26972
+runas: DBotWeakRole
+fromversion: 5.5.0
```
DeanArbel, line 32: This can be 6.0.0.

Packs/CommonScripts/Scripts/ZipStringsArrays/ZipStringsArrays.yml
```diff
@@ -0,0 +1,35 @@
+commonfields:
+  id: ZipStrings
+  version: -1
+name: ZipStrings
+script: ''
+type: python
+tags: []
```
DeanArbel, line 7: Wouldn't it make sense to make this a transformer?

Packs/CommonScripts/Scripts/ZipStringsArrays/ZipStringsArrays.yml
```diff
@@ -0,0 +1,35 @@
...
+- contextPath: zipped_list
+  description: A list of joined values by the format given.
+  type: string
+scripttarget: 0
+subtype: python3
+runonce: false
+dockerimage: demisto/python3:3.10.1.26972
+runas: DBotWeakRole
+fromversion: 5.5.0
+comment: ''
```
DeanArbel, line 33: The script is missing a description.

Packs/CommonScripts/Scripts/ZipStringsArrays/README.md
```diff
@@ -0,0 +1,24 @@
+
```
DeanArbel, line 1: make sure to add a description.

DeanArbel (commented):
See my comments.

Packs/CommonScripts/Scripts/ZipStringsArrays/ZipStringsArrays.yml
JasBeilin: it will, but not efficient. here is an example to make it clear:
list1 = ['a','b']
list2 = ['c','d']
when running as transformer on list1 with list2 as param, the script get the following arg at first run:
{'value':'a', list2:['c','d']}
the second run: 
{'value':'b':list2:['c','d']}

meaning in order to create the correct combination (without dups or missing results) we need to check for each run which index are we on and take the matching from list2.
While this is possible (by getting all the params at once, and taking for each run the index to get the correct value from list2) it is both overkill, and very specific (we need only zip when using script, and a complex logic for a transformer). I believe a script is a much better solution.

DeanArbel (approved):
Good job

Packs/CommonScripts/ReleaseNotes/1_6_43.md
```diff
@@ -0,0 +1,4 @@
+
+#### Scripts
+##### New: ZipStrings
+- You can now join strings from 2 lists. Available from Cortex XSOAR 5.5.0.
```
DeanArbel, line 4:
```suggestion
- You can now join strings from 2 lists. Available from Cortex XSOAR 6.0.0.
```

PR #18063, opened 2022-03-12
Fixed misleading arg description

ali-sawyer:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
Fixed misleading arg description for `microsoft-teams-member-add` arg `user_id`. We were initially thrown off because the arg description says the value needs to be an ID that can only be retrieved by running a different command where the user is already a member of a team. However, we tested passing the user's email address as the `user_id` and this worked. Perhaps this option was added later on, and just the arg description was not updated. Not sure if the same thing applies to other commands in this integration.

## Screenshots
N/A

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


yucohen:
@ali-sawyer Thanks for your contribution!

yucohen (approved):

PR #18101, opened 2022-03-14
fix json table issues

ilappe:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46998

## Description
the PR should fix two issues:

1. using JsonToTable as transformer in playbook failed when the `value` argument are  string from the context, the error is due to the automation get the `value` arg as list with one str element - and after discussion with @aviadl this is related to the transformer api and will not be fixed as it could cause BC

**befor**
image

**after**
image

2. when the `value` arg was array of json objects - run the `JsonToTable` from the war room print entry for each json object instead one table, the fix was to define the `value` arg as array

**before**
image

**after**
image


yaakovi (approved):
please add a sanity unit test

Issue #17421, opened 2022-02-02
[Gmail] Get label IDs


bakatzir:
**Need**
Get Gmail label IDs

**Usecase**
As a user, I want to be able to mark my email with `myLabel`, however, since `myLabel` is a custom label I will need to get its ID (different from its name) in order to be able to use the `gmail-move-mail` command.

**Definition of Done**
Add the `gmail-label-list` command.

Relevant API: https://developers.google.com/gmail/api/reference/rest/v1/users.labels/list

**Getting Started**
For details on how to start, refer to the [Cortex XSOAR Developer Hub](https://xsoar.pan.dev/docs/contributing/contributing#what-to-contribute)

aburt-demisto:
PR <- https://github.com/demisto/content/pull/18163

@bakatzir

PR #18163, opened 2022-03-16
Aburt-gmail-get-label-ids

aburt-demisto:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
https://github.com/demisto/content/issues/17421
https://github.com/demisto/etc/issues/39127

## Description
Adds a new command gmail-list-labels to list labels from a user's mailbox.
This addition requires the update of docker image demisto/google-api so that it includes the "typing" library.

## Screenshots
NA

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0
- [x] 5.0.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Gmail/Integrations/Gmail/Gmail.yml
```diff
@@ -61,6 +73,16 @@ script:
...
+    outputs:
+    - contextPath: GmailLabel.UserID
+      description: The UserID the label belongs to.
+      type: String
+    - contextPath: GmailLabel.Name
+      description: The name of the label.
+      type: String
+    - contextPath: GmailLabel.ID
+      description: The label ID.
+      type: String
```
GuyAfik, line 85: I think the definition of the context output here is under the ```gmail-delete-user``` which is wrong. You should add all of these under line 63.

Packs/Gmail/Integrations/Gmail/Gmail.yml
```diff
@@ -2245,7 +2267,8 @@ script:
       description: A name that appears in the "From:" header for email sent using
         this alias.
       type: String
-  dockerimage: demisto/google-api:1.0.0.24033
+  #dockerimage: demisto/google-api:1.0.0.24033
+  dockerimage: googleapi
```
GuyAfik, line 2271: What is this docker image? is it already in dockerfiles repo? the image should also be tagged as well.

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -929,6 +959,19 @@ def delete_user(user_key):
...
+def list_labels(user_key):
+    command_args = {
+        'userKey': user_key,
+    }
+    service = get_service(
+        'gmail',
+        'v1',
+        ['https://www.googleapis.com/auth/gmail.readonly'])
+    results = service.users().labels().list(userId=user_key).execute()
+    labels = results.get('labels', [])
```
GuyAfik, line 971:
```suggestion
    labels = results.get('labels') or []
```

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -456,6 +457,28 @@ def users_to_entry(title, response, next_page_token=None):
     }
 
 
+def labels_to_entry(title, response, user_key):
```
GuyAfik, line 460: I think this function is redundant, you can do it in the ```list_labels_command```.

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -456,6 +457,28 @@ def users_to_entry(title, response, next_page_token=None):
...
+    human_readable = tableToMarkdown(title, context, headers, removeNull=True)
+
+    return {
+        'ContentsFormat': formats['json'],
+        'Type': entryTypes['note'],
+        'Contents': response,
+        'ReadableContentsFormat': formats['markdown'],
+        'HumanReadable': human_readable,
+        'EntryContext': {'GmailLabel(val.ID == obj.ID && val.Name == obj.Name && val.UserID == obj.UserID)': context}
+    }
```
GuyAfik, line 479: To use here ```CommandResults``` wouldn't work because it is not wrapped in the main in the ```return_results``` function, right?

Packs/Gmail/Integrations/Gmail/Gmail.yml
```diff
@@ -61,6 +73,16 @@ script:
     description: Deletes a Gmail user.
     execution: false
     name: gmail-delete-user
+    outputs:
+    - contextPath: GmailLabel.UserID
```
GuyAfik, line 77: Maybe instead of ```GmailLabel.UserID``` name it ```Gmail.Labels.UserID```? I see this is the convention in this integration.

GuyAfik:
Hi @aburt-demisto, Thank you for your contribution! nice work!  a few points:

1) Can you please add release-note?
2) can you please add the new command to the README?

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: I was just following the convention of the other code. There seems to be a nice formatting function for each call.

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: That's correct. This integration is pretty old and still uses Python2 too. I had to stick with the older method.

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: Is there a reason for this? The get method allows you to specify a default if nothing is returned.

Packs/Gmail/Integrations/Gmail/Gmail.yml
aburt-demisto: Darn it, yes, I copied into the wrong section. I hate YML.

Packs/Gmail/Integrations/Gmail/Gmail.py
GuyAfik: This is incase where ```results.get('labels')``` can return None (if the api returns something like 
```
{
   "labels": null
}
``` 

In that case even if you defined a default value of a list, the labels variable would still get None. 
Many apis behave like that, I am not sure regarding the gmail api though. This is just to be on the safe side here.
In case the api doesn't behave like this you can just keep the way it was before

Packs/Gmail/Integrations/Gmail/Gmail.py
GuyAfik: Its not super important, if you think it would be better to still use that function, i am in.

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: This is what the function handles, i.e. in your example if the labels == None then it would still assign [] to the labels var.

Packs/Gmail/Integrations/Gmail/Gmail.yml
aburt-demisto: I looked through the existing code. The Gmail.Labels shows labels that are a property of a specific e-mail. These are just a list so I didn;t want to confuse the user with something that looked like it may belong to a specific e-mail.

Packs/Gmail/Integrations/Gmail/Gmail.py
GuyAfik: ```results.get('labels', [])``` handles the case where the labels key does appear in the response/dict.

Packs/Gmail/Integrations/Gmail/Gmail.yml
aburt-demisto: This is a custom Docker image. I need to add the 'typing' library to the existing demisto/google-api image.

I create the above image locally to test using `/docker_image_create base=demisto/google-api:1.0.0.24033 name=googleapi dependencies=typing` in XSOAR.

I left this value in as it would need to be change to the new Docker image.

aburt-demisto: Basically this is the only command I have seen in which the Google Client API uses typing. All the other commands don't seem to use it. If I use the standard image `demisto/google-api:1.0.0.24033` then I receive errors when running the `gmail-list-labels`. Error is along the lines of `unknown library import 'typing'`.

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: Ah yes, well spotted. Gmail has at least 1 label as you can't delete the default labels `INBOX` etc...

GuyAfik: Alright then, if so you can keep the current query.

Packs/Gmail/Integrations/Gmail/Gmail.yml
aburt-demisto: PR is in place for this now -> https://github.com/demisto/etc/issues/47444

Will update the Docker image once it becomes available.

Packs/Gmail/Integrations/Gmail/Gmail.py
```diff
@@ -929,6 +962,19 @@ def delete_user(user_key):
     return 'User {} have been deleted.'.format(command_args['userKey'])
 
 
+def list_labels(user_key):
+    command_args = {
+        'userKey': user_key,
+    }
```
GuyAfik, line 968: not sure if i am missing anything but this variable isn't used at all

GuyAfik:
Hey @aburt-demisto, nice work! 
Looks like you have conflicts, make sure to fix the build issues. 
I gave 2 more minor comments.

bakatzir:
@GuyAfik please close all the related issues when the internal PR is merged https://github.com/demisto/etc/issues/39127

Packs/Gmail/Integrations/Gmail/Gmail_test.py
```diff
@@ -128,6 +128,52 @@
 }
 
 
+EXPECTED_LIST_LABELS = {
```
GuyAfik, line 131: read the mock data from a file instead of having such a huge mock response as a variable

Packs/Gmail/Integrations/Gmail/Gmail.py
aburt-demisto: Agreed, removing.

Packs/Gmail/Integrations/Gmail/Gmail_test.py
aburt-demisto: I'm open to suggestions on how this should be done. I am just following the standard as per the other tests.

GuyAfik: not critical, I see in this integration it hasn't been used, so lets give up on that one that.

aburt-demisto: Sure, NP.

GuyAfik (approved):
@aburt-demisto looks good! nice work!

PR #18198, opened 2022-03-20
Get pack version by integration/script/pack name

GuyAfik:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42406

## Description

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -9030,6 +9030,73 @@ def polling_function(name, interval=30, timeout=600, poll_message='Fetching Resu
     return dec
 
 
+def get_pack_version(pack_name=''):
+    """
+    Get a pack version.
+    The version can be retrieved either by a pack name, script name or integration name.
+    To get the version of any pack just send the pack name as mentioned in the pack metadata file as a func argument.
+
+    To get the version of a script/integration just call the function without pack_name.
+    The version of the pack that the script/integration is part of will be returned.
+
+
+    Args:
+        pack_name (str): the pack name as mentioned in the pack metadata file to query its version.
+            use only if querying by a pack name.
+
+    Returns:
+        str: the requested pack version, empty string in-case not found.
+
+    """
+    # query by pack name
+    if pack_name:
+        entity_name = pack_name
+        body_request = {'packsQuery': entity_name}
+        query_type = 'pack'
+    # query by integration name
+    elif demisto.callingContext.get('integration'):  # True means its integration, False means its script/automation.
+        entity_name = demisto.callingContext.get('IntegrationBrand') or ''
+        body_request = {'integrationsQuery': entity_name}
+        query_type = 'integration'
+    # query by script/automation name
+    else:
+        entity_name = (demisto.callingContext.get('context') or {}).get('ScriptName') or ''
+        body_request = {'automationQuery': entity_name}
+        query_type = 'automation'
+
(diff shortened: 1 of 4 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -9030,6 +9030,96 @@ def inner(client, args):
     return dec
 
 
+def get_pack_version(pack_name=''):
+    """
+    Get a pack version.
+    The version can be retrieved either by a pack name, script name or integration name.
```
JudahSchwartz, line 9036: Sounds like from this I can set `pack_name = some_integration_name`. Is that correct? Does it then make sense that  `query_type = 'pack'` for an integration_name?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -9030,6 +9030,96 @@ def inner(client, args):
...
+    To get the version of any pack just send the pack name as mentioned in the pack metadata file as a func argument.
+
+    To get the version of a pack in which a script/integration is part of, just call the function without pack_name.
```
JudahSchwartz, line 9039:
```suggestion
    To get the version of the pack in which the calling script/integration is part of, just call the function without pack_name.
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -9030,6 +9030,96 @@ def inner(client, args):
...
+
+    pack_version = _extract_current_pack_version(
+        _packs=_get_packs_by_query(_body_request=body_request),
+        _query_type=query_type,
+        _entity_name=entity_name
+    )
+    if not pack_version and query_type == 'integration':
+        # handle the case where the display name of the integration is not the same as the integration brand
+        integration_display = _extract_integration_display_name(_integration_brand=entity_name)
+        if integration_display:
```
JudahSchwartz, line 9111: Is it always different in this case? lets add a check that it's not equal entity_name

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: well, I guess it hasn't been explained very well in the doc-string. 
My intention is that only if you want to query by a pack name - then you should pass pack_name as an input for the function, otherwise you can just call the function without this pack_name as input (will get default to empty string)

JudahSchwartz: ok, so lets update the docstring so its more clear, either a pack name, or leave it empty for current pack

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: I am not sure if its always different, I guess its good to add that check as well.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: done, take a look at the current doc-string and let me know if that's understandable.

JudahSchwartz (approved):

PR #18229, opened 2022-03-21
Support XSIAM entities in upload

yaakovpraisler:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47167
fixes: https://github.com/demisto/etc/issues/47624

## Description
Support new XSIAM entities in upload flow.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [x] No


yaakovpraisler:
Successful upload with the new XSIAM entities:
https://code.pan.run/xsoar/content/-/pipelines/2626638

Successful upload XSIAM entities with overriding existing packs (`-o` flag):
https://code.pan.run/xsoar/content/-/pipelines/2626642

Tests/Marketplace/marketplace_services.py
```diff
@@ -1798,6 +1819,49 @@ def collect_content_items(self):
...
+                            'name': content_item.get('name', ''),
+                        })
+
+                    elif current_directory == PackFolders.CORRELATION_RULES.value:
+                        self.is_siem_pack(content_item, 'CorrelationRule')
+                        folder_collected_items.append({
+                            'id': content_item.get('global_rule_id', ''),
+                            'name': content_item.get('name', ''),
+                            'description': content_item.get('description', ''),
+                        })
```
ShahafBenYakir, line 1842: We don't need to call the `is_siem_pack` on these items as we already know at this point (the same way you added the SIEM_RULES_OBJECTS) that these items indicate a SIEM pack. 
This means you can automatically add the tag.

Tests/Marketplace/marketplace_services.py
```diff
@@ -465,6 +466,20 @@ def is_feed_pack(self, yaml_content, yaml_type):
...
+        Args:
+            yaml_content: The yaml content extracted by yaml.safe_load().
+            yaml_type: The type of object to check.
+        """
+        if yaml_type == 'Integration':
+            if yaml_content.get('isFetchEvents', False) is True:
+                self._is_siem = True
+        if yaml_type in SIEM_RULES_OBJECTS:
+            self._is_siem = True
+
```
ShahafBenYakir, line 482: I think we can remove this function and use the existing `is_feed_pack` function. We can rename it to `add_pack_type_tags` and add this:
```
 if yaml_content.get('isFetchEvents', False) is True:
                self._is_siem = True
```
after the current feed condition:
```
if yaml_content.get('script', {}).get('feed', False) is True:
                self._is_feed = True
```

ShahafBenYakir (commented):
Nice

Tests/Marketplace/marketplace_services.py
```diff
@@ -1820,21 +1823,21 @@ def collect_content_items(self):
...
-                        self.is_siem_pack(content_item, 'ModelingRule')
+                        self.add_pack_type_tags(content_item, 'ModelingRule')
                         folder_collected_items.append({
                             'id': content_item.get('id', ''),
                             'name': content_item.get('name', ''),
                         })
 
                     elif current_directory == PackFolders.CORRELATION_RULES.value:
-                        self.is_siem_pack(content_item, 'CorrelationRule')
+                        self.add_pack_type_tags(content_item, 'CorrelationRule')
```
ShahafBenYakir, line 1840: This will obviously fail..... as this yml won't have the `isFetchEvents` field. 
We don't need a function here just do:
```
self._is_siem = True
```

ShahafBenYakir (approved):
After you fix my last comment you can merge

Tests/Marketplace/marketplace_services.py
yaakovpraisler: It should not fail... as the yaml_type is not Integration

PR #18386, opened 2022-03-31
fix readme to include new version

JasBeilin:
## Status
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/46421)

## Description
Updated the version the integration was tested with in readme.


MosheEichler (approved):

bakatzir:
As it works with both 3.4.x and 3.5.x versions tested, I would suggest appending, not replacing.

PR #18448, opened 2022-04-05
added DbotScore calcaulation explenation

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Description
added DbotScore calculation explanation regarding the enhancement that was done a month ago to ***AlienVault_OTX_v2*** integration.

### DIFF
--- a/Packs/AlienVault_OTX/Integrations/AlienVault_OTX_v2/README.md
+++ b/Packs/AlienVault_OTX/Integrations/AlienVault_OTX_v2/README.md
@@ -1029,3 +1029,14 @@ Queries a URL in AlienVault OTX.
 
 ## Additional Information
  - AlienVault considers non lowercased URL protocol as invalid, e.g, HTTP://www.google.com. Hence such submissions will be lowercased to ensure a seamless usage of the integration.
+
+
+## Dbot score calculation method
+First, the DbotScore return *Good* if the false_positive.assessment is “accepted”.
+Else, 
+    if there's no validation in the response then the DbotScore will return score based on the pulase_info length:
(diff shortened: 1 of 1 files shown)


Packs/AlienVault_OTX/Integrations/AlienVault_OTX_v2/README.md
```diff
@@ -1029,3 +1029,14 @@ Queries a URL in AlienVault OTX.
...
+
+## Dbot score calculation method
+First, the DbotScore return *Good* if the false_positive.assessment is “accepted”.
+Else, 
+    if there's no validation in the response then the DbotScore will return score based on the pulase_info length:
+        *Bad* if the length is larger / equal than the default threshold given by the user, 
+        *SUSPICIOUS* if it's between 0 and the default threshold,
+        *None* if the length is zero.
+    Else, if there's one validation, DbotScore will return *SUSPICIOUS*
+    Otherwise, will return Good.
```
ShahafBenYakir, line 1042: Re-ordered the logic here a bit, tell me if it makes sense
```suggestion
In case AlienVault OTX API response contains `accepted` under the `false_positive.assessment` key, the DbotScore will be set to **Good**.
Otherwise, if the response includes one validation, DbotScore will be set to **SUSPICIOUS**, if there's no validation in the response then the DbotScore will be set by the `pulse_info` length in the following manner:
   - **Bad** - If the length of is greater or equal to the default threshold given by the user.
   - **SUSPICIOUS** - If the length is shorter than the default threshold.
    - **None** - If the length is zero.

In any other case, the DbotScore will be set to **Good**.
```

PR #18460, opened 2022-04-05
Return the whole object in mirror in

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47523

## Description
in  mirror in iteration - currently we return only few fields from the updated notable 
this lead to a bug when classifier is based on a field that missed in the mirror in object 
the classifier will fail to classify the mirrored in object as the key are missed

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -44,7 +44,7 @@ MIRROR_DIRECTION = {
     'Incoming And Outgoing': 'Both'
 }
 OUTGOING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency']
-INCOMING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency', 'status_label']
+# INCOMING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency', 'status_label']
 
 # =========== Enrichment Mechanism Globals ===========
 ENABLED_ENRICHMENTS = params.get('enabled_enrichments', [])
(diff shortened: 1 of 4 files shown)


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -44,7 +44,7 @@
     'Incoming And Outgoing': 'Both'
 }
 OUTGOING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency']
-INCOMING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency', 'status_label']
+# INCOMING_MIRRORED_FIELDS = ['comment', 'status', 'owner', 'urgency', 'status_label']
```
esharf, line 47: remove it please

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -1179,11 +1179,9 @@ def get_remote_data_command(service, args, close_incident):
 
     for item in results.ResultsReader(service.jobs.oneshot(search)):
         updated_notable = parse_notable(item, to_dict=True)
-    delta = {field: updated_notable.get(field) for field in INCOMING_MIRRORED_FIELDS if updated_notable.get(field)}
-
-    if delta:
-        demisto.debug('notable {} delta: {}'.format(notable_id, delta))
-        if delta.get('status') == '5' and close_incident:
+    if updated_notable:
```
esharf, line 1182: from the original code, it looks like we don't need to check it.
was made intentionality?

esharf (approved):

yaakovi (approved):

PR #18465, opened 2022-04-06
Update dependencies (support poetrified demistos-dk)

jochman:
I've checked the version of poetry of demisto-sdk and then reverted it.

changed some deps that are not used in the content at all and sent some of the deps loose.

Run installations and tests with the poetry version in: https://github.com/demisto/demisto-sdk/pull/2007

Passed build: https://app.circleci.com/pipelines/github/demisto/content/155193/workflows/675997cd-5556-4bd4-9c8b-5b64ff58c789


ShahafBenYakir (approved):
Approving @tkatzir fyi

PR #18503, opened 2022-04-07
OTRS fix bug in fetch: look bace implementation

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47454

## Description
we store the last creation time of the newest incident in the last run object
and in the next fetch iteration - ask for incidents created after this time
if the incident was not in the queue when we search for the relevant creation time in this queue
and when the incident was moved to the queue - the creation time stored in the last run are newer
we lost this incident 

the solution will be to defined a look back interval (in days, default 1)
and when specific queue is specified - we will search for incidents created also in the looked bace period
and will filter already fetched incidents

## Must have
- [ ] Tests
- [x] Documentation


Packs/OTRS/ReleaseNotes/1_0_10.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### OTRS
+- Fixed an issue where old tickets moved to the fetched queue was missed.
```
yaakovi, line 4:
```suggestion
- Fixed an issue where old tickets moved to the fetched queue were missed.
```

yaakovi (approved):
nice work

PR #18522, opened 2022-04-10
Tweaked the Phishing - Generic v3 test to work better with multiple incidents present in the system

idovandijk:
## Status
Ready

## Description
Slightly improved the test to work better when multiple test incidents are present in the system. The test now searches for the test incident that needs to be investigated, but makes sure to only pick the latest incident to investigate.


idovandijk:
fails because of https://github.com/demisto/etc/issues/47893
which proves the test is good by the way

PR #18524, opened 2022-04-10
Elasticsearch v2 incident timestamp fix

ostolero:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/47854)

## Description
When the index is a date with a decimal ending ( e.g "2022-04-06T10:23:29.571Z"), the date is not parsed properly.

### DIFF
--- a/Packs/Elasticsearch/.pack-ignore
+++ b/Packs/Elasticsearch/.pack-ignore
@@ -1,2 +1,4 @@
 [file:Elasticsearch_v2.yml]
 ignore=IN126,BA108,BA109
+[known_words]
+Elasticsearch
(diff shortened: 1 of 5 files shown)


ostolero:
@moishce unit test failure is in the [first commit](https://github.com/demisto/content/pull/18524/commits/10746860aa87f871b12aaaa5d4b6e78e2aa4c83e)
It's not exactly the same error as the bug, since the bug fails in the incident parsing and can't be tested.
But it still fails, and passes after the fix :)

moishce (approved):

PR #18541, opened 2022-04-11
365 threathunting

tomer-pan:
## Status
- [ ] In Progress
- [x] Ready

## Description
Adding threat hunting playbook for defender 365 + as a sub plyabook in the generic playbook.

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


Packs/Microsoft365Defender/Playbooks/Microsoft_365_Defender_-_Emails_Indicators_Hunt.yml
```diff
@@ -0,0 +1,858 @@
...
+      brand: Microsoft 365 Defender
+    nexttasks:
+      '#none#':
+      - "10"
+    scriptarguments:
+      limit:
+        complex:
+          root: inputs.ResultsLimit
+      query:
+        simple: let _start = now(-${inputs.SearchTimeframe}d); EmailEvents | where
```
David-BMS, line 564: Whats the difference between task 15 and task 18?

tomer-pan: Two different tables with different data. One is to search whether the ip is a part of a url and the other is to check the sender ip.

tomer-pan: Actually I can see now that I need to change the "project away" fields for one of them

Packs/CommonPlaybooks/Playbooks/playbook-Threat_Hunting_-_Generic_6_2.yml
```diff
@@ -0,0 +1,511 @@
...
+      {
+        "position": {
+          "x": -400,
+          "y": 230
+        }
+      }
+    note: false
+    timertriggers: []
+    ignoreworker: false
+    skipunavailable: false
```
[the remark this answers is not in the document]
tomer-pan: Done

tschanfeld:
edited yaml and md files

idovandijk (commented):
Well done! There's not much to change, but it would be great if you can please address these 2 points:
1. Readability is a bit difficult because of the different queries, different results, and the merge of the results. I understand it because you explained it to me, but can you to try to make it clearer in the task names/descriptions what you are doing? Example: 
image

so in the future when we add more to this it would be easier to modify (I checked, and Defender also supports search on endpoints and more).

2. This point is more important for me - if we're looking at the other hunting playbooks they are searching indicators in SIEM systems and then outputting related indicators:
image

they're all taking similar inputs and outputting similar outputs. But in the indicator hunting in emails we are using the same inputs and outputting different results which are about emails, and they contain things like Authentication Details, Email Language, Threat Type, etc.
So what I'm thinking is: if I'm searching for hosts with a malicious SHA256 using Splunk, I might also be getting details about emails. Can we somehow separate this logic / allow the user more customizability here?

tomer-pan:
@idovandijk 
1) The name of the playbook indicates that this is an EMAIL indicator hunting playbook. If we want to add something for endpoints we would have to add a different playbook - I think it will be more object oriented and also, since the columns are different there, it's better to output it with a different playbook. As for the readability, I'm not sure I'm following - maybe a description for the task can help? As I showed you, since we get multiple results and we want to "normalize" them, these steps are needed.
2) Each playbook output to a different key and so is this playbook - I'm not sure it has to be with the same name. If someone will use the generic playbook, they can look at the outputs and decide what to do with the results.

idovandijk (approved):
As discussed just add description to explain the merge of different results. Great job!!

PR #18544, opened 2022-04-11
add content_format to CommandResults

dorschw:


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6198,18 +6215,19 @@ def to_context(self):
         if self.relationships:
             relationships = [relationship.to_entry() for relationship in self.relationships if relationship.to_entry()]
 
-        content_format = EntryFormat.JSON
-        if isinstance(raw_response, STRING_TYPES) or isinstance(raw_response, int):
-            content_format = EntryFormat.TEXT
+        if self.content_format is None:
+            self.content_format = EntryFormat.JSON
```
dorschw, line 6219: Does changing `self.content_format` here make sense? @DeanArbel 
on one hand, future calls will not enter this section and won't conclude the content format
on the other, the content is not _meant_ to be changed between `to_context()` calls.

DeanArbel: @dorschw I agree. Both options are ok, but let's stick with the better practice - work with a `local variable`.

DeanArbel (approved):
Nice

PR #18615, opened 2022-04-17
Prisma Cloud Compute Vulnerability and Compliance Reporting - deprecate

bakatzir:


yaakovi:
what about the scripts?

bakatzir:
> what about the scripts?

The are merely used in the PB, so I'm not seeing the need. The main idea is to remove and add with different IDs/move all of the items to the OOTB pack.

PR #18622, opened 2022-04-17
Panorama spyware dns commnad

omerKarkKatz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/39563)

## Description
Add a new command that will allow assigning of EDL to the Anti-Spyware profile under "DNS Signature Policies" 

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,36 @@ def get_anti_spyware_best_practice_command():
...
+def apply_dns_signature_policy_command(args) -> CommandResults:
+    anti_spy_ware_name = args.get('Anti_spyware_profile_name')
+    edl = args.get('DNS_signature_source')
+    action = args.get('Action')
+    packet_capture = args.get('Packet_capture', 'disable')
+    params = {
+        'action': 'set',
+        'type': 'config',
+        'xpath': f"/config/devices/entry[@name='localhost.localdomain']/device-group/entry[@name='{DEVICE_GROUP}']"
+                 f"/profiles/spyware/entry[@name='{anti_spy_ware_name}']",
```
GuyAfik, line 6227:
```suggestion
        'xpath': f"{XPATH_OBJECTS}/profiles/spyware/entry[@name='{anti_spy_ware_name}']",
```
XPATH_OBJECTs should refer to the correct device group already. 

image

see for example: 
```
@logger
def panorama_delete_custom_url_category(custom_url_category_name: str):
    params = {
        'action': 'delete',
        'type': 'config',
        'xpath': XPATH_OBJECTS + "profiles/custom-url-category/entry[@name='" + custom_url_category_name + "']",
        'element': "<entry name='" + custom_url_category_name + "'></entry>",
        'key': API_KEY
    }
    result = http_request(
        URL,
        'POST',
        body=params,
    )

    return result
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,36 @@ def get_anti_spyware_best_practice_command():
...
+    result = http_request(
+        URL,
+        'POST',
+        params=params,
+    )
+    res_status = result.get('response', {}).get('@status')
+    return CommandResults(outputs=result,
+                          outputs_prefix='Panorama.ApplyDNS',
+                          readable_output=f'**{res_status}**',
+                         )
```
GuyAfik, line 6245:
```suggestion
    return CommandResults(outputs=result,
                          outputs_prefix='PAN-OS.ApplyDNS',
                          readable_output=f'**{res_status}**',
                          raw_response=result
                         )
```

does the api returns just 'success'? is that why there is no table here? what is the string which is being returned for the ```res_status``` variable?
Also are we sure we want to return in the ```outputs``` the actual api result to the context? maybe we do not need it, lets discuss.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,36 @@ def get_anti_spyware_best_practice_command():
     })
 
 
+def apply_dns_signature_policy_command(args) -> CommandResults:
```
GuyAfik, line 6218: please add a doc-string to the function.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -6607,6 +6607,35 @@ script:
     - contextPath: Panorama.Spyware.BotentDomain.Max_version
       description: The botnet domain max version.
       type: String
+  - arguments:
+      - name: Anti_spyware_profile_name
+        description: The name of the anti spyware profile. if the profile exists will operate on it. else, if a new name is given a new Anti-Spyware Profile will be created.
```
GuyAfik, line 6612:
```suggestion
        description: The name of the anti spyware profile. if the profile exists, the command will operate on it, Otherwise, if a new name is given a new Anti-Spyware Profile will be created.
```

GuyAfik (commented):
Hi @omerKarkKatz nice work!

Please also add a unit-test(s) to test the new command.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -6607,6 +6607,35 @@ script:
...
+        name: Packet_capture
+        predefined:
+          - disable
+          - single-packet
+          - extended-capture
+        defaultValue: disable
+    name: pan-os-apply-dns-signature-policy
+    description: Allows assigning of EDL to the Anti-Spyware profile under "DNS Signature Policies"
+    outputs:
+      - contextPath: Panorama.ApplyDNS
```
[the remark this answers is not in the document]
omerKarkKatz: Done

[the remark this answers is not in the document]
omerKarkKatz: image
this is all the response im not sure its necessary to map it to output. maybe only raw response and human readable of the operation status like i did already

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -6607,6 +6607,35 @@ script:
...
+  - arguments:
+      - name: Anti_spyware_profile_name
+        description: The name of the anti spyware profile. if the profile exists will operate on it. else, if a new name is given a new Anti-Spyware Profile will be created.
+        required: true
+      - name: DNS_signature_source
+        description: The EDL name to link to the profile
+        required: true
+      - auto: PREDEFINED
+        description: Action on DNS queris
+        name: Action
```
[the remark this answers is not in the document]
omerKarkKatz: Done

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: if you will look closely the prefix you want to change is not the same as in the XPATH_OBJECTS  'localhost.localdomain' is missing. this prefix is similar to others that i saw dealing with the anti-spyware commands.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: I will add it after the doc-review by Shirly :)

ShirleyDenkberg:
@omerKarkKatz Doesn't the integration README need to be updated with this new command?
@GuyAfik @yaakovi Doc review completed.

omerKarkKatz:
@ShirleyDenkberg the README does need to be added but it is generated from the yml you reviewed so i waited to see if there are any spelling check changes :)

omerKarkKatz:
@GuyAfik i talked with yaakovi  about the unit tests for this command. He does not think that it is necessary. what kind of tests did you have in mind?

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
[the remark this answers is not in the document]
omerKarkKatz: after talking with Shai Yakovi we decided to remove the context output

Packs/PAN-OS/Integrations/Panorama/Panorama.py
GuyAfik: did you try it with my suggestion? check if it works, if it does not keep it the way you did.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -1,6 +1,6 @@
 category: Network Security
 commonfields:
-  id: Panorama
+  id: Panorama - SPY
```
GuyAfik, line 3: revert

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -65,8 +65,8 @@ configuration:
   additionalinfo: Use the "API Key (Recommended)" parameter instead.
 description: Manage Palo Alto Networks Firewall and Panorama. For more information
   see Panorama documentation.
-display: Palo Alto Networks PAN-OS
-name: Panorama
+display: Palo Alto Networks PAN-OS - SPY
+name: Panorama - SPY
```
GuyAfik, line 69: revert

GuyAfik:
> @GuyAfik i talked with yaakovi about the unit tests for this command. He does not think that it is necessary. what kind of tests did you have in mind?

I would really appreciate adding a unit-test here. This is a new command and we want to see it works as expected also in terms of future changes. 

two ideas:
1) test that the command results returns what it should.
2) check that the http request is being sent the way you expect it to be.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: It dosent work your way i tried it. keeping the original code here

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,36 @@ def get_anti_spyware_best_practice_command():
...
+                   f'<action><{action}/></action></entry>'
+                   f'</lists>'
+                   f'</botnet-domains>'
+    }
+    result = http_request(
+        URL,
+        'POST',
+        params=params,
+    )
+    res_status = result.get('response', {}).get('@status')
```
GuyAfik, line 6241: I suggest in case of an error, raise an exception which indicates what the error was. you should have it in the api.
There are many examples in the code. 

```suggestion
    res_status = result.get('response', {}).get('@status')
    if res_status == 'error':
        err_msg = result.get('response', {}).get('msg', {}).get('line'))
        raise DemistoException(f'Error: {err_msg})
```

Add this code section if it doesn't get caught in the http_request function.
if needed try to see what the api returns in case of error to fit the case here better. 
Other than that looks great, nice work!

GuyAfik: did you check it?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,36 @@ def get_anti_spyware_best_practice_command():
     })
 
 
+def apply_dns_signature_policy_command(args: dict) -> CommandResults:
```
GuyAfik, line 6218: add docstring please

GuyAfik (approved):

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: @GuyAfik didnt see this comment i will check it and fixed the other comments if its ok will merge :)

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: Done

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: Done

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -6215,6 +6215,47 @@ def get_anti_spyware_best_practice_command():
...
+        'type': 'config',
+        'xpath': f"/config/devices/entry[@name='localhost.localdomain']/device-group/entry[@name='{DEVICE_GROUP}']"
+                 f"/profiles/spyware/entry[@name='{anti_spy_ware_name}']",
+        'key': API_KEY,
+        'element': '<botnet-domains>'
+                   f'<lists>'
+                   f'<entry name="{edl}"><packet-capture>{packet_capture}</packet-capture>'
+                   f'<action><{action}/></action></entry>'
+                   f'</lists>'
+                   f'</botnet-domains>'
```
yaakovi, line 6241: Why is only the first part not an f-string?

yaakovi (approved):

Packs/PAN-OS/Integrations/Panorama/Panorama.py
omerKarkKatz: Changed

PR #18661, opened 2022-04-19
Aburt-new-incidents-class

aburt-demisto:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
Adds a new class called Incidents that allows for iterating through incidents without having to dela with pagination. The new class overloads the __iter__ dunder with one that yields an incident on each iteration. The pagination of incidens is taken care of in the class.

This will make searching incidents much easier for those that write scripts and have to use demisto.executeCommand with "findIncidents" each time.

## Screenshots
NA

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No (I don't think so)

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/ReleaseNotes/1_30_1.md
+++ b/Packs/Base/ReleaseNotes/1_30_1.md
@@ -1,0 +1,5 @@
+
+#### Scripts
+##### CommonServerPython
+- Added new Incidents class to CommonServerPython
+    - This new class allows for iterating through an incident search, without having to deal with pagination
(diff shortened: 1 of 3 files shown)


yaakovi:
This is a good concept, but the current implementation raises some performance concerns.
We will take it internally and prioritize it.

Closing the PR for now.

yaakovi:
This is a good concept, but the current implementation raises some performance concerns.
We will take it internally and prioritize it.

Closing the PR for now.

aburt-demisto:
Hi @yaakovi , it was just a concept that would allow for easier access to finding incidents without the user having to deal with pagination. I had hoped that the yield command would not cause too much stress on resources as it only returns one incident at a time. It could be changed so that the getIncidents commands uses "size: 1" rather than 5 at a time. This would only keep 1 incident in memory at a time when the class was instantiated.

PR #18712, opened 2022-04-24
CSP: Print fixed traceback in return error

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
instead of printing the traceback in each integration/script (what we have now - see screenshots), the `return_error` will do it.

## Screenshots
will have bad traceback line numbers
image
will have fixed traceback line numbers
image


ShahafBenYakir:
The reason we don't do it is cause to looks very very ugly. I want it only in the logs, not in the war-room.

yaakovi:
> The reason we don't do it is cause to looks very very ugly. I want it only in the logs, not in the war-room.

it will still be only in the logs, I didn't add it to the entry

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6395,7 +6397,7 @@ def return_error(message, error='', outputs=None):
             'Type': entryTypes['error'],
             'ContentsFormat': formats['text'],
             'Contents': message,
-            'EntryContext': outputs
+            'EntryContext': outputs,
```
ShahafBenYakir, line 6400: 🤔 ?

ShahafBenYakir (approved):
👍

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
yaakovi: [PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)

PR #18721, opened 2022-04-25
Splunkpy incident owner bug

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/48121

## Description
fix a bug that given a ```enable user mapping``` which is false (not set in the integration configuration), the fetch incidents still returns incidents which are the owner of 'unassigned' (or any other user) in XSOAR. 

The fix covers that in case the ```enable user mapping``` is not marked, then no owner will be returned to the incident in XSOAR.

## Screenshots

## Must have
- [x] Tests
- [ ] Documentation


GuyAfik:
failed unit-tests - https://github.com/demisto/content/pull/18721
the ```owner``` is part of the incident/notables response (which shouldn't be if user mapping is false)

JasBeilin (approved):
looks good

PR #18784, opened 2022-04-27
EWSMailSender bug fix

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [#48170](https://github.com/demisto/etc/issues/48170)

## Description
I reproduced the bug, I believe it occur because extracting the addresses using `split(",")` may yield an empty string in the case of trailing commas (,).
Therefore my fix is to filter the empty strings out of the list.
I tested my fixes nad they work.

   - [ ] Yes
       - Further details:
   - [X] No


Packs/EWSMailSender/Integrations/EWSMailSender/EWSMailSender.py
```diff
@@ -195,13 +195,17 @@ def collect_manual_attachments(manualAttachObj):
     return attachments
 
 
+def get_none_empty_addresses(addresses_str):
+    return list(filter(None, addresses_str.split(",")))
```
ShahafBenYakir, line 199: Let's add logs printing of the address_str before the return so in case the fix will not solve the issue you have better logs

ShahafBenYakir (changes requested):
We are missing a test that reproduces the issue no?

YuvHayun:
> We are missing a test that reproduces the issue no?

I added a test related to the fix

Packs/EWSMailSender/Integrations/EWSMailSender/EWSMailSender_test.py
```diff
@@ -91,3 +91,24 @@ def test_send_mail(mocker):
     assert result.get('Contents') == {
         'from': 'test@gmail.com', 'to': ['test@gmail.com'], 'subject': 'test', 'attachments': []
     }
+
+
+def test_send_mail_with_trailing_comma(mocker):
+    """
+    Given -
+        to and subject arguments to send an email.
```
ShahafBenYakir, line 99:
```suggestion
        a 'subject' which is 'test' and 'to' which is 'test@gmail.com,' (ending with a comma),
```

ShahafBenYakir (approved):

PR #18788, opened 2022-04-27
Improve xsiam build - Run xsiam flow only when tests were found. 

daryakoval:
Improve xsiam build:
- Run xsiam flow only when tests were found. 
- Run after script only if xsiam machine was locked.


daryakoval:
@yaakovi Implemented your suggestion...

PR #18806, opened 2022-04-28
Upload-flow: support upload images to new marketplace

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46785

## Description
Fix an issue where when uploading an existing pack to a new marketplace, the integration and author images are not uploaded.


solalraveh (approved):

darkushin:
The [following build](https://code.pan.run/xsoar/content/-/pipelines/2855261) shows an example of the fix. In this build, we can see that the packs `Wiz` and `CrowdStrikeFalcon` which were not included in the marketplacev2 before, are being uploaded with their images when changing the pack_metadata.json to include marketplacev2 under the `marketplaces` key.

Noy-Maimon (approved):

PR #18820, opened 2022-04-28
major: SiemAPIModule

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
An API Module that uses pydantic for validations.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/ApiModules/Scripts/SiemApiModule/SiemAPIModule.py
```diff
@@ -0,0 +1,161 @@
...
+        arbitrary_types_allowed = True
+
+    _normalize_headers = validator('headers', pre=True, allow_reuse=True)(
+        load_json
+    )
+
+
+class Credentials(BaseModel):
+    password: str
+    identifier: Optional[str]
```
yaakovi, line 56: readability:
username-password combo vs password-username - second sound weird.
like PB and J, not J and PB.
like Captain and Tennille, not Tennille and Captain
```suggestion
    identifier: Optional[str]
    password: str
```

Packs/ApiModules/Scripts/SiemApiModule/SiemAPIModule.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class IntegrationOptions(BaseModel):
+    """Add here any option you need to add to the logic"""
+
+    proxy: bool = False
+    limit: int = 1000
+
+
+class IntegrationEventsClient(ABC):
```
yaakovi, line 81: why aren't we using the BaseClient? it already has some error handling features that are lacking here

jochman: because it is breaking the requests API and is very hard to use in comparison to just sending a request.
also, we are missing errors because of that.

PR #18866, opened 2022-05-02
Update pan-os sub test playbooks deprecated panorama prefix to use pan-os prefix instead

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
remove panorama prefix deprecated commands from sub test playbooks and replace them with the same commands only to be with pan-os prefix


dantavori (commented):
why wasn't it tested in the build?

GuyAfik:
> why wasn't it tested in the build?

these are sub-playbooks that are being used in the main test-playbooks

GuyAfik:
closing this one, it will be implemented as part of the changes that the panorama/firewall test playbooks will have to adjust to the new panorama/firewall instances

PR #18871, opened 2022-05-02
add `_Test` postfix for name, id

dorschw:
The Test Playbook name&id matched the ones of the integrartion, which may cause issues.


DeanArbel (approved):
Make sure to reflect this is the `conf.json`.

dorschw (commented):
-

dorschw (commented):
-

PR #18876, opened 2022-05-02
Orca_rm_video_readme_add_metadata

bakatzir:
until we will add it back to the section


bakatzir:
@edenorca - Hey, we are moving the video to the metadata file, to present it in a nicer way in the marketplace.

MosheEichler (approved):

PR #18891, opened 2022-05-03
Fixing cache in content repo

ilaner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42551

## Description
* Changed the `prefix` to a fixed value - the cache should not be changed between commits.
* Added `dev-requirements-py2.txt` as a cache key
* store the cache key as a variable in circleci 

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


.gitlab/ci/global.yml
```diff
@@ -3,8 +3,9 @@
     key:
       files:
         - "dev-requirements-py3.txt"
+        - "dev-requirements-py2.txt"
         - "package-lock.json"
-      prefix: $CI_COMMIT_REF_SLUG
+      prefix: demisto-content
```
avidan-H, line 8: IMO we should leave the prefix as it is - AKA caching of the virtual environment would happen at the branch level (if it were to work properly at all). AFAIK the reason why it doesn't work properly is not due to the prefix but rather because our infrastructure is a [Kubernetes GitLab deployment](https://docs.gitlab.com/runner/executors/kubernetes.html) without an [s3 bucket configured for caching](https://docs.gitlab.com/runner/executors/kubernetes.html#:~:text=If%20not%20available%2C%20the%20cached%20data%20is%20downloaded%20from%20the%20configured%20storage%20and%20saved%20into%20the%20cache%20dir%20as%20a%20compressed%20file.%20The%20compressed%20file%20is%20then%20extracted%20into%20the%20build%20folder.) as well as the same cluster being used for both the `content` and `server` projects.

ilaner:
Closing, will further investigate with `enginereeing productivy` to find a way to make the cache work.

PR #18968, opened 2022-05-10
Add docker service to run-commands-against-instance

ilaner:
This step requires lint, therefore we need to configure the docker-in-docker option as we do in unit-tests step.


dorschw (commented):
Please explain the change, either in a code comment or in the PR body.

jochman (approved):

PR #18997, opened 2022-05-12
feat: SiemApiModule set limit to optional

jochman:
## Status
- [ ] In Progress
- [x] Ready

We should not set limit in general fetch-events case. Setting limit to optional and defaulting to None.

### DIFF
--- a/Packs/ApiModules/Scripts/SiemApiModule/SiemApiModule.py
+++ b/Packs/ApiModules/Scripts/SiemApiModule/SiemApiModule.py
@@ -8,7 +8,7 @@ from CommonServerPython import *
 from CommonServerUserPython import *
 
 from enum import Enum
-from pydantic import BaseConfig, BaseModel, AnyUrl, validator  # type: ignore[E0611, E0611, E0611]
+from pydantic import BaseConfig, BaseModel, AnyUrl, validator
 from requests.auth import HTTPBasicAuth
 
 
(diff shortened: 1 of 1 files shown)


Packs/ApiModules/Scripts/SiemApiModule/SiemApiModule.py
```diff
@@ -140,8 +140,12 @@ def run(self):
         stored = []
         for logs in self._iter_events():
             stored.extend(logs)
-            if len(stored) >= self.options.limit:
-                return stored[: self.options.limit]
+            if self.options.limit is not None:
+                demisto.debug(
+                    'limit reached. limit must be presented ONLY in commands and not in fetch-events.'
```
yaakovi, line 145: add the limit value and the number of events before the truncation

Packs/ApiModules/Scripts/SiemApiModule/SiemApiModule.py
```diff
@@ -140,8 +140,12 @@ def run(self):
         stored = []
         for logs in self._iter_events():
             stored.extend(logs)
-            if len(stored) >= self.options.limit:
-                return stored[: self.options.limit]
+            if self.options.limit is not None:
```
yaakovi, line 143:
```suggestion
            if self.options.limit:
```

yaakovi (approved):

PR #19038, opened 2022-05-15
changing pack with min version 6.5

daryakoval:
Testing


daryakoval:
https://code.pan.run/xsoar/content/-/pipelines/2924652 - pr with pack that was 6.5 and installed on 6.1 6.2 servers and failed

PR #19039, opened 2022-05-15
write investigation id to env

BEAdi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Writing investigation ID to env and not using `sys.exit(1)` in the code in order to see error messages.


BEAdi:
image
It is indeed shown when adding an exception to the code (where it will get anyway).

PR #19052, opened 2022-05-16
Missing dependencies fix

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: An issue where the upload flow was re-uploading packs that were not modified due to the 'missing dependencies' mechanism - these packs had misspelled dependencies, causing them to be marked as missing dependencies (and therefore re-uploaded).
Also: reduced 'upload packs with dependencies zip' to upload only modified packs or packs that their dependencies have been modified.


idovandijk:
Does this affect the sdk commands when flags are not added as well? For example if we use `demisto-sdk upload` and specify a specific pack to upload, will it require it to be a modified one?
Can this affect the packs uploaded in the build process and if so was it tested there?

idovandijk (commented):
Does this affect the `demisto-sdk upload` command when no flags are specified?
And does it affect the build process (when packs are uploaded), and if so was it tested in that way as well?

yaakovi (approved):

PR #19072, opened 2022-05-17
Sync marketplace

daryakoval:
Adding sync marketplace command, after changing marketplace url.


Tests/configure_and_test_integration_instances.py
```diff
@@ -798,10 +798,16 @@ def set_marketplace_url(servers, branch_name, ci_build_number):
             cmd = f'gsutil -m cp -r gs://{from_bucket} gs://{to_bucket}/'
             with open(output_file, "w") as outfile:
                 subprocess.run(cmd.split(), stdout=outfile, stderr=outfile)
+            try:
```
ShahafBenYakir, line 801: let's add a docstring here to explain why are we doing that sync

ShahafBenYakir (approved):

PR #19112, opened 2022-05-19
Add MySQL Modeling Rules

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Adds the XSIAM modeling rules for MySQL.

## Minimum version of Cortex XSIAM
- [x] 3.3.0

   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/MySQLEnterprise/ModelingRules/MySQLEnterpriseEventCollector/MySQLEnterpriseEventCollector.xif
```diff
@@ -0,0 +1,28 @@
+[MODEL: dataset=mysql_enterprise_raw, model=Audit, content_id=***, version=***]
+| alter XDM.Audit.operation.type = json_extract_scalar(_raw_log, ""$.class"")
+| filter XDM.Audit.operation.type = ""audit""
+| alter XDM.Database.time = json_extract_scalar(_raw_log, ""$.timestamp""),
```
avidan-H, line 4: I think this is probably an error and it's supposed to be `XDM.Audit.event_timestamp`...

Packs/MySQLEnterprise/ModelingRules/MySQLEnterpriseEventCollector/MySQLEnterpriseEventCollector.yml
```diff
@@ -0,0 +1,4 @@
+name: MySQLEventCollector
+id: MySQLEventCollector
+fromversion: 3.3.0
```
daryakoval, line 3: Lets set this for debugging purpose and check if it causes the problem:
@yaakovi 

```suggestion
fromversion: 6.0.0
```

avidan-H:
@evisochek can you review the updated modeling rules? I changed a lot of the fields to what I believe are the updated versions of the fields in the relevant models (e.g. Auth, Audit, Database) but I still think you should make sure I mapped to the correct things since some of them I was making assumptions. Also, is my [comment here](https://github.com/demisto/content/pull/19112#discussion_r878785936) correct?

Packs/MySQLEnterprise/ModelingRules/MySQLEnterpriseEventCollector/MySQLEnterpriseEventCollector.xif
evisochek: Yes

evisochek (approved):

yaakovi (approved):

avidan-H:
@anara123 @evisochek Since this pack's modeling rules are decoupled from the way the events are collected (i.e. it doesn't include an integration to fetch) do we need to include documentation here on how the event collection is configured or is this something that is known or documented on the side of xsiam platform?

avidan-H:
@evisochek regarding the dataset, when we manually tested together we were testing against the `filestream_filestream_raw` dataset, but you mentioned that the dataset was liable to change to `<vendor>_<product>_raw` which is why here I have the dataset as `mysql_enterprise_raw` in the PR. Do we have information regarding when/if it will change? The dataset name mismatch is problematic right [now](https://jira-hq.paloaltonetworks.local/browse/XDR-56028). Also, since I don't know how the collection of these events logs are configured in the platform, it's unclear to me if the dataset name will always be the same or if it's something that the user will set at configuration time, in which case we would need to document somewhere that they need to set the dataset name to match the one used here in the modeling rules.

evisochek:
> @evisochek regarding the dataset, when we manually tested together we were testing against the `filestream_filestream_raw` dataset, but you mentioned that the dataset was liable to change to `<vendor>_<product>_raw` which is why here I have the dataset as `mysql_enterprise_raw` in the PR. Do we have information regarding when/if it will change? The dataset name mismatch is problematic right [now](https://jira-hq.paloaltonetworks.local/browse/XDR-56028). Also, since I don't know how the collection of these events logs are configured in the platform, it's unclear to me if the dataset name will always be the same or if it's something that the user will set at configuration time, in which case we would need to document somewhere that they need to set the dataset name to match the one used here in the modeling rules.

Please see my reply in : https://jira-hq.paloaltonetworks.local/browse/XDR-52867

avidan-H:
I tested it manually with an instance that had a dataset and the model fields mapped correctly. I was also successfully able to upload the pack to content's test instance 2 so I will merge upon a successful build unless there are any objections.

Packs/MySQLEnterprise/pack_metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "MySQL Enterprise",
+    "description": "MySQL Enterprise Pack",
```
avidan-H, line 3:
```suggestion
    "description": "The modeling rules for MySQL Enterprise event collectors.",
```

PR #19125, opened 2022-05-20
increase sqs message visibility timeout 

jrauen:
## Status
- [ ] In Progress
- [x] Ready

## Description
The current implementation of the SQS fetch incidents has an error that occurs when the script takes longer than the visibility timeout of the receive message call, currently 5 seconds. Since the messages aren't removed from the queue until after they have been sent as incidents to XSOAR they can reappear during successive iterations of the while loop causing duplicate incidents to be created.

The proposed solution is to create the incidents and delete the queue messages during each iteration of the while loop while also increasing the visibility timeout to 30 seconds.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation

### DIFF
--- a/Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
+++ b/Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
@@ -155,19 +155,20 @@ def parse_incident_from_finding(message, parse_body_as_json=False):
 def fetch_incidents(aws_client, aws_queue_url, max_fetch, parse_body_as_json):
     try:
         client = aws_client.aws_session(service='sqs')
-        receipt_handles = []  # type: list
-        incidents = []  # type: list
+        incidents_created = 0  # type: int
         max_number_of_messages = min(max_fetch, 10)
-        while len(incidents) < max_fetch:
+        while incidents_created < max_fetch:
+            receipt_handles = []  # type: list
+            incidents = []  # type: list
             messages = client.receive_message(
                 QueueUrl=aws_queue_url,
                 MaxNumberOfMessages=max_number_of_messages,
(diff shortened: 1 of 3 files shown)


Packs/AWS-SQS/Integrations/AWS-SQS/AWS-SQS.py
```diff
@@ -177,12 +178,13 @@ def fetch_incidents(aws_client, aws_queue_url, max_fetch, parse_body_as_json):
...
-                if len(incidents) == max_fetch:
+                incidents_created += 1
+                if incidents_created == max_fetch:
                     break
 
-        demisto.incidents(incidents)
-        for receipt_handle in receipt_handles:
-            client.delete_message(QueueUrl=aws_queue_url, ReceiptHandle=receipt_handle)
+            demisto.incidents(incidents)
+            for receipt_handle in receipt_handles:
```
ilappe, line 186: @jrauen 
it's available to delete messages during the `VisibilityTimeout` ?

badarsebard: Yes. From the documentation:
```
Immediately after a message is received, it remains in the queue. To prevent other consumers from processing the message again, Amazon SQS sets a visibility timeout, a period of time during which Amazon SQS prevents other consumers from receiving and processing the message. The default visibility timeout for a message is 30 seconds.
```

Source: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

ilappe (approved):

PR #19178, opened 2022-05-24
added relationships prop to CustomIndicator

kgal-pan:
## Status
- [x] In Progress
- [ ] Ready

## Description
This PR adds a `relationships` property to `CustomIndicator` class to align with other  `Common` types (`IP`, `File`, etc).

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [x] Yes
       - Further details: There are Integrations where `CustomIndicator` class is used: (1) inside `CustomIndicatorDemo` and (2) in `ACTIIndicatorQuery`. 

For the former, I'm adding a change to update the `CustomIndicator`.

The latter are aware of this breaking change as this issue was raised because of a need of theirs to create custom indicator and set up relationship simultaneously. So changes shouldn't
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -2837,7 +2837,7 @@ class Common(object):
 
     class CustomIndicator(Indicator):
 
-        def __init__(self, indicator_type, value, dbot_score, data, context_prefix):
+        def __init__(self, indicator_type, value, dbot_score, data, context_prefix, relationships):
             """
             :type indicator_type: ``Str``
             :param indicator_type: The name of the indicator type.
@@ -2854,6 +2854,9 @@ class Common(object):
             :type context_prefix: ``Str``
             :param context_prefix: Will be used as the context path prefix.
 
+            :type relationships: ``list of EntityRelationship``
+            :param relationships: List of relationships of the indicator.
+
             :return: None
             :rtype: ``None``
             """
@@ -2868,6 +2871,7 @@ class Common(object):
                 format(context_prefix=context_prefix)
 
             self.value = value
+            self.relationships = relationships
 
             if not isinstance(dbot_score, Common.DBotScore):
                 raise ValueError('dbot_score must be of type DBotScore')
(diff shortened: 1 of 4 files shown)


Noy-Maimon (dismissed):
One thing - Note that there is no need for 3 seperate Release Notes files. you can concatenate the content of the three in to one, and have just one new version.

@kgal-pan Let's schedule a quick demo for this change

kgal-pan:
> Thank you for your contribution! One thing - Note that there is no need for 3 seperate Release Notes files. you can concatenate the content of the three in to one, and have just one new version.
> 
> @kgal-pan Let's schedule a quick demo for this change

Thanks @Noy-Maimon! Had some issues with the RNs because the Pack was updated in `master` while I was working on it so when I generated them initially the RN was `1_30_8.md` but master was then updated and another pack took over version 1.30.8. In any case, solved this by checking out the RNs from master in [`b0a344b` (#19178)](https://github.com/demisto/content/pull/19178/commits/b0a344bd58947ededd0e122b637077964fad5cee). 

I'm back next week, I'll set up a short demo. @Ni-Knight - Would you like to join the call?

Ni-Knight:
@kgal-pan - if you inv ill be there :)

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10157,7 +10169,9 @@ def create_indicator_result_with_dbotscore_unknown(indicator, indicator_type, re
                                             value=indicator,
                                             dbot_score=dbot_score,
                                             data={},
-                                            context_prefix=context_prefix)
+                                            context_prefix=context_prefix,
+                                            relationships=relationships
```
yaakovi, line 10173: [PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)
```suggestion
                                            relationships=relationships,
```

Packs/Base/ReleaseNotes/1_31_1.md
```diff
@@ -0,0 +1,6 @@
+##### Base
+- Added the `relationships` property to `CustomIndicator`.
+
+#### Scripts
+##### CommonServerPython
+- Added the `relationships` property to `CustomIndicator`.
```
yaakovi, line 6:
```suggestion
- Added the `relationships` property to `CustomIndicator`.

#### Scripts
##### CommonServerPython
- Added the *relationships* property to the **CustomIndicator** class.
```

Noy-Maimon (approved):

PR #19182, opened 2022-05-25
adding section header to .pylintrc

wissamg:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
in the nightly sdk flow we failed to run sdk_pylint_check.sh

### DIFF
--- a/Tests/scripts/sdk_pylint_check.sh
+++ b/Tests/scripts/sdk_pylint_check.sh
@@ -5,10 +5,15 @@ echo Checking for sdk related pylint errors
 pylint_tests_result="$(python3 -m pylint --errors-only ./Tests | grep demisto_sdk)"
 pylint_utils_result="$(python3 -m pylint --errors-only ./Utils | grep demisto_sdk)"
 
-if [ -n "$pylint_tests_result" ] || [ -n "$pylint_utils_result" ]; then
+if [ -n "$pylint_tests_result" ]; then
     echo Pylint result on Test directory: "$pylint_tests_result"
+fi
+if [ -n "$pylint_utils_result" ]; then
     echo Pylint result on Utils directory: "$pylint_utils_result"
     exit 1
+fi
+if [ -n "$pylint_tests_result" ]; then
+    exit 1    
 fi
 
 echo No SDK related pylint errors


Tests/scripts/sdk_pylint_check.sh
```diff
@@ -5,9 +5,9 @@ echo Checking for sdk related pylint errors
 pylint_tests_result="$(python3 -m pylint --errors-only ./Tests | grep demisto_sdk)"
 pylint_utils_result="$(python3 -m pylint --errors-only ./Utils | grep demisto_sdk)"
 
-if [ -n "$pylint_test_result" ] || [ -n "$pylint_utils_result" ]; then
-    echo Pylint result on Test directory: $pylint_test_result
-    echo Pylint result on Utils directory: $pylint_utils_result
+if [ -n "$pylint_tests_result" ] || [ -n "$pylint_utils_result" ]; then
```
tkatzir, line 8: Why not split it to 2 different `if`s?
What if there are errors only in one of them, why print the other?

yaakovi (commented):
I agree with Tal

Tests/scripts/sdk_pylint_check.sh
wissamg: Actually the other should be empty , ":"
But I will  did it in the dummy way.

yaakovi (approved):

Tests/scripts/sdk_pylint_check.sh
```diff
@@ -5,10 +5,15 @@ echo Checking for sdk related pylint errors
...
+if [ -n "$pylint_tests_result" ]; then
     echo Pylint result on Test directory: "$pylint_tests_result"
+fi
+if [ -n "$pylint_utils_result" ]; then
     echo Pylint result on Utils directory: "$pylint_utils_result"
     exit 1
 fi
+if [ -n "$pylint_tests_result" ]; then
+    exit 1    
+fi
```
tkatzir, line 17: Suggest:

```
if [ -n "$pylint_tests_result" ]; then
    echo Pylint result on Test directory: "$pylint_tests_result"
fi
if [ -n "$pylint_utils_result" ]; then
    echo Pylint result on Utils directory: "$pylint_utils_result"
fi
if [ -n "$pylint_tests_result" ] || [ -n "$pylint_utils_result" ]; then
    exit 1    
fi
```

tkatzir: Maybe even better to create (boolean) fields, instead of repeating the checks.

PR #19192, opened 2022-05-25
RHEL oracle xql

tkatzir:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Adds the XSIAM modeling rules for Oracle DB on RHEL.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


gal-forer (changes requested):
We don't use the version or content_id param, and you have an extra quotes in almost all of the xqls for example ""Oracle"" instead of "Oracle" is it on purpose ?

Also you need to try and run the mapping rules on the dataset in the query builder because there seems to be a couple of intricate functions to extract data so we need to make sure they work as intended

tkatzir:
> you have an extra quotes in almost all of the xqls for example ""Oracle"" instead of "Oracle" is it on purpose ?

It's because that's how it got copied rom the CSV.

PR #2113, opened 2018-09-06
Converted IPInfo to Python

amshamah419:
Converted over the existing IPInfo integration from JS to Py
**Includes:**
- Integration "IPInfo_PY"
- Test Playbook "IPInfo PY"

PR #19261, opened 2022-05-29
Switch deprecated commands TestMysqlQuery

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [#2113](https://github.com/demisto/demisto-sdk/pull/2113)

## Description
Switched TestMysqlQuery to use ***sql-command*** instead of ***query*** since it's deprecated.


ShahafBenYakir (commented):
I think this is a case we want to ignore, this is not an OOB content (it is not released to MP)

YuvHayun:
> I think this is a case we want to ignore, this is not an OOB content (it is not released to MP)

It's located under the "scripts" sections in the id_set and is therefore being tested during the validation.
We can ignore by pack and skip all the items coming from this pack (Developer tools)

ShahafBenYakir:
It is located under the test playbooks directory hence not visible to customers

ShahafBenYakir (approved):

PR #19366, opened 2022-06-02
[Fix] splunk-kv-store-collection-search-entry not working with values containing substring "limit"

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-12964

## Description
Fixed an issue where **splunk-kv-store-collection-search-entry** command failed when limit was provided in *query*.


yuvalbenshalom (approved):

DeanArbel:
@ilappe I left out the comment. I feel the code is understandable as is without the docs - one line below you have a call for `**query`, so if it's a str/unicode it should be wrapped by a `dict`, otherwise, it won't work.

PR #19381, opened 2022-06-02
Fix error to get script to work on particular MIME entity file type

ali-sawyer:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://xsoar.ideas.aha.io/ideas/FR-I-2619

## Description
Modify conditional to get script to work for particular file type, `MIME entity, ISO-8859 text, with very long lines, with CRLF line terminators`. The conditional was previously checking for string `MIME entity text, ISO-8859 text`, which should have matched but did not, simply because of the extra word `text`.

I realize this script is deprecated, but it is an easy fix and would be useful for the customer. This script is actually easier to fix than `ParseEmailFilesV2`, which also errors on the same file, because `ParseEmailFilesV2` uses a third-party python library.

The customer who reported the issue is unable to share the file that triggers the error, but reach out to me and I can connect you with them if needed.

@trbharathwaj  @ValerieEsposito FYI

## Screenshots
image

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_6_71.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_6_71.md
@@ -1,0 +1,4 @@
+
+#### Scripts
+##### ParseEmailFiles
+- Fixed an issue where the script errored on particular MIME entity file type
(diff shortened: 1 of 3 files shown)


bakatzir:
@ali-sawyer the py lib is created and maintained by us - https://github.com/demisto/parse-emails
to your consideration if you want to contrib to it as well.

PR #19442, opened 2022-06-07
Add API Metrics Support to UrlScan

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Description
Added Ratelimit, retries on quota errors, and API Metrics reporting

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details: Moved to Python 3. Should not break BC
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


amshamah419:
@yuvalbenshalom Can we merge this? I think it'll require a force merge due to the change from py2 to py3.

FYI @bakatzir

bakatzir:
Are we fixing https://github.com/demisto/content/blob/master/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py#L7075
 in this PR, or in a separate one?

amshamah419:
Part of me wants to say we should do it in this PR since we are force merging anyways, but since we'd be touching base pack, I'd like it to be separate just in case.

Packs/UrlScan/Integrations/UrlScan/UrlScan.py
```diff
@@ -263,13 +306,13 @@ def format_results(client, uuid):
     # Scan Lists sometimes returns empty
     num_of_attempts = 0
     relationships = []
-    response = urlscan_submit_request(client, uuid)
+    response, metric_results, _ = urlscan_submit_request(client, uuid)
```
yuvalbenshalom, line 309: is this being used?

Packs/UrlScan/Integrations/UrlScan/UrlScan.py
```diff
@@ -503,22 +546,42 @@ def get_urlscan_submit_results_polling(client, uuid):
...
+            continue
+        if metrics == ErrorTypes.QUOTA_ERROR:
+            if not is_scheduled_command_retry():
+                execution_metrics.quota_error += 1
+            items_to_schedule.append(url)
+            continue
         uuid = response.get('uuid')
         get_urlscan_submit_results_polling(client, uuid)
+        execution_metrics.success += 1
+    schedule_and_report(command_results=command_results, items_to_schedule=items_to_schedule,
```
yuvalbenshalom, line 574: please run a test(not unit test) where we get an quota error and it is bing handled by this successfully.

yuvalbenshalom (commented):
see my comments.

yuvalbenshalom:
and also, the urlsacan test playbook is not running. we must have it here.

amshamah419:
> and also, the urlsacan test playbook is not running. we must have it here.

It doesn't run since it's a partner pack. Locally running the PB checks out though

Packs/UrlScan/Integrations/UrlScan/UrlScan.py
amshamah419: It gets passed up to the submission function so we know whether or not to schedule the request.

yuvalbenshalom (approved):

PR #19465, opened 2022-06-08
Added cgroups v2 support to DockerHardeningScript

chrisbalmer:
- Added support for cgroups v2
- Fixed support for memory check_type=allocate
- Updated the Docker image to: demisto/python3:3.10.4.29342

## Status
- [ ] In Progress
- [X] Ready

## Description
When using Ubuntu 22.04 the memory check was failing due to Ubuntu changing to cgroups v2. The file path with the max RAM value was changed to a shared path. This update adds support for cgroups v2 without removing support for cgroups v1. 

While adding this support, I noticed the memory_check argument options did not match the code for allocate. I updated the code to use the `allocate` option instead of `allocation`. If a user supplied `allocate` like the argument list options supplied, the cgroup check was actually being used.

- [X] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/DockerHardeningCheck/DockerHardeningCheck.py
```diff
@@ -37,7 +38,7 @@ def check_memory(target_mem: str, check_type: str) -> str:
         str -- error string if failed
     """
     size = mem_size_to_bytes(target_mem)
-    if check_type == "allocation":
+    if check_type == "allocate":
```
JasBeilin, line 41: Can you please refer us to this change documentation?

JasBeilin (changes requested):
 Please check my comment below so we can move forward :)

Packs/CommonScripts/Scripts/DockerHardeningCheck/DockerHardeningCheck.py
chrisbalmer: I believe the original `allocation` was a typo as all of the documentation refers to `allocate`. Links to docs below:

[DockerHardeningCheck Script on XSOAR Dev Site](https://xsoar.pan.dev/docs/reference/scripts/docker-hardening-check#inputs)

> The memory check type to perform: cgroup - check memory cgroup configuration, allocate - try allocating actual memory and verify that the allocation fails. Note the allocate test on some configurations may cause the container to be killed by the linux memory manager and the whole test will then time out.

[Script YML Showing Argument Options](https://github.com/demisto/content/blob/a4b9233fbfda2e5243cdc1844fe9593b11386a22/Packs/CommonScripts/Scripts/DockerHardeningCheck/DockerHardeningCheck.yml#L19)

>   predefined:
>  - cgroup
>  - allocate

[Comments for the function in code](https://github.com/demisto/content/blob/a4b9233fbfda2e5243cdc1844fe9593b11386a22/Packs/CommonScripts/Scripts/DockerHardeningCheck/DockerHardeningCheck.py#L34)

> check_type {str} -- How to check either: cgroup (check configuration of cgroup) or allocate (check actual allocation)

JasBeilin:
Hi @chrisbalmer, please sync your master with demisto/Content master branch so the conflicts will be resolved and we can merge :)

PR #19590, opened 2022-06-15
Added cgroups v2 support to DockerHardeningScript

chrisbalmer:
- Added support for cgroups v2
- Fixed support for memory check_type=allocate
- Updated the Docker image to: demisto/python3:3.10.4.30607

## Status
- [ ] In Progress
- [X] Ready

## Description
When using Ubuntu 22.04 the memory check was failing due to Ubuntu changing to cgroups v2. The file path with the max RAM value was changed to a shared path. This update adds support for cgroups v2 without removing support for cgroups v1.

While adding this support, I noticed the memory_check argument options did not match the code for allocate. I updated the code to use the `allocate` option instead of `allocation`. If a user supplied `allocate` like the argument list options supplied, the cgroup check was actually being used.

- [X] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_7_4.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_7_4.md
@@ -2,5 +2,5 @@
 #### Scripts
 ##### DockerHardeningCheck
 - Added support for cgroups v2
-- Fixed support for memory check_type=allocate
+- Fixed an issue where memory check_type was incorrect.
 - Updated the Docker image to: demisto/python3:3.10.4.30607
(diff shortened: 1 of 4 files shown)


Packs/CommonScripts/ReleaseNotes/1_7_4.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+##### DockerHardeningCheck
+- Added support for cgroups v2
+- Fixed support for memory check_type=allocate
```
JasBeilin, line 5:
```suggestion
- Fixed Fixed an issue where memory check_type was incorrect.
```

JasBeilin (approved):

PR #19601, opened 2022-06-16
AzureAD client credentials fixes

dorschw:
1. raise when device-code methods are used while the client credentials checkbox is ticked
2. fix docs


dorschw:
deprecated and replaced, see pack/integration docs.

PR #19696, opened 2022-06-22
Fix upload not relevant for mp bug

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: recent upload failures 
https://code.pan.run/xsoar/content/-/pipelines/3132323

## Description
Fixes the upload's issue where some packs were marked as irrelevant for marketplace when they actually were.
also fix upload bug for failing pack in the 'pack with it's dependencies' step.

@yaakovpraisler FYI


Tests/Marketplace/marketplace_constants.py
```diff
@@ -270,7 +270,6 @@ class PackIgnored(object):
     PackFolders.INDICATOR_TYPES.value: "IndicatorTypes",
     PackFolders.LISTS.value: "Lists",
     PackFolders.JOBS.value: "Jobs",
-    PackFolders.LAYOUTS.value: "Layouts",
```
Noy-Maimon, line 273: Deleted since it is duplicate

Tests/Marketplace/marketplace_services.py
```diff
@@ -3693,3 +3689,33 @@ def get_last_commit_from_index(service_account):
...
+        pack_folder: containing folder of that item
+        id_set: id set dict
+
+    Returns:
+        id set dict entity if exists, otherwise {}
+    """
+
+    for id_set_entity in id_set[PACK_FOLDERS_TO_ID_SET_KEYS[pack_folder]]:
+
+        if len(entity_path.parts) == 5:  # Content items that have a sub folder (integrations, scripts, etc)
```
yaakovi, line 3709: can't we check `entity_path in list(id_set_entity.values())[0]['file_path']`

yaakovpraisler: The id-set holds only yml/json files for the content entities, so if for example a change was made in an integration python file only, then it will not be calculated because the `entity_path` is the modified python file and it compares with the integration yaml file in the id-set.

yaakovi: add it as a comment

Noy-Maimon: That was the original thing causing this bug - checking if the specific path in the id set was modified or not. this means any other file that is not the yml file that was modified will be missed.

Tests/Marketplace/marketplace_services.py
```diff
@@ -1715,9 +1715,14 @@ def filter_release_notes_by_entities_display_name(self, release_notes, modified_
...
-            # Filters the RN entries by the entity display name
-            display_names = [list(entity.values())[0]['display_name'] for entity in entities_data]
-            filtered_release_notes_entries = self.filter_entries_by_display_name(release_notes, display_names, rn_header)
+            # Filters the RN entries by the entity display name or name
+            display_names_and_names = []
+            for entity in entities_data:
+                # Since some RN's have the name and not the display name, we check both
+                display_names_and_names.extend([list(entity.values())[0]['display_name'],
+                                                list(entity.values())[0]['name']])
+            filtered_release_notes_entries = self.filter_entries_by_display_name(release_notes, display_names_and_names,
```
yaakovpraisler, line 1733: Are you sure it's neccesarry? because the display name in the id-set creating uses the same `get_display_name` function as in release notes creating, and for most entities this function takes the `name` of the entity for the `display_name`.

Tests/Marketplace/marketplace_services.py
```diff
@@ -3711,11 +3728,11 @@ def get_id_set_entity_by_path(entity_path: Path, pack_folder: str, id_set: dict)
                 return id_set_entity
 
         else:  # Other content items
-            if Path(list(id_set_entity.values())[0]['file_path']) == entity_path:
+            if Path(list(id_set_entity.values())[0]['file_path']) == str(entity_path):
```
yaakovi, line 3731: Why do you need the `str` casting?

yaakovi (commented):
revert the pack changes.

Tests/Marketplace/marketplace_services.py
Noy-Maimon: you are right - i had an issue in the testing files

Tests/Marketplace/marketplace_services.py
Noy-Maimon: at this point it is a Path object

Noy-Maimon: removed

yaakovi (approved):

PR #19710, opened 2022-06-23
Panos command update

ayman-m:

## Status
- [x] In Progress
- [ ] Ready

## Description
Added command argument 'update-context' to the pan-os command to give the user an option to store the command reponse in context.

- [x ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -514,24 +514,37 @@ def panorama_command(args: dict):
     """
     Executes a command
     """
-    params = {}
+    params = {'key': API_KEY}
     for arg in args.keys():
         params[arg] = args[arg]
-    params['key'] = API_KEY
-
-    result = http_request(
-        URL,
-        'POST',
-        body=params
-    )
-
-    return_results({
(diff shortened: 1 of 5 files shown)


GuyAfik (commented):
@ayman-m As usual, you have done an incredible work!

Notice that in some cases, because so much data could be returned in the context, it could make XSOAR web-client crush

That's the reason why until now we haven't returned context output for the `pan-os` command

I think we need to think on a way here to return a limit amount of data and not all of it, as if for example we return the entire panorama config that could be a *LOT* of data that could crush the web-client.

GuyAfik:
@ayman-m seems like you can already achieve context output by running for example:

```
!pan-os action=get xpath=/config/devices/entry[@name='localhost.localdomain']/device-group/entry[@name='Test']/address type=config extend-context=PAN-OS=.
``` 

At the moment since the raw responses can be really huge, we don't want to support context output in the near future as it could make the XSOAR web client crush. Hence closing this PR, let me know if you have any questions, or just slack me :)

PR #19745, opened 2022-06-26
Filter packs to install by min version

yaakovpraisler:
## Status
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-3234)

## Description
Fixed an issue where in the build is trying to install packs with higher min version than the server version.

   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


Tests/configure_and_test_integration_instances.py
```diff
@@ -1679,16 +1681,47 @@ def update_integration_lists(new_integrations_names: List[str], packs_not_to_ins
...
         build (Build): The build object.
     Returns:
         (Set[str]): The set of the packs names that supposed to be not installed in the pre-update.
+        (Set[str]): The set of the non hidden packs names.
     """
-    return get_turned_non_hidden_packs(modified_packs_names, build)
+    non_hidden_packs = get_turned_non_hidden_packs(modified_packs_names, build)
+    packs_with_higher_min_version = get_packs_with_higher_min_version(modified_packs_names - non_hidden_packs, build)
+    build.pack_ids_to_install = list(set(build.pack_ids_to_install) - packs_with_higher_min_version)
+    return packs_with_higher_min_version.union(non_hidden_packs), non_hidden_packs
```
dantavori, line 1697: why do we return a tuple and not one set?

Tests/configure_and_test_integration_instances.py
```diff
@@ -1679,16 +1681,47 @@ def update_integration_lists(new_integrations_names: List[str], packs_not_to_ins
...
+
+    Returns:
+        (Set[str]): The set of the packs names that supposed to be not installed because
+                    their min version is greater than the server version.
+    """
+    packs_with_higher_version = set()
+    for pack_name in packs_names:
+
+        pack_metadata = get_json_file(f"{build.content_path}/Packs/{pack_name}/pack_metadata.json")
+        server_min_version = pack_metadata.get(Metadata.SERVER_MIN_VERSION, Metadata.SERVER_DEFAULT_MIN_VERSION)
```
dantavori, line 1716: are you sure there is a default server min version? AFAIK if there is no such, it is calculated as the minimal fromversion of the pack's content items.

Tests/configure_and_test_integration_instances.py
yaakovpraisler: We need not install the non-hidden and higher-version packs and only send only the non-hidden packs to the `get_changed_integrations` method.

Tests/configure_and_test_integration_instances.py
```diff
@@ -1679,16 +1681,48 @@ def update_integration_lists(new_integrations_names: List[str], packs_not_to_ins
...
 
-def get_packs_not_to_install(modified_packs_names: Set[str], build: Build) -> Set[str]:
+def get_packs_not_to_install(modified_packs_names: Set[str], build: Build) -> Tuple[Set[str], Set[str]]:
     """
     Return a set of packs to install only in the post-update.
     Args:
         modified_packs_names (Set[str]): The set of packs to install.
         build (Build): The build object.
     Returns:
         (Set[str]): The set of the packs names that supposed to be not installed in the pre-update.
```
dantavori, line 1691:
```suggestion
        (Set[str]): The set of the pack names that should not be installed.
```

Tests/configure_and_test_integration_instances.py
```diff
@@ -1679,16 +1681,48 @@ def update_integration_lists(new_integrations_names: List[str], packs_not_to_ins
...
+        (Set[str]): The set of the non hidden packs names.
```
dantavori, line 1692:
```suggestion
        (Set[str]): The set of the non-hidden pack names (should be installed only in post update).
```

dantavori (approved):

PR #19881, opened 2022-07-04
skip firewall test playbooks

GuyAfik:
skip firewall test-playbooks due to env issues

will be resolved by [firewall-test-playbooks-failures-issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-3459)


JudahSchwartz (approved):
please add comment with bug that will solve this

PR #19885, opened 2022-07-04
Updated DBotPredictURLPhishing readme with security disclaimer

idovandijk:
## Status
Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3461

## Description
Added a security disclaimer as the DBotPredictURLPhishing script uses Rasterize which has a security recommendation for docker network hardening.

### DIFF
--- a/Packs/PhishingURL/Scripts/DBotPredictURLPhishing/README.md
+++ b/Packs/PhishingURL/Scripts/DBotPredictURLPhishing/README.md
@@ -3,7 +3,7 @@ Predict phishing URLs using a pre-trained model.
 ## Docker Security Recommendations
 ---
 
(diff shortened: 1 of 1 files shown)


Packs/PhishingURL/Scripts/DBotPredictURLPhishing/README.md
```diff
@@ -1,5 +1,10 @@
 Predict phishing URLs using a pre-trained model.
 
+## Docker Security Recommendations
```
glicht, line 3:
```suggestion
## Security Recommendations
```

glicht (changes requested):
@idovandijk would rather we simply point to the rasterize documentation and not duplicate. That way we have a single source of truth we may change over time.

glicht (approved):

PR #19926, opened 2022-07-06
CrowdStrike Api Module - reuse cached token

ilappe:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-14125

## Description
instead to generate new token for each command, cache the created token in the integration context for the next 30 min


yaakovi (commented):
release notes should be added to the relevant packs using the API module, not the API module pack.

yaakovi (approved):

PR #19947, opened 2022-07-08
Fixed an issue where it doesn't return any errors when failed to pars…

spearmin10:
…e a pcap.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
PcapConvert and PcapExtractStreams are transformer that `return_error` return successful results.
They have to raise exceptions instead of `return_error` in transformer.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PcapAnalysis/ReleaseNotes/2_4_4.md
+++ b/Packs/PcapAnalysis/ReleaseNotes/2_4_4.md
@@ -1,0 +1,6 @@
+
+#### Scripts
+- **PcapExtractStreams**
+- Fixed an issue where it doesn't return any errors when failed to parse a pcap.
+- **PcapConvert**
+- Fixed an issue where it doesn't return any errors when failed to parse a pcap.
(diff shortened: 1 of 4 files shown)


spearmin10:
@JasBeilin There were some regressions by someone modified the scripts in the reviewing process. https://github.com/demisto/content/pull/19150

JudahSchwartz (commented):
According to our [code conventions](https://xsoar.pan.dev/docs/integrations/code-conventions), best practice is to catch exceptions in the main function and return errors using the `return_error` function
Why are we removing the try catch block here? Is there a specific flow where the error is not being returned properly?

spearmin10:
@JudahSchwartz The coding conventios that it returns errors with `return_error` is geenrally true, but transformer is an exception. In transfomer, returning an error must be raised by an exception. Because results returned with `return_error` from an transformer are handled as success. That comes from the design of XSOAR.
This is a simple test playbook that branches to an error path from a task which runs PcapConvert fails. The current script (try-except-return_error) runs on the successful path even in causing an error, but the fixed code branches expectedlly to the error path.
[Test_Error_Path.zip](https://github.com/demisto/content/files/9080463/Test_Error_Path.zip)

JudahSchwartz:
Youre right, thanks for the explanation! Merging it in

PR #19977, opened 2022-07-10
Extract indicators playbook bug fix

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-13923)

## Description
Added a check on xlsm files as well in the *extract from word file* task. costumer confirmed that it fixes the issue

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


idovandijk (changes requested):
Hi @yucohen, I reviewed the changes, good job! I still have some comments that require fixing though:
1. I see that you've added the exclusion of the `xlsm` extension from the extraction command, but not from the conditional task that comes before that:
image

This would result in a situation where a file is detected and then not extracted, but we actually don't want to detect those are Word files at all.

2. You've changes the playbook to run in **quiet mode** which would defeat the purpose of the playbook, as auto-extract and auto-enrich are required in the extraction tasks. Please revert the change.
3. Please remove all the irrelevant test playbooks, looks like many were added due to the old SDK bug that added lots of irrelevant test. We should only keep ` Extract Indicators From File - Generic v2 - Test`.
4. Please update the ` Extract Indicators From File - Generic v2 - Test` test playbook to also test with an XLSM file, preferably the same file that the customer is using. In the test I noticed that we *are* able to convert XLSM files to PDF files and then extract indicators from them. So it would be good to have a check which ensures that XLSM files **are** actually parsed but without failing the playbook.

yucohen:
@idovandijk Thanks! I made all the changes
also asked for the file from the CS, waiting to reply

yaakovi (approved):

idovandijk (approved):

PR #20039, opened 2022-07-13
fix docs of VT PB

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: an content-docs failing build as a result of: https://github.com/demisto/content/pull/19970

## Description
Changed to use image names without `(` and `)`.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


efelmandar (commented):
If we change this, I believe the image filename should be changed as well in ../doc_files/ to match

PR #20048, opened 2022-07-13
Phishing att fix 2

tomer-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-14256

## Description
Changed the type of the value "IncidentFiles.MD5" in the task "Check Whether to add attachments" from "As value", which is a misconfiguration, to "From previous tasks".

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


idovandijk (approved):
Nice, this happened because of the XSOAR UI issue that doesn't display "from previous tasks" / "value"

PR #20106, opened 2022-07-19
Add false positive negative command

tyng94:
add two new commands to cortext xsoar
change image


tyng94:
pulling to wrong fork

PR #20107, opened 2022-07-19
update rn for CommonTypes 3.2.40

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
[link](https://panw-global.slack.com/archives/C011J0AP7J8/p1658175578351409)

## Description
Update the release-notes of the `CommonTypes` (3.2.40) so in case as below customers should install the `CommonTypes` anyway
image

### DIFF
--- a/Packs/CommonTypes/ReleaseNotes/3_2_40.md
+++ b/Packs/CommonTypes/ReleaseNotes/3_2_40.md
@@ -5,5 +5,5 @@
 - **Applications** - changed the type from *longText* to *markdown*
 - **Note:** 
   - In order for the changes to take effect, server versions < 6.8 should re-install the *CommonTypes* pack. If any warnings appear, click the *Install anyway* button.
-  - Re-installing the pack would reset existing configurations for those types configured. For example, incident type configured for a certain playbook would no longer be configured to use.
+  - Re-installing the pack would reset existing configurations for those types configured. For example, an incident type configured for a certain playbook would no longer be configured.
   - Any configuration attached to types from *CommonTypes* pack would need to be re-configured.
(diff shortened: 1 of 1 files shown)


JudahSchwartz (approved):

idovandijk (commented):
@GuyAfik It is worth explaining that re-installing the pack would reset existing configurations for those types (am I right?), for example an incident type configured for a certain playbook would no longer be configured to use it. A custom layout for a common incident type would also need to be reconfigured.

Need to check if this is true. If it is, maybe we want to convey that to the users in the RN, so they don't get unexpected results

GuyAfik:
> @GuyAfik It is worth explaining that re-installing the pack would reset existing configurations for those types (am I right?), for example an incident type configured for a certain playbook would no longer be configured to use it. A custom layout for a common incident type would also need to be reconfigured.
> 
> Need to check if this is true. If it is, maybe we want to convey that to the users in the RN, so they don't get unexpected results

Checked it and that is true, ill update the release-notes accordingly.

Packs/CommonTypes/ReleaseNotes/3_2_40.md
```diff
@@ -3,4 +3,7 @@
 - **Groups** - changed the type from *longText* to *markdown*
 - **Short Description** - changed the type from *longText* to *markdown*
 - **Applications** - changed the type from *longText* to *markdown*
-- **Note:** In order for the changes to take effect, server versions < 6.8 should re-install the *CommonTypes* pack.
+- **Note:** 
+  - In order for the changes to take effect, server versions < 6.8 should re-install the *CommonTypes* pack. If any warnings appear, click the *Install anyway* button.
+  - Re-installing the pack would reset existing configurations for those types configured. For example, incident type configured for a certain playbook would no longer be configured to use.
```
JudahSchwartz, line 8:
```suggestion
  - Re-installing the pack would reset existing configurations for those types configured. For example, an incident type configured for a certain playbook would no longer be configured.
```

JudahSchwartz (approved):

PR #20110, opened 2022-07-19
Pushing update to re-add missing integration

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Description
This PR updates the SDK version to include the fix for beta-integration from the SDK.
After it will be merged I will run the upload with overwrite all packs flag.


ShahafBenYakir:
Gitlab fails on unit test and lint coverage percentage. The reason is that the `CrowdStrikeFalconIntel_v2.py` has some addition that increased the coverage but it is only in the master branch. I can pull from master and fix it but it will just stale the fix for customers by coupler of hours more. Hence since there is no real issue here I am force merging.

PR #20112, opened 2022-07-19
Common delete email playbook

ayman-m:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Fixed an issue where the "Only EWS enabled" condition fails to match

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2.yml
+++ b/Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2.yml
@@ -386,29 +386,6 @@ tasks:
                       value:
                         simple: active
             iscontext: true
-      - - operator: isEmpty
-          left:
-            value:
-              complex:
-                root: modules
-                filters:
-                - - operator: isEqualString
-                    left:
-                      value:
-                        simple: modules.brand
-                      iscontext: true
-                    right:
-                      value:
-                        simple: SecurityAndCompliance
-                - - operator: isEqualString
-                    left:
-                      value:
-                        simple: modules.state
-                      iscontext: true
-                    right:
-                      value:
-                        simple: active
-            iscontext: true
     view: |-
       {
         "position": {
--- a/Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2.yml
+++ b/Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2.yml
@@ -386,6 +386,29 @@ tasks:
                       value:
                         simple: active
             iscontext: true
+      - - operator: isEmpty
+          left:
+            value:
+              complex:
+                root: modules
+                filters:
+                  - - operator: isEqualString
+                      left:
+                        value:
+                          simple: modules.brand
+                        iscontext: true
+                      right:
+                        value:
+                          simple: SecurityAndCompliance
+                  - - operator: isEqualString
+                      left:
+                        value:
+                          simple: modules.state
+                        iscontext: true
+                      right:
+                        value:
+                          simple: active
+            iscontext: true
     view: |-
       {
         "position": {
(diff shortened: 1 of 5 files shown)


idovandijk (changes requested):
As always, thanks for the contribution @ayman-m.
The input for the playbook supports "EWS" and not "EWS v2", so you are right in fixing that, but I wonder why you removed the other condition which verifies that SecurityAndCompliance is not active (checking that an active instance does not exist):
image

ayman-m:
> 

Hi @idovandijk , thanks for reviewing my updates on this playbook 

SecurityAndCompliance is the brand name of the "O365 - Security And Compliance - Content Search" integration which is a different integration and the label of the condition is used to check if Only "EWS v2" integration had an active instance.
Checking if "EWS v2" integration has an active instance is already done by the first condition which I kept, adding to that the playbook "Search And Delete Emails - EWS"  that comes after this conditional task if the branch was "Only EWS Enabled" or "Both Enabled, input is EWS" has none of the SecurityAndCompliance commands so the second condition is of no use.
Lastly, if I kept SecurityAndCompliance condition and the customer has no active instance of "O365 - Security And Compliance - Content Search" because their mail service is MS Exchange not O365 , the condition is not gonna match and the search and delete action would be skipped even if the customer has an active EWS instance integrated with their MS Exchange.

That's why I deleted the second condition

idovandijk:
I understand this part:
> SecurityAndCompliance is the brand name of the "O365 - Security And Compliance - Content Search" integration which is a different integration and the label of the condition is used to check if Only "EWS v2" integration had an active instance.
Checking if "EWS v2" integration has an active instance is already done by the first condition which I kept, adding to that the playbook "Search And Delete Emails - EWS" that comes after this conditional task if the branch was "Only EWS Enabled" or "Both Enabled, input is EWS" has none of the SecurityAndCompliance commands so the second condition is of no use.

But I don't understand this part:
> if I kept SecurityAndCompliance condition and the customer has no active instance of "O365 - Security And Compliance - Content Search" because their mail service is MS Exchange not O365 , the condition is not gonna match

If the customer has no active instance of Security And Compliance, and has EWS v2 active, then condition should be matched.
According to my test, the change of EWS v2 -> EWS, without the removal of he SecurityAndCompliance condition - yields the correct result:
image


Note that the condition for SecurityAndCompliance checks that **there is no active instance** of it, hence the "**only EWS**" condition.
Please correct me if I am wrong

ayman-m:
> I understand this part:
> 
> > SecurityAndCompliance is the brand name of the "O365 - Security And Compliance - Content Search" integration which is a different integration and the label of the condition is used to check if Only "EWS v2" integration had an active instance.
> > Checking if "EWS v2" integration has an active instance is already done by the first condition which I kept, adding to that the playbook "Search And Delete Emails - EWS" that comes after this conditional task if the branch was "Only EWS Enabled" or "Both Enabled, input is EWS" has none of the SecurityAndCompliance commands so the second condition is of no use.
> 
> But I don't understand this part:
> 
> > if I kept SecurityAndCompliance condition and the customer has no active instance of "O365 - Security And Compliance - Content Search" because their mail service is MS Exchange not O365 , the condition is not gonna match
> 
> If the customer has no active instance of Security And Compliance, and has EWS v2 active, then condition should be matched. According to my test, the change of EWS v2 -> EWS, without the removal of he SecurityAndCompliance condition - yields the correct result: image
> 
> Note that the condition for SecurityAndCompliance checks that **there is no active instance** of it, hence the "**only EWS**" condition. Please correct me if I am wrong

@idovandijk you are correct , my bad, I thought the second condition was "not empty", reviewing the condition again now, it actualy matches if no SecurityAndCompliance instances enabled, will add it back now

Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2.yml
```diff
@@ -392,22 +392,22 @@ tasks:
               complex:
                 root: modules
                 filters:
-                - - operator: isEqualString
```
idovandijk, line 395: This playbook is not really supported anymore, but if you wish to fix it too, then you should also add the EWS fix

idovandijk (approved):
Good job!

PR #20146, opened 2022-07-20
Add override

ShahafBenYakir:
## Description
Adding the override all pack option to the triggering script.


yaakovi (approved):
please check it doesn't always use override

ShahafBenYakir:
Gitlab fails on sanity test rate limit

PR #20170, opened 2022-07-21
Url Regex bug fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-14372)

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation 

test:
image


Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
```diff
@@ -193,6 +216,7 @@ def format_single_url(non_formatted_url: str) -> List[str]:
     # Common handling for unescape and normalizing
     non_formatted_url = unquote(unescape(non_formatted_url.replace('[.]', '.')))
     formatted_url = replace_protocol(non_formatted_url)
+    formatted_url = remove_special_chars_from_start_and_end_of_url(formatted_url)
     formatted_url = remove_brackets_from_end_of_url(formatted_url)
```
yaakovi, line 220: Do we still need `remove_brackets_from_end_of_url`?

yaakovi (approved):

idovandijk:
Did we test thoroughly? Had many regressions around this

altmannyarden:
@dansterenson Please wait for @Ni-Knight approval as well

Packs/CommonScripts/Scripts/FormatURL/FormatURL_test.py
```diff
@@ -325,6 +325,29 @@ def test_get_redirect_url_from_query_no_url_query_param(self, mocker):
         assert get_redirect_url_from_query(url_, urlparse(url_), 'q') == url_
         assert demisto.error.called
 
+    @pytest.mark.parametrize('url_, expected', [
+        ('[https://urldefense.com/v3/__https://google.com:443/search?66ujQIQ$]',
+         'https://urldefense.com/v3/__https://google.com:443/search?66ujQIQ$'),
+        ('(https://urldefense.us/v3/__https://google.com:443/searchERujngZv9UWf66ujQIQ$)',
+         'https://urldefense.us/v3/__https://google.com:443/searchERujngZv9UWf66ujQIQ$'),
+        ('[someURL].', 'someURL'),
+        ('[another_test)', '[another_test)'),
```
Ni-Knight, line 334: So in this case no bracket will be removed?

Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
dansterenson: I think so, "remove_special_chars_from_start_and_end_of_url" only handles the scenario where there is the same special char at the beginning and end of URL, for example: [url], (url) etc.
I guess we still want to remove single brackets from the end of the URL, no? @Ni-Knight

Packs/CommonScripts/Scripts/FormatURL/FormatURL_test.py
dansterenson: No, unless we do not distinguish between the type of the special character. Maybe let's discuss what we expect to do.

altmannyarden:
FYI,
Dont forget to update the URL extraction doc, if needed. 
https://xsoar.pan.dev/docs/indicators/url-extraction
https://github.com/demisto/content-docs/blob/master/docs/indicators/url-extraction.md

Ni-Knight:
@idovandijk - there are 120 tests for this, including odd edge cases and also additional manual testing.

Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
```diff
@@ -104,16 +105,65 @@ def replace_protocol(url_: str) -> str:
...
+        if delimiters.get(url_[0]) == url_[-1]:
+            url_ = url_[1:-1]
+        else:
+            url_ = url_[1:]
+
+    if is_ipv6(f'[{url_.split("//")[-1]}]'):
+        # In case we have an IPv6 without brackets, it's "invalid" but we will return it
+        return url_
+
+    if url_.count("/") < 3 and not is_ipv6(url_):
```
yaakovi, line 144: You assume the URL contains the protocol prefix (`https://`) - is that valid?

Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
```diff
@@ -104,16 +105,65 @@ def replace_protocol(url_: str) -> str:
...
+
+        while not url_[-1].isalpha() and not url_[-1].isnumeric():
+            url_ = url_[:-1]
+
+        last_part = url_.split('.')[-1]  # Get last part of the url
+
+        if ":" in last_part:
+            # Checks for port in URL to handle it as a special case
+            tld = last_part.split(":", 1)[0]
+            port = last_part.split(":", 1)[1]
```
yaakovi, line 155:
```suggestion
            tld, port = last_part.split(":", 1)
```

Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
```diff
@@ -104,16 +105,65 @@ def replace_protocol(url_: str) -> str:
...
+
+            if not port.isnumeric() or (not tld.isalpha() and not 1 <= int(tld) <= 255):
```
yaakovi, line 157: ?

Ni-Knight: it's an edge-case of an IP as the host or a domain with a port.

Packs/CommonScripts/Scripts/FormatURL/FormatURL.py
Ni-Knight: I don't, its just to remove the ones with more than 3, if inside handle the case of a url without  a protocol.

Ni-Knight: like - `a.b.c:12a` -> `a.b.c`

PR #20173, opened 2022-07-21
RN filtering fixes in upload flow

yaakovpraisler:
## Status
- [x] Ready

## Description
Fixed the following in the upload flow:
1. Filter release notes only for a new version of a pack, and skipping the upload only for packs without relevant release notes (after filtering) and with modified files that are relevant for both MPs.
2. Added filtering by tags for modified rn files. (should not filter by display name because the release notes file is of an old version and probably there are no relevant modified files data to compare with)
3. Fixed an issue where a change that is relevant to one MP was made in a pack (For ex. in modeling rule) and also an old release notes version was updated, then the pack was skipped when uploading to the other MP (In the above ex. to XSOAR bucket) and the old RN was not updated.


yaakovpraisler:
@ShahafBenYakir @Noy-Maimon - Ready for review
Successafull test uploads:
1. [Tests](https://code.pan.run/xsoar/content/-/pipelines/3323299) when modeling rule file name was changed and added new un-related RN. (Expected to upload the modeling rule even the RN is not related)
2. [Tests](https://code.pan.run/xsoar/content/-/pipelines/3328975) when modeling rule file name was changed without new RN. Also tests when old RN file was changed then it should be filtered by tags and updated in the relevant buckets.

Tests/Marketplace/marketplace_services.py
```diff
@@ -1696,7 +1697,10 @@ def filter_changelog_entries(self, changelog_entry: dict, version: str, modified
         filtered_release_notes = self.filter_release_notes_by_entities_display_name(filtered_release_notes_from_tags,
                                                                                     modified_files_data)
 
-        if modified_files_data and not filtered_release_notes:
+        if not filtered_release_notes and self.are_all_changes_relevant_to_more_than_one_marketplace(modified_files_data):
+            # Will return that the pack is not updated only if all the modified files are relevant also to other marketplaces
+            # besides the current one, and if there are no release notes relevant to the current marketplace (which means
+            # that the release notes are relevant to the other marketplace).
```
ShahafBenYakir, line 1703:
```suggestion
            # Will return that the pack should not be uploaded if the two followings are true:
            1. All the detected changes are for items that also belong to another MP and therefore to be sure that the changes are not relevant to this MP so we will need to check the RN to be sure (item #2).
            2. No release notes relevant to the current marketplace (which means that the changes are relevant to the other marketplace).
```

Tests/Marketplace/marketplace_services.py
```diff
@@ -1582,7 +1583,7 @@ def prepare_release_notes(self, index_folder_path, build_number, modified_rn_fil
                                 logging.debug(f"{all_relevant_pr_nums_for_unified=}")
                                 updated_entry = self._get_updated_changelog_entry(
                                     changelog, version, release_notes=modified_release_notes_lines,
-                                    pull_request_numbers=all_relevant_pr_nums_for_unified)
+                                    pull_request_numbers=all_relevant_pr_nums_for_unified, marketplace=marketplace)
```
ShahafBenYakir, line 1586: Can you remind me why we had this change?

ShahafBenYakir (commented):
Looks good from my end.
@Noy-Maimon please have a look as well. (BTW we will probably need similar logic in the refactoring of the new main)

Tests/Marketplace/marketplace_services.py
yaakovpraisler: In case someone is updating an old RN file then we will want to filter the updated RN entry (which should have the marketplace to which the upload is done)

Tests/Marketplace/marketplace_services.py
```diff
@@ -1706,6 +1710,26 @@ def filter_changelog_entries(self, changelog_entry: dict, version: str, modified
...
+        to which the upload is done.
+
+        Args:
+            modified_files_data (dict): The modified files data that are given from id-set.
+
+        Return:
+            (bool) True, if all the files are relevant to more than one marketplace.
+                   False, if there is an item that is relevant only to the current marketplace.
+        """
+        modified_items = [list(item.values())[0] for data in modified_files_data.values() for item in data]
```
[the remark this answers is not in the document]
yaakovpraisler: done

Tests/Marketplace/marketplace_services.py
```diff
@@ -1696,7 +1697,12 @@ def filter_changelog_entries(self, changelog_entry: dict, version: str, modified
         filtered_release_notes = self.filter_release_notes_by_entities_display_name(filtered_release_notes_from_tags,
                                                                                     modified_files_data)
 
-        if modified_files_data and not filtered_release_notes:
+        if not filtered_release_notes and self.are_all_changes_relevant_to_more_than_one_marketplace(modified_files_data):
+            # Will return that the pack should not be uploaded if the two followings are true:
+            # 1. All the detected changes are for items that also belong to another MP and therefore to be sure that the
+            # changes are not relevant to this MP so we will need to check the RN to be sure (item #2).
+            # 2. No release notes relevant to the current marketplace (which means that the changes are relevant to the
+            # other marketplace).
```
Noy-Maimon, line 1705:
```suggestion
            # In case all release notes were filtered out, verify that it also makes sense - by checking that the 
            # modified files are actually relevant for the other marketplace. 
```

Tests/Marketplace/marketplace_services.py
```diff
@@ -1706,6 +1712,29 @@ def filter_changelog_entries(self, changelog_entry: dict, version: str, modified
         logging.debug(f"Finall release notes - \n{changelog_entry[Changelog.RELEASE_NOTES]}")
         return changelog_entry, False
 
+    def are_all_changes_relevant_to_more_than_one_marketplace(self, modified_files_data):
+        """
+        Returns true if all the modified files are also relevant to another marketplace besides the current one
+        to which the upload is done.
+
```
Noy-Maimon, line 1719:
```suggestion
        Returns true if all the modified files are also relevant to another marketplace besides the current one
         this upload is done for.

```

Noy-Maimon (changes requested):
looks good, please see my comments

Tests/Marketplace/marketplace_services.py
```diff
@@ -1096,7 +1097,7 @@ def filter_modified_files_by_id_set(self, id_set: dict):
             if modified_entities:
                 modified_files_data[pack_folder] = modified_entities
 
-        if not self._modified_files or modified_files_data:
+        if not self._modified_files or modified_files_data or modified_rn_files_paths:
```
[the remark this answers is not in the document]
yaakovpraisler: done

Noy-Maimon (approved):

PR #20231, opened 2022-07-26
Mirroring in CS Falcon: dded the mirror mapping

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-14608

## Description
When added the new mapper for version >= 6.5, the mapping for the mirror fields are missed 
this lead to stop the mirror 

- [x] 6.5.0


yaakovi (approved):
Please fix the indentations in the file.

PR #20237, opened 2022-07-26
Added polling to log retrieval and updated PBs

aburt-demisto:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues

[link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-771) (only pan-os-query-logs)

## Description
Adds polling functionality to the PAN-OS-Query-Logs commands

## Screenshots
NA

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


aburt-demisto:
Although I have specified version 6.2.0, this should be backward compatible with 5.0.0.

GuyAfik (changes requested):
@aburt-demisto That is a wonderful work! You are surely on the way to become a content master. I am very impressed with your ability to execute the polling in such a dedicated and fast manner. a few notes:

1) Please add a unit-test to test the polling functionality, you can take a look how the unit-test is implemented for the `pan-os-commit`
2) Please update the README.md with the updated context output / arguments for the `pan-os-query-logs` command.
3) no need to create a new test playbook, you can just edit the existing one.
4) we need to make sure we return the same context output **always**, if there are any addition when `polling=true` make sure to mention it descriptions

GuyAfik:
> Although I have specified version 6.2.0, this should be backward compatible with 5.0.0.

polling is supported only from `6.2.0`, however if `polling=false`, then indeed it should be compatible with 5.0.0

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4777,127 @@ def panorama_query_logs_command(args: dict):
...
     filedigest = args.get('filedigest')
     url = args.get('url')
+    job_id = args.get('job_id')
 
-    if query and (address_src or address_dst or zone_src or zone_dst
-                  or time_generated or action or port_dst or rule or url or filedigest):
-        raise Exception('Use the free query argument or the fixed search parameters arguments to build your query.')
+    if number_of_logs > 5000:   # type: ignore
+        number_of_logs = 5000
+        return_warning('The "number_of_logs" was greater than 5000, so was reduced to 5000')
```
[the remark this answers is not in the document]
aburt-demisto: I can error out. The API doesn't support more than 5000. If you attempt, the API comes back with the error.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4777,127 @@ def panorama_query_logs_command(args: dict):
...
+            outputs_prefix='Panorama.Monitor',
+            outputs_key_field='JobID',
+            outputs=query_logs_output,
+            readable_output=tableToMarkdown('Query Logs:',query_logs_output, ['JobID', 'Status'], removeNull=True),
+            raw_response=result.response.result.job.result
+        )
+
+        poll_result = PollResult(
+            response=command_results,
+            continue_to_poll=argToBoolean(args.get('polling', 'false')),
```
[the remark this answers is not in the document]
aburt-demisto: This would then over-rule to users selection whether to use polling or not. Whether to continue to poll is then decided by the user selecting polling=True.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4777,127 @@ def panorama_query_logs_command(args: dict):
...
+                'polling': argToBoolean(args.get('polling', 'false')),
+                'interval_in_seconds': arg_to_number(args.get('interval_in_seconds', 10)),
+                'timeout': arg_to_number(args.get('timeout', 120))
+            },
+            partial_result=CommandResults(
+                outputs_prefix='Panorama.Monitor',
+                outputs_key_field='JobID',
+                outputs=result.response.result.job.result,
+                readable_output=f"Fetching logs for job ID {result.response.result.job.id}..."
+            )
```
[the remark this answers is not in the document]
aburt-demisto: I thought this would be nice to see, as there maybe be other scripts that are watching for the speicifc keys in the output?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4777,127 @@ def panorama_query_logs_command(args: dict):
...
+            hide_polling_output = True
+        else:
+            continue_to_poll = False
+            hide_polling_output = False
+            query_traffic_logs_output['Status'] = 'Completed'
+            try:
+                pretty_traffic_logs = prettify_traffic_logs(result.response.result.log.log_entries)
+            except:
+                pretty_traffic_logs = []
+            query_traffic_logs_output['Logs'] = pretty_traffic_logs
```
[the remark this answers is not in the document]
aburt-demisto: I will put something in the description in the YML file.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4777,127 @@ def panorama_query_logs_command(args: dict):
...
+            ),
+            continue_to_poll=continue_to_poll,
+            args_for_next_run={
+                'job_id': result.response.result.job.id,
+                'log-type': log_type,
+                'polling': True,
+                'interval_in_seconds': arg_to_number(args.get('interval_in_seconds', 10)),
+                'timeout': arg_to_number(args.get('timeout')),
+                'hide_polling_output': hide_polling_output
+            }
```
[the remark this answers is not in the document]
aburt-demisto: This is not part of the polling for query_logs, it's a separately function.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -188,6 +188,67 @@ class InvalidUrlLengthException(Exception):
...
+                        self.details = dict_safe_get(job_data, ['details'])
+                        self.warnings = dict_safe_get(job_data, ['warnings', 'line'])
+                        
+                    elif isinstance(job_data, str):
+                        self.id = job_data
+
+            class LogData():
+                def __init__(self, log_data: dict = {}):
+                    self.logs = dict_safe_get(log_data, ['logs'])
+                    self.log_entries = dict_safe_get(self.logs, ['entry'])
```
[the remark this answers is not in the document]
aburt-demisto: I like this SimpleNameSpace, but in testing it only maps the first level of object. I would only receive ResponseObject.response for exmaple, not all the relevant subsequent levels.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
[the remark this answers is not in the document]
GuyAfik: hmm, eventually when polling is finished, the context output is different in some way? i'd appreciate if you can **slack** me up an example of the two api responses.
1. the panorama query logs response
2. the panorama get traffic logs response

I would like to see the difference.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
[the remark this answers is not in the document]
GuyAfik: I am surprised, this should work on nested objects as well, let me check it myself and get back to you

Packs/PAN-OS/Integrations/Panorama/Panorama.py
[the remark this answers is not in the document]
GuyAfik: that's not true, what will over-rule to users whether to use polling is the `polling` argument when passing it in the `args_for_next_run` which is something you have done. 

see my previous comment

Packs/PAN-OS/Integrations/Panorama/Panorama.py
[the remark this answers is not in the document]
GuyAfik: what do you mean?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -11159,26 +11312,26 @@ def main():
...
-                                              additional_malicious=additional_malicious)
+                                                additional_suspicious=additional_suspicious,
+                                                additional_malicious=additional_malicious)
 
         elif command == 'panorama-get-url-category-from-host' or command == 'pan-os-get-url-category-from-host':
             panorama_get_url_category_command(url_cmd='url-info-host', url=args.get('url'),
-                                              additional_suspicious=additional_suspicious,
-                                              additional_malicious=additional_malicious)
+                                                additional_suspicious=additional_suspicious,
+                                                additional_malicious=additional_malicious)
```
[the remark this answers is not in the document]
aburt-demisto: I am not sure what these were?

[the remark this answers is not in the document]
GuyAfik: seems like by mistake those lines were moved

Packs/PAN-OS/Integrations/Panorama/Panorama.py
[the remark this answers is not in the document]
aburt-demisto: I'll remove the partial output for the output. It's not really needed as nothing should be done with it as it would be in a polling state.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -5545,6 +5564,166 @@ script:
     - contextPath: Panorama.Monitor.Message
       description: The message of the logs query.
       type: String
+    - contextPath: Panorama.Monitor.LogType
+      description: The log type of the query.
+      type: String
+    - contextPath: Panorama.Monitor.Logs
+      description: The logs are only returned when polling is enabled and logs are retrieved successfully.
+      type: Unknown
```
GuyAfik, line 5572:
```suggestion
```

not needed

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4689,16 +4696,22 @@ def panorama_query_logs(log_type: str, number_of_logs: str, query: str, address_
         'GET',
         params=params,
     )
+    parsed_namespace = parse_response(result)
```
GuyAfik, line 4699: what do you thing about creating adding an argument to the `http_request` lets say: `parse_as_class`, give it a default false and if its `True` you can return the `SimpleNameSpace` or the `ResponseClass`

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4724,111 @@ def panorama_query_logs_command(args: dict):
...
+                raw_response=res
+            ),
+            continue_to_poll=parsed.response.result.job.status != 'FIN',
+            args_for_next_run={
+                'query_log_job_id': parsed.response.result.job.id,
+                'log-type': log_type,
+                'polling': True,
+                'interval_in_seconds': arg_to_number(args.get('interval_in_seconds', 10)),
+                'timeout': arg_to_number(args.get('timeout'))
+            }
```
GuyAfik, line 4828:
```suggestion
```

This is not needed, see my previous comment

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4711,40 +4724,111 @@ def panorama_query_logs_command(args: dict):
...
+        
+        if not parsed.response.result.job.id:
+            raise Exception('Missing JobID status in response.')
+
+        query_traffic_logs_output = {
+            'JobID': job_id,
+            'Status': 'Pending',
+            'LogType': log_type
+        }
+        if parsed.response.result.job.status == 'FIN':
```
GuyAfik, line 4798:
```suggestion
        if parsed.response.result.job.status.upper() == 'FIN':
```

to be on the safe side

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -5545,6 +5564,166 @@ script:
...
+      type: String
+    - contextPath: Panorama.Monitor.Logs.DeviceName
+      description: The hostname of the firewall on which the session was logged.
+      type: String
+    - contextPath: Panorama.Monitor.Logs.DestinationAddress
+      description: The original session destination IP address.
+      type: String
+    - contextPath: Panorama.Monitor.Logs.DestinationUser
+      description: The username of the user to which the session was destined.
+      type: String
```
GuyAfik, line 5590: this context output is returned only if `polling=true`?

GuyAfik (changes requested):

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aburt-demisto: That could work. Let me look into it.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
aburt-demisto: That's correct. I am not sure where to indicate this though.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -188,6 +189,68 @@ class InvalidUrlLengthException(Exception):
...
+            super().__init__(**{k: v for k, v in kwargs.items() if k not in ignored_keys})
+
+        def __getattr__(self, attr):
+            """
+            If an AttributeError is raised, this method is called, if the attr was not found, Returns None.
+            """
+            if attr == 'entry':
+                return []
+            else:
+                return None
```
GuyAfik, line 211:
```suggestion
            return [] if attr == 'entry else None
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4445,7 +4508,7 @@ def panorama_get_traffic_logs(job_id: str):
     result = http_request(
         URL,
         'GET',
-        params=params,
+        params=params
```
GuyAfik, line 4511: please revert

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -4687,18 +4750,23 @@ def panorama_query_logs(log_type: str, number_of_logs: str, query: str, address_
...
```
GuyAfik, line 4753: please revert

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -11159,26 +11301,26 @@ def main():
...
-                                              additional_malicious=additional_malicious)
+                                                additional_suspicious=additional_suspicious,
+                                                additional_malicious=additional_malicious)
 
         elif command == 'panorama-get-url-category-from-host' or command == 'pan-os-get-url-category-from-host':
             panorama_get_url_category_command(url_cmd='url-info-host', url=args.get('url'),
-                                              additional_suspicious=additional_suspicious,
-                                              additional_malicious=additional_malicious)
+                                                additional_suspicious=additional_suspicious,
+                                                additional_malicious=additional_malicious)
```
GuyAfik, line 11323: please revert

GuyAfik (changes requested):
@aburt-demisto looks great! Please add the context output / command arguments to the README.md for the `pan-os-query-logs` command.

the arguments that are deprecated for polling purposes should not be added, look at `pan-os commit` README for reference

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aburt-demisto: I can't see what was changed? According to compare with master branch, they are the same.

GuyAfik: ill check it for you

idovandijk (commented):
Thanks for the contribution Adam.
Can you please elaborate on the backward compatibility of this in 5.0? In the playbook polling is set to True. Do you add this parameter only in 6.2 then?
And how will the playbook manage without polling?
Also since the test playbook for this playbook is in skipped state in our conf, I would have to see execution working correctly on an older XSOAR (5.0)

aburt-demisto:
@idovandijk 
There should be a V2 of the PBs that support poilling by default. This seemed logical to have polling enabled by default in them. The original PBs still exist for those that want to use non-polling.

The same goes for the test PB. The V2 test PB uses the V2 PB that uses polling.

aburt-demisto:
I have an Enterprise key at the moment that I can demo both with.

idovandijk:
> @idovandijk There should be a V2 of the PBs that support poilling by default. This seemed logical to have polling enabled by default in them. The original PBs still exist for those that want to use non-polling.
> 
> The same goes for the test PB. The V2 test PB uses the V2 PB that uses polling.

What I see in this PR is a modification of the original playbook which has `fromversion: 5.0.0`.
I believe what you should be doing is create a new playbook with the same name and ID in a separate file, add `fromversion: 6.2.0` to it. Then edit the existing (older) playbook and add a `toversion: 6.1.9` to it.
This way, starting from 6.2.0 users will get the new playbook, and up to that version they will get to keep the older version of it.

Am I missing something?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
GuyAfik: fixed

idovandijk:
@aburt-demisto How are you doing? Do you need any assistance with the PR?

idovandijk:
@aburt-demisto what is the change in `PAN-OS Query Logs For Indicators`? It says in the RN that that is a new playbook but it already exists in content.
Regarding the v2 playbook it would be better to name the file with `6_2` suffix instead of a `_v2` since it's not a v2 (if we end up releasing a v2 the name would be taken).

Packs/PAN-OS/ReleaseNotes/1_14_17.md
```diff
@@ -0,0 +1,8 @@
+
+#### Integrations
+##### Palo Alto Networks PAN-OS
+- Added support for *polling* to the ***pan-os-query-logs*** command.
+
+#### Playbooks
+##### New: PAN-OS Query Logs For Indicators V2
```
idovandijk, line 7: This should be `Panorama Query Logs`, and we can explain that starting in 6.2 the command will poll instead of using the GenericPolling playbook

idovandijk (commented):
Almost there! Only left to fix RN

GuyAfik (approved):
@aburt-demisto thank you so much for this incredible contribution, really well done!
Looks good to me, once @idovandijk will approve ill fix conflicts and merge it.

idovandijk (approved):
Looks good to me as well. Well done.

PR #20264, opened 2022-07-27
Add Missing `None` Type to BaseClient's Docstring

MichaelYochpaz:
## Status
- [ ] In Progress
- [ ] Ready
- [x] In Hold - (Reason for hold)

## Description
Some of `BaseClient`'s constructer parameters accept a `None` value (according to the docstring), but a `None` type isn't specified in their `types` attribute on the docstring.

This causes PyCharm (and possible other IDEs and linters?) to show a warning saying that an invalid type is used.
See the screenshot below for an example.

This PR fixes the docstring and adds a `None` type where it was missing.

## Example
Screen Shot 2022-07-27 at 14 52 04

The docstring (which specifies that None is a valid value):
```
:type headers: ``dict``
:param headers:
        The request headers, for example: {'Accept`: `application/json`}.
        Can be None.
```

   - [ ] Yes
       - Further details:
   - [x] No


bziser (changes requested):
Great,
just generate rn :)

bziser (approved):

PR #20298, opened 2022-07-28
Cmd analysis fix commandline inputs

issolahma:
## Status
- [x] In Progress
- [ ] Ready

## Description
By using the "commandline" field no encoded command can be found, and it raises an error when trying to find a B64 string.
It works when using "commandline.original".

I also changed the playbook output in the same pattern so it returns the original command line as written in the readme. If not it returns both of the original and decoded command.

## Screenshots
The default settings
b643

The result is falsely "None" when using "commandline"
b641

The result is "yes" with "commandline.original"
b642

In task "Check if base64 string exists in command-line" with the default input.
Works with "commandline.original".
b645

   - [ ] Yes
       - Further details:
   - [x ] No


issolahma:
I also fixed two typos in inputs name

MosheEichler (approved):

idovandijk:
@issolahma thank you for waiting patiently as we review this PR.

@altmannyarden I think you would know here better than me. Was this intentional that we were using `commandline` instead of `commandline.original`? Will changing it break BC?

idovandijk (commented):
@issolahma Thank you for this fix. Please update the release notes to explain the bug that was fixed. Something like:
"Fixed issues where the command-line was parsed incorrectly which caused errors in the playbook".

issolahma:
Done

idovandijk (approved):
Great job!

MosheEichler:
@issolahma Hi, can you please resolve the conflicts?

issolahma:
Oh sorry I didn't know I was supposed to do that.
Can I just change the number in the files and rename `2_2_11.md` with the new version number?

issolahma:
Okay, I think it's solved without damages.

PR #20313, opened 2022-07-31
Add Ignore Event Retries Param to SlackV3

amshamah419:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: XSUP-14848

## Description
Added parameter to allow overriding the default behavior where the listener will ignore retried events. This will result in potentially double-posted messages, but is mentioned and disclosed in the docs.


Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1416,7 +1417,7 @@ async def listen(client: SocketModeClient, req: SocketModeRequest):
     if req.envelope_id:
         response = SocketModeResponse(envelope_id=req.envelope_id)
         await client.send_socket_mode_response(response)
-    if req.retry_attempt:
+    if req.retry_attempt and IGNORE_RETRIES:
```
yuvalbenshalom, line 1420: in case IGNORE_RETRIES=False, and req.retry_attempt > 0, we should log that a retry was handled and the the attempt count.

yuvalbenshalom (changes requested):
see my comment

yuvalbenshalom (approved):

PR #20417, opened 2022-08-05
Fix AWS-GuardDuty not respecting the region parameter

ANT1OFF:
## Status
- [ ] In Progress
- [ X ] Ready

## Related Issues
fixes: AWS-GuardDuty not respecting the region parameter

## Description
Currently the AWSClient used by this integration has its region configured to always be the default region. This change overrides the default region if the region parameter is supplied.

   - [ ] Yes
       - Further details:
   - [ X ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/AWS-GuardDuty/Integrations/AWSGuardDuty/AWSGuardDuty.py
```diff
@@ -546,6 +546,9 @@ def main():
     timeout = params.get('timeout') or 1
     retries = params.get('retries') or 5
 
+    if demisto.args().get('region') is not None:
+        aws_default_region = demisto.args().get('region')
```
israelpoli, line 550: It is better to leave the meaning of the variable name as it is.

Therefore, in lines 556-557 add like this:
```
region = demisto.args().get('region')
client = aws_client.aws_session(service=SERVICE, region=region)
```

israelpoli (changes requested):
Hi @ANT1OFF 
Thank you for your contribution

please see my comment below
In addition, 
Run `demisto-sdk update-release-notes -i ~/Packs/AWS-GuardDuty` in terminal, for creating RN file, and add there a note for your changes.

you can look at in [here](https://xsoar.pan.dev/docs/documentation/release-notes), how RN file should be.

israelpoli:
@ANT1OFF 
Did you notice my comment about the RN file?

ANT1OFF:
@israelpoli done.

israelpoli (approved):
@ANT1OFF
 
Thanks a lot

PR #20437, opened 2022-08-08
PAN-OS Integration - New Command Bugfixes

adambaumeister:
This is a bugfix for some minor issues;

* Commands that were using `get_single_device()` in the topology class would fail if the target was Panorama. This is because the pan-os-python SDK does not populate the device serial number for Panorama instances. This affected the following commands;
   * pan-os-platform-reboot

To fix this issue, I've set the topology discovery function `add_device_object` to store the hostname in the relevant dictionaries instead of the serial number - this means the device dicts will always be keyed by Hostname for Panorama devices and Serial Number by firewalls. I've also updated the retrieval functions to reflect this. Previously, this was simply failing and literal `None` was being stored in the dictionary as the key for Panorama devices..

* LogForwardingProfile was specified as a valid option in the object command but was not implemented.

The fix for this was just to add the object type into the supported object dictionary.

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -9240,6 +9240,7 @@ class ObjectGetter:
         "SecurityProfileGroup": SecurityProfileGroup,
         "SecurityRule": SecurityRule,
         "NatRule": NatRule,
+        "LogForwardingProfile": LogForwardingProfile,
     }
 
     @staticmethod
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -7943,6 +7943,11 @@ class Topology:
         :param device: Either Panorama or Firewall Pandevice instance
         """
         if isinstance(device, Panorama):
+            if device.serial:
+                serial_number_or_hostname = device.serial
+            else:
+                serial_number_or_hostname = device.hostname
+
             # Check if HA is active and if so, what the system state is.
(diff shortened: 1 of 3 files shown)


adambaumeister:
@GuyAfik Fixes for some minor bugs.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -7943,6 +7943,11 @@ def add_device_object(self, device: Union[PanDevice, Panorama, Firewall]):
         :param device: Either Panorama or Firewall Pandevice instance
         """
         if isinstance(device, Panorama):
+            if device.serial:
+                serial_number_or_hostname = device.serial
+            else:
+                serial_number_or_hostname = device.hostname
```
GuyAfik, line 7949:
```suggestion
            serial_number_or_hostname = device.serial if device.serial else device.hostname
```

GuyAfik (commented):
@adambaumeister looks great! Can you please add RN and ill merge?

Packs/PAN-OS/ReleaseNotes/1_14_14.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Palo Alto Networks PAN-OS
+- Fixed issue when Panorama systems are used in target arguments for certain commands
+- Added `LogForwardingProfile` to list of objects supported by `pan-os-config-get-objects`
```
GuyAfik, line 5:
```suggestion
- Fixed an issue when Panorama systems are used in target arguments for certain commands.
```

GuyAfik (approved):
Thanks you @adambaumeister, great work as usual!

PR #20523, opened 2022-08-11
TestCollection: fix default value

dorschw:
self._tests was [none if missing](https://code.pan.run/xsoar/content/-/jobs/15412043#L3327), now it's an empty tuple

### DIFF
--- a/Tests/scripts/collect_tests/collect_tests.py
+++ b/Tests/scripts/collect_tests/collect_tests.py
@@ -662,7 +662,7 @@ def output(result: Optional[CollectionResult]):
 
 
 if __name__ == '__main__':
-    logger.info('TestCollector v20220811')
+    logger.info('TestCollector v20220811.2')
     sys.path.append(str(PATHS.content_path))
     parser = ArgumentParser()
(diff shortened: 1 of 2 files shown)


Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -662,7 +662,7 @@ def output(result: Optional[CollectionResult]):
 
 
 if __name__ == '__main__':
-    logger.info('TestCollector v20220811')
+    logger.info('TestCollector v20220811.2')
```
GuyAfik, line 665: what's this change?

GuyAfik (approved):

Tests/scripts/collect_tests/collect_tests.py
dorschw: mostly for debug purposes, to know exactly what TestCollect version is running on each build (with this fix, w/o the other fix etc)

GuyAfik: Thanks for the clarification!

PR #20529, opened 2022-08-12
TaegisXDR - Migrate to Alerts v2 API

reedium:

## Status
- [ ] In Progress
- [x] Ready

## Description
The Taegis Alerts API has migrated to newer endpoints. The existing endpoints no longer exist for pulling Alerts. The existing integrations are currently broken due to this migration.

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [x] Yes
       - Further details: The old Taegis Alerts v1 API has been decommissioned and is no longer available for use. The new API has different formats to the data (for instance, `alerts.metadata.title` instead of `alerts.message`). This means existing users of the integration will need to make a few modifications to support the new API. 
   - [ ] No

## Must have
- [x] Tests
- [x] Documentation


Packs/SecureWorks/Integrations/TaegisXDR/TaegisXDR.py
```diff
@@ -194,67 +197,111 @@ def execute_playbook_command(client: Client, env: str, args=None):
...
-    if result["data"]["alerts"]:
-        readable_output += "\n\n### Alerts\n"
-        for alert in result["data"]["alerts"]:
-            readable_output += f"* [{alert['message']}]({ENV_URLS[env]['xdr']}/alerts/{alert['id']})\n"
+    readable_output = f'## Results\nFound {len(alerts)} alerts'
 
-    outputs = result["data"]["alerts"]
+    if alerts:
+        readable_output += "\n\n### Alerts\n"
+        for alert in alerts:
```
JasBeilin, line 297: Have you consider using the TableToMarkdown function?
you can check https://xsoar.pan.dev/docs/integrations/code-conventions#tabletomarkdown for more info.

Packs/SecureWorks/Integrations/TaegisXDR/README.md
```diff
@@ -59,14 +59,18 @@ After you successfully execute a command, a DBot message appears in the War Room
...
+| ids | A list of alerts by IDs | `936c1cc1-db8f-430c-837c-1c914fcca35a` | False |
+| limit | Number of results to when `ids` is not defined | `10` | False |
+| offset | The result to start from when `ids` is not defined | `0` | False |
+| cql_query | The query to utilize when searching for Alerts | `from alert severity >= 0.6 and status='OPEN'` | False |
 
 #### Command Examples
 
 ```
 !taegis-fetch-alerts ids=`["6594e97f-a898-5b28-82b2-ea03293cdaa1"]`
+!taegis-fetch-investigation-alerts id=936c1cc1-db8f-430c-837c-1c914fcca35a
```
JasBeilin, line 73: If these are the same command, maybe we should consider deprecating the first one?

Packs/SecureWorks/Integrations/TaegisXDR/TaegisXDR.yml
```diff
@@ -66,9 +66,17 @@ script:
   - name: taegis-fetch-alerts
     arguments:
     - name: ids
-      required: true
```
JasBeilin, line 69: Removing this will cause the command to run even when "ids" were not provided. Are you sure this is your intention?

JasBeilin (changes requested):
Good job. I have some concerns regarding these changes. We should maybe consider deprecating old commands instead changing it to avoid customer's confusion. What do you think?

Also I made some comments for you to check. Let me know if you need any assistance with these changes.

Packs/SecureWorks/Integrations/TaegisXDR/README.md
reedium: I think this is a typo, I accidentally added the `taegis-fetch-investigation-alerts` line there. Technically, they're not the same command (1 is for pulling a specific alert while the other is for any alerts related to a specific investigation).

We could potentially remove `taegis-fetch-investigation-alerts` command altogether as the alerts can be fetched from the `fetch-investigation` query as well, however this command still works and I'd rather not remove this in case it's in use by any users.

I'll remove this line from the README though as that command is actually described further down in the doc.

Packs/SecureWorks/Integrations/TaegisXDR/TaegisXDR.yml
reedium: Yup, this shouldn't be a problem. The original code used our old Alerts api which did not have a good way of defining the criteria of what alerts would be pulled back, which is why the command originally required the IDs. The new API has better support for searching by specific criteria so the IDs should no longer be required.

With the modified command. we're looking to allow the user to either:

1. Pull back specific alerts based on a list of IDs
2. Pull back all alerts based on criteria. The default criteria has a limit of 10 alerts that are open and with high severity

We have a max number of results that can be returned by the API so I don't believe this should be a problem.

Packs/SecureWorks/Integrations/TaegisXDR/TaegisXDR.py
reedium: I haven't, although that does seem useful. If possible, I'm looking to make a few more smaller changes to the integration in a few weeks and would like to include this with those changes but I'd like to get this update completed first due to the current integration being broken from our Alerts v1 API decom.

reedium:
> Thank you for your contribution! Good job. I have some concerns regarding these changes. We should maybe consider deprecating old commands instead changing it to avoid customer's confusion. What do you think?

Sure, if that makes more sense I can create new commands instead of modifying the old commands. Although with the current integration, the existing commands aren't working at the moment anyway (the old API has been disabled completely) so in either case, a user is going to have to make changes on their side.

Let me know what makes the most sense.

It's worth noting in a few months, we'll be looking at releasing a new version of our Investigations API which potentially could mean additional commands if we did do this, depending on the new command names:

```
!taegis-fetch-investigation
!taegis-fetch-investigation2 <-- Alerts v2 - this change
!taegis-fetch-investigationsv2 <-- Investigations v2 - future change
```

JasBeilin:
Ok, If you are having another version soon, we can make some of these changes then (I strongly suggest the TableToMarkdown change as it handles a lot of edge cases).
Just to make sure, the fetch-alert is the only actual change, right?

reedium:
> Ok, If you are having another version soon, we can make some of these changes then (I strongly suggest the TableToMarkdown change as it handles a lot of edge cases). 

Sounds good. I'll look at moving to `TableToMarkdown` in the next PR.

> Just to make sure, the fetch-alert is the only actual change, right? If thats the case, Lets have a demo and move this forward.

Technically, the following commands were modified:
* `taegisxdr-fetch-alerts`
* `taegisxdr-fetch-incidents` - Replaced `alerts` with `alerts2`
* `taegisxdr-fetch-investigations` - Replaced `alerts` with `alerts2`

I'm available to demo the above when needed

JasBeilin:
Great, thanks for the clarification.
I will invite to a meeting, I am not sure about the timezone so let me know if it needs to be moved :)

Packs/SecureWorks/ReleaseNotes/4_0_0.json
```diff
@@ -0,0 +1 @@
+{"breakingChanges":true,"breakingChangesNotes":"Version 1 of the Alerts API has been decommissioned and replaced with Alerts v2. This migration means the fields in the alerts now differ from the original fields. Additionally, Investigations no longer use the `alerts` field, but now contain `alerts2`. Any references to the Alerts v1 fields will fail."}
```
JasBeilin, line 1: Please add the context change here as well, as this will show in the Breaking Change warning.

JasBeilin (approved):

PR #20542, opened 2022-08-14
feat(intezer): add phishing email playbook, deprecates scan host

davidt99:

## Status
- [ ] In Progress
- [x] Ready

## Description
* Add phishing playbook
* Deprecates scan host since D2 agent is deprecated

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


davidt99:
@ShacharKidor would like your review :)

davidt99:
@idovandijk the build passes, I would like your review.

Packs/Intezer/Integrations/IntezerV2/IntezerV2.yml
```diff
@@ -3,324 +3,324 @@ commonfields:
...
-    type: 8
+- display: API Key
+  name: APIKey
+  required: true
+  type: 4
+- display: Intezer Analyze Base URL
+  name: AnalyzeBaseURL
+  defaultvalue: https://analyze.intezer.com/api/
+  type: 0
+  required: false
```
JasBeilin, line 14: While this is not part of the change, are you sure you want the url to be allowed to remain empty?

Packs/Intezer/ReleaseNotes/1_4_0.md
```diff
@@ -0,0 +1,18 @@
+
+#### Integrations
+##### Intezer v2
+- Send the original file name in analyze-by-file command
```
JasBeilin, line 4: How does it impact the users? As I see it, this happens "behind the scenes", and might confuse users.

JasBeilin (commented):
Great work on this PR. I had some minor comments that I would like you to address.
Let's do a very short demo to see everything is working as expected.

Packs/Intezer/Integrations/IntezerV2/IntezerV2.yml
davidt99: If I understand correctly if it's not set, the default value is used, which is the expected behavior

Packs/Intezer/ReleaseNotes/1_4_0.md
davidt99: I can remove it, but the file name is presented in Intezer Analyze, which currently it isn't informative (it usually something like that `ad0 @ 821`.

Packs/Intezer/Integrations/IntezerV2/IntezerV2.yml
JasBeilin: The default value is what we see when configuring the integration.
If it is erased, since the param is not required It will take the value to be an empty string.
But making this param required will require breaking changes, and will not be backward compatible so let's leave it for now.
Another option that will not break backward- you can handle it as part of the code in the main function: if the value is an empty string, change t to the hardcoded string you want. But I leave it for you to decide.

davidt99: Let's leave as is. I might change it when we will completely remove the d2-related playbook since it will be a breaking change.

idovandijk (changes requested):
Thank you for your contribution and for patiently waiting for my review. You've done a great job.

These are my points from the review:
1. Please update the readme file of the deprecated playbook to reflect that it is deprecated.
2. There is no need to have separate playbook inputs for `URL` and `URL.Data`, as the former already includes the latter (please remove the redundant `URL.Data`). The same goes for File and File.EntryID.
3. You need to make a decision on whether this is a parent playbook (playbook is triggered by an incident) or a subplaybook (playbook is being called from another playbook as part off an incident), because it makes a difference in how the data is used. Since your integration is not a fetching integration, and since this playbook does not cover the full lifecycle of an incident, I believe it is more of a subplaybook - a building block that users may want to add to their existing playbooks. Based on that, I have more comments:
4. Please change all references to `${File}` and `${URL}` to refer to the playbook **inputs** instead. Thus becoming: `${inputs.File}` and `${inputs.URL}`. For example, this should be `${inputs.File.SHA256}:
image
5. Please fix the commands to meet our [context standards](https://xsoar.pan.dev/docs/integrations/context-standards-mandatory#url) - the outputs are not following are standards:
image
If you need help @JasBeilin should be able to help. If these commands are not new, then please add the new outputs without removing the older ones, or consider changing the existing ones and breaking backward-compatibility.
6. After fixing the outputs, please update the playbook to output the correct outputs:
image
all of these should be upper case.
The reason for this is that users who build and use content already expect data to be in those paths.
7. Recommendation (not mandatory): you can add the File and URL analysis tasks to the existing generic File Enrichment and URL Enrichment playbooks respectively. This will boost your exposure and improve OOTB content for users :)
If you decide to do that, make sure to check the "skip this branch...if unavailable` option: 
image
that would ensure that if users who don't have your Pack installed, but still use the generic playbooks, would not get any errors due to missing content.

Once you've finished your fixes, please schedule a short demo with @JasBeilin, @kgal-pan and me so we can see an incident in action and make sure the fixes have been incorporated correctly.

Thanks again for your hard work!

kgal-pan:
Hey @davidt99 - Any updates here?

davidt99:
@kgal-pan I just push @idovandijk requested changes so I'll schedule a demo in slack

idovandijk:
Summing up points from demo:
1. @JasBeilin to see recorded demo and add points if needed.
2. @davidt99 to decide whether to get rid of Url and keep only URL, or support both.
3. @davidt99 to test the playbook with varying number of files/URLs and investigate error in the Generic Polling flow.

Thank you for your time and effort!

JasBeilin (approved):
After seeing the demo, I think i'm good with the changes. 
I would like you to add the `deprecate` you added to the readme to the context description in the yml as well. While not a must, it will allow clients to see the description and find out its deprecated, instead of getting no result and go the readme to investigate why.

So I am approving, but will not merge before @idovandijk gives the OK after you will fix the errors and make the sure multiple files and urls in your playbook loop are working as expected when contains different number of arguments.

idovandijk (approved):
Great work guys!
Have a good weekend :)

PR #20551, opened 2022-08-15
Dataset formatter xsiem

omerKarkKatz:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3711

## Description
The purpose of this script is to process exported datasets from XSIEM as lookup tables


Utils/Dataset_formatter_xsiem/dataset_formatter_xsiem.py
```diff
@@ -0,0 +1,239 @@
...
+            - A list of the formatted col
+        """
+        return self.format_type.convert_to_xsiam_format(col)
+
+    def formatter_name(self):
+        return self.format_type.name
+
+
+def open_input_file(path: Path, f):
+    if path.suffix == '.tsv':
```
gal-forer, line 176: Do you support only tsv files ? If so please update the description in the parser

omerKarkKatz: No i support also csv files. 
This is specified in the script parameters.
'A Script to format csv/tsv files in order to upload to XSIAM'
here in case the file is a tsv format we read it with a sep='\t' flag.

gal-forer (approved):

PR #20555, opened 2022-08-15
TAXII2: return 406 on version mismatch

dorschw:
Returning 406 on invalid headers
closes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15403


DeanArbel (approved):
Nice job.

PR #20620, opened 2022-08-18
JiraV2: use the data value instead delta for mirroring 

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-14620

## Description
When mirror out mapper contain mapping for `incident status` the value in `delta` obj for this are empty 
while the data in the `data` obj are correct

server issue 
but as we can solve it via integration code - this PR contain the fix

@yaakovi need to open ticket for Server team?

### DIFF
--- a/Packs/Jira/Integrations/JiraV2/JiraV2_test.py
+++ b/Packs/Jira/Integrations/JiraV2/JiraV2_test.py
@@ -401,7 +401,7 @@ def test_update_remote_system_delta(mocker):
         {
             "incidentChanged": "17757",
             "remoteId": "17757",
-            "data": {"summary": "data"},
+            "data": {"summary": "data", "not_changes_key": "not_changes_val"},
             "delta": {"summary": "changes"},
         }
     )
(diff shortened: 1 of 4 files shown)


Packs/Jira/Integrations/JiraV2/JiraV2_test.py
```diff
@@ -400,10 +401,12 @@ def test_update_remote_system_delta(mocker):
         {
             "incidentChanged": "17757",
             "remoteId": "17757",
+            "data": {"summary": "data"},
```
yaakovi, line 404: add another key that should not be updated

Packs/Jira/ReleaseNotes/2_0_9.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Atlassian Jira v2
+- Fixed an issue where mirror out not work as expected.
```
yaakovi, line 4:
```suggestion
- Fixed an issue where the mirror out did not work as expected.
```

yaakovi (approved):

PR #20658, opened 2022-08-21
Azure Sentinel Force usage of timestamps for fetch.

amshamah419:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: XSUP-15687

## Description
I suspect the issue is with the query, so we are forcing usage of the timestamps


Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -904,7 +906,7 @@ def fetch_incidents(client: AzureSentinelClient, last_run: dict, first_fetch_tim
     last_incident_number = last_run.get('last_incident_number')
     demisto.debug(f"{last_fetch_time=}, {last_fetch_ids=}, {last_incident_number=}")
 
-    if last_fetch_time is None or last_incident_number is None:
```
yuvalbenshalom, line 907: shulden't this effect only the first fetch?'

yuvalbenshalom (commented):
why do we add a parameter and not just using it all the time if it I the first fetch?

amshamah419:
My rationale was that for some customers, time based ingestion wouldn't be ideal so I didn't want to break that for them.

Mainly was trying to preserve the existing functionality as best I can.

PR #20662, opened 2022-08-21
Fix Splunk auth issue

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15601
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15603

## Description
When using `basic=true` in the Splunk connection - the token stored in are wrong 
we shouldn't use it in this case in the command like `splunk-notable-event-edit`


yuvalbenshalom (changes requested):
looks good.
1. see my RN comment.
2. this need to be changed also in splunkpy prerelease. consider adding it here or creating another PR and connect it to this.

yuvalbenshalom (approved):

PR #20690, opened 2022-08-22
changed domain in tpb

yucohen:
changed the domain for testing, as the old one did not retrieve any data and failed the tpb


bziser (approved):
Let's try 🤞

PR #20693, opened 2022-08-22
Update scripts hours and dates

iyeshaya:
[BetweenHours, BetweenDates] Add incident context outputs

CIAC-3813


ShahafBenYakir (approved):
Great! Let's add this to the script README

PR #20749, opened 2022-08-24
TestCollect: fix version range

dorschw:


dorschw:
continued on https://github.com/demisto/content/pull/20813 due to a messy merge

PR #20814, opened 2022-08-29
Added SetWithTemplate, FormatTemplate and ConcatFormat

spearmin10:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
Build a text from a template text.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/FormatTemplate/test_data/test.json
```diff
@@ -0,0 +1,177 @@
+[
+  {
```
anara123, line 2: Please provide command examples with inputs for both commands and document the outputs. This way it will be much more clearer for the user how to use the commands.

spearmin10: @anara123 OK. I've documented them to README.

spearmin10:
@anara123 Hopefully, I would like to name `ctx_incident`, not `ctx_inc` in the argument parameters for better naming. But I got an error below when commiting. Is it possible to do with adding `ctx_incident` to `argsExceptionsList`.

image

anara123:
> @anara123 Hopefully, I would like to name `ctx_incident`, not `ctx_inc` in the argument parameters for better naming. But I got an error below when commiting. Is it possible to do with adding `ctx_incident` to `argsExceptionsList`.
> 
> image

@spearmin10 It is not possible to change `argsExceptionsList` - this requires change on the server side.
So we change it only in rare cases - because it will require new server version.
Thus you better use `ctx_inc`

Packs/CommonScripts/Scripts/SetWithTemplate/SetWithTemplate.py
```diff
@@ -0,0 +1,259 @@
...
+        ci = si
+        while ci < len(source):
+            if end_marker is not None and Formatter.__is_end_mark(source, ci, end_marker):
+                key = source[si:ci] if out is None else str(out) + source[si:ci]
+                xval = ''
+                if extractor:
+                    xval = extractor(key, dx)
+                    if self.__keep_symbol_to_null and xval is None:
+                        xval = self.__start_marker + key + self.__end_marker
+                return xval, ci + len(end_marker)
```
yaakovi, line 105: Should there be a return here?
if I'm not mistaken:
```python
__extract('> Hello my friend', extract_dt, {}, 0, '>')
```
will result in `ci=0`, `key='>'`, and `xval=None`
which will return `None, 1`

spearmin10: @yaakovi That's expected behavior. `end_marker` is only given for parsing the value within ${}. So the extracted value immidiately returns when the end marker is found.

spearmin10: I've added a docstring for it.

spearmin10:
@yaakovi Could you help me to solve the duplicated release notes? I cannot remove the one from the PR.

spearmin10:
@anara123 I've managed to solve the git issue. Please review again.

yaakovi (approved):

PR #20832, opened 2022-08-29
TestCollect: collect test dependencies

dorschw:
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3928

When collecting a test, collect all packs on which it depends (but not their tests)


bziser (approved):
Did with you a review face to face :) 
Looks good!

PR #20853, opened 2022-08-30
Fix a bug when the query includes &

israelpoli:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15597

## Description
The customer encountered a problem when a query was sent with &, and the result was not correct,
The bug was caused by an incorrect parse of the query

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ServiceNow/ReleaseNotes/2_3_21.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### ServiceNow v2
+- Updated the Docker image to: *demisto/python3:3.10.6.33415*.
+- Fixed an issue in ***servicenow-query-table*** command where the query was not parsed correctly when it contained an ampersand, returning an incorrect result.
```
yuvalbenshalom, line 5:
```suggestion
- Fixed an issue in ***servicenow-query-table*** command where the query was not parsed correctly when it contained an ampersand, returning an incorrect result.
- **Breaking changes**: using '&' to concat ServiceNow queries is deprecated. please use the '^' character instead.
```

yuvalbenshalom (changes requested):
looks good, see my comment.

PR #20857, opened 2022-08-31
PAN-OS - Fix incorrect parsing of Panorama High-Availability state XML

adambaumeister:

## Status
- [] In Progress
- [] Ready
- [X] In Hold - Pending customer validation

## Description
This fixes a bug when the instance of panorama is configured in High-availaiblity. We correctly look for the panorama HA state, but the result XML was not being parsed correctly, which results in all commands failing as it cannot resolve which server is active.

@GuyAfik @jpeng19 

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [X] Tests
- [ ] Documentation


GuyAfik (approved):
Looks great, @adambaumeister can you please pull from master to re-trigger the build?

adambaumeister:
Waiting on customer to confirm fix then will do!

adambaumeister:
@GuyAfik we haven't heard back from the customer, might be a good sign that it's working. Your call whether to merge or keep waiting. I've updated the branch from upstream.

GuyAfik:
> @GuyAfik we haven't heard back from the customer, might be a good sign that it's working. Your call whether to merge or keep waiting. I've updated the branch from upstream.

@adambaumeister lets wait, at the meantime please add release-note and bump the pack-metadata

GuyAfik:
@adambaumeister please pull from master again, fixed an issue in the pan-os build.

GuyAfik:
@adambaumeister lets push forward that one

adambaumeister:
fyi -

The original fix is verified working from a seperate user, so that part is good. However, another user has reported that after fixing the HA component, now get-ha-status does not return the correct status and peer for Panorama specifically. This was due to the mishandling of the "state" field in the XML when referring specifically to Panorama. I've added this and also updated from master, so I would just wait on the confirmation of that fix before merging.

adambaumeister:
I had to replicate the topology in the lab. The issue was due to an incorrect statement overriding the dictionary that stores the High-availability pairs. I added the following code;

```python
        # This is only true if Panorama is in HA mode as well.
        if self.ha_pair_serials:
            self.ha_pair_serials = {**self.ha_pair_serials, **ha_pair_dict}
        else:
            self.ha_pair_serials = ha_pair_dict
```

It's important to note that HA Panorama is still not fully supported, and it was never intended to be, as it is only possible to add one panorama IP into the integration at a time. This is a feature that could be added at another time if there is a reason for it.

GuyAfik (approved):
That is amazing!

PR #20892, opened 2022-09-01
TestCollect: split git diff lines by `\t`

dorschw:
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3962


Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -579,7 +579,7 @@ def _get_changed_files(self) -> tuple[str, ...]:
 
         # diff is formatted as `M  foo.json\n A  bar.py\n ...`, turning it into ('foo.json', 'bar.py', ...).
         for line in diff.splitlines():
-            match len(parts := line.split()):
+            match len(parts := line.split('\t')):
```
esharf, line 582:
```suggestion
            match len(parts := re.split(r'\s', line):
```
we can use regex here

esharf (approved):

Tests/scripts/collect_tests/collect_tests.py
dorschw: i rather keep it readable. 
`\s` is invalid, see the ticket this issue fixes - caused by files with whitespaces in their name.

esharf: \t will not work for renamed files.
Regarding the original problem, you can use the max line argument.

esharf (approved):

Tests/scripts/collect_tests/collect_tests.py
dorschw: following offline discussion, `\t` it is

PR #20920, opened 2022-09-04
TestCollect: failed nightly dependency search

dorschw:


dorschw:
test nightly: https://code.pan.run/xsoar/content/-/pipelines/3595106

PR #20961, opened 2022-09-06
Upgrade HTTPListRedirects to py3

anas-yousef:
## Status
- [ ] In Progress
- [x] Ready

## Description
Made the script python3 compatible and fixed a small bug where we did not save the response of one of the requests

## Screenshots
image
image

   - [ ] Yes
       - Further details:
   - [x] No


tkatzir (approved):
Great!

Packs/CommonScripts/Scripts/HTTPListRedirects/HTTPListRedirects.py
```diff
@@ -5,27 +5,54 @@
...
+    urls = []
+    if response.history:
+        for resp in response.history:
+            urls.append({'Data': resp.url, 'Status': resp.status_code})
+    urls.append({'Data': response.url, 'Status': response.status_code})
+    return urls
+
+
+def delete_environment_variables(use_system_proxy: str):
+    if use_system_proxy == 'false':
```
tkatzir, line 34:
```suggestion
    if use_system_proxy.lower() == 'false':
```

Packs/CommonScripts/Scripts/HTTPListRedirects/HTTPListRedirects.py
```diff
@@ -5,27 +5,54 @@
...
+        for env_var in env_variables:
+            if(env_var in os.environ):
+                del os.environ[env_var]
+
+
+def main():
+    url = demisto.args().get('url')
+    try:
+        delete_environment_variables(use_system_proxy=demisto.args().get('use_system_proxy'))
+        url = f'http://{url}' if not url.lower().startswith('http') else url
```
tkatzir, line 45: Question: If `trust_any_certificate` is `True`, doesn't it imply that it's https?

Packs/CommonScripts/Scripts/HTTPListRedirects/HTTPListRedirects.py
```diff
@@ -5,27 +5,54 @@
...
+
+        response = get_response(url=url,
+                                use_head=demisto.args()['useHead'],
+                                verify_ssl=demisto.args().get('trust_any_certificate') != 'true')
```
tkatzir, line 49:
```suggestion
                                verify_ssl=demisto.args().get('trust_any_certificate', 'true').lower() != 'true')
```
Assuming `demisto.args().get(...)` supports default values.
In any case, it's better to `lower()` the value, so we're sure we're in the same case.

Packs/CommonScripts/Scripts/HTTPListRedirects/HTTPListRedirects.py
```diff
@@ -5,27 +5,54 @@
...
+        history_urls = get_response_history(response=response)
+        demisto.results(create_command_result(history_urls=history_urls))
+
+    except Exception as e:
+        return_error(f'Failed to execute script.\nError:\n{str(e)}')
```
tkatzir, line 54:
```suggestion
        return_error(f'Failed to execute script. Error:\n{str(e)}')
```
Too many newlines...

tkatzir (approved):
Looks great!
Some minor comments and questions.
Approved.

Packs/CommonScripts/Scripts/HTTPListRedirects/HTTPListRedirects.py
anas-yousef: The opposite, if *trust_any_certificate* is *True*, that means we don't want to use the *SSL Protocol* to encrypt the request which means it's not *https* (Correct me if I'm wrong)

tkatzir: I don't think it's related.
`trust_any_certificate` is relevant only with HTTPS - It tells the client to verify the server's certificate.
It's irrelevant with HTTP.

PR #20998, opened 2022-09-06
 deprecate archerRSA scripts

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Description
After consulting @bakatzir we decided to deprecate the archerRSA python 2 scripts rather than moving them to python 3

   - [ ] Yes
       - Further details:
   - [X] No


tkatzir (commented):
Need to add a note about a replacement, or whether there's no available replacement.
Also, need to change the description.
See the templates.

Packs/ArcherRSA/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "RSA Archer",
     "description": "The RSA Archer GRC Platform provides a common foundation for managing policies, controls, risks, assessments and deficiencies across lines of business.",
     "support": "xsoar",
-    "currentVersion": "1.2.0",
+    "currentVersion": "1.3.0",
```
bakatzir, line 5: should be a revision, no new features are introduced to the pack.

bakatzir (approved):

YuvHayun:
This pr needs to be force merged since the scripts were split and the old files were deleted.

PR #21007, opened 2022-09-06
Moving to py3 RunDockerCommand

daryakoval:
Moving to py3 RunDockerCommand


Packs/CommonScripts/Scripts/RunDockerCommand/RunDockerCommand.py
```diff
@@ -0,0 +1,33 @@
+import subprocess
+
+import demistomock as demisto  # noqa: F401
+from CommonServerPython import *  # noqa: F401
+
+ret = []
```
tkatzir, line 6: Move the "meat" of the code into a method, and call it from
```
if __name__ in ["__builtin__", "builtins"]:
    main()
```

tkatzir (changes requested):
Good, with one comment

tkatzir (approved):

PR #21013, opened 2022-09-06
Salesforce ask user python2 to python3

daryakoval:
Split and added unit test.


dantavori (approved):

daryakoval:
Need force merge since I splited the script and now validation fails that I deleted file.

PR #21084, opened 2022-09-07
Fp set rule 2to3

daryakoval:
Changed 2 scripts from py2 to py3 and added scripts.


Packs/Forcepoint/Scripts/FPDeleteRule/FPDeleteRule.py
```diff
@@ -0,0 +1,54 @@
...
+import demistomock as demisto  # noqa: F401
+from CommonServerPython import *  # noqa: F401
+
+FILTER_CONFIG_PATH = "/opt/WCG/config/filter.config"
+CMD_DEL_RULE_FORMAT = "sed -i '/^{0}={1} action=[A-Za-z]*$/d' {2}"
+CMD_TRITON_RELOAD_CONFIG = "/opt/WCG/bin/content_line -x"  # && /opt/WCG/WCGAdmin runds"
+
+
+def delete_rule(ruleType):
+    if ruleType not in ["dest_domain", "dest_ip", "dest_host", "url_regex"]:
```
tkatzir, line 13: Create a constant with the rule types, and use it here and in the error message.

Packs/Forcepoint/Scripts/FPDeleteRule/FPDeleteRule_test.py
```diff
@@ -0,0 +1,38 @@
+import demistomock as demisto  # noqa: F401
+from CommonServerPython import *  # noqa: F401
+
```
tkatzir, line 3: Missing test(s) for mutual exclusion of `tritonsystem` and `remoteaccessname`

tkatzir (changes requested):
Looks good.
Same comments for all files.

Packs/Forcepoint/Scripts/FPDeleteRule/FPDeleteRule_test.py
daryakoval: Why do we need this? The coverage is 70% it does not need to be 100%. 
This test will not find any new issue. the test will use 2 mockers to test 2 lines that were mocked. It will not check any logic!

Packs/Forcepoint/Scripts/FPDeleteRule/FPDeleteRule.py
daryakoval: Why do we need to do so many changes in the code? 
I already added main, function, and unit tests to increase coverage from 0 to 70, why more effort is needed? 
It is not related to python2 to 3 issues, And if you want the code to be as more as elegant there are a lot of changes that should be done, but these changes are not likely to happen since we have no instances.

daryakoval: @dantavori

Packs/Forcepoint/Scripts/FPDeleteRule/FPDeleteRule_test.py
daryakoval: @dantavori

daryakoval: Discussed with Dan, and He told me that a unit test for this is not needed.

dantavori (approved):

PR #21100, opened 2022-09-08
Python 2 EOL | Deprecate ArcSightXML

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Description
Updated the `ArcSightXML` Pack:

- It now uses Python 3.
- Added UTs. 

Note: Coverage is low because it misses testing of XML parsing as I could not find any available examples of payloads.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation


kgal-pan:
@bakatzir - From what I can tell from the [ArcSight ESM V2 docs](https://github.com/demisto/content/blob/master//Packs/ArcSightESM/Integrations/ArcSightESMv2/README.md), ArcSight XML is no longer supported:

> ArcSight XML is no longer supported. Use the ArcSight ESM integration instead.

Should I just deprecate this Pack instead?

bakatzir:
> @bakatzir - From what I can tell from the [ArcSight ESM V2 docs](https://github.com/demisto/content/blob/master//Packs/ArcSightESM/Integrations/ArcSightESMv2/README.md), ArcSight XML is no longer supported:
> 
> > ArcSight XML is no longer supported. Use the ArcSight ESM integration instead.
> 
> Should I just deprecate this Pack instead?

You can deprecate the entire pack and refeer the customers to ArcSight ESM, yes.

PR #21132, opened 2022-09-08
Cyber Triage --> PY3

avidan-H:
## Status
- [x] Ready

## Related Issues
Py3 Hackathon


Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
```diff
@@ -0,0 +1,192 @@
...
+USE_PROXY = CT_PARAMS['use_proxy']
+
+
+def IS_2XX(x):
+    return (x / 100) == 2  # Returns true if status code (int) is 2xx
+
+
+# Delete request proxy environment variables if user does not want to use proxy
+if not USE_PROXY:
+    del os.environ['HTTP_PROXY']
```
tkatzir, line 33: This proxy env deletion appears numerous times in the code.
I guess it belongs in CommonServerPython maybe?

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
```diff
@@ -0,0 +1,192 @@
...
+    http_status_code = -1
+    e = None
+
+    # Get the request function associate with the type of http request we want to make (ex. GET or POST)
+    try:
+        request_func = getattr(requests, method)
+    except AttributeError:
+        ret_msg = f'Invalid requests method: {method}'
+        return_error(ret_msg)
+    else:
```
tkatzir, line 71: There's no `finally` in this `try-except`, so why do we need `else`?

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
```diff
@@ -0,0 +1,192 @@
...
+            'response': ret_msg,
+            'status_code': http_status_code,
+            'exception': e
+        }
+
+# Format non 2xx responses from make_rest_call(). These are cases where an error has occurred.
+
+
+def format_err_resp(json_resp):
+    return 'Error message: {}\nHTTP Status code: {}\nExceptions: {}'.format(json_resp['response'],
```
tkatzir, line 125: Can this be changed to the `f'foo {bar}'` format?

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
```diff
@@ -0,0 +1,192 @@
...
+REST_PORT = CT_PARAMS['rest_port']
+API_KEY = CT_PARAMS['api_key']
+USER = CT_PARAMS['credentials']['identifier']
+PASSWORD = CT_PARAMS['credentials']['password']
+VERIFY_SERVER_CERT = False
+REQ_HEADERS = {'restApiKey': API_KEY}
+USE_PROXY = CT_PARAMS['use_proxy']
+
+
+def IS_2XX(x):
```
tkatzir, line 27: Why isn't this method called `is_2xx` (lowercase)?
No need to decorate it with `@staticmethod`?

tkatzir (approved):
Approved, with some comments

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
avidan-H: It's just a function - not a method of a class - therefore no need.

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
avidan-H: `finally` is intended primarily for cleanup whereas the `else` clause of a `try-except` block is logic flow where that is what happens if an error _doesn't_ occur.

Packs/CyberTriage/Integrations/CyberTriage/CyberTriage.py
avidan-H: I'll replace it with `handle_proxy` from `CommonServerPython`.

PR #21134, opened 2022-09-08
Malwr --> PY3

avidan-H:
## Status
- [x] Ready

## Related Issues
Py3 Hackathon


Packs/Malwr/Integrations/Malwr/Malwr.py
```diff
@@ -0,0 +1,317 @@
...
+
+# The Malwar API from https://github.com/PaulSec/API-malwr.com
+
+
+class MalwrAPI:
+    """
+        MalwrAPI Main Handler
+    """
+    headers = {
+        'User-Agent': "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:41.0) "
```
tkatzir, line 43: Can this be a constant?
It's pretty old (Firefox 41.0 is several years old...)

Packs/Malwr/Integrations/Malwr/Malwr.py
```diff
@@ -0,0 +1,317 @@
...
+        """Request url and return the Beautifoul Soup object of html returned"""
+        if not url:
+            url = self.url
+
+        req = self.session.get(url, headers=self.headers)
+        soup = BeautifulSoup(req.content, "html.parser")
+        return soup
+
+    @staticmethod
+    def find_submission_links(req):
```
tkatzir, line 86: Can we ad unit tests? Just to make sure the regex is good.

tkatzir (approved):
Approved, but see 2 minor comments.

Packs/Malwr/Integrations/Malwr/Malwr.py
avidan-H: I'll make it a class attribute.

Packs/Malwr/Integrations/Malwr/Malwr.py
avidan-H: On it.

avidan-H:
This PR will fail on having a `sleep` statement in it. I conferred with @yaakovi on whether it required factoring the statement out and the decision was that it was unnecessary to remove in this case since it was <= 10 seconds. (It may also fail on the image being too small ¯\_(ツ)_/¯ ). So this PR will require a force merge.

avidan-H:
I'm deprecating the integration since the product no longer seems to exist (there no longer seems to be a site hosted at malwr.com). Since we're deprecating the integration the failing `Run Unit Testing And Lint` check is irrelevant IMO. So this PR will require a force merge. FYI @tkatzir.

PR #21176, opened 2022-09-12
TestCollect: fix Path syntax 

dorschw:
fix the [issue](https://github.com/demisto/content-private/runs/8306147119?check_suite_focus=true#step:30:51) found in [this PR](https://github.com/demisto/content-private/pull/623#pullrequestreview-1104141802)


omerKarkKatz (approved):
very good job!!

PR #21179, opened 2022-09-12
Ansible tower - ad-hoc command launch playbook fixes

idovandijk:
## Status
Ready

## Related Issues
Fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4090
Related to: https://jira-hq.paloaltonetworks.local/browse/CIAC-3628

## Description
- Fixed a bug where the playbook input for arguments of the ad-hoc command were not used.
- Fixed a bug where the configuration for the polling playbook was not passed from the playbook inputs.

## Screenshots

5.0.0

No


Packs/AnsibleTower/Playbooks/Ansible_Tower-_ad_hoc_command-_generic_polling.yml
```diff
@@ -302,7 +307,7 @@ inputs:
   playbookInputQuery:
 - key: Timeout
   value:
-    simple: "15"
+    simple: "10"
```
tomer-pan, line 310: @idovandijk Why did you change the timeout from 15 to 10 as the default value?

idovandijk: The playbook input was set to 15 but the task was hardcoded to use 10, so in order to preserve BC, I changed the input to be the same value that previously was hardcoded (10)

idovandijk:
@tomer-pan need ur approval again

tomer-pan (approved):

PR #21197, opened 2022-09-13
TestCollect: Skip TPBs using skipped integrations

dorschw:
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4003


Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -150,9 +158,16 @@ def _validate_args(pack: Optional[str], test: Optional[str], reason: CollectionR
                     raise ValueError(f'{test} has no path')
                 if PACK_MANAGER.is_test_skipped_in_pack_ignore(playbook_path.name, pack_id):
                     raise SkippedTestException(test, skip_place='.pack_ignore')
+                for integration in test_playbook.implementing_integrations:
```
BEAdi, line 161: I wonder if this should be taken from the id set and not from the test configuration in `conf.json`, by using `conf.get_test(test).integrations`. Do you think they should always be matching and therefore it will not be confusing?

BEAdi (commented):
Looks great!

Tests/scripts/collect_tests/collect_tests.py
dorschw: the values under conf.json shows what integrations are tested, while the id-set has all integrations used in the TPB, and I think the latter is preferred

BEAdi (approved):

PR #21213, opened 2022-09-13
Fix Phishing Campaign Test

DeanArbel:
## Status
- [x] Ready

## Description
Fix Phishing campaing test playbook.


idovandijk (approved):
thanks prince

@tomer-pan fyi he is fixing the phishing campaign test and also the bug where we don't skip the Gmail subplaybook if unavailable when we get the original email

PR #21232, opened 2022-09-14
Add Vendor as Key Field for CommonServerPython's Endpoint

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Description
This PR adds the `Vendor` field as a key when comparing Endpoint objects if it exists.


yaakovi (commented):
* Why did you modify: `Packs/ApiModules/Scripts/CoreIRApiModule/CoreIRApiModule.py`? revert if not needed.
* add release notes to the `Base` pack

MichaelYochpaz:
@yaakovi I had flake8 linting errors for the `CoreIRApiModule` pack, but it seems that it was only the test tile that required a fix to pass the CI linting step.
I've removed the changes and added release notes.

yaakovi (approved):

PR #21244, opened 2022-09-14
Added GetEntries

spearmin10:

## Status
- [ ] In Progress
- [x] Ready

## Description
A wrapper script of `getEntries` which is a build-in function. This collects entries from the war room with conditions given.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


spearmin10:
> Hi @spearmin10, thank you for your contribution. Can you please elaborate, what is the purpose of this script? what are the use cases? Also, what do you mean by incident id as an argument, I encountered the attached problem when trying to execute the script with an existing incident I have in my localhost. Screen Shot 2022-09-14 at 18 05 09

spearmin10:
@YuvHayun How did you import the script? I'm not sure why the error raises because The line 32 is below. I think that the script imported is something different.
image

spearmin10:
@YuvHayun It's a simply wrapper of buildin `getEntries`, so there are various use cases. In my use case, I wanted to collect tagged entries with specific tags to extract indicators from them.

YuvHayun:
> @YuvHayun How did you import the script? I'm not sure why the error raises because The line 32 is below. I think that the script imported is something different. image

Not sure why it's related to the imports, I ran the same command the same way twice with two different incident ids that I have both in my incidents tab - one failed and one didn't.

spearmin10:
@YuvHayun I've fixed the issue. Please review again.

Packs/CommonScripts/Scripts/GetEntries/GetEntries_test.py
```diff
@@ -0,0 +1,74 @@
+from CommonServerPython import *
```
YuvHayun, line 1: Let's add "given-when-then" documentation to this tests cases.

YuvHayun (changes requested):
Hey @spearmin10, thx for the fixes.
I added 1 minor comment.
Also, I still get the same failures when running the same command as I showed you on our call last Thursday.
Also, I think there's something inefficient in the way we print so many msgs to the war room.
Please consider changing the way the human-readable is printed to the war-room.

Packs/CommonScripts/Scripts/GetEntries/GetEntries_test.py
spearmin10: @YuvHayun Could you paste the screenshot of what the error raised?

YuvHayun (approved):

YuvHayun:
@spearmin10 merging this pr.
Thx for your contribution!

PR #21288, opened 2022-09-18
Fixed issues where None is not evaluated properly.

spearmin10:

## Status
- [ ] In Progress
- [x] Ready

### If-Then-Else
Fixed an issue where "None" is set when `stringify` is `true`.

The condition below returns `not equal` in the current behavior, but it should return `equal`.
image
image

### SetAndHandleEmpty
Fixed an issue where `an empty string == None` is evaluated as False.

Running `SetAndHandleEmpty` with the parameters below shouldn't set value to the key because the value is empty. But it actually sets `None` to the key in the current behavior.
image
image

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/ReleaseNotes/1_7_57.md
+++ b/Packs/CommonScripts/ReleaseNotes/1_7_57.md
@@ -1,0 +1,7 @@
+
+#### Scripts
+##### SetAndHandleEmpty
+- Fixed an issue where "None" is set when `stringify` is `true`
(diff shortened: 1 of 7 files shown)


YuvHayun (approved):

YuvHayun:
@spearmin10 
PR approved, thx for your contribution!

PR #21322, opened 2022-09-19
[PAN-OS] - remove "PAN-OS Create Or Edit Rule Test" TP

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Description
The `PAN-OS Create Or Edit Rule Test` test-playbook should not be executed on its own, its  a sub-playbook for the `palo_alto_panorama_test_pb` test-playbook which will soon be updated to be executed without this playbook.
This test fails on nightly because of the following error:
```
There are 9 instances of Panorama, please select one of them by using the instance_name argument in conf.json. The options are.....:
```

it also fails in the gitlab build for any pan-os build ([example](https://code.pan.run/xsoar/content/-/jobs/16638505))

### DIFF
--- a/Tests/conf.json
+++ b/Tests/conf.json
@@ -3996,13 +3996,6 @@
             ]
         },
         {
-            "playbookID": "PAN-OS Create Or Edit Rule Test",
-            "fromversion": "6.1.0",
-            "integrations": [
-                "Panorama"
-            ]
-        },
-        {
             "playbookID": "GoogleCloudSCC-Test",
             "fromversion": "5.0.0",
             "integrations": [


JudahSchwartz (approved):

JudahSchwartz:
why are we deleting this?

GuyAfik:
> why are we deleting this?

because its being used as a sub-playbook for the `palo_alto_panorama_test_pb`, it should not run on its own, it should run only when the `palo_alto_panorama_test_pb` gets executed. In the near future it will be deprecated, i am working to enhance the test-playbooks of pan-os

PR #21437, opened 2022-09-22
XDR expired IOCs

BEAdi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-16411

## Description
Expired indicators in XSOAR are not removed from XDR when no other active indicators follow the "Sync Query" (meaning when all are expired).
What causes this is that the file sent in the "iocs_to_keep" request can't be empty. The solution is to add a char to it (this command removes all the indicators that came from XSOAR TIM and are not in the list, and doesn't add new indicators that are in the list- so we should only make sure to send something that is not an existing indicator).

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CortexXDR/ReleaseNotes/4_8_18.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### Cortex XDR - IOC
+- Fixed an issue that when all indicators that follow the "Sync Query" are expired in XSOAR TIM, they will be removed from XDR IOCs.
```
DeanArbel, line 4:
```suggestion
- Fixed an issue in the command **xdr-iocs-sync** where indicators were removed from XDR when all indicators that follow the "Sync Query" expired in XSOAR TIM.
```

DeanArbel (approved):
Nice

PR #21690, opened 2022-10-12
Cortex XDR - Port Scan - Block IP V3

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-2690

## Description
Replaced "Block IP - V2" with "Block IP - V3" in the "Cortex XDR - Port Scan" playbook.

## Screenshots
Cortex_XDR_-_Port_Scan


idovandijk (changes requested):
Great job Tal!
Please change the “IP” input to the value it had before (in the v2 playbook):
image
previously it was:
image

The idea is that any external IPs detected through the hunting process, would be blocked if the user chooses to do so.
Also the new playbook has a “UserVerification” input. Since in the port scan playbook we call the subplaybook only if the user wants to **automatically** block IPs, then we should also set the new input to **False**:
image

Let me know once you fix those and I will approve :)

TalNos:
Have you checked the changes?

idovandijk (approved):
Great!

Packs/CortexXDR/Playbooks/Cortex_XDR_-_Port_Scan_README.md
```diff
@@ -61,4 +61,4 @@ There are no outputs for this playbook.
 
 ## Playbook Image
 ---
-![Cortex XDR - Port Scan](../doc_files/Cortex_XDR_-_Port_Scan.png)
\ No newline at end of file
+![Cortex XDR - Port Scan](https://raw.githubusercontent.com/demisto/content/0d732cbc4d9a4c06cb5266cca4acf5747e1663c0/Packs/CortexXDR/doc_files/Cortex_XDR_-_Port_Scan.png)
```
idovandijk, line 64: Please use relative path like it was before.
If you replaced the PNG with the new image and kept the same file name it should work fine

idovandijk (changes requested):

Packs/CortexXDR/Playbooks/Cortex_XDR_-_Port_Scan.yml
```diff
@@ -1,3264 +1,2463 @@
...
+- key: OnCall
+  value:
+    simple: "false"
+  required: false
+  description: Set to true to assign only the users that are currently on shift.
+  playbookInputQuery:
+outputs: []
+fromversion: 5.0.0
+tests:
+- Test XDR Playbook
```
idovandijk, line 2461:
```suggestion
- no test
```

idovandijk: This test was incorrectly added by the SDK, let's revert it

idovandijk (changes requested):
Great work, see comments

Packs/CortexXDR/Playbooks/Cortex_XDR_-_Port_Scan.yml
TalNos: done

idovandijk (approved):

PR #21721, opened 2022-10-13
Slack malformed users

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15577

## Description
In some cases when using a custom script, it's possible to force feed the argument a mall formed list. To reduce customer friction, we now pre-format the argument.

### DIFF
--- a/Packs/Slack/Integrations/SlackV3/SlackV3.py
+++ b/Packs/Slack/Integrations/SlackV3/SlackV3.py
@@ -2347,7 +2347,7 @@ def create_channel():
 def invite_to_channel():
     channel = demisto.args().get('channel')
     channel_id = demisto.args().get('channel_id', '')
-    users = argToList(demisto.args().get('users', []))
+    users = argToList(demisto.args().get('users', '[]').rstrip(', '))
 
     if not users:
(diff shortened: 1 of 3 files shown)


Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -2347,7 +2347,7 @@ def create_channel():
 def invite_to_channel():
     channel = demisto.args().get('channel')
     channel_id = demisto.args().get('channel_id', '')
-    users = argToList(demisto.args().get('users', []))
+    users = argToList(demisto.args().get('users', '[]').rstrip(', '))
```
yaakovi, line 2350: Why do we want to support a malformed list?

yaakovi (approved):

Packs/Slack/Integrations/SlackV3/SlackV3.py
amshamah419: :hide-the-pain: Because it's a "friction point"

PR #21724, opened 2022-10-13
Impossible Traveler - Block IP V3

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-2690

## Description
Replaced "Block IP - V2" with "Block IP - V3" in the "Impossible Traveler" playbook.

## Screenshots
Impossible_Traveler


idovandijk (changes requested):
Nice job. Please also change UserVerification to False here because the Block IP playbook is run only when blocking IPs is automatic in the impossible traveler flow

TalNos:
Also changed this, please check.

idovandijk (approved):
Well done :)

idovandijk:
> The following integrations/tests were collected by the CI build but are currently skipped. The collected tests are related to this pull request and might be critical.:
> 
> * Impossible Traveler - Test - reason: Usage limit reached (Issue 38063)

@TalNos as you can see there is a test playbook which is supposed to test the Impossible Traveler playbook. However, it looks the test is currently skipped. Let's try to unskip it. The reason that it was skipped is a usage limit issue with the Ipstack integration which the test playbook uses. Let's see if we have a working instance for that integration so we can re-enable the test.
To unskip the test we need to edit the `Content/Tests/conf.json` file:
under the `skipped_integrations`, remove the line regarding Ipstack.

PR #22084, opened 2022-11-07
tags/usecases/cateogires 366-410

bakatzir:
## Status
- [x] Ready

## Related Issues
relates to: https://jira-hq.paloaltonetworks.local/browse/CIAC-4318

### DIFF
--- a/Packs/HackerOne/pack_metadata.json
+++ b/Packs/HackerOne/pack_metadata.json
@@ -10,14 +10,9 @@
     "categories": [
         "Vulnerability Management"
     ],
-    "tags": [
-        "Security"
-    ],
-    "useCases": [
-        "Vulnerability Management"
-    ],
+    "tags": [],
+    "useCases": [],
     "keywords": [
-        "HackerOne",
         "reports",
         "vulnerability"
     ],
--- a/Packs/GSuiteAdmin/pack_metadata.json
+++ b/Packs/GSuiteAdmin/pack_metadata.json
@@ -8,7 +8,7 @@
     "email": "",
     "created": "2020-08-14T10:49:54Z",
     "categories": [
-        "IT Services"
+        "Cloud Services"
     ],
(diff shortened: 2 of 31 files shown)


Packs/HackerOne/pack_metadata.json
```diff
@@ -10,14 +10,9 @@
     "categories": [
         "Vulnerability Management"
     ],
-    "tags": [
```
martijnrusschen, line 13: Why are these removed?

Packs/HackerOne/pack_metadata.json
```diff
@@ -10,14 +10,9 @@
...
-        "Security"
-    ],
-    "useCases": [
```
martijnrusschen, line 16: Same here

Packs/HackerOne/pack_metadata.json
```diff
@@ -10,14 +10,9 @@
...
-        "Vulnerability Management"
-    ],
+    "tags": [],
+    "useCases": [],
     "keywords": [
-        "HackerOne",
```
martijnrusschen, line 20: Why is this removed?

Packs/HackerOne/pack_metadata.json
bakatzir: We are revamping the entire tags and use cases sections for the XSOAR and XSIAM Marketplaces, as they have had too many options that have cluttered the search options.

Packs/HackerOne/pack_metadata.json
bakatzir: keywords are added to the pack name words, so it is a redundancy.

bakatzir: BTW, the pack description is also used in the search bar.

GuyLibermanPA (approved):

PR #22167, opened 2022-11-11
Contrib/CDS Illumio

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Docker Image PR: https://github.com/demisto/dockerfiles/pull/9453

- Retrieves the traffic analysis of a particular port/protocol.
- Creates segmentation rule-sets for workloads.
- Create enforcement boundaries for micro-segmentation.
- These steps collectively provide the functionality of blocking a particular port.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


crestdatasystems:
Unit tests are failing because DockerImage https://github.com/demisto/dockerfiles/pull/9453 is not merged,

kgal-pan:
Unit test failing because of non-existant Docker image. @esharf - Please have a look at https://github.com/demisto/dockerfiles/pull/9453.

kgal-pan:
@crestdatasystems  - Please fill in contribution form. Your partner ID is `illumio-10397`. Also, please fix conflict in test conf.

crestdatasystems:
@esharf @kgal-pan I've updated contribs form and resolved the conflicts.

Can you please proceed further in the review as this is stale for 10 days?

esharf:
@crestdatasystems I'm in the middle of the review of this PR.
Hopefully, I will finish it today or tomorrow.
sorry for the inconvenience.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+
+        base_url = params.get("url").strip()
+        if not base_url:
+            raise ValueError("Server URL is required.")
+
+        proxy = handle_proxy()
+
+        client = PolicyComputeEngine(url=base_url, port=port, org_id=org_id)
+        client.set_proxies(
+            http_proxy=proxy.get("http", None), https_proxy=proxy.get("https", None)
```
esharf, line 1199: - if None is your default value then you don’t need to put it here (`{'a': 'b'}.get('c')` is equal to `{'a': 'b'}.get('c', None)`.
- more important is that in case the proxy is not settled, `handle_proxy()` will return `{'http': '', 'https': ''}` so you need to check it because it might lead to problems in case the SDK checks it explicitly against `None`.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+                "illumio-rule-create": rule_create_command,
+            }
+            if command in illumio_commands:
+                args = demisto.args()
+                remove_nulls_from_dictionary(trim_spaces_from_args(args))
+                return_results(illumio_commands[command](client, args))
+            else:
+                raise NotImplementedError("Command {} is not implemented".format(command))
+    except Exception as e:
+        demisto.error(traceback.format_exc())
```
esharf, line 1227:
```suggestion
```
`return_error` already takes care about it.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.yml
```diff
@@ -0,0 +1,1659 @@
...
+      secret: false
+    - auto: PREDEFINED
+      default: false
+      defaultValue: TCP
+      description: Communication protocol.
+      isArray: false
+      name: protocol
+      predefined:
+      - TCP
+      - UDP
```
esharf, line 198: you are using upper case in a part of the time a lower in other.
consider to you a single standard here.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.yml
```diff
@@ -0,0 +1,1659 @@
...
+    - contextPath: Illumio.Rule.ingress_services.href
+      description: Array of service URI and port/protocol combinations.
+      type: String
+    - contextPath: Illumio.Rule.resolve_labels_as.providers
+      description: Providers resolve labels.
+      type: String
+    - contextPath: Illumio.Rule.resolve_labels_as.consumers
+      description: Consumers resolve labels.
+      type: String
+  dockerimage: dev-illumio:1.1.2
```
esharf, line 1645: please use the demisto docker image

Packs/Illumio/.pack-ignore
```diff
@@ -0,0 +1,2 @@
+[file:IllumioCore_image.png]
+ignore=IM111
```
esharf, line 2: please try to fix it, it shouldn’t be something hard to fix. all you need is probably just reaching one of your designers. it should be very easy for them to fix it.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+                "illumio-traffic-analysis": traffic_analysis_command,
+                "illumio-virtual-service-create": virtual_service_create_command,
+                "illumio-service-binding-create": service_binding_create_command,
+                "illumio-object-provision": object_provision_command,
+                "illumio-workload-get": workload_get_command,
+                "illumio-workloads-list": workloads_list_command,
+                "illumio-enforcement-boundary-create": enforcement_boundary_create_command,
+                "illumio-enforcement-mode-update": update_enforcement_mode_command,
+                "illumio-ip-list-get": ip_list_get_command,
+                "illumio-ip-lists-get": ip_lists_get_command,
```
esharf, line 1216: what is the difference between this 2 commands?

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+    href = args.get("href")
+    validate_required_parameters(href=href)
+
+    response = client.workloads.get_by_reference(href)  # type: ignore
+    results = json.loads(json.dumps(response, cls=IllumioEncoder))
+
+    readable_output = prepare_workload_get_output(results)
+
+    return CommandResults(
+        outputs_prefix="Illumio.Workload",
```
esharf, line 836:
```suggestion
        outputs_prefix="Illumio.Workloads",
```
take into account that the command might run multiple times and it might contain multiple results (yiu already have a path named `Illumio.Workloads"`
please fix it in the yml and readme files as well.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+    href = args.get("href")
+    validate_required_parameters(href=href)
+
+    response = client.ip_lists.get_by_reference(href)
+    results = json.loads(json.dumps(response, cls=IllumioEncoder))  # convert the ip_lists objects
+
+    readable_output = prepare_ip_list_get_output(results)
+
+    return CommandResults(
+        outputs_prefix="Illumio.IPList",
```
esharf, line 1016:
```suggestion
        outputs_prefix="Illumio.IPLists",
```

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+            resolve_consumers_as=resolve_consumers_as,
+            resolve_providers_as=resolve_providers_as
+        )
+        response = client.rules.create(rule, parent=ruleset_href)
+
+    response = response.to_json()
+    readable_output = prepare_rule_create_output(response)  # type: ignore
+
+    return CommandResults(
+        outputs_prefix="Illumio.Rule",
```
esharf, line 1161:
```suggestion
        outputs_prefix="Illumio.RuleSet",
```
not sure about this one

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+                "illumio-service-binding-create": service_binding_create_command,
+                "illumio-object-provision": object_provision_command,
+                "illumio-workload-get": workload_get_command,
+                "illumio-workloads-list": workloads_list_command,
+                "illumio-enforcement-boundary-create": enforcement_boundary_create_command,
+                "illumio-enforcement-mode-update": update_enforcement_mode_command,
+                "illumio-ip-list-get": ip_list_get_command,
+                "illumio-ip-lists-get": ip_lists_get_command,
+                "illumio-ruleset-create": ruleset_create_command,
+                "illumio-rule-create": rule_create_command,
```
esharf, line 1218: what is the difference between this 2 commands?

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
```diff
@@ -0,0 +1,1232 @@
...
+        if command == "test-module":
+            return_results(test_module(client))
+        else:
+            illumio_commands = {
+                "illumio-traffic-analysis": traffic_analysis_command,
+                "illumio-virtual-service-create": virtual_service_create_command,
+                "illumio-service-binding-create": service_binding_create_command,
+                "illumio-object-provision": object_provision_command,
+                "illumio-workload-get": workload_get_command,
+                "illumio-workloads-list": workloads_list_command,
```
esharf, line 1212: what is the difference between this 2 commands?

esharf (changes requested):
some common comments:
- check for None before comparing optional variables with int variables (`None > 1` will raise a `TypeError`).
- please use `f'string: {s}` rather than `'string: {}'.format(s)`.
- try to use `dict_safe_get(data, ['a', 'b'])` rather than `data.get('a', {}).get('b')`.

esharf:
@idovandijk can you check the playbooks here?

Packs/Illumio/Integrations/IllumioCore/IllumioCore.yml
kgal-pan, line 1645:
```suggestion
  dockerimage: demisto/illumio:1.0.0.37679
```

idovandijk (commented):
@crestdatasystems thank you for the contribution.
I've dug into the documentation of the Illumio Core product to understand some of the terms that you are using. I'm basing my review on my understanding which may not be perfect, please correct me if I got something wrong.

### Rapid ransomware containment - Illumio
1. You have a task where you're getting all workloads, and then updating their enforcement mode to "selective". According to the documentation, only managed workloads can be modified to be in selective enforcement mode, but in your playbook you're not checking whether the workloads you found are managed before trying to change them to selective enforcement mode:
image
2. If I understand correctly, if the traffic is allowed, then you are getting the workloads which were communicating in the said port and protocol. Then you are creating a virtual service which will serve as a label for the allowed service. Then you bind that service to the workloads where the traffic should be allowed. What I don't understand is the part where you're later creating a ruleset and a rule which are then applied to any IP. Can you clarify the objective of those tasks?

### Update enforcement mode - Illumio
3. Why is this playbook using global context + loops? Can it use inputs and outputs instead? Loops should be used when the subplaybook should make different decisions for each individual input, but it seems in this case that the playbook is linear and doesn't require looping. The downside of not using inputs is that this playbook is now hardcoded to update the enforcement mode to "selective" and won't be reusable for other use-cases.

After answering these points - can you suggest time windows for us to have a demo of the contribution? I'd like to understand better how Illumio works and the exact flow of the playbook.

Let me know if you have any questions.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
crestdatasystems: > not sure about this one

Rule and ruleset are different so won't be possible to update this change

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
crestdatasystems: Make sense, updated the same.

Tests/conf.json
```diff
@@ -121,7 +121,7 @@
         },
```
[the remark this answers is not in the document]
crestdatasystems: Done

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
crestdatasystems: One will get the list of all workloads based on the arguments provided. Other will get the specific workload based on HREF provided.

Packs/Illumio/Integrations/IllumioCore/IllumioCore.py
crestdatasystems: Ruleset is the superset where all the rules will be linked, where the rule is specific with more details. All rules should be linked to any ruleset.

crestdatasystems:
@esharf We've made all the changes suggested by you.

regarding the common comments, 
1. Client asked us specifically to not use the f-string. 
2. dict_safe_get is actually good idea that can be taken up in consideration for the upcoming integrations. However, RN changing data.get to dict_safe_get will result in a good amount of rework and may introduce regression so will skip this for this particular release.

crestdatasystems:
@esharf @idovandijk, We've observed that the unit test stage of CI trying to test the "test_module" function of the main file which results in the error which is expected because it is not the test case.

Can you please check and help why this has been changed previously it was not checking the "test_module" function of the integration script file.

https://app.circleci.com/pipelines/github/demisto/content/210043/workflows/afbaa700-a2f4-4de7-99df-d8357833b43b/jobs/603577/tests#failed-test-0

crestdatasystems:
@idovandijk For the playbook-related queries please find my answers.

1. I'll check for the code and API docs and will get back to you.

2. First we're getting traffic for any specific port then we're blocking all the traffic. Now if the traffic allowed is yes then it will create a rule and ruleset which will allow traffic for specific period of time. 

3. This playbook is part of the Main playbook(Rapid ransomware containment - Illumio). The purpose of this playbook is to loop certain tasks to update workloads enforcement mode. Also, we need certain exit conditions to stop the looping. To check that condition in the Main playbook, we are not able to access subplaybook's context that's why we have used global context.


Let me know if you have any other doubts or concerns.

esharf:
@crestdatasystems regarding the `test_nodule` problem in the unit tests.
this happened because pytest recognized this function as a test because it has a `test_` prefix please change it to something like `command_test_module`.
secondly please check my comments again there are a few comments you didn't address, please check them.
I’m available in [DFIR](https://dfircommunity.slack.com/team/UQWL4FQJ0) if you need any help with this.

crestdatasystems:
@esharf I've updated the image and addressed all the comments.

@esharf @idovandijk we're available to take a call tomorrow any time between 11:00 AM to 6 PM IST. Please let me know when we can have a call to complete this PR.

idovandijk:
Hi @crestdatasystems, I will be out of office for the next week. Since I don't want to have to delay this any further I will kindly ask that you let @efelmandar take the review from here. I will assign him as a reviewer.

crestdatasystems:
@esharf @efelmandar, let me know when we can schedule the demo for Illumio. We're available to have a demo call today or tomorrow anytime between 11:00 AM to 6 PM IST.

efelmandar:
@crestdatasystems Thank you again for the demo, I see that you finished fixing the playbooks according to the comments so we can go ahead with approving the PR, @esharf FYI

esharf (approved):

PR #22189, opened 2022-11-13
Resolve dependencies in order to hide d2

DinaMeylakh:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-4539)

## Description

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


DinaMeylakh:
Had to change the script id and filename from TestRegPathReputation to TestRegPathReputationBasicLists in order to not fail on the validation about changing the pack name

bakatzir:
As a lesson learned, asking @dantavori to review.

context - D2 i deprecated as of 6.6, we want to hide the pack from the MPs.

dantavori (approved):

DinaMeylakh:
It seems like if there is a dependency like we saw, the pipeline will still fail because the dependency search there is being searched against the marketplace itself and not the current branch changes. So in order to hide d2, I'll need to first update the marketplace to have no dependency issues and only then I could hide d2. 

So in this pr I will update the dependencies and open another one after this is merged to hide d2.

DinaMeylakh:
@dantavori The pipeline fails on `No such file or directory: 'Packs/D2/Integrations'`
https://code.pan.run/xsoar/content/-/jobs/18429176

Maybe a force merge is in order?

PR #22133, opened 2022-11-09
add-CortexAttackSurfaceManagement-pack

bakatzir:
## Status
- [x] Ready

## Description
Add CortexAttackSurfaceManagement-pack per @andrew-expanse and gal zilberstein approval.
Also - alphabetize the list.

PR #22190, opened 2022-11-13
Revert "add-CortexAttackSurfaceManagement-pack" - HOLD

bakatzir:
Reverts demisto/content#22133


ShahafBenYakir (approved):

yaakovpraisler (changes requested):
The `Rapid7_Nexpose` dependency was added [here](https://github.com/demisto/content/pull/22177) to fix the core pack dependencies issue.
If we are reverting the pack `CortexAttackSurfaceManagement` from being a core pack then we should consider to revert the dependency as well.

PR #22249, opened 2022-11-15
Pre update fix

daryakoval:
## Status
- [ ] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4487

   - [ ] Yes
       - Further details:
   - [x] No
  

When adding xsiam content item (for example XSIAM Event Collector) to the existing XSOAR-only pack, we are adding "marketplacev2" to the pack's metadata marketplace field.
xsiam-server step fails on pre-update since this pack does not exist in the marketplace-v2 production bucket.

The same issue happens when adding xsoar marketplace to xsiam-only packs. (xsoar server fails on pre-update since the pack does not exist in xsoar prod bucket).

This causes unnecessary Force Merges, that can be avoided.

Solution:
Check if a new marketplace was added to the pack's metadata marketplaces field, if yes skip pre-update installations on the build step, for the new marketplace.

## Must have
- [x] Tests
- [ ] Documentation


Tests/configure_and_test_integration_instances.py
```diff
@@ -270,6 +275,19 @@ def test_integration_with_mock(self, instance: dict, pre_update: bool):
...
+        Args:
+            diff: the git diff for pack_metadata file, between master and branch
+        Returns:
+            (bool): whether new (current) marketplace was added to the pack_metadata or not
+        """
+        spaced_diff = " ".join(diff.split())
+        if (f'+ "{self.marketplace_name}"' in spaced_diff) and not (f'- "{self.marketplace_name}"' in spaced_diff):
+            return True
+        else:
+            return False
```
dantavori, line 289:
```suggestion
        return (f'+ "{self.marketplace_name}"' in spaced_diff) and not (f'- "{self.marketplace_name}"' in spaced_diff)
```

dantavori (approved):

PR #22323, opened 2022-11-17
Added CheckIndicatorValue and Indicator Registration Polling - Generic

spearmin10:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
Indicators inserted/updated to the Threat Intel DB by the auto extractor are asynchronously executed even in in-line mode. So it takes a time to get them after that, and tasks cannot get them soon after the auto extraction task.
The playbook of `Indicator Registration Polling - Generic` with `CheckIndicatorValue` script waits for indicators available on the DB using the Generic Polling.

## Short Demo Video
https://drive.google.com/file/d/174y24HpjjLGVBZAtnl8y4-Ei_X2oqulA/view?usp=share_link

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CommonScripts/Scripts/CheckIndicatorValue/CheckIndicatorValue.py
+++ b/Packs/CommonScripts/Scripts/CheckIndicatorValue/CheckIndicatorValue.py
@@ -1,0 +1,60 @@
+import base64
+import urllib.parse
+from typing import Any, Dict, Optional
+
+import demistomock as demisto  # noqa: F401
+from CommonServerPython import *  # noqa: F401
+
+
+def exists_indicator(indicator: str) -> bool:
+    if contents := execute_command('getIndicator', {'value': indicator}):
+        if len(contents) > 0 and contents[0].get('value') in [indicator]:
+            return True
+    return False
+
+
+def decode_indicator(indicator: str, encoding: Optional[str]) -> str:
+    if not encoding or encoding == 'none':
+        return indicator
+    elif encoding == 'base64':
+        return base64.b64decode(indicator.encode()).decode()
+    elif encoding == 'url-encoding':
+        return urllib.parse.unquote(indicator)
+    else:
+        raise DemistoException(f'Unknown encoding mode: {encoding}')
+
+
+def check_indicators(indicators: List[str], encoding: str) -> List[Dict[str, Any]]:
+    # Decode and dedup indicators
+    pairs = {decode_indicator(encoded_indicator, encoding): encoded_indicator for encoded_indicator in indicators}
+
+    # Check if each indicator exists
+    return [{
+        'Indicator': indicator,
+        'EncodedIndicator': encoded_indicator,
+        'Exists': exists_indicator(indicator)
+    } for indicator, encoded_indicator in pairs.items()]
+
+
+def main():
+    try:
+        args = demisto.args()
+        encoding = args.get('encoding') or 'none'
+        indicators = argToList(args.get('indicator') or [])
+
+        outputs = check_indicators(indicators, encoding)
+        count = sum(1 for ent in outputs if ent.get('Exists'))
+
+        return_results(CommandResults(
+            outputs_key_field='Indicator',
+            outputs_prefix='CheckIndicatorValue',
+            outputs=outputs,
+            readable_output=f'{count} indicators exist.',
+            raw_response=outputs
+        ))
+    except Exception as e:
+        return_error(f'Failed to execute CheckIndicatorValue. Error: {str(e)}')
+
+
+if __name__ in ('__main__', '__builtin__', 'builtins'):
+    main()
--- a/Packs/CommonPlaybooks/Playbooks/playbook-Indicator_Registration_Polling_-_Generic.yml
+++ b/Packs/CommonPlaybooks/Playbooks/playbook-Indicator_Registration_Polling_-_Generic.yml
@@ -1,0 +1,170 @@
+contentitemexportablefields:
+  contentitemfields:
+    propagationLabels:
+    - all
+fromversion: 5.0.0
+description: This playbook polls all indicators to check if they exist.
+id: Indicator Registration Polling - Generic
+inputs:
+- description: The indicators to check
+  key: indicators
+  playbookInputQuery: null
+  required: true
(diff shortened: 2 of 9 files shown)


ERNIERENDON2023:
Imthetruth 2

ERNIERENDON2023:
`I’m the truth 2 ``

idovandijk:
Thank you @spearmin10 for the contribution. This PR is intended as a workaround for a platform issue that you're experiencing. Please allow me to do an internal investigation about the issue that you are originally trying to solve here, before I move on to review this PR. I will update with the findings once I have them.

Packs/CommonScripts/Scripts/CheckIndicatorValue/CheckIndicatorValue.py
```diff
@@ -0,0 +1,60 @@
...
+    return False
+
+
+def decode_indicator(indicator: str, encoding: Optional[str]) -> str:
+    if not encoding or encoding == 'none':
+        return indicator
+    elif encoding == 'base64':
+        return base64.b64decode(indicator.encode()).decode()
+    elif encoding == 'url-encoding':
+        return urllib.parse.unquote(indicator)
```
JasBeilin, line 22: How should the customers know which encoding to use?
Also, please use our formatters, as it might cause bugs when we save something one way, but look it up another way.
You have FormatURL Script for urls, for example.

JasBeilin (changes requested):
Great contribution, thank you! Could you send demonstration video? (slack is good)

idovandijk (changes requested):
Hi @spearmin10, thanks for contributing this content. After checking internally - it seems that indeed the indicator **extraction** is a separate event than the **creation** of the indicator in the database, which makes this contribution a very useful one.

I do have some comments about this though:
1. What is purpose of the URLEncode transformer here?
image
2. Why are we outputting the **encoded** indicator in the script CheckIndicatorValue? Isn't it supposed to be the **decoded** one?
3. The playbook `Indicator Registration Polling - Generic` is generic as the name suggests, so anything specific to URLs cannot be added there:
image
what you can do instead is add that as a playbook input instead, so it would be configurable for users.
4. It's problematic that you created your own logic for unquoting URLs. We have the FormatURL script for that. @JasBeilin will elaborate.

spearmin10:
@idovandijk @JasBeilin 
That's come from the design of `Generic Polling`. The `Ids` parameters given to it cannot contain commas.
The way that the playbooks/script uses encoded indivators as the polling `id` is to allow it any indicators even if they contains commas. That's not for URLs, but all the type of indicators.
e.g. Type: `Compaign` - Value: `Destructive wiper attacks against Ukraine, Latvia and Lithuania`

For example, we have the two indicators in array containing commas below.

```
[
  [0] Destructive wiper attacks against Ukraine, Latvia and Lithuania
  [1] Palo Alto Networks, Inc
]
```

If we give them to the `Ids` of  the Generic Polling without any encoding, `GenericPollingScheduledTask` called in the Generic Polling simply converts to commna separated list and gives them to the scheduled command (`CheckIndicatorValue`),

``` 
In GenericPollingScheduledTask:

    pollingCommandArgs[args.pollingCommandArgName] = idsToPoll.join(',');
    var res = executeCommand(args.pollingCommand, pollingCommandArgs);
```

so it gets the value below.
`Destructive wiper attacks against Ukraine, Latvia and Lithuania,Palo Alto Networks, Inc`

Then, by `argToList` in CheckIndicatorValue

`indicators = argToList(args.get('indicator') or [])`

it finally gets the indicators below.
```
[
  [0] Destructive wiper attacks against Ukraine
  [1] Latvia and Lithuania
  [2] Palo Alto Networks
  [3] Inc
]
```

That's not expected values, and that doesn't work well.

So, we need to use something IDs not containing commas for polling apart from indicators. That's why I've used the percent-encoded (url-encoded) indicators as polling IDs. Any encoding to encode commas is OK such as base64 as well as url-encoding. However the playbook is Generic. It's best to use an OOTB transformer to encode them. The `URLEncode` transformer is available on the brand new instance, so it would be the best encoding for it.

In that case, we give the following polling IDs by the URLEncode to `Ids` of the Generic Polling.
```
[
  [0] Destructive wiper attacks against Ukraine%2C Latvia and Lithuania
  [1] Palo Alto Networks%2C Inc
]
```

so the CheckIndicatorValue will be given the values below.
`Destructive wiper attacks against Ukraine%2C Latvia and Lithuania,Palo Alto Networks%2C Inc`

Then, by `argToList`, it finally gets the indicators below.
```
[
  [0] Destructive wiper attacks against Ukraine%2C Latvia and Lithuania
  [1] Palo Alto Networks%2C Inc
]
```

But, they are polling IDs, not indicators. That's why it decodes them internally as indicators.
```
[
  [0] Destructive wiper attacks against Ukraine%2C Latvia and Lithuania
  [1] Palo Alto Networks%2C Inc
]
```

image

idovandijk:
@spearmin10 thanks for the explanation, this is an awesome workaround. I still wonder about points 2 and 4. Waiting for your reply.

spearmin10:
> @spearmin10 thanks for the explanation, this is an awesome workaround. I still wonder about points 2 and 4. Waiting for your reply.

@idovandijk 
For 2., as I mentioned, encoded indicators are used for **polling IDs**, not indicators. To give polling IDs to the Generic Polling, that needs to give encoded indicators.

For 4., as I mentioned, url-encoding is not used for encoding/decoding URLs. So FormatURL doesn't work in this case.

Packs/CommonScripts/Scripts/CheckIndicatorValue/CheckIndicatorValue.py
spearmin10: The demo video is recored in the Contribution Registration form.

idovandijk (approved):
@spearmin10 thank you for clarifying, I also viewed the demo video - looks great.

JasBeilin (approved):

PR #22341, opened 2022-11-18
ReversingLabs A1000 2.0.7

MislavReversingLabs:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Corrected the bug with an unusuable `result-limit` argument in the `advanced_search` funcction. Now the argument is called `resultLimit`.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml
+++ b/Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml
@@ -326,7 +326,7 @@ script:
       defaultValue: '5000'
       description: Maximum number of results
       isArray: false
-      name: result_limit
+      name: resultLimit
       required: false
       secret: false
     deprecated: false
--- a/Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml
+++ b/Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml
@@ -337,7 +337,7 @@ script:
     - contextPath: ReversingLabs.a1000_advanced_search_report
       description: A1000 classification report
       type: Unknown
-  dockerimage: demisto/reversinglabs-sdk-py3:2.0.0.36560
+  dockerimage: demisto/reversinglabs-sdk-py3:2.0.0.37494
   feed: false
   isfetch: false
   longRunning: false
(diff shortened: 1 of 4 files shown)


MislavReversingLabs:
`Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml: [BC104] - Possible backwards compatibility break, Your updates to this file contains changes to a name or an argument of an existing command(s).`

I did change the command argument but it was never implented correctly in the first place so this technically not a breaking change. It never worked properly before this change.
That's why I selected that this update does not contain breaking changes. Can we ignore this check in this case?

Packs/ReversingLabs_A1000/Integrations/ReversingLabsA1000v2/ReversingLabsA1000v2.yml
```diff
@@ -326,7 +326,7 @@ script:
       defaultValue: '5000'
       description: Maximum number of results
       isArray: false
-      name: result_limit
+      name: resultLimit
```
ostolero, line 329: revert

ostolero (changes requested):
Hi @MislavReversingLabs thank you for this fix, good catch!
I think it might be better to leave the argument as is and just fix the input for the 'getArg' command. Is there another reason to change the name?
You're right that this argument was never functional because of this bug, but if a customer did use it in a playbook or script, changing the argument name might throw an error.  That's why we should try and keep this backwards compatible if possible. Let me know if these changes look good to you and we can proceed :)

ostolero (commented):
All the changes look great let's do a short demo :)

ostolero:
Hi @MislavReversingLabs,
 as this is a small fix we should just see the command works as expected with the new argument. 
We can either schedule a short call or you can record a short video of the command.
If you want to schedule a call I'll reach out through email and we can set it up.
If you prefer to send over a video make sure it includes running the command, showing the output in the war room, and viewing the context output as well.
Let me know what works best for you or if there are any questions :)
Thanks

MislavReversingLabs:
> 

Great. I'll record a video with all of the mentioned steps and send the link to you via Slack.

MislavReversingLabs:
@ostolero Video sent to Slack.

ostolero (approved):
Demo was great, we can merge 👍

PR #22394, opened 2022-11-21
Added LDAP Authentication (OpenLDAP) Test to the conf.json

ShacharKidor:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [XSUP-14831](https://jira-hq.paloaltonetworks.local/browse/XSUP-14831).
related MR: [1564](https://code.pan.run/xsoar/content-test-conf/-/merge_requests/1564).
related MR: [1572](https://code.pan.run/xsoar/content-test-conf/-/merge_requests/1572).

## Description
- Added the 'SSL Version' integration parameter that allows the user to specify a specific version for the SSL protocol (the default value is None).

- Added the ability to run the ad-authenticate command with a username that isn't a full DN. 

- Added the new LDAP Authentication (OpenLDAP) instance to the conf.json tests, so the 'LDAP Authentication - Test' will be tested on both 'OpenLDAP' and 'Active Directory' instances. 

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0
- [ ] 6.8.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


ShacharKidor:
Tested the 'LDAP Authentication - Test' on both instances: 
OpenLDAP_-_Test_Mon_Nov_21_2022

ShacharKidor:
Hi @glicht, 
I added the fixes we talked about. 

What do you think about the security issue found [here ](https://github.com/demisto/content/pull/22394#issuecomment-1324104570)? those SSL versions are old, but we added them intentionally to allow customers to use them if they want/need to. Should we ignore the security warning? 

I haven't completely finished the PR cause I would like to add a few unit tests to increase the integration's coverage (originally it didn't have any), but you can review the code in the meantime.

Thanks.

Packs/OpenLDAP/Integrations/OpenLDAP/OpenLDAP.py
```diff
@@ -30,6 +30,15 @@ class LdapClient:
...
     CIPHERS_STRING = '@SECLEVEL=1:ECDHE+AESGCM:ECDHE+CHACHA20:DHE+AESGCM:DHE+CHACHA20:ECDH+AESGCM:DH+AESGCM:' \
                      'ECDH+AES:DH+AES:RSA+ANESGCM:RSA+AES:!aNULL:!eNULL:!MD5:!DSS'  # Allowed ciphers for SSL/TLS
+    SSL_VERSIONS = {
+        'SSLv23': ssl.PROTOCOL_SSLv23,
+        'TLSv1': ssl.PROTOCOL_TLSv1,
+        'TLSv1_1': ssl.PROTOCOL_TLSv1_1,
+        'TLSv1_2': ssl.PROTOCOL_TLSv1_2,
+        'TLS': ssl.PROTOCOL_TLS,
+        'TLS_CLIENT': ssl.PROTOCOL_TLS_CLIENT,
+        'TLS_SERVER': ssl.PROTOCOL_TLS_SERVER
```
glicht, line 40: We don't need this one. It is for SERVER only.

ShacharKidor: Removed it. Thanks.

glicht (approved):

Packs/OpenLDAP/Integrations/OpenLDAP/OpenLDAP.py
```diff
@@ -108,12 +131,12 @@ def _get_tls_object(self):
             # Trust any certificate = False means that the LDAP server's certificate must be valid -
             # i.e if the server's certificate is not valid the connection will fail.
             tls = Tls(validate=ssl.CERT_REQUIRED, ca_certs_file=os.environ.get('SSL_CERT_FILE'),
-                      version=ssl.PROTOCOL_TLS)
+                      version=self._get_ssl_version())
```
JudahSchwartz, line 134: we dont want  ssl.PROTOCOL_TLS to be the default? We're passing None as default here

Packs/OpenLDAP/Integrations/OpenLDAP/OpenLDAP.yml
```diff
@@ -85,6 +85,19 @@ configuration:
   - Start TLS
   required: true
   type: 15
+- additionalinfo: The SSL\TLS version to use in SSL or Start TLS connections. Default is None.
```
JudahSchwartz, line 88: Default is None doesnt really explain what will happen when I dont set this param. None should be part of the dropdown and default. We should give it a better name than None

JudahSchwartz (approved):
Nice! 
I think we should clarify what choosing None means and add it as an option. 
Great work on this!

Packs/OpenLDAP/Integrations/OpenLDAP/OpenLDAP.py
ShacharKidor: No, I consulted with @glicht and we decided that None will be the default. 
Although I think that eventually None = PROTOCOL_TLS = PROTOCOL_SSLv23 (at least that is what I understood from the ssl code).

Packs/OpenLDAP/Integrations/OpenLDAP/OpenLDAP.yml
ShacharKidor: Thank you.
I improved the description. Added a recommendation to use the TLS_CLIENT version with a link to a reference. 
Please Let me know what you think.

PR #22421, opened 2022-11-22
Maltiverse fix bc

Shellyber:
## Status
- [x] Ready

## Description
The domain command is not returning the resolvedIP anymore in the new API version 1.1
Removed the outputs 

   - [x] Yes


Shellyber:
Need to force merge this PR since we're breaking BC (intentional) and validation fails on it.

PR #22441, opened 2022-11-23
AWS - Update the credentials

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
Relates: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-4586)

## Description
Added credentials for the *Access Key* and *Secret Key*.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [ ] Documentation


amshamah419:
Even though we aren't breaking BC, I'm wondering if we should include a BC notice to tell customers that the credentials should be migrated to the credential store as a better security practice. @glicht WDYT?

glicht:
> Even though we aren't breaking BC, I'm wondering if we should include a BC notice to tell customers that the credentials should be migrated to the credential store as a better security practice. @glicht WDYT?

if we are not breaking, then no need.

Packs/ApiModules/Scripts/AWSApiModule/AWSApiModule.py
```diff
@@ -30,8 +40,7 @@ def __init__(self, aws_default_region, aws_role_arn, aws_role_session_name, aws_
         self.aws_role_session_duration = aws_role_session_duration
         self.aws_role_policy = aws_role_policy
         self.aws_access_key_id = aws_access_key_id
-        self.aws_secret_access_key = aws_secret_access_key
-        self.aws_session_token = aws_session_token
+        self.aws_secret_access_key, self.aws_session_token = extract_session_from_access(aws_secret_access_key, aws_session_token)
```
glicht, line 43: I think we need to fix `aws_session` function. It looks like the code won't go into the part to log in with access_key + session_token. See: https://github.com/demisto/content/blob/f24dbfef8a47c6cc2f6665dfb89dd3438af410be/Packs/ApiModules/Scripts/AWSApiModule/AWSApiModule.py#L137

mmhw: I fixed it

glicht (approved):
Looks good

Packs/ApiModules/Scripts/AWSApiModule/AWSApiModule_test.py
```diff
@@ -176,3 +176,29 @@ def test_AWSClient_without_session_token():
...
+                         [
+                             ('access_key@@@session_token', None, ('access_key', 'session_token')),
+                             ('test1', None, ('test1', None)),
+                             ('test1', 'test2', ('test1', 'test2')),
+                             ('test1@@@test2', 'test3', ('test1@@@test2', 'test3')),
+                             ('', None, ('', None)),
+                             (None, '', (None, '')),
+                             (None, None, (None, None))
+                         ])
+def test_extract_session_from_access(access_key, session_token, expected):
```
bziser, line 191: Great work!! :)

bziser (approved):

PR #22509, opened 2022-11-27
SNOW comments and work notes without sys_journal_field

darkushin:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4009

## Description
Support getting comments and work notes using `sysparm_display_value=all`. We use 'all' and not 'true' as this returns both system and display value. Using both values we can compute the timezone offset, by computing the difference between them.

   - [ ] Yes
       - Further details:
   - [x] No


darkushin:
Hi @jrosenpanw,
We will appreciate your help here. We added a new option to retrieve work notes that requires fewer permissions, i.e. the elevated access to the `sys_journal_field` table is not required anymore. We updated the Readme section that addressed it, but we want to make sure that it is clear enough.
Thank you!

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -489,6 +491,67 @@ def split_fields(fields: str = '', delimiter: str = ';') -> dict:
...
+            retrieved_last_note = False
+            continue
+        note_dict = {
+            "sys_created_on": created_on_UTC.strftime(DATE_FORMAT),
+            "value": note_value,
+            "sys_created_by": created_by,
+            "element": note_type
+        }
+        notes.append(note_dict)
+        retrieved_last_note = True
```
yuvalbenshalom, line 524: what if 2 notes were added?

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -2208,6 +2280,18 @@ def parse_dict_ticket_fields(client: Client, ticket: dict) -> dict:
...
+def get_timezone_offset(full_response):
+    """
+    Receives the full response of a ticket query from SNOW and computes the timezone offset between the timezone of the
+    instance and UTC.
+    """
+    local_time = datetime.strptime(full_response.get('result', {}).get('sys_created_on', {}).get('display_value', ''),
+                                   DATE_FORMAT)
+    utc_time = datetime.strptime(full_response.get('result', {}).get('sys_created_on', {}).get('value', ''), DATE_FORMAT)
+    offset = utc_time - local_time
+    return offset
```
yuvalbenshalom, line 2292: you should try/catch each parsing here separately and add more information to the exception that will explain what we failed to parse.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -489,6 +491,67 @@ def split_fields(fields: str = '', delimiter: str = ';') -> dict:
...
+        if 'Mirrored from Cortex XSOAR' in note:
+            if retrieved_last_note:  # add to last note only in case the note was not filtered by the time filter
+                notes[-1]['value'] += '\n\n Mirrored from Cortex XSOAR'
+            continue
+        note_info, note_value = note.split('\n')
+        created_on, created_by = note_info.split(' - ')
+        created_by = created_by.split(' (')[0]
+
+        # convert note creation time to UTC
+        created_on_UTC = datetime.strptime(created_on, DATE_FORMAT) + time_info.get('timezone_offset')
```
yuvalbenshalom, line 510: handle the aception here and return a more detailed error of what we failed to parse.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -489,6 +491,67 @@ def split_fields(fields: str = '', delimiter: str = ';') -> dict:
...
+    for note in notes_split:
+        if not note:
+            continue
+        if 'Mirrored from Cortex XSOAR' in note:
+            if retrieved_last_note:  # add to last note only in case the note was not filtered by the time filter
+                notes[-1]['value'] += '\n\n Mirrored from Cortex XSOAR'
+            continue
+        note_info, note_value = note.split('\n')
+        created_on, created_by = note_info.split(' - ')
+        created_by = created_by.split(' (')[0]
```
yuvalbenshalom, line 507: check if we failed to find the created_by

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -76,6 +76,11 @@ configuration:
   - Incoming And Outgoing
   required: false
   type: 15
+- additionalinfo: Select this checkbox to retrieve comments and work notes without accessing the `sys_field_journal` table.
```
yuvalbenshalom, line 79: we will probably need more documentation in the readme about when this is needed

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2_test.py
```diff
@@ -152,6 +153,131 @@ def test_split_fields_with_special_delimiter():
     assert False
 
 
+def test_convert_to_notes_result():
```
yuvalbenshalom, line 156: please add empty note cases

yuvalbenshalom (changes requested):
looks good. see my comments

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
darkushin: I added some more details in line 2490 in the readme, but I also asked Jane Rosen to help us here so it will be clear enough.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -489,6 +491,67 @@ def split_fields(fields: str = '', delimiter: str = ';') -> dict:
...
+def split_notes(raw_notes, note_type, time_info):
+    notes: List = []
+    notes_split = raw_notes.split('\n\n')
+    retrieved_last_note = False
+    for note in notes_split:
+        if not note:
+            continue
+        if 'Mirrored from Cortex XSOAR' in note:
+            if retrieved_last_note:  # add to last note only in case the note was not filtered by the time filter
+                notes[-1]['value'] += '\n\n Mirrored from Cortex XSOAR'
```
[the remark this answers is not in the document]
darkushin: In line 517 we create a dictionary with the details of the note and add the 'value' key to it. Then, in line 524 we set the retrieved_last_note to true, to indicate that the previous note was retrieved (and therefore the key 'value' exists).

The reason for all this is the structure of the notes that we get from SNOW: the notes appear as a single string, separated by "/n/n" to indicate different notes and by a "/n" within each note to separate between the creation time & created-by information and the actual value of the note (see ServiceNowv2_test.py line 175 for an example). However, when a note is created through XSOAR, we attach to it \n\n Mirrored from Cortex XSOAR. This makes the split to notes more difficult, so the purpose of this if is to address this case and add this message back to the note.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
darkushin: This boolean was added to handle the case of a note created in XSOAR (as I explained [here](https://github.com/demisto/content/pull/22509#discussion_r1046698364) about the structure of notes). It shouldn't be a problem even if two notes are added as we handle them in the for loop here and the '\n\n Mirrored from Cortex XSOAR' message should always be added to the previous note that was retrieved.

yuvalbenshalom (approved):

Shellyber (approved):
Looks good. 
Let's do a short demo just for the mirroring part, to check that everything works as expected

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.yml
```diff
@@ -76,6 +76,23 @@ configuration:
...
+  display: Use Display Value
+  name: use_display_value
+  required: false
+  type: 8
+- additionalinfo: 'Select the date format of your ServiceNow instance. Mandatory when using the `Use Display Value` option. To find the date format of your instance see https://docs.servicenow.com/bundle/tokyo-platform-administration/page/administer/time/task/t_PersonalizeTheSystemDateFormat.html'
```
yuvalbenshalom, line 84: please move the explanation of where to get the formats to the readme, and leave a here a direction to the place in the readme. this is because additional info is a hover popup. so it is not clickable and you cannot copy from it.

yuvalbenshalom (changes requested):
looks good, see my comment

yuvalbenshalom (approved):

PR #22518, opened 2022-11-27
change tags to pack in xsiam

JasBeilin:
## Status

- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4112

## Description
changed the xsiam data onboarding tag according to out infrastructure.
Need force as validate fails to run on that much packs.


MLainer1 (commented):
Looks good 🥇 , just make sure there are no more "Data Source" tags without the prefix in the repo

PR #22560, opened 2022-11-29
Added proxy support

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue
](https://jira-hq.paloaltonetworks.local/browse/XSUP-18853)

## Description
Added proxy support to the integration

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


ShahafBenYakir (changes requested):
Please see lines 94-95 in the python code

dansterenson:
I had issues testing it on my machine since we do not have working credentials. google-cloud-translate was updated two months ago and from what I read [here](https://github.com/googleapis/python-translate/issues/80), it should be supporting proxies.

Customer wrote to CS that "i’ve just tested it and if i set the Proxy setting, it’s working and the translation works.".

ShahafBenYakir (approved):

PR #22599, opened 2022-11-30
getFaildTasks bug fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-18384)

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/IntegrationsAndIncidentsHealthCheck/Scripts/GetFailedTasks/GetFailedTasks.py
```diff
@@ -116,9 +138,14 @@ def get_incident_data(incident: dict, rest_api_instance: str = None):
...
-    tasks = get_incident_tasks_using_rest_api_instance(incident, rest_api_instance) if rest_api_instance \
-        else get_incident_tasks_using_internal_request(incident)
+    if rest_api_instance:
+        tasks = get_incident_tasks_using_rest_api_instance(incident, rest_api_instance)
+    else:
+        try:
+            tasks = get_incident_tasks_using_internal_request(incident)
+        except Exception:
+            # if using_internal_request fails, using rest api call
+            tasks = get_incident_tasks_using_rest_api_instance(incident, rest_api_instance)
```
DeanArbel, line 148: This doesn't seem right:
1. Catching a bare Exception is too broad - do we have a specific error message/status that matches the use case of MT failure? No need to retry in other cases.
2. `rest_api_instance` isn't set here, so this shouldn't work. Use `get_rest_api_instance_to_use()` here to determine whether there's a possible fallback.

DeanArbel (changes requested):

DeanArbel (approved):
Good job

PR #22626, opened 2022-12-02
EXPANDR-89 ASM - Tenable

johnnywilkes:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
https://jira-hq.paloaltonetworks.local/browse/EXPANDR-89

## Description
Adding Tenable.io Enrichment to ASM Pack/Playbook
"Cortex ASM - Enrichment" - edited to subplaybooks
"Cortex ASM - Tenable.io Enrichment" - new: looks up information on asset, if found parses for emails to set in service owner field
"Cortex ASM - ServiceNow CMDB Enrichment" - new playbook - same behavior
"Cortex ASM - AWS Enrichment" - new playbook - same behavior

## Screenshots
Provided if necessary

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [X] 6.5.0

   - [ ] Yes
       - Further details:
   - [X] No

## Must have
- [ ] Tests
- [X] Documentation


ostolero:
Hi @johnnywilkes! thank you for this contribution, and I'm sorry this review is being delayed. 
We have quite a few cr to handle and are trying to go over them as soon as possible, I do see most of your changes are related to test playbooks and this is why @JasBeilin  has added @efelmandar from our Security team to review them. 
@efelmandar please update here after you review this or if this is being delayed. Thank you.

efelmandar:
## Review Draft
Hi @johnnywilkes, thank you for this contribution and sorry for the delayed response, I finished reviewing your PR and have some small comments and suggestions regarding the logic of the playbooks and tasks, please see them listed below for easier reference:

### Cortex ASM - Enrichment
1. From my understanding this is the main playbook which include all the individual sub-playbooks for each enrichment (Cloud AWS, ServiceNow CMDB and Tenable.io) which are not dependent on one another, while in the current logic the flow is linear and not running them in parallel (each in it's own branch), I would recommend changing the logic to run each sub-playbook on separate pathes which would make the logic simpler and allow adding other enrichments easier in the future.
image

2. After the conditional check if ASM is enabled you get the service information and than check if there was a result and if the provider is AWS in two different check, this can be done in a single task with the `AND` operator.
image

### `Cortex ASM - Tenable.io Enrichment`
1. In the first task which checks that `Tenable.io` instance is enable and we `inputs.IPAddress` is defined, the module check will result a bug since you return the module's object and not boolean value, so I would recommend to change the `Get` value to be the module's state where `modules.brand` equals to `Tenable.io` as you did but than check that `modules.state` Equals `active`.
image

johnnywilkes:
@efelmandar
Thank you for your review.  Some responses:
**Cortex ASM - Enrichment**
1. We need to run enrichment serially for now because of a limitation with the GridFieldSetup.  I found that in testing that if multiple subplaybooks were updating the same field with that automation that one of the changes could be missed/overwritten.
2. I'd prefer to keep it the way it is because we will be adding more cloud providers (GCP and Azure soon) and it makes sense to do two checks instead of one.

**Cortex ASM - Tenable.io Enrichment**
1. I don't really understand what you are asking.  This is the syntax we have used for a variety of other playbooks and verified that is doesn't cause an error.  Do you have an example of what you would prefer or would you like to meet to discuss?

efelmandar:
@johnnywilkes Thank you for clarifying these points for me and since this is the case it's ok to keep it as is.
Regarding the last point, after your response I went to double check it again in a different environment and the error must have been due to something in my env so please kindly disregard this one.

Thank you again for the contribution and hard work it really shows in your PR's!

@JasBeilin everything done and we are good to merge

JasBeilin (approved):
Thank you for the contribution, great job!

PR #22645, opened 2022-12-04
Ews v2 bug fetch

daryakoval:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-18650

## Description
The bug is that we send API call with filter by: datetime_received, but saving datetime_created for the next fetch (to fetch from datetime_created of the email).
In our environment datetime_received is the same as datetime_created, and I think in most cases it will be the same, but in the customer's environment, it is different for some reason.
In the fix, I changed the logic to save datetime_received and filter the next fetch by datetime_received too.
I will look into the logs to understand why it is still not working.

Waiting for confirmation from customer. 

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


Shellyber (commented):
Please add a test.

Packs/EWS/ReleaseNotes/2_0_2.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### EWS v2
+- Fixed an issue with ***fetch-incidents***, which caused by different created and received datetime of the incident.
```
Shellyber, line 4:
```suggestion
- Fixed an issue with the ***fetch-incidents*** command where incidents were missed due to different DateTime of the incident.
```

Shellyber (approved):
Nice!

PR #22655, opened 2022-12-04
Tor Exit feed fix

yucohen:
## Status
- [ ] In Progress
- [ ] Ready
- [x] In Hold - waiting for a session with the customer to see if this is the real cause if the bug

## Description
Added a validation on the value of the indicator. if it is not a valid IP (the feed only fetches IP type) then do not add it to the indicators list.


JudahSchwartz (commented):
this isnt waiting for me is it?

yucohen:
@JudahSchwartz No
This is probably not the right solution for the bug
closed

PR #22722, opened 2022-12-06
VMWare timeout issue fixes

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [XSUP-19058](https://jira-hq.paloaltonetworks.local/browse/XSUP-19058)

## Description
A draft pr testing the possibility the issue is related to TLS versions.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/VMware/Integrations/VMware/VMware.py
+++ b/Packs/VMware/Integrations/VMware/VMware.py
@@ -12,6 +12,15 @@ from pyVmomi import vim, vmodl  # type: ignore
 from vmware.vapi.vsphere.client import create_vsphere_client
 
 
+SSL_VERSIONS = {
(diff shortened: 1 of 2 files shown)


ShacharKidor (approved):
Looks good :) 
@YuvHayun - if this would be the final solution please add the info about the SSL version new parameter to the readme and the RN as well. 

Also worth checking what is the python default value for the SSL context and setting it as your default. (In my case it was None).

YuvHayun:
> Looks good :) @YuvHayun - if this would be the final solution please add the info about the SSL version new parameter to the readme and the RN as well.
> 
> Also worth checking what is the python default value for the SSL context and setting it as your default. (In my case it was None).

I took the one that is already configured to be the default one.

PR #21630, opened 2022-10-06
Fixed bug in GetFailedTasks script

yaakovpraisler:
## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-15983

## Description
Fixed a bug in the GetFailedTasks where the tenant name was added twice when requesting via Demisto Rest API.

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [ ] Documentation

PR #22739, opened 2022-12-06
GetFailedTasks get tenant name fix

dansterenson:
Reverted some of the code that was removed [here](https://github.com/demisto/content/pull/21630). In case the user is using RestAPI in MT env, we should get the tenant name. CS also tested this version and confirmed it was working as expected.

## Status
- [ ] In Progress
- [x] Ready

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


DeanArbel (approved):
Nicely done

PR #22857, opened 2022-12-08
Domain regex upgrade to support only natural language.

Ni-Knight:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Better regex that catches only letters and not unseen chars.
https://regex101.com/r/kIytNC/1

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


MLainer1:
Do we also want to update the domain regex in CSP?

MLainer1 (approved):
🥇

PR #22955, opened 2022-12-11
Cisco FirePower

LiorQM:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [x] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3602,842 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+    # In case import arguments weren't inserted when validate_only is false.
+    if not validate_only and not all((rule_import_mode, rule_group_ids)):
+        raise ValueError('rule_import_mode and rule_group_ids must be inserted when validate_only is "False".')
+
+    file_entry = demisto.getFilePath(entry_id)
+    filename = file_entry['name']
+    file_type = os.path.splitext(filename)[1]
+
+    if file_type not in ('.txt', '.rules'):
+        raise ValueError('Supported file formats are ".txt" and ".rules".')
```
dorschw, line 3998:
```suggestion
        raise ValueError('Supported file formats are ".txt" and ".rules", got {file_type}')
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3602,842 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+    basepolicy_id = args.get('basepolicy_id', '')
+    name = args.get('name', '')
+    description = args.get('description')
+    inspection_mode = args.get('inspection_mode')
+    replicate_inspection_mode = arg_to_optional_bool(args.get('replicate_inspection_mode'))
+
+    update_arguments = (name, basepolicy_id, description, inspection_mode)
+
+    if not any(update_arguments):
+        raise ValueError(f'Please enter one of the update arguments: {update_arguments}')
```
dorschw, line 4287: here it'd be a list of none, please use a string,
```suggestion
        raise ValueError(f'Please enter one of the update arguments: name, basepolicy_id, descrpition, inspection_mode')
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3602,842 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+        client = Client(
+            base_url=base_url,
+            username=username,
+            password=password,
+            verify=verify_ssl,
+            proxy=proxy,
+        )
+
+        if command == 'test-module':
+            return_results('ok')
```
dorschw, line 4428: not testing anything?

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
```diff
@@ -316,3 +343,994 @@ def test_raw_response_to_context_access_policy(list_input, list_output):
...
+    """
+    message_by_method = {
+        'POST': 'Created',
+        'GET': 'Fetched',
+        'PUT': 'Updated',
+        'DELETE': 'Deleted',
+    }
+
+    if not command_results.readable_output:
+        assert False
```
dorschw, line 384:
```suggestion
    assert command_results.readable_output
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
```diff
@@ -316,3 +343,994 @@ def test_raw_response_to_context_access_policy(list_input, list_output):
...
+
+    assert message_by_method[method] in command_results.readable_output
+
+    if method != 'DELETE':
+        context_prefix = '.'.join((INTEGRATION_CONTEXT_NAME, expected_output_prefix))
+
+        assert command_results.outputs_prefix == context_prefix
+        assert_output_has_no_links(command_results.outputs)  # type: ignore # outputs is Optional[object]
```
dorschw, line 392: please use a specific ignore, not blanket.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
```diff
@@ -316,3 +343,994 @@ def test_raw_response_to_context_access_policy(list_input, list_output):
...
+
+    if not command_results.readable_output:
+        assert False
+
+    assert message_by_method[method] in command_results.readable_output
+
+    if method != 'DELETE':
+        context_prefix = '.'.join((INTEGRATION_CONTEXT_NAME, expected_output_prefix))
+
+        assert command_results.outputs_prefix == context_prefix
```
dorschw, line 391: please compare the expected output, for each command. See `HelloWorld_test.py`.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
```diff
@@ -316,3 +343,994 @@ def test_raw_response_to_context_access_policy(list_input, list_output):
...
+
+    if status_code == HTTPStatus.OK:
+        assert_command_results(
+            command_results=command_results,
+            method=method
+        )
+
+    elif status_code == HTTPStatus.NOT_FOUND:
+        assert command_results.readable_output == \
+            f'The Intrusion Policy ID: "{args["intrusion_policy_id"]}" does not exist.'
```
dorschw, line 631: please split to two tests, here and in other similar cases.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1,111 +1,527 @@
...
+                return items, raw_response
+
+            raw_items: list[dict[str, Any]] = []
+            raw_responses: list[dict[str, Any]] = []
+
+            # Keep calling the API until the required amount of items have been met.
+            while remaining_items > 0:  # type: ignore[operator]
+                raw_response = func(
+                    self,
+                    limit=remaining_items if remaining_items <= api_limit else api_limit,  # type: ignore[operator]
```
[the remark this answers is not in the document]
LiorQM: Thanks! added it here and elsewhere.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1,111 +1,527 @@
...
+
+                if raw_item is None:
+                    break
+
+                raw_responses.append(raw_response)
+                raw_items += raw_item
+
+                # Calculate the offset and limit for the next run.
+                received_items = len(raw_item)
+                remaining_items -= received_items  # type: ignore[operator]
```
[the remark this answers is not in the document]
LiorQM: I've changed the code a bit so it won't appear here, but basically the warning is because `remaining_items` might be `None`, although this scenario isn't possible.

Now I have 2 `type: ignore` in the decorator and they are each for the same warning ( a value that might be `None` even though it is within an `if` statement that promises it isn't).

I'd like to know your opinion on these cases, should I added a redundant `x is not None` even though this statement is saved withing a variable, for example: `is_automatic`. Or am I using `type: ignore` correclty?

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3602,842 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+
+    command_results = get_commandresults(
+        raw_response=raw_response,
+        command_headers_by_keys=INTRUSION_POLICY_HEADERS_BY_KEYS,
+        command_title=f'Fetched {INTRUSION_POLICY_TITLE}',
+        command_context=INTRUSION_POLICY_CONTEXT,
+    )
+
+    if not is_get_request:
+        command_results.raw_response = raw_responses
```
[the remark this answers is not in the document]
LiorQM: Because of `pagination` we may have to make several API requests to the server, for example: max `limit=10` and the user wanted `limit=30` which results in 3 `raw_response`. How should I handle those 3 `raw_response` with the actual `items` that are supposed to be returned to the user.

What should I do?
- Edit the 1st `raw_response` to hold the rest (not too fond of this).
- Create a `CommandResults` for each `raw_response` that will also hold the related `context_outputs` (this will result in a `readable_output` that is split to multiple pieces).
-  Leave it as is.
- Return all the `context_outputs` and `readable_output` (`items`)  in the first or last `CommandResults` and the rest of the `CommandResults` will hold `raw_responses` (this will result in 1 full `readable_output` and empty `readable_outputs` that will be sent).

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3602,842 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+        raw_response = client.delete_intrusion_policy(
+            intrusion_policy_id=intrusion_policy_id,
+        )
+
+    except DemistoException as exc:
+        if 'UUID cannot be null' in str(exc):
+            return CommandResults(
+                readable_output=f'The Intrusion Policy ID: "{intrusion_policy_id}" does not exist.'
+            )
+
```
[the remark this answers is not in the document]
LiorQM: Thanks, will do it from now on. Edited.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: Added.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: My bad 😅.
Changed it.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: This is from the old implementation.

This should be alright because if we've reached this point that means that the `__init__` which has a request within it (for the `bearer token`) has worked.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
LiorQM: Silly me 😅.
Added.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
LiorQM: Added `arg-type`.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
LiorQM: Done.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower_test.py
LiorQM: Added more asserts everywhere 🙂

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
dorschw: please add this as a comment in 4428, so future devs know it :)

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1,111 +1,531 @@
...
+            Raises:
+                ValueError: In case the user has mixed between automatic and manual pagination arguments.
+
+            Returns:
+                tuple[Union[list, dict], Union[list, dict]]:
+                    All the items combined within raw response, All the raw responses combined
+            """
+            is_automatic: bool = limit is not None and limit > 0
+            is_manual: bool = (page is not None and page > 0) or (page_size is not None and page_size > 0)
```
dorschw, line 136: please wrap with `bool(   )`

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -616,39 +1853,284 @@ def raw_response_to_context_rules(items: Union[Dict, List]) -> Union[Dict, List]
...
+                item_to_add[0]: item_to_add[1],
+                **context_output
+            }
+
+        context_outputs.append(context_output if context_output else item)
+
+    return context_outputs
+
+
+def get_commandresults(
```
dorschw, line 1978: please rename `parse_results`

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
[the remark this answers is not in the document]
dorschw: Understood, let's leave it as it is, thanks

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
dorschw: Cool, please add a comment explaining it, and a unit test for failure as well.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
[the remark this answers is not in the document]
dorschw: why is the type `Optional[int]` if it's not possible? Can we make it just `int`?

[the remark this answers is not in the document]
dorschw: yes, please add this check. It's redundant indeed, but preferred over muting mypy.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3606,844 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+    name = args.get('name', '')
+    basepolicy_id = args.get('basepolicy_id', '')
+    description = args.get('description')
+    inspection_mode = args.get('inspection_mode')
+    replicate_inspection_mode = arg_to_optional_bool(args.get('replicate_inspection_mode'))
+
+    update_arguments = (name, basepolicy_id, description, inspection_mode)
+
+    if not any(update_arguments):
+        raise ValueError(f'Please enter one of the update arguments: {update_arguments}')
```
dorschw, line 3732: please mention the field names explicitly, otherwise it will show as `Please enter one of the update arguments: "", "", "", None` or similar

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3606,844 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+        raise ValueError('rule_import_mode and rule_group_ids must be inserted when validate_only is "False".')
+
+    file_entry = demisto.getFilePath(entry_id)
+    filename = file_entry['name']
+    file_type = os.path.splitext(filename)[1]
+
+    if file_type not in ('.txt', '.rules'):
+        raise ValueError(f'Supported file formats are ".txt" and ".rules", got {file_type}')
+
+    with open(file_entry['path'], 'rb') as file_handler:
```
dorschw, line 4006: `rb`? is this a binary file?

dorschw (commented):
Fantastic job, please see the last comments, then I believe we'll be good to go.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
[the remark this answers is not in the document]
LiorQM: No problem, added an `if` to the two areas that had this mypy error.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: added a comment and a unit-test 🙂

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: wrapped!

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: renamed!

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: Silly me, can't believe I've actually done that. 😅
Fixed.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: Changed to `r`.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -319,8 +921,643 @@ def deploy_to_devices(self, force_deploy, ignore_warning, version, device_ids) -
...
+                'id': basepolicy_id
+            }
+        })
+
+        return self._http_request(
+            method='PUT',
+            url_suffix=f'policy/intrusionpolicies/{intrusion_policy_id}',
+            params=params,
+            json_data=body,
+            timeout=600
```
dorschw, line 1063: is the container configured to stay up this long? Please make sure the command (and other commands with timeout >300s) have `execution-timeout` configured in the yml.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -616,39 +1853,284 @@ def raw_response_to_context_rules(items: Union[Dict, List]) -> Union[Dict, List]
...
+        if contexts_to_delete:
+            context_output = delete_keys_from_dict(item, contexts_to_delete)
+
+        if item_to_add:
+            context_output = {
+                item_to_add[0]: item_to_add[1],
+                **context_output
+            }
+
+        context_outputs.append(context_output if context_output else item)
```
dorschw, line 1973:
```suggestion
        context_outputs.append(context_output or item)
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -616,39 +1853,284 @@ def raw_response_to_context_rules(items: Union[Dict, List]) -> Union[Dict, List]
...
+        ValueError: In case the user has entered both GET and LIST arguments, raise an error.
+
+    Returns:
+        bool: Are the GET arguments true.
+    """
+    is_get_request = any(get_args)
+    is_list_request = any(list_args)
+
+    if is_get_request and is_list_request:
+        raise ValueError('GET and LIST arguments can not be supported simutanlesy.')
```
dorschw, line 2054:
```suggestion
        raise ValueError('GET and LIST arguments can not be supported simultaneously.')
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3606,845 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+
+    command_results = parse_results(
+        raw_response=raw_response,
+        command_headers_by_keys=INTRUSION_RULE_HEADERS_BY_KEYS,
+        command_title=f'Fetched {INTRUSION_RULE_TITLE}',
+        command_context=INTRUSION_RULE_CONTEXT,
+    )
+
+    if not is_get_request:
+        command_results.raw_response = raw_responses
```
dorschw, line 3881: why not do it for get requests?

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3606,845 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+    """
+    intrusion_rule_id = args['intrusion_rule_id']
+    rule_data = args.get('rule_data', '')
+    rule_group_ids = argToList(args.get('rule_group_ids'))
+    update_strategy = args.get('update_strategy', 'OVERRIDE')
+
+    is_merge = update_strategy == 'MERGE'
+
+    if not any((rule_data, rule_group_ids)):
+        raise ValueError('At least rule_data or rule_group_ids must be entered, if not both of them.')
```
dorschw, line 3912:
```suggestion
        raise ValueError('rule_data, rule_group_ids or both must be populated.')
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
```diff
@@ -1469,120 +3606,845 @@ def get_task_status_command(client: Client, args: Dict) -> Tuple[str, Dict, Dict
...
+
+    # Rule groups must be entered when merging.
+    if is_merge and not rule_group_ids:
+        raise ValueError('rule_group_ids must be entered when merging.')
```
dorschw, line 3916:
```suggestion
        raise ValueError('rule_group_ids must be populated when merging.')
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: In the commands there are two function calls, the functions handle `GET` and `LIST` request according to the user's input.

In all of the list requests I return a `Tuple` that holds two arguments, the first is the appended `items` within all the responses and the second is the the actual `raw response`
All the list requests are handled in a regular manner (single request), therefore only a `raw_response` is returned.

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: From what I saw in other integrations, I've implemented it this way:
Added `execution-timeout` to commands descriptions and removed `timeout` from their requests.

Did I do it right?

dorschw: hmm doesn't removing the timeout from the request make it use the default=30s? I think we should explicitly pass `timeout=600`

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
dorschw: Let's move this part to row 3857 (as part of the `if` block), to keep the flow clear.
By the way, we can remove the `is_get_request` variable at all - in line 3854 just do `if check_is_get_requect(...):`

dorschw: please apply it in all functions, I see it's a recurring thing. The flow is a bit confusing this way

LiorQM: The thing is `parse_results` has to be called after the `if` and `else` statements.
I can implement this suggestion If I hold an empty variable `raw_responses = None` and also add this argument to the `parse_results` method which will set it if it exists.
What do you think?

You may see the new implementation below, also the `command_results.raw_response` has passed to the method.
```
raw_responses = None

    if check_is_get_request(
        get_args=[intrusion_rule_id],
        list_args=[sort, filter_string, expanded_response, limit, page, page_size]
    ):
        raw_response = client.get_intrusion_rule(
            intrusion_rule_id=intrusion_rule_id,
        )

    else:  # is_list_request
        raw_response, raw_responses = client.list_intrusion_rule(
            page=page,
            page_size=page_size,
            limit=limit,
            sort=sort,
            filter_string=filter_string,
            expanded_response=expanded_response,
        )

    command_results = parse_results(
        raw_response=raw_response,
        command_headers_by_keys=INTRUSION_RULE_HEADERS_BY_KEYS,
        command_title=f'Fetched {INTRUSION_RULE_TITLE}',
        command_context=INTRUSION_RULE_CONTEXT,
        raw_responses=raw_responses,
    )
```

Packs/CiscoFirepower/Integrations/CiscoFirepower/CiscoFirepower.py
LiorQM: I removed it because I noticed there is a global argument in XSOAR called `execution-timeout` to control this situation.
I figured this out while going through some other `Packs`.

Here is an image of its use:
image

And an image of the command description to notice the user to use that argument, if needed:
image

If a new argument is needed instead of this as you suggested `timeout`, I'll add it immediately.

dorschw: Yeah, the built-in is what we need, no need to declare it. The only thing I'd change is `requests`' `timeout` value to 600, so when `execution-timeout` is used, `requests` won't limit the time to just 30 seconds (the default iirc)

dorschw (approved):

PR #22979, opened 2022-12-12
fixed separated prs link

JasBeilin:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4775

- Changed the seperator of appending text in Jira to " , " instead of "," as it made one concatenated string (links were not seperated causing 404 when clicked on)

- Passing the action if "relates"/ "related" or "fixes"/"fixed" were not find in order to pass builds that does not require linking, without compromising on incorrectly linking. ("fixed" won't work at the moment due to incorrect regex, we want to fail it so developer will know).


Packs/Jira/ReleaseNotes/2_0_19.md
```diff
@@ -0,0 +1,5 @@
+
+#### Integrations
+##### Atlassian Jira v2
+- Fixed an issue where appending data to existing string in ***jira-append-to-field*** the separator made links unavailable.
```
BEAdi, line 4:
```suggestion
- Fixed an issue where when appending data to an existing string in ***jira-append-to-field*** command, the separator made links unavailable. 
```

BEAdi (approved):

PR #23096, opened 2022-12-15
Command-Line Analysis - added missing "else" path to playbook

idovandijk:
## Status
Ready

## Description
Fixed another bug where the condition that validates the input had 2 "yes" paths instead of "yes" and "else".

No


ssokolovich (commented):
Please also update the image for the playbook 
(so the else path will be visible - instead of the existing 'yes').

ssokolovich (approved):

PR #23098, opened 2022-12-15
Tanium threat response v2 fetch fix

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes:[CIAC-4621](https://jira-hq.paloaltonetworks.local/browse/CIAC-4621)

## Description
Fix for the extraction of **filter_by_label_name** integration param.

   - [ ] Yes
       - Further details:
   - [x] No


Packs/TaniumThreatResponse/Integrations/TaniumThreatResponseV2/TaniumThreatResponseV2.py
```diff
@@ -424,7 +420,7 @@ def test_module(client, data_args):
         raise ValueError(f'Please check your credentials and try again. Error is:\n{str(e)}')
 
 
-def fetch_incidents(client, alerts_states_to_retrieve, label_names_to_retrieve, last_run, fetch_time, max_fetch):
+def fetch_incidents(client, alerts_states_to_retrieve, label_name_to_retrieve, last_run, fetch_time, max_fetch):
```
JudahSchwartz, line 423: label_names_to_retrieve isnt a list anymore? What changed? Lets add type refs

JudahSchwartz (commented):
Please also add a unittest that reproduces the bug

Packs/TaniumThreatResponse/Integrations/TaniumThreatResponseV2/TaniumThreatResponseV2.py
YuvHayun: It's not anymore - the issue is divided into two parts:
1. The integration didn't know how to handle cases where empty label names were given.
2. Unlike what was written in the first place, the API can handle only one label at a time.
that's why label_names_to_retrieve > label_name_to_retrieve

YuvHayun: I'll add type refs & unit tests.

JudahSchwartz: is that a change that the api made? Or we just implemented this wrong in the first place?

Packs/TaniumThreatResponse/Integrations/TaniumThreatResponseV2/TaniumThreatResponseV2_test.py
```diff
@@ -1450,11 +1451,12 @@ def test_fetch_new_incidents(requests_mock):
     last_run = {'time': '2021-09-26T14:02:59.000000Z', 'id': '2'}
     fetch_time = '3 days'
     max_fetch = 2
+    filter_label_name = ''
```
JudahSchwartz, line 1454: lets assert the url we got with these 2 tests with requestsmock

YuvHayun: They're being asserted, requests_mock.get won't retrieve the incidents if the call wasn't made to the exact URL.|
I'm even tried to change the URL in the requests_mock.get and saw that the unit-test failed

JudahSchwartz: if it wasnt added it would still succeed because
```
 requests_mock.get(BASE_URL + '/plugin/products/detect3/api/v1/alerts?'
                                 '&state=unresolved&sort=-createdAt&limit=500&offset=0'
```
is there without the suffix, no?

YuvHayun: There're two unit tests, one of a fetch_incidents with an empty filter_label_name that ensures that the 'labelName' query param is not being added to the URL suffix.

And the second one with a labelName that ensures the 'labelName' query param is being added to the URL suffix. at the right way.

JudahSchwartz (approved):

PR #23144, opened 2022-12-19
SentinelOne: add the EventID as an key

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-18961

## Description
problem: the output was unique by the `processID` which is not unique and few alerts may contained the same `processID`
solution:
add the EventID and used it as an output key in order to unique the results that added some new output keys

## Screenshots

**fail unit test befor fix:**
image

**passed unit test after fix:**
image


Packs/SentinelOne/ReleaseNotes/3_0_3.md
```diff
@@ -0,0 +1,4 @@
+
+#### Integrations
+##### SentinelOne v2
+Fixed an issue where some context entries the was identified as duplication in ***sentinelone-get-events*** command .
```
DeanArbel, line 4:
```suggestion
- Added new outputs to **sentinelone-get-events** command.
- Fixed an issue in **sentinelone-get-events** where some results were unified under the same key in the context, resulting in fewer context entries than expected.
```

DeanArbel (approved):
Good job

PR #23157, opened 2022-12-20
add latest to native image config

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Description
add the "latest" to the native image config

### DIFF
--- a/Tests/docker_native_image_config.json
+++ b/Tests/docker_native_image_config.json
@@ -30,6 +30,37 @@
             "tesseract"
          ],
          "docker_ref": "demisto/py3-native:8.1.0.40949"
+      },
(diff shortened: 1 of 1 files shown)


tkatzir (commented):
Shouldn't the `ignored_content_items` be empty?

GuyAfik:
> Shouldn't the `ignored_content_items` be empty?

if you recall, for each content item we define which native images are supported individually, for example:
```
"ignored_content_items":[
      {
         "id":"UnzipFile",
         "reason":"Failed unit-test: test_unrar_no_password",
         "ignored_native_images":[
            "native:8.1"
         ]
      }
   ]
```

PR #23167, opened 2022-12-20
adding redirect_std_out integration param to vmware

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Description
Added the **redirect_std_out** integration param, used for troubleshooting.

   - [x] Yes
       - Further details:
   - [ ] No

### DIFF
--- a/Packs/VMware/Integrations/VMware/VMware.yml
+++ b/Packs/VMware/Integrations/VMware/VMware.yml
@@ -622,7 +622,7 @@ script:
       required: true
       isArray: true
     outputs: []
-  dockerimage: demisto/vmware:2.0.0.39940
+  dockerimage: demisto/vmware:2.0.0.40991
   runonce: false
   script: ""
   subtype: python3
--- a/Packs/VMware/Integrations/VMware/VMware.yml
+++ b/Packs/VMware/Integrations/VMware/VMware.yml
@@ -17,6 +17,11 @@ configuration:
   additionalinfo: Username and password used to login into the system.
   displaypassword: ""
   hiddenusername: false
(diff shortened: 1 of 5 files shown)


Packs/VMware/ReleaseNotes/1_2_6.json
```diff
@@ -0,0 +1 @@
+{"breakingChanges":true,"breakingChangesNotes": "Added the **redirect_std_out** integration param, use this param when encountering issues with docker timeout."}
```
JudahSchwartz, line 1: are you sure this needs to be a breaking change? Adding a param that has a default?

Packs/VMware/Integrations/VMware/VMware.yml
```diff
@@ -17,6 +17,11 @@ configuration:
   additionalinfo: Username and password used to login into the system.
   displaypassword: ""
   hiddenusername: false
+- name: redirect_std_out
```
JudahSchwartz, line 20: lets give a default of false in the yml and also the code

JudahSchwartz (approved):

Packs/VMware/ReleaseNotes/1_2_6.json
YuvHayun: Yes, because the default is different from what occurs now.

Packs/VMware/Integrations/VMware/VMware.yml
YuvHayun: It's a checkbox it's False by default.
Anyway there's also a fault false in the code.

PR #17689, opened 2022-02-16
Panos network operations base

adambaumeister:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: link to the issue

## Description
@GuyAfik 

This is a PR that includes only the **pan-os-platform** series of commands (17 commands). It also removes the decorators, but maintains the dataclasses. Otherwise, the code is unchanged from the previous PR; https://github.com/demisto/content/pull/17430

## Minimum version of Cortex XSOAR
- [ ] 6.0.0
- [ ] 6.1.0
- [X] 6.2.0
- [ ] 6.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

PR #23201, opened 2022-12-21
[PAN-OS] - fix bug in pan-os-platform-get-ha-state command

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-19555)

## Description
Fixes an issue where if a firewall is enabled as HA, the **pan-os-platform-get-ha-state command** commands still returns that its not enabled.

## Must have
- [x] Tests
- [ ] Documentation


GuyAfik:
failed UT - https://app.circleci.com/pipelines/github/demisto/content/217663/workflows/dcb7ec8f-d954-480e-a689-b735290c20c1/jobs/618024

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -10724,7 +10724,7 @@ def get_device_state(topology: Topology, target: str):
     @staticmethod
     def get_ha_status(
         topology: Topology, device_filter_str: Optional[str] = None, target: Optional[str] = None
-    ) -> List[ShowHAState]:
+    ) -> List[ShowHAState] | ShowHAState:
```
JudahSchwartz, line 10727: why return 1 instead of a list of 1? Lets be consistent with our return types

GuyAfik: This is how it was originally written by @adambaumeister, the reason was related to the way the context shows the data, i don't recall exactly why, if we change it that might affect the way context is showing the data, so i prefer not to change it if it was implemented that way in the first place.

I had the same question here: https://github.com/demisto/content/pull/17689/files#r808922114 (you can read)

JudahSchwartz (approved):

PR #23231, opened 2022-12-22
Jira mirroring bug

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-18171
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-20237

## Description
Fixed some issues: 
1. the mirror out did not work and got a key error - fixed that with safe get.
2. the mirror out for the status field did not work - we did not take the right transactions in the code.
3. the URL in the get_user_data function was wrong following a recent change in the code.


DeanArbel (approved):
Looks good

Packs/Jira/Integrations/JiraV2/JiraV2.py
```diff
@@ -830,7 +831,8 @@ def edit_status(issue_id, status, issue):
     if not issue:
         issue = {}
     j_res = list_transitions_data_for_issue(issue_id)
-    transitions = [transition.get('name') for transition in j_res.get('transitions')]
+    # When changing the status we search the transition that leads to this status
+    transitions = [transition.get('to', {}).get('name', '') for transition in j_res.get('transitions')]
```
yuvalbenshalom, line 835: is it that the relevant transitions has not been found? or that it found wrong transitions?
I am asking because I think this code generally worked, and I want to make sure it is not a BC break in Jira versions.

yuvalbenshalom (commented):
looks good, see my question.

Packs/Jira/Integrations/JiraV2/JiraV2.py
yucohen: We found the transitions names, but when we change the status, the correct thing to do is to check if the transition can lead to the status (therefore taking the *to* section). Sometimes the transitions have the same names as the statuses (and this is probably why we did not catch this earlier), but in this case they dont, so comparing the transitions names to the status name is wrong.
@JasBeilin Do you have anything to add here?

yuvalbenshalom: @yucohen @JasBeilin so if I understand correctly, after this change we are collecting statuses and not transitions. if this is true then the names of the variable should reflect this.

JasBeilin: This is what we are currently doing (at least by declaration).
```def edit_status(issue_id, status, issue):
    # check for all authorized transitions available for this user
    # if the requested transition is available, execute it.```
    
 It is part of this function, the only way to change status is by transition. It's not collecting as much as it is finding the actual transition leading to the status. Until now, not sure how come we did not encounter it - we looked for a transition with the same name of the status so it a transition was leading to it but had different name it would not be able to change the status.

PR #23278, opened 2022-12-26
Run Release Notes Review only on XSOAR-supported Packs

kgal-pan:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-2954

## Description
The changes introduced here are:

* Changed files returned from _Get changed files_ job in _Review Release Notes_ workflow  includes only release notes.
* Added env var  (`env.CHANGED_FILES_DELIMITER`) to hold the delimiter of the changed files.
* Added new script (`Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py`) that returns only release notes from XSOAR-supported Packs.
* Unit test for new script.
* Removed usage of `poetry` in the workflow as it was installing unused dependencies from `pyproject.toml` and `poetry.lock` unnecessarily elongating the build. Replaced `poetry` with using `pip` which took down the build time from ~4m to right under 2m (ignore the other runs, they were for tuning performance):

image

<details><summary><h2>Build Tests</h2></summary>

### Case 1: Single Community-supported Pack

[Build link](https://github.com/demisto/content/actions/runs/3782269822/jobs/6429869741#step:7:1)

**Expected:**
Doc review to be skipped.

**Actual**

```bash
Release notes found in PR: Packs/HelloWorld/ReleaseNotes/1_3_0.md
No need to run doc review. Exiting...
```

### Case 2: 2 Packs, 1 XSOAR-supported, 1 Community-supported

[Build link](https://github.com/demisto/content/actions/runs/3782380225/jobs/6430072356#step:7:1)

**Expected**
Doc review only for XSOAR-supported release notes

**Actual**

```bash
Release notes found in PR: Packs/HelloWorld/ReleaseNotes/1_3_0.md,Packs/SentinelOne/ReleaseNotes/3_1_0.md
Release notes requiring review: Packs/SentinelOne/ReleaseNotes/3_1_0.md}
Starting docs review...

================= Starting Doc Review =================
Running only on release notes

Checking file Packs/SentinelOne/ReleaseNotes/3_1_0.md
...
```
</details>


kgal-pan:
@ShacharKidor  - Please review. Current state of PR has the [_Review Release Notes_ workflow failing on purpose](https://github.com/demisto/content/actions/runs/3787376475/jobs/6439123618) because I wanted to test the scenario of a XSOAR-supported Pack (SentinelOne) that has invalid release notes along with a community-supported Pack (HelloWorld):

```
Release notes found in PR: Packs/HelloWorld/ReleaseNotes/1_3_0.md,Packs/SentinelOne/ReleaseNotes/3_1_0.md
Release notes requiring review: Packs/SentinelOne/ReleaseNotes/3_1_0.md
Starting docs review...

================= Starting Doc Review =================
Running only on release notes

Checking file Packs/SentinelOne/ReleaseNotes/3_1_0.md
```

I will remove the updates to the two Packs once we confirm the rest of the code is OK.

.github/workflows/review-release-notes.yml
```diff
@@ -12,9 +15,13 @@ jobs:
           fetch-depth: 2
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v18.7
+        uses: tj-actions/changed-files@v35.1.0
         with:
-          separator: ';'
+          separator: ${{ env.CHANGED_FILES_DELIMITER }}
+          files: |
+            Packs/**/ReleaseNotes/**
```
ShacharKidor, line 22: Nice addition :) 
@DinaMeylakh - please confirm that we are not missing anything by doing that.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,122 @@
...
+    parser.add_argument("release_notes", help="Comma separated list of release notes paths")
+    parser.add_argument('-d', '--delimiter', help="The delimiter that separates the changed files names,"
+                                                  " defined in yq 'env.CHANGED_FILES_DELIMITER' .github/workflows/"
+                                                  "review-release-notes.yml", default=",")
+
+    args = parser.parse_args()
+
+    try:
+        output = main(args)
+        print(output)
```
ShacharKidor, line 118: Lets use click.secho intead of the 2 prints here

ShacharKidor (changes requested):
@kgal-pan - Great work! 

I've added a few comments - please review them.
Also, let's discuss tomorrow - would like to verify a few things.

.github/workflows/review-release-notes.yml
```diff
@@ -1,6 +1,9 @@
 name: Review Release Notes
 on: pull_request
 
+env:
+  CHANGED_FILES_DELIMITER: ","
```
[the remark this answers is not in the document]
kgal-pan: No specific reason, we can change it back to semicolon if needed. I reviewed the PR/issue but couldn't find a specific comment about a comma (only single quotes). In any case, both `,` and `;` are [valid in filenames](https://www.ibm.com/docs/en/zos/2.1.0?topic=pages-zos-data-set-unix-file-naming-conventions) so it shouldn't matter  much which delimiter is used. 

In any case, since we're adding the following to the job:

```yaml
          files: |
            Packs/**/ReleaseNotes/**
```

We're only going to get a list back of release notes standard format `$MAJOR_MINOR_REVISION.md` so no commas or semicolons should appear in the filenames we're going to process through the doc review.

Utils/github_workflow_scripts/run_docs_review.py
```diff
@@ -47,9 +47,9 @@ def parse_changed_files_names() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description='Parse the changed files names.')
     parser.add_argument('-c', '--changed_files',
                         help="The files that are passed to docs review (passed as one string).")
-    parser.add_argument('-d', '--delimiter', help="the delimiter that separates the changed files names (determined in"
-                                                  " the call to tj-actions/changed-files in "
-                                                  "review_release_notes script).")
+    parser.add_argument('-d', '--delimiter', help="The delimiter that separates the changed files names,"
+                                                  " defined in yq 'env.CHANGED_FILES_DELIMITER' .github/workflows/"
+                                                  "review-release-notes.yml", default=",")
```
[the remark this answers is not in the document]
kgal-pan: Same as https://github.com/demisto/content/pull/23278#discussion_r1057776370.

> By 'yq' you mean YAML processor?

Yep, [this one](https://github.com/mikefarah/yq) :)

.github/workflows/review-release-notes.yml
kgal-pan: From what I see in https://github.com/demisto/content/blob/CIAC-4970-run-release-notes-check-only-if-xsoar-supported-pa/Utils/github_workflow_scripts/run_docs_review.py#L27, we're specifying that only release notes will be processed through the doc review:

```python
doc_reviewer = DocReviewer(file_paths=files_for_review,
                               release_notes_only=True, # <-----------
                               known_words_file_paths=['Tests/known_words.txt'],
                               load_known_words_from_pack=True,
                               no_camel_case=True)
```

So I believe there's no need to have any other changed files (besides release notes) in this context.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: Done in [`c17a1c0` (#23278)](https://github.com/demisto/content/pull/23278/commits/c17a1c00014fc95d0692b7cca6f27518f899c8e0).

.github/workflows/review-release-notes.yml
```diff
@@ -12,22 +16,30 @@ jobs:
...
-          echo "Starting docs review"
-          poetry run python Utils/github_workflow_scripts/run_docs_review.py --changed_files "$changed_files" --delimiter $delim
-          echo "Finished docs review successfully"
\ No newline at end of file
+          release_notes="${{ steps.changed-files.outputs.all_changed_files }}"
+          echo "Release notes found in PR: ${release_notes}"
+          release_notes_to_review=$(Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py "$release_notes" --delimiter ${{ env.CHANGED_FILES_DELIMITER }})
+
+          # Check if there are any release notes
+          if [ $? -eq 0 ] && [ -n "$release_notes_to_review" ]; then
```
DinaMeylakh, line 37: The script should work when passing empty arguments, so I feel like this is redundant here and could potentially hide deeper problems with the sdk command.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+        - `delimiter` (`str`): The delimiter to separate the release note paths
+
+    Returns"
+        `str` of joined release notes to review
+    """
+
+    return delimiter.join(rns_to_review)
+
+
+def main(args: argparse.Namespace) -> str:  # pragma: no cover
```
DinaMeylakh, line 70: Lets give this function a proper name

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+    parser.add_argument("release_notes", help="Comma separated list of release notes paths")
+    parser.add_argument('-d', '--delimiter', help="The delimiter that separates the changed files names,"
+                                                  " defined in yq 'env.CHANGED_FILES_DELIMITER' .github/workflows/"
+                                                  "review-release-notes.yml", default=",")
+
+    args = parser.parse_args()
+
+    try:
+        output = main(args)
+        click.secho(output)
```
DinaMeylakh, line 119: If we already use click here, why not use click.option instead of argparse?

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+    parses the release notes into a paths and checks whether they are part
+    of an XSOAR-supported Pack.
+
+    Returns the output, list of paths of release notes requiring review.
+
+    """
+
+    arg_dict: Dict[str, Any] = vars(args)
+
+    delimiter: str = arg_dict.get("delimiter", ",")
```
DinaMeylakh, line 83:
```suggestion
    delimiter: str = arg_dict.get("delimiter", ";")
```

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+        return ""
+
+
+if __name__ == "__main__":  # pragma: no cover
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("release_notes", help="Comma separated list of release notes paths")
+    parser.add_argument('-d', '--delimiter', help="The delimiter that separates the changed files names,"
+                                                  " defined in yq 'env.CHANGED_FILES_DELIMITER' .github/workflows/"
+                                                  "review-release-notes.yml", default=",")
```
DinaMeylakh, line 113: The script should stand on its own, please refrain from specifying env variables.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+    of an XSOAR-supported Pack.
+
+    Returns the output, list of paths of release notes requiring review.
+
+    """
+
+    arg_dict: Dict[str, Any] = vars(args)
+
+    delimiter: str = arg_dict.get("delimiter", ",")
+    release_notes_arg: List[Path] = convert_files_to_paths(arg_dict.get("release_notes", []).split(delimiter))
```
DinaMeylakh, line 84: This convert_files_to_paths fuction makes this script less efficient, we already iterate over the paths in line 94, and it can be converted to a path there, no need to iterate twice

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+    if release_notes_arg:
+        for rn in release_notes_arg:
+            pack_name = rn.parts[1]
+            if is_pack_xsoar_supported(pack_name):
+                release_notes_to_review.append(str(rn))
+
+        if release_notes_to_review:
+            return format_output(release_notes_to_review, delimiter)
+        else:
+            return ""
```
DinaMeylakh, line 102: redundant as lines 103-104 capture this case

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Args:
+        - `rns_to_review` (``List[str]``): List of release note paths.
+        - `delimiter` (`str`): The delimiter to separate the release note paths
+
+    Returns"
+        `str` of joined release notes to review
+    """
+
+    return delimiter.join(rns_to_review)
```
DinaMeylakh, line 67: I see no need for a whole function declaration just for this line

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+    Checks whether the supplied Pack is from a XSOAR-supported
+
+    Args:
+        - `pack_name` (``str``): The name of the Pack
+
+    Returns:
+        `bool` representing whether Pack is XSOAR-supported or not
+    """
+
+    pack_metadata = Path("Packs", pack_name, "pack_metadata.json")
```
DinaMeylakh, line 28: lets get this path from demisto-sdk constants instead of hard coded

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if pack_metadata.exists():
+        with pack_metadata.open(encoding="UTF-8") as pm:
+            metadata: Dict[str, Any] = json.load(pm)
```
DinaMeylakh, line 32: let's add try and except here

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
```diff
@@ -0,0 +1,123 @@
...
+            if metadata.get("support") == "xsoar":
```
DinaMeylakh, line 33: This should also be done with constants, where do we get the "xsoar" supported constants?

DinaMeylakh (changes requested):
I believe there are some changes in this pr that are out of scope and require a different issue if voted to change, as it could reintroduce previous bugs and workflow failures, also maybe we need to rethink where the answer to this issue should be. 

So:
1. If you want to change the workflow setup, please open a different issue and a pr so you can test thoroughly. The truth is it's a delicate workflow and it took a while to make it stable, so I'm skeptical of introducing changes where there is no need.
2. Good work on the logic of finding the support level of each pack, but maybe it will be better if it sits inside the demisto-sdk command? That way if a customer also uses this command in their build, they can use this cool feature. It will also resolve you from handling exceptions and supporting a command line script. @ShacharKidor wdyt?
3. As I said in my comments, please refrain from adding a lot of logic to the yml as it is harder to test and maintain.

.github/workflows/review-release-notes.yml
```diff
@@ -1,6 +1,10 @@
 name: Review Release Notes
 on: pull_request
 
+env:
+  CHANGED_FILES_DELIMITER: ","
+  PYTHONWARNINGS: ignore
```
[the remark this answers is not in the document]
kgal-pan: This prevents the job log from displaying warnings concering Python code. I realized that I didn't specify the type of warning I wanted to ignore, specifically `DeprecationWarning`. I think preventing prints of `DeprecationWarning` makes the job log more readable.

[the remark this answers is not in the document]
kgal-pan: Resolved in [demisto/content@`b64bbe4` (#23278)](https://github.com/demisto/content/pull/23278/commits/b64bbe400121741bcb2b030d03666b6c20f4ca4e).

.github/workflows/review-release-notes.yml
```diff
@@ -12,22 +16,30 @@ jobs:
...
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v18.7
-        with:
-          separator: ';'
-      - name: Setup Python
-        uses: actions/setup-python@v3
+        uses: tj-actions/changed-files@v35.1.0
         with:
-          python-version: '3.9'
```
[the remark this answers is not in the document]
kgal-pan: The runner image we're currently using (`ubuntu-latest`) [comes preinstalled with Python 3.10](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#language-and-runtime).

.github/workflows/review-release-notes.yml
```diff
@@ -12,22 +16,30 @@ jobs:
...
-      - name: Install Python Dependencies
-        run: |
-          poetry install --with ci
+          separator: ${{ env.CHANGED_FILES_DELIMITER }}
+          files: |
+            Packs/**/ReleaseNotes/**
+          since_last_remote_commit: false
+
+      - name: Install SDK
+        run: python3 -m pip install demisto-sdk
```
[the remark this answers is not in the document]
kgal-pan: > Can I have a quick link to the meaning of specifying files like that?

Sure, [here is the link](https://github.com/tj-actions/changed-files#inputs) to the input details of this action and here's one explaining [`glob` matching patterns](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet).

> I think there is little advantage to doing this here. It's harder to test if files went missing and all we gain is that the Review Release Notes step is skipped in the Review Release Notes pipeline, which can be confusing.

I disagree that there's little advantage of adding this filtering. The advantages of adding this `glob` I see are:

- We only process files that are relevant to this scope. This prevents issues such as processing file names that include characters that might break the delimiter such as the one we see [here](https://github.com/demisto/content/pull/23278#discussion_r1057776370). 
- In PRs with a large amount of files, there's a performance penalty for bringing in all PR changed files. This makes the job run unnecessarily long. 

> How do you run python if you don't install it?
See https://github.com/demisto/content/pull/23278#discussion_r1059736434

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: Addressed in [demisto/content@`a6b811a` (#23278)](https://github.com/demisto/content/pull/23278/commits/a6b811a763ae1c8ec4fe5789fc7203b5bcd56f07)

ShacharKidor:
> I believe there are some changes in this pr that are out of scope and require a different issue if voted to change, as it could reintroduce previous bugs and workflow failures, also maybe we need to rethink where the answer to this issue should be.
> 
> So:
> 
> 1. If you want to change the workflow setup, please open a different issue and a pr so you can test thoroughly. The truth is it's a delicate workflow and it took a while to make it stable, so I'm skeptical of introducing changes where there is no need.
> 2. Good work on the logic of finding the support level of each pack, but maybe it will be better if it sits inside the demisto-sdk command? That way if a customer also uses this command in their build, they can use this cool feature. It will also resolve you from handling exceptions and supporting a command line script. @ShacharKidor wdyt?
> 3. As I said in my comments, please refrain from adding a lot of logic to the yml as it is harder to test and maintain.

@DinaMeylakh - Thank you for your thorough review.

1. Agree - @kgal-pan the changes in the poetry and python installations are unrelated to this issue... if you think we need to change those things let's open a new PR/Issue (and have the relevant reviewers). 

2. Also agree - I think it would be beneficial to run the SDK command locally with the option to ignore partners/community packs. It can also help with testing/debugging. Having said that, all we need to do is to move the logic from the script to the SDK command (doc reviewer).
@kgal-pan - WDYT?

.github/workflows/review-release-notes.yml
```diff
@@ -12,22 +16,30 @@ jobs:
...
+
       - name: Review Release Notes
+        if: steps.changed-files.outputs.all_changed_files != '' && steps.changed-files.outputs.all_changed_files != ' '
```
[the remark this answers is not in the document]
kgal-pan: See https://github.com/demisto/content/pull/23278#discussion_r1059738164, second bullet.

.github/workflows/review-release-notes.yml
kgal-pan: Relevant to discussion https://github.com/demisto/content/pull/23278#discussion_r1059743911.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: TBH, not sure why we need to use `click` at all here when `print` will do just fine. I made the change because of suggestion made [here](https://github.com/demisto/content/pull/23278#discussion_r1057763696).

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: I think it actually makes it easier to test out the expected script output as specified in [UT](https://github.com/demisto/content/blob/CIAC-4970-run-release-notes-check-only-if-xsoar-supported-pa/Utils/github_workflow_scripts/github_workflow_scripts_tests/test_get_xsoar_supported_release_notes.py#L80).

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: To catch what? If the file exists? if the file can be parsed as JSON? 

In both cases I think that by the time we're in this stage of the flow, we're expecting the pack metadata to exist and be valid JSON since the SDK bumps the version in it and created the RN with the `update-release-notes` command. If the file doesn't exist or is not parsable as JSON, the `update-release-notes` command would fail and the release notes wouldn't get generated in the first place.

If we do still want to handle an exception, we're running `main` under `try..catch` block so I guess we would `raise` some type of exception according to what we're attempting to catch.

Utils/github_workflow_scripts/get_xsoar_supported_release_notes.py
kgal-pan: Addressed in [`cecfe10` (#23278)](https://github.com/demisto/content/pull/23278/commits/cecfe10f03509fe33a3b4d3b0f56b9c8876ec104)

kgal-pan:
> 1. Agree - @kgal-pan the changes in the poetry and python installations are unrelated to this issue... if you think we need to change those things let's open a new PR/Issue (and have the relevant reviewers).

Sounds good, I'll open a separate issue for this.

> 2. Also agree - I think it would be beneficial to run the SDK command locally with the option to ignore partners/community packs. It can also help with testing/debugging. Having said that, all we need to do is to move the logic from the script to the SDK command (doc reviewer).
>    @kgal-pan - WDYT?

Cool, I'll close this PR and open one in the SDK with said changes.

PR #23286, opened 2022-12-26
Parse emails entry bug

daryakoval:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-4930)

## Description
When running `getEntry` with EntryID for the file that was uploaded using DataCollection task, the script fails with the error: `Item Not Found (8).`

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/ParseEmailFilesV2/ParseEmailFilesV2.py
```diff
@@ -87,12 +87,10 @@ def extract_file_info(entry_id: str) -> tuple:
 
         file_path = result[0]['Contents']['path']
         file_name = result[0]['Contents']['name']
-        result = demisto.executeCommand('getEntry', {'id': entry_id})
-        if is_error(result):
-            return_error(get_error(result))
 
-        file_metadata = result[0]['FileMetadata']
-        file_type = file_metadata.get('info', '') or file_metadata.get('type', '')
+        dt_file_type = demisto.dt(demisto.context(), f"File(val.EntryID=='{entry_id}').Type")
```
dantavori, line 91: 1. is it always a list?
2. before the change we did "info or type", do we preserve the logic if we extract only the type now?

daryakoval: 1. In my experiments it was always a `list`, but better to put the check, because there is no documentation what the return type.
2. Before the change we took the type (info) from FileMetadata, and I changed it to take the type from context. The value was previously under the info key, and I noticed that the same value is under the Type key in the context.

dantavori: 1 - let's make sure that if it's not a list (i.e. a string) we still assign a value to file_type from the dt result

dantavori (approved):

PR #23377, opened 2022-12-28
Changed phishing to assign less-busy-user instead of administrator

idovandijk:
## Status
Ready

## Related Issues
Related to: https://jira-hq.paloaltonetworks.local/browse/CRTX-70287

## Description
Changed the task that assigns an analyst to assign a "less-busy-user" instead of the user with the role of Administrator.

No


idovandijk:
Tested on XSOAR NG and works fine

tomer-pan (approved):

PR #23442, opened 2023-01-01
If-Then-Else enhancement

melamedbn:
## Status
- [ ] In Progress
- [x] Ready

## Description
With this enhancement, a user could put two different conditions in the script.

- [x] 5.0.0
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [ ] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/IfThenElse/IfThenElse.js
```diff
@@ -21,6 +21,15 @@ function getValue(args, condition, options, lhs) {
     throw `Invalid condition: ${condition}`;
 }
 
+function getValueB(args, second_conditionB, optionsB, lhsB) {
```
gal-forer, line 24: @melamedbn You don't need to duplicate the function and hard code the array inside the for, just send it as a variable to the function and also make the

gal-forer (changes requested):

gal-forer (approved):

PR #23542, opened 2023-01-04
Moving to common scripts part two

maimorag:
## Status
- [x] In Progress

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5042

## Description
moving DBotPredictURLPhishing script from **Phishing URL** to **common scripts**.

- [X] 6.0.0

   - [X] Yes


Packs/CommonScripts/pack_metadata.json
```diff
@@ -47,6 +47,14 @@
         "RemoteAccess": {
             "mandatory": false,
             "display_name": "RemoteAccess v2"
+        },
+        "rasterize": {
```
Shellyber, line 51: Why did we add these dependencies? 
It shouldn't be mandatory: true

Shellyber (commented):
Looks good.
See my comment regarding the dependencies.
Also, add in the template the link to the issue

Packs/CommonScripts/pack_metadata.json
maimorag: the script DBotPredictURLPhishing.py uses rasterize and Whois

maimorag:
DBotPredictURLPhishing will stay in PhishingURL folder

PR #23559, opened 2023-01-04
trigger build on content private custom branch

esharf:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5145

requierd: https://github.com/demisto/content-private/pull/640

job when branch exist: https://code.pan.run/xsoar/content/-/jobs/20092547
triggered: https://github.com/demisto/content-private/actions/runs/3841781671/jobs/6542391473#step:2:672

job when branch doesn’t exist: https://code.pan.run/xsoar/content/-/jobs/20092742
triggered: https://github.com/demisto/content-private/actions/runs/3841826580/jobs/6542484745#step:2:671

## Description
added new ability for content-private build trigger with custom branch on content-private repo.

### DIFF
--- a/.gitlab/ci/on-push.yml
+++ b/.gitlab/ci/on-push.yml
@@ -17,7 +17,9 @@ trigger-private-build:
     - if: '$NIGHTLY'
   script:
     - echo "====Trigger Private Build===="
-    - python3 Utils/trigger_private_build.py --github-token $GITHUB_TOKEN
+    - git ls-remote --exit-code --heads "https://uesr:$GITHUB_TOKEN@github.com/demisto/content-private.git" "refs/heads/$CI_COMMIT_BRANCH" >> /dev/null && PRIVATE_BRANCH_NAME=$CI_COMMIT_BRANCH
+    - "[ -z $PRIVATE_BRANCH_NAME ] && PRIVATE_BRANCH_NAME='master'"
+    - python3 Utils/trigger_private_build.py --github-token $GITHUB_TOKEN --private-branch-name $PRIVATE_BRANCH_NAME
     - sleep 60
     - python3 Utils/get_private_build_status.py --github-token $GITHUB_TOKEN
   timeout: 2 hours
--- a/.gitlab/ci/on-push.yml
+++ b/.gitlab/ci/on-push.yml
@@ -17,8 +17,12 @@ trigger-private-build:
     - if: '$NIGHTLY'
   script:
     - echo "====Trigger Private Build===="
-    - git ls-remote --exit-code --heads "https://uesr:$GITHUB_TOKEN@github.com/demisto/content-private.git" "refs/heads/$CI_COMMIT_BRANCH" >> /dev/null && PRIVATE_BRANCH_NAME=$CI_COMMIT_BRANCH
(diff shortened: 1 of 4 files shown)


Utils/trigger_private_build.py
```diff
@@ -111,10 +116,14 @@ def main():
     # get github token parameter
     arg_parser = argparse.ArgumentParser()
     arg_parser.add_argument('--github-token', help='Github token')
+    arg_parser.add_argument('--private-branch-name', help='Name of the branch in the private repository', default='master')
```
DeanArbel, line 119: Add a note - if not provided, will default on `master` branch.

.gitlab/ci/on-push.yml
```diff
@@ -17,7 +17,9 @@ trigger-private-build:
     - if: '$NIGHTLY'
   script:
     - echo "====Trigger Private Build===="
-    - python3 Utils/trigger_private_build.py --github-token $GITHUB_TOKEN
+    - git ls-remote --exit-code --heads "https://uesr:$GITHUB_TOKEN@github.com/demisto/content-private.git" "refs/heads/$CI_COMMIT_BRANCH" >> /dev/null && PRIVATE_BRANCH_NAME=$CI_COMMIT_BRANCH
```
DeanArbel, line 20: It's not clear what you're trying to achieve here. Please add a note, or extract to a script with a meaningful name.

DeanArbel (dismissed):

DeanArbel (commented):
LGTM
@omerKarkKatz please review as well.

Utils/trigger_private_build.py
```diff
@@ -123,28 +136,40 @@ def main():
...
         # get the workflows ids before triggering the build
-        pre_existing_workflow_ids = get_dispatch_workflows_ids(github_token, 'master')
+        pre_existing_workflow_ids = get_dispatch_workflows_ids(github_token, private_branch_name)
 
         # trigger private build
-        payload = {'event_type': f'Trigger private build from content/{branch_name}',
-                   'client_payload': {'commit_sha1': branch_name, 'is_infra_build': 'True'}}
-
-        res = requests.post(TRIGGER_BUILD_URL,
+        if private_branch_name == 'master':
```
omerKarkKatz, line 142: Is there a shared way to pass the parameters here so it will be on the same fields? 
Now the case is client_payload for dispatch on master, and inputs for the custom private branch. 
This makes a messy check in the private build instead of using the same convention

omerKarkKatz (commented):
Hey over all looks good!
One question, why did you decide to implement the check_if_brach_exist.sh as a bash script and not a python script? 
It is less readable and will be harder to maintain in the future.

Utils/trigger_private_build.py
esharf: the only solution for that is to move to inputs only.
but this will prevent us from giving a title for the build which might be a little annoying since you can’t search it.

esharf:
> Hey over all looks good! One question, why did you decide to implement the check_if_brach_exist.sh as a bash script and not a python script? It is less readable and will be harder to maintain in the future.

there is no need for any installation and will work in almost every environment

DeanArbel (approved):
Looks good

PR #23562, opened 2023-01-05
feat: Lacework integration was updated to use APIv2 [APIv1 Deprecation] 

alannix-lw:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
N/A

## Description
This pull request updates the Lacework integration pack to use the new v2 APIs as the v1 APIs are being deprecated.  New terminology is used to match the Lacework product - utilizing "Alerts" instead of "Events" and implementing new Incident Fields and Incident Types to match.

## Screenshots

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [x] Yes
       - Further details: The response data that is provided by the v2 APIs in Lacework is slightly different than what was provided with the v1 endpoints.  The existing commands were ported over to use the new API, but the returned data will likely require some user interaction.  As such, the major version was bumped, and a breaking change notification was created to notify users.
   - [ ] No

## Must have
- [ ] Tests
- [x] Documentation


efelmandar:
Hi @alannix-lw and thank you for this contribution, I started reviewing your PR and overall the changes looks good.
One thing I do want to mention is that, it's generally best to use existing fields whenever possible to minimize content duplication and make the fields more global (e.g. searchable through system search and shared across incident types). 

However, if you feel that the new fields you created are necessary, please let me know the reasoning and we can discuss further.

here are the existing fields which can be used instead of the ones created:

|Existing field|New Lacework field
|---|---|
Alert Category | Lacework Alert Category
Alert ID  | Lacework Alert ID 
Alert Source | Lacework Alert Source
Alert Sub-Category | Lacework Alert Sub-Category
Alert Type | Lacework Alert Type
Event ID | Lacework Event ID

alannix-lw:
Thanks for the feedback @efelmandar!   I just pushed an update which attempts to use existing Incident Fields - although what I found within my XSOAR instance differed slightly from your recommendations.  I didn't have `Alert Sub-Category` I just had `Sub Category` and I didn't have `Alert Type` I only had `Alert Type ID`.  Can you validate that those are the appropriate types to use?

I would also like to leave the Event ID mapping as Lacework Event ID for the time being - that was the mapping in previous releases, and "Events" in Lacework are going through a bit of an evolution.  Once we have finalized that transition, I can update the plugin to have more appropriate mappings to default incident fields.

efelmandar:
Hi @alannix-lw, Thanks for pushing the changes, I've reviewed them and they look good. You were correct about the fields, so there is no need to change them. Can you let me know what time would be best to schedule a live demo? @ilaner FYI

ilaner:
Hi @alannix-lw,
Haven’t heard from you since I published my review.
Do you need any help with something? Please don’t hesitate to reach out to me here or on Slack.

Thanks again.

ilaner (approved):

PR #23580, opened 2023-01-05
Splitting indicator extraction test playbook to multiple test playbooks.

Ni-Knight:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4905

## Description
Instead of having one single linear test playbooks for all indicator types its been split into multiple ones in accordance with the indicator type. The steps in the playbooks were also split to allow better understanding of what exactly failed when the test playbook ran.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


altmannyarden:
I think we should add those playbooks to the formatters scripts or add split playbooks to the existing test playbooks

Ni-Knight:
@altmannyarden - we shouldn't test things twice, formatters are already checked in unitests and playbooks. Those tests check all the edge cases of the formatter. These tests check the full extraction flow and not only the formatting, adding these to the formatter tests will not add value.

PR #23597, opened 2023-01-08
[ElasticSearchV2] - enhancement and bug fix

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-19716)

## Description
1. add support to run any dsl query
2. fixed an issue where even if page and size are provided, the **search** command will always return the same amount of results when the *query_dsl* argument was provided.

## Must have
- [x] Tests
- [ ] Documentation


Packs/Elasticsearch/Integrations/Elasticsearch_v2/Elasticsearch_v2.py
```diff
@@ -671,15 +673,20 @@ def get_time_range(last_fetch: Union[str, None] = None, time_range_start=FETCH_T
...
     body = {"query": raw_query}
-    response = es.search(index=index or FETCH_INDEX, body=body)
-    return response
+    requested_index = index or FETCH_INDEX
+
+    try:
+        return es.search(index=requested_index, body=body, size=size, from_=page)
+    except Exception as e:
+        demisto.debug(f'got an error when searching with {body}, error:\n{e}')
+        return es.search(index=requested_index, body=raw_query, size=size, from_=page)
```
JudahSchwartz, line 689: looks like in the case that raw_json is provided and not meant to be wrapped we're relying on an exception here. 
Lets do the following instead

```
def execute_raw_query(es, raw_query, index=None, size=None, page=None):
    try:
        raw_query = json.loads(raw_query)
        if raw_query.get('query'):
            demisto.debug('query provided already has a query field. Sending as is')
            return es.search(index=requested_index, body=raw_query, size=size, from_=page)
    except TypeError as e:
        demisto.info(f"unable to convert raw query to dictionary, use it as a string\n{e}")

    body = {"query": raw_query}
    requested_index = index or FETCH_INDEX

    return es.search(index=requested_index, body=body, size=size, from_=page)

```

GuyAfik: looks good, but the catch of the exception should be i think `ValueError` and not `TypeError`, I will catch them both

JudahSchwartz (approved):
nice

Packs/Elasticsearch/ReleaseNotes/1_3_11.md
```diff
@@ -0,0 +1,6 @@
+#### Integrations
+##### Elasticsearch v2
+- Fixed an issue where in the ***search*** and ***es-search*** commands the *page* and *size* were ignored when using the *query_dsl* argument.
```
JudahSchwartz, line 3:
```suggestion
- Fixed an issue where in the ***search*** and ***es-search*** commands, the *page* and *size* arguments were ignored when using the *query_dsl* argument.
```

PR #23768, opened 2023-01-11
Microsoft ATP TPB fix

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [CIAC-5192](https://jira-hq.paloaltonetworks.local/browse/CIAC-5192)

## Description
Switched the id of the software the TPB tried to get since the previous one was deleted.

   - [ ] Yes
       - Further details:
   - [x] No


Packs/MicrosoftDefenderAdvancedThreatProtection/TestPlaybooks/playbook-Microsoft_Defender_Advanced_Threat_Protection-Test.yml
```diff
@@ -1645,7 +1648,7 @@ tasks:
     conditions:
     - label: "yes"
       condition:
-      - - operator: isNotEmpty
+      - - operator: isEmpty
```
ostolero, line 1651: This is not the same condition, was this intentional?

ostolero (commented):
Hi, can we have the same workflow without relaying on a specific software name?
I see we have  a 'microsoft-atp-list-software' command, can we rely on this output?

Packs/MicrosoftDefenderAdvancedThreatProtection/TestPlaybooks/playbook-Microsoft_Defender_Advanced_Threat_Protection-Test.yml
YuvHayun: Since I had to replace the previous software with a new one, I couldn't find any software with vulnerabilities so I had to change it to isEmpty.. I can change it to just take the first element from a list but then then we wouldn't be able to predict the tasks afterwards (which conditions to put)

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
```diff
@@ -6685,4 +6685,4 @@ script:
 defaultmapperin: Microsoft Defender For Endpoint Mapper
 fromversion: 5.0.0
 tests:
-- No tests (auto formatted)
+- Microsoft Defender Advanced Threat Protection - Test
```
ostolero, line 6688: Let's add here also-
Microsoft Defender Advanced Threat Protection - Test dev
Microsoft Defender Advanced Threat Protection - Test self deployed
Microsoft Defender Advanced Threat Protection - Test prod

ostolero (approved):
Ok makes sense, let's just add the other tests in the yml as well :)

PR #23780, opened 2023-01-12
add device-group input to pan-os-commit-configuration playbook

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-19163)

## Description
Added the *device-group* argument as an input for the *pan-os-commit-configuration* playbook.


idovandijk (approved):
Looks good!
BTW no need for the filter `where inputs.device-group is not empty`, but since you tested and it works it's fine to keep it :)

GuyAfik:
> Looks good! BTW no need for the filter `where inputs.device-group is not empty`, but since you tested and it works it's fine to keep it :)

removed the filter :)

PR #23811, opened 2023-01-13
Reimplment precommit

ilaner:


pyproject.toml
```diff
@@ -122,6 +123,44 @@ db-dtypes = "^1.0.1"
...
+    "W605", # invalid-escape-sequence
+]
+
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+# fixable = []
+# unfixable = []
+
+line-length = 120
+target-version = "py310"
+src = ["Packs/Base/Scripts/CommonServerPython", "Tests/demistomock", "Packs/ApiModules/Scripts/*"]
```
dorschw, line 144:
```suggestion
src = ["Packs/Base/Scripts/CommonServerPython/CommonServerPython.py", "Tests/demistomock", "Packs/ApiModules/Scripts/*.py"]
```

ilaner: We need the directory

dorschw (approved):

ilaner:
This needs to be force merged.
Reason for the force merge is the deleted files of `.pre-commit-config.yaml` which will no longer neede and will be generated automatically, and `__init__.py` in the repo root which is not needed and makes `mypy` not identify issues in infra scripts.

PR #23832, opened 2023-01-15
took out force

JudahSchwartz:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: **CIAC-5186**

## Description
A race condition was being caused where editing and then immediately merging a pr would trigger this workflow to run, merge, and then this workflow would ovverride the branch with the master commit. Setting force to false will cause only fast-forwarding changes to be accepted

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


avidan-H (approved):

ShacharKidor:
@JudahSchwartz - do we know what was the reason for setting there the force in the first place? 
I'm asking to make sure we are not messing up some other cases by changing that,

JudahSchwartz:
@ShacharKidor according to @avidan-H  it was there as a sort of "just in case something weird happened" clause

PR #23851, opened 2023-01-16
Fix MITRE name and external_id filtering

BEAdi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-20288

## Description
Because of a change in stix2 python library, using `Filter("external_references.external_id", "=", mitre_id)` and `Filter('name', '=', name)` doesn’t work, because they were supported internally in the library. The only filter that is used in the code and works in this version is `Filter("type", "=", "attack-pattern")`, because it was supported in the request itself (with "added_after", "id", "type" and "version" filters).

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


BEAdi:
Adding that the runtime of the "attack-pattern" command decreased by about a minute compared to the previous version, because I moved the request outside, so now we make one request for all inputs instead of 8 requests for each requested attack pattern (4 times, each with a different name, in which 2 requests where sent every time).

A passing test-playbook after the change for verifying that the indicator that is outputted has the same name as the one asked for:
image

MLainer1 (approved):
🥇

PR #23935, opened 2023-01-19
pan-os command documentation fixes

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-20398

## Description
Improved the docs of the pan-os command type argument.


Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -3601,9 +3601,21 @@ script:
       name: reporttype
     - description: The report name.
       name: reportname
-    - defaultValue: keygen,config,commit,op,report,log,import,export,user-id,version
```
JudahSchwartz, line 3604: double check that this isnt a breaking change

JudahSchwartz (approved):

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
GuyAfik: checked it out and this isn't a breaking change as you can only pass in 1 type per api request and not a list

PR #23953, opened 2023-01-19
URLHaus TPB was fixed

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5331)

## Description
Task id 5: We have changed the condition for DBotScore from checking that it is equal to 3, since the URLHaus sometimes changes the score, resulting in the test Playbook failing, to checking that it is not empty.
Task id 15: This task was removed. Even though the instance was not configured with a proxy, this command would rarely fail and return a 502 (proxy error). Perhaps there is a problem on the other side or a limit to the number of calls in a given period of time. This command ('urlhaus-download-sample') is still called in this test playbook.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


ostolero (approved):
Looks great!

PR #24021, opened 2023-01-23
add-FiltersAndTransformers

bakatzir:
Adding the base/core pack of Filters and transformers to Expanse MP.

## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/EXPANDR-2297


bakatzir:
@yaakovpraisler @daryakoval should I ask review from someone else?

ShahafBenYakir:
Adding a core pack is risky and not enough to run the build. Have we checked about dependencies?

bakatzir:
> Adding a core pack is risky and not enough to run the build. Have we checked about dependencies?

If you'll instruct me what to do - I'll do that. Bare in mind that this is already a core pack for XSOAR and XSIAM MP.

ShahafBenYakir:
Good point regarding it being a core pack.
We need to verify all dependencies are from within this close core packs group. The best will be to run the find dependencies SDK command on the filtersAndTransformers pack to validate that all required dependencies are in this list.

bakatzir:
> Good point regarding it being a core pack. We need to verify all dependencies are from within this close core packs group. The best will be to run the find dependencies SDK command on the filtersAndTransformers pack to validate that all required dependencies are in this list.

@ShahafBenYakir 
```
Found dependencies result for FiltersAndTransformers pack:
{}
```

ShahafBenYakir (approved):

PR #24127, opened 2023-01-26
Fix get identifier

mmhw:
## Status
- [ ] In Progress
- [x] Ready

   - [ ] Yes
       - Further details:
   - [x] No


JasBeilin (commented):
Didn't we said that it should put empty string inside if not found?

mmhw:
> Didn't we said that it should put empty string inside if not found?

Right, I added

JasBeilin (approved):

daryakoval (approved):

esharf (approved):

PR #24228, opened 2023-01-31
Improve Performance of Review Release Notes Workflow

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://jira-hq.paloaltonetworks.local/browse/CIAC-5496

## Description
This PR includes changes to improve the performance of Review Release Notes workflow. 

* Return only release notes from _Get changed files_ step [using glob](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet).

* Added env var `(env.CHANGED_FILES_DELIMITER)` to hold the delimiter of the changed files.

* Added conditional to skip installing the SDK and run the doc review when no release notes were changed.

* Removed usage of `poetry` in the workflow as it was installing unused dependencies from `pyproject.toml` and `poetry.lock` unnecessarily elongating the build. Replaced `poetry` with using `pip`.

The advantages of these changes are:

* We only process files that are relevant to this scope. This prevents issues such as processing file names that include characters that might break the delimiter such as the one we see https://github.com/demisto/content/pull/23278#discussion_r1057776370. 
* In PRs that have no release notes, there's no need to run the rest of the workflow at all. This not only shortens the workflow run time but also releases computing for the rest of the workflows that are run within the runner pool so the rest of the build should be faster as well.
* In continuation to the previous bullet, PRs with a large amount of files, there's a performance penalty for bringing in all PR changed files. This makes the job run unnecessarily long.

The changes above helped take down the build time from ~4m to right under 2m (ignore the other runs, they were for tuning performance):

See https://jira-hq.paloaltonetworks.local/browse/CIAC-5045 for more details.

Test workflows:
- [x] Changed both community and XSOAR-supported Packs (HelloWorld, SentinelOne), [workflow](https://github.com/demisto/content/actions/runs/4105393749/jobs/7082184209) ran and _Install SDK_ and _Review Release Notes_ steps were triggered as expected.
- [x]  No release notes changed, [workflow](https://github.com/demisto/content/actions/runs/4105455623/jobs/7082334410) ran and _Install SDK_ and _Review Release Notes_ steps were skipped as expected.


kgal-pan:
[Secret detection workflow failure](https://github.com/demisto/content/actions/runs/4105455624/jobs/7082334827) caused by https://jira-hq.paloaltonetworks.local/browse/CIAC-5551.

.github/workflows/review-release-notes.yml
```diff
@@ -1,32 +1,38 @@
...
-      - name: Setup Poetry
-        uses: Gr1N/setup-poetry@v7
-      - name: Install Python Dependencies
+          separator: ${{ env.CHANGED_FILES_DELIMITER }}
+          files: |
+            Packs/**/ReleaseNotes/**
+          since_last_remote_commit: false
+
+      - name: Install SDK
+        if: ${{ steps.changed-files.outputs.all_changed_files }}
```
ShahafBenYakir, line 28: what doe4s this condition means?

ShahafBenYakir (approved):
let's wait with the merge to Sunday

.github/workflows/review-release-notes.yml
kgal-pan: This condition is added to two steps (_Install SDK_ and _Review Release Notes_). We're checking whether there are any outputs from the _Get Changed Files_ (which should only include Release Notes with this change) step. 

If not, we skip both _Install SDK_ and _Review Release Notes_ since there's no need to run them if there are no Release Notes to review. 

The elegant solution would be to terminate the job successfully in _Install SDK_ step and not have to have the same condition twice but apparently [there's no built-in way to do so in GitHub Actions yet](https://github.com/actions/runner/issues/662).

See [here](https://github.com/tj-actions/changed-files#outputs) for more info about the `changed-files` output used in the conditional.

ShahafBenYakir: @kgal-pan 
>(which should only include Release Notes with this change)

Not sure I understand why it will include only release notes files, according to docs:
>Returns all changed files i.e. a
combination of all added, copied, modified
and renamed files (ACMR)

kgal-pan: It will only include release notes because of the [glob pattern supplied to the `files` key](https://github.com/demisto/content/blob/CIAC-4970-run-release-notes-check-only-if-xsoar-supported-pa/.github/workflows/review-release-notes.yml#L23-L24).

ShahafBenYakir: great

ShahafBenYakir: great

PR #24295, opened 2023-02-02
Update FileType UT for SDK 1.10.0

dorschw:
following https://github.com/demisto/demisto-sdk/pull/2677. 
The unit tests here will fail here unless using SDK 1.10.0.


dantavori (approved):

dorschw:
implemented in another PR.

PR #24336, opened 2023-02-05
[CommonServerPython] - add retry on api rate limit in send_events_to_xsiam func

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

https://jira-hq.paloaltonetworks.local/browse/XSUP-17281

## Description
Added retry mechanism to handle occasions where sometimes sending events could fail on rate limit. 

## Must have
- [x] Tests
- [ ] Documentation


Packs/PrismaSaasSecurity/Integrations/SaasSecurityEventCollector/SaasSecurityEventCollector.py
```diff
@@ -270,18 +271,24 @@ def main() -> None:  # pragma: no cover
...
-                )
-                demisto.setIntegrationContext({})
-            except Exception as e:
-                demisto.info(f'got error when trying to send events to XSIAM: [{e}]')
-                demisto.setIntegrationContext({'events': events})
-                demisto.info(f'set the following events into integration context: {events}')
+            for i in range(1, 4):
+                try:
+                    demisto.info(f'sending the following amount of events into XSIAM: {len(events)}, retry number {i}')
+                    send_events_to_xsiam(
```
anara123, line 277: I think we better insert this fix to the send_events_to_xsiam function in CommonServerPython script.

GuyAfik: no problem, will be done.

JudahSchwartz (approved):
Nice!

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -8457,9 +8460,117 @@ def test_send_events_to_xsiam_error_handling(self, mocker, requests_mock, error_
...
+                [
+                    (429, None), (429, None), (429, None)
+                ],
+                3,
+                3,
+                False
+            ),
+            (
+                [
+                    (401, None), (401, None)
```
ShahafBenYakir, line 8481: I think we don't use the second item here, why do we need it?
```suggestion
                    (401, None)
```

ShahafBenYakir (commented):
Looks good, see my comments

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10865,13 +10868,29 @@ def events_error_handler(res):
         ).format(xsiam_url=xsiam_url, headers=json.dumps(headers, indent=8), status_code=res.status_code, error=error)
 
         demisto.error(header_msg + api_call_info)
-        raise DemistoException(header_msg + error, DemistoException)
+        raise DemistoException(header_msg + error, DemistoException, res)
```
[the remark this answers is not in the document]
GuyAfik: but the error is a `string`, i prefer to check the `status_code` by using the entire `Response` object. I don't think that the `error` contains the status code.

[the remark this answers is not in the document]
GuyAfik: i use the response object here:

```
        except DemistoException as err:
            if err.res.status_code != 429 or attempt_num == num_of_attempts:
                raise err
            else:
                time.sleep(1)  # pylint: disable=sleep-exists
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10865,15 +10868,34 @@ def events_error_handler(res):
...
-    res = client._http_request(method='POST', full_url=urljoin(xsiam_url, '/logs/v1/xsiam'), data=zipped_data,
-                               headers=headers,
-                               error_handler=events_error_handler)
-    if res.get('error').lower() != 'false':
-        raise DemistoException(header_msg + res.get('error'))
+
+    # retry mechanism in case there is a rate limit (429) from xsiam.
+    for attempt_num in range(1, num_of_attempts + 1):
+        demisto.debug('Sending events into xsiam: attempt number {attempt_num}'.format(attempt_num=attempt_num))
+        ok_codes = (200, 429) if attempt_num < num_of_attempts else ()
```
JudahSchwartz, line 10879: will sending in an empty tuple work like you expect? Should probably just put 200 explicitly

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10865,15 +10868,34 @@ def events_error_handler(res):
...
+            method='POST',
+            full_url=urljoin(xsiam_url, '/logs/v1/xsiam'),
+            data=zipped_data,
+            headers=headers,
+            error_handler=events_error_handler,
+            ok_codes=ok_codes,
+            resp_type='response'
+        )
+        demisto.debug('received status code: {status_code}'.format(status_code=response.status_code))
+        if response.status_code == 200:
```
JudahSchwartz, line 10890: best not to use break. Flip the loop to a do while loop with this being one of the conditions

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: yes its working, cause by default we don't send any tuple

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: changed the implementation to your suggestion :)

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
JudahSchwartz: what do you mean? we will send in ()

JudahSchwartz (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
GuyAfik: changed to `None` cause that is the default when not sending anything to the `ok_codes`, when you don't send anything to `ok_codes` it will fallback to the `_is_status_code_valid` and will go into the `response.ok` function which is:

```
@property
    def ok(self):
        """Returns True if :attr:`status_code` is less than 400, False if not.

        This attribute checks if the status code of the response is between
        400 and 600 to see if there was a client error or a server error. If
        the status code is between 200 and 400, this will return True. This
        is **not** a check to see if the response code is ``200 OK``.
        """
        try:
            self.raise_for_status()
        except HTTPError:
            return False
        return True
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10869,13 +10872,35 @@ def events_error_handler(res):
...
+        )
+        status_code = response.status_code
+        demisto.debug('received status code: {status_code}'.format(status_code=status_code))
+        if status_code == 429:
+            time.sleep(1)
+        attempt_num += 1
+
+    raw_response = response.json()
+    if raw_response.get('error').lower() != 'false':
+        raise DemistoException(header_msg + raw_response.get('error'))
```
ShahafBenYakir, line 10903: Let's extract this whole part into an external function:
```suggestion

def xsiam_api_call_with_retries(num_of_attempts: int):
    # retry mechanism in case there is a rate limit (429) from xsiam.
    status_code = None
    attempt_num = 1

    while status_code != 200 and attempt_num < num_of_attempts + 1:
        demisto.debug('Sending events into xsiam, attempt number {attempt_num}'.format(attempt_num=attempt_num))
        # in the last try we should raise an exception if any error occurred, including 429
        ok_codes = (200, 429) if attempt_num < num_of_attempts else None
        response = client._http_request(
            method='POST',
            full_url=urljoin(xsiam_url, '/logs/v1/xsiam'),
            data=zipped_data,
            headers=headers,
            error_handler=events_error_handler,
            ok_codes=ok_codes,
            resp_type='response'
        )
        status_code = response.status_code
        demisto.debug('received status code: {status_code}'.format(status_code=status_code))
        if status_code == 429:
            time.sleep(1)
        attempt_num += 1

    raw_response = response.json()
    if raw_response.get('error').lower() != 'false':
        raise DemistoException(header_msg + raw_response.get('error'))
        
        
xsiam_api_call_with_retries(num_of_attempts)
```

ShahafBenYakir (approved):
Great

PR #24341, opened 2023-02-05
Adding Log that telling which pack I am installing now

daryakoval:


tkatzir (approved):
Great! Approved.
Maybe:
Pack(s) instead of Packs?

PR #24599, opened 2023-02-14
Downgrade aella fromversion

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [21769159](https://code.pan.run/xsoar/content/-/jobs/21769159)

## Description
Fixed an issue where the integration wasn't supported by lower version machines.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/Aella_StarLight/Integrations/AellaStarLight/AellaStarLight.yml
+++ b/Packs/Aella_StarLight/Integrations/AellaStarLight/AellaStarLight.yml
@@ -90,4 +90,4 @@ script:
   dockerimage: demisto/python3:3.10.10.47713
 tests:
 - No test
-fromversion: 6.8.0
+fromversion: 6.5.0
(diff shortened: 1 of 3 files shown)


yaakovpraisler (approved):

JudahSchwartz (approved):
no reason at all this was set to 6.8?

YuvHayun:
> no reason at all this was set to 6.8?

Nope, it was formatted to the default fromversion.

JudahSchwartz:
@YuvHayun why was our default version not good enough? Should we switch the default?

YuvHayun:
> @YuvHayun why was our default version not good enough? Should we switch the default?

see my updated comment for the link ^
It wasn't good because the pack was already supporting lower versions.

PR #24622, opened 2023-02-15
[Pre-Commit] Add a step to convert line separators to `LF`

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Description
Add a pre-commit step to convert line separators to `LF`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,8 @@ repos:
     -   id: check-toml
     -   id: check-xml
     -   id: check-merge-conflict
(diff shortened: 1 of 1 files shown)


MichaelYochpaz:
@dorschw Git diff shows all of the lines were updated because the file was converted from `CRLF` to `LF`.
I've opened a test PR with changes to a file using `CRLF` and the pre-commit step seems to fail successfully :)

https://github.com/demisto/content/actions/runs/4182345320/jobs/7245321864

PR #24635, opened 2023-02-15
Stix bug

eyalpalo:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-21108)

## Description
added support to extract indicators from stix file when the format is: value valueIN('value1', 'value2', 'value3',..., 'value n').


Packs/Base/Scripts/StixParser/StixParser.py
```diff
@@ -34,6 +34,7 @@
 
 # Pattern Regexes - used to extract indicator type and value
 INDICATOR_OPERATOR_VAL_FORMAT_PATTERN = r"(\w.*?{value}{operator})'(.*?)'"
+INDICATOR_IN_VAL_PATTERN = r"(\w.*?valueIN)\(+('.*?')\)"
```
dantavori, line 37: this regex did not match the example in the test for me, got an idea why? shouldn't it be something like:
```suggestion
INDICATOR_IN_VAL_PATTERN = r"(\w*?value IN )\(+('.*?')\)"
```
not sure what is the exact pattern, please consult with Dror.

Ni-Knight: line 243 - `trimmed_pattern = pattern.replace(" ", "")`

it really confused me when I looked at it 😅

eyalpalo: After consulting with Dror, the value I entered is correct (the spaces are removed before using the regex).
An additional comment was added

dantavori (approved):

PR #24677, opened 2023-02-16
remove non existing command from readme

yucohen:

## Status
- [ ] In Progress
- [x] Ready

## Description
This command was planned to be implemented in some [contrib PR](https://github.com/demisto/content/pull/19445) but then was removed, and we missed to remove it from the readme.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


chrisbalmer:
If https://github.com/demisto/content/pull/24688 is approved and merged, this documentation could be left in. I added the command back in and the docs match up as the original code I believe used code from the O365 integration which is where I also got the code from and just adjusted to fit. So the args and outputs align.

PR #24698, opened 2023-02-17
[Azure Active Directory Groups] authorization_code support

ennozdd:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
added authorization code flow support for delegated permissions
modified apimodule to handle non default scopes when generating proper login urls

## Screenshots
image

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


yucohen (changes requested):
@ennozdd Thanks for your contribution! Great work!
1. Lets not change the API module and try to handle all the changes in the integration code itself.
2. Why is the scopes parameter is needed? The endpoints we reach in the integration are determined, so therefore we can 'hardcode' the scope.

ennozdd:
> @ennozdd Thanks for your contribution! Great work!
> 
>     1. Lets not change the API module and try to handle all the changes in the integration code itself.
> 
>     2. Why is the scopes parameter is needed? The endpoints we reach in the integration are determined, so therefore we can 'hardcode' the scope.

hardcoding in my understanding might give the app more permission than necessary, for example an administrator might want to give permission only to add/remove members from a group or just read access for that matter. If hardcoding is a standard for scope handling that's fine I can revert that change.

yucohen:
@ennozdd We actually using the default scope in the API module for the login url (you can see [this microsoft docs](https://learn.microsoft.com/en-us/azure/active-directory/develop/scopes-oidc#the-default-scope) 
So lets revert this change
If you need help or any consultation please reach out here, or on slack (you can also e-mail me at yucohen@paloaltonetworks.cpm)

yucohen (approved):

PR #24740, opened 2023-02-20
PAN-OS Bugfix

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
relates: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5740)

## Description
When running one of the rule listing commands using a `disabled` filter, the function `filter_rules_by_status` could raise an error in cases where the `disabled` tag data returned by the API is structured differently.

For example: `<disabled>yes</disabled>` in the response will work, while `<disabled @name=test>yes</disabled>` will cause an error.

## Screenshots
Screenshot 2023-02-20 at 12 07 01


Packs/PAN-OS/ReleaseNotes/1_16_10.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+##### Palo Alto Networks PAN-OS
+- Fixed a bug where using the **disabled** filter would not work and return an error in some cases on in the following commands:
```
GuyAfik, line 4:
```suggestion
- Fixed an issue where using the **disabled** filter would not work and return an error in some cases on in the following commands:
```

GuyAfik (approved):
very nice, do you mind adding a unit-test for it?

Packs/PAN-OS/ReleaseNotes/1_16_10.md
MichaelYochpaz:
```suggestion
- Fixed a bug where using the **disabled** filter would not work and return an error in some cases in the following commands:
```

DeanArbel (approved):

PR #24783, opened 2023-02-21
Changed PrivateIPs and InternalDomains lists to system

idovandijk:
## Status
Ready

## Description
Changed the XSOAR lists to `system:true` for server tests.

No


idovandijk:
@liorblob fyi

BenJoParadise (approved):

PR #24886, opened 2023-02-26
InsightVM / Nexpose Bugfixes

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

### Bugs fixed:
- **nexpose-delete-site** and **nexpose-start-site-scan** commands won't work when using the `id` and `site` args respectively.
- **nexpose-start-site-scan** might fail if one of the site's assets is missing the `ip` key in the API response (encountered this while testing the previous bug)


Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -4928,7 +4928,20 @@ def start_site_scan_command(client: Client, site_id: str | None = None, site_nam
 
     else:
         assets = client.get_site_assets(site.id)
-        hosts_list = [asset["ip"] for asset in assets]
+
```
bziser, line 4931: please add test to this use case :)

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -5404,7 +5417,7 @@ def main():  # pragma: no cover
         elif command == "nexpose-delete-vulnerability-exception":
             results = delete_vulnerability_exception_command(client=client, vulnerability_exception_id=args.pop("id"))
         elif command == "nexpose-delete-site":
-            results = delete_site_command(client=client, **args)
+            results = delete_site_command(client=client, site_id=args.pop("id", None), **args)
```
bziser, line 5420: we don't have this command in the test playbook? how didn't it fail there?

MichaelYochpaz: The error is `delete_site_command() received an unexpected keyword arg 'id'`, which happened because we passed the`id` key (because of `**args`) as it's on the YAML file instead of the correct `site_id` key.

There is a test for this command, but at the function level (same for all other tests), so it doesn't use `demisto.args`, we just pass it the parameters.

The test looks like this:
```python
@pytest.mark.parametrize("site_id",
                         [
                             ("1",),
                         ])
def test_delete_site_command(mocker, mock_client: Client, site_id: str):
    http_request = mocker.patch.object(BaseClient, "_http_request", return_value={})
    result = delete_site_command(client=mock_client, site_id=site_id)

    http_request.assert_called_with(
        url_suffix=f"/sites/{site_id}",
        method="DELETE",
        resp_type="json",
    )

    assert result.outputs is None
```

bziser: because of it I asked about teat **playbook** and not unit test :)

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
MichaelYochpaz: Done in a838794102587104526ea99e959186702fb711b9

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -4928,7 +4928,22 @@ def start_site_scan_command(client: Client, site_id: str | None = None, site_nam
...
+
+            # In some cases there is an IP address in the "addresses" field, but not in the "ip" field.
+            elif asset.get("addresses"):
+                for address in asset["addresses"]:
+                    if address.get("ip"):
+                        hosts_list.add(address["ip"])
+
+        hosts_list = list(hosts_list)
+        # Using sort to allow consistent unit-testing. Seems like set -> list conversion results in a random order.
+        hosts_list.sort()
```
bziser, line 4946: if we need this `sort` only for testing, so do it in the test.
No need to add time complexity to the code because of unit testing.

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/test_data/api_mock/client_get_site_assets.json
```diff
@@ -0,0 +1,518 @@
...
+                    {
+                        "name": "http.banner",
+                        "value": "Apache/2.4.25 (Debian)"
+                    },
+                    {
+                        "name": "http.banner.server",
+                        "value": "Apache/2.4.25 (Debian)"
+                    },
+                    {
+                        "name": "http.banner.x-powered-by",
```
bziser, line 73: better to change it to fake names

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
MichaelYochpaz: Fixed be236d36ce6e82f49e52312bf3fcd63912e98170

PR #24915, opened 2023-02-27
Updating "cs-falcon-update-incident-comment" new command

ewelina3007:
This new command has two arguments:
1. ids -> mandatory:true, is array:true, description:"A comma-separated list of incident IDs."
2. comment -> mandatory:true, description:"A comment which should be added to the CrowdStrike incident."

### DIFF
--- a/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
+++ b/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
@@ -2896,6 +2896,16 @@ script:
     - contextPath: CrowdStrike.VulnerabilityHost.cve.id
       description: Unique identifier for a vulnerability as cataloged in the National Vulnerability Database (NVD).
       type: String
+  - arguments:
+    - description: A comma-separated list of incident IDs.
+      isArray: true
+      name: ids
+      required: true
+    - description: A comment added to the CrowdStrike incident.
+      name: comment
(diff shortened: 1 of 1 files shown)


dansterenson:
Duplication of this: https://github.com/demisto/content/pull/24915
Closing

yucohen (approved):

yucohen (changes requested):
@ewelina3007 Thanks for your contribution! 
Lets set up a short demo (it can also be a video attached to the PR) to demonstrate the new command

ewelina3007:
@yucohen yes we can schedule a call and I can present usage of the command. I will appreciate if you could schedule a call during business hours in CET timezone. Would be great if you could add to the meeting my company email address: ewelina.mozejko@euroclear.com

ewelina3007:
Actually, I will upload a video tomorrow with the show case - so the call is not necessary :)

ewelina3007:
Hello @yucohen,
Unfortunately, I cannot record the video as I would then need to disclosure some CrowdStrike incident information (teams does not allow me to present only part of the window to hide e.g. hostname info). Hence I agreed with the manager that we can show you on the live demo instead of creating the recording. I hope that you understand that - sorry! . I will appreciate if you could schedule a call during business hours in CET timezone. Would be great if you could add to the meeting my company email address: ewelina.mozejko@euroclear.com

yucohen:
@ewelina3007 Sent you an invite
Let me know if that works for you :)

ewelina3007:
@yucohen date works for me, see you tomorrow :)

PR #25005, opened 2023-03-02
[InsightVM / Nexpose] "duration" Error Bugfix

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-21812

## Description
So basically, it seems like the `nexpose-start-site-scan` command starts a scan, and then returns the scan that was just created by making another API call to get the scan (which can be manually ran with `nexpose-get-scan`) on the new scan ID.

What I assume happens (couldn't validate because we can't run this command on our instance), is that since the scan is still running, the `duration` key, specifying how long the scan took, is missing.
I was unable to find any mention of this key being optional on API responses on the [docs](https://help.rapid7.com/insightvm/en-us/api/index.html#operation/getScan) whatsoever.

I've added checks if the `duration` key is available before using it.


Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -2475,7 +2475,11 @@ def normalize_scan_data(scan_data: dict) -> dict:
         include_none=True,
     )
 
-    result["TotalTime"] = readable_duration_time(scan_data["duration"])
+    if "duration" in scan_data:
+        result["TotalTime"] = readable_duration_time(scan_data["duration"])
+
+    else:
+        result["TotalTime"] = "No duration data was found."
```
MichaelYochpaz, line 2482: @bziser I've checked and debugged the old integration by checking out commit 6d2266fd29997611a64988f9362faa349e8ab4c5, and it seems like previously, if the `duration` key was missing, the value of `TotalTime` ended up being `0 minutes`.

I thought returning a message (since its value is a string anyway) saying it couldn't be found would be more accurate, but I guess it could be considered as breaking BC.

Please let me know what you think.
(see PR's description for more context about the bug and why I believe it happened)

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -2475,7 +2475,11 @@ def normalize_scan_data(scan_data: dict) -> dict:
         include_none=True,
     )
 
-    result["TotalTime"] = readable_duration_time(scan_data["duration"])
+    if "duration" in scan_data:
+        result["TotalTime"] = readable_duration_time(scan_data["duration"])
```
DeanArbel, line 2479: If duration is present but is empty, this will also raise an error. This should cover both cases.
```suggestion
    if scan_data.get("duration"):
        result["TotalTime"] = readable_duration_time(scan_data["duration"])
```

DeanArbel (approved):
This solution looks ok, I have just one small adjustment.

PR #25041, opened 2023-03-05
[NetskopeCollector] Support up to 30k events per type

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-21081

## Description
Added support for fetching more than 1 page per event type (up to 3).
Fixed an issue in the v2 implementation where the time filter was malformed.


Shellyber (approved):
@DeanArbel Looks good.
Just need to fix the import error from the test file

PR #25106, opened 2023-03-06
Filter Fixes for 'Cortex XDR - Get File Path from alerts by hash' pla…

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5905

## Description
To support cases in which there are multiple SHA256s within the playbook input, the 'Equals' filter operation used in the playbook tasks should be changed to the 'In list' filter operation.

## Screenshots
An example of the current filters used by the playbook tasks:
Screenshot 2023-03-06 at 21 40 03


Packs/CortexXDR/ReleaseNotes/4_10_20.md
```diff
@@ -0,0 +1,6 @@
+
+#### Playbooks
+
+##### Cortex XDR - Get File Path from alerts by hash
+
+- Playbook now supports the retrieval of file paths for multiple SHA256 file hashes.
```
ssokolovich, line 6: "This playbook..."

ssokolovich (approved):

ssokolovich (approved):

PR #25107, opened 2023-03-06
Update DragosWorldview.py to break loop on edge case

aanton660:
total_pages from the /indicators endpoint can be 0, which causes an infinite loop in pagination as raw_response['page'] (1,2,3, ..., N) will never be equal to 0 to break the loop. This occurs when there have been no IOCs published from WorldView in the last 48 hours. 

example query response with total_pages being 0 (as no new IOCs have been published in last 48 hours)

2023-03-06 10_04_32-Dragos WorldView API and 8 more pages - Work - Microsoft​ Edge


aanton660:
@yucohen thanks for the feedback! I'm actually a consumer of the product, not a developer, so I do not have a development environment set up to run the command. Would you be able to update the release notes for me? Glad to assist however I can with the process.

yucohen:
@aanton660 Sure, I'll do that for you

yucohen (approved):

PR #25134, opened 2023-03-07
Updated email regex and formatter to allow idna

Ni-Knight:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5885

## Description
Fixes email extraction when the email address is part of a URL query. Also fixed the regex to allow none ASCII (IDNA) domains.
See here for new regex - https://regex101.com/r/t09fC7/3

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


Packs/CommonScripts/Scripts/ExtractEmailFormatting/ExtractEmailFormatting.py
```diff
@@ -20,16 +20,21 @@ def extract_email(email_address: str) -> str:
...
                               "(?:(?:\\\\|\^{3})u[a-f\d]{4})?"
                               "([\w.!#$%&'*+/=?^_`{|}~-]{1,64}"
                               "\[?@]?[\w.-]{1,255}(?:\[?\.]?"
                               "[A-Za-z]{2,}){1,2})", re.IGNORECASE)
 
-    try:
-        return re.findall(email_format, email_address)[0]
+    email_address = re.match(email_format, email_address)
 
-    except IndexError:
```
Ni-Knight, line 32: re.match will always return, either a string or `None` so no need to catch `IndexError` it is done by the new `if` clause in line 35.

Packs/Base/TestPlaybooks/playbook-Email_extraction_test.yml
```diff
@@ -447,23 +447,23 @@ tasks:
...
     id: "79"
-    taskid: a8a8ccfd-d429-4dd6-8c4a-5e4d971c1588
+    taskid: 4adf005c-e997-442f-8cf2-7f7d03833c5b
     type: condition
     task:
-      id: a8a8ccfd-d429-4dd6-8c4a-5e4d971c1588
+      id: 4adf005c-e997-442f-8cf2-7f7d03833c5b
       version: -1
-      name: non extraction - お@example.com
+      name: non ascii - お@example.com
```
Ni-Knight, line 455: The email `お@example.com` was checked as invalid, due to the change allowing international addresses every char above U+0072 is valid within the local-part and domain of an email address (See https://www.rfc-editor.org/rfc/rfc6531). We've adjusted the regex and now the formatter and test-playbook.

PR #25158, opened 2023-03-08
[PAN-OS] Fetch Incidents Job Query Bug

samuelFain:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22003

## Description
This PR will fix a PAN-OS fetch incidents bug where job queries for certain log types that take longer to retrieve than others were sent even if the jobs were not completed, resulting in missing incidents.

## Screenshots

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -1324,6 +1324,20 @@ def get_job_status(job_id: str, target: Optional[str] = None):
 
     return result.get('response', {}).get('result', {}).get('job', {}).get('status', '')
 
+def verify_job_complete(job_id: str, max_tries: int):
+    """Verify a Panorama job is completed.
+    Will wait 1 second between each try and will try a maximum of max_tries.
+
+    Args:
+        job_id (str): Panorama job id.
+        max_tries (int): maximum number of tries to check the job status.
+    """    
+    for _ in range(max_tries):
+        if get_job_status(job_id) == 'FIN':
+            return
+        else:
+            time.sleep(1)
+
 
 @logger
 def panorama_push_status(job_id: str, target: Optional[str] = None):
@@ -13092,10 +13106,7 @@ def get_query_entries(log_type: str, query: str, max_fetch: int) -> List[Dict[An
     demisto.debug(f'{job_id=}')
 
     # verify the job is finished before proceeding
-    for _ in range(GET_JOB_ID_MAX_RETRIES):
-        if get_job_status(job_id) == 'FIN':
-            break
-        time.sleep(1)
+    verify_job_complete(job_id, GET_JOB_ID_MAX_RETRIES)
 
     # second http request: send request with job id, valid response will contain a dictionary of entries. 
     query_entries = get_query_entries_by_id_request(job_id)
--- a/Packs/PAN-OS/Integrations/Panorama/Panorama.py
+++ b/Packs/PAN-OS/Integrations/Panorama/Panorama.py
@@ -52,6 +52,7 @@ QUERY_DATE_FORMAT = '%Y/%m/%d %H:%M:%S'
 FETCH_DEFAULT_TIME = '1 day'
 MAX_INCIDENTS_TO_FETCH = 100
 FETCH_INCIDENTS_LOG_TYPES = ['Traffic', 'Threat', 'URL', 'Data', 'Correlation', 'System', 'Wildfire', 'Decryption']
+GET_JOB_ID_MAX_RETRIES = 15
 
 XPATH_SECURITY_RULES = ''
 DEVICE_GROUP = ''
@@ -1295,6 +1296,33 @@ def panorama_push_to_template_stack_command(args: dict):
     else:
         # no changes to commit
         return_results(result['response']['msg']['line'])
+
+
+def get_job_status(job_id: str, target: Optional[str] = None):
+    """Get Panorama job status.
+
+    Args:
+        job_id (str): Panorama job id.
+        target (Optional[str], optional): target command argument (if available). Default is None.
+
+    Returns:
+        string: a sting representing the job status.
+    """        
+    params = {
+        'type': 'op',
+        'cmd': f'<show><jobs><id>{job_id}</id></jobs></show>',
+        'key': API_KEY
+    }
+    if target:
+        params['target'] = target
+
+    result = http_request(
+        URL,
+        'GET',
+        params=params
+    )
+
+    return result.get('response', {}).get('result', {}).get('job', {}).get('status', '')
 
 
 @logger
@@ -13063,7 +13091,13 @@ def get_query_entries(log_type: str, query: str, max_fetch: int) -> List[Dict[An
     job_id = get_query_by_job_id_request(log_type, query, max_fetch)
     demisto.debug(f'{job_id=}')
 
-    # second http request: send request with job id, valid response will contain a dictionary of entries.
+    # verify the job is finished before proceeding
+    for _ in range(GET_JOB_ID_MAX_RETRIES):
+        if get_job_status(job_id) == 'FIN':
+            break
+        time.sleep(1)
+
+    # second http request: send request with job id, valid response will contain a dictionary of entries. 
     query_entries = get_query_entries_by_id_request(job_id)
 
     # extract all entries from response


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13055,7 +13056,19 @@ def get_query_entries_by_id_request(job_id: str) -> Dict[str, Any]:
...
+        response = http_request(URL, 'GET', params=params)
+        status = response.get('response', {}).get('result', {}).get('job', {}).get('status', '')
+        demisto.debug(f'Job ID response status: {status}')
+        if status == 'FIN':
+            demisto.debug(f'{response=}')
+            return response
+        else:
+            demisto.debug('Job not completed, Retrying in 1 second...')
+            time.sleep(1)
+    return {}
```
GuyAfik, line 13071:
```suggestion
    try_number = 1
    while try_number <= GET_LOG_JOB_ID_MAX_RETRIES:
        response = http_request(URL, 'GET', params=params)
        status = response.get('response', {}).get('result', {}).get('job', {}).get('status', '')
        demisto.debug(f'Job ID {job_id} response status: {status}, {try_number=}')
        if status == 'FIN':
            demisto.debug(f'{response=}')
            return response
        else:
            demisto.debug(f'Job {job_id} has not completed, Retrying in 1 second...')
            time.sleep(1)
        try_number += 1
    return {}
```

please extract the `job_id` as well and log it out.
i think maybe logging the response would also be good even if its not finished, would help us in the future in case of a bug.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -52,6 +52,7 @@
 FETCH_DEFAULT_TIME = '1 day'
 MAX_INCIDENTS_TO_FETCH = 100
 FETCH_INCIDENTS_LOG_TYPES = ['Traffic', 'Threat', 'URL', 'Data', 'Correlation', 'System', 'Wildfire', 'Decryption']
+GET_JOB_ID_MAX_RETRIES = 15
```
GuyAfik, line 55:
```suggestion
GET_LOG_JOB_ID_MAX_RETRIES = 15
```

GuyAfik (commented):
@samuelFain nicely done
1) can you please add UT(s) for this?
2) can you please add RN and bump packmetadata?

Packs/PAN-OS/Integrations/Panorama/Panorama.py
samuelFain: Done.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
samuelFain: I added `job_id` to the log and made changes so that the response will be logged in both cases (finished or not).
Not sure how the "while" loop is better than the "for" loop in this situation...

GuyAfik: its just more clean instead of doing 

```
for _ in range(...)
```

in my opinion, but you don't have to accept it :)

GuyAfik: if log the try number then for option is the better one.

```
for try_num in range(1, GET_LOG_JOB_ID_MAX_RETRIES + 1)....
```

samuelFain: Looks good, added the `try_num` to the log 👍

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13055,7 +13056,19 @@ def get_query_entries_by_id_request(job_id: str) -> Dict[str, Any]:
...
     """
     params = assign_params(key=API_KEY, type='log', action='get', job_id=job_id)
-    return http_request(URL, 'GET', params=params)
+    
+    # if the job has not finished, wait for 1 second and try again (until success or max retries)
+    for try_num in range(1, GET_LOG_JOB_ID_MAX_RETRIES + 1):
+        response = http_request(URL, 'GET', params=params)
+        status = response.get('response', {}).get('result', {}).get('job', {}).get('status', '')
+        demisto.debug(f'Job ID {job_id}, response status: {status}')
+        demisto.debug(f'{response=}')
```
ShacharKidor, line 13065:
```suggestion
        demisto.debug(f'raw response: {response}')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13055,7 +13056,19 @@ def get_query_entries_by_id_request(job_id: str) -> Dict[str, Any]:
...
+        if status == 'FIN':
+            return response
+        else:
+            demisto.debug(f'{try_num=}. Job not completed, Retrying in 1 second...')
```
ShacharKidor, line 13069:
```suggestion
            demisto.debug(f'Attempt number: {try_num}. Job not completed, Retrying in 1 second...')
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13055,7 +13056,19 @@ def get_query_entries_by_id_request(job_id: str) -> Dict[str, Any]:
...
+            time.sleep(1)
```
ShacharKidor, line 13070: @yuvalbenshalom - would appreciate your opinion here about using the 1 sec sleep. 
The logic here is kind of internal polling which is limited by the GET_LOG_JOB_ID_MAX_RETRIES.
Please advice.

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
```diff
@@ -6215,6 +6215,26 @@ def test_get_parsed_incident_entries(
...
+    def test_get_query_entries_by_id_request(self, mocker, response, debug_msg, expected_result):
+        """
+        Given:
+        - a valid query id
+
+        When:
+        - get_query_entries_by_id_request function is called
+
+        Then:
+        - assert that the returned response is valid, or retry if the job is not finished
```
ShacharKidor, line 6230: Please explain here in a few words about each tested case, 
Like: 
"""
        Given:
        1.  A valid query id with a success (FIN status)
        2. A valid query id... 
        3. etc
        When:
        - get_query_entries_by_id_request function is called
        Then:
        assert that the returned response is valid, or retry if the job is not finished
        Verify that: 
        1. ...
        2. ...
        3. ...
        4. ...

Packs/PAN-OS/ReleaseNotes/1_16_15.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Fixed an issue where job queries for certain log types that take longer to retrieve than others were sent even if the jobs were not completed, resulting in missing incidents.
```
ShacharKidor, line 6:
```suggestion
- Fixed an issue in the **Fetch incidents** mechanism where job queries for certain log types that take longer to retrieve than others were sent even if the jobs were not completed, resulting in missing incidents.
```

ShacharKidor (changes requested):
@samuelFain - Great job!

Added a few minor suggestions.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13055,7 +13056,19 @@ def get_query_entries_by_id_request(job_id: str) -> Dict[str, Any]:
...
+    return {}
```
ShacharKidor, line 13071:
```suggestion
    demisto.debug(f'Maximum attempt number: {try_num} has reached. Job ID {job_id} might be not completed which could result in missing incidents.')
    return {}
```

Packs/PAN-OS/Integrations/Panorama/Panorama_test.py
samuelFain: Done.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
samuelFain: Done.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
samuelFain: Done.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
yuvalbenshalom: @samuelFain this can be problematic. if we may wait 10+ sec, considering other delays in the fetch, we may hit the 60 sec in some cases.

why don't we save the Ids in the context and re try them in the next fetch? we are doing a similar flow in splunkpy when getting drill down on fetch. you can check this out.

samuelFain: Due to the short job life, the context solution wont work for longer fetch intervals.
For now we will handle this by lowering the max retries value.

ShacharKidor (approved):

PR #25173, opened 2023-03-09
St GitHub iam test fix

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5501

## Description
Allowed status code 404 in get_users() function

## Screenshots

   - [x] No


Packs/GitHub/Integrations/GitHub_IAM/GitHub_IAM.py
```diff
@@ -24,6 +24,7 @@ def get_user(self, input_type, user_term):
         return self._http_request(
             method='GET',
             url_suffix=uri,
+            ok_codes=[200, 404]
```
dantavori, line 27: Actually it will make the regular iam-get-user won't work correctly when we search a non-existing user. so instead I recommend putting a try-except block in the test-module itself.

dantavori (approved):

PR #25174, opened 2023-03-09
Fix microsoft cas duplicate incidents

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-21861)

## Description
1. fixed an issue where the occurred time of an incident was not in a UTC time format while we query the api with a timestamp that is based on UTC format.
2. fixed an issue where we didn't consider the miliseconds of the last incident that occurred in the next fetch which caused incidents to be re-fetched in case all of them occurred at the same second (but not at the same milisecond).
3. now 1 milisecond will be added for the last incident fetched timestamp to avoid such a case in the future. 

for more info read the issue, there is a very detailed explanation there. 

## Must have
- [x] Tests
- [ ] Documentation


Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
```diff
@@ -664,9 +661,30 @@ def list_users_accounts_command(client: Client, args: dict):
...
+    if fetch_start_time_datetime.microsecond == 0:  # type: ignore
+        return int(start_fetch_timestamp) * 1000
+
+    timestamp = str(start_fetch_timestamp).replace('.', '')
+    if len(timestamp) == 11:
+        return int(timestamp) * 100
+    elif len(timestamp) == 12:
+        return int(timestamp) * 10
+    elif len(timestamp) == 13:
+        return int(timestamp)
```
MLainer1, line 675: I'll try to think of a clever way for it, as it can create a bug in the future

Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
```diff
@@ -664,9 +661,30 @@ def list_users_accounts_command(client: Client, args: dict):
...
+
+
+def timestamp_to_datetime_string(timestamp: int, include_miliseconds: bool = True):
```
MLainer1, line 678: Can we use `timestamp_to_datestring` from csp instead?

MLainer1 (commented):
Great

Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
GuyAfik: yes, good idea

Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
GuyAfik: i agree its not the best practice, however i have unit-tests that check it works as expected in all kinds of forms of timestamps.

see the `test_format_fetch_start_time_to_timestamp` unit-test.

in my case there could be 3 situations:

1. the miliseconds of the date is 0 (multiply by 1000)
2. the miliseconds of that date is of form: <number>00 (multiply by 100)
3. the miliseconds of that date is of form: <number><number>0 (multiply by 10)
4. the miliseconds of that date is of form: <number><number><number> (no need to multiply)

the api knows to work **only** with 13 digits timestamps (which include also the milliseconds)

if you still got a better idea, i'd be happy to hear :)

Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
GuyAfik: done

JudahSchwartz:
@GuyAfik please explain the change being made here in the description

GuyAfik:
> @GuyAfik please explain the change being made here in the description

done

JudahSchwartz (approved):
looks good! nice catch

Packs/MicrosoftCloudAppSecurity/Integrations/MicrosoftCloudAppSecurity/MicrosoftCloudAppSecurity.py
GuyAfik: changed

PR #25178, opened 2023-03-09
Added the vsys param to the register_ip_tag command

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22176

## Description
Added the vsys param to the following commands:
- register_ip_tag command
- unregister_ip_tag_command
- register_user_tag command
- unregister_user_tag command

   - [x] No


Shellyber:
I tested it locally in our env - it seems that the API does recognize the vsys param in this command (it will also throw an error if the vsys is not valid) 
I can't test a change since we only have vsys1
I will send it to the customer to verify

Shellyber:
@ShahafBenYakir / @GuyAfik Ready for review

ShahafBenYakir:
@Shellyber please add a verification for the vsys param inside the unit test.

Shellyber:
@ShahafBenYakir done

GuyAfik:
@Shellyber how will this behave if we run those commands against panorama instance and not a firewall?

Shellyber:
@GuyAfik The behavior is the same behavior, it takes the default vsys, even if it's not passed to the command. 
So empty string or nothing at all its the same.
But I'll update to keep the same behavior.

GuyAfik (approved):

PR #25233, opened 2023-03-12
[TimeComponents] Fix Unit-Test Failure

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5982)

## Description
Fix unit-test failure.
See JIRA ticket for context about the issue.


MichaelYochpaz:
Keeping this as draft atm as I need to check whether "CDT" is actually a valid value or not.

MichaelYochpaz:
After some debugging, the "CDT" abbreviation (as returned by the pytz lib) is actually "Cuba Daylight Time" and NOT "Central Daylight Time" as Google returns when looking up "CDT Time".

So, that means the "CDT" value is actually valid.

bziser (approved):

PR #25327, opened 2023-03-16
add rate limit handling

jthom-vmray:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
~~fixes: link to the issue~~

## Description
The app could not handle rate limiting, which lead to failing playbook runs. This PR introduces a new setting that handles rate limiting, if set.

## Screenshots
~~Paste here any images that will help the reviewer~~

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

## Must have
- [x] Tests
- [x] Documentation


Packs/VMRay/Integrations/VMRay/VMRay.py
```diff
@@ -142,10 +148,40 @@ def find_error(may_be_error_inside):
...
+        seconds_to_wait = random.uniform(1.0, 5.0)
+        try:
+            seconds_to_wait += float(r.headers.get("Retry-After", 0))
+        except ValueError:
+            pass
+
+        demisto.debug(f"Rate limit exceeded! Waiting {seconds_to_wait} seconds and then retry.")
+        time.sleep(seconds_to_wait)  # pylint: disable=sleep-exists
+        reset_file_buffer(files)
+        return http_request(method, url_suffix, params, files, ignore_errors, get_raw)
```
dansterenson, line 183: In this case, it would be better to assign the response to a variable and then continue with the logic of the function. 
```suggestion
        r = http_request(...)
```

Why are the request args different from the one above? 
```suggestion
        r = requests.request(
        method, url, params=params, headers=HEADERS, files=files, verify=USE_SSL, proxies=PROXIES
    )
```

Packs/VMRay/Integrations/VMRay/README.md
```diff
@@ -17,6 +17,7 @@ The Playbooks accelerate incident response and make security operations more sca
     | API Key | True |
     | Use system proxy settings | False |
     | Trust any certificate (not secure) | False |
+    | Retry requests when API is rate limited |  | False |
```
dansterenson, line 20:
```suggestion
    | Retry requests when API is rate limited | False |
```

Packs/VMRay/Integrations/VMRay/VMRay.py
```diff
@@ -1,3 +1,6 @@
+import random
+import time
+
```
dansterenson, line 3:
```suggestion
import urllib3
```

Packs/VMRay/Integrations/VMRay/VMRay.py
```diff
@@ -10,14 +13,15 @@
...
+RETRY_ON_RATE_LIMIT = demisto.params().get("retry_on_rate_limit", True)
 SERVER += '/rest/'
 USE_SSL = not demisto.params().get('insecure', False)
 HEADERS = {'Authorization': 'api_key ' + API_KEY}
 ERROR_FORMAT = 'Error in API call to VMRay [{}] - {}'
 RELIABILITY = demisto.params().get('integrationReliability', DBotScoreReliability.C) or DBotScoreReliability.C
 
 # disable insecure warnings
-requests.packages.urllib3.disable_warnings()
+requests.packages.urllib3.disable_warnings()  # type: ignore[attr-defined]  # pylint: disable=no-member
```
dansterenson, line 24:
```suggestion
# Disable insecure warnings
urllib3.disable_warnings()
```

Packs/VMRay/Integrations/VMRay/VMRay_test.py
```diff
@@ -122,3 +124,35 @@ def test_build_finished_job(mocker):
...
+                }
+            }
+        ]
+    )
+
+    from VMRay import http_request
+    start = time.perf_counter()
+    response = http_request("GET", "submission/123")
+
+    assert 2.0 < time.perf_counter() - start <= 7.0
```
dansterenson, line 157: How does it ensure that the rate limit logic is working? It would be easier to check that the _http_request() function was called twice.

Packs/VMRay/ReleaseNotes/1_1_7.md
```diff
@@ -0,0 +1,3 @@
+#### Integrations
+##### VMRay
+- Handle API rate limiting.
```
dansterenson, line 3:
```suggestion
- Added a new *retry_on_rate_limit* parameter to handle a rate limit http error.
```

dansterenson (changes requested):
Nice and thanks for your contribution. Please see my notes.

Packs/VMRay/Integrations/VMRay/VMRay.py
jthom-vmray: The arguments differ, because it is a recursive function call and not a direct request. We need to call `http_request` recursively here, since it is possible that retrying only once could end up in a rate limit that was triggered by another waiting request.

Packs/VMRay/Integrations/VMRay/VMRay_test.py
jthom-vmray: Good catch! I did not think of using a `call_count`. I will change that.

Packs/VMRay/Integrations/VMRay/VMRay.py
dansterenson: Missed the recursion. So what is the stop condition? How do you make sure you won't enter an infinite loop and get a stack overflow? you should count the number of tries.

dansterenson (changes requested):

dansterenson (approved):
Nice

PR #25332, opened 2023-03-16
Ignore domain look alike from email user

Ni-Knight:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22404

## Description
At the moment if an email has a user name that looks like a domain we will extract it as such. This change should fix this issue by grabbing the "@" sign with the user name. That way the formatter knows its a user-name and not a domain and drops it.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
+++ b/Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
@@ -1,6 +1,16 @@
 import demistomock as demisto
 from CommonServerPython import *
 from CommonServerUserPython import *
+import json
+
+def string_to_markdown(text):
+    text = json.loads(text)
+    md = ''
+    for key, values in text.items():
+        md += f"### {key}\n"
+        for value in values:
+            md += f"- {value}\n"
+    return md
 
 
 def read_file_with_encoding_detection(filePath, maxFileSize):
@@ -39,7 +49,7 @@ def extract_indicators_from_file(args):
         'Type': entryTypes['note'],
         'ContentsFormat': formats['text'],
         'Contents': indicators_hr,
-        'HumanReadable': indicators_hr
+        'HumanReadable': string_to_markdown(indicators_hr)
     }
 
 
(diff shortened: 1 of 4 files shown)


Packs/CommonScripts/ReleaseNotes/1_11_55.md
```diff
@@ -0,0 +1,10 @@
+
+#### Scripts
+
+##### ExtractIndicatorsFromTextFile
+- Updated the code for a better human readable response.
+- Updated the Docker image to: *demisto/python3:3.10.10.49934*.
+
+##### ExtractDomainAndFQDNFromUrlAndEmail
+- Fixed a typo in the comments.
```
Shellyber, line 9:
```suggestion
```
No need for this line

Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
```diff
@@ -1,6 +1,35 @@
 import demistomock as demisto
 from CommonServerPython import *
 from CommonServerUserPython import *
+import json
+
+
+def string_to_markdown(indicators: str) -> str:
```
Shellyber, line 7: Why do we need this function?
tableToMarkdown won't do the trick?

Shellyber (commented):
Nice work

Packs/CommonScripts/Scripts/ExtractIndicatorsFromTextFile/ExtractIndicatorsFromTextFile.py
Ni-Knight: I tried, it didnt work for me :(
Maybe I've done it wrong?

Ni-Knight: Its, not a table though, its a list so even if it works it will end up looking bad.

Ni-Knight: @Shellyber

PR #25392, opened 2023-03-20
TAXII Server 2 - Updated the way the server serves file hashes and other SCOs

Ni-Knight:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6028

## Description
Correct handling of specific SCOs like files, ASN and win key registry.

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


Packs/Base/ReleaseNotes/1_31_80.md
```diff
@@ -0,0 +1,6 @@
+#### Scripts
+
+##### CommonServerPython
+
+ - Updated the "FeedIndicatorType" class adding `Software` as an indicator.
+ -
```
ShahafBenYakir, line 6:
```suggestion
 Updated the "FeedIndicatorType" class adding `Software` as an indicator.

```

Packs/TAXIIServer/ReleaseNotes/2_0_30.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### TAXII2 Server
+- Updated the Docker image to: *demisto/flask-nginx:1.0.0.50558*.
+
+- Fixed an issue where the server served files, ASN and Windows registry key indicators wrong.
```
ShahafBenYakir, line 7:
```suggestion
- Fixed an issue where the integration served File, ASN, and Windows registry key indicators in a wrong way.
```

Packs/TAXIIServer/Integrations/TAXII2Server/TAXII2Server.py
```diff
@@ -821,6 +823,54 @@ def create_stix_object(xsoar_indicator: dict, xsoar_type: str, extensions_dict:
     return stix_object, extension_definition, extensions_dict
 
 
+def build_sco_object(stix_object: Dict[str, Any], xsoar_indicator: Dict[str, Any]) -> Dict[str, Any]:
```
ShahafBenYakir, line 826: Can you add a unit test to this specific function?

Ni-Knight: Yes, on it.

Packs/TAXIIServer/Integrations/TAXII2Server/TAXII2Server_test.py
```diff
@@ -589,10 +589,29 @@ def test_convert_sco_to_indicator_sdo_with_type_file(mocker):
 
     output = convert_sco_to_indicator_sdo(ioc, xsoar_indicator)
     assert 'file:hashes.' in output.get('pattern', '')
-    assert 'MD5' in output.get('pattern', '')
+    assert 'SHA-1' in output.get('pattern', '')
     assert 'pattern_type' in output.keys()
 
 
+def test_build_sco_object(mocker):
```
ShahafBenYakir, line 596: Since you asked me last time we talked (and I agree) it would be a bit better to use `parameterized`. It will allow each use case to be independent such that if one use case fails the rest will still run, and also it makes it pretty convenient to explain the different cases you are testing in each run. 
And so it is not very different than what you did:
```python
@pytest.mark.parametrize('indicator, sco_indicator',[
    (xsoar_indicators[0], sco_indicators[0])
    (xsoar_indicators[1], sco_indicators[1])
    (xsoar_indicators[2], sco_indicators[2])
])
def test_build_sco_object(indicator, sco_indicator):
    """
        Given
            Case 1: xsoar indicator which is of type A (whatever is special in this case)
            Case 2: xsoar indicator which is of type B
            Case 3: xsoar indicator which is of type C
        When
            Running build_sco_object
        Then
            Case 1: validate that the resulted object has all hashes bla bla
            Case 2: validate that the resulted object does not have hashes bla bla
            Case 3: ...
    """
    output = build_sco_object(indicator["stix_type"], indicator["xsoar_indicator"])
    assert output == sco_indicator
``` 

This is just an example that needs to be fine-tuned (probably fix variable namings and complete docstring) and tested.

Ni-Knight: Thank you!

ShahafBenYakir (approved):

PR #25448, opened 2023-03-22
added -gt instead != in pipelines stop util

esharf:
## Status
- [x] Ready

## Description
this PR makes sure that we are canceling only older builds.


eyalpalo (approved):

bziser:
@esharf can you add in the description of the PR what are you trying to do here and why?

esharf:
> @esharf can you add in the description of the PR what are you trying to do here and why?

Done

PR #25467, opened 2023-03-23
split large gzip files to smaller chunks

thefrieddan1:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5771

## Description
XSIAM has an upper limit for sending events of upto 10 Mib 
We implement a batching mechanism, where the limit is ~1 Mib per payload.

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -38,10 +38,12 @@ def __line__():
     return cf.f_back.f_lineno  # type: ignore[union-attr]
 
 
-# 42 - The line offset from the beggining of the file.
+# 43 - The line offset from the beggining of the file.
 _MODULES_LINE_MAPPING = {
     'CommonServerPython': {'start': __line__() - 43, 'end': float('inf')},
 }
+
+IDEAL_XSIAM_FILE_SIZE = 2**20
 
 
 def register_module_line(module_name, start_end, line, wrapper=0):
@@ -10854,6 +10856,24 @@ def xsiam_api_call_with_retries(
     return response
 
 
+def round_up(x):
+    return int(x) + int((x > 0) and (x - int(x)) > 0)
+
+
+def is_zipped_data_size_ideal(zipped_data):
+    return sys.getsizeof(zipped_data) <= IDEAL_XSIAM_FILE_SIZE
+
+
+def split_xsiam_events(data, size_of_zipped_data):
+    number_of_list = round_up(size_of_zipped_data / IDEAL_XSIAM_FILE_SIZE)
+    data_list = str.split(data, '\n') if isinstance(data, str) else data
+    chunk_size = round_up(len(data_list) / number_of_list)
+    return [
+        data_list[i: i + chunk_size]
+        for i in range(0, len(data_list), chunk_size)
+    ]
+
+
 def send_events_to_xsiam(events, vendor, product, data_format=None, url_key='url', num_of_attempts=3):
     """
     Send the fetched events into the XDR data-collector private api.
@@ -10961,15 +10981,30 @@ def send_events_to_xsiam(events, vendor, product, data_format=None, url_key='url
 
     zipped_data = gzip.compress(data.encode('utf-8'))  # type: ignore[AttributeError,attr-defined]
     client = BaseClient(base_url=xsiam_url)
+    if is_zipped_data_size_ideal(zipped_data):
+        send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
+                                   zipped_data)
+        demisto.updateModuleHealth({'eventsPulled': amount_of_events})
+    else:
+        demisto.info(f'will split data to chunks of {IDEAL_XSIAM_FILE_SIZE} to send to XSIAM server.')
+        data_chunks = split_xsiam_events(data, sys.getsizeof(zipped_data))
+        for data_chunk in data_chunks:
+            amount_of_events = len(data_chunk)
+            data_chunk = '\n'.join(data_chunk)
+            zipped_data = gzip.compress(data_chunk.encode('utf-8'))  # type: ignore[AttributeError,attr-defined]
+            send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
+                                       zipped_data)
+            demisto.updateModuleHealth({'eventsPulled': amount_of_events})
 
+
+def send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
+                               zipped_data):
     raw_response = xsiam_api_call_with_retries(
         client, xsiam_url, zipped_data, headers, num_of_attempts, events_error_handler
     ).json()
-
     if raw_response.get('error').lower() != 'false':
         raise DemistoException(header_msg + raw_response.get('error'))
-
-    demisto.updateModuleHealth({'eventsPulled': amount_of_events})
+    return raw_response
 
 
 def is_scheduled_command_retry():


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -38,11 +38,13 @@ def __line__():
     return cf.f_back.f_lineno  # type: ignore[union-attr]
 
 
-# 42 - The line offset from the beggining of the file.
+# 43 - The line offset from the beggining of the file.
```
thefrieddan1, line 41: not mandatory change. it is a comment. I have changed the actual number in the code from 42 to 43 in the past and i now see that i did not change the number in the comment.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10854,6 +10856,24 @@ def xsiam_api_call_with_retries(
     return response
 
 
+def round_up(x):
+    return int(x) + int((x > 0) and (x - int(x)) > 0)
```
bziser, line 10860: why not:
```suggestion
import math

def round_up(x):
    return math.ceil(x)

```

bziser (commented):
please add inputs types and return types to the functions

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10854,6 +10856,24 @@ def xsiam_api_call_with_retries(
...
+def split_xsiam_events_to_chunks(data, size_of_zipped_data):
+    num_of_sub_lists = round_up(size_of_zipped_data / XSIAM_EVENT_CHUNK_SIZE)
+    data_list = str.split(data, '\n') if isinstance(data, str) else data
+    chunk_size = round_up(len(data_list) / num_of_sub_lists)
+    return [
+        data_list[i: i + chunk_size]
+        for i in range(0, len(data_list), chunk_size)
+    ]
+
+
```
bziser, line 10876: maybe these functions need to be Internal function like `def events_error_handler(res)`? 
because we won't want that integration could call them, only `send_events_to_xsiam`, no?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10986,15 +11009,45 @@ def events_error_handler(res):
...
+    :param headers: A dictionary containing headers for the request.
+    :type num_of_attempts: ``int``
+    :param num_of_attempts: The number of attempts to send the data to XSIAM.
+    :type xsiam_url: ``str``
+    :param xsiam_url: The URL of the XSIAM server.
+    :type zipped_data: ``bytes``
+    :param zipped_data: The zipped data to send.
+
+    :return: Bool or DemistoException
+    :rtype: ``Bool`` or ``DemistoException``
```
bziser, line 11043: what about the type of `raw_response`?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10986,15 +11009,45 @@ def events_error_handler(res):
...
+    :type headers: ``dict``
+    :param headers: A dictionary containing headers for the request.
+    :type num_of_attempts: ``int``
+    :param num_of_attempts: The number of attempts to send the data to XSIAM.
+    :type xsiam_url: ``str``
+    :param xsiam_url: The URL of the XSIAM server.
+    :type zipped_data: ``bytes``
+    :param zipped_data: The zipped data to send.
+
+    :return: Bool or DemistoException
```
bziser, line 11042: `return` is the same as `rtype`?

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -7,7 +7,7 @@
 import sys
 import warnings
 
-import dateparser
+# import dateparser
```
bziser, line 10: why not to remove?

bziser:
adding @DeanArbel to review also :)

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
thefrieddan1: should not have been left un commented. 
will revert it.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
thefrieddan1: for some reason i dont understand yet, it parses the docstring successfully with rtype and fails with return. 
In validations when it runs the commonServerDocs.sh script.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10986,15 +11009,45 @@ def events_error_handler(res):
...
     zipped_data = gzip.compress(data.encode('utf-8'))  # type: ignore[AttributeError,attr-defined]
     client = BaseClient(base_url=xsiam_url)
+    data_chunks = split_xsiam_events_to_chunks(data, sys.getsizeof(zipped_data))
+    for data_chunk in data_chunks:
+        amount_of_events = len(data_chunk)
+        data_chunk = '\n'.join(data_chunk)
+        zipped_data = gzip.compress(data_chunk.encode('utf-8'))  # type: ignore[AttributeError,attr-defined]
+        send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
+                                   zipped_data)
+        demisto.updateModuleHealth({'eventsPulled': amount_of_events})
```
DeanArbel, line 11019: This should actually be called just once after the for loop with the sum total of events created.

DeanArbel (changes requested):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10956,6 +10959,26 @@ def send_events_to_xsiam(events, vendor, product, data_format=None, url_key='url
 
     header_msg = 'Error sending new events into XSIAM.\n'
 
+    def split_xsiam_events_to_chunks(data, size_of_zipped_data):
+        """
+        Splits a list of XSIAM events into chunks of size XSIAM_EVENT_CHUNK_SIZE.
+
+        :type data: ``list or str``
+        :param data: A list of XSIAM events or a string containing XSIAM events separated by newline characters.
+        :typ: size_of_zipped_data: ``int``
```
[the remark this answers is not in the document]
thefrieddan1: how do i run the docstring check locally ?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10879,6 +10882,31 @@ def xsiam_api_call_with_retries(
...
+    :param data: The string of data to split.
+    :type chunk_size: ``int``
+    :param chunk_size: The maximum size of each chunk.
+
+    :return: A list of chunks containing the data.
+    :rtype: ``list``
+    """
+    chunk = []
+    curr_size = 0
+    for data_part in data.split('\n'):
```
DeanArbel, line 10899: Shouldn't this function also support data that's a `list`?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10879,6 +10882,31 @@ def xsiam_api_call_with_retries(
...
+        chunk.append(data_part)
+        curr_size += sys.getsizeof(data_part)
+        if curr_size > chunk_size:
```
DeanArbel, line 10902:
```suggestion
        if curr_size >= chunk_size:
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10879,6 +10882,31 @@ def xsiam_api_call_with_retries(
     return response
 
 
+def split_data_to_chunks(data, chunk_size):
```
DeanArbel, line 10885: Let's change `chunk_size` to `target_chunk_size`

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10879,6 +10882,31 @@ def xsiam_api_call_with_retries(
...
+    """
+    Splits a string of data into chunks of a specified size.
```
DeanArbel, line 10887:
```suggestion
    Splits a string of data into chunks of an approximately specified size. The actual size can be lower.
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
DeanArbel, line 10902: If you check the size after the addition, the size will always be greater than the `target_chunk_size` (except for the last iteration).
Let's switch the order a bit, and instead check the size before addition - if it's greater, then `yield` the data, if not then add it and continue.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10879,6 +10882,31 @@ def xsiam_api_call_with_retries(
...
+
+    :type data: ``str``
+    :param data: The string of data to split.
+    :type chunk_size: ``int``
+    :param chunk_size: The maximum size of each chunk.
+
+    :return: A list of chunks containing the data.
+    :rtype: ``list``
```
DeanArbel, line 10895: It returns an iterable of a list and not a list. I think that `collections.Iterable[list]` might be the right option.
https://www.jetbrains.com/help/pycharm/type-syntax-for-docstrings.html#return-type

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10984,17 +11012,47 @@ def events_error_handler(res):
...
+    for data_chunk in data_chunks:
+        amount_of_events += len(data_chunk)
+        data_chunk = '\n'.join(data_chunk)
+        zipped_data = gzip.compress(data_chunk.encode('utf-8'))  # type: ignore[AttributeError,attr-defined]
+        send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
+                                   zipped_data)
+    demisto.updateModuleHealth({'eventsPulled': amount_of_events})
+
 
+def send_to_xsiam_with_retries(client, events_error_handler, header_msg, headers, num_of_attempts, xsiam_url,
```
DeanArbel, line 11027: I think this doesn't need to be its own function (especially because it has a big signature that just gets passed to another function as is). Perhaps we can add this functionality via an optional argument to the existing `xsiam_api_call_with_retries()` e.g. `return_type='json'`, which will also add an implicit error handling - I'm not 100% sure that's the right way to go either.

I'll leave it up to you to decide how to handle this

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
DeanArbel: After adding this change, `curr_size` can also be changed to `chunk_size`.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
thefrieddan1: in  send_events_to_xsiam flow
If data is a list it is joined into a str.
So data is a string.

I can however not do the join to a string if data is a list.
But then if data is a string will I split it to a list ?

Or should I check in the split_data_to_chunks method if data is a list or a string ? 
I prefer defined data structures. at least in terms of collections vs. str or int 
But I understand it might not be a valid preference in python.
What do you say ?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10875,10 +10883,39 @@ def xsiam_api_call_with_retries(
         if status_code == 429:
             time.sleep(1)
         attempt_num += 1
-
+    if is_json_response:
+        response = response.json()
+    if response.get('error', '').lower() != 'false':
+        raise DemistoException(header_msg + response.get('error'))
```
DeanArbel, line 10889: The error check can only be done for json, so this check should be done only for that case.
```suggestion
        if response.get('error', '').lower() != 'false':
            raise DemistoException(header_msg + response.get('error'))
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10842,15 +10847,18 @@ def xsiam_api_call_with_retries(
 
     :type headers: ``dict``
     :param headers: headers for the request
+    
+    :type header_msg: ``str``
+    :param header_msg: The header message.
```
DeanArbel, line 10852:
```suggestion
    :param header_msg: The error message prefix in case of an error.
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
DeanArbel: Python is very tolerable for both. In this case, it'll make much more sense for this function to get a `list` as an input rather than an `str` with a `'\n'` delimiter - so the default case should be a `list` rather an `str` - if you prefer to support just one usecase, I suggest to support a `list`.
Of course it'll be easy to support both usecases as the `for` loop iterates a list - making it support both just require an extra if before the `for` loop.

DeanArbel (changes requested):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -10826,7 +10829,9 @@ def xsiam_api_call_with_retries(
     zipped_data,
     headers,
     num_of_attempts,
-    events_error_handler=None
+    error_msg='',
+    events_error_handler=None,
```
DeanArbel, line 10833: Let's keep the original order of the args.
```suggestion
    events_error_handler=None,
    error_msg='',
```

DeanArbel (approved):
Looks good, I have just a couple of notes.

bziser (approved):

ShahafBenYakir:
@DeanArbel @thefrieddan1 @bziser 
Please check the `amount_of_events` variable it is declared twice and is getting overwritten. We should remove the unnecessary usage.

PR #25592, opened 2023-03-29
Un-Deprecate Sec&Comp V1, Improve troubleshooting docs

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: :hide-the-pain:

## Description
For some customers, V1 may work where V2 does not. For this reason, we have chosen to un-deprecate the integration.
Also added some documentation to help with troubleshooting.


Packs/MicrosoftExchangeOnline/Integrations/SecurityAndCompliance/README.md
```diff
@@ -45,6 +43,33 @@ To access the Security & Compliance Center, the user account needs to be a globa
 7. Click **Done**.
    
 
+### Enabling Client Side Basic Authentication
+
```
yuvalbenshalom, line 47: please add a disclaimer here that say that it is mandatory to enable basic auth and at the moment Microsoft are not providing any other way in order to make the delegated commands work.

yuvalbenshalom (commented):
see my comment

yuvalbenshalom (approved):

PR #25665, opened 2023-04-02
[AWS] - fix issue with SessionDuration parameter

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22984

## Description
It seems like the `sessionDuration` parameter in AWS integration can become empty string instead of `None` which causes failures in commands that do not pass this argument, make sure even if empty string is shown, then the value of the parameter will be `None` to avoid failing on `ValueError`.

## Must have
- [x] Tests
- [ ] Documentation


GuyAfik:
failed unit-test

MLainer1 (approved):
🍍

JudahSchwartz (approved):

PR #25683, opened 2023-04-02
Moved the Reporter Email Address field from Phishing to CommonTypes pack

idovandijk:
## Status
Ready

## Related Issues
Moving the field because of https://jira-hq.paloaltonetworks.local/browse/CIAC-4697

## Description
Moved the `Reporter Email Address` field from the Phishing pack to the CommonTypes pack.
This is needed because the field needs to be used in an upcoming pack, and Phishing is no longer a core pack.

Yes, but not really because CommonTypes is a Core pack


AradCarmi (approved):

idovandijk:
@dantavori Could you please force merge this PR?
The reason force merge is needed is because I moved a field from Phishing to CommonTypes. The purpose of this is to use the field in an upcoming pack without creating a dependency on Phishing.

PR #25770, opened 2023-04-09
added coreRestApi integration to conf.json

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

Demisto rest Api integration is deprecated, using Core Rest API instead which contain the same commands.


ostolero:
Looks good, don't forget to update the google secret as well.
But do it as close to merging as possible as this affects nightly tests as well.

PR #25828, opened 2023-04-13
Dou collector missing events

omerKarkKatz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-23092

## Description
Bug in Duo collector that might be related to loss of events.
demisto.setLastRun is called before the sending of the events to XSIAM.
That means that if the send-events-to-xsaim will fail the last_run already changed and the events will be lost.

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/DuoAdminApi/ReleaseNotes/3_2_10.md
+++ b/Packs/DuoAdminApi/ReleaseNotes/3_2_10.md
@@ -1,0 +1,6 @@
+
+#### Integrations
+
+##### Duo Event Collector
+
+- Fixed an issue caused by updating the last_run before sending events to XSIAM. This may lead to loss of events upon a failuer. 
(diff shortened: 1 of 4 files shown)


Packs/DuoAdminApi/ReleaseNotes/3_2_10.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### Duo Event Collector
+
+- Fixed an issue caused by updating the last_run before sending events to XSIAM. This may lead to loss of events upon a failuer.
```
DeanArbel, line 6:
```suggestion
- Fixed a rare issue where events were missing, caused by a failed creation of events in XSIAM. In the future, in cases where events fail to be created, they will be fetched again on the next run.
```

PR #25853, opened 2023-04-16
Test Playbook - Cortex XDR Malware - Incident Enrichment

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5512

## Description
A test playbook for the ‘Cortex XDR Malware - Incident Enrichment’ playbook.

## Screenshots
Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0
- [ ] 6.8.0


Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+    nexttasks:
+      '#default#':
+      - "140"
+      Verified:
+      - "136"
+    separatecontext: false
+    conditions:
+    - label: Verified
+      condition:
+      - - operator: isExists
```
ssokolovich, line 1875: I think we should change the `isExist` --> `isNotEmpty`
I am afraid that in case of a `null,` it will take the "verified" path instead of the "ELSE".
So if you see that it creates issues leave it as it is ( I want to be sure that we can handle another alert from xdr in case this specific will be deleted). If that works ok - maybe we should change that in the whole PB.

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+      condition:
+      - - operator: isEqualString
+          left:
+            value:
+              simple: incident.deviceosname
+            iscontext: true
+          right:
+            value:
+              complex:
+                root: Endpoint
```
ssokolovich, line 2898: In case we have 2 endpoints with the same name for example - we might get more than 1 endpoint. and then it will be a list.

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+      Verified:
+      - "136"
+    separatecontext: false
+    conditions:
+    - label: Verified
+      condition:
+      - - operator: isEqualString
+          left:
+            value:
+              simple: incident.isolated
```
ssokolovich, line 2989: I'm not sure we are setting this in the incident enrichment PB. Can you check pls that we are indeed setting this up through the PB?

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+        }
+      }
+    note: false
+    timertriggers: []
+    ignoreworker: false
+    skipunavailable: false
+    quietmode: 0
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "164":
```
ssokolovich, line 2339: can we move this task and other "account" context key checks to the Account section below?
image

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+    separatecontext: false
+    conditions:
+    - label: Verified
+      condition:
+      - - operator: isExists
+          left:
+            value:
+              complex:
+                root: PaloAltoNetworksXDR.Incident.alerts
+                accessor: '[0]'
```
ssokolovich, line 420: If that is an array, can we check if there are any alerts there? I prefer not to point to only 1 element in the array.

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
```diff
@@ -0,0 +1,4565 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "182":
+    id: "182"
+    taskid: 3511a1b2-fd06-4375-87a4-af6be9ca684d
+    type: condition
+    task:
+      id: 3511a1b2-fd06-4375-87a4-af6be9ca684d
+      version: -1
+      name: Verify Device Status
```
ssokolovich, line 2923: duplicated with task 207. 
Note that it is not the only one... also 183<->200
and 178<->204 and some more

ssokolovich (commented):
Add some more. The most important comment -> we have some duplicated tasks from the first row of checks and the other dedicated sections (like the endpoint info).
Let's remove those that are on the first row, and keep the tasks that sectionalized :)

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
TalNos: changed to 'PaloAltoNetworksXDR.Incident.alerts' and used the 'Is not empty' condition.

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
TalNos: Removed the task from the 'check incident fields'

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
TalNos: Replaced all conditions of the 'endpoint context data' check to 'In'

Packs/CortexXDR/TestPlaybooks/Test_Playbook_-_Cortex_XDR_Malware_-_Incident_Enrichment.yml
TalNos: Removed the following duplicated tasks:
 1- 'Verify Device Status' - task number 182
2- 'Verify Isolated' - task number 183
3- 'Verify Device Name' - task number 179
4- 'Verify Device MAC Address' -task number  178
5- 'Verify Device OS Name' - task number 180
6- 'Verify Device ID' - task number 176
7 - 'Verify Device Local IP' - task number 177

ssokolovich (approved):

Packs/CortexXDR/ReleaseNotes/4_10_38.md
```diff
@@ -0,0 +1,6 @@
+
+#### Playbooks
+
+##### Cortex XDR Malware - Incident Enrichment
+
+- Internal code improvements.
```
ssokolovich, line 6: pls remove the `-` at the beginning

PR #25913, opened 2023-04-17
SlackV3: fix missing user messages

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22898

## Description
a wrong condition cause the bug that user message was considered as bot message and ignored

### DIFF
--- a/Packs/Slack/Integrations/SlackV3/SlackV3.py
+++ b/Packs/Slack/Integrations/SlackV3/SlackV3.py
@@ -1280,7 +1280,7 @@ def is_bot_message(data: dict) -> bool:
         return True
     elif event.get('subtype') == 'bot_message':
         return True
-    elif data.get('authorizations', [{}])[0].get('is_bot', False):
+    elif not (event.get('user') or data.get('user', {}).get('id') or data.get('envelope_id')):
         return True
     else:
         return False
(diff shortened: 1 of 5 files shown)


Packs/Slack/Integrations/SlackV3/SlackV3.py
```diff
@@ -1280,7 +1280,7 @@ def is_bot_message(data: dict) -> bool:
         return True
     elif event.get('subtype') == 'bot_message':
         return True
-    elif data.get('authorizations', [{}])[0].get('is_bot', False):
+    elif not (event.get('user') or data.get('user', {}).get('id') or data.get('envelope_id')):
```
amshamah419, line 1283: Looks good. One thing we can also do is to make sure that the User string doesn't start with "B". For bots, a user ID is always in the format BXXX123. It's just another little check we can do.

amshamah419 (approved):
Looks good to me, just see my comments.

Packs/Slack/ReleaseNotes/3_1_31.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### Slack v3
+
+- Fixed an issue where the Slack listener missed user messages.
```
amshamah419, line 6:
```suggestion
- Fixed an issue where the Slack listener would miss some user messages when mirroring an incident.
```

amshamah419 (approved):

PR #25975, opened 2023-04-19
Qualys: update PB to try get the scan metadata after creation

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-23455

## Description
As there is cases where the PB launch a Qualys scan and after that when we run the GenericPolling PB
it fail due to the `qualys-report-list` command return nothing 
i added 2 changes
1. in case `qualys-report-list` run with a given id and there is no results - raise exception
2. in the Launch_And_Fetch_Scan_Based_Findings_Report_-_Qualys playbook i added task to try get the scan after it created, with 3 retries in case of error (5 sec interval)

## Screenshots
image


DeanArbel (approved):
Very nice

DeanArbel (approved):

Packs/qualys/Playbooks/playbook-Launch_And_Fetch_Scan_Based_Findings_Report_-_Qualys.yml
```diff
@@ -155,6 +155,45 @@ tasks:
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "6":
+    id: "6"
+    taskid: 86f0e348-c8e9-4d57-845b-e77ccabb919d
+    type: regular
+    task:
+      id: 86f0e348-c8e9-4d57-845b-e77ccabb919d
+      version: -1
+      name: Try to get the created scan metadata
```
altmannyarden, line 179: Please change to `Get the created scan metadata`.
Also, we said - you need to add a conditional task after this task to verify that you got the scan ID. if you didnt (after the retries) you will NOT enter the polling subplaybook.

Packs/qualys/ReleaseNotes/1_2_6.json
```diff
@@ -0,0 +1 @@
+{"breakingChanges":true,"breakingChangesNotes": "The command qualys-report-list will return error in case of no results returned for a given report id"}
```
altmannyarden, line 1: why does this change consider as breaking changes?

altmannyarden: oh, the change in the integration :) all good.

PR #26048, opened 2023-04-23
Creating Test Playbooks Checking Empty Inputs

AdiPeret:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-6258)
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6447

## Description
Creating New Test Playbooks to test what happens when providing empty inputs
Fixes a minor bug in the Unisolate Endpoint - Cybereason where the input.hostname is defined returns true when the actual hostname field is empty.

- [x] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [X] No


idovandijk (changes requested):
Great job on everything. I'll write my comments below. Some 

### General Question
What are the changes done in the TestData folder? If irrelevant please revert (PyCharm > Right click on the file > Git > Rollback)

### Block IP - Generic V3_Test
It seems that in this test, the tested data was inputted directly to the subplaybook. That means that in your case we can't use these inputs as we want to test it with empty inputs:
image

I assume that you copied the subplaybook when you added your test, so it was copied with the modified inputs. We actually need to import it from the Playbook Library to ensure that it's using the **default inputs** configured for it:
image

### Calculate Severity - Generic v2 - Test
In this case, you kept the default inputs which is usually correct, but since the existing test is also using the default inputs, I suggest to remove the `CriticalUsers`, `CriticalEndpoints`, `CriticalGroups` inputs. Important note about removing input:
If you simply delete the value, it will later change back to the playbook's default, so this:
image
will become this:
image
See [this issue ](https://github.com/demisto/etc/issues/47507) if interested.
What we should do to make the playbook take empty inputs in this case, is give it some key that doesn't exist in context. This will simulate customer environments where they could have removed the default inputs, or get them from some other place in the context.
You can do something like this:
image

### Detonate File - Generic Test
Excellent job.

### Detonate URL - Generic Test
Excellent job as well.

### Domain Enrichment - Generic v2
Good job. Just a note: True is different than the default (yours had its default changed on your server)
image
however, for our test it doesn't really matter whether it's True or False because there won't be a domain.

### Endpoint Enrichment - Generic v2.1 - Test
Looks good!

### Extract Indicators From File - Generic v2 - Test
Great

### File Enrichment - Generic v2 - Test
Great

### Get endpoint details - Generic - test
Great

### IP Enrichment - Generic v2 - Test
Great

### Retrieve File from Endpoint - Generic V3 Test
Great

### URL Enrichment - Generic v2 - Test
Great


If you have any questions you can DM me or ask here :)

Packs/CommonPlaybooks/TestPlaybooks/playbook-Retrieve_File_from_Endpoint_-_Generic_V3_Test.yml
```diff
@@ -0,0 +1,307 @@
+id: Retrieve File from Endpoint - Generic V3 Test
```
idovandijk, line 1: Did you intend for this to be a new test for v3? If so, please add also to `conf.json` with the relevant test playbook and integration needed.

Packs/Cybereason/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "Cybereason",
     "description": "Endpoint detection and response to manage and query malops, connections and processes.",
     "support": "partner",
-    "currentVersion": "2.1.5",
```
idovandijk, line 5: Let's make this a revision since the fix is very small. To do this you can manually edit this file to 2.1.6, and manually rename the `2_2_0.md` release notes file to `2_1_6.md`.

AdiPeret:
As there is no change in the logic of the testing I don't see a reason to
update to V3

בתאריך יום ד׳, 3 במאי 2023, 13:49, מאת Ido van Dijk ‏<
***@***.***>:

Packs/Cybereason/Playbooks/playbook-Unisolate_Endpoint_-_Cybereason.yml
```diff
@@ -1,32 +1,50 @@
 id: Unisolate Endpoint - Cybereason
-version: -1
+version: 6
```
idovandijk, line 2: Looks like **format** did not work earlier on this file. Possibly because of the issue with the environment you had.
Please format this file, then commit + push. The format will remove all these unnecessary fields that got added, and also bring back the original `fromversion` and `tests` properties of the file.

idovandijk (changes requested):
Looks good, needs a couple of small changes.

Changes applied

Packs/Cybereason/Playbooks/playbook-Unisolate_Endpoint_-_Cybereason.yml
AdiPeret: Done

Packs/Cybereason/pack_metadata.json
AdiPeret: Done

Packs/Cybereason/ReleaseNotes/2_1_6.md
```diff
@@ -0,0 +1,6 @@
+
+#### Playbooks
+
+##### Unisolate Endpoint - Cybereason
+
+- Fixed a minor bug in the Unisolate Endpoint - Cybereason where the input.hostname is defined returns true when the actual hostname field is empty.
```
ShirleyDenkberg, line 6:
```suggestion
Fixed a minor bug in the Unisolate Endpoint - Cybereason where the input.hostname returned true when the actual hostname field was empty.
```

ShirleyDenkberg:
@michalgold @idovandijk Doc review completed.

Packs/Cybereason/Playbooks/playbook-Unisolate_Endpoint_-_Cybereason.yml
```diff
@@ -144,5 +157,8 @@ inputs:
   playbookInputQuery:
 outputs: []
 tests:
-- Cybereason Test
+- No tests (auto formatted)
```
idovandijk, line 160: Looks like the SDK `format` command did not keep the existing test. It should have:
https://github.com/demisto/etc/issues/47716
https://jira-hq.paloaltonetworks.local/browse/CIAC-531

feel free to +1 my Jira issue about this, otherwise we are going to accidentally remove a lot of tests :)
regarding your PR, please bring back the test `Cybereason Test`

AdiPeret: done

AdiPeret:
Please check again, I think we are done.

idovandijk (approved):
Excellent job. NG is probably failing due to build issue.

PR #26097, opened 2023-04-27
ServiceNowV2: use the sys_id as the unique field for last_run

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-23283

## Description
Problem: 
when store the last fetch incident ids in the last run - we used the `name` of the incident as the unique id
in some case the name aren't unique

Fix: 
use the `sys_id` field from the tickets as the unique id, it should appear in all the SNOW table entries 
see [snow documentation](https://docs.servicenow.com/en-US/bundle/utah-platform-administration/page/administer/table-administration/concept/c_UniqueRecordIdentifier.html)


DeanArbel (approved):
Good job

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -2157,6 +2157,11 @@ def fetch_incidents(client: Client) -> list:
 
     severity_map = {'1': 3, '2': 2, '3': 1}  # Map SNOW severity to Demisto severity for incident creation
 
+    # remove duplicate incidents which were already fetched
+    tickets_response = filter_incidents_by_duplicates_and_limit(
+        incidents_res=tickets_response, last_run=last_run, fetch_limit=client.sys_param_limit, id_field='sys_id'
+    )
```
MLainer1, line 2163: Just a question, why did you move the incidents filtering above the tickets parsing? Is there a meaningful reason that would be beneficial for more integrations that use the lookback feature as well.

ilappe: i move it because the field `name` was configured as the `id field` but in some case the `name` aren't unique 
so i used the `sys_id` instead
and `sys_id` was in the `tickets_response`

PR #26243, opened 2023-05-02
CrowdStrike Falcon Intel Feed Actors: add offset to last run

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-23889

## Description
Currently the offset is not saved in case there is more then `limit` indicators 
save the offset and made small refactoring


DeanArbel (approved):
Beautiful implementation.

PR #26352, opened 2023-05-07
Deprecate Mandiant feed

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6516

## Description
- Deprecate the Manidant Feed
- Hide the deprecated pack

   - [x] No


bakatzir (commented):
Let's add
hidden: true
in a PR tomorrow, so the update will be uploaded to the bucket.

bakatzir (approved):

PR #26361, opened 2023-05-07
PreproccessEmail: correct the if condition

ilappe:
## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-24033)

## Description
when the emailsubject field of the incident are `None` the script fail due to incorrect if condition

## Screenshots
**updated unit test failed before the fix**
image

**updated unit test pass after the fix**
image


DeanArbel (approved):
Looks good

PR #26456, opened 2023-05-11
added core packs lists as required to be up-to-date from master

dantavori:
image


Tests/Marketplace/core_packs_list.json
```diff
@@ -1,5 +1,6 @@
 {
   "core_packs_list": [
+    "SplunkPy",
```
ShahafBenYakir, line 3: This is testing, right?

ShahafBenYakir (changes requested):
I am "requesting changes" only so we will not accidentally merge the SplunkPy changes, once removed you can send me again.

ShahafBenYakir (approved):

PR #26405, opened 2023-05-09
Revert bad docker

JudahSchwartz:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: link to the issue

## Description
A few sentences describing the overall goals of the pull request's commits.

## Screenshots
Paste here any images that will help the reviewer

## Minimum version of Cortex XSOAR
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

## Does it break backward compatibility?
   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation

PR #26558, opened 2023-05-16
Update Hackuity integration: API key auth, Hackuity.Findings.Status.LastClosedAt

Rogdham:
## Status

- [x] Ready

## Description
This PR makes the following changes:
1. Authentication is done by API key instead of user+password. This simplifies the code quite a lot. This is a breaking change.
2. New data is added in output: `Hackuity.Findings.Status.LastClosedAt`
3. Update logo

- [x] 6.0.0

   - [x] Yes
       - Further details: Authentication was previously done with user+password, now through an API key. Some details 

## Must have
- [x] Tests
- [x] Documentation


Rogdham:
Summary of the checks in error below.

This PR is now ready to review!

---

### The docker image tag is not the latest numeric tag, please update it.

I did not change it since it was been reverted in #26405

### [IN129] - You've removed integration parameters, the removed parameters are '{'login'}'

This has been documented as a breaking change.

### [IN116] - You've added required, the field is 'apikey'

This has been documented as a breaking change.

thefrieddan1 (changes requested):
Hi @Rogdham,
thank you for your contribution we appreciate your effort, I reviewed the content in the PR and would like to provide some feedback:
Because the changes you introduce here are breaking changes I would recommend 2 possible options :
1. Make the login and password hidden in the yaml file. ( do not remove them )
    And keep the apikey changes with slight modifications. ( example for hidden apikey: 
https://github.com/demisto/content/blob/master/Packs/AbuseDB/Integrations/AbuseDB/AbuseDB.yml
2. Or keep the breaking changes and add a note to the README file. Basically adding the breakingChangesNotes part from the json in the README as well. example : 
https://github.com/demisto/content/blob/master/Packs/MicrosoftGraphSecurity/ReleaseNotes/2_1_21.md

The advantage of option 1 is that it does not break compatibility to users already using the integration. New users configuring the integration will not see the hidden user and password since its hidden and will configure it with the new apikey.

The advantage of option 2 is that its easier to add a note in the README.
But it might make existing users of the integration dealing the the breaking change.
So its up to you to decide how to proceed from here.

Rogdham:
Hello @thefrieddan1 and thank you for your review!

We have chosen option 2, tell me if the wording change in the release notes looks better to you!

thefrieddan1 (approved):
LGTM

PR #26624, opened 2023-05-18
Fix mismatch versions bug

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6621

## Description
Added an api call to sync marketplace. The platform is caching packs versions, so if we install a build bucket on a machine and later on use it on another build, we can try installing a pack with a higher version than existing in the production bucket.


daryakoval (approved):
Great job, let this XSIAM work.

PR #26665, opened 2023-05-21
Fix `IN154` Validation on Failing Integrations

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-3678)

## Description
Add `IN154` error to the ignore list before releasing an update to the validation.


bakatzir:
I really prefer to not have this error is ignorable in Cortex XSOAR supported packs.
Please work with @Ni-Knight to decide on a reliability and have this validation as mandatory.

For community packs it can be ignorable.

Thanks.

MichaelYochpaz:
@bakatzir Issue is, that these integrations don't return an indicator (which means no DBot score),
So at the current state, the reliability field (if added) will be useless.

Packs/VulnDB/Integrations/VulnDB/VulnDB.py
```diff
@@ -360,22 +360,40 @@ def vulndb_get_cve_command(args: dict, client: Client):
...
+
     data = {
         "ID": cve_id,
         "CVSS": cvss_metrics_details[0].get("score", "0") if cvss_metrics_details else "0",
-        "Published": result.get('vulndb_published_date', '').rstrip('Z'),
-        "Modified": result.get('vulndb_last_modified', '').rstrip('Z'),
-        "Description": result.get("description", '')
+        "Published": result.get('vulndb_published_date').rstrip('Z'),
+        "Modified": result.get('vulndb_last_modified').rstrip('Z'),
+        "Description": result.get("description")
```
bziser, line 375: why removed the '' ? 
If you will run it on None, you will get:
`AttributeError: 'NoneType' object has no attribute 'rstrip'`

MichaelYochpaz: Right, my bad.

```suggestion
        "Published": result.get('vulndb_published_date', '').rstrip('Z'),
        "Modified": result.get('vulndb_last_modified', '').rstrip('Z'),
```

bakatzir:
so if `Pipl` is the only one left, let's ignore it, and please open an issue for a re write, I'll pormote and we will get rid of the ignorable error after it.

PR #26690, opened 2023-05-22
Support label feature fixes

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Description
we cannot use the `changed-files` for the `handle-external-pr` workflow because the `pull_request_target` event operates on the base branch and not on the branch that triggered the workflow, because of that we got the following error:

```
  Error: Similar commit hashes detected: previous sha: d547ae0268b99a344c9f9d19032f98f80493bc67 is equivalent to the current sha: d547ae0268b99a344c9f9d19032f98f80493bc67.
  Error: Please verify that both commits are valid, and increase the fetch_depth to a number higher than 50.
  Error: Process completed with exit code 1.
```

this commit is a commit of merging to the master branch which is not what we want in this case, we want the commits of the branch that triggered the workflow.

that happens because the checkout step checks out to the master branch instead of the branch that triggered the workflow.

so now instead of using `changed-files` we can simply use the `Github` module to request all the changed files from the pull request which is easier and more concise and we do not depend on the github workflow's ref branches.

see also the two workflows for the test branch:
1. review-release-notes which checks out to the triggered branch: https://github.com/demisto/content/actions/runs/5039367295/jobs/9037462417?pr=26678

see that the github ref for it is: `GITHUB_REF: refs/pull/26678/merge` which is the current pull request.

2. handle external pr which checks out to the master branch causing the error: https://github.com/demisto/content/actions/runs/5039367206/jobs/9037510926?pr=26678

see that the github ref for it is: `GITHUB_REF: refs/heads/master` which causes the issue

references: https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

### DIFF
--- a/.github/workflows/handle-new-external-pr.yml
+++ b/.github/workflows/handle-new-external-pr.yml
@@ -1,5 +1,7 @@
 name: Handle New External PRs
-on: pull_request
+on:
+  pull_request_target:
+    types: [opened]
 
 permissions:
   contents: read
@@ -7,7 +9,7 @@ permissions:
 jobs:
   handle_new_external_pr:
     runs-on: ubuntu-latest
-    if: github.repository == 'demisto/content'
+    if: github.repository == 'demisto/content' && github.event.action == 'opened' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


GuyAfik:
see this workflow with the changes on this branch that executed successfully: https://github.com/demisto/content/actions/runs/5041947755/jobs/9042100904?pr=26690

ShacharKidor (approved):

PR #26927, opened 2023-05-23
Enable branch protection for all contributor branches

amshamah419:
## Status
- [ ] In Progress
- [x] Ready


.github/workflows/protect-infra-directories.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   check_changes:
     runs-on: ubuntu-latest
-    if: startsWith(github.head_ref, 'contrib')
+    if: contains(github.base_ref, 'contrib')
```
amshamah419, line 12:
```suggestion
    if: startsWith(github.head_ref, 'contrib') || ${{ github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'contrib') }}
```

This could also work as well. @dorschw what's your take?

dorschw: the latter seems more explicit and easier to maintain, if it performs the same let's go with it

dorschw (approved):

PR #26929, opened 2023-05-23
collect content items will now use id and folder

thefrieddan1:
to decide if to replace / display items in metadata.

## Status
- [] In Progress
- [x ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CRTX-82975
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6712

## Description
in order to decide if displaying the right version of content item used id. however it seems like same id can be found for different content item types. now using content item id + folder to decide if its the latest content item.


dorschw (commented):
Nice

yaakovpraisler (commented):
@thefrieddan1 - Are we sure that there are no content items with same ids in the same directory?
and if so, @dorschw - do we have a validation to prevent adding another content item of the same type with same id?

thefrieddan1:
@yaakovpraisler Anar confirmed now that the same content items of the same type should not have same id.

dorschw (approved):

yaakovpraisler (approved):

PR #26954, opened 2023-05-23
Ruff: Phase 2 (05-23)

dorschw:
| Code    |     Name                                                                                                | Explanation                                                                                                         |
| ------- | ------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| A003    | [builtin-attribute-shadowing](https://beta.ruff.rs/docs/rules/builtin-attribute-shadowing/)                   | Class attribute {name} is shadowing a python builtin                                                               |
| B006    | mutable-argument-default                                                                                      | Do not use mutable data structures for argument defaults                                                           |
| B007    | unused-loop-control-variable                                                                                  | Loop control variable {name} not used within loop body                                                             |
| B008    | [function-call-in-default-argument](https://beta.ruff.rs/docs/rules/function-call-in-default-argument/)       | Do not perform function call {name} in argument defaults                                                           |
| B011    | assert-false                                                                                                  | Do not assert False (python -O removes these calls), raise AssertionError()                                        |
| B015    | useless-comparison                                                                                            | Pointless comparison. This comparison does nothing but waste CPU instructions. Either prepend assert or remove it. |
| B018    | useless-expression                                                                                            | Found useless expression. Either assign it to a variable or remove it.                                             |
| N813    | [camelcase-imported-as-lowercase](https://beta.ruff.rs/docs/rules/camelcase-imported-as-lowercase/)           | Camelcase {name} imported as lowercase {asname}                                                                    |
| PLW0602 | global-variable-not-assigned                                                                                  | Using global for {name} but no assignment is done                                                                  |
| PTH107  | os-remove                                                                                                     | os.remove() should be replaced by Path.unlink()                                                                    |
| PTH113  | os-path-isfile                                                                                                | os.path.isfile() should be replaced by Path.is_file()                                                              |
| RET507  | [superfluous-else-continue](https://beta.ruff.rs/docs/rules/superfluous-else-continue/)                       | Unnecessary {branch} after continue statement                                                                      |
| SIM102  | collapsible-if                                                                                                | Use a single if statement instead of nested if statements                                                          |
| SIM108  | if-else-block-instead-of-if-exp                                                                               | Use ternary operator {contents} instead of if-else-block                                                           |
| SIM110  | reimplemented-builtin                                                                                         | Use {repl} instead of for loop                                                                                     |
| SIM114  | [if-with-same-arms](https://beta.ruff.rs/docs/rules/if-with-same-arms/)                                       | Combine if branches using logical or operator                                                                      |
| SIM115  | open-file-with-context-handler                                                                                | Use context handler for opening files                                                                              |
| SIM116  | [if-else-block-instead-of-dict-lookup](https://beta.ruff.rs/docs/rules/if-else-block-instead-of-dict-lookup/) | Use a dictionary instead of consecutive if statements                                                              |
| SIM117  | [multiple-with-statements](https://beta.ruff.rs/docs/rules/multiple-with-statements/)                         | Use a single with statement with multiple contexts instead of nested with statements                               |
| SIM118  | in-dict-keys                                                                                                  | Use {key} in {dict} instead of {key} in {dict}.keys()                                                              |
| SIM201  | negate-equal-op                                                                                               | Use {left} != {right} instead of not {left} == {right}                                                             |
| SIM210  | if-expr-with-true-false                                                                                       | Use bool({expr}) instead of True if {expr} else False                                                              |
| SIM211  | if-expr-with-false-true                                                                                       | Use not {expr} instead of False if {expr} else True                                                                |
| SIM212  | if-expr-with-twisted-arms                                                                                     | Use {expr_else} if {expr_else} else {expr_body} instead of {expr_body} if not {expr_else} else {expr_else}         |
| SIM300  | yoda-conditions                                                                                               | Yoda conditions are discouraged, use {suggestion} instead                                                          |
| TRY201  | [verbose-raise](https://beta.ruff.rs/docs/rules/verbose-raise/)                                               | Use raise without specifying exception name                                                                        |


kobymeir (approved):

dorschw:
see https://github.com/demisto/content/pull/27732 instead

PR #26967, opened 2023-05-24
Set some scripts only XSOAR marketplace

israelpoli:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CRTX-75064

## Description
set up some scripts that will only upload to XSOAR as a result
that the command getMLModel is not supported in XSIAM

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


idovandijk (commented):
@israelpoli Can't we exclude the whole pack from XSIAM instead of individual items? Managing separate items will be confusing in the future

israelpoli:
@idovandijk 
Base and commonScripts are must be in XSIAM,
PhishingURL was set to only be in XSOAR

PR #26982, opened 2023-05-24
[Rapid7 InsightVM] Fix `token` Parameter BC Issue

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-24552)

## Description
Fix `token` parameter having unexpected values in some cases (string instead of dict, and empty string instead of None) after upgrading to 1.2.0+ and using a previous instance.


Packs/Rapid7_Nexpose/ReleaseNotes/1_2_9.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### Rapid7 InsightVM
+
+- Fixed an issue where usage of the **token** parameter may cause a Python error ("'str' object has no attribute 'get'"), or authentication issues, after upgrading the pack from versions that are < 1.2.0.
```
bziser, line 6:
```suggestion
- Fixed an issue where usage of the **token** parameter may cause aמ error or authentication issues.
```

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -5357,11 +5357,23 @@ def main():  # pragma: no cover
...
+        if params.get("token") and isinstance(params["token"], str):
+            token = params["token"]
+
+        else:
+            if identifier := params.get("token", {}).get("identifier"):
+                token = identifier
+
+            else:  # If token.identifier is an empty string, use None instead.
+                token = None
+
```
bziser, line 5372:
```suggestion
        token = None
        if params.get("token"):
               token = params["token"] if isinstance(params["token"], str) else params.get("token", {}).get("identifier")
```

bziser (commented):
Great

Packs/Rapid7_Nexpose/Integrations/Rapid7_Nexpose/Rapid7_Nexpose.py
```diff
@@ -5357,11 +5358,22 @@ def main():  # pragma: no cover
...
+        # A workaround for fixing compatibility issues when upgrading existing instances that are < 1.2.0.
+        # ('token' field was converted from type 0 to type 9)
+        token = None
+
+        if params.get("token"):
+            if isinstance(params["token"], str):
+                token = params["token"]
+
+            elif params["token"].get("identifier"):
+                token = params["token"]["identifier"]
```
bziser, line 5370:
```suggestion
            else:
                token = params["token"]["identifier"]
```
why needed the elif?

MichaelYochpaz:
```python
        token = None

        if params.get("token"):
            if isinstance(params["token"], str):
                token = params["token"]

            elif params["token"].get("identifier"):
                token = params["token"]["identifier"]
```

If there is no `token` (can happen in case it's of type `str`) or `token["identifier"]` does not exist, or is an empty string, `token` will remain as `None` (this is mainly to avoid a case where `identifier` will be able to have an empty string value, which can cause further bugs in the future)

bziser (approved):

PR #27017, opened 2023-05-28
O365 feed raise on error

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24426

## Description
We now raise an error for 503 status codes. This approach preserves most of the original functionality changed in https://github.com/demisto/etc/issues/47337

### DIFF
--- a/Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
+++ b/Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
@@ -119,6 +119,9 @@ class Client:
                                 f'Check your Server URL parameter.\n\n{err}')
             except requests.exceptions.HTTPError as err:
                 demisto.debug(f'Got an error from {feed_url} while fetching indicators {(str(err))} ')
+                if err.response.status_code == 503:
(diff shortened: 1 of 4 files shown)


Packs/FeedOffice365/Integrations/FeedOffice365/FeedOffice365.py
```diff
@@ -119,6 +123,13 @@ def build_iterator(self) -> List:
                                 f'Check your Server URL parameter.\n\n{err}')
             except requests.exceptions.HTTPError as err:
                 demisto.debug(f'Got an error from {feed_url} while fetching indicators {(str(err))} ')
+                if err.response.status_code == 503:
+                    raise Exception(f'The service located at {feed_url} is unavailable while fetching '
+                                    f'indicators {(str(err))} ')
+                elif err.response.status_code == 400 and region == GERMANY:
```
yuvalbenshalom, line 129: after this if you should have a general raise for al the other cases, so we wold not silence the other http errors. we should fail on the also.

yuvalbenshalom (changes requested):
see my comment

yuvalbenshalom (approved):

PR #27026, opened 2023-05-28
Test Playbook for 'Cortex XDR Malware - Investigation And Response'

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6789

## Description
Test playbook for the ‘Cortex XDR Malware - Investigation And Response’ playbook.

## Screenshots
Test_Playbook_-_Cortex_XDR_Malware_-_Investigation_And_Response


Tests/conf.json
```diff
@@ -20,6 +20,20 @@
...
+        {
+            "integrations": [
+                "WildFire-v2",
+                "Cortex XDR - IR",
+                "Joe Security"
+            ],
+            "instance_names": [
+                "WildFire-v2",
+                "Cortex XDR - IR",
+                "Joe_Security"
```
ostolero, line 32: We need to set instance names only if we have multiple credentials for the same integration, this doesn't seem the case so we can remove this.

TalNos: Fixed, thanks.

Tests/conf.json
```diff
@@ -20,6 +20,20 @@
...
+            ],
+            "playbookID": "Test_Playbook_-_Cortex_XDR_Malware_-_Investigation_And_Response",
```
idovandijk, line 29: "playbookID" should be the ID which should be the same as the name: `Test Playbook - Cortex XDR Malware - Investigation And Response`

Tests/conf.json
```diff
@@ -20,6 +20,20 @@
...
+            "fromversion": "6.5.0"
```
idovandijk, line 30: Your test PB has fromversion 6.8.0. This should be the same

idovandijk (commented):
Great job, please see my comments.
In addition, for use-case playbooks I prefer testing using our newer method with the "Create Test Incidents" integration. If you want we can do a session explaining how it works. I did it for Phishing for example.
The idea with the newer method is that we use actual incident JSON to create the incident, and then we also test the mapper and classifier in the process. We can simulate a real incident with the test.

But this is still great, and definitely better than no test :)

idovandijk (approved):
Great job Tal!

TalNos:
duplicated. changes will be made by the following PR: 
Test Playbook - 'Cortex XDR Malware - Investigation And Response' #27402
https://github.com/demisto/content/pull/27402

PR #27190, opened 2023-06-04
Remove default values from yml - phase 1

ilappe:
## Related Issues
https://jira-hq.paloaltonetworks.local/browse/CIAC-4906

## Description
currently the default values like `hidden: false` are removed in server side 
this cause a lot of noises in contribution process when the yml have a lot of diff

we do it without bump the pack versions & release notes.
phase 1 - only yml integrations & scripts with the following fields (name:count in files):

`{'isremotesyncin':1, 'ismappable':4, 'hidden':286, 'isremotesyncout':1, 'enabled':166, 'defaultEnabled':9, 'beta':14, 'hiddenusername':16, 'vcShouldKeepItemLegacyProdMachine':1, 'defaultValue':18}`
~ 300 packs


ilappe:
running `git diff origin/master | grep '\-  ' | grep -v ": *false" > test_out` to get all the deleted text that not contained `false`
it's look like there is only format diff

PR #27352, opened 2023-06-11
CSP - Added rank class (new grid field) and rank to the Domain class (new grid field)

Ni-Knight:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
This is due to a request by domain tools to add a `Rank` grid field to a domain type indicator to allow for data from ranking sources such as majestic.
 
- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -119,10 +119,10 @@ def fix_traceback_line_numbers(trace_str):
     """
     def is_adjusted_block(start, end, adjusted_lines):
         return any(
-            block_start < start < end < block_end 
+            block_start < start < end < block_end
             for block_start, block_end in adjusted_lines.items()
         )
-        
+
     for number in re.findall(r'line (\d+)', trace_str):
         line_num = int(number)
         module = _find_relevant_module(line_num)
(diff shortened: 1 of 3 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -4614,6 +4636,9 @@ def to_context(self):
             if self.certificates:
                 domain_context['Certificates'] = self.create_context_table(self.certificates)
 
+            if self.rank:
+                domain_context['rank'] = self.create_context_table(self.rank)
```
eyalpalo, line 4640: So in that case well have context path ```rank.rank```?

Ni-Knight: no:
rank = [{"source": "x", "rank": "y"}, ...]

But it should be capitalized

Ni-Knight: That's how grids work in CSP (you can see ExternalReference, Hash, DNSRecord and so on).

eyalpalo (approved):

PR #27370, opened 2023-06-11
Add API Execution Metrics to Whois

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6900

## Description
This PR adds the API execution metrics to the `ip`, `domain` Reputation Commands and to the`whois` integration command. 

It also includes a bit of a code cleanup and Python 3 feature usages such as:
- Type hints.
- Function docstrings.
- Removal of unused arguments.
- Moving global integration parameters and command arguments to local scope.
- Removal of unneeded helper functions, handle socket opening and reading in the same function.
- Remove nested `try..catch` and handle all exceptions on the command function level. This was done to enable API execution metrics definition on the command function depth.
- Removed unneeded `domain_command` function as it has the same logic as `whois_command`.

## Screenshots
image

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0
- [x] 6.8.0

   - [x] Yes
       - Further details: The `CommonServerPython::append_metrics` method is used to return results from command functions. This function ensures that we're backwards compatible with versions < 6.8.0 since it checks `is_demisto_version_ge('6.8.0')`.
   - [ ] No

## Must have
- [x] Tests
- [x] Documentation


Packs/Whois/Integrations/Whois/test_data/ip_using_ciac-6900.json
kgal-pan, line 1: These new files are the printed context used to compare the output of the same command using this branch and `master`. For example, this file is the result of running the command:
```
!ip ip=8.8.8.8,1.1.1.1 using="ciac-6900"
```

We can see that this JSON is identical to the one in `ip_using_master.json`.

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8494,235 @@ def prepare_readable_ip_data(response):
 '''COMMANDS'''
 
 
-def domain_command(reliability):
```
kgal-pan, line 8441: This function was removed because it's the same logic, input as `whois_command`. No need to maintain two identical functions.

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -7130,9 +7130,113 @@
...
+    # Connection Errors
+    ipwhois.exceptions.NetError: "connection_error",
+
+    # Rate Limit Errors
+    ipwhois.exceptions.HTTPRateLimitError: "quota_error",
+    ipwhois.exceptions.WhoisRateLimitError: "quota_error",
+}
+
+
+def has_rate_limited_result(cmd_results: List[CommandResults]) -> bool:
```
kgal-pan, line 7162: Not sure if this function is needed, depends on https://github.com/demisto/content/pull/27370/files#r1245304220

Packs/Whois/Integrations/Whois/test_data/whois_using_ciac-6900.json
kgal-pan, line 1: We can remove this after demo

Packs/Whois/Integrations/Whois/test_data/whois_using_master.json
kgal-pan, line 1: We can remove this after demo

Packs/Whois/Integrations/Whois/test_data/ip_using_master.json
kgal-pan, line 1: We can remove this after demo

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -7234,85 +7299,54 @@ def get_whois_raw(domain, server="", previous=None, rfc3490=True, never_cut=Fals
...
+            entry = tlds[tld]
             host = entry["host"]
-        except KeyError:
-            context = ({
-                outputPaths['domain']: {
-                    'Name': domain,
-                    'Whois': {
-                        'QueryStatus': 'Failed'
-                    }
-                },
```
ShahafBenYakir, line 7253: Where is this behavior?

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -7234,85 +7299,54 @@ def get_whois_raw(domain, server="", previous=None, rfc3490=True, never_cut=Fals
...
-        return host
+            return host
+        else:
+            raise WhoisInvalidDomain(f"Can't parse the root server from domain '{domain}'")
 
-    else:
-        raise WhoisException("No root WHOIS server found for domain.")
+    except (KeyError, TypeError, ValueError) as e:
+        demisto.error(f"Could not get root server from domain '{domain}': {e.__class__.__name__} {e}")
+        raise WhoisInvalidDomain(f"Can't parse the root server from domain '{domain}'")
```
ShahafBenYakir, line 7319: Seems that we are not using the SHOULD_ERROR flag?

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8617,28 +8742,52 @@ def main():
...
+                ))
+
             elif command == 'domain':
-                domain_command(reliability)
+                results.extend(whois_command(
+                    reliability=reliability,
+                    query=args.get('domain'),  # type: ignore
+                    is_recursive=argToBoolean(args.get("recursive", 'false')),
+                    should_error=should_error,
+                ))
```
ShahafBenYakir, line 8784:
```suggestion
                domain_to_query = args.get('domain') 
                is_recurisve = argToBoolean(args.get("recursive", 'false'))
                cmd_res = whois_command(
                    reliability=reliability,
                    query=domain_to_query),  # type: ignore
                    is_recursive=,is_recurisve
                    should_error=should_error,
                    )
                results = cmd_res
```

Packs/Whois/Integrations/Whois/Whois.py
ShahafBenYakir: But it is not entirely correct as the domain logic should have the "generic command" [outputs](https://xsoar.pan.dev/docs/integrations/context-standards-mandatory#domain) whereas the whois command does not.

Packs/Whois/ReleaseNotes/1_5_0.json
```diff
@@ -0,0 +1,4 @@
+{
+	"breakingChanges": true,
+	"breakingChangesNotes": "When supplying a comma-separated list of domains or IPs to the `domain`,`ip` and `whois` commands, and one of the domains/IPs is invalid, the command will return results for the valid domains/IPs in addition to the error entry for the invalid domains/IPs"
```
ShahafBenYakir, line 3: Can we make this shorter? I want users to be able to easily understand how it affects them, maybe something like that?

```suggestion
	"breakingChangesNotes": "As opposed to before where the Whois integration failed to finish analyzing a list of inputs if given one invalid domain/IP, the integration now returns a combination of both success results and errors "
```

ShahafBenYakir (commented):
Great effort here!
TBH it is preatty tough to follow all the changes I went over as deeply as I could.

Packs/Whois/Integrations/Whois/Whois.py
ShahafBenYakir: Also It is usually recommended  to create such wrappers so you could easily extract different arguments for different functions.

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8297,28 +8322,34 @@ def parse_nic_contact(data):
     return handle_contacts
 
 
-def get_whois(domain, normalized=None, is_recursive=True):
-    if normalized is None:
-        normalized = []
+def get_whois(domain: str, is_recursive=True):
```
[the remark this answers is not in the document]
kgal-pan: Exactly. I noticed that the two instances where this method was used the `normalized` argument was not set.

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: I see what you mean. In the case with the code in `domain_command` and `whois_command` currently available in `master`, what is returned is identical:

```python
# domain_command

demisto.results({
            'Type': entryTypes['note'],
            'ContentsFormat': formats['markdown'],
            'Contents': str(whois_result),
            'HumanReadable': tableToMarkdown('Whois results for {}'.format(domain), md),
            'EntryContext': dbot_score,
        })

# whois_command

demisto.results({
            'Type': entryTypes['note'],
            'ContentsFormat': formats['markdown'],
            'Contents': str(whois_result),
            'HumanReadable': tableToMarkdown('Whois results for {}'.format(domain), md),
            'EntryContext': context_res,
        })
```

Therefore, since there doesn't seem to be a wrapper in this case, I decided to remove one of them for easier maintainability.

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8297,28 +8322,34 @@ def parse_nic_contact(data):
...
+        # checks for largest matching suffix inside tlds dictionary
+        suffix_len = max([len(suffix) for suffix in tlds if query.endswith('.{}'.format(suffix))] or [0])
+        # if suffix(TLD) was found increase the length by one in order to add the dot before it. --> .com instead of com
+        if suffix_len != 0:
+            suffix_len += 1
+        suffixless_query = query[:-suffix_len]
+        domain = query
+        # checks if query includes subdomain
+        if suffixless_query.count(".") > 0:
+            domain = query[suffixless_query.rindex(".") + 1:]
```
tkatzir, line 8348: Does this work for both `amazon.com` and `amazon.co.uk`?

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8297,28 +8322,34 @@ def parse_nic_contact(data):
...
+        return domain
+    except Exception:
+        demisto.error(f"Error parsing domain from query '{query}'...")
```
tkatzir, line 8351:
```suggestion
        demisto.error(f"Error parsing domain from query '{query}'.")
```

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
+    """
+    Performs an Registration Data Access Protocol (RDAP) lookup for an IP.
+
+    See https://ipwhois.readthedocs.io/en/latest/RDAP.html
+
+    Arguments:
+        - `ip` (``str``): The IP to perform the lookup for.
+        - `retry_count` (``int``): The number of times to retry the lookup in case of rate limiting error.
+        - `rate_limit_timeout` (``int``): How long in seconds to wait before retrying the lookup in case of rate limiting error.
+
```
tkatzir, line 8485: Missing some parameters.

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
-    rate_limit_errors_suppressed: bool = bool(get_param_or_arg(
-        'rate_limit_errors_suppressed', 'rate_limit_errors_suppressed') or RATE_LIMIT_ERRORS_SUPPRESSEDL_DEFAULT)
+def ip_command(ips: str, reliability: str, rate_limit_retry_count: int, rate_limit_wait_seconds: int, should_error: bool = False) -> List[CommandResults]:
+    """
+    Performs RDAP lookup for the IP(s) and returns a list of CommandResults.
+    Sets API execution metrics functionality (if supported) and adds them to the list of CommandResults.
+
+    Args:
+        - `ips` (``str``): Comma-separated list of IPs to perform the RDAP lookup for.
+        - `reliability` (``DBotScoreReliability``): RDAP lookup source reliability.
```
tkatzir, line 8515: The type of `reliability`, is it `str` or `DBotScoreReliability`?

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
+                    result = CommandResults(readable_output=f"No results returned for IP {ip}", entry_type=EntryType.WARNING)
+
+            results.append(result)
+
+        except Exception as e:
+            demisto.error(f"Exception type {e.__class__.__name__} caught performing RDAP lookup for IP {ip}: {e}")
+
+            output = {
+                'query': ip,
+                'raw': f"Query failed for {ip}: {e.__class__.__name__} {e}"
```
tkatzir, line 8575:
```suggestion
                'raw': f"Query failed for {ip}: {e.__class__.__name__}, {e}"
```

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
+                entry_type=EntryType.NOTE,
+                content_format=EntryFormat.MARKDOWN,
+                readable_output=tableToMarkdown('Whois results for {}'.format(domain), md),
+                raw_response=str(whois_result)
+            )
+
+            results.append(result)
+
+        except Exception as e:
+            demisto.error(f"{e.__class__.__name__} caught performing whois lookup with domain '{domain}'")
```
tkatzir, line 8653:
```suggestion
            demisto.error(f"Exception of type {e.__class__.__name__} was caught while performing whois lookup with the domain '{domain}'")
```

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
+                    'Whois': {
+                        'QueryStatus': f"Failed whois lookup: {e}"
+                    }
+                },
+            })
+
+            if should_error:
+                results.append(CommandResults(
+                    outputs=output,
+                    readable_output=f"{e.__class__.__name__} caught performing whois lookup with domain '{domain}': {e}",
```
tkatzir, line 8673:
```suggestion
                    readable_output=f"Exception of type {e.__class__.__name__} was caught while performing whois lookup with the domain '{domain}': {e}",
```

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8438,140 +8469,232 @@ def prepare_readable_ip_data(response):
...
+                    entry_type=EntryType.ERROR,
+                    raw_response=str(e)
+                ))
+            else:
+                results.append(CommandResults(
+                    outputs=output,
+                    readable_output=f"{e.__class__.__name__} caught performing whois lookup with domain '{domain}': {e}",
```
tkatzir, line 8680:
```suggestion
                    readable_output=f"Exception of type {e.__class__.__name__} was caught while performing whois lookup with the domain '{domain}': {e}",
```

tkatzir (commented):
It's good. Some minor comments.

And one general note: There are many logs like: "Going to do this task..."
but there's no "Finished the task"

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8617,28 +8770,65 @@ def main():
...
-            ret_value = ip_command(ip, reliability)
-            if ret_value:
-                return_results(ret_value)
+            ip = args.get('ip', '')
+            rate_limit_retry_count: int = arg_to_number(get_param_or_arg(
+                param_key='rate_limit_retry_count', arg_key='rate_limit_retry_count'), arg_name="rate_limit_retry_count") or RATE_LIMIT_RETRY_COUNT_DEFAULT
+            rate_limit_wait_seconds: int = arg_to_number(get_param_or_arg(
+                param_key='rate_limit_wait_seconds', arg_key="rate_limit_wait_seconds")) or RATE_LIMIT_WAIT_SECONDS_DEFAULT
+            rate_limit_errors_suppressed: bool = argToBoolean(get_param_or_arg(
+                param_key='rate_limit_errors_suppressed', arg_key='rate_limit_errors_suppressed')) or RATE_LIMIT_ERRORS_SUPPRESSEDL_DEFAULT
```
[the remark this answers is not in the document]
kgal-pan: I thought it would be easier to feed the arguments specific to the IP command to the function and makes it easier to test (rather than `mock.patch` each one of them inside the UT). I wrote a UT for it as well, `test_ip_command` .

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: This is now added in the `whois_command` when an exception is caught, lines 8661-8668.

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: Right, `SHOULD_ERROR`'s only function is deciding whether the entry returned is an error vs a warning:

```python
if should_error:
     result = CommandResults(readable_output=f"No results returned for IP {ip}", entry_type=EntryType.ERROR)
else:
     result = CommandResults(readable_output=f"No results returned for IP {ip}", entry_type=EntryType.WARNING)
```

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: Which ones?

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: Yep, there's a UT specifically for that:

```python
@pytest.mark.parametrize(
    'query,expected',
    [("app.paloaltonetwork.com", "paloaltonetwork.com"),
     ("test.this.google.co.il", "google.co.il"),
     ("app.XSOAR.test", "app.XSOAR.test")]
)
def test_get_domain_from_query(query, expected):
    assert get_domain_from_query(query) == expected
```

Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -8606,8 +8732,10 @@ def setup_proxy():
 
 
 def main():
-    LOG('command is {}'.format(str(demisto.command())))
+    demisto.debug(f"command is {demisto.command()}")
     command = demisto.command()
+    args: Dict[str, str] = demisto.args()
+    should_error = argToBoolean(demisto.params().get('with_error', False))
```
kgal-pan, line 8738: Honestly, not sure we need `should_error` at all anymore since we return an error entry (instead of using `return_error`) in all command flows. The only usage it has at this point is to return a warning entry instead of an error entry which I'm not sure is needed as well.

Packs/Whois/Integrations/Whois/Whois.yml
```diff
@@ -57,14 +58,6 @@ configuration:
   section: Collect
   advanced: true
   required: false
-- display: Suppress Rate Limit errors
```
kgal-pan, line 60: We're no longer parsing the suppress integration instance in our Python code.

Is this OK to remove the integration setting? It's not a required field.

Packs/Whois/Integrations/Whois/Whois.py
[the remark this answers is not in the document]
ShahafBenYakir: I understand but if the only thing it causes is the usage of mocks then let's revert it. The unit tests are not the core priority, the code clearness, and standards are more important.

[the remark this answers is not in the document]
kgal-pan: https://github.com/demisto/content/pull/27370/commits/40a6cbfe2e250f8b1e2ec36dbf65806792ad41e5

Packs/Whois/Integrations/Whois/Whois.py
ShahafBenYakir: Let's do this, since merging both logics into one amends you need to take thee extraction of args outside from the whois_command (different from what happens today), let's add a wrapper command for each of the flows (whois_command and domain_command where each is making the needed preparations relevant foe its flow (extraction of args, initialization of variables, etc...)

Packs/Whois/Integrations/Whois/Whois.py
ShahafBenYakir: Sounds like removing it will break BC and suddenly users who used to not fail on this will start to fail no?

Packs/Whois/Integrations/Whois/Whois.yml
ShahafBenYakir: @tkatzir can you have a look?

Packs/Whois/Integrations/Whois/Whois.yml
```diff
@@ -463,9 +456,6 @@ script:
     - description: 'The number of seconds to wait each iteration when getting a Rate Limit response.'
       defaultValue: 120
       name: rate_limit_wait_seconds
-    - description: 'Whether Rate Limit errors should be supressed or not.'
```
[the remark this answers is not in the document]
ShahafBenYakir: @tkatzir also here

Packs/Whois/Integrations/Whois/Whois.yml
tkatzir: AFAIK, the "standard" to remove a config is to make it hidden.

tkatzir: After discussing this with @ShahafBenYakir, it's ok.
My main concern was about breaking custom code.

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: https://github.com/demisto/content/pull/27370/commits/5fd5e32059294aed675c92a01b757f3257911d77

Packs/Whois/Integrations/Whois/Whois.yml
kgal-pan: > removing this field yields another break in BC for customers no? Customers that had this flag on will now get errors instead warnings no?

The usage of the flag in version 1.4.16 is in this section of the `get_whois_ip`:

```python
try:
        return ip_obj.lookup_rdap(depth=1, retry_count=retry_count, rate_limit_timeout=rate_limit_timeout)
except urllib.error.HTTPError as e:
        if rate_limit_errors_suppressed:
            demisto.debug(f'Suppressed HTTPError when trying to lookup rdap info. Error: {e}')
            return
        demisto.error(f'HTTPError when trying to lookup rdap info. Error: {e}')
        raise e
```

So basically if the flag is on and we get some type of generic HTTP error, we return `None` from the `get_whois_ip` command. However, we never actually check the value of `response` in `ip_command` method:

```python
        response = get_whois_ip(args)

        result = CommandResults(
            outputs_prefix='Whois.IP',
            outputs_key_field='query',
            outputs=response,
            readable_output=tableToMarkdown('Whois results:', readable_data),
            raw_response=response,
            indicator=ip_output
        )

        results.append(result)
    return results
```

And in `main` we check if the `ret_value` exists (which it always does):
```python
        if command == 'ip':
            demisto_args = demisto.args()
            ip = demisto_args.get('ip')
            ret_value = ip_command(ip, reliability)
            if ret_value:
                return_results(ret_value)
            else:
                return_error(f'Failed to lookup ip {ip}')
```

So an error is never returned in that case, just an empty command result.

 If the suppress flag is off and an exception occurs, we raise the exception and it gets propagated and handled in the `main` `try/catch` block where a `return_error` is specified.

With the changes in this PR (including removal of the suppress flag) we now return an error entry when the `should_error` parameter is set and in cases when:

* The response is empty from `get_whois_ip` method.
* An exception (including `HttpError`) is raised 

If the `should_error` parameter is unset, we return a warning entry in those cases.

kgal-pan: Had a short call with @ShahafBenYakir about this. We decided to bring back the suppress flag so now if an `urrlib.error.HttpError` is encountered, we will return `None` from the `get_whois_ip` method and use the `with_error` integration instance to decide whether to return an error or warning entry. 

Changes done in https://github.com/demisto/content/pull/27370/commits/1176153aa910e031e45afd0410f348352d7a563e.

kgal-pan:
https://code.pan.run/xsoar/content/-/jobs/30607595 failing because of a bug in the SDK reported in https://jira-hq.paloaltonetworks.local/browse/CIAC-7723.

Tests/conf.json
```diff
@@ -2372,7 +2372,8 @@
         {
             "integrations": "Whois",
             "playbookID": "whois_test",
-            "fromversion": "4.1.0"
+            "fromversion": "5.0.0",
+            "timeout": 5000
```
kgal-pan, line 2376: @ostolero - I'm increasing the timeout because of constant timeouts across all 6.8+ instances, e.g.

https://code.pan.run/xsoar/content/-/jobs/30751591#L1829
https://code.pan.run/xsoar/content/-/jobs/30751592#L1788

kgal-pan: I changed it to 2000 and the test playbook finished successfully.

.guardrails/ignore
```diff
@@ -1,2 +1,3 @@
 Integrations/Active_Directory_Query/key.pem
-poetry.lock
\ No newline at end of file
+poetry.lock
+Packs/Whois/Integrations/Whois/Whois_test.py
```
ShahafBenYakir, line 3: What is this?

ShahafBenYakir (commented):
Looks ok to me in general

ShahafBenYakir (approved):

.guardrails/ignore
kgal-pan: The Guardrails job was because of the `pickle`ing of test data in https://github.com/demisto/content/blob/CIAC-5844-whois-add-metrics/Packs/Whois/Integrations/Whois/Whois_test.py#L332-L333, see https://docs.guardrails.io/docs/vulnerabilities/python/insecure_processing_of_data. 

Since we know what we're un`pickle`ing, not sure this is actually a threat. WDYT?

ShahafBenYakir: Instead of this file use `// guardrails-disable-line` on the line

kgal-pan: https://github.com/demisto/content/pull/27370/commits/db03805d9abaed746d9a387fbabe01d3c1c6c85e

PR #27405, opened 2023-06-13
[Build] Fix Dependency Issues

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-4543)

## Description
Currently, in the pre_update check that installs packs from master and assures they're installed properly, the required dependencies are taken from master (API call to marketplace), while the check whether each pack is deprecated, is done locally from the branch.
This causes issues when deprecating a pack and changing dependencies on the same PR.

This PR fixes the behavior to get deprecation data from the same API response used for fetching the dependencies instead of checking locally.

Test build before the fix: https://code.pan.run/xsoar/content/-/pipelines/5513561
Test build after the fix: https://code.pan.run/xsoar/content/-/pipelines/5514657
(only XSOAR instance builds are relevant to the changes)


bziser (commented):
looks good, please add a test which simulates this situation as we discussed

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -32,20 +32,6 @@
 SUCCESS_FLAG = True
 
 
-def is_pack_deprecated(pack_path: str) -> bool:
```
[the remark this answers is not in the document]
MichaelYochpaz: No, except for `Tests/Marketplace/Tests/search_and_install_packs_test.py` which apparently is what caused the run-unit-tests step to fail, as it has:
```python
mocker.patch.object(script, 'is_pack_deprecated', return_value=False)
```

I'll update that.

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -112,7 +98,8 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
             if dependant in dependants_ids and is_required and dependency.get('id') not in checked_packs:
                 dependencies_data.append({
                     'id': dependency.get('id'),
-                    'version': dependency.get('extras', {}).get('pack', {}).get('currentVersion')
+                    'version': dependency.get('extras', {}).get('pack', {}).get('currentVersion'),
+                    'deprecated': dependency.get('extras', {}).get('pack', {}).get('deprecated'),
```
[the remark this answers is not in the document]
MichaelYochpaz: Doesn't work as this is a build script which doesn't import `CommonServerPython`.

Tests/Marketplace/Tests/search_and_install_packs_test.py
```diff
@@ -1,23 +1,26 @@
...
-    "id": "HelloWorld",
-    "currentVersion": "1.1.10"
-}
-MOCK_AZURESENTINEL_SEARCH_RESULTS = {
-    "id": "AzureSentinel",
-    "currentVersion": "1.0.2"
-}
+
+def load_json_file(file_path: str):
+    with open("test_data/" + file_path, 'r') as json_file:
```
bziser, line 16:
```suggestion
    with open(f"test_data/{file_path}", 'r') as json_file:
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -434,46 +369,57 @@ def search_pack_and_its_dependencies(client: demisto_client,
...
+                current_packs_to_install.append(dependency)
+
+    lock.acquire()
+
+    if one_pack_and_its_dependencies_in_batch:
+        if batch_packs_install_request_body is None:
+            batch_packs_install_request_body = []
+        pack_and_its_dependencies = \
+            {p['id']: p for p in current_packs_to_install if p['id'] not in packs_in_the_list_to_install}
+        if pack_and_its_dependencies:
```
bziser, line 413:
```suggestion
        if pack_and_its_dependencies := {
            p['id']: p
            for p in current_packs_to_install
            if p['id'] not in packs_in_the_list_to_install
        }:
```

bziser (commented):
great!
mine initial cr :) 
@ilaner take a look too please

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -434,46 +369,57 @@ def search_pack_and_its_dependencies(client: demisto_client,
...
-                if pack['id'] not in packs_to_install:
-                    packs_to_install.append(pack['id'])
-                    installation_request_body.append(pack)
-        lock.release()
+    api_data = get_pack_dependencies(client, pack_id, lock)
+
+    if not api_data:
+        return
+
+    pack_data = api_data["packs"][0]
```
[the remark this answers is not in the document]
MichaelYochpaz: Because as far as I know, a successful response for this endpoint should always have the same structure, and I think it would be better to fail on an error so that we'll know something is wrong, than treating it gracefully as an empty list and continuing.

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,52 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
         create_dependencies_data_structure(response_data, next_call_dependants_ids, dependencies_data, checked_packs)
 
 
-def get_pack_dependencies(client: demisto_client, pack_data: dict, lock: Lock):
-    """ Get the pack's required dependencies.
+def get_pack_dependencies(client: demisto_client, pack_id: str, lock: Lock):
```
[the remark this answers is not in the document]
MichaelYochpaz:
```suggestion
def get_pack_dependencies(client: demisto_client, pack_id: str, lock: Lock) -> dict | None:
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -94,13 +78,12 @@ def is_pack_hidden(pack_id: str) -> bool:
     return False
 
 
-def create_dependencies_data_structure(response_data: dict, dependants_ids: list, dependencies_data: list,
-                                       checked_packs: list):
-    """ Recursively creates the packs' dependencies data structure for the installation requests
-    (only required and uninstalled).
+def create_dependencies_data_structure(response_data: dict, dependants_ids: list, dependencies_data: list, checked_packs: list):
```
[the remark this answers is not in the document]
MichaelYochpaz: It doesn't return anything, it's a recursive function that gets sent a pointer (`dependencies_data`) to an empty list created before calling it, and fills it recursively with all the dependencies (it's being called recursively for each dependency to find its dependencies).

I thought of rewriting it in a more-simple and less-complicated non-recursive function, but that has a potential of breaking stuff and will require some thorough testing.

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,51 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
...
-            create_dependencies_data_structure(response_data, dependants_ids, dependencies_data, dependants_ids)
-            if dependencies_data:
-                dependencies_str = ', '.join([dep['id'] for dep in dependencies_data])
-                logging.debug(f'Found the following dependencies for pack {pack_id}: {dependencies_str}')
-            return dependencies_data
-        if status_code == 400:
-            logging.error(f'Unable to find dependencies for {pack_id}.')
-            return []
-        msg = response_data.get('message', '')
-        raise Exception(f'status code {status_code}\n{msg}\n')
```
MichaelYochpaz, line 163: Removed this section since `demisto_client.demisto_api.rest.py` raises an APIException if an error code is received on `RESTClientObject.request()`:
```python
        if not 200 <= r.status <= 299:
            raise ApiException(http_resp=r)
```

So in case an error code is returned, it skips straight to `except ApiException as api_ex:` making this code section redundant.

bziser:
Adding @yucohen and @daryakoval to review too

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,51 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
...
     Returns:
-        (list) The pack's dependencies.
+        dict: API response data for the /search/dependencies endpoint.
     """
     global SUCCESS_FLAG
-    pack_id = pack_data['id']
-    logging.debug(f'Getting dependencies for pack {pack_id}')
+
+    api_endpoint = "/contentpacks/marketplace/search/dependencies"
+    body = [{"id": pack_id}]  # 'version' key can be specified alongside "id". Not including it will fetch the latest version.
```
bziser, line 126:
```suggestion
    body = [{"id": pack_id}]  # Not including 'version' key will fetch the latest version.
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,51 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
...
+
+    logging.debug(f"Fetching dependencies for pack '{pack_id}'.\n"
+                  f"Sending POST request to {api_endpoint} with body: {json.dumps(body)}")
```
bziser, line 129:
```suggestion
    logging.debug(f"Fetching dependencies for pack '{pack_id}'.\n"
                  f"Sending POST request to {api_endpoint} with body: {body}")
```

Tests/Marketplace/search_and_install_packs.py
bziser: `if not 200 <= r.status <= 299:` so need to be checked `if 200 <= status_code < 300`

MichaelYochpaz: You're right, shouldn't have removed this section:
```python
        if 200 <= status_code < 300:
            dependencies_data: list = []
            dependants_ids = [pack_id]
            response_data = response_data.get('dependencies', [])
            create_dependencies_data_structure(response_data, dependants_ids, dependencies_data, dependants_ids)
            if dependencies_data:
                dependencies_str = ', '.join([dep['id'] for dep in dependencies_data])
                logging.debug(f'Found the following dependencies for pack {pack_id}: {dependencies_str}')
            return dependencies_data
```
I'll revert that

Tests/Marketplace/search_and_install_packs.py
MichaelYochpaz:
```suggestion
    body = [{"id": pack_id}]  # Not specifying a "version" key will result in the latest version of the pack being fetched.
```

Tests/Marketplace/search_and_install_packs.py
MichaelYochpaz: `json.dumps()` is more accurate for debugging purposes, as it's formatted in the actual format that is used in HTTP requests (ex: using `"` instead of `'`), and can be copy & pasted straight to CURL, Postman, etc.

Differences:
```python
test = {"a": "test", "b": [1,2,3], "c": {1: 2}}

str(test)
>> "{'a': 'test', 'b': [1, 2, 3], 'c': {1: 2}}"

json.dumps(test)
>> '{"a": "test", "b": [1, 2, 3], "c": {"1": 2}}'
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -434,45 +368,54 @@ def search_pack_and_its_dependencies(client: demisto_client,
             Each list contain one pack and its dependencies.
 
     """
-    pack_data = {}
-    if pack_id not in packs_to_install:
-        pack_display_name = get_pack_display_name(pack_id)
-        if pack_display_name:
-            pack_data = search_pack(client, pack_display_name, pack_id, lock)
```
daryakoval, line 441: Why you deleted this function?

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,51 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
...
+                          f"Response code '{api_ex.status}'\nResponse: '{api_ex.body}'\nResponse Headers: '{api_ex.headers}'")
+
     except Exception as ex:
         with lock:
             SUCCESS_FLAG = False
-        logging.exception(f"The request to get pack {pack_id} dependencies has failed. {ex}.")
+        logging.exception(f"API call to fetch dependencies of '{pack_id}' has failed.\nError: {ex}.")
 
-
-def search_pack(client: demisto_client,
```
daryakoval, line 175: Make sure that we are not using this function in upload flow/private build/ other places in content

Tests/Marketplace/search_and_install_packs.py
daryakoval, line 126: Have you verified this in XSOAR, XSOAR-NG, XSIAM?

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -124,102 +108,51 @@ def create_dependencies_data_structure(response_data: dict, dependants_ids: list
         create_dependencies_data_structure(response_data, next_call_dependants_ids, dependencies_data, checked_packs)
 
 
-def get_pack_dependencies(client: demisto_client, pack_data: dict, lock: Lock):
-    """ Get the pack's required dependencies.
+def get_pack_dependencies(client: demisto_client, pack_id: str, lock: Lock) -> dict | None:
```
daryakoval, line 111: Make sure we are not using this function in another build flows.

daryakoval (changes requested):
The same code runs for XSOAR, XSIAM, and XSOAR-NG... So we should test it for all envs. 
Please trigger Nightly, Test Upload flow, and add here links, to make sure we didn't break anything.

Tests/Marketplace/search_and_install_packs.py
MichaelYochpaz: The display name returned by`get_pack_display_name` was used in the past in the API call, but the code has changed since then, and now it's only used in logs returned by `search_pack` (which was removed).

`search_pack` is an API call that is used only to fetch the latest version of the pack, which is then passed to the dependencies API call.

This is redundant, since if we just pass a pack's ID without a version to the dependencies API call (on `get_pack_dependencies`), we already get the latest version.

Both of these functions are not in use anywhere else in the code.

Tests/Marketplace/search_and_install_packs.py
MichaelYochpaz: Already tested, it's not.

Tests/Marketplace/configure_and_install_packs.py
```diff
@@ -39,16 +39,15 @@ def install_packs_from_content_packs_to_install_path(servers, pack_ids, marketpl
     Install pack_ids from "$ARTIFACTS_FOLDER/content_packs_to_install.txt" file, and packs dependencies.
 
     Args:
-        hostname:
         pack_ids: the pack IDs to install.
         servers: XSIAM or XSOAR Servers to install packs on it.
     """
-    install_packs_one_by_one: bool = marketplace_tag_name == XSIAM_MP
+    use_multithreading = marketplace_tag_name != XSIAM_MP
```
bziser, line 45: why you changed from `==` to `!=`?

MichaelYochpaz: I changed the`install_packs_one_by_one` flag to `multithreading` on https://github.com/demisto/content/pull/27405/commits/57b3e7d37ab17bf35f5b491fb0545218d039d0ab since it was confusing (at least to me), but now it's the opposite (`install_packs_one_by_one` is like  `not multithreading`), so I had to reverse it.

To make it easier to follow - this condition means that we install packs one by one (without multithreading) when the `XSIAM_MP` is used. So if the tag is different, we use multithreading.

Tests/Marketplace/search_and_install_packs.py
MichaelYochpaz: Already did, it's not in use anywhere else

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -35,47 +34,49 @@
...
-    return tools.get_pack_metadata(str(pack_metadata_path)).get('hidden', False)
+    if check_locally:
+        pack_metadata_path = Path(PACKS_FOLDER) / pack_id / PACK_METADATA_FILENAME
+
+        if not pack_metadata_path.is_file():
+            return True
+
+        return tools.get_pack_metadata(str(pack_metadata_path)).get('hidden', False)
+
+    if not check_locally:
```
MichaelYochpaz, line 59:
```suggestion
    else:
```

MichaelYochpaz:
@ilaner @daryakoval @yaakovpraisler @yucohen The build keeps failing only for general known issues (I confirmed with @yucohen), I think we can start the review process and just ignore them for now.

This is the nightly build I triggered:
https://code.pan.run/xsoar/content/-/pipelines/5681897

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -86,143 +86,88 @@ def is_pack_hidden(pack_id: str) -> bool:
...
             is_required = dependants[dependant].get('level', '') == 'required'
-            if dependant in dependants_ids and is_required and dependency.get('id') not in checked_packs:
-                dependencies_data.append({
-                    'id': dependency.get('id'),
-                    'version': dependency.get('extras', {}).get('pack', {}).get('currentVersion')
-                })
-                next_call_dependants_ids.append(dependency.get('id'))
-                checked_packs.append(dependency.get('id'))
+            if dependant in dependants_ids and is_required and dependency['id'] not in checked_packs:
+                dependencies_data.append(dependency)
```
daryakoval, line 114: How does the "dependency" object look now? With all data? not only ID and version?

Is the 'version': dependency.get('extras', {}).get('pack', {}).get('currentVersion') line handled?

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -417,77 +362,118 @@ def search_pack_and_its_dependencies(client: demisto_client,
...
+                     f"Pack deprecation status will be checked using local pack metadata.")
+
+        if is_pack_deprecated(pack_id=pack_id, check_locally=True):
+            logging.warning(f"Pack '{pack_id}' is deprecated (hidden) and will not be installed.")
+            return
+
+    else:
+        # On pre-update, we use current prod data, so deprecated packs are not supposed to be passed to this function.
+        # If they are - the API call will fail with a 400 status "item not found" error.
+        logging.info(f"Detected pre-update run mode ('BUCKET_UPLOAD'='{is_post_update}'. "
```
yaakovpraisler, line 402: What does `'BUCKET_UPLOAD'='{is_post_update}'` means? Why BUCKET_UPLOAD will be True if it's a post-update in a regular build?

MichaelYochpaz: This is just a log for debugging, it will look something like:
> Detected pre-update run mode ('BUCKET_UPLOAD'='True'). Packs deprecation status will be checked using Marketplace API.

There is, however, no need to print this variable since there are already different messages if it's true or false, so I'll update that to something like this:
```suggestion
        logging.info(f"Detected pre-update run mode."
```

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -675,33 +662,34 @@ def search_and_install_packs_and_their_dependencies(pack_ids: list,
...
-    packs_to_install: list = []  # we save all the packs we want to install, to avoid duplications
-    installation_request_body: list = []  # the packs to install, in the request format
-    batch_packs_install_request_body: list = []    # list of lists of packs to install if install packs one by one.
+    packs_to_install: list = []  # Packs we want to install, to avoid duplications
+    installation_request_body: list = []  # Packs to install, in the request format
+    batch_packs_install_request_body: list = []  # List of lists of packs to install if not using multithreading .
     # Each list contain one pack and its dependencies.
-    packs_in_the_list_to_install: list = []    # list of packs that are already in the list to install.
+    collected_dependencies: list = []  # List of packs that are already in the list to install.
+    is_post_update = os.getenv('BUCKET_UPLOAD', 'false').casefold() == 'true'
```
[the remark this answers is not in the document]
MichaelYochpaz: This function (`search_pack_and_its_dependencies`) is called both on pre-update, and on post-update.(BUCKET_UPLOAD` is set to `True` when running on post-update. It's a pre-existing env variable that I'm using.

The issue is, that at first, I just used the marketplace API for checking whether a pack is deprecated or not, but then we found that this caused issues, since if we deprecate a pack in the current PR, the API call will fail (pack wasn't uploaded to the bucket, so it returns a message saying the requested pack ID doesn't exist). Like how for example, the Microsoft Defender pack which has a test that deprecated it, failed the build because the API call on post-update failed, since this pack wasn't in the marketplace.

So the solution was, that if we're running on pre-update, we check for deprecation status using the marketplace. If we're running on post-update, we'll check deprecation status locally by checking the file (this way we will see the deprecation status according to the changes on the PR.)

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -35,47 +34,49 @@
...
-    Tests are not being collected for deprecated packs and the pack is not installed in the build process.
+def is_pack_deprecated(pack_id: str, check_locally: bool = True, pack_api_data: dict | None = None) -> bool:
+    """
+    Check whether a pack is deprecated or not.
+    Can be checked locally (pack_metadata.json), or using Marketplace API response data.
+
     Args:
-        pack_path (str): The pack path
+        pack_id (str): ID of the pack to check.
+        check_locally (bool): Whether to check locally (pack_metadata file) or not (will use Marketplace API data instead).
```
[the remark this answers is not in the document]
ilaner: why do we need to check locally? What is the use case to check locally instead query the api?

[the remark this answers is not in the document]
MichaelYochpaz: @ilaner When we're in post-update, the new bucket is created using the branch and not 'master'.
This means, that if I, for example, deprecate a pack named 'X' in my PR, this pack won't be uploaded to the bucket.

Because of this, if I send a marketplace API call to fetch information about 'X' (to check if it's deprecated) during the post-update phase (where the bucket is updated to the branch), the server returns a '400' response code, saying that this ID doesn't exist (since it's not on the marketplace).

We could perhaps just catch such exceptions, and deduce that if we get a '400' status code, the pack is deprecated.
Not sure if this is a good idea, though, and if there can be other cases where we get a '400' response code and want it to fail the build.

Tests/Marketplace/search_and_install_packs.py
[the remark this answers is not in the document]
MichaelYochpaz: After a meeting, we have concluded that checking for whether we're in post-update mode using the `BUCKET_UPLOAD` env var is invalid, since it doesn't cover all cases.

I'm working on a fix for that.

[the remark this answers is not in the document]
MichaelYochpaz: Fixed on 35dd44b0372cfe14040ba1832120765be56ceba0

Tests/Marketplace/configure_and_install_packs.py
```diff
@@ -82,27 +85,28 @@ def xsoar_configure_and_install_flow(options, branch_name: str, build_number: st
     if not content_path:
         raise Exception('Could not find content path')
 
-    # all packs that should be installed
+    # Create a list of all packs that should be installed
     packs_to_install = set(Build.fetch_pack_ids_to_install(options.pack_ids_to_install))
-    logging.info(f'packs to install before filtering by minServerVersion {packs_to_install}')
+    logging.info('Packs to install before filtering by minServerVersion:' + ', '.join(packs_to_install))
```
bziser, line 90: why you changed it from f'string to join?

MichaelYochpaz: Just to have it printed in a more readable manner.
(`minServerVersion: X, Y, Z` instead of `minServerVersion {'X', 'Y', 'Z'}`)

There is however a missing space there
```suggestion
    logging.info('Packs to install before filtering by minServerVersion: ' + ', '.join(packs_to_install))
```

If you think the previous print is better, let me know and I'll revert that.

bziser: If you want to print it like this, I would do it this way:
```suggestion
    logging.info(f'Packs to install before filtering by minServerVersion: {", ".join(packs_to_install)}')
```
But I don't think it is so meaningful that you need to change all the logs to be like this and make them more clumsy

MichaelYochpaz: I'll revert that then

Tests/Marketplace/search_and_install_packs.py
```diff
@@ -675,33 +664,33 @@ def search_and_install_packs_and_their_dependencies(pack_ids: list,
...
         with ThreadPoolExecutor(max_workers=130) as pool:
             for pack_id in pack_ids:
-                if is_pack_hidden(pack_id):
-                    logging.debug(f'pack {pack_id} is hidden, skipping installation and not searching for dependencies')
-                    continue
-                pool.submit(search_pack_and_its_dependencies,
-                            client, pack_id, packs_to_install, installation_request_body, lock,
-                            packs_in_the_list_to_install, install_packs_one_by_one,
-                            batch_packs_install_request_body)
+                pool.submit(search_pack_and_its_dependencies, pack_id=pack_id, **kwargs)
```
ilaner, line 693: No need to check if it's hidden anymore?

ilaner (approved):
Good job!

yaakovpraisler (approved):
Looks good to me as well

bziser (approved):

MichaelYochpaz:
Ran a test whether the original ticket's issue (deprecating a pack, and removing it as a dependency from other packs causes build issue) has been resolved:

Reproducing the issue on master:
https://github.com/demisto/content/pull/27598/commits/4b3936b46650ab6197c4e6d702da4aa589326811
https://code.pan.run/xsoar/content/-/pipelines/5799866

Then when merging the current fix branch into the testing branch:
https://github.com/demisto/content/pull/27598/commits/c840c2a4a8e975f39a7b654ba5f7fc15fc8b97af
https://code.pan.run/xsoar/content/-/pipelines/5802635

PR #27416, opened 2023-06-13
fix Test_XDR_Playbook_general_commands TPB

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-6961)

## Description
Instead of sending arbitrary alert that might be removed by XDR, we query all the correlation alerts we have in the instance and then put the first one we find.

In that way its being done dynamically rather then giving specific `alert_id` as an input.

image

image

image

image


ostolero (approved):
Great work!

PR #27530, opened 2023-06-18
Delete offset from CrowdStrike fetch

ilaner:
## Related Issues
related: https://jira-hq.paloaltonetworks.local/browse/XSUP-25354
fixes: [CrowdStrike Missing incidents](https://jira-hq.paloaltonetworks.local/browse/XSUP-24966)


Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -2335,18 +2326,14 @@ def fetch_incidents():
...
-        last_incident_fetched = current_fetch_info_incidents.get('last_fetched_incident')
-        new_last_incident_fetched = ''
-
         fetch_query = demisto.params().get('incidents_fetch_query')
 
         if fetch_query:
-            fetch_query = "start:>'{time}'+{query}".format(time=last_fetch_time, query=fetch_query)
-            incidents_ids = demisto.get(get_incidents_ids(filter_arg=fetch_query, offset=offset), 'resources')
+            fetch_query = "start:>'{time}'+{query}".format(time=start_fetch_time, query=fetch_query)
+            incidents_ids = demisto.get(get_incidents_ids(filter_arg=fetch_query), 'resources')
```
yuvalbenshalom, line 2333: I am not sure the integration knows how to handle a case where we get more new incidents than the limit, without using the offset at all.

yuvalbenshalom (commented):
see my comment, I wold like to understand the order that the detections are returned from CS and what happens if we get more detections than the limit, not sure it will work without the offset at all.

ilaner:
@yuvalbenshalom 
The lookback handles this scenario and increases the limit.
However, we need to provide the new limit from look back to the query, did here

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -2298,7 +2284,11 @@ def fetch_incidents():
     if not isinstance(last_run, list):
         last_run = migrate_last_run(last_run)
     current_fetch_info_detections: dict = last_run[0]
+    current_fetch_info_detections.pop("offset", None)
+
     current_fetch_info_incidents: dict = last_run[1]
+    current_fetch_info_incidents.pop("offset", None)
```
yuvalbenshalom, line 2290: can we move this to migrate?

yuvalbenshalom (changes requested):
see my comments, and please add release notes.
tag this as fixes both relevant issues.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
ilaner: moved

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -2252,17 +2252,7 @@ def get_mapping_fields_command() -> GetMappingFieldsResponse:
...
-    offset = current_fetch_info.get('offset', 0)
-    if not last_fetch_time:
-        last_fetch_time, _ = parse_date_range(FETCH_TIME, date_format=DATE_FORMAT)
-    prev_fetch = last_fetch_time
-    last_fetch_timestamp = int(parse(last_fetch_time).timestamp() * 1000)
-    return last_fetch_time, offset, prev_fetch, last_fetch_timestamp
-
-
-def migrate_last_run(last_run: dict[str, str]) -> list[dict]:
+def migrate_last_run(last_run: dict[str, str] | list[dict]) -> list[dict]:
```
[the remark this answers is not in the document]
ilaner: added

yuvalbenshalom (approved):

PR #27541, opened 2023-06-18
Fixed empty installation 

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Description
We always try to install packs on the servers, even if the packs list is empty. Added a check on that list and considered the installation as a success in case there are no packs to install.

### DIFF
--- a/Tests/Marketplace/search_and_install_packs.py
+++ b/Tests/Marketplace/search_and_install_packs.py
@@ -340,7 +340,7 @@ def install_packs(client: demisto_client,
     """
     global SUCCESS_FLAG
     if not packs_to_install:
-        logging.info("There is not packs to install on servers. Consolidating installation as success")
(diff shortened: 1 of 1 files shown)


Tests/Marketplace/search_and_install_packs.py
```diff
@@ -339,6 +339,9 @@ def install_packs(client: demisto_client,
         sleep_interval (int): The sleep interval, in seconds, between install attempts.
     """
     global SUCCESS_FLAG
+    if not packs_to_install:
+        logging.info("There is not packs to install on servers. Consolidating installation as success")
```
yaakovpraisler, line 343:
```suggestion
        logging.info("There are no packs to install on servers. Consolidating installation as success")
```

yaakovpraisler (approved):

kobymeir (approved):

PR #27683, opened 2023-06-24
Align credentials stores - part 22

maimorag:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
Relates:[ link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5274)

## Description
Change the type of credentials, from type 4 to type 9.

- [x] 6.5.0

  - [x] No

### DIFF
--- a/Packs/DuoAdminApi/Integrations/DuoAdminApi/DuoAdminApi.py
+++ b/Packs/DuoAdminApi/Integrations/DuoAdminApi/DuoAdminApi.py
@@ -7,8 +7,8 @@ from CommonServerPython import *  # noqa: F401
 # Setup
 
 HOST = demisto.getParam('hostname')
-INTEGRATION_KEY = demisto.getParam('integration_key')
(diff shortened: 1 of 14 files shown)


Packs/DuoAdminApi/Integrations/DuoAdminApi/DuoAdminApi.py
```diff
@@ -8,7 +8,7 @@
 
 HOST = demisto.getParam('hostname')
 INTEGRATION_KEY = demisto.getParam('integration_key')
```
Shellyber, line 10: I think the INTEGRATION_KEY should be part of the credentials (the identifier). Please verify and update accordingly

maimorag: updated

Shellyber (approved):
Nice

PR #27724, opened 2023-06-26
added mp to tim processing layouts

yucohen:
changed the mp to be xsoar only


yaakovpraisler (approved):

altmannyarden (approved):
You might need to add release notes

PR #27732, opened 2023-06-26
update ruff rules

dorschw:
Only using ruff rules that are one of the following:
1. don't exist in the repo
2. are autofixable
3. hand-picked


pyproject.toml
```diff
@@ -384,9 +439,8 @@ ignore = [
 ]
 extend-exclude = ["CommonServerPython", "test_data"]
 
-# Allow autofix for all enabled rules (when `--fix`) is provided.
 # fixable = []
-# unfixable = []
+unfixable = ["RUF001"]
```
kobymeir, line 443: the name of the check is missing 
```suggestion
unfixable = [
    "RUF001" # ambiguous-unicode-character-string
]
```

kobymeir (changes requested):

kobymeir (approved):

PR #27904, opened 2023-07-04
Core REST API - fix test module

adi88d:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-7209)

## Description
test-module now is making request to **user** suffix without checking the response data,
this fix raise exception if the response is invalid.

## Screenshots
image

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [ ] 6.5.0

   - [ ] Yes
       - Further details:
   - [ ] No

## Must have
- [ ] Tests
- [ ] Documentation


DeanArbel (approved):

dansterenson:
@DeanArbel @adi88d Are we sure thats the right solution? I think the proper solution is to add the fix to the SendRequest function so it fill fail with a proper message when the HTML response is returned.

adi88d:
> @DeanArbel @adi88d Are we sure thats the right solution? I think the proper solution is to add the fix to the SendRequest function so it fill fail with a proper message when the HTML response is returned.

But in this case we want to fix the **test-module** command, maybe we want to send some request with html response? we dont want it to fail

PR #27922, opened 2023-07-04
enhance pre-commit

ilaner:
* Add poetry hooks to validate the poetry relevant files
* Add update-additional-dependencies hook to auto-update the dependencies for mypy


poetry.lock
dorschw, line 1: are the changes to `poetry.lock` necessary?

ilaner: yes, they trigger the `update-additional-dependencies` step

.pre-commit-config_template.yaml
```diff
@@ -15,6 +15,15 @@ repos:
       - id: name-tests-test
         files: .+_test.py$
 
+  - repo: https://github.com/python-poetry/poetry
```
dorschw, line 18: can we make it only run when pyproject/poetry.lock are changed

dorschw (approved):

PR #27938, opened 2023-07-05
Fixed cveReputationV2 to handle None value

Ni-Knight:
## Status
- [x] In Progress
- [ ] Ready

## Description
Fixes an issue where the script failed when cvss was None

- [ ] 6.0.0
- [ ] 6.1.0
- [ ] 6.2.0
- [x] 6.5.0

   - [ ] Yes
       - Further details:
   - [x] No


Packs/CommonScripts/Scripts/CveReputationV2/CveReputationV2.py
```diff
@@ -17,7 +17,14 @@ def get_dbot_score(resCmd: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
             data = cve.get('Contents', {})
-            cvss = int(data.get('cvss', -1))
+            cvss = data.get('cvss', -1)
+
+            if isinstance(cvss, dict):
+                cvss = float(data.get('cvss').get('Score', -1))
+            elif not cvss:
+                cvss = -1
+            else:
+                cvss = float(cvss)
```
yuvalbenshalom, line 27: the original code dose not handle the case where data.get('cvss').get('Score', -1) returned None, so float() will fail. 
suggestions(mind the indentations):
```suggestion
            if not cvss:
                cvss = -1
            else if isinstance(cvss, dict):
                Score = data.get('cvss').get('Score', -1))
                if Score:
                    cvss =  float(Score)
                else:
                    cvss = -1
```
only thing is that I don't think we have tests for the 4 possiblties flows.

yuvalbenshalom (commented):
see my comment

yuvalbenshalom (approved):

PR #28082, opened 2023-07-11
Fix unsecure variable mismatch and align across integrations

ryokubaka:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
None

## Description
The integrations for Recorded Future have the ability to ignore SSL verification, however some integrations have a mismatch between the property variable and the variable used in the actual code execution. This mismatch results in SSL verification even when the associated property is set to `true`.

This PR addresses the mismatch, and aligns all SSL verification overrides to `unsecure`, rather than `insecure`.

Changes:
 - RecordedFuture
   - Changes the property variable from `unsecure` -> `insecure` 
   - Changes the code reference variable from `unsecure` -> `insecure` to align with standards
 - RecordedFutureLists
   - Fixes the mismatched code variable from `unsecure` -> `insecure`
 - RecordedFuturePlaybookAlerts
   - Fixes the mismatched code variable from `unsecure` -> `insecure`
 

N/A

   - [ ] Yes
       - Further details:
   - [X] No


maimorag:
Hi @ryokubaka,
Our standard is using "insecure".
Could you please elaborate on which behavior you're trying to implant?

ryokubaka:
Hi @maimorag - I can switch this to `insecure` if that is the standard. I used `unsecure` as that is what it seemed like the integration maintainer (Recorded Future) was using - but I don't know that it really matters, so long as it works.

The expected behavior is that when `insecure = true`, SSL verification is disabled.

maimorag (approved):

PR #28116, opened 2023-07-12
Remove PEP484 hook

dorschw:
will be replaced with ruff's RUF013
related: https://github.com/demisto/demisto-sdk/pull/3227


dorschw:
done in #3227

PR #28132, opened 2023-07-13
Integrations & Scripts: added the missed config param field "required: false"

ilappe:
## Related Issues
related: https://jira-hq.paloaltonetworks.local/browse/CIAC-4906

## Description
Added the missed field `required: false` in the configuration params for params doesn't contained `required`

run `git diff --cached | grep -E '^(\-|\+) '` to be sure only `required: false` is changed
results: 7023 entries of `+  required: false`
in 1013 yml files


DeanArbel (approved):

ilappe:
[content nightly ](https://code.pan.run/xsoar/content/-/pipelines/5761053)

ilaner:
the `unit-test-and-lint` failures are only `mypy` errors, which will not fail in `lint -a`, so we can force merge.

ilappe:
[content nightly after force merge](https://code.pan.run/xsoar/content/-/pipelines/5787040)

PR #28141, opened 2023-07-13
fix regex of getting the version of the server

eyalpalo:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
fix the regex of getting the version of the server. During the upload version, we need to get the version of the server to determine which packs to install. We had a bug in the regex as we searched using match and not search.
This bug caused us to return the default version


eyalpalo:
@yucohen FIY

bziser (commented):
please add test

yucohen (approved):

bziser (approved):

PR #28311, opened 2023-07-19
Darktrace Event Collector - Added the marketplacev2 to the yml file

AradCarmi:
Fix for missing `marketplacev2` for the event collector.


AradCarmi:
A force merge is needed as the event collector YAML file is missing the marketplace V2 field and we want to fix this without updating the pack version.

dantavori (approved):

PR #28347, opened 2023-07-20
added marketplacev2

AradCarmi:
Fix master [failure](https://code.pan.run/xsoar/content/-/pipelines/5818608)


AradCarmi:
Force merge is needed as the PR fixes master [issue](https://code.pan.run/xsoar/content/-/jobs/29519723).

PR #28351, opened 2023-07-20
CrowdStrike Falcon: add condition to before get_incidents_entities

ilappe:
## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-26344)

## Description
due to changes in [this pr](https://github.com/demisto/content/issues/27530)
an if condition is removed and this lead to call the function `get_incidents_entities` with empty `incidents_ids` and the API throw `400 - reason: Bad Request`


ilaner (changes requested):
can you change in all incident types?

ilappe:
> can you change in all incident types?

@ilaner not relevant to other types

ilaner (approved):
Great job!

PR #28409, opened 2023-07-23
Read Reviewers from Content Roles JSON

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7508

## Description
Read and assign Content and Security reviewers from `content-roles.json` instead of constants in `handle_external_pr.py`.

## Must have

- [x] Tests: https://github.com/demisto/content/pull/28487
- [x] Documentation


.github/workflows/handle-new-external-pr.yml
```diff
@@ -32,11 +32,17 @@ jobs:
         run: |
           poetry install --with ci
 
+      - name: Set Absolute Content Roles File Path to Env
+        id: content_roles_path
+        run: echo "CONTENT_ROLES_FILE_PATH=$(realpath .github/content_roles.json)" >> "$GITHUB_ENV"
```
kgal-pan, line 37: I don't like this approach (setting env vars with the path to the file). I think the better solution IMO would be to supply the path as an argument to `handle_external_pr.py`. The thing is that `handle_external_pr.py` isn't implemented as a CLI (i.e. `argparse` or `click`) and already utilizes env vars for reading other CI inputs (e.g. `EVENT_PAYLOAD`, `CONTENTBOT_GH_ADMIN_TOKEN`) so I continued with this train-of-thought. 

I think to generify `handle_external_pr` we need to refactor the whole thing and not sure it's worth the investment at this point. 

Let me know your thoughts!

dorschw (approved):

.github/workflows/handle-new-external-pr.yml
kgal-pan: Modified to use `utils.load_json`.

PR #28428, opened 2023-07-23
Carbon Black EDR (Live Response API): fix get file issue on Linux

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-25715

## Description
When trying to get file from Linux endpoint - the command failed due to wrong code that try to get the parent dir in order to know the file size.

i remove it  as the size of bytes to get in getFile is optional 
see :
https://github.com/vmware-archive/cbapi/blob/master/sensor_apis/example/cblr_cli.py#L869
https://developer.carbonblack.com/reference/enterprise-response/latest/live-response-api/#get-file


ilappe:
[integration-Carbonblackliveresponse.yml.txt](https://github.com/demisto/content/files/12139919/integration-Carbonblackliveresponse.yml.txt)

DeanArbel (approved):
Nice

PR #28507, opened 2023-07-25
Set Python 3.10 to Contribution Form Verification and Create Internal PR flows

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7571

## Description
`ruff` precommit hook modified the line in https://github.com/demisto/content/pull/28409/files#diff-c6f310753e3d75d96f4db44595fbcdbe9f12c0c2d0d0c4625ffb38379c7ca223R20 as a part of [PEP 605 syntax](https://beta.ruff.rs/docs/rules/non-pep604-annotation/).

image

This change caused the following build failures:

- https://github.com/demisto/content/actions/runs/5660517661/job/15336567674
- https://github.com/demisto/content/actions/runs/5660664006/job/15336881770

According to https://github.com/python/cpython/issues/86399#issuecomment-1093889908, the ability to use bitwise OR with generic types was added in Python 3.10. This PR sets the Python version to 3.10 in 2 flows (Contribution form filled verification, Create Internal PR from External Merged).

## Must have
- [ ] Tests
   - [x] Contribution form filled verification (when not filled): https://github.com/demisto/content/actions/runs/5667172366/job/15355365700?pr=28525#step:6:14
   - [x] Contribution form filled verification (when filled): https://github.com/demisto/content/actions/runs/5667201125/job/15355452721?pr=28525#step:6:15
   - [x] Create Internal PR from External Merged: https://github.com/demisto/content/actions/runs/5667247630/job/15355593358
- [ ] Documentation


kgal-pan:
@dorschw - Should we consider removing https://github.com/demisto/content/blob/master/pyproject.toml#L393 as well?

ShacharKidor (approved):

kgal-pan:
After further review, I found in https://github.com/python/cpython/issues/86399#issuecomment-1093889925 that generic types for OR were only added in Python 3.10. In the `create-internal-pr-from-external` workflow, we're installing Python 3.9:

```bash
yq '.jobs.create_internal_pr.steps[1].with' .github/workflows/create-internal-pr-from-external.yml
python-version: '3.9'
```

So I think the better solution here would be to set `python-version: 3.10` instead. That way we don't have to comment out `UP007` from `ruff`. @ShacharKidor @ilaner - WDYT?

ShacharKidor:
> After further review, I found in [python/cpython#86399 (comment)](https://github.com/python/cpython/issues/86399#issuecomment-1093889925) that generic types for `OR` were only added in Python 3.10. In the `create-internal-pr-from-external` workflow, we're installing Python 3.9:
> 
> ```shell
> > yq '.jobs.create_internal_pr.steps[1].with' .github/workflows/create-internal-pr-from-external.yml
> 
> python-version: '3.9'
> ```
> 
> So I think the better solution here would be to set `python-version: 3.10` instead. That way we don't have to comment out `UP007` from `ruff`. @ShacharKidor @ilaner - WDYT?

@ kgal-pan - it's a good idea to update, but we will need to test it carefully to ensure it won't break anything else. 
Also, this is not the only flow that was broken due to this ruff change - according to this [error](https://github.com/demisto/content/actions/runs/5660664006/job/15336881770) we need to fix also the `contribution_form_filled_verification` flow.

kgal-pan:
> this is not the only flow that was broken due to this ruff change - according to this [error](https://github.com/demisto/content/actions/runs/5660664006/job/15336881770) we need to fix also the `contribution_form_filled_verification` flow.

Actually, both job failures I mentioned in the description (which is the one you've just mentioned in this comment) stem from the same module (`utils.py:L20`):

```
  File "/home/runner/work/content/content/Utils/github_workflow_scripts/utils.py", line 20, in <module>
    CallArgs = Iterable[tuple[Any] | tuple[Any, dict]]
TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'types.GenericAlias'
```

So whichever solution we choose, it'll fix both flows.

ShacharKidor:
@kgal-pan - ok good! so let's try to update the python version, and test it carefully to see if it works smoothly. 
@ilaner - do you agree?

kgal-pan:
- Contribution Form Filled label verification (both when filled and when not) jobs ran successfully (no `TypeError` thrown), see https://github.com/demisto/content/actions/runs/5667201125/job/15355452721?pr=28525#step:6:15.

ilaner:
We assume in the `pre-commit` command that the infra files run in python 3.10. We should verify that all infra run in 3.10 and not in versions below.

PR #28530, opened 2023-07-26
Mitre indicators by open incidents bug fix

dansterenson:
Managed to reproduce. When the server http response was as follows:
res = {'total': 0, 'iocs': None, 'searchAfter': None, 'accountErrors': None, 'totalAccounts': 0}

res.get('iocs', []) will return [] only if iocs key does not exist in the dict. if the iocs is None in the dict it will return None and fail afterwords in the for loop.

Added UT to cover this case.

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-26443)

## Must have
- [x] Tests
- [x] Documentation


DeanArbel (approved):
Good job

PR #28567, opened 2023-07-27
Added the option to pass chunk size to send_events_to_xsiam

ShahafBenYakir:
## Description
Added the option to pass chunk size to send_events_to_xsiam so specific integration can request for a different chunk size than the default one.


DeanArbel (approved):
Looks good, please consider putting an upper limit to chunk_size, as the current API limit is up to 10Mib.

PR #28666, opened 2023-08-01
Remove `BA125` from `.pack-ignore`

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Description
We originally added `BA125` to `.pack-ignore` of these packs because their integrations have `test-module` listed as a command in their YAML file, causing the validation to fail when people run `format` and don't remove the added`test-module` section in their README file.

Now that the SDK version has been updated on the `content` repository, we can restore the validation for these integrations and trust that people won't cause master's build to fail, since the validation will fail during the PR stage, blocking them from merging, and they'll have to remove the `test-module` section from the README file to merge.

### DIFF
--- a/Packs/HarfangLabEDR/.pack-ignore
+++ b/Packs/HarfangLabEDR/.pack-ignore
@@ -1,8 +1,5 @@
 [file:Hurukai.yml]
-ignore=IN126,BA124,BA125
-
-[file:README.md]
-BA125
+ignore=IN126,BA124
 
 [known_words]
 harfanglab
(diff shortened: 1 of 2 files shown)


bziser (approved):

bziser:
@MichaelYochpaz  You ran validate on these files to see that it passing, right?

MichaelYochpaz:
@bziser Yep, both integrations' READMEs are passing the validation (we removed the `test-module` section in the last fix, and they haven't changed since).

PR #28800, opened 2023-08-06
Pack upload support

ostolero:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Must have
- [ ] Tests
- [ ] Documentation


Tests/Marketplace/upload_packs.py
```diff
@@ -1196,9 +1196,12 @@ def option_handler():
                               "https://googleapis.dev/python/google-api-core/latest/auth.html"),
                         required=False)
     parser.add_argument('-d', '--pack_dependencies', help="Full path to pack dependencies json file.", required=False)
-    parser.add_argument('-p', '--pack_names',
+    parser.add_argument('-pn', '--pack_names',
                         help=("Target packs to upload to gcs."),
                         required=True)
+    parser.add_argument('-p', '--pack_names_flag',
+                        help=("Indication if the -p flag is used and only specific packs are uploded"),
+                        default=False)
```
JudahSchwartz, line 1204: what are both these flags now? Doesnt at all seem intuitive to need both

ostolero: We needed to add a flag that indicates we are using the -p option. 
This function receives a list of packs chosen by collect_tests  in the -p flag in the original implementation (regardless if we initially entered packs with -p, whatever is chosen to upload is passed here). so I wanted to keep this initial list but also have a way to signal we are really using the -p option. I can change the names if you think it would be clearer.

JudahSchwartz: i gues then lat che=ange to something like -upload-specific-packs or something that sounds more like a mode.

But it seems to be more intuitive to control it with 1 flag. if no packs are provided, I know i need to upload all. So maybe in the case of is_bucket_upload_flow we can simply not pass the pack-names flag?

yaakovpraisler: I agree with @JudahSchwartz, let's change the flag name to something like `--upload-specific-pack`

Tests/scripts/prepare_content_packs_for_testing.sh
```diff
@@ -84,9 +84,10 @@ else
 fi
 echo "BUCKET_UPLOAD = $BUCKET_UPLOAD, FORCE_BUCKET_UPLOAD = $FORCE_BUCKET_UPLOAD"
 
+PACK_UPLOAD_FLAG=false
```
yaakovpraisler, line 87: Also here let's change the name to `UPLOAD_SPECICF_PACKS`
and I think we can configure this variable from the beginning of the process, meaning when triggering the flow we can pass the `UPLOAD_SPECICF_PACKS=true` and of course by default it will be false.

yaakovpraisler (commented):
Looks very good!
Commented just on the var names

yaakovpraisler (approved):
Good job! 💪 
Don't forget to remove the debug print logs.

PR #28864, opened 2023-08-09
Fix destroy services to consider successful runs

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Description
The shutdown command was always getting a TTL and therefore always shut down with TTL, and did not consider the test succession (we want to shutdown the machine if the tests pass).
Also, moved the files that indicates rather the test had pass to the artifacts folder


kobymeir (approved):

yucohen:
The test-upload-flow set fails due to a known and unrelated issue with some pack readme

PR #28885, opened 2023-08-10
Panorama get address group command fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-27023)

## Description
Changed "show" to "get" in the API call params. When using "get" it gets the candidate config (before commit) instead of "show" param which gets only the active configuration (which is the configuration after the commit) :


GuyAfik (approved):
nice!

Packs/PAN-OS/ReleaseNotes/2_1_1.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Fixed an issue where the ***panorama-get-address-group*** command did not get address groups that were not committed.
```
GuyAfik, line 6:
```suggestion
- Fixed an issue where the ***pan-os-get-address-group*** command did not get address groups that were not committed.
```
panorama-get-address-group is deprecated.

dansterenson:
Force merge needed: 
Failed Tests:  - PAN-OS - Add Anti-Spyware Security Profile To Rule - Test (Mock Disabled)

Fails on 6.8.0 machine, the test playbook fromversion is 6.9.0 and it seems like a bug in the build.

PR #28930, opened 2023-08-13
added generic method to retry requests to XSIAM machines

eyalpalo:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-7715)

## Description
Added a generic method to implement the retry mechanism when requesting the machines to uninstall/install and similar.


Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +19,128 @@
...
+    Returns:
+        True if the pack is still installed, False otherwise.
+
+    """
+    def success_handler(response_data):
+        installed_packs = ast.literal_eval(response_data)
+        installed_packs_ids = [pack.get('id') for pack in installed_packs]
+        if pack_id in installed_packs_ids:
+            return True, None
+        return False, None
```
kobymeir, line 131:
```suggestion
            return (True, None) if pack_id in installed_packs_ids else (False, None)
```

kobymeir (changes requested):

daryakoval:
Why the generic function is in the uninstall file? Maybe we need to move to some tools file, and use it in all existing functions that we did retries there?

eyalpalo:
@daryakoval any idea of a specific file we should move it into?

kobymeir (approved):

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -183,51 +280,40 @@ def uninstall_pack(client: demisto_client,
...
+        if ALREADY_IN_PROGRESS in ex.body:
+            wait_succeeded = wait_until_not_updating(client)
+            if not wait_succeeded:
+                raise Exception(
+                    "Failed to wait for the server to exit installation/updating status"
+                ) from ex
+
+    return generic_retries_request(client=client,
+                                   retries_message=f'Failed to uninstall pack: {pack_id}',
+                                   exception_message='The request to uninstall packs has failed.',
```
dantavori, line 308: de we need to differentiate between these two? they are pretty much the same

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +19,126 @@
...
+    """
+    try:
+        for attempt in range(attempts_count - 1, -1, -1):
+            try:
+                if should_try_handler and not should_try_handler():
+                    # if the method exist and we should not try again.
+                    return True, None
+
+                # should_try_handler return True, we are trying to send request.
+                logging.info(f"{prior_message}, Attempt: {attempts_count - attempt}/{attempts_count}")
```
dantavori, line 65:
```suggestion
        for attempt in range(attempts_count):
            try:
                if should_try_handler and not should_try_handler():
                    # if the method exist and we should not try again.
                    return True, None

                # should_try_handler return True, we are trying to send request.
                logging.info(f"{prior_message}, attempt: {attempts_count +1}/{attempts_count}")
```
otherwise `attempt` should be named `attempts_left`

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +19,126 @@
 ALREADY_IN_PROGRESS = "create / update / delete operation is already in progress (10102)"
 
 
+def generic_retries_request(client: demisto_client,
```
dantavori, line 22:
```suggestion
def generic_request_with_retries(client: demisto_client,
```

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +19,126 @@
...
+                    http_exception_handler(http_ex)
+                if not attempt:  # exhausted all attempts, understand what happened and exit.
+                    raise Exception("Failed to perform http request to the server") from http_ex
+                logging.debug(f"Process failed, got error {http_ex}")
+
+            # There are more attempts available, sleep and retry.
+            logging.debug(f"{retries_message}, sleeping for {sleep_interval} seconds.")
+            sleep(sleep_interval)
+    except Exception as e:
+        logging.exception(f'{exception_message}. Additional info: {str(e)}')
```
dantavori, line 106: we are not raising the exception on purpose?

dantavori (commented):
looks good, see notes below:

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +19,126 @@
...
+
+                # should_try_handler return True, we are trying to send request.
+                logging.info(f"{prior_message}, Attempt: {attempts_count - attempt}/{attempts_count}")
+                response, status_code, headers = demisto_client.generic_request_func(client,
+                                                                                     path=path,
+                                                                                     method=method,
+                                                                                     accept=accept,
+                                                                                     _request_timeout=request_timeout)
+
+                if 200 <= status_code < 300 and status_code != 204:
```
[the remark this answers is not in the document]
eyalpalo: This check was part of the request mechanism we had before, maybe @daryakoval  could answer in what cases we get 204.

Tests/Marketplace/search_and_uninstall_pack.py
eyalpalo: Yes, tried to keep the implementation we had before. It does fail the build itself as we log exception.

Tests/Marketplace/search_and_uninstall_pack.py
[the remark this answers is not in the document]
kobymeir: 204 in some cases the un/install packs returned a 204, which isn't a good response, it means that the operation didn't work on their end.

Tests/Marketplace/search_and_uninstall_pack.py
kobymeir: @eyalpalo  i prefer the option for changing the variable name, rather than changing the loop direction

Tests/Marketplace/search_and_uninstall_pack.py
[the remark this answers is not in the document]
dantavori: sounds specific to the endpoint - 204 means "successful delete" which may be valid in different API endpoints

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +20,126 @@
...
+    """
+    try:
+        for attempts_left in range(attempts_count - 1, -1, -1):
+            try:
+                if should_try_handler and not should_try_handler():
+                    # if the method exist and we should not try again.
+                    return True, None
+
+                # should_try_handler return True, we are trying to send request.
+                logging.info(f"{prior_message}, attempts_left: {attempts_count - attempts_left}/{attempts_count}")
```
dantavori, line 66:
```suggestion
                logging.info(f"{prior_message}, attempt: {attempts_count - attempts_left}/{attempts_count}")
```

Tests/Marketplace/search_and_uninstall_pack.py
```diff
@@ -18,6 +20,126 @@
...
+                response, status_code, headers = demisto_client.generic_request_func(client,
+                                                                                     path=path,
+                                                                                     method=method,
+                                                                                     accept=accept,
+                                                                                     _request_timeout=request_timeout)
+
+                if 200 <= status_code < 300 and status_code != 204:
+                    if success_handler:
+                        # We have a method to run as we were returned a success status code.
+                        return success_handler(response)
```
dantavori, line 76: correct me if I'm wrong, except here we always return `None` in the second item of the response tuple, no?

also, consider returning the response itself by default instead of the status bool, because then the only way to return the response data using this method is by providing `success_handler=lambda res: res`.
if you want additional information like status, a common practice is to do something like:
```
res, err = func()
```
where `err` is the error data or `None` if no error.

PR #28936, opened 2023-08-13
Update core_packs_mpv2_list.json

ShahafBenYakir:
Removed default playbook from core packs list


idovandijk (approved):

kgal-pan:
https://github.com/demisto/content/pull/27370 - causing failure in https://code.pan.run/xsoar/content/-/jobs/30962401.

ShahafBenYakir:
Merging as NG can't lock a mission and in any case this is not related to NG

PR #29069, opened 2023-08-20
Update Active Directory Query V2 to use LDAP docker image

anas-yousef:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24646

## Description
Changed the docker image of the **Active Directory Query V2** integration to the LDAP docker image, which now uses a [debian](https://github.com/demisto/dockerfiles/pull/18780) based docker image.

## Must have
- [ ] Tests
- [ ] Documentation


Tests/conf.json
```diff
@@ -2543,29 +2543,34 @@
             "playbookID": "Active Directory Test",
             "integrations": "Active Directory Query v2",
             "instance_names": "active_directory_ninja",
+            "memory_threshold": 100,
```
ShacharKidor, line 2546: @anas-yousef  - please explain those changes.

ShacharKidor (approved):

Tests/conf.json
anas-yousef: The build was failing, as can be seen [here](https://code.pan.run/xsoar/content/-/jobs/31474955#L2562) (line 2562), due to insufficient memory.

JasBeilin:
Need a force merge due to known NG issues.

anas-yousef:
This PR requires a force merge because:
The [XSOAR NG](https://code.pan.run/xsoar/content/-/jobs/31542094) build is failing on a test playbook (Active Directory - manual pagination check), and when I investigated the test playbook on the build machine, I can see that the tasks are configured without any inputs, which makes the test playbook fail. I tried running the test playbook on a disposable tenant, and the test playbook passed and all tasks are show with inputs, and the test playbook passes on all XSOAR 6.x build machines.
image (3)
image (4)
This is from my XSOAR NG disposable tenant, where we can see that the test playbook does indeed succeed.
Active_Directory_-_manual_pagination_check_Tue_Aug_22_2023

PR #28279, opened 2023-07-19
new github workflow for new label

edik24:
## Contributing to Cortex XSOAR Content

When a community contribution is done for partner pack - need to get partner to review and approve.
A new label was added to the content repository and a new github workflow that will fail if the "partner-approved" label was not added to the PR.

If the contribution form verified that this is a partner - the label will be added automatically.
If not, reviewer will need to add this label to pass the new workflow

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-6819)

## Description
A few sentences describing the overall goals of the pull request's commits.

## Must have
- [ ] Tests
- [ ] Documentation

PR #29172, opened 2023-08-23
Fix gtihub workflow for partner-approved label flow

edik24:

The initial [PR](https://github.com/demisto/content/pull/28279)
had a condition that made the partner-approved flow to run only when PR is opened and not when something changes in the PR.
Fixed it.

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/check-partner-approved-label.yml
+++ b/.github/workflows/check-partner-approved-label.yml
@@ -9,7 +9,7 @@ permissions:
 jobs:
   check_label_job:
     runs-on: ubuntu-latest
-    if: github.repository == 'demisto/content' && github.event.action == 'opened' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
+    if: github.repository == 'demisto/content' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/check-partner-approved-label.yml
```diff
@@ -9,7 +9,7 @@ permissions:
 jobs:
   check_label_job:
     runs-on: ubuntu-latest
-    if: github.repository == 'demisto/content' && github.event.action == 'opened' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
+    if: github.repository == 'demisto/content' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
```
ShacharKidor, line 12: So we want it to run not only when the PR is opened right? that is the fix.

ShacharKidor (changes requested):
@edik24 , 

Please explain in the pr body what happened here, and add a link to the PR you added with the workflow.

.github/workflows/check-partner-approved-label.yml
edik24: Yes, in this case the flow was running only when PR opened but not after the change of the branch

ShacharKidor (approved):

PR #29204, opened 2023-08-24
removed release_notes_generator from prepare testing bucket

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8015

## Description
The release notes generator is not needed. I checked, and we used its outputs only then the test-instance.yml was running, but we do not run it anymore

## Must have
- [ ] Tests
- [ ] Documentation


kobymeir (approved):

yucohen:
deleting a file causing validations to fail. This file is no longer in use

PR #29218, opened 2023-08-25
Graph single mail html

Anonymoo53:
## Status
- [x] In Progress
- [] Ready
- [] In Hold - (Reason for hold)

## Related Issues
fixes: link to the issue

## Description
This update clones functionality from the GraphMail integration for retrieving the HTML bodies of email into the Listener integration. The listener integration does not pull the HTML bodies of email and is hardcoded to only pull text based representations of emails. This change does not completely align the two integrations and there are still significant differences between them. 

## Must have
- [x] Tests
- [x] Documentation


RosenbergYehuda (commented):
Awesome work, @Anonymoo53!

You've done an excellent job, and I really appreciate it. I want to mention that when we're using the same code for multiple integrations, it's a good idea to write it once in the MicrosoftGraphMailApiModule (in this case). Then, both integrations can use it without duplicating the code.

This approach simplifies things because we only have to deal with one version of the code for tasks like testing and fixing bugs. So, could you please take out all the duplicated code from both sets and put it into the module? 
Thanks!

Anonymoo53:
Hi @RosenbergYehuda,

The changes have been made to push all duplicate code into the shared module. The original version of the APIModule does not contain any unit tests but all the migrated code is covered by the existing code - or tests that have been aded in this PR - within the individual integrations. Given the size of the current test files it is not insignificant to compare all of the existing test functionality and migrate these into the module. If possible, I would like to look at completing this current PR as-is, leaving the unit tests to be completed at the integration-level and then migrate the relevant unit tests to the API module at a later date.

Anonymoo53:
@RosenbergYehuda Broke this PR with recent changes. Will raise a new one after fixing.

PR #29233, opened 2023-08-27
fixing pr to jira linking issue

kobymeir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixed: https://jira-hq.paloaltonetworks.local/browse/CIAC-8124

## Description
The patterns matching only supported "fixes" and relates" and false support for "related" and "fixed"


Utils/github_workflow_scripts/jira_integration_scripts/link_pr_to_jira_issue.py
```diff
@@ -68,24 +68,25 @@ def trigger_generic_webhook(options):
...
     # Handle cases where the PR did not intend to add links:
     if ("fixes:" not in pr_body.lower()
             and "relates:" not in pr_body.lower()
             and "fixed:" not in pr_body.lower()
             and "related:" not in pr_body.lower()):
-        print("Did not detect Jira linking pattern.")
-        return
+        print("Did not detect Jira linking pattern.")  # noqa: T201
+        # Exiting with exit code 1 will fail the workflow, blocking the PR from being merged.
+        sys.exit(1)
```
JasBeilin, line 80: We do not want to fail if it was not added on purpose...

JasBeilin (changes requested):
Great, just check my comment as we do not want to fail the build if it should not be linked.

JasBeilin (approved):

PR #29294, opened 2023-08-29
[pre-commit template file] Fix validate hook

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Description
Since there are situations where we delete the `-g` flag [here](https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/pre_commit/hooks/validate_format.py#L11), therefore it must be last in the list.
After the changes I make in this [PR](https://github.com/demisto/demisto-sdk/pull/3514), this issue can no longer be recovered again

### DIFF
--- a/.pre-commit-config_template.yaml
+++ b/.pre-commit-config_template.yaml
@@ -69,7 +69,7 @@ repos:
     args: ["-ud", "-n", "--no-validate", "-g"]
 
   - id: validate
-    args: ["-g", "--skip-pack-dependencies"]
+    args: ["--skip-pack-dependencies", "-g"]
 
   - id: secrets
     args: ["--ignore-entropy"]


.pre-commit-config_template.yaml
```diff
@@ -89,7 +89,7 @@ repos:
   rev: v0.982
   hooks:
   - id: mypy
-    args: [--ignore-missing-imports, --check-untyped-defs, --show-error-codes, --follow-imports=silent, --allow-redefinition, --python-version=3.10]
+    args: [--ignore-missing-imports, --check-untyped-defs, --show-error-codes, --follow-imports=silent, --allow-redefinition]
```
dorschw, line 92: Careful ⚠️  does this depend on the SDK feature branch?

PR #29299, opened 2023-08-29
Devcontainer maintanence

ilaner:
The devcontainer step was broken.
Fixing the CreateCommand of the continaer, and update the features hash. Will run `demisto-sdk pre-commit` to cache the pre-commit depedencies


.devcontainer/createCommand.sh
```diff
@@ -4,9 +4,14 @@ set -e
...
-sudo chown demisto .venv
-sudo chown demisto node_modules
 sudo chown demisto /workspaces
+sudo chown demisto /workspaces/content
+sudo chown -R demisto /workspaces/content/.vscode
+sudo chown -R demisto /workspaces/content/.git
+sudo chown -R demisto /workspaces/content/.venv
+sudo chown -R demisto /workspaces/content/node_modules
+sudo chown -R demisto /workspaces/content/package-lock.json
+
```
kobymeir, line 14:
```suggestion
sudo chown -R demisto /workspaces/content
```

.devcontainer/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apt-get update && apt-get install dos2unix git python2 curl -y \
     && cd /features \
     # locking to the latest master commit in this repo to prevent breaking changes
     # We should update this commit hash from time to time to
-    && git checkout 1869e5931cfe0517f75d58cb70863a6b4874c487
+    && git checkout 96bff0097028001e6e4126c5528d37cb8c13e785
```
kobymeir, line 19: where did this magic value come from? and can it be computed instead of magic. or document we you get it from.

.devcontainer/createCommand.sh
```diff
@@ -19,5 +24,11 @@ rm -f .env
...
+echo "Setting up git safe directory"
+git config --global --add safe.directory /workspaces/content
+
 echo "Setting up content dependencies"
-.hooks/bootstrap
\ No newline at end of file
+.hooks/bootstrap
+
+echo "Run demisto-sdk pre-commit to cache dependencies"
+poetry run demisto-sdk pre-commit > /dev/null 2>&1 || true
```
kobymeir, line 34:
```suggestion
poetry run demisto-sdk pre-commit 2>&1 >/dev/null || true
```
redirects is read from left to right. so first point stderr to stdout, then point stdout to null.

why not point it to a log file for forensic if needed?

.devcontainer/devcontainer.json
```diff
@@ -104,8 +97,8 @@
 // "overrideFeatureInstallOrder": [
 // 	"ghcr.io/devcontainers/features/common-utils:1"
 // ],
-"onCreateCommand": "dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && chmod +x .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
-"postStartCommand": "poetry install",
+"onCreateCommand": "sudo dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && sudo chmod +x .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
+"postStartCommand": "poetry install && poetry run demisto-sdk pre-commit > /dev/null 2>&1 || true",
```
kobymeir, line 101: same here for bash stderr redirects

.devcontainer/devcontainer.json
```diff
@@ -104,8 +97,8 @@
 // "overrideFeatureInstallOrder": [
 // 	"ghcr.io/devcontainers/features/common-utils:1"
 // ],
-"onCreateCommand": "dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && chmod +x .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
-"postStartCommand": "poetry install",
+"onCreateCommand": "sudo dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && sudo chmod +x .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
```
kobymeir, line 100: why the last command it executed via bash if it's changed to be executed?
and why do you need the dos2unix? why not fix the file to begin with?
and why not make it executable in the git repository to begin with?

.devcontainer/devcontainer.json
```diff
@@ -1,4 +1,4 @@
-// Development container for Content. Demisto-SDK 1.14.5
+// Development container for Content
```
kobymeir, line 1:
```suggestion
// Development container for Content.
```

kobymeir (changes requested):

.devcontainer/createCommand.sh
ilaner: This heavily reduces the performance, as it take time to change the ownership of all the content repo.
Those are the directories that needed for the environment installation, which the permissions are not set for them yet.

For some reason, after the `onCreateCommand`, the `demisto` user in the container has writable access to all the workspace, just during this command the permissions are missing.

.devcontainer/Dockerfile
ilaner: This is the latest commit of this repo:
https://github.com/devcontainers/features

Basically, we shouldn't install the Dev Container features directly from this repo. The way to install features is to add this in the configuration file:
```
"features": {
    "ghcr.io/devcontainers/features/go:1": {
        "version": "1.18"
    },
    "ghcr.io/devcontainers/features/docker-in-docker:1": {
        "version": "latest",
        "moby": true
    }
```

However, GP blocks it, so our workaround is to install the features directly from the repo in the Dockerfile. We set to the latest commit as we had a case which this repo had breaking changes which we did not caught.

.devcontainer/createCommand.sh
ilaner: We don't care about the output of the pre-commit. We do it, so the pre-commit dependencies will be cached when running `demisto-sdk pre-commit` in the future.

.devcontainer/devcontainer.json
ilaner: 1. checking if chmod is reduntant
2. we had issues when creating devcontainer from Windows machines where Windows automatically changed the `.devcontainer/createCommand.sh`, and if failed when running this script from the container.
3. I will try.

ilaner: 1+3: dont need the chmod

.devcontainer/createCommand.sh
MichaelYochpaz:
```suggestion
sudo chown -R demisto /workspaces/content /workspaces/content/.vscode /workspaces/content/.git demisto /workspaces/content/.venv /workspaces/content/node_modules /workspaces/content/package-lock.json
```
`chown` accepts multiple folder / file parameters

.devcontainer/Dockerfile
kobymeir: @ilaner  why not ask GP to approve this URL?

.devcontainer/devcontainer.json
```diff
@@ -104,8 +97,8 @@
 // "overrideFeatureInstallOrder": [
 // 	"ghcr.io/devcontainers/features/common-utils:1"
 // ],
-"onCreateCommand": "dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && chmod +x .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
-"postStartCommand": "poetry install",
+"onCreateCommand": "sudo dos2unix -n .devcontainer/createCommand.sh .devcontainer/createCommand_unix.sh && bash .devcontainer/createCommand_unix.sh",
+"postStartCommand": "poetry install && poetry run demisto-sdk pre-commit >/dev/null 2>&1 || true",
```
kobymeir, line 101: why not create a postStartCommand.sh like you do in the onCreateCommand?

.devcontainer/createCommand.sh
```diff
@@ -19,5 +19,11 @@ rm -f .env
...
+echo "Setting up git safe directory"
+git config --global --add safe.directory /workspaces/content
+
 echo "Setting up content dependencies"
-.hooks/bootstrap
\ No newline at end of file
+.hooks/bootstrap
+
+echo "Run demisto-sdk pre-commit to cache dependencies"
+poetry run demisto-sdk pre-commit >/dev/null 2>&1 || true
```
kobymeir, line 29: why are you suppressing the pre-commit run result?

kobymeir (changes requested):

ilaner:
@kobymeir 
Dev Container has a problem with self signed certificate, which GP has:
https://github.com/microsoft/vscode-remote-release/issues/6092

Hopefully this issue will be resolved and we will no longer need this workaround to install features.

.devcontainer/createCommand.sh
ilaner: We don't care about the result. It will just cache the pre-commit so `git commit` will be fast in the first run.

kobymeir (approved):

PR #29433, opened 2023-09-04
Servicenow api fix

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-27758

## Description
- Fixed an issue when using the `v2` API version the mirror-in and get_attachemtns would fail, since it is not supported in this version

image


Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
```diff
@@ -687,6 +688,10 @@ def send_request(self, path: str, method: str = 'GET', body: dict | None = None,
                 'Accept': 'application/json',
                 'Content-Type': 'application/json'
             }
+        # The attachments table does not support v2 api version
+        if get_attachments:
```
ShahafBenYakir, line 692: Does this have any impact on customers who are expecting to use v2 in any way?

ShahafBenYakir (approved):
Let's add a disclaimer to the description of the command so it would also appear in the integration README explaining that this command will always use v1 as it is not supported in v2.

Packs/ServiceNow/Integrations/ServiceNowv2/ServiceNowv2.py
Shellyber: No, it shouldn't.

PR #29510, opened 2023-09-06
[collect test] Added run of TPB of API-module dependencies

bziser:
## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-7960)

## Description
Collect Test didn't collected all the tests of the integrations that using ApiModule when only ApiModule was changed.
[Example](https://code.pan.run/xsoar/content/-/jobs/32852083) for running of the updated flow for a change in MicrosoftApiModule.


Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -885,6 +885,11 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
 
             case FileType.SCRIPT | FileType.PLAYBOOK:
                 try:
+
+                    if "ApiModule" in yml.id_:
+                        logger.debug(f"Found changes in ApiModule = {yml.id_}")
```
dorschw, line 890: add to the log, what are the implications? What happens because we found changes to the ApiModule?

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -927,6 +932,22 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
...
+    def _collect_integrations_using_apimodule(self, api_module_id) -> CollectionResult | None:
+        integrations_using_apimodule = self.id_set.api_modules_to_integrations.get(api_module_id, [])
+        collection_result_of_apimodule = None
+        for integration in integrations_using_apimodule:
+            try:
+                integration_collected = self._collect_yml(integration.path)
+                if collection_result_of_apimodule:
+                    collection_result_of_apimodule += integration_collected
+                else:
+                    collection_result_of_apimodule = integration_collected
```
dorschw, line 944: instead of the if/else logic, consider using a list, appending to, and then call `.union` on it

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -927,6 +932,22 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
...
+            except Exception as e:
+                logger.info(e)
```
dorschw, line 946: add to the log - why do we skip in this case, rather than crash everything? If we tolerate it, devs may not notice it

dorschw, line 946: why `info` rather than `warning`/`error`?

dorschw (commented):
Nice, pleas add a unit test.

Tests/scripts/collect_tests/collect_tests.py
bziser: we want only to write to the log and not crash it if the tests of the related integration are not good

Packs/ApiModules/Scripts/MicrosoftApiModule/MicrosoftApiModule.py
```diff
@@ -1553,6 +1553,7 @@ def reset_auth() -> CommandResults:
     :return: Message about resetting the authorization process.
     """
     demisto.debug(f"Reset integration-context, before resetting {get_integration_context()=}")
+    demisto.debug("YAYYYY COLECTED TEST")
```
dorschw, line 1556: lol

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -927,6 +932,19 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
                 allow_incompatible_marketplace=override_support_level_compatibility,
             )
 
+    def _collect_integrations_using_apimodule(self, api_module_id) -> CollectionResult | None:
+        integrations_using_apimodule = self.id_set.api_modules_to_integrations.get(api_module_id, [])
+        collection_result_of_apimodule = []
```
dorschw, line 937:
```suggestion
        result = []
```
the function is short, so its name implies what the result is

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -927,6 +932,19 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
...
+        for integration in integrations_using_apimodule:
+            try:
+                integration_collected = self._collect_yml(integration.path)
+                collection_result_of_apimodule.append(integration_collected)
```
dorschw, line 941: do not use `_collect_yml` directly. Use `collect_single` instead.

Tests/scripts/collect_tests/id_set.py
```diff
@@ -199,6 +200,8 @@ def __init__(self, marketplace: MarketplaceVersions) -> None:
 
             self.test_playbooks = self.id_to_test_playbook.values()
             self.modeling_rules = self.path_to_modeling_rule.values()
+            self.api_modules_to_integrations = {
+                script.object_id: script.imported_by for script in scripts if "ApiModule" in script.object_id}
```
dorschw, line 204: Consider changing both here and the other place with `object_id.endswith("ApiModule")` instead of `"ApiModule" in script.object_id`

Tests/scripts/collect_tests/id_set.py
```diff
@@ -23,22 +23,22 @@ class IdSetItem(DictBased):
     See the IdSet class to see how it's parsed.
     """
 
-    def __init__(self, id_: Optional[str], dict_: dict):
+    def __init__(self, id_: str | None, dict_: dict):
```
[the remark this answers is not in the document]
bziser: Since there are so few changes, I don't think it is necessary

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -885,6 +885,11 @@ def _collect_yml(self, content_item_path: Path) -> CollectionResult | None:
 
             case FileType.SCRIPT | FileType.PLAYBOOK:
                 try:
+
+                    if "ApiModule" in yml.id_:
```
dorschw, line 889:
```suggestion
                    if yml.id_.endswith("ApiModule"):
```

Tests/scripts/collect_tests/collect_tests.py
dorschw, line 1: add a unit test case please

Tests/scripts/collect_tests/id_set.py
```diff
@@ -199,6 +200,8 @@ def __init__(self, marketplace: MarketplaceVersions) -> None:
 
             self.test_playbooks = self.id_to_test_playbook.values()
             self.modeling_rules = self.path_to_modeling_rule.values()
+            self.api_modules_to_integrations = {
+                script.object_id: script.imported_by for script in scripts if script.object_id.endswith("ApiModule")}
```
ilaner, line 204: I don't think you need the endswith condition. 
If you do it sounds like a bug in the graph

ilaner (approved):
Looks good!
See minor comment

PR #29513, opened 2023-09-06
Added 'UserVerification' as input for 'NGFW Scan' playbook

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8301

## Description
The 'NGFW Scan' playbook, which uses the 'Block IP - Generic v3' playbook as a sub-playbook, does not contain the 'UserVerification' playbook input, which affects the 'User Verification Prompt' section of the 'Block IP - Generic v3' playbook. Due to the absence of the UserVerification playbook input in the NGFW Scan playbook, the 'Block IP - Generic v3' playbook flow will always result in data collection tasks that require manual intervention by the user.

To accomplish this, the input 'UserVerification' must be added to the 'NGFW Scan' playbook. Additionally, 'Block IP - Generic v3' (task number 23) should have a value set to the input 'UserVerification' within the 'NGFW Scan' playbook.


Packs/Core/ReleaseNotes/2_0_13.md
```diff
@@ -0,0 +1,8 @@
+
+#### Playbooks
+
+##### NGFW Scan
+
+- Added the *'UserVerification'* playbook input to determine whether user verification is required before blocking malicious IP addresses using the **'Block IP - Generic v3'** sub-playbook.
+
+- Updated the configuration for the *'UserVerification'* input within the **'Block IP - Generic v3'** sub-playbook.
```
altmannyarden, line 8: From false to true? What do you mean with updated the configuration?

TalNos: The configs for the UserVerification input within Block IP - Generic v3 sub-playbook was changed. Before it was set to 'true' and now it is set to the value of UserVerification input within NGFW Scan playbook

TalNos: fixed

PR #29605, opened 2023-09-11
Remove unnecessary recommendations from extensions.json

samuelFain:
#### Remove unnecessary extension recommendations from extensions.json
List of removed extensions:

-  "ms-python.black-formatter"
- "rangav.vscode-thunder-client"
- "Gruntfuggly.todo-tree"
- "shd101wyy.markdown-preview-enhanced"
- "timonwong.shellcheck"


.devcontainer/devcontainer.json
```diff
@@ -62,14 +62,10 @@
 				"njpwerner.autodocstring",
 				"VisualStudioExptTeam.vscodeintellicode",
 				"yzhang.markdown-all-in-one",
```
ilaner, line 64: do we need this?

ilaner:
let's add to the list:
* ms-python.debugpy (better python debug experience)
* EditorConfig.EditorConfig (support the `.editorconfig` file we now have in content)
* DavidAnson.vscode-markdownlint (we already use it in sdk format, this lints and autofixes MD files.
* ms-vscode.powershell (powershell integrations...)

.devcontainer/devcontainer.json
samuelFain: Not really, I removed on extension (there were two) but the OOTB VS Code markdown support is good enough in my opinion.

ilaner (approved):
Great job

PR #29679, opened 2023-09-14
[XDR] - update the list risky users/hosts commands

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-8369)

## Description
See the issue description

## Must have
- [x] Tests
- [ ] Documentation


Packs/ApiModules/Scripts/CoreIRApiModule/CoreIRApiModule.py
```diff
@@ -3909,6 +3910,8 @@ def _warn_if_module_is_disabled(e: DemistoException) -> None:
             else:
                 raise
 
+        if arg_to_number(outputs.get("score")) == 0:  # type: ignore[union-attr]
```
dansterenson, line 3913: arg_to_number raises exception if arg is None. Is it ok?

dansterenson, line 3913: arg_to_number raises exception if arg is None. Is it ok?

dansterenson (approved):
Nice

Packs/ApiModules/Scripts/CoreIRApiModule/CoreIRApiModule.py
GuyAfik: it should because `score` doesn't exist, but maybe i should return that the `score` is not available for the user/host specified

ill update it, good catch

GuyAfik: apparently `arg_to_number` raises exception only when `required=True` and by default its `False`
I updated the implementation to be more clear, take a look

PR #29724, opened 2023-09-19
Mandiant Threat Intelligence - Use Local Filtering

chrishultin:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Use local filtering in MATI integration in order to ensure that updates to indicator information are retrieved, even if they no longer fit the "search" criteria in use. (e.g. an indicator is created as "malicious", but is later marked as "benign".  Previous versions would ignore the "benign" update)
## Must have
- [ ] Tests
- [ ] Documentation


Packs/MandiantAdvantageThreatIntelligence/Integrations/MandiantAdvantageThreatIntelligence/MandiantAdvantageThreatIntelligence.py
```diff
@@ -1048,8 +1049,22 @@ def get_new_indicators(
...
             filter(last_updated_filter(start_date), new_indicators_list)  # type: ignore
         )
+        return new_indicators_list
+    else:
+        updated_indicators = []
+        # For Indicators of Compromise only
+        for indicator in new_indicators_list:
+            # Check if indicator should be added no matter what
+            # E.g. it meets the `param` requirements
+            if indicator.get("mscore") >= minimum_mscore:
```
MichaelYochpaz, line 1059: Do we expect `mscore` is optional here?
Because if it should always be there, let's use `indicator
```suggestion
            if indicator["mscore"] >= minimum_mscore:
```

If it's an optional key, please handle such cases where it's missing.

At its current state, if this key is missing, the condition would become:
`if None >= minimum_mscore:`, which would raise an unexpected error.

Packs/MandiantAdvantageThreatIntelligence/Integrations/MandiantAdvantageThreatIntelligence/MandiantAdvantageThreatIntelligence.py
```diff
@@ -1048,8 +1049,22 @@ def get_new_indicators(
...
+                updated_indicators.append(indicator)
+            else:
+                existing_indicators = list(IndicatorsSearcher(value=indicator.get("value")))
+                if len(existing_indicators) > 0 and int(existing_indicators[0].get("total")) > 0:
```
MichaelYochpaz, line 1063: Same note here. Is `total` an optional key? Because at its current state, if it's missing, it will try to convert `None` to an `int`, resulting in an unexpected error of:
`TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

If you expect this key to always be there, please use:
```suggestion
                if len(existing_indicators) > 0 and int(existing_indicators[0]["total"]) > 0:
```

If not, handle cases where it's missing.

Packs/MandiantAdvantageThreatIntelligence/Integrations/MandiantAdvantageThreatIntelligence/MandiantAdvantageThreatIntelligence.py
```diff
@@ -1048,8 +1049,22 @@ def get_new_indicators(
...
+                    updated_indicators.append(indicator)
+        return updated_indicators
 
-    return new_indicators_list
+    return []
```
MichaelYochpaz, line 1067: This line is never used, since you have:
```python
if X:
    return something

else:
   return something else

return something  # <-- This line never runs
```

Packs/MandiantAdvantageThreatIntelligence/Integrations/MandiantAdvantageThreatIntelligence/MandiantAdvantageThreatIntelligence.py
```diff
@@ -1048,8 +1049,22 @@ def get_new_indicators(
...
+    else:
+        updated_indicators = []
+        # For Indicators of Compromise only
+        for indicator in new_indicators_list:
+            # Check if indicator should be added no matter what
+            # E.g. it meets the `param` requirements
+            if indicator.get("mscore") >= minimum_mscore:
+                updated_indicators.append(indicator)
+            else:
+                existing_indicators = list(IndicatorsSearcher(value=indicator.get("value")))
```
MichaelYochpaz, line 1062:
```suggestion
                existing_indicators = list(IndicatorsSearcher(value=indicator["value"]))
```
The `value` parameter of `IndicatorSearcher` expects an `str`, and you're passing it `str | None` when using `.get()`

MichaelYochpaz (commented):
Hey, thank you for the contribution.
Looks pretty good overall. Left a few minor notes.

chrishultin:
Thanks for the feedback! It's all been addressed, please let me know if you've got any other comments or concerns.

MichaelYochpaz (approved):
Looks good, good job :)

PR #29836, opened 2023-09-25
[TextFromHTML] Rundssoar contrib automations

rundssoar:

## Status
- [x] In Progress
- [x] Ready

## Related Issues
n/a

## Description
This Pull Request enhances the "TextFromHTML" Automation by:
- Adding an option to output the extracted Text to the Context
- For the "body" tag, allow to fall back to the full HTML for extraction. This is helpfull if the HTML may not contain a proper body tag (e.g when the HTML is an E-Mail body, that isn't formated properly). In this case, it is now possible to fall back to the whole HTML as the Body and extract the Text from that.
- Properly convert <br> Tags into "\n" characters to maintain Linebreaks in the text. To keep the text somewhat reasonable, the number of successive Linebreaks is limited to 2 (\n\n). If more linebreaks are found, they are collapsed to two linebreaks.
- Collapse multible Space and Tab Characters into a single Space
- Remove trailing whitespace from extracted Text

In Addition, the Script was adjusted to address the Warnings generated by demisto-sdk 1.20.4
Existing tests have been adjusted to fit the new logic of removing trailing whitespace

Overall the Changes were designed in a way, to maintain backwards compability. Besides the slightly changed output, generated, the Automation will keep working as before if none of the new arguments are provided.

## Must have
- [x] Tests
- [x] Documentation


Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
 
 
-def get_plain_text(html_regex):
+def get_plain_text(html: str):
     data = ''
-    if html_regex and html_regex.group(0):
-        data = re.sub(r'<.*?>', '', html_regex.group(0))
+    if html:
+        data = re.sub(r'<br>', '\n', html, flags=re.M + re.S + re.I + re.U)
+        data = re.sub(r'<.*?>', '', html, flags=re.M + re.S + re.I + re.U)
```
MichaelYochpaz, line 12: All these RegEx flags seem unneeded here.

Also, I know this isn't your regex but a previous one, but the `?` there is redundant, since `*` also captures cases where the pattern appears 0 times.

```suggestion
        data = re.sub(r'<.*>', '', html)
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
         entities = {'quot': '"', 'amp': '&', 'apos': "'", 'lt': '<', 'gt': '>', 'nbsp': ' ',
                     'copy': '(C)', 'reg': '(R)', 'tilde': '~', 'ldquo': '"', 'rdquo': '"', 'hellip': '...'}
         for e in entities:
             data = data.replace(f'&{e};', entities[e])
+        data = re.sub(r'[ \t]{2,}', ' ', data)
+        data = re.sub(r'(\r?\n){3,}', '\n\n', data)
+        data = re.sub(r'\s+$', '', data)
```
MichaelYochpaz, line 19: These also change the current behavior of the script and break backwards-compatibility.
I'm also not sure if it's a good idea to "cleanup" the text, since some users might want to see it in the original format it was in untouched, just extracted from the HTML.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
+
+        if body and body.group(0):
+            return body.group(0)
+        elif allow_fallback and html_tag.lower() == 'body':
+            return html
+    return ''
+
+
+def text_from_html(args: dict):
+    html = args.get('html', '')
```
MichaelYochpaz, line 35: Please revert this. "html" is a required parameter that should always be assigned.
We want it to fail if it doesn't exist for some reason (and not assign it an empty string value)/

```suggestion
    html = args['html']
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
     html_tag = args.get('html_tag', 'body')
+    allow_fallback = str(args.get('allow_body_fallback', 'false')).lower() == 'true'
 
-    body = re.search(fr'<{html_tag}.*/{html_tag}>', html, re.M + re.S + re.I + re.U)
+    body = get_body(html, html_tag, allow_fallback)
     data = get_plain_text(body)
+
     return data if data != '' else 'Could not extract text'
```
MichaelYochpaz, line 42: This function was kind of the `main` function before, no need to have both since it just adds unneeded complexity.
Please move this function's logic inside the new `main` function.

MichaelYochpaz (changes requested):
Hey @rundssoar, thank you for your contribution :)
I've left a few notes for you to address. If you need any help, feel free to contact me on the DFIR Slack channel.

Also note that you're missing a releaes-notes and version bump for the pack.
You can generate that using `demisto-sdk update-release-notes -i Packs/CommonScripts -u revision`

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: This only was changed to address the demisto-sdk linter warning: 
W9019 text_from_html: It is best practice to use .get when accessing the arg/params dict object rather then direct access.
I'll revert the change

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: I've combined them into a main function

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: This is now an Optional feature

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: I've reduced the flags, but i kept the re.I flag, as i've seen cases, where HTML contained Upper case Tags, and the regex without the flag just wouldn't replace them.
If necessary, this can be stated in the "breaking changes" section, though i'd rather consider this a fix, as for me this was the inteded behavious from the beginning

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
 import re
 
 
-def get_plain_text(html_regex):
+def get_plain_text(html: str):
     data = ''
-    if html_regex and html_regex.group(0):
-        data = re.sub(r'<.*?>', '', html_regex.group(0))
+    if html:
+        data = re.sub(r'<br>', '\n', html, flags=re.M + re.S + re.I + re.U)
```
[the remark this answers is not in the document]
rundssoar: I've made this an optional Feature, and updated the regex to actually consider all the variants in the referenced article.
I've also reduce the flags, but kept the re.I flag, as i've seen HTML with uppercase characters, and the regex wouldn't work othewise.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
     return data if data != '' else 'Could not extract text'
 
 
+def main():
+    try:
+        text = text_from_html(demisto.args())
+        context_path = demisto.args().get('context_path', None)
+        result = CommandResults(
+            outputs_prefix=context_path,
+            outputs=text if context_path else None,
```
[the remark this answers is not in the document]
rundssoar: I've made it a constant "TextFromHTML" context key, but kept it optional if the context key is created or not.
We could make it always the case, but for the sake of backwords-compability and to avoid blowing up the context of existing playbooks, if the feature isn't required, i made it optional.
 For people just using this in the warroom, putting it to the context may not be what they want.
 Not sure, what the best practise here would be.
 Would it be better to always have the context key, but only populate it if needed?

rundssoar:
I've made the first changes as requested.
I'll finish the Readme / Versioning, etc. once the technical details are done, so i don't have to do it repeatedly.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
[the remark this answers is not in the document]
rundssoar: Should be done in [5074e5c](https://github.com/demisto/content/pull/29836/commits/5074e5c36904a21cdd35c2229c6ffae7c7440a0f)

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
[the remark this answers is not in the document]
rundssoar: Good Catch,
fixed in 8cc6040

rundssoar: In Regards to the `?` in the tag regex, it is actually necessary here, because in this case, it tells the `*` to be "lazy"  (see https://www.regular-expressions.info/repeat.html).
If we remove it, it actually breaks the script.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,57 @@
...
+            readable_output=text
+        )
+
+        return_results(result)
+    except Exception as ex:
+        return_error(message="failed to extract text", error=ex)
```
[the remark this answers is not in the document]
rundssoar: Fixed in in [5074e5c](https://github.com/demisto/content/pull/29836/commits/5074e5c36904a21cdd35c2229c6ffae7c7440a0f)

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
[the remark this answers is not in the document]
rundssoar: @MichaelYochpaz Can you let me know, if this is OK now, or what you would recommend?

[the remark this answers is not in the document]
MichaelYochpaz: Adding additional stuff to context is not considered breaking backwards-compitability (removing / changing keys does), but if you think there could be cases where users might not want to have it added to context, adding an optional parameter to determine whether to add it to context sounds fine to me.
Up to you.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
MichaelYochpaz: I see now what you're saying about the "lazy" issue there.
As for the case-insensitive flag, though, I still don't see why it's needed.
`.` captures all ASCII characters, both upper and lower case.

It might be relevant for the `<br />` RegEx if you want to capture `<BR />` for example too, but shouldn't add anything here.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
MichaelYochpaz: This warning is incorrect (it's not "best practice"), I raised an internal issue to remove it.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: You're right, the `.` captures them all.
Not sure, what it was i rember being an issue. I'm reverting that flag there.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
```diff
@@ -13,14 +13,52 @@ args:
...
   default: true
   defaultValue: body
   description: Specify HTML tag to extract the text from within.
+- name: allow_body_fallback
+  default: true
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: 'Allow using the input HTML as a fallback for the Body, if no body-tag is found (default: false).
```
MichaelYochpaz, line 27:
```suggestion
  description: 'Allow using the input HTML as a fallback for the Body, if no body-tag is found.
```

We don't write default values in the description by convention. This is already added automatically in the UI (which determines the default type from the `default` key).

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
```diff
@@ -13,14 +13,52 @@ args:
...
+    This only applies, if html_tag is set to body.'
+  defaultValue: "false"
+- name: replace_line_breaks
+  default: true
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: 'Replace `br` in `html` with linebreaks in the output  (default: false).'
```
MichaelYochpaz, line 36:
```suggestion
  description: 'Replace `br` in `html` with linebreaks in the output.'
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
```diff
@@ -13,14 +13,52 @@ args:
...
+  defaultValue: "false"
+- name: trim_result
+  default: true
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: 'Trim the extracted result (default: false).
```
MichaelYochpaz, line 44:
```suggestion
  description: 'Trim the extracted result.
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
```diff
@@ -13,14 +13,52 @@ args:
...
+     When set to true, leading and trailing whitespaces are removed and blocks of more than 3 consecutive whitespaces are collapsed to two.'
+  defaultValue: "false"
+- name: output_to_context
+  description: Store the converted Text in the Context.
```
MichaelYochpaz, line 48:
```suggestion
  description: Store the extracted text in context.
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,59 @@
...
+        body = re.search(fr'<{html_tag}.*/{html_tag}>', html, re.M + re.S + re.I + re.U)
+
+        if body and body.group(0):
+            return body.group(0)
+        elif allow_fallback and html_tag.lower() == 'body':
+            return html
+    return ''
+
+
+def main():
```
MichaelYochpaz, line 37:
```suggestion
def main():  # pragme: no cover
```
Let's add this to exclude `main` from unit-tests coverage since it's non-relevant for unit-tests, and the coverage dropped (probably due to merging the functions) as can be seen here ("Generate coverage reports"):
https://app.circleci.com/pipelines/github/demisto/content/306249/workflows/08600a31-4e0b-416f-b799-fafe00961e7a/jobs/721876

MichaelYochpaz (commented):
Good job (and great unit-tests btw) :)
I've added a few additional minor notes.

Please re-run the `demisto-sdk generate-docs` command after applying the changes to the YAML file.

Note that some unit-tests are failing:
https://app.circleci.com/pipelines/github/demisto/content/306249/workflows/08600a31-4e0b-416f-b799-fafe00961e7a/jobs/721876

Also, the [validations step](https://app.circleci.com/pipelines/github/demisto/content/306249/workflows/08600a31-4e0b-416f-b799-fafe00961e7a/jobs/721878) fails because of an issue we had on master. Please pull from master to your PR and then it should work properly

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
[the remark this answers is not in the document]
rundssoar: I'm sticking with making it optional.

rundssoar:
Hi,
I've added the ReadMe Update.
Unfortunatelly, adding the command example keeps failing with "The provided example for cmd ... has failed, did not get any example for None. please add it manually."
If you have any ideay, why, please let me know
i've copied the existing example to the updated Readme.

Packs/CommonScripts/Scripts/TextFromHTML/README.md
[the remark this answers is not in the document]
rundssoar: done

Packs/CommonScripts/ReleaseNotes/1_12_30.md
```diff
@@ -0,0 +1,10 @@
+
+#### Scripts
+
+##### TextFromHTML
+
+- Added the *allow_body_fallback* argument, which allows to use the full *html* as the body in case the input *html* does not have a `body` tag
+- Added the *replace_line_breaks* argument, which allows to replace `br` tags in the html with line breaks in the extracted text
+- Added the *trim_result* argument, which allows to replace leading and trailing whitespaces as well as collapsing multible empty lines in the extracted text to a single empty line
+- Added the *output_to_context* argument, which allows to store the extracted text in context
```
ShirleyDenkberg, line 9:
```suggestion
- Added the *allow_body_fallback* argument, which allows using the full *html* as the body in case the input *html* does not have a `body` tag.
- Added the *replace_line_breaks* argument, which allows replacing `br` tags in the html with line breaks in the extracted text.
- Added the *trim_result* argument, which allows replacing leading and trailing whitespaces as well as collapsing multiple empty lines in the extracted text to a single empty line.
- Added the *output_to_context* argument, which allows storing the extracted text in context.
```

Packs/CommonScripts/Scripts/TextFromHTML/README.md
```diff
@@ -9,26 +10,44 @@ Extract regular text from the given HTML
...
 ## Inputs
+
 ---
 
 | **Argument Name** | **Description** |
 | --- | --- |
-| html | The HTML to strip tags from |
+| html | The HTML to strip tags from. |
+| html_tag | Specify HTML tag to extract the text from within. |
+| allow_body_fallback | Allow using the input HTML as a fallback for the Body, if no body-tag is found. This only applies, if html_tag is set to body. |
```
ShirleyDenkberg, line 28:
```suggestion
| html_tag | Specify the HTML tag to extract the text from within. |
| allow_body_fallback | Allow using the input HTML as a fallback for the body, if no body tag is found. This only applies, if html_tag is set to body. |
```

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.yml
```diff
@@ -7,23 +7,59 @@ type: python
...
   default: true
   defaultValue: body
   description: Specify HTML tag to extract the text from within.
+- name: allow_body_fallback
+  default: true
+  auto: PREDEFINED
+  predefined:
+  - "true"
+  - "false"
+  description: 'Allow using the input HTML as a fallback for the Body, if no body-tag is found. This only applies, if html_tag is set to body.'
```
ShirleyDenkberg, line 27:
```suggestion
  description: 'Allow using the input HTML as a fallback for the body, if no body tag is found. This only applies, if html_tag is set to body.'
```

ShirleyDenkberg:
@MichaelYochpaz Doc review completed.

Packs/CommonScripts/Scripts/TextFromHTML/README.md
```diff
@@ -9,26 +10,44 @@ Extract regular text from the given HTML
 | Tags | Utility |
 | Cortex XSOAR Version | 5.0.0 |
 
+## Used In
+
+---
+This script is used in the following playbooks and scripts.
+
+* TextFromHTML_test_playbook
+
```
MichaelYochpaz, line 19: Remove this. Test playbooks are just used internally (like unit-tests) and aren't shipped / used by users.
```suggestion
```

MichaelYochpaz (commented):
@rundssoar I think this is because you named the file `command_examples` instead of `command_examples.txt`.
Other than that and the small note I added, PR looks good. Good job! :)

Just handle those and Shirley's doc-review notes, and it should be good to go for a merge.

Packs/CommonScripts/Scripts/TextFromHTML/README.md
rundssoar: Wasn't intentional, generate-docs automatically seems to add this.

MichaelYochpaz: That's odd, the SDK should skip those. I'll open an internal ticket for fixing that.

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
```diff
@@ -5,26 +5,59 @@
...
+        data = re.sub(r'<\/?br\s?\/?>', '\n', html, flags=re.I) if replace_line_breaks else html
+
+        data = re.sub(r'<.*?>', '', data)
         entities = {'quot': '"', 'amp': '&', 'apos': "'", 'lt': '<', 'gt': '>', 'nbsp': ' ',
                     'copy': '(C)', 'reg': '(R)', 'tilde': '~', 'ldquo': '"', 'rdquo': '"', 'hellip': '...'}
         for e in entities:
             data = data.replace(f'&{e};', entities[e])
+
+        if trim_result:
+            data = re.sub(r'[ \t]{2,}', ' ', data)
```
MichaelYochpaz, line 20:
```suggestion
            data = re.sub(r'\s{2,}', ' ', data)
```

MichaelYochpaz (approved):
Hey, PR looks good besides this minor RegEx suggestion.
I sent you an email regarding the demo.

Good job :)

Packs/CommonScripts/Scripts/TextFromHTML/TextFromHTML.py
rundssoar: Making this change would result in line 21 being useless, as \s also matches Linebreaks.
The Ideas of Line 20 is: Remove occurences of two or more consecutive space or tab characters from the extracted text, whereas line 21 intends to avoid large blocks of "empty" lines.
changing the regex to \s, would also match strings like "some line\n\nwith a empty line between", which isn't intended.
I would be happy to merge them together, but for line 21, having 3 or more is intentional, to allow for single empty lines, whereas for tabs or space, in most cases, having two or more is already something i wanted to avoid.

PR #29863, opened 2023-09-26
CortexXDRIR Custom Resolution Status Mapping

mohlcyber:
## Status
- [x] In Progress
- [ ] Ready

## Description
Proposal to provide the flexibility with custom resolution status mappings. We receive more requests to provide more flexible mechanism to allow custom mappings. Currently we do hard code the mappings.

## Must have
- [ ] Tests
- [x] Documentation


MichaelYochpaz:
Closing this following a Slack conversation with @mohlcyber and @altmannyarden, as this should be implemented as a custom mapper (that we need to add support for) and not using custom lists.

PR #30073, opened 2023-10-05
[content-nightly] - install packs from artifacts + use Packs folder from latest upload commit

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-7562)

## Nightly
[nightly](https://code.pan.run/xsoar/content/-/pipelines/6755434)

## Description
1. Install all the packs from the `pack_ids_to_install.txt` artifacts that `collect_test.py` creates which contains all the necessary packs that needs to be installed for the nightly.
2. run the **content-nightly** `Packs` folder from the latest successful upload commit in order to test our production buckets properly.  
3. Fixes an issue where the `packs_to_install.txt` artifact collected also `hidden` packs which are not needed.
4. no need to pass `--post-commit` when using `validate -a`.

## Must have
- [x] Tests
- [ ] Documentation


.gitlab/ci/.gitlab-ci.global.yml
```diff
@@ -210,6 +210,31 @@
   - set | grep -E "^CI_.*=|^GITLAB.*=" | sort
   - section_end "Gitlab CI Build Parameters"
 
+.checkout-upload-commit-content-nightly: &checkout-upload-commit-content-nightly
+  - section_start "Checkout upload commit content nightly"
+  - | 
+    if [[ -n "${NIGHTLY}" && "${CI_COMMIT_BRANCH}" == "install_packs_nightly_bugs" ]]; then
```
GuyAfik, line 216:
```suggestion
    if [[ -n "${NIGHTLY}" && "${CI_COMMIT_BRANCH}" == "master" ]]; then
```

after finishing up the testing it should be `master`

.gitlab/ci/.gitlab-ci.global.yml
```diff
@@ -210,6 +210,31 @@
...
+      echo "content nightly: checking out to the last successful upload commit $LAST_UPLOAD_COMMIT for the Packs folder"
+      git status
+      git reset $LAST_UPLOAD_COMMIT -- ./Packs
+      git checkout -- ./Packs
```
JudahSchwartz, line 220: why not `git checkout $LAST_UPLOAD_COMMIT -- ./Packs` ? whats reset doing here

GuyAfik: see the last comment here: https://stackoverflow.com/questions/6119036/how-to-revert-a-folder-to-a-particular-commit-by-creating-a-patch

GuyAfik: You can use git reset to reset the index which will also include removing files that were added in more recent commits (git checkout on it's own doesn't do this):

JudahSchwartz: lets add comments explaining each line and why its neccessary

GuyAfik: done

Tests/scripts/validate.sh
```diff
@@ -1,16 +1,22 @@
...
         for item in $PACKS_TO_UPLOAD_SPACED; do
             python3 -m demisto_sdk validate -i Packs/"$item" --post-commit --graph --skip-pack-dependencies
         done       
-    else 
-        python3 -m demisto_sdk validate -a --post-commit --graph --skip-pack-dependencies
+    else
+        if [ -n "$NIGHTLY" ]; then
+            PREV_VER=$LAST_UPLOAD_COMMIT
+        else
+            PREV_VER="origin/master"
```
daryakoval, line 17: Instead of adding this variable here, you can add it as a variable to the gitlab.yml file. There is an option to define different variables under different conditions(rules). For example: https://docs.gitlab.com/ee/ci/yaml/#rulesvariables
We already using different variables under different rules, look in the on-push file :).

yucohen:
Looks good from my side, great work!

Tests/scripts/validate.sh
GuyAfik: defined https://github.com/demisto/content/blob/master/.gitlab/ci/.gitlab-ci.global.yml#L184,
is it considered to be defined globally? i am not sure.

daryakoval: Instead of defined globally, I should write - predefined when the build starts. 
Sorry for that. 
It is ok you can implement my suggestion because of the way it is defined. 
You can ignore this comment. thanks.

JudahSchwartz (approved):
awesomesauce

yucohen (approved):

PR #30158, opened 2023-10-13
Revert "Add gitlab mirror self trigger step"

michal-dagan:
Reverts demisto/content#29992


yucohen (approved):

michal-dagan:
A force merge is needed as run-validation failed because a CI file that was added to the previous PR was reverted.

PR #30220, opened 2023-10-17
[Prisma Cloud Compute] Commands Arguments Description Update

BEAdi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8638

## Description
~Change errors to info in cases where the argument provided doesn't exist.~
~Change hostname argument to required in cases where an error returned if it is not provided. This is a BC, now it will be caught before running the command instead of getting an error after running it.~
Update command arguments descriptions in order to clarify where hostname argument values should be taken from.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2564,22 +2600,6 @@ def main():
...
-            return_results(results=get_container_hosts_list(client=client, args=demisto.args()))
-        elif requested_command == 'prisma-cloud-compute-profile-container-forensic-list':
-            return_results(results=get_profile_container_forensic_list(client=client, args=demisto.args()))
-        elif requested_command == 'prisma-cloud-compute-host-forensic-list':
-            return_results(results=get_profile_host_forensic_list(client=client, args=demisto.args()))
-        elif requested_command == 'prisma-cloud-compute-custom-feeds-ip-add':
-            return_results(results=add_custom_ip_feeds(client=client, args=demisto.args()))
-        elif requested_command == 'prisma-cloud-compute-console-version-info':
-            return_results(results=get_console_version(client=client))
-        elif requested_command == 'prisma-cloud-compute-custom-feeds-ip-list':
```
yuvalbenshalom, line 2581: how is this deletion related to the PR?

yuvalbenshalom (commented):
see my comment.
there is one deletion I don't understand, and also I am not sure the original bug and fix expectations are correct.
I written it the ticket.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
BEAdi: These lines are duplicate with the lines above it, I assume it happened by mistake so I fixed it here. As I've worked on those commands I noticed it and removed these lines.

yuvalbenshalom (approved):

PR #30335, opened 2023-10-22
FindSimilar logs

tkatzir:
Add logs to various FindSimilar scripts.

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue


anara123 (commented):
Add a log when the script starts and one before we call demisto.executeCommand, and maybe even print their args

anara123 (approved):

PR #30361, opened 2023-10-23
Made premium packs free. Added new alert states to the Code42 Check I…

jdullum:
…ncydr Status And Close XSOAR Incident playbook. Deprecated commands.

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
The main goal of this PR is to move the playbooks from the Code42 Insider Threat Remediation premium pack to the Code42 free pack. It also fixes a minor bug with the "Check Incydr Status and Close XSOAR Incident" playbook, and deprecates commands that are no longer supported. 

## Must have
- [x] Tests
- [x] Documentation


Packs/Code42/Integrations/Code42/Code42.py
ilappe, line 1: our policy is to remaining the functionality of a deprecated commands as is
so please return back the deprecated commands functionality

Packs/Code42/Integrations/Code42/Code42_test.py
ilappe, line 1: as mentioned above, we will need back the test for the deprecated commands 
as the implementation of them should remain as is.

ilappe (changes requested):
Thanks @jdullum for your great effort in this contribution!
i reviewed the code related files, please see my few comments 
in addition please note the [validation error](https://app.circleci.com/pipelines/github/demisto/content/311646/workflows/a5dac0ef-a72b-4970-ad2d-3abc3df52a3b/jobs/725761?invite=true#step-119-10549_92) 

@melamedbn will review the playbooks etc.

Packs/Code42/Integrations/Code42/Code42.py
jdullum: Hi @ilappe, these commands no longer work because the backend APIs have been removed. We followed our API deprecation process, which included notifying our customers, so no one should be using them anymore. It looks like we forgot to mark these commands as deprecated in XSOAR, but we did previously update our playbooks to stop using them. I also can't simply leave the code as-is, because the functionality has been removed from py42 too. While we would prefer to remove the commands altogether since they are broken and no longer supported by our product, doing so caused build errors. If you have another suggestion for us to do in cases like this, please let me know. I did mark the version as a breaking change, and documented it in the release notes.

ilappe: @jdullum 
so if this are the case, i think it will be helpful for existing user to know about the alternative command within the war room
so can you please, in the readable_output of the deprecated command, to print `Deprecated. please use the <the alternative command> command` ?

jdullum: Sure!

ilappe (approved):
Thanks @jdullum, great work!
@melamedbn, did you finish with the review?

melamedbn (approved):

ilappe:
pre commit fail due to change in gitlab files, will be fixed in the internal PR

PR #30459, opened 2023-10-26
fixing contrib rule for docker builds

kobymeir:
## Status
- [ ] In Progress
- [x] Ready

## Description
fixing contrib rule for docker builds


kobymeir:
force merge: fixing contrib rule for docker builds

PR #30471, opened 2023-10-26
[The Hive] Use new endpoint in fetch incidents

ilaner:
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-29349

* This will use the `query` endpoint to fetch `TheHive` cases
* Add support for lookback


Packs/TheHiveProject/Integrations/TheHiveProject/TheHiveProject_test.py
```diff
@@ -523,3 +523,26 @@ def test_update_observable_command(requests_mock):
     assert "Updated Observable" in response.readable_output
     assert response.outputs_prefix == 'TheHive.Observables'
     assert response.outputs_key_field == 'id'
+
+
+def test_migrate_last_run(mocker, requests_mock):
```
yuvalbenshalom, line 528: please add doc string.
also sold be test when the migration gets an updated context

yuvalbenshalom (changes requested):
see my comments.
please bump version and add RN.
also there are multiple recommit errors. what is your opinion there?

yuvalbenshalom (approved):

PR #30501, opened 2023-10-29
NextToken in CommandResults

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Add support for returning a `NextToken` received from a pagination API endpoint to the context using the `CommandResults` class.

Why this is needed:
1. Next_tokens have the unique attribute that they need to be overridden on every run. CommandResults does not support this use case so the DT must be manually created by each individual developer. 
2. Because of the previous point, the regular outputs returned by the command must also have a custom DT created because the args `outputs_key_field` and 'outputs_prefix` would interfere with the next_token's DT.
3. There is no standard, well-known way to achieve the override-on-every-run feature which creates non-uniform paths to next_tokens across the repo (such as `Module.PrefixNextToken.NextToken` or `Module.Prefix.NextToken`, none of which match the standard `Module.PrefixNextToken` used by the TPMs)

## Must have
- [ ] Tests
- [ ] Documentation


ShahafBenYakir:
@JudahSchwartz @jlevypaloalto I am missing some background, why is this change needed? Is there a ticket? Did someone ask for this?

jlevypaloalto:
@ShahafBenYakir I've added the need in the Description. I've personally had the need on more than one occasion and I know that others have too.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6807,6 +6807,11 @@ class CommandResults:
     :type execution_metrics: ``ExecutionMetrics``
     :param execution_metrics: contains metric data about a command's execution
 
+    :type replace_existing: ``bool``
+    :param replace_existing: Replace the context value at outputs_prefix if it exists.
```
JudahSchwartz, line 6811: give the nexttoken example here

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -2613,6 +2613,46 @@ def test_empty_readable_outputs(self):
...
+        Then:
+        - Raise an errror.
+        """
+        from CommonServerPython import CommandResults
+        res = CommandResults(
+            outputs="next_token",
+            outputs_prefix="PathToValue",
+            replace_existing=True
+        )
+        with pytest.raises(DemistoException, match='outputs_prefix must be a nested path to replace an existing key.'):
```
JudahSchwartz, line 2653: need to assert false after tocontext, otherwise even if exception doesnt happen test will pass

JudahSchwartz (approved):
nice work!

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
jlevypaloalto: I got this message when I added a dot to the path so that an error was not raised:
`Failed: DID NOT RAISE <class 'CommonServerPython.DemistoException'>`

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
jlevypaloalto: done

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -6807,6 +6807,11 @@ class CommandResults:
...
+            Works only if outputs_prefix is a path to a nested value i.e., contains a period.
+            Used primarily for next tokens.
```
JudahSchwartz, line 6813:
```suggestion
            For example, the next token should always be overwritten. This response can be returned as follows: ...
```

JudahSchwartz (approved):
nice!

PR #30531, opened 2023-10-30
Fix truncated data from PAN OS

anas-yousef:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-29603
 
## Description
Fix truncated data that is returned from the API by supplying a query parameter **show-detail**, and when set to **yes**, it will return the full data, and not just a preview of them.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13962,6 +13967,10 @@ def main():  # pragma: no cover
         command = demisto.command()
         LOG(f'Command being called is: {command}')
 
+        # Log the API version
+        demisto.debug('Retrieving the API version')
+        demisto.debug(f'API version: {get_pan_os_version()}')
```
jlevypaloalto, line 13972: Great idea. Maybe we should add this condition so as not to burden regular runs.   
```suggestion
        if is_debug_mode():
            demisto.debug('Retrieving the API version')
            demisto.debug(f'API version: {get_pan_os_version()}')
```

anas-yousef: Agree

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -13962,6 +13967,11 @@ def main():  # pragma: no cover
         command = demisto.command()
         LOG(f'Command being called is: {command}')
 
+        # Log the API version
+        if is_debug_mode():
+            demisto.debug('Retrieving the API version')
+            demisto.debug(f'API version: {get_pan_os_version()}')
```
GuyAfik, line 13973:
```suggestion
            demisto.debug(f'PAN-OS Version (debug-mode): {get_pan_os_version()}')
```
I would put it under the `test-module` because assuming that debug-mode would be test on `test-module` this will be executed before the `test-module`

anas-yousef:
@JasBeilin
No need to add show-detail to pan-os, only the command where we get the job ID, show that in command example in README.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5290,7 +5290,7 @@ def build_logs_query(address_src: Optional[str], address_dst: Optional[str], ip_
 @logger
 def panorama_query_logs(log_type: str, number_of_logs: str, query: str, address_src: str, address_dst: str, ip_: str,
                         zone_src: str, zone_dst: str, time_generated: str, time_generated_after: str, action: str,
-                        port_dst: str, rule: str, url: str, filedigest: str):
+                        port_dst: str, rule: str, url: str, filedigest: str, show_detail: str):
```
JasBeilin, line 5293:
```suggestion
                        port_dst: str, rule: str, url: str, filedigest: str, show_detail: str = False):
```

Lets make sure we do not affect other commands that might be using this function.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5350,7 +5353,8 @@ def panorama_query_logs_command(args: dict):
     job_id = args.get('query_log_job_id')
     illegal_chars = {'@', '#'}
     ignored_keys = {'entry'}
-
+    show_detail = args.get('show-detail', 'no')
```
JasBeilin, line 5356: Why not use the arg-to-boolean function in CSP?

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -6001,6 +6001,13 @@ script:
     - description: whether to hide the polling result (automatically filled by polling).
       name: hide_polling_output
       deprecated: true
+    - description: Show the data of the fields `after-change-preview`, and `before-change-preview`, or keep them truncated. The full data are under the fields `after-change-detail`, and `before-change-detail`. If this argument was set to `no` when retrieving a job ID, then the data of the fields of that job ID will be truncated, and not if set to `yes`.
```
JasBeilin, line 6004:
```suggestion
    - description: Whether to show only `after-change-preview`, and `before-change-preview`, or get full data for it.  The full data are under the fields `after-change-detail`, and `before-change-detail`.
```

Packs/PAN-OS/Integrations/Panorama/README.md
```diff
@@ -3294,7 +3294,8 @@ The query logs in Panorama.
...
 | timeout           | The timeout (in seconds) when polling. Default is 120.                                                                                                            | Optional | 
-| interval_in_seconds | The interval (in seconds) when polling. Default is 10.                                                                                                            | Optional | 
+| interval_in_seconds | The interval (in seconds) when polling. Default is 10.                                                                                                            | Optional |
+| show-detail       | Show the data of the fields `after-change-preview`, and `before-change-preview`, or keep them truncated. The full data are under the fields `after-change-detail`, and `before-change-detail`. If this argument was set to `no` when retrieving a job ID, then the data of the fields of that job ID will be truncated, and not if set to `yes`. Possible values are: yes, no. Default is no.                                                                                                             | Optional |
```
JasBeilin, line 3298: to be changed according to yml.

Packs/PAN-OS/ReleaseNotes/2_1_12.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Added the argument **show-detail** to the commands **pan-os**, and **pan-os-query-logs**, which gives the ability to return the full data of the fields *after-change-preview*, and *before-change-preview*, under the fields *after-change-detail*, and *before-change-detail* in the raw response.
```
JasBeilin, line 6: I can see it was only added in 1 place.

```suggestion
- Added the argument **show-detail** to the commands **pan-os**, and **pan-os-query-logs**, to retrieve the full data of the fields *after-change-preview*, and *before-change-preview*, under the fields *after-change-detail*, and *before-change-detail* in the raw response.
```

JasBeilin (changes requested):
Great, please check my comments.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
anas-yousef: Because the API accepts either 'yes' or 'no', and not bool values

Packs/PAN-OS/ReleaseNotes/2_1_12.md
anas-yousef: True, only to the command pan-os-query-logs. Fixed it

Packs/PAN-OS/Integrations/Panorama/Panorama.py
JasBeilin: So you need to handle the case where you receive empty value, cause it will fail in the current code.

JasBeilin (changes requested):
Good, just handle the new arg correctly and we can move forward.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5454,6 +5457,9 @@ def panorama_query_logs_command(args: dict):
     job_id = args.get('query_log_job_id')
     illegal_chars = {'@', '#'}
     ignored_keys = {'entry'}
+    # We added the `or` operator in case the show-detail argument has an empty string, which can happen when
+    # updating an integration
```
JasBeilin, line 5461:
```suggestion

```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5454,6 +5457,9 @@ def panorama_query_logs_command(args: dict):
...
+    show_detail = args.get('show-detail') or 'no'
```
JasBeilin, line 5462:
```suggestion
    show_detail = args.get('show-detail', 'no') or 'no'
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -5452,6 +5455,8 @@ def panorama_query_logs_command(args: dict):
     job_id = args.get('query_log_job_id')
     illegal_chars = {'@', '#'}
     ignored_keys = {'entry'}
+    # Any value other than 'yes' will act as 'no'
```
JasBeilin, line 5458:
```suggestion
    # The API refers to any value other than 'yes' as 'no'.
```

JasBeilin (approved):

PR #30626, opened 2023-11-02
SentinelOne v 3 2 12

munna-metron:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Must have
- [ ] Tests
- [ ] Documentation


Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -3434,8 +3433,8 @@ def main():
     fetch_type = params.get('fetch_type', 'Threats')
     first_fetch_time = params.get('fetch_time', '3 days')
     fetch_severity = params.get('fetch_severity', [])
-    fetch_incidentStatus = params.get('fetch_incidentStatus', [])
-    fetch_threat_incident_statuses = params.get('fetch_threat_incident_statuses', [])
+    fetch_incidentStatus = params.get('fetch_incidentStatus', ["UNRESOLVED"])
+    fetch_threat_incident_statuses = params.get('fetch_threat_incident_statuses', ["UNRESOLVED"])
```
MosheEichler, line 3437: Is `fetch_incidentStatus` (if there is a value) not a string?
Why do you give a list by default?
I think you can use the [argToList](https://github.com/demisto/content/blob/master/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py#L1824) function in such a case

Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
```diff
@@ -3302,13 +3302,12 @@ def fetch_threats(client: Client, args):
...
     current_fetch = args.get('current_fetch')
     incident_statuses = args.get('fetch_threat_incident_statuses')
-    resolved = 'true' if incident_statuses and 'RESOLVED' in incident_statuses else 'false'
 
     threats = client.get_threats_request(limit=args.get('fetch_limit'),
                                          created_after=args.get('last_fetch_date_string'),
                                          site_ids=args.get('fetch_site_ids'),
                                          incident_statuses=','.join(incident_statuses).lower() if incident_statuses else None,
-                                         resolved=resolved)
+                                         include_resolved_param=False)
```
MosheEichler, line 3310: Can you please explain what is the reason for this change?

MosheEichler (commented):
Hi @contributor_user,
Good work :)

Plesae see my comments,
@melamedbn Please take a look on the Classifier changes.

Thanks again

Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
munna-metron: If there is a value in these fields of config`fetch_incidentStatus` and `fetch_threat_incident_statuses` it should be a list. Provided default value if we didn't found the fields in the params, in order to avoid breakage of the code flow.
Example: If we select the IN_PROGRESS, UNRESOLVED and RESOLVED while configuring the value of the above fields  will be ['IN_PROGRESS', 'RESOLVED', 'UNRESOLVED'].

Packs/SentinelOne/Integrations/SentinelOne-V2/SentinelOne-V2.py
munna-metron: Required behaviour would be, Content pack should fetch the selected incident statues of the threat provide in the config. If we provide the resolve parameter that going to override the incident_statuses in the get threats api call to server. But that is not a required behaviour this a bug we found recently in the content pack so we fixed it.

MosheEichler (approved):
@melamedbn Please review the Classifier changes.

melamedbn:
Hi @munna-metron,

I couldn't understand how the changes you've made to the classifier effect the issue with using incident type with the phrase 'incident' in its name.
Please explain why you mapped both SentinelOne Incident and Incident, it doesn't seem like the right implementation.

We can schedule a short call if necessary.

Best regards,
Ben

munna-metron:
Hi @melamedbn , 
I will try to explain the issue, what happened is when we created the mirroring feature  for Sentinelone. some how we mapped some of the incidents to `incident` incident type, but our classifier only classifies the events which had the `SentineloOne Incident` as incident type. And we fixed this issue in code in V3.2.10 and corrected it with `SentinelOne Incident ` so that our classifier identifies the incident. But some of the customers are still seeing some unclassified incidents in their Xsoar, the reason is they had some incident which pulled before 3.2.10 version. So in-order to this backward compatible we mapped this "incident" to our classifier. I hope this make sense.
if you think we need a call we can have it
Regards

melamedbn:
> Hi @melamedbn , I will try to explain the issue, what happened is when we created the mirroring feature for Sentinelone. some how we mapped some of the incidents to `incident` incident type, but our classifier only classifies the events which had the `SentineloOne Incident` as incident type. And we fixed this issue in code in V3.2.10 and corrected it with `SentinelOne Incident ` so that our classifier identifies the incident. But some of the customers are still seeing some unclassified incidents in their Xsoar, the reason is they had some incident which pulled before 3.2.10 version. So in-order to this backward compatible we mapped this "incident" to our classifier. I hope this make sense. if you think we need a call we can have it Regards

Thanks for explaining the flow.
I would still appreciate a call to see it live.

Ben

PR #30702, opened 2023-11-06
Ciac 8762/support rawjson for test incidents integration

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8762

## Description
Added a new command that has 2 arguments:
1. raw json
2. entry id to a json file
the command creates an incident object that is parsed to an incident via the fetch incident command (same as the other command that was already implemented)


DeanArbel (approved):
Well done

PR #30746, opened 2023-11-08
[wait_for_upload.py] increased TIMEOUT

michal-dagan:
## Status
- [ ] In Progress
- [x] Ready


yaakovpraisler:
@michal-dagan Why do we need to increase the timeout where the upload-flow shouldn't take more than 2-3 hours?

michal-dagan:
> @michal-dagan Why do we need to increase the timeout where the upload-flow shouldn't take more than 2-3 hours?

Due to limited resources, it now takes more than 6 hours in the new environment.

yaakovpraisler (approved):

PR #30858, opened 2023-11-13
fixing clean machine script, not having an artifacts folder created

kobymeir:
## Status
- [ ] In Progress
- [x] Ready

## Description
fixing clean machine script, not having an artifacts folder created.


tkatzir (approved):
Good catch!

PR #29165, opened 2023-08-23
DBotPredictURLPhishing - timeout argument

MLainer1:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CRTX-87358)

## Description
Added the rasterize_timeout argument to the script

## Must have
- [ ] Tests
- [ ] Documentation

PR #30903, opened 2023-11-14
Revert docker image - DBotPredictURLPhishing

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-30112

## Description
ForceMerge is needed here.
Reverting the docker image since in the new one, scikit-learn package was upgraded to 1.3.1 but it does not work with the current model (ML). 

## Must have
- [ ] Tests
- [ ] Documentation


tkatzir (approved):
Good. Approved.

tkatzir:
@rshunim ,
Please post a comment here with the details.
Both on the error message, and a link to Stack Overflow (or the previous PR where this was discussed,)
so future developers could find it easily.

rshunim:
We are reverting the docker image since the scikit-learn package was upgraded in the new docker image and then a version of this package was created once against the model. An option to update the model is available at @Yohan Guez, I also have attached the error and a [link](https://github.com/demisto/content/pull/29165#issuecomment-1694671914) to Github for another place where we met with this problem.
The error we got is: "Error: 'ColumnTransformer' object has no attribute '_name_to_fitted_passthrough'" as a result of mismatch between scikit-learn and model versions.

yuvalbenshalom (approved):

PR #30925, opened 2023-11-15
MISP v3: make TPB sequentially

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8593

## Description
its looks like inconsistent MISP server error when running "misp-add-attribute" in concurrency mode for the same event id

as few errors appear each time in another task from this group of tasks 
will update t heTPB to be in sequence.

before:
image

after:
image

take a bout 0.5 min more


DeanArbel (approved):
It's a shame it's taking longer, but as long as it's stable, this solution is preferable.

PR #31058, opened 2023-11-22
[Azure Sentinel] Fix first fetch

ilaner:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-29015

## Description
* Fetch by timestamp if `last_incident_number` is 0 or None, beacuse it means that no incidents were fetched.
* Update the last timestamp to now when no incidents were fetched.


Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -1356,7 +1356,9 @@ def process_incidents(raw_incidents: list, min_severity: int, latest_created_tim
             latest_created_time = incident_created_time
         if incident.get('IncidentNumber') > last_incident_number:
             last_incident_number = incident.get('IncidentNumber')
-
+    if not raw_incidents:
+        # if we don't have any raw incidents, we want to keep the last incident id and update the last_created_time to now
+        latest_created_time = datetime.utcnow()
```
yuvalbenshalom, line 1361: if last created time exists(so it is not 0) shouldn't we just preserve it and not override it with now?

yuvalbenshalom (changes requested):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
ilaner: You right, we can keep the old time.
The latest_created_time always exist so we can delete this line.

The only change is to continue to fetch by timestamp if the latest incident number is 0.

yuvalbenshalom (approved):

PR #31111, opened 2023-11-26
[CommonServerPython] Fix Polling Failure If `polling` Parameter is Missing

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Description
Currently, when there's a polling command with a polling argument (for example, `polling=true/false`, that decides whether to use polling for the command or not), if the polling argument is optional and not required (in my case, to not break BC. It just has a default value of `false`), and is not assigned (on playbooks, default parameters aren't assigned automatically to tasks), there's an "Error: Argument is neither a string nor a boolean." error:
Screenshot 2023-11-26 at 13 40 37

This happens due to `args.get(polling_arg_name)` having a `None` value, resulting in `argToBoolean(None)` being called.
This is resolved by assigning a default value of `False` instead of `None`.

An example of this error can be seen [in the following unit-test error](https://code.pan.run/xsoar/content/-/jobs/37031100), which fails due to the polling argument not being used when running the command in the unit-test.


Packs/Base/ReleaseNotes/1_32_49.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+
+##### CommonServerPython
+
+- Fixed an issue where polling commands might not work properly when  the `polling` parameter is missing.
```
MichaelYochpaz, line 6:
```suggestion
- Fixed an issue where polling commands might not work properly if the `polling` parameter is no provided.
```

GuyAfik (commented):
LGTM, @JudahSchwartz please take a look at this one, related to the polling decorator.

DeanArbel (approved):
LGTM

JudahSchwartz (approved):
nice catch

PR #31122, opened 2023-11-26
Fixes For 'Email Address Enrichment - Generic v2.1'

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9028

## Description
As part of the common playbook update effort, I reviewed 'Email Address Enrichment - Generic v2.1' and noticed the following issues:
1- Playbook input configurations - updated the following playbook input values:
'InternalDomains' playbook input - the value was configured as 'inputs.InternalDomains'. Removed the value and left it empty.
'Domain' playbook input - the value was configured as 'Account.Email.Domain'. Removed the value and left it empty.
2- Playbook input descriptions - The description of the 'UseReputationCommand' playbook input has changed to include 'The default value is false'.
3- Automation inputs configurations - In the 'Check Reputation' task, the 'email' input argument value was changed from 'inputs.Email' to 'Account.Email.Address' where 'Account.Email.NetworkType Equals External'.
5- Missing playbook outputs.


Packs/CommonPlaybooks/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
```diff
@@ -714,22 +718,18 @@ inputs:
...
-      filters:
-      - - operator: isNotEmpty
-          left:
-            value:
-              simple: inputs.Domain
-            iscontext: true
+      root: Account.Email
       transformers:
       - operator: uniq
+      accessor: Domain
```
ssokolovich, line 724: R u sure that you want to add all the domains (such as URL's or any extracted domains) from each incident?

Packs/CommonPlaybooks/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
```diff
@@ -714,22 +718,18 @@ inputs:
...
   value:
     complex:
-      root: inputs.Domain
-      filters:
-      - - operator: isNotEmpty
-          left:
-            value:
-              simple: inputs.Domain
-            iscontext: true
+      root: Account.Email
```
ssokolovich, line 721: usually, that won't be a domain ... maybe regex can help here

ssokolovich (commented):
Don't forget to add RN :)

idovandijk (changes requested):
@TalNos seems we've been working on the same playbook but with different fixes.
Can you please see the note in https://docs.google.com/spreadsheets/d/1IlruglY95NjARv3MgewjaQzEkxXNIplfVxQFnNVHe6w/edit#gid=381054217 ? The ticket is https://jira-dc.paloaltonetworks.com/browse/CIAC-9029

There is an issue where the internal emails are being enriched in the reputation command task. Here is the fix I added:
image


In addition, why are we removing the default inputs? They currently work with `Entity Enrichment - Generic v3/Phishing v2`. I'd say we want to keep them.
Without passing the internal domains and all domains we won't have the domain squatting task run. Let me know if you wanna discuss this :)

Packs/CommonPlaybooks/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
TalNos: removed it

Packs/CommonPlaybooks/Playbooks/playbook-Email_Address_Enrichment_-_Generic_v2.1.yml
TalNos: removed it

Packs/CommonPlaybooks/ReleaseNotes/2_4_33.md
```diff
@@ -0,0 +1,8 @@
+
+#### Playbooks
+
+##### Email Address Enrichment - Generic v2.1
+
+- Removed *'InternalRange'* playbook input's default value.
```
ssokolovich, line 6: I think u have added it back right?

ssokolovich:
> In addition, why are we removing the default inputs? They currently work with Entity Enrichment - Generic v3/Phishing v2. I'd say we want to keep them.

@idovandijk  - just to emphasize why @TalNos removed those defaults - as those `inputs.XXXX` are only being inherited in case you are using this playbook as a sub-playbook. This means that in the cases that you mentioned, it will work properly - but if I want to use this playbook as a sub playbook where those default inputs do not exist, I still need to point them properly.

TBH, I usually don't wanna remove default values from PB inputs in existing Playbooks as it might cause an issue, but here I see that differently.
Here is an example from the `Entity_Enrichment_-_Generic_v3.yml`( where those values are passed properly).
image

idovandijk:
> > In addition, why are we removing the default inputs? They currently work with Entity Enrichment - Generic v3/Phishing v2. I'd say we want to keep them.
> 
> @idovandijk - just to emphasize why @TalNos removed those defaults - as those `inputs.XXXX` are only being inherited in case you are using this playbook as a sub-playbook. This means that in the cases that you mentioned, it will work properly - but if I want to use this playbook as a sub playbook where those default inputs do not exist, I still need to point them properly.
> 
> TBH, I usually don't wanna remove default values from PB inputs in existing Playbooks as it might cause an issue, but here I see that differently. Here is an example from the `Entity_Enrichment_-_Generic_v3.yml`( where those values are passed properly). image

I agree. Eventually it comes down to whether the inputs are inherited from the default or are custom inputs. In the case of Entity Enrichment I'm not sure, but anything that uses the default inputs will fail if we delete them without specifying the new ones in that playbook

idovandijk (approved):

PR #31196, opened 2023-11-29
Deprecate 'Get endpoint details - Generic' Playbook

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9067

## Description
As part of the common playbook update effort, I reviewed 'Get endpoint details - Generic' and noticed it only runs the 'endpoint' command which is already implemented in 'Endpoint Enrichment - Generic v2.1'.
Upon checking, I discovered that only the 'NGFW Internal Scan' XSIAM playbook uses 'Get endpoint details - Generic' as a sub-playbook. 
Hence, the 'Get endpoint details - Generic' playbook should be deprecated and replaced with the 'Endpoint Enrichment - Generic v2.1', a sub-playbook of NGFW Internal Scan' XSIAM.


Packs/CommonPlaybooks/Playbooks/playbook-Get_endpoint_details_-_Generic.yml
ssokolovich, line 1: note that his PB is being used as a sub-playbook in a non-deprecated parent playbook. Maybe we can switch that quickly?

ssokolovich (approved):

PR #30897, opened 2023-11-14
[transformers] Enhance to be more durable

mmhw:
## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
Related: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-8741)

## Description
Enhance transformers to be more durable.

This is done by several types of changes:
- Change the script to `JS`
- Set `isArray` to `true`
- Add try/except and throw errors only by `return_error`

## TPB
IsEmailAddressInternal_Test

SetAndHandleEmpty_Test

SetIfEmpty_Test

PR #31222, opened 2023-11-30
[IsEmailAddressInternal] Fix an issue with **domain** argument

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-9069)

## Description
Fixed an issue where when the **domain** argument received a single value (rather than a list) it did not work as expected (issue was introduced in 1.12.53).

### DIFF
--- a/Packs/CommonScripts/Scripts/IsEmailAddressInternal/IsEmailAddressInternal.yml
+++ b/Packs/CommonScripts/Scripts/IsEmailAddressInternal/IsEmailAddressInternal.yml
@@ -41,3 +41,4 @@ scripttarget: 0
 fromversion: 5.0.0
 tests:
 - IsEmailAddressInternal_Test
+- TestIsEmailAddressInternal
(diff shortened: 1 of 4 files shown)


Packs/CommonScripts/Scripts/IsEmailAddressInternal/IsEmailAddressInternal.yml
```diff
@@ -41,3 +41,4 @@ scripttarget: 0
 fromversion: 5.0.0
 tests:
 - IsEmailAddressInternal_Test
+- TestIsEmailAddressInternal
```
DeanArbel, line 44: was this missed?

DeanArbel (approved):

Packs/CommonScripts/Scripts/IsEmailAddressInternal/IsEmailAddressInternal.yml
mmhw: It's an old TPB and it didn't run in the previous build [here](https://github.com/demisto/content/pull/30897), so it didn't fail and we didn't catch the bug

PR #31377, opened 2023-12-10
Feed Elastic: fetch in batches

ilaner:
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-8889
This will fetch indictaors in batches


Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch_test.py
```diff
@@ -291,3 +297,14 @@ def test_extract_api_from_username_password_username_api_key():
     import FeedElasticsearch as esf
     username = esf.API_KEY_PREFIX + 'api_id'
     assert esf.extract_api_from_username_password(username, 'api_key') == ('api_id', 'api_key')
+
+
```
yuvalbenshalom, line 301: check recommit error

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.py
```diff
@@ -208,36 +209,53 @@ def get_demisto_indicators(search, tags, tlp_color):
...
+    last_ids = []
+    for ioc in reversed(ioc_lst):
+        calculate_time = dateparser.parse(ioc.get(client.time_field))
+        if calculate_time and (not last_calculated_time or calculate_time > last_calculated_time):
+            last_calculated_time = calculate_time
+            last_ids.append(ioc.get('id'))
+        else:
+            break
+    if last_calculated_time is None:
+        last_calculated_time = datetime.now()
```
yuvalbenshalom, line 223: if last calculated time is None, meaning probably we did not fetch any indicators, shodden we just preserve the preserve time we fetched with on the current fetch?

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.yml
```diff
@@ -100,7 +105,7 @@ configuration:
   name: feedExpirationInterval
   type: 1
   required: false
-- defaultvalue: '240'
+- defaultvalue: '10'
```
yuvalbenshalom, line 108: what was this changed?

yuvalbenshalom (commented):
looks good.
see my questions.

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.yml
ilaner: This is the fetch interval.
SInce in the previous version it fetched all the indicators in one batch, it was making sense to set the feed interval for 4 hours.
In the current version it will fetch in batches, so it doesn't really make sense to wait 4 hours until the next batch.

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.py
ilaner: We want to update the last run to now so the query will be as efficient as possible.
If we didn't receive any indicators, we can update it to the `now` time.

yuvalbenshalom (approved):

PR #31402, opened 2023-12-11
[PAN-OS Policy Optimizer] Add pagination support to `pan-os-po-get-rules`

MichaelYochpaz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-8844)

## Description
Add pagination support to the `pan-os-po-get-rules` command (which previously fetched only up to 200 items), using 2 new parameters - `limit` and `page_size`.


DeanArbel (approved):
Looks good

PR #31414, opened 2023-12-12
Fix js get incident tasks by state

ShahafBenYakir:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9172

## Description
Changed `includes` to `indexOf` as it is not supported in ecma 5 which means it will not work in xsoar 6.9


adi88d (approved):
nice

ShahafBenYakir:
Old script does not have test PB and is written in JS so no unit tests.

PR #31418, opened 2023-12-12
Fix "unexpected keyword argument" Error

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Currently, if more args than specified in the yml would be added to the If-Elif transformer, it would throw an error as the args are unpacked directly into the `ConditionParser` constructor and it has only the yml args a parameters.
To solve this, this PR adds the parameter `**_` to the constructor so that the unexpected args can be safely ignored.

## Must have
- [ ] Tests
- [ ] Documentation


JudahSchwartz (approved):
Noice!

PR #31497, opened 2023-12-16
Ews header deduplication bug

ieichen137:

## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
This is a quick bug fix for the recent additions to the EWS_0365 which added the rfc compliant specifications along with the header deduplication prevention.

We just had one email come through where the DATE header value stored in the O365 object was different than the value stored in the EML bytes. 

When parsing using the python email library, since when using the SMTP and SMTPUTF8 policies, some  headers have specific restrictions on how many can be included. DATE has a restriction of 1, and since the header deduplication works off key/value pair, as some headers can and should appear more than once, the DATE header was attempted to be added again with the altered value from the microsoft api data, which threw this uncaught exception. 

This change adds a simple try/except around the add_header call to catch this specific error:

```
 ValueError: There may be at most <n> <header name> headers in a message
```

where <n> is the limit and <header name> is the name of the header. 

the exception logic is this:

```python
...
try:
    email_content.add_header(header.name, header.value)
except ValueError as err:
    if "There may be at most" not in str(err):
        raise err
```

Which will catch a value error and check if it matches the error message for this problem. If it does, the function will continue as normal. If it does not, the error is raised. 

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/MicrosoftExchangeOnline/Integrations/EWSO365/EWSO365.py
+++ b/Packs/MicrosoftExchangeOnline/Integrations/EWSO365/EWSO365.py
@@ -2023,7 +2023,11 @@ def get_item_as_eml(client: EWSClient, item_id, target_mailbox=None):      # pra
                         header.name.lower(),
                         header.value,
                 ) not in attached_email_headers and header.name.lower() != "content-type":
-                    email_content.add_header(header.name, header.value)
+                    try:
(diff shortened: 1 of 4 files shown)


amshamah419 (approved):
Great work!

PR #31531, opened 2023-12-18
changed test to correspond to the latest changes on xsoar8

ShahafBenYakir:
## Status
- [x] Ready
- [ ] In Progress

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-31482

## Description
In XSOAR8 `getIncidents` always returns `total:0` (breaking changes), so we cannot use it in our loop. Instead, we check every loop if the page contains exactly 100 (page size) results, if so it means we might have a next page. We finish the loop when either we have pulled all results or when the number of results is smaller than 100.

## Must have
- [x] Tests


ShahafBenYakir:
Failing UT reproducing the issue https://github.com/demisto/content/actions/runs/7245826775/job/19736592053

JudahSchwartz:
you only changed a test file here?

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
```diff
@@ -163,24 +162,27 @@ def search_incidents(args: Dict):   # pragma: no cover
...
-        demisto.debug(
-            f'before filtering {len(current_page_found_incidents)=} '
-            f' {args=} {page=}'
-        )
+
+        # When current_page_found_incidents is None it means the requested page was empty
+        if not current_page_found_incidents:
+            break
+
+        demisto.debug(f'before filtering {len(current_page_found_incidents)=} {args=} {page=}')
```
JudahSchwartz, line 176: would add a log for after filtering len too

JudahSchwartz (approved):
📃 🚀 📟

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
ShahafBenYakir: roger

ShahafBenYakir: @JudahSchwartz but there is line 180

ShahafBenYakir:
code.pan.run is failing on a known issue with cloud tenants' installation of packs, see [this](https://panw-global.slack.com/archives/C03A9TJ57FT/p1702919593245019?thread_ts=1702898307.617109&cid=C03A9TJ57FT).

PR #31579, opened 2023-12-20
[pre-commit] remove __init__folders, fix script_runner and add commit mode

ilaner:
* fix issue with script runner when the script errors resulted error code 0 (not pytest failures, but failures executing the script)
* Remove __init__ folders from content (if we want to add `__init__`, we need to add in all of them, since those packages has priority over others.
* Add commit mode to the template


ilaner:
this will need a force merge because we delete`__init__.py` files and this is not allowed in our validate:
image

kobymeir (approved):

PR #31699, opened 2023-12-24
e2e tests SDK Nightly

MLainer1:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue
](https://jira-dc.paloaltonetworks.com/browse/CIAC-9281)
## Description
Changes:
- Added TestCollector for SDK Nightly, to collect only the relevant sanity tests
- Added support for the DEMISTO_SDK_NIGHTLY variable in collect_tests.py and install_and_configure_integrations.py

## Must have
- [ ] Tests
- [ ] Documentation


MLainer1:
Nightly run https://code.pan.run/xsoar/content/-/pipelines/7252399

Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -1399,6 +1399,12 @@ def get_e2e_packs(self) -> set[str]:
         return {"TAXIIServer", "EDL", "QRadar", "Slack"}
 
 
+class SDKNightlyTestCollector(TestCollector):
+
+    def _collect(self) -> CollectionResult | None:
+        return CollectionResult.union((self.sanity_tests))  # type: ignore
```
GuyAfik, line 1405:
```suggestion
        return CollectionResult.union((self.sanity_tests))  # type: ignore[<kind_of_error>]
```

can we avoid it type: ignore? and if not at least state which type of mypy rule should be ignored

Tests/scripts/install_content_and_test_integrations.sh
```diff
@@ -23,6 +23,7 @@ fi
 CONF_PATH="./Tests/conf.json"
 
 [ -n "${NIGHTLY}" ] && IS_NIGHTLY=true || IS_NIGHTLY=false
+[ -n "${DEMISTO_SDK_NIGHTLY}" ] && DEMISTO_SDK_NIGHTLY=true || DEMISTO_SDK_NIGHTLY=false
```
GuyAfik, line 26: what's this change?
can you add a comment here?

GuyAfik (commented):
@MLainer1 🏆 
See my small comments

Tests/scripts/collect_tests/collect_tests.py
MLainer1: yes

Tests/scripts/install_content_and_test_integrations.sh
MLainer1: If the DEMISTO_SDK_NIGHTLY variable is set, we set it to true, else, we set it to false.
This way we will always have a value

Tests/configure_and_test_integration_instances.py
```diff
@@ -220,7 +220,7 @@ def __init__(self, options):
         self.git_sha1 = options.git_sha1
         self.branch_name = options.branch
         self.ci_build_number = options.build_number
-        self.is_nightly = options.is_nightly
+        self.is_nightly = options.is_nightly or options.is_sdk_nightly
```
ilaner, line 223: if sdk nightly is true, is nightly should be false.

ilaner (approved):
Just see comments

Tests/scripts/install_content_and_test_integrations.sh
dorschw: It may be confusing to devs in the future, I think code accessing this env var should consider it optional (may be missing) and use a default, instead.

PR #31778, opened 2023-12-26
Pre commit config template config update

ilaner:
* add fixable property
* add min_py_version property
* save pytest coverage reports in the pre-commit folder and use parallel mode.

### DIFF
--- a/.pre-commit-config_template.yaml
+++ b/.pre-commit-config_template.yaml
@@ -94,16 +94,21 @@ repos:
     language: docker
     files: Packs\/.*_test\.py$
     exclude: test_data
+    env:
+      COVERAGE_FILE: /src/.pre-commit/coverage/.coverage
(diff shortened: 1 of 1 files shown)


.pre-commit-config_template.yaml
```diff
@@ -7,10 +7,13 @@ repos:
   - id: check-json
   - id: check-yaml
   - id: check-ast
+    min_py_version: "3.7"
```
dorschw, line 10: can we use a constant? this value is repeated multiple times

ilaner: I don't want to overcomplicte this YAML with anchors.
In addition, when we upgrade the hook versions we will need to check this value, as it's possible that the newer version dropped support (for example, ruff supports now python 3.8 and up)

PR #31903, opened 2024-01-02
Improve performance for DBotPredictURLPhishing

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] Blocked

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-9357)

## Description
- Improve the runtime of the **DBotPredictURLPhishing** script by lowering the times server commands are called.
- Fixed an issue where the output field `current_url` of the command `!rasterize type=json` would include all URLs in the `url` argument instead of the URL associated with the individual output. 

## Must have
- [ ] Tests
- [ ] Documentation


JudahSchwartz (approved):
looks great!!!

Packs/rasterize/Integrations/rasterize/rasterize.py
```diff
@@ -873,40 +872,45 @@ def add_filename_suffix(file_names: list, file_extension: str):
 
 
 def rasterize_command():  # pragma: no cover
-    url = demisto.getArg('url')
+    urls = argToList(demisto.getArg('url'))
```
tkatzir, line 875: `url` can either be a list, or a single str.
This is to support url's that contain a comma.

tkatzir (changes requested):
Looks good!
One comment re `url` type.

tkatzir:
@jlevypaloalto,
You should also support file_name array.
See previous PR: https://github.com/demisto/content/pull/33406

Packs/rasterize/Integrations/rasterize/rasterize.py
tkatzir: @jlevypaloalto, I'm not sure.
As you wrote it, if `url` contains a comma - It will be split to an array of several "urls" (in quote because the url's will be messed up,) instead of staying only one url input.

jlevypaloalto: Fixed, please review.

jlevypaloalto:
This PR needs a force-merge because the build does not deal with headless-chrome.

PR #31936, opened 2024-01-03
Added round transformer for playbook "wait until datetime"

altmannyarden:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-31808)

## Description
Added another transformer for the sleep commend to round non int values. 

## Must have
- [ ] Tests
- [ ] Documentation


idovandijk (approved):
Great. Maybe also explain the reason - sleep doesn't work with fractions so we had to round the number

PR #31997, opened 2024-01-07
[PAN-OS] Change if-elif to match case

jlevypaloalto:
## Status
- [x] In Progress
- [ ] Ready

## Description
Change the command if-elif tree to a match case tree in main().

## Must have
- [ ] Tests
- [ ] Documentation


GuyAfik (approved):
add rn
Please check it properly with TPB

PR #32009, opened 2024-01-07
AWS SecHub Fix for AWS Filter Limitation

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-31890

## Description
We now use the same IDs we saved prior for the filter and instead just filter them from the results received.
## Must have
- [ ] Tests
- [ ] Documentation


Packs/AWS-SecurityHub/ReleaseNotes/1_3_26.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### AWS Security Hub Event Collector
+
+- Fixed an issue where event collection would fail while filtering large amounts of events.
+- Updated the Docker image to: *demisto/boto3py3:1.0.0.84082*.
```
amshamah419, line 7:
```suggestion
- Updated the Docker image to: *demisto/boto3py3:1.0.0.84273*.
```

Packs/AWS-SecurityHub/Integrations/AWSSecurityHubEventCollector/AWSSecurityHubEventCollector.py
```diff
@@ -107,6 +108,10 @@ def get_events(client: "SecurityHubClient", start_time: dt.datetime | None = Non
 
         response = client.get_findings(**kwargs)
         result = response.get('Findings', [])
+
+        # Filter out events based on id_ignore_set
+        result = [event for event in result if event.get('Id') not in id_ignore_set]
```
MichaelYochpaz, line 113: Not something too important, but I think this should be more appropriate:
```suggestion
        result = [event for event in result if event['Id'] not in id_ignore_set]
```
since we always expect ID in responses (and if not, we want to raise an error since something went wrong), plus we already fetch the ID using `event['id']` (and not using `get`) in other places, so it will fail anyway in such case.

MichaelYochpaz (approved):
Looks good to me, good job :)

PR #32137, opened 2024-01-11
CS detonate file PB bug

anas-yousef:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-32005

## Description
Add error handling for the commands `cs-fx-get-full-report` and `cs-fx-get-report-summary`, where the report could hold error messages that weren't handled.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1092,6 +1118,15 @@ def get_report_summary_command(
...
         response = client.get_report_summary(single_id)
+        if response.get('resources'):
+            is_polling = argToBoolean(polling)
+            # We can extract the error from the response object under resources
+            error_message = validate_sandbox_report(response.get('resources', []))
+            if error_message:
+                if is_polling:
+                    raise DemistoException(error_message)
+                if not is_polling:
+                    return_warning(error_message)
```
anas-yousef, line 1129: The condition checking if we are using polling may be redundant, since we use the command `cs-fx-get-full-report` to retrieve the report of the submitted file when polling, and not the `cs-fx-get-report-summary` command (the function `get_report_summary_command` is the entry point for the `cs-fx-get-report-summary` command)

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1302,7 +1337,7 @@ def get_results_function_args(outputs, extended_data, item_type, interval_in_sec
 
 
 def pop_polling_related_args(args: dict) -> None:
-    for key in ('submit_file', 'enable_tor', 'interval_in_seconds', 'polling', 'environment_id'):
+    for key in ('submit_file', 'enable_tor', 'interval_in_seconds', 'environment_id'):
```
anas-yousef, line 1340: The function `pop_polling_related_args` is called before requesting a status update when running the `upload file` command with polling, and we specifically need the `polling` argument, so we can know if we should throw an error or just a warning when analyzing the report.

anas-yousef:
@ssokolovich The playbook is ready for review, if you want, we can do a small demo to demonstrate the changes.
@JasBeilin Once you finish the review, I will update the unit tests accordingly.

Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_Intelligence_Sandbox_v2.yml
```diff
@@ -59,17 +59,17 @@ tasks:
...
+      id: f38c9e4a-4710-434e-83fe-aeb69b6f67c5
       version: -1
       name: CrowdStrike Falcon Intelligence Sandbox Get Full Report
       description: Get a full version of a sandbox report.
-      script: CrowdStrike Falcon X|||cs-fx-get-full-report
+      script: CrowdStrike Falcon X Test|||cs-fx-get-full-report
       type: regular
       iscommand: true
-      brand: CrowdStrike Falcon X
+      brand: CrowdStrike Falcon X Test
```
ssokolovich, line 72: No need for brand - especially for `test`

Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_Intelligence_Sandbox_v2.yml
```diff
@@ -59,17 +59,17 @@ tasks:
...
+    taskid: f38c9e4a-4710-434e-83fe-aeb69b6f67c5
     type: regular
     task:
-      id: c7c659db-15a7-4c9b-8f3d-72713eefe5f1
+      id: f38c9e4a-4710-434e-83fe-aeb69b6f67c5
       version: -1
       name: CrowdStrike Falcon Intelligence Sandbox Get Full Report
       description: Get a full version of a sandbox report.
-      script: CrowdStrike Falcon X|||cs-fx-get-full-report
+      script: CrowdStrike Falcon X Test|||cs-fx-get-full-report
```
ssokolovich, line 69: remove the exact instance + test

Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_Intelligence_Sandbox_v2.yml
```diff
@@ -157,7 +157,7 @@ tasks:
                       iscontext: true
                     right:
                       value:
-                        simple: CrowdStrike Falcon X
+                        simple: CrowdStrike Falcon X Test
```
ssokolovich, line 160: Same here - pls remove `test`

Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_Intelligence_Sandbox_v2.yml
ssokolovich, line 1: please update the image of this PB in the /docs

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1000,7 +1001,15 @@ def get_full_report_command(
...
         if response.get('resources'):
+            is_polling = argToBoolean(polling)
             is_command_finished = True  # flag used when commands
+            # We can extract the error from the response object under resources
+            error_message = validate_sandbox_report(response.get('resources', []))
+            if error_message:
+                if is_polling:
+                    raise DemistoException(error_message)
+                if not is_polling:
+                    return_warning(error_message)
```
JasBeilin, line 1012: Why are we changing it here and not in the polling process?

JasBeilin, line 1012: Why are we changing it here and not in the polling process?

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
JasBeilin, line 1129: same, why do it as part of the polling process?

JasBeilin (commented):
Why are

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1043,6 +1046,24 @@ def get_full_report_command(
         ]
     return command_results, is_command_finished
 
+def validate_sandbox_report(report_resources: list[dict[str, Any]]) -> str:
```
dorschw, line 1049: `validate` method returning a string with the error is a bit unconventional.

Instead, please create a custom exception type (above) and raise it here (I'd include the word `Warning` in its name), catch wherever this code is used and act (`return_warning`) accordingly.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1375,11 +1403,32 @@ def run_polling_command(client, args: dict, cmd: str, upload_function: Callable,
             timeout_in_seconds=6000)
 
         command_result = CommandResults(scheduled_command=scheduled_command)
+    elif post_function:
+        # Validate the polling results
+        validate_submit_file_polling_results(command_result)
     return command_result
 
+def validate_submit_file_polling_results(command_results: list[CommandResults] | CommandResults):
```
dorschw, line 1411: missing return type hints

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1375,11 +1403,32 @@ def run_polling_command(client, args: dict, cmd: str, upload_function: Callable,
...
+
+    Args:
+        command_results (list[CommandResults] | CommandResults): The results of the polling method.
+
+    Raises:
+        DemistoException: If the results contain an error message, stating that the sandbox analysis was not able to run
+        properly.
+    """
+    command_results = command_results if isinstance(command_results, list) else [command_results]
+    for command_result in command_results:
```
dorschw, line 1422:
```suggestion
    for command_result in always_iterable(command_results, CommandResults):
```
make sure to import `always_iterable` from `more_itertools` _(included in all our py3 images)_

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1375,11 +1403,32 @@ def run_polling_command(client, args: dict, cmd: str, upload_function: Callable,
             timeout_in_seconds=6000)
 
         command_result = CommandResults(scheduled_command=scheduled_command)
+    elif post_function:
```
dorschw, line 1406:
```suggestion
    elif post_function is not None:
```

dorschw, line 1406: is the `post_function` used?

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1375,11 +1403,32 @@ def run_polling_command(client, args: dict, cmd: str, upload_function: Callable,
...
+        # Validate the polling results
+        validate_submit_file_polling_results(command_result)
```
dorschw, line 1408: here the code doesn't use the output of `validate`, is this intended?

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1043,6 +1046,24 @@ def get_full_report_command(
...
+
+    Returns:
+        str: The error message found in the sandbox, or an empty string if no error is found.
+    """
+    for resource in report_resources:
+        resource_id = resource.get('id')
+        for sandbox_entity in resource.get('sandbox', []):
+            if error_message := sandbox_entity.get('error_message'):
+                error_type = sandbox_entity.get('error_type')
+                return (f'Sandbox report for resource id {resource_id} returned an error of'
```
dorschw, line 1064: this current mechanism returns on the first warning (and stop running on others), is this the intention?

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1001,6 +1000,10 @@ def get_full_report_command(
         response = client.get_full_report(id_)
         if response.get('resources'):
             is_command_finished = True  # flag used when commands
+            # We can extract the error from the response object under resources
```
JasBeilin, line 1003:
```suggestion
            # We can extract the error from the response object under resources section, as it looks like the error section is always empty.
```

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1043,6 +1046,24 @@ def get_full_report_command(
...
+
+    Args:
+        report_resources (list[dict[str, Any]]): Report resources of the report. They hold the data about
+        any error messages returned from the report.
+
+    Returns:
+        str: The error message found in the sandbox, or an empty string if no error is found.
+    """
+    for resource in report_resources:
+        resource_id = resource.get('id')
```
JasBeilin, line 1060:
```suggestion
        resource_id = resource['id']
```
Id should always be present.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1043,6 +1046,24 @@ def get_full_report_command(
         ]
     return command_results, is_command_finished
 
+def validate_sandbox_report(report_resources: list[dict[str, Any]]) -> str:
+    """This function checks for any error messages in the sandbox report.
```
JasBeilin, line 1050: since we are using report and this is not generic, lets add a raw report for example.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1092,6 +1113,11 @@ def get_report_summary_command(
 
     for single_id in argToList(ids):
         response = client.get_report_summary(single_id)
+        if response.get('resources'):
+            # We can extract the error from the response object under resources
```
JasBeilin, line 1117: same here, we want to explain to the next dev why we are looking inside resources when we have error section.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1375,11 +1403,32 @@ def run_polling_command(client, args: dict, cmd: str, upload_function: Callable,
             timeout_in_seconds=6000)
 
         command_result = CommandResults(scheduled_command=scheduled_command)
+    elif post_function:
+        # Validate the polling results
+        validate_submit_file_polling_results(command_result)
     return command_result
```
JasBeilin, line 1409:
```suggestion
    elif post_function:
        # Validate the polling results
    return  post_function(command_result) if post_function is not None else command_result
```

as discussed, we want to make sure we use the post function when needed and keep it generic as possible, without hurting other functionality.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX_test.py
```diff
@@ -120,6 +121,51 @@ def json(self):
...
+    mocker.patch.object(Client, 'upload_file', return_value=UPLOAD_FILE_HTTP_RESPONSE)
+    mocker.patch.object(Client, 'send_uploaded_file_to_sandbox_analysis', return_value=SEND_URL_TO_SANDBOX_ANALYSIS_HTTP_RESPONSE)
+    mocker.patch.object(Client, 'get_full_report', return_value=GET_FULL_REPORT_HTTP_RESPONSE_ERROR_MESSAGE)
+    with pytest.raises(DemistoException) as e:
+        run_polling_command(client, args, 'cs-fx-upload-file', upload_file_command,
+                                          get_full_report_command, 'FILE', post_function=validate_sandbox_report)
+    assert 'Sandbox was not able to analyze one of the files, failing with error' in str(e)
+
+
+@pytest.mark.parametrize('command', [(get_full_report_command), (get_report_summary_command)])
```
JasBeilin, line 148: Im not a big fan of using the parametrize to distinct 2 running function, I think a class is better here in general. Since the code is really the same in this case Im ok with it here but it is not the best practice to do so,

JasBeilin (changes requested):
Nice!

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: My bad, I meant to reference `post_function`.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
```diff
@@ -1043,6 +1046,24 @@ def get_full_report_command(
...
+        any error messages returned from the report.
+
+    Returns:
+        str: The error message found in the sandbox, or an empty string if no error is found.
+    """
+    for resource in report_resources:
+        resource_id = resource.get('id')
+        for sandbox_entity in resource.get('sandbox', []):
+            if error_message := sandbox_entity.get('error_message'):
+                error_type = sandbox_entity.get('error_type')
```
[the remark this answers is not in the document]
anas-yousef: I am not sure, in my case, when I got an `error_message`, I would get an `error_type` with it, but I can't say that for all reports.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: Nice

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: Yes, I forgot to reference it, and instead directly referenced `validate_submit_file_polling_results`.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: Yeah, because the whole point of the post function, in this case, is to validate the content of the report analysis, and if it finds any error messages, then it will raise an error, if no error is found, then we continue normally.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: I made it as such:

`elif post_function is not None:`
`    # Validate the polling results`
`    post_function(command_result)`
`return command_result`

just because `post_function` does not return anything, it only validates the content of the polling results, and if there is an error in the report, then it raises an exception, otherwise, it does not return anything.

anas-yousef:
@ssokolovich Validation is telling me that the playbook input `Timeout` is not used, I checked in the [master's branch](https://github.com/demisto/content/blob/master/Packs/CrowdStrikeFalconX/Playbooks/Detonate_File_-_CrowdStrike_Falcon_Intelligence_Sandbox_v2.yml), and I can indeed see that this input is not used, I will delete it.

Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py
anas-yousef: I believe yes, in the cases that the sandbox returned an error when I was testing them, only one entry was returned, holding the error object, I believe even if further entries are found, returning one is enough, since this means the sandbox report faced some issues.

JasBeilin (approved):

PR #32331, opened 2024-01-21
bumped version to all core packs

yucohen:
Upgraded core packs to freeze their local version


DeanArbel (approved):

yucohen:
forcing as build fails unrelated

PR #32552, opened 2024-01-31
[CommonScripts] Fix comparison key in ProvidesCommand script

pl-brault:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
When comparing the instance name between the enabled instances and the integration commands , the check was made on the "display" key  which wasn't the same as the "brand" key retrieved before in `integration_search_res`.

Example:
Here is the display:
image

Here is the brand:
image

## Must have
- [ ] Tests
- [ ] Documentation


YuvHayun:
Hi @pl-brault, thx for your contribution.
I'm not sure I fully understand the issue, can you try to elaborate a bit about the problem?

pl-brault:
> Hi @pl-brault, thx for your contribution. I'm not sure I fully understand the issue, can you try to elaborate a bit about the problem?

Hello, 

This Script find all available integration instance for a specific command passed in argument.
To find which integration that implement the command, the script  run two API requests:
- one that list all integration with their commands (line 18,19)
- one that list all available integration (line 28)

Line 41 the script construct a dict based on the the "brand" for all integration enabled.
Then when iterating (line 56-64) on all all commands for all integrations, if we want only enabled instance, the script search for the "display name" of the integration in the dict constructed line 41, therefore if the brand and the display name of an integration isn't the same the "integration" isn't return even if it has the command implemented and is enable.

For example:
For the integration EclecticIQ  the brand is `EclecticIQ Platform` and the display name is `EclecticIQ Platform (Deprecated) (Partner Contribution)`
therefore if we search for `EclecticIQ Platform (Deprecated) (Partner Contribution)` in the dict that has the key `EclecticIQ Platform` it's never return

```python
integration_instances_enabled = {
    "EclecticIQ Platform": True
}

# This line is never true because we compare the display name which isn't in integration_instances_enabled 
"EclecticIQ Platform (Deprecated) (Partner Contribution)" in integration_instances_enabled
```

YuvHayun:
> > Hi @pl-brault, thx for your contribution. I'm not sure I fully understand the issue, can you try to elaborate a bit about the problem?
> 
> Hello,
> 
> This Script find all available integration instance for a specific command passed in argument. To find which integration that implement the command, the script run two API requests:
> 
> * one that list all integration with their commands (line 18,19)
> * one that list all available integration (line 28)
> 
> Line 41 the script construct a dict based on the the "brand" for all integration enabled. Then when iterating (line 56-64) on all all commands for all integrations, if we want only enabled instance, the script search for the "display name" of the integration in the dict constructed line 41, therefore if the brand and the display name of an integration isn't the same the "integration" isn't return even if it has the command implemented and is enable.
> 
> For example: For the integration EclecticIQ the brand is `EclecticIQ Platform` and the display name is `EclecticIQ Platform (Deprecated) (Partner Contribution)` therefore if we search for `EclecticIQ Platform (Deprecated) (Partner Contribution)` in the dict that has the key `EclecticIQ Platform` it's never return
> 
> ```python
> integration_instances_enabled = {
>     "EclecticIQ Platform": True
> }
> 
> # This line is never true because we compare the display name which isn't in integration_instances_enabled 
> "EclecticIQ Platform (Deprecated) (Partner Contribution)" in integration_instances_enabled
> ```

Thx a lot for the clarification.
Now I fully understand the issue, I'll review it quickly.

Packs/CommonScripts/ReleaseNotes/1_13_32.md
```diff
@@ -0,0 +1,7 @@
+
+#### Scripts
+
+##### ProvidesCommand
+
+- Fixed an issue where the comparison key wasn't the same when checking that a command is part of enabled instances.
```
YuvHayun, line 6:
```suggestion
- Fixed an issue where the script would miss enabled commands due to a mismatch between comparison key when checking that a command is part of enabled instances.
```

YuvHayun (changes requested):
Nice catch, please add tests and note my comment and pre-commit failures.

YuvHayun:
Moving to instance test.

YuvHayun:
Merging to intenrnal PR, thx a lot for your contribution.

PR #32556, opened 2024-01-31
Add pyzbar to ReadQRCode

jlevypaloalto:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [XSUP-32537](https://jira-dc.paloaltonetworks.com/browse/XSUP-32537)

## Description
Add the `pyzbar` library to ReadQRCode to be used when `cv2` fails.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/ReadQRCode/ReadQRCode.py
```diff
@@ -45,7 +50,7 @@ def extract_info_from_qr_code(entry_id: str, **_) -> CommandResults:
 
 def main():
     try:
-        return_results(extract_info_from_qr_code(**demisto.args()))
+        return_results(extract_info_from_qr_code(**demisto.args()['entry_id']))
```
JudahSchwartz, line 53:
```suggestion
        return_results(extract_info_from_qr_code(demisto.args()['entry_id']))
```

JudahSchwartz (approved):
looks good, lots of tests please for qr codes of all sorts.

Packs/CommonScripts/Scripts/ReadQRCode/ReadQRCode.py
```diff
@@ -10,18 +11,22 @@ def read_qr_code(filename: str) -> str:
     img = cv2.imread(filename)
```
enes-oezdemir, line 11: Just something to mention, cv2 is utilizing c library calls so if those libraries issue a warning the task will be considered a failure. Warnings are written to stderr so xsoar thinks the task failed.

I faced this issue with libpng.

I would recommend redirecting stderr here

from wurlitzer import pipes

with pipes() as (out, err):
    cv2.imread(filename)
    
demisto.info(err.read())

JudahSchwartz: Thanks for the tip! @jlevypaloalto will look into it

jlevypaloalto: Hi @enes-oezdemir, thanks for your input!
It would be helpful if you could provide an example of a warning in cv2 that concerns our use-case.

enes-oezdemir: here is a warning when runnign cv2.imread() with an image in my environment

libpng warning: bKGD: invalid

Packs/CommonScripts/Scripts/ReadQRCode/ReadQRCode.py
```diff
@@ -1,35 +1,38 @@
 import demistomock as demisto  # noqa
 from CommonServerPython import *  # noqa
+from pyzbar import pyzbar
 import cv2
 # pylint: disable=E1101  # disable pylint not recognizing cv2's attributes.
 
 
-def read_qr_code(filename: str) -> str:
+def read_qr_code(filename: str) -> str | list[str]:
```
JudahSchwartz, line 8: dont return a string or a list. 
a single string should be put into a list of 1

jlevypaloalto: what's the point? it's extra work and the output is exactly identical

JudahSchwartz: its bad practice to have a variable return type to your functions

Packs/CommonScripts/Scripts/ReadQRCode/ReadQRCode.py
enes-oezdemir: Untitled

For example, this is a file, I know that this is not a valid QR code but cv2.imread will complain even when loading the image (cv2.imread). This behavior might result in a task error even if you have a valid QR code

jlevypaloalto: Thanks for your help @enes-oezdemir! 😇
I'll add the context manager to the code.

JudahSchwartz (approved):
awesome

PR #32622, opened 2024-02-01
Recorded Future Feed - bug fix: malicious threshold integration parameter

randomizerxd:
'malicious_threshold' in code did not match 'threshold' integration parameter. We recently contributed PR 32316 which introduced this bug where the integration will always fail. 

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
change 'threshold' integration parameter to 'malicious_threshold' for clarity between 'suspicious_threshold'

## Must have
- [ ] Tests
- [ ] Documentation


YuvHayun:
Hi @randomizerxd, thx a lot for the clarification.
However, changing the name field of an integration parameter isn't allowed because it can break BC.
What we usually do in this cases, is hiding the existing param and adding a new one to the integration while making sure that each call to this certain param is trying to receive its info either from the previous name or the new one.

However, I don't recommend doing so for this issue since what customers see is the display (which is indeed malicious_threshold). therefore I don't think this can confuse the customers.

Let me know if you still prefer to change the name or close the pr.

YuvHayun:
Closing this pr, a fix will be merged in [this](https://github.com/demisto/content/pull/32688) pr.

PR #32718, opened 2024-02-05
[UploadFlow] Fix git diff

yaakovpraisler:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-9661)

## Description
Fixed git diff in collect-tests to compare the last upload commit against the current commit (and not origin/master)


Tests/scripts/collect_tests/collect_tests.py
```diff
@@ -1124,7 +1124,7 @@ def _get_git_diff(self) -> FilesToCollect:
             logger.info('bucket upload: getting last commit from index')
             previous_commit = get_last_commit_from_index(self.service_account, self.marketplace)
             if self.branch_name == 'master':
-                current_commit = 'origin/master'
+                current_commit = os.getenv("CI_COMMIT_SHA") or "origin/master"
```
dorschw, line 1127: why use a default here? I think it's safe to just take `os.environ["CI_COMMIT_SHA"]`

yaakovpraisler: Why not? I think it's better to have a default here

kobymeir (approved):

Tests/scripts/collect_tests/collect_tests.py
dorschw: DMd you

PR #32778, opened 2024-02-07
[FeedElasticSearch] Fix ids in last run 

ilaner:
In the previous version, we intorduced fetching indicators from elastic search in batches. We save in the last run the last indicator timestamp and the ids which were fetched in the previous version, to handle duplicates.

There is an issue where the ids list is based on the `datetime` object and not the timestamp, which causes this list to be inaccurate.

related: https://jira-dc.paloaltonetworks.com/browse/XSUP-33218

### DIFF
--- a/Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch_test.py
+++ b/Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch_test.py
@@ -307,4 +307,4 @@ def test_last_run():
                {"id": "4", "calculatedTime": "2023-01-17T14:33:00.000Z"}]
     last_update, last_ids = update_last_fetch(MockClient(), ioc_lst)
     assert set(last_ids) == {"4", "3"}
-    assert datetime.fromtimestamp(last_update // 1000).isoformat() == "2023-01-17T14:33:00+00:00"
+    assert datetime.fromtimestamp(last_update // 1000).isoformat() == "2023-01-17T14:33:00"


Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.py
```diff
@@ -210,18 +210,27 @@ def get_demisto_indicators(search, tags, tlp_color):
...
     last_ids = []
     for ioc in reversed(ioc_lst):
         calculate_time = dateparser.parse(ioc.get(client.time_field))
-        if calculate_time and (not last_calculated_time or calculate_time >= last_calculated_time):
-            last_calculated_time = calculate_time
+        if not calculate_time:
+            demisto.info(f"ioc {ioc.get('name')} if missing {client.time_field}")
+            break
+        calculate_timestamp = int(calculate_time.timestamp() * 1000)
+        if not last_calculated_timestamp or calculate_timestamp == last_calculated_timestamp:
```
yuvalbenshalom, line 222:
```suggestion
        if not last_calculated_timestamp or calculate_timestamp >= last_calculated_timestamp:
```
I think it sold be >=, not ==
am I missing something? maybe aUT is needed.

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.py
```diff
@@ -210,18 +210,27 @@ def get_demisto_indicators(search, tags, tlp_color):
...
         else:
+            demisto.debug(f"FeedElasticSearch: {last_calculated_timestamp=}")
+            demisto.debug(f"FeedElasticSearch: {calculate_timestamp=}")
             break
-    if last_calculated_time is None:
-        last_calculated_time = datetime.now()
-    return last_calculated_time, last_ids
+    if last_calculated_timestamp is None:
+        last_calculated_timestamp = int(datetime.now().timestamp() * 1000)
+    demisto.debug(f"FeedElasticSearch: The length of the indicators of the last time: {len(last_ids)}")
```
yuvalbenshalom, line 231:
```suggestion
    demisto.info(f"FeedElasticSearch: The length of the indicators of the last time: {len(last_ids)}")
```

yuvalbenshalom (changes requested):
see my comment

Packs/FeedElasticsearch/Integrations/FeedElasticsearch/FeedElasticsearch.py
ilaner: It is the same behavior
The time is sorted so we should never encounter cases like this.
There is already a unit test where the last fetched ids have the same time

yuvalbenshalom (approved):

PR #32794, opened 2024-02-08
Fix type issue in CommandResult input tags

AradCarmi:
## Description
Fixed an issue where the **Tags** input for ***CommandResult*** was incorrectly set as a boolean instead of a list.


MichaelYochpaz (approved):
Nice 👍🏼

PR #32857, opened 2024-02-12
Splunkpy cache incidents by window

anas-yousef:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-32803

## Description
Fix an issue where fetch incidents would fetch the same incident twice if we stay in the fetch window for a long period of time. We solve this by caching last fetched IDs with respect to their fetch window, and not by a specific timeline.

## Must have
- [ ] Tests
- [ ] Documentation


JasBeilin (approved):
NICE

PR #32871, opened 2024-02-13
Symantec Collector

jlevypaloalto:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-8703)

## Description
Create the **Symantec Cloud Workload Protection Event Collector**

## Must have
- [ ] Tests
- [ ] Documentation


JudahSchwartz (commented):
looks very good! We need to add a ton of logs, wayt more than you think are neccessary, at least while it is under development and not in our environment. We should not be debugging in customer environments as much as possible.

PR #32875, opened 2024-02-13
Return XML Object

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready

## Description
When `BaseClient._http_request` is called with `resp_type='xml'`, an XML object is created but not returned. This PR fixes this issue.

## Must have
- [ ] Tests
- [ ] Documentation


jlevypaloalto:
This breaks BC with [FortiSIEMV2.py](https://github.com/demisto/content/blob/b3627fbe5e2df3cde6bb340668207863a09c210b/Packs/FortiSIEM/Integrations/FortiSIEMV2/FortiSIEMV2.py), aborting.

PR #32962, opened 2024-02-15
Fix AWSRecreateSG EC2 breaking change bug

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-33573)

## Description
AWSRecreateSG fails after the recent changes to `AWS - EC2` [(version 1.3.0)](https://github.com/demisto/content/pull/31416/)
The problems are caused by the change in the object returned when `AWS - EC2` commands are called using the `demisto.executeCommand` function in the script . This PR uses a new path to extract the required data correctly.

## Must have
- [ ] Tests
- [ ] Documentation


johnnywilkes (approved):
tested in XSIAM and XSOAR.  This makes the script no longer backwards compatible, but I don't see this as a concern

JudahSchwartz (approved):

Packs/AWS-Enrichment-Remediation/Scripts/AWSRecreateSG/AWSRecreateSG.py
```diff
@@ -86,13 +86,6 @@ def sg_fix(sg_info: list, port: int, protocol: str, assume_role: str, instance_t
                     and rule['IpProtocol'] == protocol
                 ):
                     change = True
-                elif (
```
johnnywilkes, line 112: this is not a duplicate, please don't remove

johnnywilkes (changes requested):
please re-add this block

Packs/AWS-Enrichment-Remediation/Scripts/AWSRecreateSG/AWSRecreateSG.py
johnnywilkes: This elif is when the offending rule references the port exactly.  The next elif is when the port i part of a range the therefore we need to run it past the split_rule() function.

jlevypaloalto: The `elif` block I removed is a duplicate of the `if` block before it not the `elif` block after it.

johnnywilkes: you are right, good catch

johnnywilkes (approved):

PR #32990, opened 2024-02-19
ServiceNowV2: fix bug in upload-file command

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-33532

## Description
_Problem_:
In the `upload-file` command there is redundant `copy` to copy the file from fileEntry path to local 
this cause an exception when the file name contain invalid characters (especial `/`).
_Fix_:
Remove the redundant code that cause the bug.


DeanArbel (approved):
Good job

PR #32991, opened 2024-02-19
fix a bug in fetch-indicators 

maimorag:
## Status
- [x] In Progress

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-33366)

## Description
Fixing a bug where fetch retrieves the feed all indicators are processed as new.


Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -512,19 +545,26 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
...
     search_query_per_page = client.search_query(params_dict)
+    last_timestamp = ''
     while len(search_query_per_page.get("response", {}).get("Attribute", [])):
         demisto.debug(f'search_query_per_page number of attributes:\
                       {len(search_query_per_page.get("response", {}).get("Attribute", []))}\
                         page: {params_dict["page"]}')
         indicators = build_indicators(search_query_per_page, attribute_types, tlp_color, params.get('url'), reputation, feed_tags)
         demisto.createIndicators(indicators)
         params_dict['page'] += 1
+        last_timestamp = search_query_per_page['response']['Attribute'][-1]['timestamp']
```
eyalpalo, line 561: what if the response is empty?

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -259,6 +262,36 @@ def get_ip_type(ip_attribute: Dict[str, Any]) -> str:
     return FeedIndicatorType.ip_to_indicator_type(ip_attribute['value'])
 
 
+def update_indicators_iterator(indicators_iterator: List[Dict[str, Any]],
```
eyalpalo, line 265: Lets remove this method if we dont use it.

eyalpalo (changes requested):

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -210,7 +210,8 @@ def handle_file_type_fields(raw_type: str, indicator_obj: Dict[str, Any]) -> Non
     indicator_obj['fields'][raw_type.upper()] = hash_value
 
 
-def build_params_dict(tags: List[str], attribute_type: List[str], limit: int, page: int) -> Dict[str, Any]:
+def build_params_dict(tags: List[str], attribute_type: List[str], limit: int, page: int, from_timestamp: str | None = ''
```
eyalpalo, line 213: How can it be none? if the last_run does not contain the timestamp then we assigning empty string to it (not None)

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -506,25 +520,33 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
...
-                      {len(search_query_per_page.get("response", {}).get("Attribute", []))}\
-                        page: {params_dict["page"]}')
+                      {len(search_query_per_page.get("response", {}).get("Attribute", []))} page: {params_dict["page"]}')
         indicators = build_indicators(search_query_per_page, attribute_types, tlp_color, params.get('url'), reputation, feed_tags)
-        demisto.createIndicators(indicators)
+        if LIMIT > 2000:
+            for iter_ in batch(indicators, batch_size=2000):
+                demisto.createIndicators(iter_)
+        else:
+            demisto.createIndicators(indicators)
```
dantavori, line 543:
```suggestion
        for iter_ in batch(indicators, batch_size=2000):
            demisto.createIndicators(iter_)
```
if LIMIT <= 2000 then the first batch will be the only one.
if you do it because LIMIT might change in the `clean_user_query()` method, see my note about this method below.

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -506,25 +520,33 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
     Returns: List of indicators.
 
     """
+    global LIMIT
```
dantavori, line 523:
```suggestion
```
you do this line only if you want to set the global.
also, instead of passing it to other methods - you can just use it in them directly

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -506,25 +520,33 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
...
     tlp_color = params.get('tlp_color')
     reputation = params.get('feedReputation')
     tags = argToList(params.get('attribute_tags', ''))
     feed_tags = argToList(params.get("feedTags", []))
     attribute_types = argToList(params.get('attribute_types', ''))
     query = params.get('query', None)
-    params_dict = clean_user_query(query) if query else build_params_dict(tags=tags, attribute_type=attribute_types, limit=2000,
-                                                                          page=1)
+    last_run = demisto.getLastRun().get('timestamp') if demisto.getLastRun() else ''
```
dantavori, line 530:
```suggestion
    last_run = demisto.getLastRun().get('timestamp') or ""
```

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -506,25 +520,33 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
...
+            for iter_ in batch(indicators, batch_size=2000):
+                demisto.createIndicators(iter_)
+        else:
+            demisto.createIndicators(indicators)
         params_dict['page'] += 1
+        last_run = search_query_per_page['response']['Attribute'][-1]['timestamp']
         search_query_per_page = client.search_query(params_dict)
     if error_message := search_query_per_page.get('Error'):
         raise DemistoException(f"Error in API call - check the input parameters and the API Key. Error: {error_message}")
+    demisto.setLastRun({'timestamp': last_run, 'params': params_dict})
```
dantavori, line 549: you are storing "params" but never use it, no?

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -506,25 +520,33 @@ def fetch_attributes_command(client: Client, params: Dict[str, str]):
...
+                      {len(search_query_per_page.get("response", {}).get("Attribute", []))} page: {params_dict["page"]}')
         indicators = build_indicators(search_query_per_page, attribute_types, tlp_color, params.get('url'), reputation, feed_tags)
-        demisto.createIndicators(indicators)
+        if LIMIT > 2000:
+            for iter_ in batch(indicators, batch_size=2000):
+                demisto.createIndicators(iter_)
+        else:
+            demisto.createIndicators(indicators)
         params_dict['page'] += 1
+        last_run = search_query_per_page['response']['Attribute'][-1]['timestamp']
```
dantavori, line 545: is it guaranteed that response.Attribute exists under search_query_per_page and non empty?

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -229,20 +232,31 @@ def build_params_dict(tags: List[str], attribute_type: List[str], limit: int, pa
...
 
 
-def clean_user_query(query: str) -> Dict[str, Any]:
+def clean_user_query(query: str, limit: int, page: int = 1, from_timestamp: str | None = None) -> Dict[str, Any]:
     """
     Takes the query string created by the user, adds necessary argument and removes unnecessary arguments
     Args:
         query: User's query string
     Returns: Dict which has only needed arguments to be sent to MISP
     """
```
dantavori, line 246: this is no longer the purpose of the method. If you keep it this way, need to rename it and update the docstring.

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
```diff
@@ -229,20 +232,31 @@ def build_params_dict(tags: List[str], attribute_type: List[str], limit: int, pa
...
     try:
         params = json.loads(query)
         params["returnFormat"] = "json"
         params.pop("timestamp", None)
+        if 'page' not in params:
+            params["page"] = page
+        if 'limit' not in params:
+            params["limit"] = limit
+        else:
+            LIMIT = params["limit"]
```
dantavori, line 257:
```suggestion
        params["limit"] = params.get("limit") or LIMIT
```
also I didn't understand why we set the LIMIT global (note that it's not a best practice to do it, usually globals are meant to be constants. someone else who reads might miss this spot and when he sees later in the code the usage of LIMIT, he will assume it's 2000 (it just happened to me).

dantavori (dismissed):

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
maimorag: no, it was in the original fetch- not sure if to keep it or delete it. Any thoughts?

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
maimorag: yes if it doesn't exist it exits the while loop

Packs/FeedMISP/Integrations/FeedMISP/FeedMISP.py
maimorag: after consult it was decided to remove it

PR #33065, opened 2024-02-21
ServiceNow mirror: fix bug when mirror not started

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-33571

## Description
In case of [trigger-incident-is-not-servicenow](https://xsoar.pan.dev/docs/reference/integrations/service-now-v2#configure-incident-mirroring-when-the-trigger-incident-is-not-servicenow)
there is cases when we create the SNOW ticket and after that set the XSOAR incident mirror related fields
this cause the incident to be updated then the SNOW ticket, and `get-remote-date` to return empty

_**Fix:**_
check if `isFetch` is False (in `rigger-incident-is-not-servicenow` it will be False and then return the remote data even the XSOAR incident are updated then the SNOW ticket,

### DIFF
--- a/Packs/ServiceNow/ReleaseNotes/2_5_55.md
+++ b/Packs/ServiceNow/ReleaseNotes/2_5_55.md
@@ -1,0 +1,6 @@
+
+#### Integrations
+
+##### ServiceNow v2
+
+- Fixed an issue where mirroring did not work in case mirror fields set direct after ServiceNow ticket is created.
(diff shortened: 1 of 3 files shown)


tkatzir (approved):
Approved.
Before merging, please wait for @DeanArbel 's approval as well.

Packs/ServiceNow/ReleaseNotes/2_5_55.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### ServiceNow v2
+
+- Fixed an issue where mirroring did not work in case mirror fields set direct after ServiceNow ticket is created.
```
DeanArbel, line 6:
```suggestion
- Fixed an issue where mirroring did not update fields that were updated before the creation time of the incident.
```

DeanArbel (approved):

PR #33080, opened 2024-02-22
[Native Image] Update native image tag

samuelFain:
## Description
Update native image tag to `demisto/py3-native:8.6.0.88298`.
After support for `openpyxl` and `python-pptx` packages was added in this PR https://github.com/demisto/dockerfiles/pull/24500, we're upgrading the native image reference to the latest version.


GuyAfik (approved):
@samuelFain Please update the PR description with the reason that we're updating the production image

PR #33134, opened 2024-02-28
fix failing paths

dorschw:
Fixed paths that would fail on (soon to be introduced) [path validations](https://github.com/demisto/demisto-sdk/pull/4084/files)

**Content behaviors are not affected.**


dorschw:
Will be done gradually

PR #33222, opened 2024-03-05
Formatter fix

Ni-Knight:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-34063

## Description
The PR fixes an edge case in which the regex captures a CIDR with an extra character (a valid url character). The formatter will try and strip the extra char and check again if the input is a valid CIDR.


DeanArbel (approved):
Looks good

PR #33225, opened 2024-03-05
Add CertStream Integration Pack 

efelmandar:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
This is a new Pack for the CertStream integration

## Must have
- [x] Tests
- [x] Documentation


Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+    Args:
+        host (str): The URL for the websocket connection
+        fetch_interval (int): The interval in minutes to check for updates to the homographs list
+    """
+
+    while True:
+        with websocket_connections(host) as (message_connection):
+            demisto.info("Connected to websocket")
+
+            while True:
```
JudahSchwartz, line 51: why do we need this inner while true?

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+                now = datetime.now()
+                context = json.loads(demisto.getIntegrationContext()["context"])
+                last_fetch_time = datetime.strptime(context["fetch_time"], '%Y-%m-%d %H:%M:%S')
```
JudahSchwartz, line 54: move` '%Y-%m-%d %H:%M:%S'` too a constant at the top of the file

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+                if now - last_fetch_time >= timedelta(minutes=fetch_interval):
+                    context["homographs"] = get_homographs_list(context["word_list_name"])
+                    context["fetch_time"] = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+                    demisto.setIntegrationContext({"context": json.dumps(context)})
+
+                try:
+                    message = message_connection.recv()
+                    fetch_certificates(message)
+
+                except ConnectionClosed:
```
JudahSchwartz, line 66: log here

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+    """ Executes to long running loop and checks if an update to the homographs is needed
+
+    Args:
+        host (str): The URL for the websocket connection
+        fetch_interval (int): The interval in minutes to check for updates to the homographs list
+    """
+
+    while True:
+        with websocket_connections(host) as (message_connection):
+            demisto.info("Connected to websocket")
```
JudahSchwartz, line 49: are there any more details from message_connection that we can log?

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+
+            while True:
+                now = datetime.now()
+                context = json.loads(demisto.getIntegrationContext()["context"])
+                last_fetch_time = datetime.strptime(context["fetch_time"], '%Y-%m-%d %H:%M:%S')
+                fetch_interval = context["list_update_interval"]
+
+                if now - last_fetch_time >= timedelta(minutes=fetch_interval):
```
JudahSchwartz, line 57: add logs in the if and else to follow the logic

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+                    context["homographs"] = get_homographs_list(context["word_list_name"])
+                    context["fetch_time"] = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+                    demisto.setIntegrationContext({"context": json.dumps(context)})
```
JudahSchwartz, line 60: use `set_to_integration_context_with_retries` from `commonserverpython`

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+def main():  # pragma: no cover
+    command = demisto.command()
+    params = demisto.params()
+    host: str = params["url"]
+    word_list_name: str = params["list_name"]
+    list_update_interval: int = params.get("update_interval", 30)
+    levenshtein_distance_threshold: float = float(params.get("levenshtein_distance_threshold", 0.85))
+    logging.getLogger('websockets.client').setLevel(logging.ERROR)
+
+    demisto.setIntegrationContext({"context": json.dumps({
```
JudahSchwartz, line 334: lets discuss this in demo, what are we setting to the context here?
note, in case of container crash/disable enable, this will be run again, is this wanted behavior?

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+        "word_list_name": word_list_name,
+        "list_update_interval": list_update_interval,
+        "levenshtein_distance_threshold": levenshtein_distance_threshold,
```
JudahSchwartz, line 337: levenshtein_distance_threshold, word_list_name is only set by param right? Why store this in the context?

Packs/CertStream/Integrations/CertStream/CertStream.py
```diff
@@ -0,0 +1,354 @@
...
+            continue
+
+        else:
+            return json.loads(user_list["data"])
+
+    demisto.error("List of words not found")
+    raise OSError
+
+
+def levenshtein_distance(original_string: str, reference_string: str) -> float:
```
JudahSchwartz, line 285: this function needs about 300 tests (if we dont want to use a library for this)

Packs/CertStream/Integrations/CertStream/CertStream.py
JudahSchwartz, line 57: `   if now - last_fetch_time >= timedelta(minutes=fetch_interval):`
the server should only call after the fetch interval, even if not, why do we need this check?

JudahSchwartz (commented):
looks great! nice code. Lets do a demo before a final review

Packs/CertStream/Integrations/CertStream/CertStream.py
Ni-Knight: Its just the Levenshtein algorithm in Python, we can use a library but that would make us need a different docker. We can add one or two tests with known distances as a safety.

Packs/CertStream/Integrations/CertStream/CertStream.py
Ni-Knight: This allows the user to update the list.

Packs/CertStream/Integrations/CertStream/CertStream.py
efelmandar: I got an error when I tried to use `set_to_integration_context_with_retries()` are you sure it needs to be changed to that?

JudahSchwartz: its better in case theres any flakiness from the server, what error did you get? Thus might be a moot point since not sure context should be used at all

JudahSchwartz (approved):
last point is to switch to globals like we discussed, great work!

efelmandar:
Thank you for your review, I finished making the changes please review again 🙏

efelmandar:
A force merge is needed due to the reason that validation is wrongly failing for less than 70% unit tests coverage which is not mandatory for a Community Support Level pack

PR #33295, opened 2024-03-11
Rasterize Fix Arrays Support

tkatzir:
## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-34587

PR #33301, opened 2024-03-11
[Rasterize] Fix a bug where input doesn't support array

pl-brault:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
When executing the *rasterize* command with an array of url and an array of file_name so that the filename matches the url to be screenshot the command isn't working properly.
The two input are not the same type: 
- the input *url* the type is  *isArray:True* 
- the input *file_name* the type is *isArray:False*

Therefore the command was launch multiple time for each file_name.

I fix the issue by removing the *isArray* on the input *url*, so now both inputs (url and file_name) for the command *rasterize* have the same type. If there is multiple inputs (an array of url is passed as the input of the command) the command is launch for each url and behave correctly.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/rasterize/Integrations/rasterize/rasterize.py
+++ b/Packs/rasterize/Integrations/rasterize/rasterize.py
@@ -658,23 +658,21 @@ def perform_rasterize(path: str,
     if browser:
         support_multithreading()
         with ThreadPoolExecutor(max_workers=MAX_CHROME_TABS_COUNT) as executor:
-            paths = argToList(path)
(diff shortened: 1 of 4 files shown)


jbabazadeh:
Hi @pl-brault 
Thank you for opening a pull request. We already have a PR to Fix this issue [PR 33295](https://github.com/demisto/content/pull/33295) that will be merge soon.

PR #33343, opened 2024-03-13
[Rasterize] Added the possibility to pass an array of filename to match the URLs.

pl-brault:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Added the possibility to pass an array of filename to match the URLs.
The change is that both `file_name` and `url` support an array, and then first url matches first file_name, second url matches second file_name, etc...

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/rasterize/Integrations/rasterize/rasterize.py
+++ b/Packs/rasterize/Integrations/rasterize/rasterize.py
@@ -570,7 +570,7 @@ def screenshot_image(browser, tab, path, wait_time, navigation_timeout, full_scr
         else:
             demisto.info(f"Response Body not available available after {operation_time} seconds.")
 
-    return ret_value, response_body
+    return ret_value, response_body, path
 
 
 def screenshot_pdf(browser, tab, path, wait_time, navigation_timeout, include_url):  # pragma: no cover
@@ -658,13 +658,7 @@ def perform_rasterize(path: str,
     if browser:
         support_multithreading()
         with ThreadPoolExecutor(max_workers=MAX_CHROME_TABS_COUNT) as executor:
-            demisto.debug(f'path type is: {type(path)}')
-            if type(path) == list:
-                demisto.debug('path type is list')
-                paths = argToList(path)
-            else:
-                demisto.debug('path type is str')
-                paths = [path]
+            paths = argToList(path)
             demisto.debug(f"rasterize, {paths=}, {rasterize_type=}")
             rasterization_threads = []
             rasterization_results = []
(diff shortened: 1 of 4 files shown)


Packs/rasterize/Integrations/rasterize/rasterize.py
```diff
@@ -658,13 +658,7 @@ def perform_rasterize(path: str,
     if browser:
         support_multithreading()
         with ThreadPoolExecutor(max_workers=MAX_CHROME_TABS_COUNT) as executor:
-            demisto.debug(f'path type is: {type(path)}')
-            if type(path) == list:
-                demisto.debug('path type is list')
-                paths = argToList(path)
-            else:
-                demisto.debug('path type is str')
-                paths = [path]
```
tkatzir, line 667: This is important.
It supports paths that don't start with `http`.
Paths are allowed to contain. a comma.

Packs/rasterize/Integrations/rasterize/rasterize.py
```diff
@@ -866,25 +860,25 @@ def rasterize_command():  # pragma: no cover
     rasterize_type = RasterizeType(demisto.args().get('type', 'png').lower())
     wait_time = int(demisto.args().get('wait_time', 0))
     navigation_timeout = int(demisto.args().get('max_page_load_time', DEFAULT_PAGE_LOAD_TIME))
-    file_name = demisto.args().get('file_name', 'url')
+    file_names = argToList(demisto.args().get('file_name', 'url'))
```
tkatzir, line 863: Please add a debug log message that prints `len(file_names)`.

Packs/rasterize/Integrations/rasterize/rasterize.py
```diff
@@ -866,25 +860,25 @@ def rasterize_command():  # pragma: no cover
...
     if rasterize_type == RasterizeType.PDF or str(rasterize_type).lower == RasterizeType.PDF.value:
         file_extension = "pdf"
-    file_name = f'{file_name}.{file_extension}'  # type: ignore
 
     rasterize_output = perform_rasterize(path=url, rasterize_type=rasterize_type, wait_time=wait_time,
                                          navigation_timeout=navigation_timeout, include_url=include_url,
                                          full_screen=full_screen)
     demisto.debug(f"rasterize_command response, {rasterize_type=}, {len(rasterize_output)=}")
-    for current_rasterize_output in rasterize_output:
+    for current_rasterize_output, file_name in zip(rasterize_output, file_names):
```
tkatzir, line 874: Need to handle the case where `len(file_names) < len(paths)`
Note that you don't have `paths` isn't here...

Packs/rasterize/Integrations/rasterize/rasterize.py
```diff
@@ -866,25 +860,25 @@ def rasterize_command():  # pragma: no cover
...
         # demisto.debug(f"rasterize_command response, {current_rasterize_output=}")
+        file_name = f'{file_name}.{file_extension}'  # type: ignore
```
tkatzir, line 876: For better readibility
```suggestion
    for current_rasterize_output, current_file_name in zip(rasterize_output, file_names):
        # demisto.debug(f"rasterize_command response, {current_rasterize_output=}")
        file_name = f'{current_file_name}.{file_extension}'  # type: ignore
```

tkatzir (commented):
Looks good! Thank you for your contribution.
Added some comments and notes.

Packs/rasterize/Integrations/rasterize/rasterize.py
BestPig: The following code
```python
if type(path) == list:
```

is not working as expected, because the XSOAR never pass a list object as command argument.
When "Is array" is checked on an argument, it will have as effect to not loop on the integration
but the argument will still be a `str` for the command in all cases.
So checking that the type is a list is never true when it's called from a command or a playbook.
The only way to handle list correctly is by calling argToList, I understand your argument that path can contain comma,
but in this case you should pass your argument correctly as a way argToList can handle it like passing the str as a JSON list `["te,st1", "test,2"]`

The old code
```python
paths = [path]
```
is invalid, because passing a list like the XSOAR expect it `test,test2` or `["test1", "test2"]` will just broke as it will ends with a list `["test,test2"] and it's not what we are expecting.

tkatzir: You can pass a key from the Context that has a list value.
Then `type(path) == list` will be `True`.

BestPig: I forgot this case, but in that case calling argToList directly is not an issue, as it will just return the list, so everything is fine.

merit-maita (commented):
thank you for your Contribution!
please address @tkatzir comments, and let's set up a demo

Packs/rasterize/Integrations/rasterize/rasterize.py
tkatzir: Please check what will be the output of rasterizing this URL: https://www.google.com/maps/@48.8588254,2.2644638,12z

tkatzir:
@BestPig,
We will do it ourselves: https://github.com/demisto/content/pull/33406

@merit-maita FYI.

merit-maita:
@BestPig thank you for your Contribution! but as Rotem has mentioned we'll do it internally, so i'll close the pr.

PR #33390, opened 2024-03-17
AWS EC2 fixed an issue where roleArn in params was ignored

MosheEichler:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-33746

## Description
Fixed an issue where roleArn in params was ignored.

## Must have
- [ ] Tests
- [ ] Documentation


jlevypaloalto:
If this is the solution we should check if the other AWS integrations have the same problem and need the same treatment.

PR #33425, opened 2024-03-19
Gmail: fix fetch duplicate incident

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-34805

## Description
Due to a wrong condition, in case the latest occurred email was = to the `next_last_fetch` it's ignored
this lead to a case like following:
incident A occurred time `10:00:01`
incident B occurred time `10:00:02`
--> `next_last_fetch=10:00:02` instead `10:00:03`
and in the next fetch it was fetched again


DeanArbel (approved):
@ilappe please add a UT for this case

PR #33484, opened 2024-03-21
Hatching Triage 1.0.17 - Dbot Score - triage-get-sample-summary

mohlcyber:
## Status
- [ ] In Progress
- [x] Ready

## Description
- Added Dbot score calculation and output to **triage-get-sample-summary** command.
- Updated the Docker image to: *demisto/python3:3.10.14.90585*.
- Added new playbook **Detonate File - Hatching Triage** to analyze files

## Must have
- [x] Tests
- [x] Documentation


mohlcyber:
Hi Team, see that the pre-commit failed because of missing tests related to linting. I added new the HatchingTriage_test.py new to the commit as this integration didn't had any testings and i added a few more. Could you please advise how best to proceed. Or should i remove the HatchingTriage_test.py?

melamedbn:
Hi @mohlcyber,

Thanks for contributing to Cortex XSOAR! I've reviewed your contribution and would be happy if you could clarify why you executed task 9, which is the same as task 7 before the generic polling command.

Besides that, the generic polling command should also make task 8 (sleep) irrelevant since the generic polling command executes that exact logic.

Best regards,
Ben

moishce:
Hi @mohlcyber, we haven’t heard from you in a while.
Did you understand Ben's questions?
Do you need any help with the pull request?

Please feel free to reach out to me here or on Slack.

mohlcyber:
Hi Guys, sorry for the delayed response. I aligned the playbook to the already existing Playbook (called Detonate URL - Hatching Triage) that is part of the content pack. Basically this is a one to one copy just aligned for file submissions. Fully agree that this might not be needed but don't have the context why it was done in the first instance inside the playbook in the content pack. Happy to adjust or do we want to get the feedback from the Hatching Triage team first?

ssokolovich:
> Hi Guys, sorry for the delayed response. I aligned the playbook to the already existing Playbook (called Detonate URL - Hatching Triage) that is part of the content pack. Basically this is a one to one copy just aligned for file submissions. Fully agree that this might not be needed but don't have the context why it was done in the first instance inside the playbook in the content pack. Happy to adjust or do we want to get the feedback from the Hatching Triage team first?

Hey @mohlcyber,
Since no one responded here, I'm assuming that nothing was decided about your last question. 
So let me put out the options and let me know your thoughts:

### Option 1:
leave it as it is (similar to the other playbook that you had).
### Option 2:
Make a minor change by removing tasks # 8 and # 9. This flow will get you straight to the generic polling, where the sleep logic is happening OOTB. **This will reduce the playbook size and also the investigation.**
### Option 3:
Make this entire playbook a command with the polling function.[ Here is an example from our repo where submission was done with a polling command.](https://github.com/demisto/content/blob/a715596158c4ce9bfab26d46f4818fcbf3d11fd8/Packs/CrowdStrikeFalconX/Integrations/CrowdStrikeFalconX/CrowdStrikeFalconX.py#L1471)
In this option, this playbook is just a command that everyone can use wherever they like in the system :) 


TBH, I prefer options 2/3 as they are better in my opinion. Waiting for your feedback 🙏 
Cheers!

ssokolovich (commented):
Hey @mohlcyber,
Still waiting for your response to my last comment above ⬆️. Let me know if you require any clarification from our end.
Cheers!

mohlcyber:
Hi @ssokolovich,

thanks for taking a look. Fully agree Option 2 and 3 would be most preferred. Unfortunatley i don't have access to Hatching Triage to test Option 2 and 3 anymore. If we don't get any response from Hatching Triage can we just go ahead with the PR for now until we have another opportunity to extend the integration with the scheduling command when we have access to the solution?

Many thanks and all the best,
Martin

PR #33582, opened 2024-03-27
remove tmp file from repo

ilaner:
Remove temp file from repo which cause the pytest for StringSifter to fail


GuyAfik (approved):

kobymeir:
force merge: accidental file added

PR #33637, opened 2024-03-28
Gitlab v2.2.22 - Argument Description Change

mohlcyber:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
Argument Description for gitlab-file-get and gitlab-raw-file-get misleading

## Description
- Updated argument description in ***gitlab-file-get***
- Updated argmuent description in ***gitlab-raw-file-get***

## Must have
- [x] Tests
- [x] Documentation


anas-yousef:
@mohlcyber
Thank you so much for your contribution!
Nice catch. I believe it is better to do the URL encoding in the code, and not as part of the argument value.
I updated the PR to reflect this behavior. Update me if you feel anything is off, otherwise, we can go ahead and push these changes.

mohlcyber:
> @mohlcyber Thank you so much for your contribution! Nice catch. I believe it is better to do the URL encoding in the code, and not as part of the argument value. I updated the PR to reflect this behavior. Update me if you feel anything is off, otherwise, we can go ahead and push these changes.

Looks amazing man. Thanks so much. Much more elegant solution.

PR #33654, opened 2024-03-31
[PrismaCloud v2] Issue With Docker Timeout Instead `urlib` Timeout

shmuel44:
## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CRTX-109102)

## Description
Fixed an issue where the ***test button*** failed on container timeout when server url is not reachable.

When using the test module, the retries mechanism is now disabled.
This means that the API will only be called once.


Packs/PrismaCloud/ReleaseNotes/4_3_1.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Prisma Cloud v2
+
+Fixed an issue where the ***test module*** failed on Docker timeout when the url is not reachable.
```
yuvalbenshalom, line 6:
```suggestion
Fixed an issue where the ***test module*** failed on container timeout when server url is not reachable.
```

yuvalbenshalom (changes requested):
see my comment

Packs/PrismaCloud/ReleaseNotes/4_3_1.md
shmuel44: I also changed from `test module` to `test button` (validate BA-125)

yuvalbenshalom (approved):

PR #33706, opened 2024-04-01
Path validations: fix `InvalidDepthZeroFile`

dorschw:
related: https://github.com/demisto/demisto-sdk/pull/4084

The only change that requires an update is in the `Druva` pack (change the existing, commit-locked image url), will be done in a separate PR, together with the required release notes. All other changes do not require release notes.


dorschw:
@yuvalbenshalom please force-merge, only failure is BA115 (deleted files) as we move them to their correct place.

yuvalbenshalom (approved):

PR #33776, opened 2024-04-07
Pre-Commit gitlab - update graph instead of create in nightly

ilaner:
In the pre-commit -a in gitlab we first set up the graph and then run the command, which will set up a new graph if necessary. We need to unset the DEMISTO_SDK_GRAPH_FORCE_CREATE variable so we will use the created graph


adi88d (approved):
Nice!

PR #33804, opened 2024-04-08
[pre-commit] Exclude coverage for community support level

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Description
Exclude coverage for community support level


mmhw:
This is the behavior in `lint` too so it works as expected.

PR #33807, opened 2024-04-08
SkyHigh Security Bug Fix

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-35710

## Description
- Changed the response type to `response`. 
- Hadled the case where the response returns 200 OK with no data.
- Hadled the case where the response returns data.

## Must have
- [x] Tests


Packs/SkyhighSecurity/Integrations/SkyhighSecurity/SkyhighSecurity_test.py
```diff
@@ -57,26 +61,39 @@ def test_status_update_command(mocker):
...
-
     with requests_mock.Mocker() as m:
-        m.post('https://www.example.com/shnapi/rest/external/api/v1/queryActivities', json={})
+        m.post('https://www.example.com/shnapi/rest/external/api/v1/queryActivities', status_code=200)
         main()
+    assert response.call_args[0][0].get('HumanReadable') == 'No activities found for anomaly ID 1111'
 
-    assert response.call_args[0][0].get('HumanReadable') == '**No entries.**\n'
+
+def test_anomaly_activity_list_command_with_response(mocker):
```
ShahafBenYakir, line 85: Add docstring

Packs/SkyhighSecurity/Integrations/SkyhighSecurity/SkyhighSecurity.py
```diff
@@ -28,6 +29,18 @@
...
+    Args:
+        csv_data: data in csv format
+    Returns:
+        the same data in json formal
+    """
+    reader = csv.DictReader(io.StringIO(csv_data))
+    json_data = list(reader)
+    return json_data
+
+
```
ShahafBenYakir, line 43: It looks like instead of a normal JSON parser we had until now, we are changing to using csv.DictReader, this means the command never worked?

ShahafBenYakir (commented):
Good job!
- Can we add a test function to csv2json specifically?
- I see the customer has not yet replied the fix is working, you don't want to wait before merging?

Packs/SkyhighSecurity/Integrations/SkyhighSecurity/SkyhighSecurity.py
Shellyber: Yes, never worked before.
The response headers are in text/csv.
Not sure how it wasn't caught before the first merge

Shellyber:
@ShahafBenYakir I'll add the test to the CSV function
Also will send this latest version to the customer to test.
The previous integration fix was only for the byte error. 
After some digging and testing, I saw that it would not work for a regular response.
Will send this version as well for confirmation

Packs/SkyhighSecurity/Integrations/SkyhighSecurity/SkyhighSecurity.py
ShahafBenYakir: @MosheEichler FYI

MosheEichler: As @Shellyber says, the response appears to be in CSV format. However, the function may work when there are responses, but in this case, it appears that there are none (content length is 0).

ShahafBenYakir (approved):

PR #33833, opened 2024-04-08
Changes related to the Securonix Release: 2.0.20

scnx-knathani:

## Status
- [ ] In Progress
- [x] Ready

#### Updates in the Securonix Integrations

- Updated the ***securonix-list-activity-data*** and ***securonix-list-violation-data*** commands for the improved *query* argument by adding the support for the escape characters "\", "*", and "?" when passing them to the API.
- Added a new argument "policy_type" to the ***securonix-list-violation-data*** command.
- Updated the retry mechanism for the ***securonix-list-violation-data*** command to dynamically set the "to" argument to the current time based on the per-specified policy types.
- Updated the ***securonix-list-activity-data*** command to support the optional index key in the query argument.
- Fixed an issue with the handling of special characters in the HR (Human Readable) for the commands.
- Added the **Securonix Policy Type** incident field.

## Must have
- [x] Tests
- [x] Documentation


ssokolovich (commented):
Hey @scnx-knathani 

Thanks for your contribution. Everything looks great from my perspective.
The only thing that bothers me is that the new incident field holds the `shortText` type when one field above in the layout is a `multi-select` and refers to policies. 
So my question here is it possible that the field might contain more than one type if there are multiple policies? if that isn't the intent -> maybe we should change the field to be multi-select and the name to be "Securonix Policy Type**s**"

Moreover, just a side note for you, my suggestion is always to try using the existing fields (such as `Policy Type`). Just a thought for you :) 

image

scnx-knathani:
Hi @ssokolovich,

Thanks for reviewing the PR. Policy type is single value and this could be different for incidents based on the policy it's violated. For now we intend to use the custom field so it keep label in parity with other fields in the layout with Securonix prefix.

Please let us know your if you need anything else from our side!

> Hey @scnx-knathani
> 
> Thanks for your contribution. Everything looks great from my perspective. The only thing that bothers me is that the new incident field holds the `shortText` type when one field above in the layout is a `multi-select` and refers to policies. So my question here is it possible that the field might contain more than one type if there are multiple policies? if that isn't the intent -> maybe we should change the field to be multi-select and the name to be "Securonix Policy Type**s**"
> 
> Moreover, just a side note for you, my suggestion is always to try using the existing fields (such as `Policy Type`). Just a thought for you :)
> 
> image

ssokolovich:
> Thanks for reviewing the PR. Policy type is single value and this could be different for incidents based on the policy it's violated. For now we intend to use the custom field so it keep label in parity with other fields in the layout with Securonix prefix.
> 
> Please let us know your if you need anything else from our side!

As for the prefix, NP - this is up to you.
As for the type - I still want to check if there is a possibility of multiple policies being fetched within 1 incident. If so, there is a possibility that the policies might have different types - which means that we can get multiple values in a single short text field. 
If that's not the case, and indeed, for every incident we will see only 1 policy -> shortText makes sense.

scnx-knathani:
> > Thanks for reviewing the PR. Policy type is single value and this could be different for incidents based on the policy it's violated. For now we intend to use the custom field so it keep label in parity with other fields in the layout with Securonix prefix.
> > Please let us know your if you need anything else from our side!
> 
> Thanks for your feedback - @scnx-knathani ! As for the prefix, NP - this is up to you. As for the type - I still want to check if there is a possibility of multiple policies being fetched within 1 incident. If so, there is a possibility that the policies might have different types - which means that we can get multiple values in a single short text field. If that's not the case, and indeed, for every incident we will see only 1 policy -> shortText makes sense.

There won't be possibility to get multiple policies for one Incident and one policy will have only one type.

ssokolovich (approved):

Packs/Securonix/Integrations/Securonix/Securonix.py
```diff
@@ -464,6 +495,21 @@ def extract_closing_comments(activity_data: List[Dict[str, Any]], close_states_o
     return " | ".join(closing_comments)
 
 
+def escape_spotter_query(original_query: str) -> str:
```
mmhw, line 498: Please add unit test

Packs/Securonix/Integrations/Securonix/Securonix.yml
```diff
@@ -393,17 +396,28 @@ script:
...
     - description: End date/time for which to retrieve activity data (in the format MM/dd/yyyy HH:mm:ss).
       name: to
       required: true
-    - description: Free-text query, For example, query="resourcegroupname=WindowsSnare and policyname=Possible Privilege Escalation - Self Escalation".
+    - description: "Free-text query, For example, query=\"resourcegroupname=WindowsSnare and policyname=Possible Privilege Escalation - Self Escalation\".\nNote: Use the XSOAR automation browser to run the command if the \"query\" argument contains the special characters."
       name: query
     - description: Paginate next set of results.
       name: query_id
+    - description: "Provide the policy type for retrying if the violations are not found in the initial attempt. The types of policies that can be retried are: \"Land Speed\", \"TIER2\", \"DIRECTIVE\", \"BEACONING\"."
+      name: policy_type
```
mmhw, line 420: If it is a closed list, you can use predefined, as for example in line 907.

mmhw (commented):
Hi @scnx-knathani,
Good work :)

Please note my comments.

Thanks again

Packs/Securonix/Integrations/Securonix/Securonix.yml
scnx-knathani: It's not closed list. Right now retries needed for these 4 types, that's why we mentioned but in future it could be added for more policy types.

Packs/Securonix/Integrations/Securonix/Securonix.py
scnx-knathani: This function is covered as a part of the command unit test cases where we're providing query which will eventually fall to this function and validates it's behavior.

Packs/Securonix/Integrations/Securonix/Securonix.py
```diff
@@ -1605,11 +1655,36 @@ def list_activity_data(client: Client, args) -> Tuple[str, Dict, Dict]:
...
+                                     headers=headers, removeNull=True)
+
+    pagination_data = {
+        "totalDocuments": activity_data.get('totalDocuments'),
+        "message": activity_data.get('message'),
+        "queryId": activity_data.get('queryId')
+    }
+
+    entry_context = {
+        'Securonix.Activity(val.id === obj.id)': remove_empty_elements(pagination_data)
```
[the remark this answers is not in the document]
scnx-knathani: For this command, we have two outputs to be returned. One is for the activity data, and second one is for the pagination data for the response.
We want that there should be only one entry for the pagination related response. So, I have provided the `id` key which is not being used in the context for that specific path. Due to that, we want have duplicate pagination entry in the context.

mmhw (approved):

PR #33848, opened 2024-04-09
SplunkPy: raise error in case error msg return from Splunk 

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-35748

## Description
Splunk return status 200 even if there is error in the search query 
but the response contained error message in the response body

image
 

Fix:
raise error in case this error message is returned.


DeanArbel (approved):
Looks good

PR #33868, opened 2024-04-10
demisto/polyswarm:1.0.0.91366 | 0-100 | PR batch #1/1

samuelFain:
Auto updated docker tags for the following content items:
Packs/PolySwarm/Integrations/PolySwarmV2/PolySwarmV2.yml


tcarmeli1:
The integration is partner-supported and has not been changed for a few years. 
However, the API has changed. 
We need to decide whether to rewrite or deprecate the integration (?).

tcarmeli1:
Was merged directly to the main PR https://github.com/demisto/content/pull/34465

PR #33922, opened 2024-04-11
moved init of auth to startup instead of first request

JudahSchwartz:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-35918)

## Description
The initialization of what the auth header was done was only done in the first request, making the wrong header be used for the first request

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/GenericSQL/Integrations/GenericSQL/GenericSQL.py
+++ b/Packs/GenericSQL/Integrations/GenericSQL/GenericSQL.py
@@ -81,7 +81,7 @@ class Client:
             connect_parameters_dict['driver'] = 'FreeTDS'
             connect_parameters_dict.setdefault('autocommit', 'True')
         elif dialect == MS_ODBC_DRIVER:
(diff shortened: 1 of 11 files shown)


Packs/GenericWebhook/ReleaseNotes/1_0_29.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Generic Webhook
+
+- Fixed an issue where when the *username* parameter was set with `_header:customheader`, the first request after startup would use the *Authorization* header instead.
```
ShahafBenYakir, line 6:
```suggestion
- Fixed an issue where when the *username* parameter was set with `_header:customheader`, the first request after startup would use the `Authorization` header instead.
```

PR #33951, opened 2024-04-15
Fix 3.11 issues

dorschw:
Fixing various issues that would prevent using a Python3.11-based Native Image. 
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-9389


dorschw:
nightly: https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/959235

PR #34024, opened 2024-04-17
Pre-Commit: copy coverage report folder to artifacts in GitLab

ilaner:


.gitlab/ci/.gitlab-ci.global.yml
```diff
@@ -468,6 +468,7 @@
           fi
         fi
       fi
+    - cp -r coverage_report artifacts/coverage_report || true
```
ilappe, line 471: i think is better not to use the `|| true` as if there is any error we will see it here 
instead to fail on the `upload_code_coverage_report ` without details

ilappe (changes requested):

ilappe (approved):

kobymeir (approved):

PR #34086, opened 2024-04-22
Url enrichment

rundssoar:

## Status
- [ ] In Progress
- [x] Ready

## Description
Use the "Is not Empty" condition rather than the "isExists" condition when checking if there are URLs to analyze.
This resolves an issue, if the input is given an empty string, that the "rasterize" command would fail because it can't access an empty string.
This can happen, for example: If there are no URLs in the Context, but the "replace match" transformer is used. For some reason, the output then is an empty string, rather than nothing (as the input was)

There are no Changes in the Layout / other logic, only the Condition was replaced

## Must have
- [x] Tests
- [x] Documentation


ssokolovich (approved):

maimorag (approved):
Hi @rundssoar,
Good work :)

PR #34093, opened 2024-04-24
added setting to dateparser

yucohen:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-36389

## Description
On specific cases, where the ms in the time parsing is starting with many 0, the parser doesn't return the ms and therfore the splitting doesn't change the value, leading to the last_update_without_ms to contain the timezone offset. As we do not use this offset, removed it from the return value. 

failing test
image


yucohen:
Test Playbook - Cortex XDR Malware - Incident Enrichment https://jira-dc.paloaltonetworks.com/browse/CIAC-9931.
Test Playbook - Cortex XDR - Retrieve File by sha256 - https://jira-dc.paloaltonetworks.com/browse/CIAC-9932.

PR #34158, opened 2024-05-01
Add Zoom Mail integration

amshamah419:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10380

## Description
Phase One for Zoom Mail integration release

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1224 @@
...
+        return {
+            "success": False,
+            "error": "Failed to retrieve access token from ZoomMail API",
+        }
+
+    def _http_request(self, *args, **kwargs):
+        """
+        Override the _http_request method to include the access token in the headers.
+        """
+        if not self.access_token or (time.time() - self.token_time) >= 3500:
```
yuvalbenshalom, line 76: if access token is not saved to integration context, what of saving and checking its creation time?
BTW, I don't  think we should save it, unless there are relevant limitations in the docs.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.yml
```diff
@@ -0,0 +1,201 @@
...
+    required: false
+  - display: Fetch Mailbox
+    name: fetch_from
+    type: 0
+    required: false
+    additionalinfo: Mailbox to fetch incidents from
+  - display: Fetch Query
+    name: fetch_query
+    type: 0
+    required: false
```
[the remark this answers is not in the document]
amshamah419: Yes. This param just concats to the query we already send.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
amshamah419: I don't think it's being saved so much as we are holding it in memory as part of the ZoomMail class. I do think the time comparison we can remove however since we aren't storing it.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1224 @@
...
+
+    for msg in messages:
+        message_id = msg.get("id")
+        thread_id = msg.get("threadId", "")
+        message_details = client.get_email_message(
+            email=fetch_from, message_id=message_id
+        )
+        internal_date = float(message_details.get("internalDate")) / 1000.0
+
+        msg.update({"internalDate": internal_date})
```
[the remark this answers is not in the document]
amshamah419: Because we are looping off the list of fetched emails and then getting the internal time for each one. At the end of the run, I want to sort a list of message IDs and their create time so I can grab all emails at that end time.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1224 @@
...
+            account_id=account_id,
+            verify=verify_certificate,
+            proxy=proxy,
+            default_email=default_email,
+        )
+
+        COMMAND_FUNCTIONS: dict[str, Callable] = {
+            "fetch-incidents": lambda: fetch_incidents(client, params),
+            "test-module": lambda: the_testing_module(client, params),
+            f"{ZOOM_MAIL_COMMAND_PREFIX}-email-move-trash": lambda: trash_email_command(
```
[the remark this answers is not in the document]
amshamah419: It's needed to defer the invocation of the function until it's called, otherwise it'll evaluate when the dictionary is initialized.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
yuvalbenshalom: ok, so lets remove.
No variable shed be saved on the container between runs.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail_test.py
```diff
@@ -0,0 +1,1899 @@
...
+            "next_page_token": "",
+        }
+        mock_list_response = load_test_data("test_data/fetch/fetch_list_response.json")
+        self.client.list_emails = MagicMock(return_value=mock_list_response)
+
+        fetch_incidents(self.client, self.params)
+
+        incidents = mock_incidents.call_args[0][0]
+        assert len(incidents) == 2
+        assert incidents[0]["name"] == "Zoom Encrypted Email"
```
yuvalbenshalom, line 1017: is this really testing encrypted emails?

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1219 @@
...
+        )
+        internal_date = float(message_details.get("internalDate")) / 1000.0
+
+        msg.update({"internalDate": internal_date})
+
+        # Check if it's a new message, and if it's either all messages (fetch_threads is False) or only thread starters
+        if (internal_date > last_fetch_info["internalDate"] or (
+            internal_date == last_fetch_info["internalDate"] and message_id not in last_fetch_info["ids"]
+        )) and (not fetch_threads or message_id == thread_id):
+            incident = zoom_mail_to_incident(message_details, client, fetch_from)
```
yuvalbenshalom, line 416: the following case needs unit tests:
Deduplecation of ids that exist in the last_fetch_info["ids"]
filtering of threads when fetch_threads is true, not filtering threads when it is false.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1224 @@
...
+            new_processed_ids.add(message_id)
+            message_dates.append(internal_date)
+
+    if message_dates:
+        new_internal_date = max(message_dates)
+        new_ids = [
+            msg["id"] for msg in messages if msg["internalDate"] == new_internal_date
+        ]
+
+        last_fetch_info = {"internalDate": new_internal_date, "ids": new_ids}
```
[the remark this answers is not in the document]
yuvalbenshalom: this needs to be completed.

yuvalbenshalom (changes requested):
added some last comments

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
amshamah419: Actually re-reading this, I think it should stay. It's not being persisted to context or anything so the credentials are safe, just held in memory for the client. When the container dies, so does the client. If a call takes longer then the access token could theoretically expire before it's used again which could be useful for lists where the container is kept alive.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail_test.py
amshamah419: No, it's just based off the mock data I had at the time.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -417,32 +428,31 @@ def fetch_incidents(client: ZoomMailClient, params: dict[str, str]) -> None:
...
+            internal_date == last_fetch_info["internalDate"] and message_id not in last_fetch_info["ids"]
         )) and (fetch_threads or (not fetch_threads and message_id == thread_id)):
             incident = zoom_mail_to_incident(message_details, client, fetch_from)
             incidents.append(incident)
-            message_dates.append(internal_date)
+            if not last_page_fetch_token:
+                message_dates.append(internal_date)
 
-    if message_dates:
+    if message_dates and not next_page_token:
```
yuvalbenshalom, line 438: if multiple pages has messages of the same time, we need to save ids of the latest time from multiple pages, if time is inclusive.

this is a delicate case. I think you may need to save the local max time between pages. lets talk about this.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -417,32 +428,31 @@ def fetch_incidents(client: ZoomMailClient, params: dict[str, str]) -> None:
...
     # If we don't collect any new messages, but we have a next fetch token,
     # we are fetching old messages and need to stop.
-
     if len(incidents) == 0 and next_page_token:
         demisto.info("No new messages found, stopping fetch.")
         next_page_token = None
 
+    # If we dont have a token, we are fetching new messages and need to update the last fetch info
+    if next_page_token is None:
+        last_fetch_info["internalDate"] = now_epoch
```
yuvalbenshalom, line 454: I think that internalDate should never be now_epoch.
there are 3 cases for internalDate.
1. if it is the first run, use first run calculation
2. if there are new mail, take the last mail time make sure it is later then current internalDate
3. if there are no mails, preserve internalDate.

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1349 @@
...
+
+    demisto.info(f"Found {len(messages)} messages to process prior to filtering.")
+
+    for msg in messages:
+        message_id = msg.get("id")
+        thread_id = msg.get("threadId", "")
+        message_details = client.get_email_message(
+            email=fetch_from, message_id=message_id
+        )
+        internal_date = float(message_details.get("internalDate")) / 1000.0
```
yuvalbenshalom, line 425: is this still relevant?

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
```diff
@@ -0,0 +1,1349 @@
...
+            msg["id"] for msg in messages if msg["internalDate"] == new_internal_date
+        ]
+
+        last_fetch_info = {"internalDate": new_internal_date, "ids": new_ids}
+
+    # If we don't collect any new messages, but we have a next fetch token,
+    # we are fetching old messages and need to stop.
+    if len(incidents) == 0 and next_page_token:
+        demisto.info("No new messages found, stopping fetch.")
+        next_page_token = None
```
yuvalbenshalom, line 450: since time filtering works, we should go over all pages when they exist. and not to delete next_page_token. also I do not understand why it is relevant that no incident were found to delete the next page.

yuvalbenshalom (changes requested):
see my comments

Packs/ZoomMail/Integrations/ZoomMail/ZoomMail.py
yuvalbenshalom: on second glance you just need to explain here that pages ae arranged backwards by time, so if all messages has been filtered, new new messages will arrive in the next page because thy are older.

yuvalbenshalom (approved):

PR #34222, opened 2024-05-06
Jamf Pro Parsing Rule Update

eepstain:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-36721)

## Description

## Must have
- [ ] Tests
- [ ] Documentation


Packs/jamf/ParsingRules/jamf/jamf.xif
```diff
@@ -1,3 +1,3 @@
 [INGEST:vendor="jamf", product="pro", target_dataset="jamf_pro_raw", no_hit=keep]
-filter webhook ~= "\d{7,13}"
+filter webhook != null AND to_string(webhook) ~= "eventTimestamp[^\d]+\d{7,13}"
 | alter _time = to_timestamp(to_integer(json_extract_scalar(webhook, "$.eventTimestamp")), "millis");
```
cweltPA, line 3: The [JAMF Pro docs](https://developer.jamf.com/developer-guide/docs/webhooks) mention only that the timestamp is an epoch timestamp, but do not mention whether it is in given in seconds, milliseconds, or microseconds. Hence perhaps we shouldn't take for granted it is in milliseconds and determine the time unit explicitly, for e.g., 
```c
| alter tmp_eventTimestamp = webhook -> eventTimestamp
| filter to_string(tmp_eventTimestamp) ~= "\d{10}|\d{13}|\d{16}|\d{19}" 
| alter 
    tmp_epoch_int = to_integer(tmp_eventTimestamp),
    tmp_epoch_len = to_integer(len(tmp_eventTimestamp))
| alter tmp_time = if( 
    // in_case of nanoseconds, we convert them to microseconds since to_timestamp does not support nanoseconds
    tmp_epoch_len = 10, to_timestamp(tmp_epoch_int, "SECONDS"), 
    tmp_epoch_len = 13, to_timestamp(tmp_epoch_int, "MILLIS"),  
    tmp_epoch_len = 16, to_timestamp(tmp_epoch_int, "MICROS"),  
    tmp_epoch_len = 19, to_timestamp(to_integer(divide(tmp_epoch_int, 1000)), "MICROS")) 
```

Packs/jamf/ParsingRules/jamf/jamf.xif
```diff
@@ -1,3 +1,3 @@
 [INGEST:vendor="jamf", product="pro", target_dataset="jamf_pro_raw", no_hit=keep]
-filter webhook ~= "\d{7,13}"
+filter webhook != null AND to_string(webhook) ~= "eventTimestamp[^\d]+\d{7,13}"
```
cweltPA, line 2: maybe we could replace the _[^\d]_ with _\D_ or first try to extract the _eventTimestamp_ and apply the filter on that field directly - 
```c
| alter tmp_eventTimestamp = webhook -> eventTimestamp
| filter to_string(tmp_eventTimestamp) ~= "\d{10}|\d{13}|\d{16}|\d{19}" 
```

On any case I believe the epoch should be at least 10 digits long.

Packs/jamf/ReleaseNotes/2_2_1.md
```diff
@@ -0,0 +1,6 @@
+
+#### Parsing Rules
+
+##### Jamf Modeling Rule
+
+Updated the Parsing Rule logic to consider the webhook field with NULL values.
```
cweltPA, line 6: Seems the issue was the usage of the regex operator on the JSON object rather than having null values, 
So I would adjust the RN accordingly, perhaps something like: 
`Fixed an issue with the timestamp format validation. `

cweltPA (changes requested):

Packs/jamf/ParsingRules/jamf/jamf.xif
eepstain: Its based on the data we have, no changes will be conducted.
If its not mentioned, then its not worth the effort.
The complaint was on a parsing rule error, not an enhancment.

Packs/jamf/ReleaseNotes/2_2_1.md
eepstain: The regex received NULL, the issue wasn't with the regex...

Packs/jamf/ParsingRules/jamf/jamf.xif
eepstain: Another form of implementation, I am using less alter operation, so maybe this offer is less effective..
Not critical.

PR #34230, opened 2024-05-06
add argument to pre-commit validate hook

dorschw:
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-10404


dorschw:
[nightly](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/5847147)

PR #34323, opened 2024-05-12
Revert "[Marketplace Contribution] Common Dashboards - Content Pack Update"

maimorag:
Reverts demisto/content#34282


adi88d:
we need to revert those changes because we already have file with the name `UseCaseAdoptionMetrics.py`

PR #34392, opened 2024-05-16
Added functionality to YaraScan to handle multiple yara rules at once

ioannispanagopoulos:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
- Updated the YaraScan Script from the Yara Pack to be able to handle multiple yara rules at once following the format of the yara module, which is:
        yara.compile(sources={
              'namespace1':'rule dummy { condition: true }',
              'namespace2':'rule dummy { condition: false }'
        })

## Must have
- [ ] Tests
- [ ] Documentation


Ni-Knight (approved):

ioannispanagopoulos:
Hello @jlevypaloalto is there an ETA for this, so I can transfer that information to my team if possible of course?

Packs/Yara/Playbooks/playbook-YARA_-_File_Scan.yml
```diff
@@ -268,10 +268,10 @@ tasks:
           root: inputs.EntryID
           transformers:
           - operator: uniq
-      yaraRule:
+      yaraRules:
```
ssokolovich, line 271: Please note that changing this value, and not changing the YML - breaking the original logic.
Please revert those changes in PB.
Also, let's keep the original script argument name - so it won't break any logic for other users.

Packs/Yara/ReleaseNotes/1_0_6.md
```diff
@@ -0,0 +1,16 @@
+#### Scripts
+
+##### YaraScan
+
+- Updated the script to be able to handle multiple yara rules following the format of the yara module
```
ssokolovich, line 5: no need for a detailed description - please keep it high-level and more to the point.

ssokolovich (changes requested):

Packs/Yara/Playbooks/playbook-YARA_-_File_Scan.yml
```diff
@@ -271,7 +271,7 @@ tasks:
       yaraRule:
         complex:
           root: foundIndicators
-          accessor: rawrule
+          accessor: rawrules
```
ssokolovich, line 274: @Ni-Knight - Does the `FoundIndicators` path contain `rawrule` or `rawrules`?

Packs/Yara/Scripts/YaraScan/YaraScan.py
```diff
@@ -31,7 +31,7 @@ def main():
     if len(fileInfos) < 1:
         return_error('No files were found for scanning, please check the entry IDs')
 
-    yaraRuleRaw = args.get('yaraRule')
+    yaraRulesRaw = args.get('yaraRules')
```
ssokolovich, line 34: @Ni-Knight - Does the script argument name is `yaraRule` or `yaraRules`?

Packs/Yara/Scripts/YaraScan/YaraScan.py
```diff
@@ -31,7 +31,7 @@ def main():
     if len(fileInfos) < 1:
         return_error('No files were found for scanning, please check the entry IDs')
 
-    yaraRuleRaw = args.get('yaraRule')
```
jlevypaloalto, line 34: Let's not change argument names to keep backwards compatibility.

Packs/Yara/Scripts/YaraScan/YaraScan_test.py
```diff
@@ -2,22 +2,13 @@
...
 
 
 def test_main(mocker):
-    rule = '''rule PE_file_identifier
-{
-    meta:
-        author = "Adam Burt"
-        description = "Detects PE files"
-        date = "12/08/2016"
+    rule1 = "rule PE_file_identifier {meta: author = \"Adam Burt\" description = \"Detects PE files\" date = \"12/08/2016\" strings: $MZ = \"MZ\" ascii condition: $MZ at 0}"  # noqa: E501
```
jlevypaloalto, line 9: Let's add this as a new test case and leave the old one in place.

Packs/Yara/Scripts/YaraScan/YaraScan.py
```diff
@@ -42,12 +42,12 @@ def main():
                 "HasMatch": False,
                 "HasError": False,
                 "MatchCount": 0,
-                "Matches": list(),
-                "Errors": list()
+                "Matches": [],
+                "Errors": []
             }
             try:
-                cRule = yara.compile(source=yaraRuleRaw)
```
jlevypaloalto, line 49: Would this not break backwards compatibility?

jlevypaloalto (commented):
Hi @ioannispanagopoulos,
Good work :)

Packs/Yara/Scripts/YaraScan/YaraScan.py
ioannispanagopoulos: Yeah it would, because the yara module requires a dictionary instead of a string to handle multiple yara rules at once, as mentioned here https://yara.readthedocs.io/en/stable/yarapython.html. To keep backwards compatibility it would require bigger changes to the automation itself to handle both cases, where input is either a single or multiple yara rules. How would you prefer to going forward with this?

jlevypaloalto: Breaking backwards compatibility is not an option but bigger changes are welcome. Maybe you can add a new argument?

ioannispanagopoulos: @jlevypaloalto in this case then I would also need to change the yaraRule argument to optional, because it doesn't make sense to also provide a single rule if you want to provide multiple rules to begin with. I believe this option is optimal as it wouldn't break backwards compatibility. However there is the case where someone can run the automation without providing any rules, as both yaraRule and yaraRules will be optional arguments. Then in that case how would you prefer to handle this case?

jlevypaloalto: @ioannispanagopoulos the implementation is up to you. I'd prefer if you make the argument support both a single and multiple rules.

Packs/Yara/ReleaseNotes/1_0_6.md
```diff
@@ -0,0 +1,10 @@
+#### Scripts
+
+##### YaraScan
+
+- Updated the script to be able to handle multiple yara rules following the format of the yara module 
+
+#### Playbooks
```
ssokolovich, line 7:
```suggestion
```

Packs/Yara/ReleaseNotes/1_0_6.md
```diff
@@ -0,0 +1,10 @@
...
+
+##### YARA - File Scan
```
ssokolovich, line 9:
```suggestion
```

Packs/Yara/Scripts/YaraScan/YaraScan.py
ioannispanagopoulos: @jlevypaloalto Updated all files with the suggestions made and made it so the automation uses the yaraRuleRaw argument as it is with the added functionality for multiple rules. Let me know if there is any other change needed for this.

ioannispanagopoulos (commented):
Just noticed that the conflicts come from another commit done to the same Pack. However it doesn't seem to be merged yet as updating the branch didn't fix this, so can't resolve the conflict arising at the moment. If there is anything else needed please let me know.

ssokolovich:
@ioannispanagopoulos Thanks for updating us. I still can see that you need to resolve the Release notes thing. Just create a newer version of the Release Notes (with a higher ID like `1_0_7`) and push your update there (so the 1_0_6 can be pulled from the master).

As for my review, the PR looks good, and I don't have any comments for now. Please note that you still need approval from either @Ni-Knight or @jlevypaloalto

ioannispanagopoulos:
Hello @Ni-Knight @jlevypaloalto is there maybe an ETA for this, as I would like to provide it to my team?

Packs/Yara/Scripts/YaraScan/YaraScan.py
```diff
@@ -42,16 +42,29 @@ def main():
...
-            except Exception as err:
-                thisMatch['HasError'] = True
-                thisMatch['Errors'].append(str(err))
-                entries.append(thisMatch)
-                continue
+                rules = json.loads(yaraRuleRaw)
```
Ni-Knight, line 49: Why would a rule file be in JSON format and not YARA format?

YARA:
```
rule rule_1 : tag{ ... }

rule rule_2{...}
```

This isn't a JSON file so it will always throw an exception. The original script supports multiple rules in a single file. It does not support multiple sources. We can add that support by adding a new argument named `filepaths` or better yet `sources`. These should be multiple entry ids as we are running through XSOAR.

Ni-Knight: An entry ID is always a string, you need to open it within the script, see `line 20`. 

Either the script is called directly or via another script it has 2 inputs:
1. The rules.
  1a. A string of Rules (This is handled today via `yaraRule`)
  1b. A list of entry IDs which are yar files. (Not handled today)
2. Files to scan.

In the case of 1b the script should open the relevant files and compile them using the dictionary structure which should be built within the script.

PR #34430, opened 2024-05-19
AlienVault OTXv2 TPB: replace IP used in TPB

dorschw:
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10551


sapirshuker (approved):
Great!

PR #34551, opened 2024-05-27
[ASM]  EXPANDR-6928 Certificate changes

capanw:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
https://jira-dc.paloaltonetworks.com/browse/EXPANDR-6928

## Description
Adding Certificate enrichment by Venafi

## Must have
- [ ] Tests
- [ ] Documentation


johnnywilkes (changes requested):
@capanw , made some comments in the Jira ticket

Packs/Venafi/ReleaseNotes/2_0_1.md
```diff
@@ -0,0 +1,5 @@
+#### Integrations
```
johnnywilkes, line 1: I don't think you need release notes if just updating the marketplace

johnnywilkes (approved):
@capanw , thanks for the changes

capanw:
@Shellyber @ssokolovich, This is approved from our team. Please review. Thank you!!

Packs/CortexAttackSurfaceManagement/Playbooks/Cortex_ASM_-_Certificate_Enrichment.yml
```diff
@@ -0,0 +1,782 @@
...
+    isautoswitchedtoquietmode: false
+    isoversize: false
+    nexttasks:
+      '#none#':
+      - "9"
+    note: false
+    quietmode: 0
+    scriptarguments:
+      CN:
+        simple: ${inputs.Hostname}
```
ssokolovich, line 132: please use unique transformer here - in case an array was provided :)

Packs/CortexAttackSurfaceManagement/Playbooks/Cortex_ASM_-_Certificate_Enrichment.yml
```diff
@@ -0,0 +1,782 @@
...
+                - - left:
+                      iscontext: true
+                      value:
+                        simple: modules.state
+                    operator: isEqualString
+                    right:
+                      value:
+                        simple: active
+                root: modules
+          operator: isExists
```
ssokolovich, line 78: Usually, I will suggest using `Is Not Empty` - but if that works for you, then we should be fine.
The main reason for using the Is not Empty is to avoid situations when the key will be created but without a value.

ssokolovich (commented):
Looks good to me! I left some notes from my end that might improve the PB - but they are not mandatory. 
Cheers!

Packs/CortexAttackSurfaceManagement/Playbooks/Cortex_ASM_-_Certificate_Enrichment.yml
capanw: Hello @ssokolovich, we do not expect an array here, instead we always expect a unique single hostname, if its passed via inputs. Thank you.

ssokolovich: I know that you won't expect but sometimes the user might change that behavior :) up to you.

Packs/CortexAttackSurfaceManagement/Playbooks/Cortex_ASM_-_Certificate_Enrichment.yml
capanw: Thank you for the suggestion. We usually select "Is defined" as an operator in the playbook and it shows up as "isExists" in the yml file. 

But, I will bring this up with the team and probably start using this suggestions going forward.

Shellyber (approved):

PR #34564, opened 2024-05-27
Ciac 9706

edik24:
Fixing CIAC-9706

In the current flow when we check if tim reviewer is needed - we use graph functions for example:
BaseContent.from_path(CONTENT_PATH / file)

When a new pack contributed, the graph doesn't find it and throws an error, that stops the handle external pr.
I've added a check in the is_tim_content that if the pack is new (not found in master), I'll checkout the branch and check in the branch.
If the branch checkout is not successful (sometimes it's private and we don't have permissions to checkout) - the script won't ask review for tim (and if needed, should be asked manually).

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-9706)

## Description
Fixing CIAC-9706 in the handle external PR flow


Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,13 +255,56 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+
+def is_tim_content(pr_files: list[str], external_pr_branch: str, repo_name: str) -> bool:
     """
     This is where the actual search for feed:True or relevant tags or categories are being searched
     according to the login in is_tim_reviewer_needed
 
     Arguments:
     - pr_files: List[str] The list of files changed in the Pull Request.
+    - 'external_pr_branch': str : name of branch to checkout
+    - 'repo_name': str : name of repository
```
ShahafBenYakir, line 307: Same as in is_tim_reviewer_needed

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -268,6 +313,8 @@ def is_tim_content(pr_files: list[str]) -> bool:
         if 'CONTRIBUTORS.json' in file:
             continue
         integration = BaseContent.from_path(CONTENT_PATH / file)
```
ShahafBenYakir, line 315: let's change it to content_object since it is not an integration for sure
```suggestion
        content_object = BaseContent.from_path(CONTENT_PATH / file)
```

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -268,6 +313,8 @@ def is_tim_content(pr_files: list[str]) -> bool:
...
+        if not integration:
```
ShahafBenYakir, line 316:
```suggestion
        if not integration:
        # This means we were not able to find the file in content repo
```

Utils/github_workflow_scripts/utils.py
```diff
@@ -248,3 +250,24 @@ def get_doc_reviewer(content_roles: dict[str, Any]) -> str:
...
+    """
+    pack_metadata = {}
+
+    for pack_dir in pack_dirs:
+        if pack_metadata := get_pack_metadata(pack_dir):
+            print(f"pack metadata was retrieved for pack {pack_dir}")  # noqa: T201
+        else:
+            print(f'Could not find pack support level for pack {pack_dir}')  # noqa: T201
+
+    return pack_metadata
```
ShahafBenYakir, line 273: This should not return one pack dicy but all the pack dicts that were changed in this PR, maybe one has the TIM tag where the other does not.

Utils/github_workflow_scripts/handle_external_pr.py
edik24: I don't understand what should be the same? the description of the arguments is the same and the relevant arguments are listed.

What should I fix?

Utils/github_workflow_scripts/utils.py
edik24: agreed, added support form multiple packs

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -291,11 +338,13 @@ def is_tim_reviewer_needed(pr_files: list[str], support_label: str) -> bool:
     Arguments:
     - pr_files: tThe list of files changed in the Pull Request
     - support_label: the support label of the PR - the highest one.
+    - 'external_pr_branch': str : name of branch to checkout
+    - 'repo_name': str : name of repository
```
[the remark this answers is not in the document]
ShahafBenYakir: This was not posted for a reason and this is the answer to the information missing

ShahafBenYakir (commented):
Added missing comment

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
-    integrations_checked = []
-    for file in pr_files:
-        if 'CONTRIBUTORS.json' in file:
-            continue
-        integration = BaseContent.from_path(CONTENT_PATH / file)
-        if not isinstance(integration, Integration) or integration.path in integrations_checked:
-            continue
-        integrations_checked.append(integration.path)
-        if integration.is_feed:
+    print(f'check_if_pack_or_integration_is_feed - content object {content_object}')
```
ShahafBenYakir, line 267: remove

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+    if isinstance(content_object, Integration):
+        print(f'check_if_pack_or_integration_is_feed - content object {content_object}')
```
ShahafBenYakir, line 269: remove

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+        if content_object.is_feed:
             return True
-        pack = integration.in_pack
+    try:
+        pack = content_object.in_pack  # type: ignore
+        print(f' check_if_pack_or_integration_is_feed the pack is: {pack} ')
```
ShahafBenYakir, line 274: remove

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
         tags = pack.tags
         categories = pack.categories
         if TIM_TAGS in tags or TIM_CATEGORIES in categories:
             return True
+    except Exception as er:
+        print(f"The pack is not TIM: {er}")
```
ShahafBenYakir, line 280: What exceptions this is intended to catch?

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+        ):
+            for file in pr_files:
+                if 'CONTRIBUTORS.json' in file or 'Author_image' in file or 'README.md' in file or ".pack-ignore" in file:
+                    continue
+                content_object = BaseContent.from_path(CONTENT_PATH / file)
+                is_tim_needed = check_if_pack_or_integration_is_feed(content_object)
+                if is_tim_needed:
+                    return True
+    except Exception as er:
+        print(f"couldn't checkout branch to get metadata, error is {er}")
```
ShahafBenYakir, line 344: Why would this happen?

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
     return False
 
 
-def is_tim_content(pr_files: list[str]) -> bool:
+def check_if_pack_or_integration_is_feed(content_object: dict) -> bool:
```
ShahafBenYakir, line 258:
```suggestion
def check_if_item_is_tim(content_object: dict) -> bool:
```

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+    for file in pr_files:
+        if 'CONTRIBUTORS.json' in file or 'Author_image' in file or 'README.md' in file or ".pack-ignore" in file:
+            continue
+        content_object = BaseContent.from_path(CONTENT_PATH / file)
+        if not content_object:
+            # This means we were not able to find the file in content repo, and the contribution is new
+            print(f'for file {file}, the pack doesn\'t exist in Master and going to be searched in the remote branch')
+            return check_new_pack_metadata(pr_files, external_pr_branch, repo_name)
+        return check_if_pack_or_integration_is_feed(content_object)
     return False
```
ShahafBenYakir, line 372: It behaves the same but is a bit more clear
```suggestion
    is_tim = False
    for file in pr_files:
        if 'CONTRIBUTORS.json' in file or 'Author_image' in file or 'README.md' in file or ".pack-ignore" in file:
            continue
        content_object = BaseContent.from_path(CONTENT_PATH / file)
        if not content_object:
            # This means we were not able to find the file in content repo, and the contribution is new
            print(f'for file {file}, the pack doesn\'t exist in Master and going to be searched in the remote branch')
            is_tim = check_new_pack_metadata(pr_files, external_pr_branch, repo_name)
        else:
            is_tim = check_if_pack_or_integration_is_feed(content_object)
    return is_tim
```

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+    Returns: returns True or False if tim reviewer needed
+    """
+    for file in pr_files:
+        if 'CONTRIBUTORS.json' in file or 'Author_image' in file or 'README.md' in file or ".pack-ignore" in file:
+            continue
+        content_object = BaseContent.from_path(CONTENT_PATH / file)
+        if not content_object:
+            # This means we were not able to find the file in content repo, and the contribution is new
+            print(f'for file {file}, the pack doesn\'t exist in Master and going to be searched in the remote branch')
+            return check_new_pack_metadata(pr_files, external_pr_branch, repo_name)
```
ShahafBenYakir, line 370: Why aren't we always checking out the contributor repo? it is more accurate...
Also, you are calling check_new_pack_metadata for every file, this should not be the case it should only be called once...

Utils/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +255,124 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+                if 'CONTRIBUTORS.json' in file or 'Author_image' in file or 'README.md' in file or ".pack-ignore" in file:
+                    continue
+                content_object = BaseContent.from_path(CONTENT_PATH / file)
+                is_tim_needed = check_if_pack_or_integration_is_feed(content_object)
+                if is_tim_needed:
+                    return True
+    except Exception as er:
+        print(f"couldn't checkout branch to get metadata, error is {er}")
+        # if the checkout didn't work for any reason, will try to go over files manually
+        return check_files_of_pr_manually(pr_files)
```
ShahafBenYakir, line 346: unless we have a really good reason why a checkout wouldn't work in some cases I would like to remove this.

Utils/github_workflow_scripts/handle_external_pr.py
ShahafBenYakir: Basically the logic of the check_new_pack_metadata function should be the logic of the is_tim_content

Utils/github_workflow_scripts/utils.py
```diff
@@ -120,19 +120,21 @@ def __init__(self, repo: Repo, branch_to_checkout: str, fork_owner: str | None =
...
-                        payload = json.loads(github_event_path)
-                    except ValueError:
-                        print('failed to load GITHUB_EVENT_PATH')  # noqa: T201
-                        raise ValueError(f'cannot checkout to the forked branch {branch_to_checkout} of the owner {fork_owner}')
-                    # forked repo name includes fork_owner + repo name, for example foo/content.
-                    forked_repo_name = payload.get("pull_request", {}).get("head", {}).get("repo", {}).get("full_name")
-                    self.repo.create_remote(name=forked_remote_name, url=f"https://github.com/{forked_repo_name}")
-                else:
-                    raise
+                if f'{forked_remote_name} already exists' not in str(error):
```
ShahafBenYakir, line 123: can you explain this if?

ShahafBenYakir (commented):
See my comments

Utils/github_workflow_scripts/handle_external_pr.py
edik24: the BaseContent.from_path(CONTENT_PATH / file) sometimes returns "not found" for example for images or for items that are not formatted correctly (like incident_field or incident_type missing a mandatory key) 

when not found - content_object.in_pack returns an error. 
This is what I want to catch.

Utils/github_workflow_scripts/handle_external_pr.py
edik24: This will happen if we don't have permissions to checkout remote branch. this happens sometimes.
depends on the remote settings, now all of them allow to checkout their branch

Utils/github_workflow_scripts/handle_external_pr.py
edik24: is_tim_content is only doing the:
BaseContent.from_path(CONTENT_PATH / file)
and based on the results it chooses whether to go to check_new_pack_metadata - which will checkout the remote branch if this is a new pack and not part of the master.

or if this is in the master and the from_path returned an item - I will go to check_if_item_is_tim

you're correct about running for each file, I've added break after.
about checking out every time - I think it saves time that if we don't need to checkout (if this is not a new pack - since most of the PR's are not new) then no need to go there.

Utils/github_workflow_scripts/handle_external_pr.py
edik24: As I noted before - there are some cases (even now) when the checkout is not successful.
At first I just returned false - so if the checkout wasn't successful, we just not assign TIM reviewer.
But then I thought that if i'm already working on this area, I should cover all options - so I added that.

So the logic:
if in the pack exists in master - check the items from graph
if the pack new - try to checkout and then check items from graph
If can't checkout - check items manually

in the last days checkout failed for those PR's (checkout is also done to check support level):
https://github.com/demisto/content/pull/34859 - https://github.com/demisto/content/actions/runs/9522341776/job/26251687499
https://github.com/demisto/content/pull/34827 - https://github.com/demisto/content/actions/runs/9484881582/job/26135465734
https://github.com/demisto/content/pull/34769 - https://github.com/demisto/content/actions/runs/9403458275/job/25900055331

so it is happening from time to time

Utils/github_workflow_scripts/utils.py
edik24: Yes,
when we call the checkout - it first creates a new forked_remote_branch name in line 117:
forked_remote_name = f'{fork_owner}_{repo_name}_{branch_to_checkout}_remote'

so if there was already a try for checkout, for example when checking the support level, the next try to checkout will fail and return an error.

In this case the except clause will ignore the "already exists" error and would continue to line 139 to checkout.

ShahafBenYakir: add this as a comment to the code

.github/github_workflow_scripts/handle_external_pr.py
```diff
@@ -253,35 +254,96 @@ def is_requires_security_reviewer(pr_files: list[str]) -> bool:
...
+    try:
+        pack = content_object.in_pack  # type: ignore
         tags = pack.tags
         categories = pack.categories
         if TIM_TAGS in tags or TIM_CATEGORIES in categories:
             return True
+    except Exception as er:
+        print(f"The pack is not TIM: {er}")
+        return False
+    return False
```
edik24, line 277: add finally instead of 2 falses

ShahafBenYakir (approved):

PR #34623, opened 2024-05-29
EDL: fix error in integration

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-37761

## Description
fix issue where the integration stop to working due to error about log file not exist

fix:
dd condition to check if file exist before reading it


DinaMeylakh (approved):
Good catch, add a unittest and move the checking of the wip to the start of the function as discussed.

PR #34788, opened 2024-06-09
Skip sending Slack Notifications for Tests

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10814

## Description
Add a conditional to check if the PR title contains prefix, `[tT]est`. If the PR title includes this prefix, we do not send a Slack notification.

## Must have
- [x] Test: Skip Slack notification when the external PR title starts with `Test |`. https://github.com/demisto/content/actions/runs/9435799540/job/25989716652?pr=34791.
- [x] Test: Send Slack notification when the external PR title doesn't start with `Test |`. https://github.com/demisto/content/actions/runs/9435764689/job/25989637953#step:10:554.
- [x] Documentation


.github/workflows/handle-new-external-pr.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           LABEL: "Contribution Form Filled"
 
       - name: Send Notification
+        if: startsWith(github.event.pull_request.title, 'Test |') != true
```
RotemAmit, line 57:
```suggestion
        if: startsWith(github.event.pull_request.title, 'Test') != true
```
I think that it will be better if we will check only test.

kgal-pan: https://github.com/demisto/content/pull/34788/commits/58e748db31f0109d4e48ec9357035d42e31e2746

RotemAmit (approved):
Nice catch!

PR #34806, opened 2024-06-09
Revert "add: update octoxlabs (#34645)"

yaakovpraisler:
## Status
- [ ] In Progress
- [x] Ready


Shellyber (approved):

ahmetkotan:
why do you need this revert?

yaakovpraisler:
I thought we'd need to revert this because of some internal issues that were caused by these changes, but I found another solution so we don't need it.

PR #34811, opened 2024-06-10
Added backslashes handling to drilldown enrichment 

ShacharKidor:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [CIAC-10779](https://jira-dc.paloaltonetworks.com/browse/CIAC-10779)

## Description
As described in this issue - [CIAC-10779](https://jira-dc.paloaltonetworks.com/browse/CIAC-10779) - we didn't handle cases of drill-down search queries that include user names. User names in Splunk usually contain backslashes and as described here - [splunk docs](https://docs.splunk.com/Documentation/SplunkCloud/latest/Search/Backslashes#:~:text=To%20pass%20a%20literal%20backslash), in order to pass a literal backslash in an argument to Splunk we must escape the backslash by using the double-slash ( \\ ) string.

In this PR I handled the user name backslashes issue in the drill-down search builder function, similar to the existing handling for this issue in the identity enrichment. 

I referred to the `'user'` and the `'src_user'` Splunk fields (the same as we did for the Identity enrichment). 
We can add more fields in the future if needed. 

## Must have
- [x] Tests
- [x] Documentation


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -972,14 +974,31 @@ def build_drilldown_search(notable_data, search, raw_dict, is_query_name=False):
             if not is_query_name:
                 demisto.error(f'Failed building drilldown search query. Field {raw_field} was not found in the notable.')
             return ""
+
         if prefix:
-            replacement = get_fields_query_part(notable_data, prefix, [field], raw_dict)
+            demisto.debug(f"field is User and Prefix - {field}")
+            if field in USER_RELATED_FIELDS:
+                add_backslash = True
```
yuvalbenshalom, line 981: why this variable "add_backslash"  is needed and not just pass true if we enter this condition?
like: 
eplacement = get_fields_query_part(notable_data, prefix, [field], raw_dict, true)

yuvalbenshalom (commented):
looks good.
see my comment.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
ShacharKidor: @yuvalbenshalom, I added it just for readability. I thought it would be clearer that way. 
But you are right, it is redundant, will change that.

yuvalbenshalom (approved):

PR #34903, opened 2024-06-18
skipping autopep8 on nightlies

AradCarmi:
## Description
Skipping the autopep8 on pre-commit nightlies due to an issue with this step.


YuvHayun (approved):
Such an amazing work!

PR #34982, opened 2024-06-20
Test Playbook for XSIAM 'Impossible Traveler Response'

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10770

## Description
Test playbook for the 'Impossible Traveler Response' XSIAM playbook.


Packs/Core/TestPlaybooks/Test_Playbook_-_Impossible_Traveler_Response.yml
```diff
@@ -0,0 +1,2689 @@
...
+      "yes":
+      - "256"
+    separatecontext: false
+    conditions:
+    - label: "yes"
+      condition:
+      - - operator: isNotEmpty
+          left:
+            value:
+              simple: BlockIndicators
```
ssokolovich, line 2206: Does this key will be created with the provided IPs? I double-checked them with VT and seems like they r not malicious... maybe we should mark one of them as malicious to get this key?

Tests/conf.json
```diff
@@ -21,6 +21,17 @@
     "testTimeout": 160,
     "testInterval": 20,
     "tests": [
+        {
+            "playbookID": "Test Playbook - Impossible Traveler Response",
+            "integrations": [
+                "VirusTotal (API v3)",
+                "Microsoft Graph"
```
ssokolovich, line 28: do you need whois as well for geolocation?

ssokolovich: Also, I see that you are referring to AD here - but you didn't mention it in this file 
image
image

TalNos: removed the AD mentioned in the 'printerrorentry' tasks.

TalNos:
All flows in the playbook end in 'Close alert', which closes the mock incident that runs the test playbook, so the playbook cannot be tested as once it closes, the test cannot continue.

PR #35056, opened 2024-06-25
remove zipfile from native ignore

talzich:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10936

## Description
Remove ZipFile from skip list. The package was originally ignored because of an issue with pyminizip as described in CIAC-1036. Later, pyminizip was replaced altogether so ignoring ZipFile is no longer necessary.


shmuel44 (approved):
Nice!

PR #35059, opened 2024-06-25
Panorama: fix a bug in the fetch_incidents

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-38710

## Description
Currently, due to the bug we increase and increase the `max_fetch` for some log type 
this cause the `nlogs` to be > 5000 which is invalid as the valid value are 0-5000
`Exception: Request Failed. with status: 400. Reason is: Illegal value for parameter nlogs [5200]. Should be between 1 to 5000`


jlevypaloalto:
@ilappe, looks great!
The logic changes are great :)

jlevypaloalto (approved):

PR #35071, opened 2024-06-26
AzureLogAnalytict: fix validat params

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-38605

## Description
in case of authentication using the client credentials/certificate flow, there is a bug in the `validate_params` which require a client secret also.
the instance must have client_secret or certificate&private key

removing this validation as it's already validated in the last if condition 
```
elif not enc_key and not (certificate_thumbprint and private_key):
            raise DemistoException('Key or Certificate Thumbprint and Private Key must be provided.')
```


michal-dagan (approved):
Great!

PR #35084, opened 2024-06-26
RN- removing invalid header(content item doesn't exist)

maimorag:
## Description
Fixing the last RN of Lacework, since it contained a mapper header that doesn't exist in the classifier folder.

### DIFF
--- a/Packs/Lacework/.pack-ignore
+++ b/Packs/Lacework/.pack-ignore
@@ -1,6 +1,9 @@
 [file:Lacework_image.png]
 ignore=IM111
 
+[file:2_0_0.md]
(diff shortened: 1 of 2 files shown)


Packs/Lacework/.pack-ignore
```diff
@@ -1,6 +1,9 @@
 [file:Lacework_image.png]
 ignore=IM111
 
+[file:2_0_0.md]
+ignore=RN115
```
dorschw, line 5: if we remove the faulty one, why ignore it here?

dantavori (approved):

PR #35094, opened 2024-06-27
pyews test fails - add EWS v2 to commands

tcarmeli1:
## Status
- [x]  Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10696

## Description
pyews tpb fails since it use send-mail of other configured integrations, added EWS v2 for each command


sapirshuker (approved):
Great!

PR #35162, opened 2024-07-01
SplunkPy:get drilldown search in correct format

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-38778

## Description
currently, the [feature](https://github.com/demisto/content/issues/34327) that enable the multiple drilldown enrichment queries, assume the `drilldown_searches` key will handled only cases where there is more then one drilldown query, while one query will be in `drilldown_search`

this is not true,
the `drilldown_searches` will contained also one drilldown search, and `drilldown_search` are the 'old' format

test befor fix:
image

test after fix:
image


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy_test.py
```diff
@@ -2645,3 +2645,43 @@ def test_handle_message(item: dict | results.Message, expected: bool):
...
+    
+    assert demisto.error.call_count == 0, 'Something was wrong in the drilldown_enrichment process'
+
+@pytest.mark.parametrize('drilldown_data, expected',
+                         [({'drilldown_search': 'test'},['test']),
+                         ({'drilldown_searches': ['{"search_1":"test_1"}','{"search_2":"test_2"}']},[{'search_1':'test_1'},{'search_2':'test_2'}]),
+                         ({'drilldown_searches': '{"search_1":"test_1"}'}, [{'search_1':'test_1'}])])
+def test_get_drilldown_searches(drilldown_data, expected):
+    """
+    Given:  - drildown search in various formats.
```
ShacharKidor, line 2682: It would be better to write here the tested cases in detail as I did in similar tests.

ShacharKidor (changes requested):
@ilappe, 
Nice and clean! Thanks for handling this issue. 
Added minor comments - see below. 

Also - please link to the PR the second issue we discussed if it would be fixed by this PR as well.

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy_test.py
```diff
@@ -1579,15 +1579,15 @@ def test_drilldown_enrichment(notable_data, expected_result):
           "{\"name\":\"View all test involving user=\\\"$user$\\\"\",\"search\":\"index=\\\"test\\\"\\n| where user ="
                   "$user|s$\",\"earliest\":,\"latest\":}"],
       '_raw': "src=\'test_src\', user='test_user'"},
-     'Failed getting the drilldown timeframe for notable test_id'),
+     'drill-down was not properly configured for notable test_id'),
```
ShacharKidor, line 1582: @ilappe - let's talk about it again... I  Don't think we need to change it.

ShacharKidor, line 1582: @ilappe - let's talk about it again... I Don't think we need to change it.

ShacharKidor (changes requested):

ShacharKidor (approved):

PR #35163, opened 2024-07-01
[DBotPredictURLPhishing] automation fails with certain mailto URI/URLs

jlevypaloalto:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-38872)

## Description
`DBotPredictURLPhishing` has a bug in the case where the responses from the `rasterize` command for single URLs had more than one output and contained an error.

The bug was triggered after the `rasterize` outputs  were merged into a single list.
Screenshot 2024-07-14 at 12 24 37
When the outputs were removed from the `res_rasterize` in the `weed_rasterize_errors` command, there were no corresponding URLs in the `urls` list so the `IndexError: pop index out of range` error was raised.
Screenshot 2024-07-14 at 12 42 04
As a fix, this PR adds the function `return_and_remove_additional_results` which removes and returns the extra outputs so they don't effect the logic.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PhishingURL/ReleaseNotes/1_1_16.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+
+##### DBotPredictURLPhishing
+
+- Fixed an issue in which an error was raised when the `rasterise` command returned multiple responses for a single URL.
```
JudahSchwartz, line 6: think thats spelled wrong

Packs/PhishingURL/Scripts/DBotPredictURLPhishing/DBotPredictURLPhishing.py
```diff
@@ -482,15 +482,15 @@ def rasterize_command(urls: Union[list[str], str], rasterize_timeout: int) -> li
         }
     )
     demisto.debug(f'Rasterize Data: {res_rasterize}')
-    return [res['Contents'] for res in res_rasterize]
+    return [res['Contents'] for res in res_rasterize][:len(urls)]
```
JudahSchwartz, line 485: how do we know the first x are the relevant ones?

Packs/PhishingURL/Scripts/DBotPredictURLPhishing/DBotPredictURLPhishing.py
```diff
@@ -482,15 +482,15 @@ def rasterize_command(urls: Union[list[str], str], rasterize_timeout: int) -> li
...
 
 
 def rasterize_urls(urls: list[str], rasterize_timeout: int) -> list[dict]:
     res_rasterize = rasterize_command(urls, rasterize_timeout)
     if len(res_rasterize) < len(urls):  # check for errors in the response
         demisto.info(f'Rasterize response is too short, running command for each URL\n{res_rasterize=}\n{urls=}')
         rasterize_runs = map(rasterize_command, urls, [rasterize_timeout] * len(urls))
-        res_rasterize = sum(rasterize_runs, [])
+        res_rasterize = [run[0] for run in rasterize_runs]
```
JudahSchwartz, line 493: this is no longer a number, on purpose? Explain this change

JudahSchwartz (commented):
explain this change a bit more in the description of the pr

Packs/PhishingURL/Scripts/DBotPredictURLPhishing/DBotPredictURLPhishing.py
jlevypaloalto: It never was a number it was always a list. In this case I'm just picking out the first of each so as not to take more than one response per URL.

Packs/PhishingURL/Scripts/DBotPredictURLPhishing/DBotPredictURLPhishing.py
jlevypaloalto: It's the same as when we use `response[0]` for a `demisto.executeCommand` output across the repo. Any further additions to the response by the server is ignored.

Packs/PhishingURL/ReleaseNotes/1_1_16.md
jlevypaloalto: I'll fix

Packs/PhishingURL/ReleaseNotes/1_1_17.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+
+##### DBotPredictURLPhishing
+
+- Fixed an issue in which an error was raised when the `rasterize` command returned multiple responses for a single URL.
```
JudahSchwartz, line 6: note the addition of the new param and how it works

jlevypaloalto: That was added in the RN in `rasterize`

Packs/PhishingURL/Scripts/DBotPredictURLPhishing/DBotPredictURLPhishing.py
```diff
@@ -457,6 +457,13 @@ def extract_created_date(entry: dict):
     return None
 
 
+def return_and_remove_additional_results(results: list, from_index):
+    '''Return and remove the extra unneeded results returned from a command call.'''
+    if results[from_index:]:
+        return_results(results[from_index:])
+        del results[from_index:]
```
JudahSchwartz, line 464: log how many removed

JudahSchwartz (approved):

jlevypaloalto:
This PR needs a force-merge as `rasterize` does not work in builds. Issue: [CIAC-10379](https://jira-dc.paloaltonetworks.com/browse/CIAC-10379)

PR #35180, opened 2024-07-02
Defang Added

mandarnaik016:

## Status
- [ ] In Progress
- [x] Ready

## Description
Defangs IP, Mail and URL address to prevent them from being recognized.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/AzureWAF/Integrations/AzureWAF/AzureWAF.py
+++ b/Packs/AzureWAF/Integrations/AzureWAF/AzureWAF.py
@@ -40,6 +40,11 @@ class AzureWAFClient:
                 'grant_type': DEVICE_CODE,
                 'resource': 'https://management.core.windows.net',
                 'scope': 'https://management.azure.com/user_impersonation offline_access user.read'
+            },
+            'Client Credentials': {
+                'grant_type': CLIENT_CREDENTIALS,
+                'resource': None,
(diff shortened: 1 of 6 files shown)


israelpoli:
Hi @mandarnaik016 

Thanks for your contribution.

Here are some thoughts before we proceed with this contribution, considering it’s for XSAOR support level:

It might be worth checking [here](https://hellfire0x01.medium.com/defanging-the-urls-and-keep-yourself-safe-3c194dfac163) to define the default of the script and what additional options can be configured via the arguments.

(We need to have a default defang. For example, only the last dot in an IP address or in the case of a URL the last dot, along with replacing `https` with `hxxps` without replacing the colon.

In addition to the default option, there should be arguments that allow configuring the script to defang all dots in an IP address or in the case of a URL also defang colons and other such options.)

Additionally, please replace all IP addresses, URLs, or email addresses in the file `_test.py` with the addresses already present in the file `Packs/Base/.secrets-ignore`

After doing all this, run demisto-sdk update-release-notes -i Packs/CommonScripts and modify the *.md file to describe the new script you are adding, considering the fixed [templates](https://xsoar.pan.dev/docs/documentation/release-notes#general) of the ReleaseNotes files.

Thanks

PR #35206, opened 2024-07-03
Fixed CortexCoreIR http_request

maimorag:
## Status
- [x] In Progress

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-39134

## Description
- Adding a function to CommonServerPython to check if the platform is using an engine.
- Adding to FORWARD_USER_RUN_RBAC check that the platform is using an engine.
- Removing **using_base_client_http_request** parameter from coreIRAPIModule _http_request.


dantavori (approved):
nice

maimorag:
The build fails on ״Max retries exceeded with url: /containers/reset (Caused by ConnectTimeoutError״ when running with multiple machine build since only one instance is available and it takes time until each machine can lock the integration. Also, the playbooks takes 30min.

https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/6870501 passed successfully.
@dantavori can you force merge?

PR #35209, opened 2024-07-03
Prevent changing infra files

dorschw:
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-9545

image


tkatzir (commented):
Very nice!

Please rename to something like: protect-files.yml
To make it more generic, as we might add other files in the future.

tkatzir (approved):
Great!

PR #35277, opened 2024-07-07
DemistoRESTAPI: added logs

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-37987

## Description
added a log for the case ew tried to get the tenant name but it not exist.


DeanArbel (approved):
Well done

PR #35315, opened 2024-07-09
fix-ews-get-attachment

tcarmeli1:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-39244

## Description
fix ews-get-attachment - for item attachment does not have attachment field


jbabazadeh (approved):
Nice!

PR #35337, opened 2024-07-09
Make General Fields available to All Types

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11083

## Description
The _Email_, _Title_ and _Events_ general fields are not available for selection in common mapping:

image

This prevents users from selecting these OOTB fields and results in them creating new custom fields as was the case of https://github.com/demisto/content/pull/35256.
This PR sets the Email and Title fields to be available for all types.

## Must have
- [ ] Tests
- [ ] Documentation


ssokolovich (commented):
I assume we will need to add RN here + and test that with MP bypass URL. Let me know if you need help with that

PR #35357, opened 2024-07-10
Add new parameters for Sophos Central API

nikstuckenbrock:
## Status
- [ ] In Progress
- [x] Ready

## Description
The Sophos Central Endpoint API provides a parameter for looking up devices by IP addresses. This PR adds this parameter to the related command. Furthermore the Sophos API does a partial string matching to check the IP. In probably the most cases you want a full match, therefor I also added an optional parameter to do a full match after retreiving the results.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
+++ b/Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
@@ -2009,7 +2009,7 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
         args.get("ids"),
         args.get("view"),
         min(int(args.get("limit", "")), 100),
-        args.get("ip_addresses"),
+        args.get("ip_addresses", []).split(","),
     )
     full_match_ip = args.get("full_match_ip") == "true" if args.get("full_match_ip") else None,
     items = results.get("items")
@@ -2029,9 +2029,13 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
             object_data = {field: item.get(field) for field in table_headers}
             # The Sophos API does partial string matching, which can be unwanted for IP addresses.
             if full_match_ip:
-                matching_ips = [
-                    ip for ip in args.get("ip_addresses", []) if ip in object_data.get("ipv4Addresses", []) or ip in object_data.get("ipv6Addresses", [])
-                ]
+                query_ips = args.get("ip_addresses", []).split(",")
+                ips = []
+                if (data := object_data.get("ipv4Addresses", [])) is not None:
+                    ips.extend(data)
+                if (data := object_data.get("ipv6Addresses", [])) is not None:
+                    ips.extend(data)
+                matching_ips = [ip for ip in query_ips if ip in ips]
                 if not matching_ips:
                     continue
             assigned_products = item.get("assignedProducts")
--- a/Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
+++ b/Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
@@ -2009,7 +2009,7 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
         args.get("ids"),
         args.get("view"),
         min(int(args.get("limit", "")), 100),
-        args.get("ip_addresses", []).split(","),
+        argToList(args.get("ip_addresses")),
     )
     full_match_ip = args.get("full_match_ip") == "true" if args.get("full_match_ip") else None,
     items = results.get("items")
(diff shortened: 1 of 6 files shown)


Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
```diff
@@ -2022,6 +2027,17 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
     if items:
         for item in items:
             object_data = {field: item.get(field) for field in table_headers}
+            # The Sophos API does partial string matching, which can be unwanted for IP addresses.
+            if full_match_ip:
+                query_ips = argToList(args.get("ip_addresses"))
+                ips = []
+                if (data := object_data.get("ipv4Addresses", [])) is not None:
+                    ips.extend(data)
```
samuelFain, line 2035: The check `is not None` is redundant because `object_data.get("ipv4Addresses", [])` will always return a list, either the value associated with the key "ipv4Addresses" exists, or an empty list otherwise.
You can make the code more concise and readable like so:
```suggestion
                ips.extend(object_data.get("ipv4Addresses", []))
```

Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
```diff
@@ -2022,6 +2027,17 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
...
+                if (data := object_data.get("ipv6Addresses", [])) is not None:
+                    ips.extend(data)
```
samuelFain, line 2037: Same as comment above:
```suggestion
                ips.extend(object_data.get("ipv6Addresses", []))
```

Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
```diff
@@ -2006,7 +2009,9 @@ def sophos_central_endpoint_list_command(client: Client, args: dict) -> CommandR
         args.get("ids"),
         args.get("view"),
         min(int(args.get("limit", "")), 100),
+        argToList(args.get("ip_addresses")),
     )
+    full_match_ip = args.get("full_match_ip") == "true" if args.get("full_match_ip") else None,
```
samuelFain, line 2014: The setting of `full_match_ip` is a little confusing as it calls `args.get("full_match_ip")` twice.
I suggest using CommonServerPython's `argToBoolean` function for a much more complete check, like so:
```suggestion
    full_match_ip = True if argToBoolean(args.get("full_match_ip")) else None,
```

samuelFain (changes requested):
@nikstuckenbrock Thank you so much for your contribution!
Please see my review comments.
There is a unit test failing in _SophosCentral_test.py_ that needs addressing called `test_sophos_central_endpoint_list_command`. Please see more information [here](https://github.com/demisto/content/actions/runs/9886748843/job/27315050834?pr=35357).

Additionally, in order to confirm that the changes you've implemented are working as expected, a demo is needed.
I suggest that you record and share with me a short video demonstrating  the modifications you made functioning as described. Alternatively, we can arrange a meeting for a live demo instead. Whatever suits you best.

Thanks again!

Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
nikstuckenbrock: I would say the check isn't redundant depending on the `object_data` that is returned by the API. Consider the `object_data` to look like this:

```
object_data = {
    "ip_addresses": None
}
```

Extending the `ips` will then lead to a `TypeError`.

Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
nikstuckenbrock: Same as my comment above.

nikstuckenbrock:
Hey @samuelFain ,

thank you very much for your review and feedback. I've added some ideas of mine and applied most of the feedback. Could you please take a look at it again? You can find a brief demo [here](https://drive.google.com/file/d/1WDlJka979QRrR3eXP0568P_iYrow_mbM/view?usp=sharing). Please tell me when you've watched it so I can delete it afterwards. Thanks in advance.

Packs/SophosCentral/Integrations/SophosCentral/SophosCentral.py
samuelFain: Gotcha, so in that case your version is perfectly logical. No need for changes here.

samuelFain (approved):
@nikstuckenbrock Thank you for the screen recording demo! I've reviewed it and you can go ahead and delete it now. Everything looks good, so I'm going to proceed with merging the PR.

Thank you once again for your contribution!

PR #35526, opened 2024-07-21
Update mypy.sh

kobymeir:
## Status
- [x] In Progress
- [ ] Ready

## Description
Update mypy.sh


kobymeir:
force merge: needed a commit to confuse collect tests with an additional commit

PR #35597, opened 2024-07-25
DBotUpdateLogoURLPhishing_test : change tasks order

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11311

## Description
Currently, the TPB add image and after that delete it
in case of error in deletion, the next add Image will fail with the message `This logo name already exists in the model. Please use another logo name`
like happened in [nightly](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/7251229/raw#L9741)

fix:
change the tasks order, the first step after delete context will try to delete the image and will pass even in case of error (image not exist)

PB before:
image

PB after:
image


omerKarkKatz (approved):
Nice job!

PR #35620, opened 2024-07-28
Stop Interpreting Special Characters

BEAdi:
Removed the `release_changes` from github command, the body is updated in a different place.


jbabazadeh (approved):
great! :) 
add the ticket and the related infra branch

PR #35717, opened 2024-08-01
Fix set-junk-rules regular and global commands

YuvHayun:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: [XSUP-40124](https://jira-dc.paloaltonetworks.com/browse/XSUP-40124)

## Description
Fixed an issue where **ews-global-junk-rules-set** and **ews-junk-rules-set** would fail on empty values for trusted and blocked domains and email sections.

### DIFF
--- a/Packs/MicrosoftExchangeOnline/Integrations/EwsExtensionEXOPowershellV3/EwsExtensionEXOPowershellV3.ps1
+++ b/Packs/MicrosoftExchangeOnline/Integrations/EwsExtensionEXOPowershellV3/EwsExtensionEXOPowershellV3.ps1
@@ -794,16 +794,18 @@ class ExchangeOnlinePowershellV3Client
             # Import and Execute command
             $cmd_params = @{
                 "Identity"                 = $mailbox
-                "BlockedSendersAndDomains" = @{Add = $add_blocked_senders_and_domains
-                    Remove                         = $remove_blocked_senders_and_domains
-                }
-                "TrustedSendersAndDomains" = @{Add = $add_trusted_senders_and_domains
-                    Remove                         = $remove_trusted_senders_and_domains
-                }
                 "TrustedListsOnly"         = $trusted_lists_only
                 "ContactsTrusted"          = $contacts_trusted
                 "Enabled"                  = $enabled
                 "Confirm"                  = $false
+            }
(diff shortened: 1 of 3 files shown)


Packs/MicrosoftExchangeOnline/Integrations/EwsExtensionEXOPowershellV3/EwsExtensionEXOPowershellV3.ps1
```diff
@@ -1760,6 +1764,22 @@ function SetJunkRulesCommand([ExchangeOnlinePowershellV3Client]$client, [hashtab
     return $human_readable, $entry_context, $raw_response
 }
 
+function  CreateAddAndRemoveSections([string[]]$items_to_add, [string[]]$items_to_remove ){
```
JudahSchwartz, line 1767: what about a case where theyre both empty?

Packs/MicrosoftExchangeOnline/Integrations/EwsExtensionEXOPowershellV3/EwsExtensionEXOPowershellV3.ps1
```diff
@@ -1760,6 +1764,22 @@ function SetJunkRulesCommand([ExchangeOnlinePowershellV3Client]$client, [hashtab
...
+    if ((-not [string]::IsNullOrEmpty($items_to_add)) -and (-not [string]::IsNullOrEmpty($items_to_remove))){
```
JudahSchwartz, line 1768: kind of ugly with treating this as 3 cases. Can we say "add it if it exists to an object"? like we would do with python

something like 
```

$object = @{}

if (-not [string]::IsNullOrEmpty($items_to_add)) {
    $object.Add = $items_to_add
}

if (-not [string]::IsNullOrEmpty($items_to_remove)) {
    $object.Remove = $items_to_remove
}`
```

Packs/MicrosoftExchangeOnline/Integrations/EwsExtensionEXOPowershellV3/EwsExtensionEXOPowershellV3.ps1
YuvHayun: In that case it will be empty and not populated (already tested)

JudahSchwartz (approved):
✊ 🐚

PR #35736, opened 2024-08-04
fix expire_password_command

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [XSUP-40197](https://jira-dc.paloaltonetworks.com/browse/XSUP-40197)


Packs/Okta/Integrations/Okta_v2/Okta_v2.py
```diff
@@ -174,19 +174,12 @@ def set_password(self, user_id, password):
             json_data=body
         )
 
-    def set_temp_password(self, user_id):
+    def expire_password(self, user_id, params):
```
JudahSchwartz, line 177: would add `temporary_password` as an argument here and let it take care of the logic of what gets sent to the api

JudahSchwartz (approved):
nice catch!

PR #35784, opened 2024-08-07
URLScan integration: check response status idiomatically instead of strict comparison against `200`

luis261:
## Description
Just a suggestion, feel free to reject of course. Might make sense to use the `.ok` attribute as provided by `requests.Response` instead of literally comparing against `200`.

> [!NOTE]
> Technically, this does constitute a behavior change, according to `help(requests.Response.ok)` :

Returns True if :attr:`status_code` is less than 400, False if not.

**This attribute checks if the status code of the response is between
400 and 600 to see if there was a client error or a server error. If
the status code is between 200 and 400, this will return True. This
is **not** a check to see if the response code is ``200 OK``.**

---

Btw, along the lines of failing early or more strict/robust error handling: long term, it might make sense to consider switching from directly checking status codes to instead calling `Response.raise_for_status` and handling any (expected) raised exceptions appropriately, instead. Switching to `.ok` for now would help pave the way towards this, at least a tiny bit.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
+++ b/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
@@ -119,7 +119,7 @@ def http_request(client, method, url_suffix, json=None, retries=0):
         return_warning('Your available rate limit remaining is {} and is about to be exhausted. '
                        'The rate limit will reset at {}'.format(str(rate_limit_remaining),
                                                                 r.headers.get("X-Rate-Limit-Reset")))
-    if r.status_code != 200:
+    if not r.ok:
         if r.status_code == 429:
             return {}, ErrorTypes.QUOTA_ERROR, rate_limit_reset_after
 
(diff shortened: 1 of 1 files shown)


luis261:
Btw, here are some nits, couldn't help but notice while skimming through `URLScan.py` :
- it should say "5 seconds" instead of "5 minutes" [here](https://github.com/demisto/content/blob/master/Packs/UrlScan/Integrations/UrlScan/UrlScan.py#L316)

The parens around the exception type don't do anything here, if you're going for a tuple you'd need `(requests.exceptions.ConnectionError',)` instead.
https://github.com/demisto/content/blob/1d0d89ef6bc96bce3ab74fd76154dda5bee3553a/Packs/UrlScan/Integrations/UrlScan/UrlScan.py#L179

I don't like just bunching up unrelated changes into a single PR so I did not commit anything, also not worth it to go through the whole overhead including signing off on 2 more contribution forms, just wanted to leave this here FYI

samuelFain:
Hi @luis261!
I see this PR is in a `Draft` status. Kindly let me know if or when can I start reviewing this contribution.

Thank you!

luis261:
Hey 😀 @samuelFain

Depends on the definition of ready for review. Wouldn't necessarily recommend this to be merged as is (at least not without further discussion). Mainly meant as a suggestion/to start a conversation.

Either way, should probably wait until #35785 and #35780 are through.

PR #35785, opened 2024-08-07
URLScan integration: properly handle timeouts by returning an error (instead of just reporting via normal results + debug)

luis261:
## Status
- [ ] In Progress
- [x] Ready

## Description
As part of one of the use cases I've delevoped in the XSOAR instance of my org for automatically triaging, analyzing and responding to suspicious requests blocked/logged by our firewall/IPS, there is a subplaybook which tries to reconstruct what caused a request by a given user to be fired off (e.g. could be directly clicking on a link in an E-Mail/messenger, being redirected directly from another site, clicking on a hyperlink, arriving from a search machine query etc).

As part of this subcomponent of the investigation, I'm using URLScan (for inspecting HTTP transactions, hrefs etc for a given site which is suspected as a "referrer"). Recently, I've noticed a lot of failures in subsequent playbook steps. I tracked the cause down to URLScan submissions failing in the polling stage without causing an actual failure in technical terms (XSOAR warroom/context results). There is just a normal results message which states "The operation timed out. Please try again with a longer timeout period.", but no error.

This causes actual error handling (which would be in place otherwise) to be skipped and the playbook to instead fail on subsequent tasks which depend on expected URLScan output in the context. I've since quickly worked on a workaround for that playbook, but to avoid such unexpected behavior generally, we need to actually fail on polling timeouts, instead of just emitting them like a normal result.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
+++ b/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
@@ -220,7 +220,7 @@ def poll(target, step, args=(), kwargs=None, timeout=60,
         values.put(last_item)
         tries += 1
         if max_time is not None and time.time() >= max_time:
-            demisto.results('The operation timed out. Please try again with a longer timeout period.')
+            demisto.error('The operation timed out. Please try again with a longer timeout period.')
             demisto.debug('The operation timed out.')
             return False
         time.sleep(step)  # pylint: disable=sleep-exists
--- a/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
+++ b/Packs/UrlScan/Integrations/UrlScan/UrlScan.py
@@ -220,9 +220,8 @@ def poll(target, step, args=(), kwargs=None, timeout=60,
         values.put(last_item)
         tries += 1
         if max_time is not None and time.time() >= max_time:
-            demisto.error('The operation timed out. Please try again with a longer timeout period.')
             demisto.debug('The operation timed out.')
-            return False
+            demisto.error('The operation timed out. Please try again with a longer timeout period.')
         time.sleep(step)  # pylint: disable=sleep-exists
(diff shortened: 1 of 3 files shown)


luis261:
I just went with simply replacing `demisto.results` with `demisto.error` for now.
It's the simplest approach and when reading the rest of the integration,
I think this would produce the desired behavior without clashing with any of the
polling constructs in an unexpected/undesired way.

However, it might still make sense to directly raise an exception instead.
If we decide to do it that way, I would suggest using [`TimeoutError`](https://docs.python.org/3/library/exceptions.html#TimeoutError)
or a custom exception instead. (could be named URLScanTimeoutException
or PollingTimeoutException or simply PollingTimeout)

```python
class PollingTimeout(Exception): # TODO consider inheriting from TimeoutError instead
    ...
```

This way, other code "upstream" in the integration could theoretically have a convenient
way to specifically intercept such polling timeouts in cases where that might be sensible.

samuelFain:
Hi @luis261!
I see this PR is in a `Draft` status. Kindly let me know if or when can I start reviewing this contribution.

Thank you!

luis261:
Hello @samuelFain, thank you for reaching out!

Yes, this is actually ready for review (we just need to work out the details regarding the preferred way the error should be thrown in tandem).

What I meant to do was avoid conflicts for the metadata/version regarding my multiple concurrently opened PRs for this integration
Seems like I marked #35780 as ready by accident, when it really should have been this one.

samuelFain:
Hi @luis261, Please see my latest comment in https://github.com/demisto/content/pull/35780.

Thanks!

luis261:
Thanks @samuelFain, [here's my reply in 35780](https://github.com/demisto/content/pull/35780#issuecomment-2286408215)

PR #35826, opened 2024-08-11
JiraV3/fetch-attachments/fix

AradCarmi:
## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-40404)

## Description
Fixed an issue where the mirroring-in process recreated an attachment with any field change.


Packs/Jira/Integrations/JiraV3/JiraV3.py
anas-yousef, line 1: Why not add the logic to the function **get_updated_remote_data**, similar to the **Mirroring comments** condition?
**get_attachments_entries_for_fetched_incident** is also used when we fetch for the first time, and we always want to fetch all the attachments in our first fetch.

AradCarmi: It is the same logic but when you first fetch you don't need to check the times as we want to fetch all of the attachments.

anas-yousef: Exactly, that is why I believe it is better to add the logic only to the **get_updated_remote_data** function. We do the filtering of the comments in that function, and to keep it consistent, we should also do the filtering of the attachments there.

AradCarmi: It is possible but the code currently keeps the behavior fetch behavior. You don't refer to the semantics but only to the design of the if statement right?

anas-yousef: True true, the code will preserve the exact behavior of the fetch.
Yup, just so it stays consistent with the if statement of the **comment** section.

anas-yousef (approved):
Awesome!

dantavori (approved):

PR #35862, opened 2024-08-14
Recoreded future risk score fix

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: [XSUP-40598](https://jira-dc.paloaltonetworks.com/browse/XSUP-40598)

## Description
Changed the indicator field type to be *number* instead of *shorttext* to help the user compare values.


MLainer1 (approved):
🆒

Shellyber:
@ShahafBenYakir This will require force merge - old validation fails on changing the type. 
It shouldn't break BC.

Shellyber:
@ShahafBenYakir What I did is to change the fromversion and added toversion to the old one.
And added the new one with fromversion. 
That is so we can keep the naming

PR #35864, opened 2024-08-14
add comment in pyproject.toml

dorschw:
mention that this pyproject only affects GitHub Actions


RosenbergYehuda (approved):
Awesome!
Virtual Bonusly is on its way!

PR #35967, opened 2024-08-21
Dynamic-section scripts returning unwanted metrics

DinaMeylakh:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-40442)

## Description
When running scripts from dynamic sections, the usual way we check if the current run is a script fails. In that case, the change is to check via a different path in the calling context.

## Must have
- [x] Tests
- [ ] Documentation


DinaMeylakh:
TDD: failing ut commit [bbfee4c](https://github.com/demisto/content/pull/35967/commits/bbfee4cb17851b7b5ab2dafafceb919b49cca584) in this [job](https://github.com/demisto/content/actions/runs/10489559741/job/29054456473?pr=35967)

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -4751,6 +4751,36 @@ def test_integration_return_results_execution_metrics_command_results(mocker):
...
 
+def test_dynamic_section_script_return_results_execution_metrics_command_results(mocker):
+    """
+    Given:
+      - List of CommandResult and dicts that contains an execution metrics entry
+      - The command currently running is a dynamic-section script
+    When:
+      - Calling return_results()
+    Then:
+      - demisto.results() is called 1 time (without the execution metrics entry)
```
ShahafBenYakir, line 4762: your code is expecting 2 calls

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -7117,7 +7117,11 @@ def is_integration_command_execution():
     try:
         return demisto.callingContext['context']['ExecutedCommands'][0]['moduleBrand'] != 'Scripts'
     except (KeyError, IndexError, TypeError):
-        return True
+        try:
+            # In dynamic-section scripts ExecutedCommands is None and another way is needed to verify if we are in a Script.
+            return demisto.callingContext['context']['ScriptName'] == ''
```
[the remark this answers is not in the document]
DinaMeylakh: In dynamic sections, the check before raises an exception since ExecutedCommands is None, and we assume it is an integration. It seems that usually when running any scripts (dynamic-sections or not) the `context.ScriptName` parameter is set. But I think that running an integration command can also result sometimes in a context.ScriptName, therefore I rather leave the previous check as well.

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
DinaMeylakh: fixed

ShahafBenYakir (approved):

PR #35976, opened 2024-08-21
filter first array element for process email attachments

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
[XSUP-40561](https://jira-dc.paloaltonetworks.com/browse/XSUP-40561)

## Description
Since we don't support multiple attachments investigation, filtering to only first attachment header.


ShahafBenYakir (approved):
Need to make sure we are documenting the single file limitation

PR #36036, opened 2024-08-27
Fix str replace

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11633

## Description
`replace` should only occur on strings


omerKarkKatz (approved):
Nice

PR #36038, opened 2024-08-27
CrowdStrikeFalcon: fix tag mirroring

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-40819

## Description
currently, changes in tags in CS side didn't reflect via mirror in.

### DIFF
--- a/Packs/CrowdStrikeFalcon/ReleaseNotes/2_0_13.md
+++ b/Packs/CrowdStrikeFalcon/ReleaseNotes/2_0_13.md
@@ -3,5 +3,5 @@
 
 ##### CrowdStrike Falcon
 
-- Fixed an issue where tags of detections doesn't mirrored in.
+- Fixed an issue where new detection tags were not mirrored into XSOAR.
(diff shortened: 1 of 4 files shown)


ilappe:
still, there is a problematic scenario, when deleting a tag in CS side, it's remain in the tags in XSOAR
need to investigate if this by design in server side (consider of use case of XSOAR incident which have it's own tags which are not related to XS

Packs/CrowdStrikeFalcon/ReleaseNotes/2_0_13.md
```diff
@@ -0,0 +1,7 @@
+
+#### Integrations
+
+##### CrowdStrike Falcon
+
+- Fixed an issue where tags of detections doesn't mirrored in.
```
dantavori, line 6:
```suggestion
- Fixed an issue where new detection tags were not mirrored into XSOAR.
```

dantavori (approved):

PR #36043, opened 2024-08-27
Whois return JSON in raw_response

kgal-pan:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-41040

## Description
Return a dictionary as raw response in the new flow when calling `domain` or `whois` commands.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/Whois/Integrations/Whois/Whois.py
+++ b/Packs/Whois/Integrations/Whois/Whois.py
@@ -9086,7 +9086,7 @@ def whois_and_domain_command(command: str, reliability: str) -> list[CommandResu
                         headers=hr_headers,
                         removeNull=True,
                     ),
-                    raw_response=str(domain_data),
+                    raw_response=dict(domain_data),
                 )
             )
         except Exception as e:
--- a/Packs/Whois/ReleaseNotes/1_5_16.json
+++ b/Packs/Whois/ReleaseNotes/1_5_16.json
@@ -1,0 +1,4 @@
+{
+    "breakingChanges": true,
+    "breakingChangesNotes": "Commands <code>domain</code> and <code>whois</code> will now return a dictionary instead of a string when 'Use legacy context' is not enabled. "
+}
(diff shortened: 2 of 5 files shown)


Packs/Whois/Integrations/Whois/Whois.py
```diff
@@ -9086,7 +9086,7 @@ def whois_and_domain_command(command: str, reliability: str) -> list[CommandResu
                         headers=hr_headers,
                         removeNull=True,
                     ),
-                    raw_response=str(domain_data),
+                    raw_response=dict(domain_data),
```
ShahafBenYakir, line 9089: Let's fix for all raw responses

Packs/Whois/ReleaseNotes/1_5_16.json
```diff
@@ -0,0 +1,4 @@
+{
+    "breakingChanges": true,
+    "breakingChangesNotes": "Commands <code>domain</code> and <code>whois</code> will now return a dictionary instead of a string when 'Use legacy context' is not enabled. "
+}
```
ShahafBenYakir, line 4:
```suggestion
    "breakingChangesNotes": "Commands ***domain*** and ***whois*** will now return a dictionary instead of a string when 'Use legacy context' is disabled. "
}
```
 Also this is true only for raw response so let's make sure it is clear from the note.

ShahafBenYakir (changes requested):

Packs/Whois/Integrations/Whois/Whois.py
kgal-pan: The other `raw_response` are used in the legacy flow. This is the only place that the new flow returns a raw response. I thought we agreed to make this change only in the new flow?

ShahafBenYakir (approved):

PR #36122, opened 2024-09-03
O365 issue 1607

edibleShell:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/content-docs/issues/1607

## Description
The send-mail argument importance states it has a default value of normal when the default value is set to low in the integration settings. This is setting customers email notification priority to low which means data-collection tasks and other emails from the instance might be ignored.

## Must have
- [ ] Tests
- [ ] Documentation


adi88d:
Hi @edibleShell,
We cannot change the default value for this argument because it will break backward compatibility but I will update the description.
Thanks

adi88d (approved):

PR #36124, opened 2024-09-03
Proofpoint bug

eyalpalo:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Must have
- [ ] Tests
- [ ] Documentation


Packs/ProofpointThreatResponse/Integrations/ProofpointThreatResponseEventCollector/ProofpointThreatResponseEventCollector.py
```diff
@@ -375,10 +357,9 @@ def main():  # pragma: no cover
     first_fetch, _ = parse_date_range(params.get('first_fetch', '3 days') or '3 days',
```
dantavori, line 357: let's remove the first fetch (otherwise we reach a time out without the delta, correct me if I'm wrong)
instead, take 10 minutes ago hard coded.
but perhaps worth checking with Meital before doing so, to ensure it is not by design (i.e. we did it only because we always did it until we decided to change the guideline).

PR #36243, opened 2024-09-11
[PAN-OS] fix pan-os-edit-rule no working for some elements

jlevypaloalto:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-41514)

## Description
Fixed the issue in which running the `pan-os-edit-rule` with `element_to_change=source-user` would cause an error.
The error was caused when "source-user" was looked up from the `SECURITY_RULE_ARGS` dictionary causing a `KeyError` as the `SECURITY_RULE_ARGS` dictionary was made for a different command and doesn't hold the relevant keys.
In this PR, a new dictionary `ELEM_TO_CONTEXT` was added that holds the correct keys and values.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -99,7 +97,22 @@
...
 }
-
+ELEM_TO_CONTEXT = {
+    'source': 'Source',
+    'destination': 'Destination',
+    'application': 'Application',
+    'action': 'Action',
+    'category': 'Category',
+    'description': 'Description',
+    'disabled': 'Disabled',
```
sapirshuker, line 107: Why do you want to add new dict instead of adding the source-user value to the SECURITY_RULE_ARGS dict?

Packs/PAN-OS/ReleaseNotes/2_2_2.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Fixed an issue in which the `pan-os-edit-rule` command would fail for some elements.
```
sapirshuker, line 6:
```suggestion
- Fixed an issue in which the ***pan-os-edit-rule*** command would fail for some elements.
```
[release notes examples](https://xsoar.pan.dev/docs/documentation/release-notes#examples)

Packs/PAN-OS/Integrations/Panorama/Panorama.py
jlevypaloalto: Because the `SECURITY_RULE_ARGS` dict was made for a different command and which parses over the the keys. Adding keys will break the behavior.

JudahSchwartz (approved):
looks good, add a test

PR #36335, opened 2024-09-16
Claroty xdome updates

tomlandes:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
1. Fix a bug that happens only in cases where there are at least `2 * max_fetch` incident that have the exact same `update_time`.
2. Make the `alert_type` & `alert_class` fields searchable (customer request).
3. Add 5 custom incident fields: `alert_descriptio`, `mitre_technique_enterprise_ids`, `mitre_technique_enterprise_names`, `mitre_technique_ics_ids`, `mitre_technique_ics_names`

## Must have
- [ ] Tests
- [ ] Documentation


adi88d:
Hi @tomlandes,
is it ready for review?

adi88d:
please fill the [contribution registration form] so we can start the review process
thanks

tomer-pan:
@tomlandes 
Looks good.
I would prefer if you could use the common fields like "MITRE Technique ID" and "MITRE Technique Name" for the relevant fields. I noticed you have 2 pairs, one for ICS and one for Enterprise but you can use a transformer in the mapper so a single field will hold both values.
For example, for the "MITRE Technique ID", you can use "Append" transformer to join the "mitre_technique_enterprise_ids" and "mitre_technique_ics_ids" fields.
Eventually, it will look like this in the UI:
image

And like this in the mapper json:
image

You can do the same for both Mitre fields.

Also, for the alert description field, you can use this common field: "Original Description", I don't think it's necessary to create a new field for this.

In general, we prefer using common fields, when it's suitable, over creating/using custom fields.

Packs/ClarotyXDome/Integrations/XDome/XDome.py
```diff
@@ -19,9 +20,9 @@
 
 SMALLEST_TIME_UNIT = timedelta(seconds=1)
 
-MAX_REQUEST_LIMIT = 5000
+MAX_REQUEST_LIMIT = 5_000
 
-DEFAULT_REQUEST_LIMIT = 1000
+DEFAULT_REQUEST_LIMIT = 1_000
```
adi88d, line 25: why did you change the **DEFAULT_REQUEST_LIMIT** and **MAX_REQUEST_LIMIT**?

it might break backward compatibility for the users?

adi88d (approved):
Hi @tomlandes,
Thank you for your contribution! the code looks good!

We're ready for a demo. Please check [this page](https://xsoar.pan.dev/docs/contributing/demo-prep), and let me know when you're available for one over DFIR. 
Feel free to reach out to me with any questions - I'm available here or on slack :)
Thanks again

Packs/ClarotyXDome/Integrations/XDome/XDome.py
tomlandes: Note that There's no changes to the actual values because python ignores the `_`. I was just trying to improve readability

tomlandes:
@adi88d 
Thank you for the review and also thank for the help with the mapper `json`.

**Regarding the `alert_description` field**
in the context of this integration `incident = alert-device-pair` therefor the `alert_description` is describing only the alert but not the incident. So we don't want to create confusion around this.

**Regarding the MITRE fields**
the problem is that we could face a duplication problem when append the ICS & the enterprise fields. but maybe we can make it work. I'll look into it.

**Regarding the demo**
let's set a time :)

tomlandes:
> I think it's not needed to add these new fields at all and if you want to add the existing mitre fields to the layout - you are welcome to do so but I saw that the techniques and tatitcs IDs are already present in the indicators section in the layout.

As I understand it, we have a client that might care about separating the ICS MITRE techniques from the enterprise MITRE techniques. So I wanted them to have both options, the OOTB combined field and the separated custom fields.
Do you see it as a bad practice or a problem?

tomer-pan:
@tomlandes It's not a bad practice - if it helps for certain customers or scenarios to separate the fields, it's okay.

tomer-pan (approved):

PR #36338, opened 2024-09-16
Fix mypy error flaky for google.cloud library

israelpoli:
## Description
Following a mypy error that sometimes occurs, ignore was added to the import line of the google.cloud library


israelpoli:
Force merge:
Shouldn't affect any job except pre-commit, and it's been tested
`Hotfix for Nightly`

PR #36413, opened 2024-09-22
Fix for large upload playbooks

TalNos:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-41883

## Description
Playbook task 'Calculate Verdict' fails due to an interface-to-string error (InterfaceToString - Unsupported type []string of value [] (52). Two reasons account for the failure of the task:

1 - an incorrect filter is being used within the get 'DBotScore.Score' function. The extracted 'DBotScore.Score' value is filtered based upon whether the 'DBotScore.Indicator' **_equals_** 'Core.OriginalAlert._all_events.os_actor_process_image_sha256' or 'Core.OriginalAlert._all_events.actor_process_image_sha256' or 'Core.OriginalAlert._all_events.causality_actor_process_image_sha256' 
Since 'DBotScore.Indicator' and 'Core.OriginalAlert._all_events.' context keys may contain multiple values (array), we should use the '**_contains_**' filter rather than the 'equal' filter. 

2- Lack of transformers to handle the results of the get 'DBotScore.Score' function, if an array of results is returned. Get 'DBotScore.Score' can return arrays of results, which are then checked using the 'Greater or equal' filter, which cannot handle arrays, resulting in an interface-to-string error as the array cannot be compared to a single number.
Since the outcome can also be an array, I have added the 'unique', 'sort', and 'last array element' transformers to obtain the highest or lowest score that can then be compared as a single number.


Packs/Core/ReleaseNotes/3_0_68.md
```diff
@@ -0,0 +1,10 @@
+
+#### Playbooks
+
+##### Large Upload Alert
+
+- Added transformers to some conditions within the *'Calculate Verdict'* task to obtain the highest or lowest score in cases where an array of results was returned.
+- Fixed an issue which caused the *'Calculate Verdict'* task to fail due to an incorrect filter being used in certain conditions of the *'Calculate Verdict'* playbook task.
+- Fixed the context path for risk scores of risky hosts and users within the *'malicious'* condition of the *'Calculate Verdict'* task.
+- Updated task numbers 32 and 48's descriptions and names.
```
ssokolovich, line 9: Didn't see that on your PR - maybe the format reverted it?

Packs/Core/ReleaseNotes/3_0_68.md
```diff
@@ -0,0 +1,10 @@
+
+#### Playbooks
+
+##### Large Upload Alert
+
+- Added transformers to some conditions within the *'Calculate Verdict'* task to obtain the highest or lowest score in cases where an array of results was returned.
```
ssokolovich, line 6:
```suggestion
- Added transformers to the conditions within the *'Calculate Verdict'* task to obtain the highest or lowest score in cases where an array of results was returned.
```

Packs/Core/ReleaseNotes/3_0_68.md
TalNos: Screenshot 2024-09-24 at 17 04 19
Screenshot 2024-09-24 at 17 04 50
Screenshot 2024-09-24 at 17 05 31
here are the changes and I committed another change to task 48 name

Packs/Core/ReleaseNotes/3_0_68.md
TalNos: done

ssokolovich (approved):

PR #36573, opened 2024-10-01
Extract Indicators From File PB: not convert docx to pdf

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-41176

## Description
currently, we consider a file which contain `xml` in the `file.info` - as file to be converted to pdf
this lead to convert also a docx file with `vnd.openxmlformats-officedocument.wordprocessingml.document` in the `file.info`

adding additional check to filter out `doc/docx` files


idovandijk (changes requested):
Great, but we also need to add the same filter to the conversion task (not just the conditional task), otherwise the condition may pass (due to other supported files), and then the conversion will run (against both supported and unsupported files)

idovandijk (approved):
Awesome

AdiPeret (approved):

PR #36595, opened 2024-10-04
Flashpoint Release 201

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

## Description
- Fixed an issue with the resource URL incident field in cases where it is not populated by default.

## Must have
- [x] Tests
- [x] Documentation


Packs/Flashpoint/Integrations/Ignite/Ignite.py
```diff
@@ -142,7 +142,7 @@
 ALERT_RESOURCE_URL = {
     'communities': '/search/context/communities/{}',
     'marketplaces': '/search/context/marketplaces/{}',
-    'media': '/search/results/media?include.media_id={}',
+    'media': '/search/results/media?include.date=all+time&include.media_id={}',
```
tcarmeli1, line 145: why did you changed the the original url?

tcarmeli1 (changes requested):
Hi @crestdatasystems ,
Good work :)
Please see my comments.
General comment- Please add new unit tests for the new functionality you added.
Thanks again

Packs/Flashpoint/Integrations/Ignite/Ignite.py
crestdatasystems: Hi @tcarmeli1, The old `media` based URL was not having the parameter to search it on bases of all time. So, we have added date parameter to resolve the issue.

crestdatasystems:
> Hi @crestdatasystems , Thank you for your contribution! Good work :) Please see my comments. General comment- Please add new unit tests for the new functionality you added. Please feel free to reach out to me with any questions - I'm available here or on slack :) Thanks again

Hi @tcarmeli1,
 Thanks for reviewing the PR.
 The intention of this release was to resolve the issue of unavailability of the Resource URL.
-  In some of type of alerts, the Resource URL is directly available from the API response.
- To resolve the issue, we have forged custom Resource URL based on the type of Alert.
- So, there is no change in any major functionality. And we also wanted to check that the existing test cases should pass, and there should be only change in test data to check that existing functionality haven't affected due to this release.
- This was the reason to not modify the test cases.

Packs/Flashpoint/Integrations/Ignite/Ignite.py
tcarmeli1: So it will not lead to backwards compatibility right?

tcarmeli1:
Hi @crestdatasystems ,
Thank you for your response!
It depends on your decision but i would suggests adding one more test in order to test the functionality when a Resource URL is not provided by the API.
In addition, I would appreciate if you can please create a Demo video so i can go on and merge this pr, we can also conduct a Demo meeting if you prefer.
Thank you so much!

Packs/Flashpoint/Integrations/Ignite/Ignite.py
crestdatasystems: The change doesn't affect any functionality. Its only related to the platform side evaluation of the parameters.

tcarmeli1 (approved):
Watched the demo- all looks great.
Great work!

PR #36602, opened 2024-10-06
CrowdStrikeFalcon: ioc count in case of rate limit

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-42040

## Description
currently, if a rate limit occurred in CS API - the user will not know about this
fix:
in case there is a rate limit indication in the raw response - return it to the user

test befor fix:
image

test after fix:
image


DeanArbel (approved):
Good job

PR #36683, opened 2024-10-10
FireEye HX: fix context path

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-42448

## Description
the context path defined in the yml for the command `fireeye-hx-search` have the prefix `FireEyeHX.Search.results`
while the actual data are under `FireEyeHX.Search.results.data`
fo example: in the playbooks we search for the key  `FireEyeHX.Search.results.File Name` which the correct one are `FireEyeHX.Search.results.data.File Name`


ilappe:
Apparently it's not BC because it never worked
and customers are not supposed to change something that worked

DeanArbel (approved):
Nice

PR #36692, opened 2024-10-10
[Pack][RecordedFuture][v1.8.1] Fix the api client to use proxy when corresponding parameter is set to true.

recordedfuture-dev:
# Recorded Future v1.8.1

## Status
- [x] Ready

## Related Issues
Fixing issue where Proxy is not used by integration despite the integration is configured to use proxy.

## Description
The issue was due to the fact that we didn't provided proxy=True to the Client() class, when user set proxy=True in the integration settings.

## Must have
- [x] Tests
- [x] Documentation


aaron1535 (approved):

recordedfuture-dev:
Thank you!

PR #36794, opened 2024-10-22
Update DomainTools_Iris.py

bdanjoux:

## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: Running !domaintools-whois with raw-response=true returns nothing (an empty dict)

## Description
This PR introduces a "contents" in the results of the !domaintools-whois command to bring it inline with the other commands of this integration that return both Human Readable results and Raw results

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/DomainTools_Iris/Integrations/DomainTools_Iris/DomainTools_Iris.py
+++ b/Packs/DomainTools_Iris/Integrations/DomainTools_Iris/DomainTools_Iris.py
@@ -1551,7 +1551,7 @@ def parsed_whois_command():
     results = CommandResults(
         indicator=domain_indicator,
         readable_output=human_readable,
-        outputs=parsed,
+        raw_response=parsed,
         ignore_auto_extract=True
     )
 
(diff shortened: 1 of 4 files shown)


aaron1535:
Hi @bdanjoux,
Thank you for your contribution.

In order to move forward with your contribution process, I want to understand the use case.
Usually, we use `raw_response` when we make any API call for the requested command, and in `raw_response` we will return the response returned from the endpoint
The use of `outputs` is to return information to the context (for which, of course, we need to arrange the information we want to return in an orderly dictionary)
Please explain what you want to do
Thank you

bdanjoux:
Hi @aaron1535 

I want to make sure that the command `domaintools-whois` returns the raw response of the API, as it currently only returns human readable results.

Indeed, using `raw_response` is all that's needed here, I will modify the PR accordingly.

aaron1535:
@bdanjoux Indeed, this is the right way to do what you asked for.
Please add release notes so we can merge the change
You can find more explanation [here](https://xsoar.pan.dev/docs/documentation/release-notes)

aaron1535:
@bdanjoux. We are waiting for the partner's approval so that we can merge
in the meantime
Please update your Docker image to the latest version `1.0.0.114678`

aaron1535 (approved):

PR #36816, opened 2024-10-27
Added assets fix and snapshot fix

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/xsup-40703

## Description
- Added handling for large assets (over 1 MB)
- added the ability to re-run the fetch assets in case of timeouts
- Chaged snapshot IDs to start with a timestamp
 
## Must have
- [x] Tests


Packs/qualys/Integrations/Qualysv2/Qualysv2.py
```diff
@@ -2803,6 +2807,30 @@ def add_fields_to_events(events, time_field_path, event_type_field):
...
+            asset['DETECTION']['RESULTS'] = results[:results_characters_lim]
+            asset['isTruncated'] = True
+            demisto.debug(f'Truncated Asset ID: {host_id}{detection_str} to {results_characters_lim}')
+            demisto.debug(json.dumps(asset))
+
+        # For debug
+        for key, val in asset.items():
+            if (val_size := get_size_of_object(val)) > ASSET_SIZE_LIMIT:  # 1 MB
+                demisto.debug(f'Data under key "{key}" has size of {val_size}:\n'
+                              f'{str(val)[:10000]}...')
```
moishce, line 2831: We still need it?

moishce (approved):

Packs/qualys/Integrations/Qualysv2/Qualysv2.py
ShahafBenYakir: In theory, the fix should do the trick but I think that for future cases where other keys might be problematic, it could be nice to have a better overview of other problematic keys (probably will not happen but for any case that may come)

PR #36874, opened 2024-10-28
CiscoUmbrellaReporting: Added optional "categories" argument to supporting commands

psybertool:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
This is an addition to the already merged PR #36253 
The optional `categories` argument has been added to all commands which supports it according to the [Cisco Umbrella API](https://developer.cisco.com/docs/cloud-security/api-reference-reports-reporting-overview/) (in the initial PR the argument itself was only added to two of the commands but 7 other commands support it as well).

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/CiscoUmbrellaReporting/ReleaseNotes/1_1_4.md
+++ b/Packs/CiscoUmbrellaReporting/ReleaseNotes/1_1_4.md
@@ -1,4 +1,11 @@
 #### Integrations
 ##### Cisco Umbrella Reporting
-- Added the optional `categories` argument to all commands which support the category filter.
+- Added the *categories* argument to following commands:
+    - ***umbrella-reporting-destination-list***
+    - ***umbrella-reporting-category-list***
+    - ***umbrella-reporting-identity-list***
+    - ***umbrella-reporting-event-type-list***
+    - ***umbrella-reporting-file-list***
+    - ***umbrella-reporting-threat-list***
+    - ***umbrella-reporting-summary-list***
 - Updated the Docker image to: *demisto/python3:3.11.10.113941*.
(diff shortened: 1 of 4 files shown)


Packs/CiscoUmbrellaReporting/ReleaseNotes/1_1_4.md
```diff
@@ -0,0 +1,4 @@
+#### Integrations
+##### Cisco Umbrella Reporting
+- Added the optional `categories` argument to all commands which support the category filter.
```
yaakovpraisler, line 3:
```suggestion
- Added the *categories* argument to following commands:
    - ***umbrella-reporting-destination-list***
    - ***umbrella-reporting-category-list***
    - ***umbrella-reporting-identity-list***
    - ***umbrella-reporting-event-type-list***
    - ***umbrella-reporting-file-list***
    - ***umbrella-reporting-threat-list***
    - ***umbrella-reporting-summary-list***
```

yaakovpraisler (commented):
Hey @psybertool, thank you for your contribution!
Please check if it's possible to add for it a test in the unit tests file.

psybertool:
Hi @yaakovpraisler 
Thanks for the review and the inputs.

The additional argument itself has already been reflected in the unit tests with the original PR, see here: https://github.com/demisto/content/pull/36253/files#diff-fa10be96fb8f1fd92f047cd405c8fbb259f0da4383fd656fcac8f5f3d696c748

The categories argument (which is always optional) only limits the result sent back from the Umbrella API, therefore it doesn't change the behaviour of the UmbrellaReporting integration itself. Therefore additional unit tests (other than the already added in the original PR) somehow doesn't make that much sense right? The other optional arguments which are similar to the categories argument aren't tested neither.

yaakovpraisler (approved):

PR #36884, opened 2024-10-29
QRadar - Generic - fixed issue with username enrichment

idovandijk:
## Status
Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-43037

## Description
Fixed an issue where the username from the QRadar offense was not being enriched due to incorrect context path in playbook `QRadar - Generic` -> task `Entity Enrichment - Generic v3`.
The fix preserves BC and still sets Account.Username if QRadar.Username is empty.

### DIFF
--- a/Packs/QRadar/Playbooks/playbook-QRadar_Generic.yml
+++ b/Packs/QRadar/Playbooks/playbook-QRadar_Generic.yml
@@ -527,6 +527,7 @@ tasks:
       id: 670d8449-b528-465d-8b3f-6e22cda63229
       version: -1
       name: Was the SOC email address provided?
+      description: ""
       type: condition
       iscommand: false
       brand: ""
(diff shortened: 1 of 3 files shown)


Packs/QRadar/Playbooks/playbook-QRadar_Generic.yml
```diff
@@ -1805,10 +1872,26 @@ tasks:
...
-          root: Account
-          accessor: Username
+          root: QRadar.Username
           transformers:
+          - operator: SetIfEmpty
+            args:
+              applyIfEmpty: {}
+              defaultValue:
+                value:
+                  simple: Account.Username
```
AdiPeret, line 1882: Shouldn't that be changed to QRadar.Username?

idovandijk: Wanted to add a screenshot but it won't let me (browser?)
this is part of the `SetIfEmpty` transformer, but above it you can see QRadar.Username. So if QRadar Username doesn't exist, it will fallback to Account.Username (as it was before)

AdiPeret (approved):

PR #37177, opened 2024-11-11
CRTX-139456 Parsing Rule Fixes for Cisco Catalyst and Azure Security Center

cweltPA:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [CRTX-139456](https://jira-dc.paloaltonetworks.com/browse/CRTX-139456)

## Description
This PR addresses parsing rule issues in two packs, which cause the *`_time`* value to be populated with the default fallback value of *`_insert_time`* : 

| # | Pack | Dataset |
| :-| :------------------------| :--------------------------------------|
| 1 |  Cisco Catalyst                 | `cisco_catalyst_raw`                             | 
| 2 |  Azure Security Center   | `microsoft_defender_for_cloud_raw`|

## Must have
- [ ] Tests
- [ ] Documentation


Packs/AzureSecurityCenter/ReleaseNotes/2_0_33.md
OBavly, line 1: I would avoid focusing on the error and more on the fix itself. Consider changing to 'Added support for timestamp in 'datetime' format

Packs/CiscoCatalyst/ReleaseNotes/1_0_7.md
OBavly, line 1: Same thing with the description

OBavly (changes requested):

Packs/AzureSecurityCenter/ReleaseNotes/2_0_33.md
cweltPA: In general I follow the guidelines described here: [release-notes#examples-and-best-practices](https://xsoar.pan.dev/docs/documentation/release-notes#examples-and-best-practices). 

The underlying data type (string or datetime) is an internal implementation issue which is not relevant in the customer's point of view. The datatype is set based on the actual values, and thus should be the same across all tenants using the designated data source/collector ("Azure Event Hub"). On this case, we overlooked this issue, the treated the timestamp field as a string (without an explicit conversion from datetime to string), which caused the parsing rule to fail and report on the corresponding errors. This fix is not adding new functionality but rather fixes a bug on existing functionality which was allegedly supported from release 2.0.30 as declared there on the release notes: 
`Added parsing rule for _time extraction for Microsoft Defender for Cloud events, ingested via "Azure Event Hub".`
From the customer's point of view, as long as he ingested the logs via the Azure Event Hub, the parsing rule should have worked beginning from release 2.0.30, so it would be more accurate to reflect on the existing functionality fix rather than allegedly adding/expanding functionality.

Packs/CiscoCatalyst/ReleaseNotes/1_0_7.md
cweltPA: Same here, there was a typo on a format that was declared already to be supported. 
We didn't add support for new/additional formats, just fixed bugs on formatted which have already assumed to be supported.

OBavly (approved):

PR #37327, opened 2024-11-20
O365 - Security And Compliance - Search Action - Preview: add DT condition

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-42892

## Description
In case that the `search` not found , the GenericPolloing in the `O365 - Security And Compliance - Search Action - Preview` playbook will fail because it take the incorrect search name (the that didn't found)

adding a `DT` to the playbook to use the `*_Preview` search in that case


BEAdi (commented):
Looks good

BEAdi (approved):

PR #37329, opened 2024-11-20
Add a check for Security Approved for contributions

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12235

## Description
If a contributor has opened a PR which requires the Security team to review it, then the merge should be blocked until they have reviewed the PR. This check will block a merge only if the PR is from a contribution and only if the "Security Review" label has been added.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/security-label-check.yml
+++ b/.github/workflows/security-label-check.yml
@@ -1,7 +1,7 @@
 name: Security Label Check
 on:
   pull_request:
-    types: [opened, edited, labeled, synchronize]
+    types: [opened, edited, labeled, synchronize, unlabeled]
 jobs:
   security-check:
     runs-on: ubuntu-latest
--- a/.github/workflows/security-label-check.yml
+++ b/.github/workflows/security-label-check.yml
@@ -5,7 +5,7 @@ on:
 jobs:
   security-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


RotemAmit (changes requested):
Great idea! Thank you very much!
Just some small changes.

amshamah419:
Here's the testing PR - https://github.com/demisto/content/pull/37337

.github/workflows/security-label-check.yml
```diff
@@ -0,0 +1,27 @@
+name: Security Label Check
+on:
+  pull_request:
+    types: [opened, edited, labeled, synchronize, unlabeled]
+jobs:
+  security-check:
+    runs-on: ubuntu-latest
+    if: 'startsWith(github.head_ref, ''contrib'') || (github.event.pull_request.head.repo.fork == true && contains(github.event.pull_request.base.ref, ''contrib''))'
```
RotemAmit, line 8:
```suggestion
    if: github.repository == 'demisto/content' && github.event.pull_request.head.repo.fork == true && contains(github.head_ref, 'xsoar-bot-contrib-ContributionTestPack') == false
```
Let's make it with the same condition as the rest of the github actions. Also, this way it is more general, in case that in the future the 'contrib' prefix will change.

amshamah419: Sure, we can do that

RotemAmit (approved):
Good job on the PR and the catch!

PR #37341, opened 2024-11-21
Add send-notification command to ServiceNow integration

nikstuckenbrock:
## Status
- [x] In Progress
- [ ] Ready

## Description
Add send-notification command to ServiceNow integration to directly interact with ServiceNow in ask or data collection tasks.
## Must have
- [x] Tests
- [x] Documentation 

Just as a sidenote: I will be on vacation for the next week, therefor it will take a week for me to respond.


amshamah419:
Hi @nikstuckenbrock ,

Thank you again for opening a pull request. Unfortunately, we cannot accept it, as this is a significant change which requires an in-house design due to it using a builtin command which is not supported on all platforms quite yet. 

nikstuckenbrock:
Hi @amshamah419,

thanks for your response. I know that it is not supported by all platforms, that's why i created the PR. We can still fork the integration locally, but that's probably bad practice. Could you explain a little bit more why this change i such of a significant change? Because it is downwards capable and only adds a new command?

nikstuckenbrock:
Hi @amshamah419,

I wanted to follow up and ask if there have been any changes or updates regarding this topic?

PR #37346, opened 2024-11-21
Rubrik Release 1.4.0

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

## Description
- Added the "ip" and "domain" commands.
- Added new playbook: "Rubrik Workload Analysis" to increase the incident severity based on the workload risk levels and threat information.
- Added new automation script: "RubrikSetIncidentSeverityUsingWorkLoadRiskLevel" to set the incident severity using the workload data provided from the argument.

## Must have
- [x] Tests
- [x] Documentation


Packs/RubrikPolaris/Integrations/RubrikPolaris/RubrikPolaris.py
```diff
@@ -2092,6 +2113,92 @@ def prepare_context_hr_suspicious_file_list(snappable_investigations_data: dict,
...
+    '''
+    invalid_ip_addresses = []
+    valid_ip_addresses = []
+    for ip in ips_list:
+        ip = ip.strip().strip('\"')
+        if ip:
+            if is_ip_valid(ip, accept_v6_ips=True):
+                valid_ip_addresses.append(ip)
+            else:
+                invalid_ip_addresses.append(ip)
```
amshamah419, line 2198:
```suggestion
    for ip in ips_list:
        ip = ip.strip('\"')
        if is_ip_valid(ip, accept_v6_ips=True):
            valid_ip_addresses.append(ip)
        else:
            invalid_ip_addresses.append(ip)
```

amshamah419 (approved):
Hi @crestdatasystems Looks great! Just a couple comments which can help performance-wise

crestdatasystems:
> Hi @crestdatasystems Looks great! Just a couple comments which can help performance-wise

Thanks @amshamah419, for reviewing PR.

Please let us know if you need anything else from our side in order to finish it!

amshamah419:
Hi @crestdatasystems we are currently waiting on a security review. Once done, we can merge. - @idovandijk any eta?

crestdatasystems:
Hi @amshamah41, Its been more than two weeks since we have committed suggested changes. Can we please make some progress from here?

amshamah419:
@idovandijk are we okay to merge?

idovandijk (changes requested):
Hi there @crestdatasystems, thank you for your continued patience here as I finally got to review your files! You are continuously working hard on more and more playbooks and it's wonderful to see that happening.

I'll list my comments below for the playbook, and one for the script, numbered purely for convenience purposes:

### Rubrik Workload Analysis - Rubrik Security Cloud
1. Can you please explain the need for DeleteContext in the playbook? Are incidents re-run at any point? If this was done just for testing, let's remove it.

3. If the IPs and Domains are not inputted, do we expect them to be associated and found in the incident? If so, why? What is the trigger for this playbook?
image

I find this potentially problematic because we check inputs, if they don't exist we search for indicators by their association to the incident, and if we get empty results we take the inputs anyway (even if they were empty to begin with).

4. Also related to point 3 - can we set default context input to these playbook inputs? 
image

For example, if IPs are mapped to fields on incident ingestion, we can perhaps use `incident.rubrikips`, or if we want this to be a generic playbook that could be used anywhere, we may want to use `IP.Address` (the default context where extracted IPs go).

5. Please add information to the playbook inputs - what is the syntax we expect: is it comma-separated IPs, a list of IPs, or something else? Worth adding an example.

6. Should we have an OR between the conditions that check if inputs were specified? This means if one of them is empty.

7. Please remove the periods from the task names. You can browse the[ playbook conventions documentation ](https://xsoar.pan.dev/docs/playbooks/playbook-conventions) for guidance.

### RubrikSetIncidentSeverityUsingWorkLoadRiskLevel
1. Missing a description of some sort that explains what severity is changed - is it XSOAR incident severity? If so, how does `increase_severity_by` input behave if for example we are at severity 3 and increasing by 2, but the maximum is 4?

To summarize I think the direction is great, I'd like to get a short demo and to get your reply for the comments as I'm still slightly confused on the incident and data flow here.
If there's anything you're not sure about, struggling to do, or have any question, feel free to ping me here.

Finally, I want to assure you that now that I've reviewed your PR, I will make sure to stay available to help get this merged quickly. Waiting to hear back, good luck :)

crestdatasystems:
> Hi there @crestdatasystems, thank you for your continued patience here as I finally got to review your files! You are continuously working hard on more and more playbooks and it's wonderful to see that happening.
> 
> I'll list my comments below for the playbook, and one for the script, numbered purely for convenience purposes:
> 
> ### Rubrik Workload Analysis - Rubrik Security Cloud
> ...
> ### RubrikSetIncidentSeverityUsingWorkLoadRiskLevel
> ...

Thanks @idovandijk for reviewing the PR. We have updated the PR as per your suggested changes and we are also adding the following points as per your comments over the playbook and the script.

### Rubrik Workload Analysis - Rubrik Security Cloud

1. The playbook is designed to re-run any number of times.
- Because the main perspective of the playbook is to do analysis on the indicators which keeps gets changing as we run any kind of command, script or any other playbook.
2. The playbook will give first priority to the indicators supplied in the playbook inputs.
- If there are no indicators supplied, then the "Get Indicators" task will fetch the indicators from the current incident.
3. We had added check in the condition for checking if there are any indicators present or not.
- If not, then the path will reach to the end of the playbook.
4. The reason for not setting the default context for the playbook input is that the:
- We are already handling the blank inputs scenario in the latter part of the playbook.
- If we directly set the playbook inputs, then user might be confused whether they should change it or not, and what could be the impact.
5. Certainly, we have updated the description and which will tell the user how to use it and what would be the impact.
6. We are already using the OR condition for checking the playbook inputs.
- If one of supplied, then we will not use the incident indicators.
7. We have removed the periods from the task names.

## RubrikSetIncidentSeverityUsingWorkLoadRiskLevel

1. We have updated the description that it will change the XSOAR incident severity.
- We also updated the description for the "increase_severity_by" argument, as it follows internal instruction for the XSOAR incident severity.

idovandijk:
Thank you @crestdatasystems, I can confirm the changes you made! We're almost done now.
I'd like to get a short demo of the playbook, but if you prefer, I'm also completely fine with a recorded video of successful playbook execution, showing the task results of the playbook tasks and the context sidebar.

Please let me know which one you prefer, and we can proceed accordingly.

idovandijk (approved):
Well done!

PR #37430, opened 2024-11-25
get_detections_entities:

bdanjoux:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: CASE 03318743

## Description

The July commit https://github.com/demisto/content/commit/f53aa8711a013ebe4a893358dfad8f4519105ec7 related to then upcoming Crowdstrike Raptor release incorrectly updated the mechanism by which detection summaries are fetched.

The "legacy" API for fetching detection summaries: /detects/entities/summaries/GET/v1
was incorrectly amended to use /alerts/entities/alerts/v2 for clients who didn't tick the "legacy API" integration instance configuration boolean.

This was incorrect because the endpoint /detects/entities/summaries/GET/v1 is not deprecated in the CrowdStrike API documentation and also because /alerts/entities/alerts/v2 is for fetching alerts summaries, not detection summaries.

Subsequently, passing a valid detection id to an instance configured to use the new API endpoints fails when it works if you tick "use legacy API".

I removed mentions of LEGACY_VERSION and the incorrect new API endpoint.
I also corrected the argument description for cs-falcon-list-incident-summaries: ids
this command fetches incident details and subsequently takes incident IDs, not detection IDs

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
+++ b/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
@@ -5006,8 +5006,7 @@ def detections_to_human_readable(detections):
     detections_readable_outputs = []
     for detection in detections:
         readable_output = assign_params(status=detection.get('status'),
-                                        max_severity=detection.get('max_severity_displayname') if LEGACY_VERSION else
-                                        detection.get('severity_name'),
+                                        max_severity=detection.get('max_severity_displayname'),
                                         detection_id=detection.get('detection_id'),
                                         created_time=detection.get('created_timestamp'))
         detections_readable_outputs.append(readable_output)
--- a/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
+++ b/Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
@@ -1590,9 +1590,9 @@ def get_detections_entities(detections_ids: list):
         :param detections_ids: IDs of the requested detections.
         :return: Response json of the get detection entities endpoint (detection objects)
     """
-    ids_json = {'ids': detections_ids} if LEGACY_VERSION else {"composite_ids": detections_ids}
-    url = '/detects/entities/summaries/GET/v1' if LEGACY_VERSION else '/alerts/entities/alerts/v2'
-    demisto.debug(f"Getting detections entities from {url} with {ids_json=}. {LEGACY_VERSION=}")
+    ids_json = {'ids': detections_ids}
+    url = '/detects/entities/summaries/GET/v1'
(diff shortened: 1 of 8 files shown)


Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -1590,9 +1590,9 @@ def get_detections_entities(detections_ids: list):
         :param detections_ids: IDs of the requested detections.
         :return: Response json of the get detection entities endpoint (detection objects)
     """
-    ids_json = {'ids': detections_ids} if LEGACY_VERSION else {"composite_ids": detections_ids}
-    url = '/detects/entities/summaries/GET/v1' if LEGACY_VERSION else '/alerts/entities/alerts/v2'
-    demisto.debug(f"Getting detections entities from {url} with {ids_json=}. {LEGACY_VERSION=}")
+    ids_json = {'ids': detections_ids}
+    url = '/detects/entities/summaries/GET/v1'
+    demisto.debug(f"Getting detections entities from {url} with {ids_json=}.")
```
amshamah419, line 1595: Hi @bdanjoux I'm not sure if we should remove the support for legacy version. Is there a specific issue you are experiencing?

bdanjoux: Hi @amshamah419 I am experiencing an issue where an integration instance is not able to fetch detection summaries, that instance is configured with "use the legacy API" unticked.

As I explain in the PR description, the API endpoint /detects/entities/summaries/GET/v1 is not deprecated, it is very much current, please check the Crowdstrike API documentation.

This problem was introduced in the commit https://github.com/demisto/content/commit/f53aa8711a013ebe4a893358dfad8f4519105ec7    back in July.

in that July PR, many commands were updated as their endpoint was being deprecated and replaced with a new one, hence the LEGACY switch.

However, the command to fetch detection summaries should not have been modified, as its underlying endpoint was not deprecated. Additionally, what that still valid endpoint was replaced with, is an endpoint for alert summaries, not detection summaries.

This is why my integration instance is failing to retrieve detection summaries.

I input a valid detection ID to the detection summary fetching command, but the integration command feeds it to an endpoint that expects an alert ID instead.

In summary, I am not removing support for the legacy API endpoint, I am only trimming an incorrectly amended integration command.

amshamah419: Hi @bdanjoux Can you please open an issue with CS? I think this is something which should be investigated since it could have effects elsewhere.

amshamah419:
Hi @bdanjoux ,
Thank you again for opening a pull request. Unfortunately, we cannot accept it, as the change which was made originally was by design and works in our testing. If you are experiencing an issue with the command, please open a bug report with your CS so we can investigate further. 

PR #37498, opened 2024-12-02
AzureSentinel: restriction on mirror out

ilappe:
## Related Issues
related: https://jira-dc.paloaltonetworks.com/browse/XSUP-43987

## Description
update the remote incident only when some of the mirrored fields was changed


Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -705,6 +705,12 @@ def update_incident_request(client: AzureSentinelClient, incident_id: str, data:
     Returns:
         Dict[str, Any]: the response of the update incident request
     """
+    # we will run the mirror-out update only if there is relevant changes or need to close the remote ticket
+    relevant_keys_delta = delta.keys() & OUTGOING_MIRRORED_FIELDS.keys()
+    if not relevant_keys_delta and not close_ticket:
+        demisto.debug('No relevant changes to update in the remote system')
+        return {}
```
mmhw, line 712:
```suggestion
    # we will run the mirror-out update only if there is relevant changes or need to close the remote ticket
    close_ticket = close_incident_in_remote(delta, data)
    relevant_keys_delta = delta.keys() & OUTGOING_MIRRORED_FIELDS.keys()
    if not relevant_keys_delta and not close_ticket:
        demisto.debug('No relevant changes to update in the remote system')
        return {}
```
I think it would be more appropriate to move all of this to the `update_remote_system_command` function, and add a new parameter to the `update_remote_incident` function - `close_ticket`.

mmhw (changes requested):
Good job!

mmhw (approved):
Great

PR #37863, opened 2024-12-30
Fix-wildfire-malware-playbook

ssokolovich:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-12496)

## Description
Fixed an issue with the condition that checks if the alert was generated in the last 24h.
The fixes includes the following:
- fixed the context key path with the right prefix
- fixed the unix timestamp to be in millieseconds instead of nano seconds

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Core/Playbooks/playbook-WildFire_Malware.yml
+++ b/Packs/Core/Playbooks/playbook-WildFire_Malware.yml
@@ -722,11 +722,19 @@ tasks:
       - - operator: greaterThan
           left:
             value:
-              simple: alert.autime
+              complex:
+                root: alert
+                accessor: autime
+                transformers:
+                - operator: division
+                  args:
+                    by:
+                      value:
+                        simple: "1000000"
             iscontext: true
           right:
             value:
-              simple: TimeNowUnix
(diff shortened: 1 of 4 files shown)


altmannyarden (approved):

Packs/Core/Playbooks/playbook-WildFire_Malware.yml
```diff
@@ -722,11 +722,19 @@ tasks:
       - - operator: greaterThan
           left:
             value:
-              simple: alert.autime
+              complex:
```
ssokolovich, line 725:
```suggestion
              complex:nice
```

Packs/Core/Playbooks/playbook-WildFire_Malware.yml
```diff
@@ -722,11 +722,19 @@ tasks:
...
+                root: alert
```
ssokolovich, line 726:
```suggestion
                root: alertdam
```

PR #37904, opened 2025-01-02
SplunkPy: update the mirror-in mechanism

ilappe:
## Related Issues
fixed: https://jira-dc.paloaltonetworks.com/browse/XSUP-44952

## Description

- changed the command used for mirror-in to be only `get-modified-remote-data`.
- changed the Splunk query for the mirror-in to be more efficient.
- raise exception in case of `get-remote-data`
- use caching for UserMapping
- take in account the potential collision when using the Enrichment and mirror-in deatures.


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -177,8 +183,61 @@ def get_splunk_user_by_xsoar_command(self, args):
                                             headers=['XsoarUser', 'SplunkUser'])
         )
 
+    def update_xsoar_user_in_notables(self, notables_data):
+        """In case of `should_map_user` is True, update the 'owner' in the notables to be the mapped XSOAR user.
```
DeanArbel, line 187:
```suggestion
        """In case of `should_map` is True, update the 'owner' in the notables to be the mapped XSOAR user.
```

Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -177,8 +183,61 @@ def get_splunk_user_by_xsoar_command(self, args):
...
+
+        Args:
+            notables_data (list[dict]): The notables to be updated.
+        """
+        if self.should_map:
```
DeanArbel, line 192: better to just return if not, this will remove the indentation that follows

DeanArbel (approved):
Impressive work, very well done. I'll follow up this review in another meeting

PR #38129, opened 2025-01-14
Use unique userId instead of userName (display name) for ZoomAsk output entry

ali-sawyer:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://xsoar.ideas.aha.io/ideas/FR-I-4373

## Description
It is necessary to use a unique identifier rather than display name to uniquely identify the user who responded to the `ZoomAsk` question, because users can change their Zoom display names at will.

## Must have
- [ ] Tests
- [ ] Documentation


mayyagoldman:
Hey @ali-sawyer ,  

Thank you so much for your effort on this PR and for taking on this task!  

- I’d like you to modify the code so that it utilizes `userID` together with `userName` rather than replacing it. Please adjust the code and make the necessary changes accordingly. For example, in the `check_and_handle_entitlement` function, `userName` is used appropriately in a string:  

```python
reply = message.get('reply', f'Thank you {user_name} for your response {text}.')
```  
Replacing `userName` with `userID` in this context would be unsuitable.  

- I also noticed in the PR description that you mentioned:  

*“It is necessary to use a unique identifier rather than the display name to uniquely identify the user who responded to the ZoomAsk question, because users can change their Zoom display names at will.”*  

However, the original ticket states:  
*“Modify Zoom Integration to return user ID when running ZoomAsk. Currently, the integration only returns the display name (`userName`) from the webhook event in the function `handle_zoom_response()`. The `userId` is required, and ideally, this would be saved to context for reference in other tasks.”*  

The ticket only requests adding `userId` without mentioning a specific issue with using `userName`. Could you provide an example where relying on `userName` causes a problem? This would help me better understand and confidently evaluate your changes.  

- Additionally, please consider whether these changes might introduce a breaking change. If so, this should be clearly reflected in the release notes.  

- Finally, please add tests to validate your new additions. This will ensure the modifications are robust and work as intended.  

Let me know if you need any further clarification or assistance! 😊

mayyagoldman (changes requested):
Hey @ali-sawyer ,  

Thank you so much for your effort on this PR and for taking on this task!  

- I’d like you to modify the code so that it utilizes `userID` together with `userName` rather than replacing it. Please adjust the code and make the necessary changes accordingly. For example, in the `check_and_handle_entitlement` function, `userName` is used appropriately in a string:  

```python
reply = message.get('reply', f'Thank you {user_name} for your response {text}.')
```  
Replacing `userName` with `userID` in this context would be unsuitable.  

- I also noticed in the PR description that you mentioned:  

*“It is necessary to use a unique identifier rather than the display name to uniquely identify the user who responded to the ZoomAsk question, because users can change their Zoom display names at will.”*  

However, the original issue states:  
*“Modify Zoom Integration to return user ID when running ZoomAsk. Currently, the integration only returns the display name (`userName`) from the webhook event in the function `handle_zoom_response()`. The `userId` is required, and ideally, this would be saved to context for reference in other tasks.”*  

The issue only requests adding `userId` without mentioning a specific issue with using `userName`. Could you provide an example where relying on `userName` causes a problem? This would help me better understand and confidently evaluate your changes.  

- Additionally, please consider whether these changes might introduce a breaking change. If so, [this should be clearly reflected in the release notes](https://xsoar.pan.dev/docs/documentation/release-notes#breaking-changes-version).  

- Finally, please add tests to validate your new additions. This will ensure the modifications are robust and work as intended.  

Let me know if you need any further clarification or assistance! 😊

ali-sawyer:
@mayyagoldman Responding to your 4 points in order:

1. Completed.
2. With the pack in its current state, when using the `ZoomAsk` script, the `ZoomAsk` response is returned as a War Room entry where the owner of the entry is the user's Zoom display name, for example, "user 123". The customer's goal is to get the _XSOAR user identity_ who responded to the `ZoomAsk` question. So using display name poses a problem, because the display name is any arbitrary value (e.g., "user 123") and doesn't correspond to an XSOAR user. But by using `userId` which is the user's email address in Zoom, we can confidently link the user to an XSOAR user, since XSOAR also uses email address for user identification. Hope that helps. If it's still unclear, I think looking at a test together will help clear it up.
3. IMO this is not a breaking change, it will only change the owner of the `ZoomAsk` response entry, i.e. it will slightly change the appearance of the entry.
4. My changes are very small. There is already a unit test function to test the `check_and_handle_entitlement()` function (`test_check_and_handle_entitlement()`), I just updated it slightly to add the new `user_id` parameter.

ali-sawyer:
I do not have a way to test `ZoomAsk`, so I have reached out to @sapirshuker to request her help with testing.

Packs/Zoom/Integrations/Zoom/Zoom.py
```diff
@@ -785,8 +785,9 @@ async def handle_zoom_response(request: Request, credentials: HTTPBasicCredentia
             account_id = payload['accountId']
             robot_jid = payload['robotJid']
             to_jid = payload['toJid']
-            user_name = payload['userName']
-            entitlement_reply = await check_and_handle_entitlement(action, message_id, user_name)
+            user_name = payload['userId']
```
mayyagoldman, line 788:
```suggestion
            user_name = payload['userName']
```

mayyagoldman (changes requested):
Hey @ali-sawyer ,  

Awesome! This looks great overall.  

I have a small comment regarding reverting `userName`—you left your old change there, so it seems it wasn’t fully reverted.  

Additionally, you mentioned: *"But by using userId, which is the user's email address in Zoom."* However, when I tested your code, the `userId` wasn’t an email address but rather some kind of ID. If this isn’t your intention, perhaps the payload contains a different field with the email, and you could use that instead? Unfortunately, I can’t share a screenshot of the `userId` we’re receiving, as it contains our admin ID, but let me know if you’d like further clarification on this.  

Also, please notice that there are now some conflicts in the PR—make sure to resolve them.  

Let me know if you have any questions! 😊

ali-sawyer:
@mayyagoldman Good catch! Added a lookup to get the user email address based on their user ID. I still don't have a way to test. If you can re-test now, that is greatly appreciated.

mayyagoldman (approved):

PR #36540, opened 2024-09-26
wallixbastion 3.0.0: add new commands, return id from created objects…

yoennb:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Description
This major release removes deprecated commands/arguments, adds a bunch of new command with support for null arguments and improves workflow by returning the ID of created objects in the command output. Details in the release notes.

## Must have
- [x] Tests
- [x] Documentation

PR #38193, opened 2025-01-16
Wallixbastion 3.1.0

yoennb:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
none

## Description
This version adds several commands requested by our customers, adds more documentation on command outputs which may not always be present, and removes an unnecessary parameter from wab-add/edit-notification.

It also includes an update of the WALLIX logo.

## Must have
- [x] Tests
- [x] Documentation


yoennb:
CI failing because of the mentioned breaking change. If this can cause issues, we can also mark the argument as deprecated (but I have not seen a field similar to the one that exists for commands)

talzich (commented):
Hi @yoennb, and thank you for your contribution!

Other than the comment on the `pack_metadata` file, can you please elaborate on why the breaking change is needed and why those arguments are no longer in use? 

Thanks again,
Tal

yoennb:
Hello @talzich, 

It was noticed during the [last PR review](https://github.com/demisto/content/pull/36540#discussion_r1807799178) that this argument could only take a single value. I eventually confirmed with our teams that there is no plan to change this API any time soon, thus the decision to simply remove it for the sake of clarity.
If you prefer, we can also leave this argument as non required and explicitly state in the description that the value will be ignored.

Thanks,

Yoenn

yoennb:
Hello @talzich,

Perfect then, we can schedule a call in the coming days. I am available on Slack to discuss about it.

talzich (approved):

PR #38329, opened 2025-01-26
Fix an issue in Tenable_io fetch 

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-45828

## Description
Improve the fetch code, by reseting the last index fetched when starting a new day.
Seem the audit logs are fetched per day, so need to reset the last index fetch when starting a new day.

## Must have
- [ ] Tests
- [ ] Documentation


AradCarmi (commented):
Nice!
Just a heads up, in case the day just changed, if we reset the `index_audit_logs` after checking if there are audit logs we will not ingest the current logs in the same **fetch cycle** only the one after it.

jbabazadeh (approved):
great job!

PR #38351, opened 2025-01-27
[EmailAskUserResponse] Add Suffix arg

DeanArbel:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-45386

## Description
Added a suffix argument to remove known terms from the response.

### DIFF
--- a/Packs/CommonScripts/Scripts/EmailAskUserResponse/EmailAskUserResponse.js
+++ b/Packs/CommonScripts/Scripts/EmailAskUserResponse/EmailAskUserResponse.js
@@ -7,6 +7,9 @@ if (res[0].Type==entryTypes.note || res[0].Type==entryTypes.file) {
     if (prefix && text.startsWith(prefix)) {
         text = text.slice(prefix.length);
     }
+    if (suffix && text.endsWith(suffix)) {
+        text = text.slice(0, -1 * suffix.length);
+    }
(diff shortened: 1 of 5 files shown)


Packs/CommonScripts/Scripts/EmailAskUserResponse/EmailAskUserResponse.js
```diff
@@ -11,9 +18,6 @@ if (res[0].Type==entryTypes.note || res[0].Type==entryTypes.file) {
                 .replace(/__NL____NL__/g,'')
                 .replace(/__NL__/g,'\n');
     response = text.replace(/<(?:.|\n)*?>/gm, '').trim("\n").split("\n")[0].trim();
-    if (prefix && response.startsWith(prefix)) {
```
dantavori, line 14: moving it up doesn't change the logic a bit?

dantavori (approved):

Packs/CommonScripts/Scripts/EmailAskUserResponse/EmailAskUserResponse.js
DeanArbel: It does, but this is actually an intended result, as removing it in the original place will remove the result of the white space trim. As such, this way achieves the intended behavior better

DeanArbel:
Force merging since there are no tests.

PR #38512, opened 2025-02-06
Panoaram get-jobs command

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-44394

## Description
Adding a fix for get-jobs command, since the given ID exists just in one device and we didn't handle it the code.
It was failing on ID not found since it was searching it in all devices instead of the correct one (and only one).
Actually try-except block was added.

## Must have
- [ ] Tests
- [ ] Documentation


jbabazadeh (commented):
great!

jbabazadeh (approved):

rshunim:
@yuvalbenshalom @jbabazadeh 
ForceMerge is needed here due to 4 TPBs which fail in the nightly too
(same errors here)

PR #38522, opened 2025-02-06
CloseInvestigationAsDuplicate - add an option for error when the incident was not closed

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-46392

## Description
Currently when a problem occurs and the incident is not closed, there is no error and the message is "0 incidents marked as duplicates".
I added a boolean argument for raise_error (default is False) and edited the msg to be more informative.

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/CloseInvestigationAsDuplicate/CloseInvestigationAsDuplicate.py
```diff
@@ -1,7 +1,27 @@
...
+    content = response[0].get('Contents') or response[0].get('contents')
+    return content == NO_INCIDENT_CLOSED_MSG
+
+
+def main():
+    current_incident_id = demisto.incidents()[0]['id']
+    duplicate_id = demisto.args()['duplicateId']
+    res = demisto.executeCommand("linkIncidents", {"incidentId": duplicate_id, "linkedIncidentIDs": current_incident_id,
+                                                   "action": "duplicate"})
+    raise_error = demisto.args().get('raise_error', False)
```
israelpoli, line 19:
```suggestion
    raise_error = argToBoolean(demisto.args().get('raise_error', False))
```

To prevent the value `false` from being received as a string, which would cause the condition in line 20 to evaluate as `True`.

Packs/CommonScripts/Scripts/CloseInvestigationAsDuplicate/CloseInvestigationAsDuplicate.py
```diff
@@ -1,7 +1,27 @@
...
+    if raise_error and is_incident_not_closed(res):
```
israelpoli, line 20: The failure message when the server returns the message **"0 incidents marked as duplicates"** should be present even when **raise_error** is **false**, just without raising an error.

rshunim: Done

israelpoli (approved):
Perfect

rshunim:
@yuvalbenshalom Force Merge is needed here due to CJ105 validation

PR #38582, opened 2025-02-10
Updated Panorama creds

noydavidi:
Updated panorama creds


noydavidi:
Im closing the pr in order the secret will merged into prod. Also, changed the device group param to "Shared" with approve from @jlevypaloalto.

PR #38850, opened 2025-03-03
fix tbp instead of doing image ocr on wikipedia to google

ilaredo:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [CIAC-12709](https://jira-dc.paloaltonetworks.com/browse/CIAC-12709)

## Description
The old version of the TBP tried to perform image OCR on Wikipedia, which has a lot of words. I changed it to Google, which has fewer words to parse. This change is acceptable because the main focus is on rasterize.

## Must have
- [ ] Tests
- [ ] Documentation


tkatzir (approved):
Great!

PR #38862, opened 2025-03-03
Flashpoint ignite - is_time_sensitive() related changes

ShacharKidor:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
relates: [CIAC-12908](https://jira-dc.paloaltonetworks.com/browse/CIAC-12908)

## Description
- Improved implementation of all the integration commands where the number of retries was lowered to 0 and the request timeout was lowered to 15 seconds when a command is called during an enrichment process.

## Must have
- [ ] Tests
- [x] Documentation


LHousehold:
Hello, this is Luke from Flashpoint. We've tested this change and it functions correctly and as expected. Thank you for your patience as we've been working through this.

On the technical side we have been discussing and we do have a concern that even a small network hiccup would cause the enrichment command to fail and harm the experience for customers. We'd like to suggest changing the 10 seconds to 15, and changing the 0 retries to 2 (the first retry happens instantly). Considering the feedback from yourselves and the customer, we agree that the current retry method is too heavy for enrichment purposes, but having no retries feels like a risk as well. If your team is okay with 15s and 2 retries, we can agree to this change immediately.

Additionally when we make these changes, we can also observe by how much this particular customers' timing is improved and discuss if we need to further reduce the timing or restrategize. 

What are your thoughts?

ShacharKidor:
Hi Luke,

Thanks for the update and for validating the fix.

I’ll increase the timeout to 15 seconds as suggested. However, regarding the retry count, we believe it's best to keep it at 0. During enrichment, minimizing the number of calls is critical to ensure optimal performance—additional retries could significantly degrade the user experience.

Let me know your thoughts.

Best regards,
Shachar

ShacharKidor:
@yuvalbenshalom  - Please force merge the PR. 
The [ST111] validation is failing for missing section headers, I didn't add the since this is a partner pack.

PR #38919, opened 2025-03-05
Fix duplicate alerts issue

RosenbergYehuda:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-47192

## Description
In the fetch process, we maintain a lookback to ensure we always overlap the latest fetch, which can potentially cause duplicates. We filter these duplicates using their timestamps.

In this case, we observed that the vendor unexpectedly changed timestamps, leading to duplicate alerts.

In this PR, we implemented a patch to update the saved alert timestamp in the lookback if it changes.

This only affects the lookback events we store and does not directly impact the created alerts.

## Must have
- [ ] Tests
- [ ] Documentation


JasBeilin (approved):
Nice catch!

PR #39101, opened 2025-03-18
Silent - Azure AD PIM role settings change

idovandijk:
## Status
Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12913

## Description
Silent playbook for "Azure AD PIM role settings change".

This playbook addresses the following alerts:

- Azure AD PIM role settings change

Playbook Stages:

Triage:

- Collect initial alert data regarding the Azure AD role settings modification.

Investigation:

- Analyze the impact of the role settings changes, categorizing them as high or moderate risk.
- Determine if the changes affect privileged roles in Azure AD.
- Correlate recent user activity with related security alerts.
- Assess the user's risk status in Azure AD and Cortex XDR.
- Evaluate the reputation of the IP address used for the changes.

Containment:

- Automatically revoke the user's active sessions across all applications.
- Present findings to an analyst for review and potential user account disablement.

Requirements:

For response actions, you need the following integrations:

- Cortex Core - Investigation and Response
- Microsoft Graph User

Azure_AD_PIM_role_settings_change_Mon_Mar_31_2025


Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
```diff
@@ -0,0 +1,1498 @@
...
+        }
+      }
+    note: false
+    timertriggers: []
+    ignoreworker: false
+    skipunavailable: true
+    quietmode: 0
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "18":
```
AdiPeret, line 469: The Entire Block for tasks 18, 17, 20. why are they dependent on the check if the user is risky? Checking to see if the user has any detections in Azure is another test we do in parallel to Azure risk score, isn't it?

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
```diff
@@ -0,0 +1,1498 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "30":
+    id: "30"
+    taskid: 8996bb86-10ac-4a74-9200-50e4ef4b0086
+    type: title
+    task:
+      id: 8996bb86-10ac-4a74-9200-50e4ef4b0086
+      version: -1
+      name: Caller IP Reputation
```
AdiPeret, line 897: why is that block part of the investigation? wouldn't it be better to have an early containment phase? or is it there only because we currently lack the ability to block IPs in Azure?

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
```diff
@@ -0,0 +1,1498 @@
...
+    isautoswitchedtoquietmode: false
+  "42":
+    id: "42"
+    taskid: 9a3038b3-65fd-4028-9476-cce5c54f1292
+    type: regular
+    task:
+      id: 9a3038b3-65fd-4028-9476-cce5c54f1292
+      version: -1
+      name: Close alert
+      description: Closes the current alert.
```
AdiPeret, line 1350: image
you close the investigation if the playbook does not find anything as False Positive if the playbook doesn't, find anything g it does not mean its a false positive.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
idovandijk: If the user doesn't appear as Risky in AAD, there is no point to look for Risk Detections, as they appear only for users that are risky - hence the dependency

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
idovandijk: Correct. We can't block the IP in Azure, but if the IP is returned with bad reputation it will increase the threshold used in our verdict

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
```diff
@@ -0,0 +1,1498 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "33":
+    id: "33"
+    taskid: f6dd5341-3aaf-4944-aacc-b4b8c88dc8a9
+    type: condition
+    task:
+      id: f6dd5341-3aaf-4944-aacc-b4b8c88dc8a9
+      version: -1
+      name: Analyze evidence
```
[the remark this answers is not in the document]
idovandijk: Yes, because we are counting the changes of each tier, I thought it would be more consistent to implement the rest of the checks as counts too. It looked better in the tasks and the logic. Greater than 0 means we have a detection. If you think it's confusing I can change it for the checks which will always be either 0 or 1

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
```diff
@@ -0,0 +1,1498 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "35":
+    id: "35"
+    taskid: acfd8198-e7cf-4933-97ce-e3adfa48a2a0
+    type: title
+    task:
+      id: acfd8198-e7cf-4933-97ce-e3adfa48a2a0
+      version: -1
+      name: Remediation
```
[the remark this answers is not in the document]
idovandijk: Technically - if we decide it's a TP we will perform soft remediation and then wait. The analyst task asks to review the evidence and decide whether disabling the account is also required. I think that since we are going with a threshold in this detector, adding more logic to distinguish "soft" TP vs "hard" TP in order to skip the analyst task will complicate it and won't necessarily create accurate results.
I could improve the wording in the task to make it more clear that disabling is not a suggestion but an option. WDYT?

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Azure_AD_PIM_role_settings_change.yml
idovandijk: Will look for an alternative reason, good point

idovandijk: Going with `Resolved - Auto Resolve`

idovandijk:
Pushed all changes and fixed conflicts

ShirleyDenkberg:
@AdiPeret Doc review completed.

AdiPeret (approved):

PR #39337, opened 2025-03-27
Pin github actions

james-otten-pan:
GitHub recommends pinning actions to a full length commit SHA, as this is currently the only method of using an action as an immutable release.

For more information refer to: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions


eyalpalo:
Hey @james-otten-pan 
As this is an infra internal decision, we decided not to approve this and the similar PRs with these changes.
Thank you for your willingness to contribute.

PR #39629, opened 2025-04-20
Coreir fix exception parsing

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Description
Found an issue that when running in cloud tenants (X8, XSAIm, etc...) we do not work with requests to run the API call, and therefore some error handling we added in the past needed to be modified.


eyalpalo (approved):

ShahafBenYakir:
The Cortex XDR Malware - Incident Enrichment test is failing but it seems like an issue with the instance and is investigated be @merit-maita here https://jira-dc.paloaltonetworks.com/browse/CIAC-13523.
Force merging.

PR #39693, opened 2025-04-23
CiscoUmbrellaReporting bugfix category filter

psybertool:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description

The validation of the categories which can be sent to the Cisco Umbrella API uses the python function isdigit() which returns only true for positive numbers. However, the Cisco Umbrella API accepts also negative numbers (to filter out categories explicitly)

So `[10, 34, 63]` is a valid list of categories but `[10, 34, -63]` is also valid. In the current version, the second argument throws a `ValueError()` although it is valid.

To be able to filter out categories explicitly, I added a `lstrip('-')` right before the `isdigit()` check which results in accepting strings of positive and also negative numbers.

## Must have
- [ ] Tests
- [ ] Documentation


mmhw (approved):
Hi @psybertool ,
Good work :)

This looks great.

Thanks again!

PR #39540, opened 2025-04-09
Reintroduce earliest_occurrence_time_fieldname and latest_occurrence_…

pmunaretto:
…time_fieldname parameters in SplunkPy integration

## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [X] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: n/a

## Description

In version 2.1.10 of the SplunkPy integration, the ability to customize the `earliest_fetch_time_fieldname` and `latest_fetch_time_fieldname` parameters in the oneshot fetch search was removed (#14365). These parameters are essential for dispatching the fetch search based on `index_earliest` and `index_latest` time filters, rather than the default `earliest_time` and `latest_time`.

Although the code for these parameters was reintroduced in version 2.3.0 (#16576), the corresponding parameters in the configuration file were not, making it impossible to set these values. The new parameter names are `earliest_occurrence_time_fieldname` and `latest_occurrence_time_fieldname`.

This pull request aims to reintroduce the `earliest_occurrence_time_fieldname` and `latest_occurrence_time_fieldname` parameters in the SplunkPy integration's configuration file (`SplunkPy.yml`) and ensure they are properly utilized in the code (`SplunkPy.py`).

By reintroducing these parameters, users can customize the oneshot fetch search based on specific time filters, enhancing the flexibility and usability of the SplunkPy integration.

## Must have
- [X] Tests
- [X] Documentation

PR #39860, opened 2025-05-08
[SplunkPy] New param notable_time_source

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/content/pull/39540

## Description
Added support for **Advanced: Time type to use when fetching events** parameter that defines which timestamp will be used to filter the events:
 *creation time*: filters based on when the event actually occurred (current behavior).
 *index time (beta)*: *beta feature* – filters based on when the event was ingested into splunk.  
  this option is still in testing and may not behave as expected in all scenarios.  
  when using this mode, you can set the parameter `fetch backwards window for the events occurrence time (minutes)` to `0`, since indexing time ensures no delaybased gaps.

## Must have
- [ ] Tests
- [ ] Documentation


ilappe (approved):
Grate work!

PR #39891, opened 2025-05-11
Red canary domain context

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-50484

## Description
Fixed an issue where in some cases the get-detection command did not return the domain info to context

## Must have
- [x] Tests

### DIFF
--- a/Packs/RedCanary/Integrations/RedCanary/RedCanary.py
+++ b/Packs/RedCanary/Integrations/RedCanary/RedCanary.py
@@ -205,7 +205,7 @@ def process_timeline(detection_id):
     ips = []
     processes = []
     for activity in res:
-        if activity.get('type') not in ["activity_timelines.LatestIndicationSeen", "activity_timelines.EventActivityOccurred"]:
+        if activity.get("type") not in ["activity_timelines.LatestIndicationSeen", "activity_timelines.EventActivityOccurred"]:
             continue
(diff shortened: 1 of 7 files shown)


Packs/RedCanary/Integrations/RedCanary/RedCanary.py
```diff
@@ -206,14 +205,13 @@ def process_timeline(detection_id):
     ips = []
     processes = []
     for activity in res:
-        if activity["type"] != "activity_timelines.ActivityOccurred":
+        if activity.get('type') not in ["activity_timelines.LatestIndicationSeen", "activity_timelines.EventActivityOccurred"]:
```
JudahSchwartz, line 208: what happened with ActivityOccurred? never relevant anymore?
where are these two values from?

JudahSchwartz (approved):
nice!

Packs/RedCanary/Integrations/RedCanary/RedCanary.py
Shellyber: I don't know. I don't have access to the product. 
But what I used is from a customer response (deleted secrets, of course)
It is from his response, so I could replicate the issue via unittest.
I didn't delete the old value cause I can't be sure it is irrelevant, this way we're handling both cases

PR #39897, opened 2025-05-12
XSUP-50451/azure-sentinel-list-incidents

sapirshuker:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-50451

## Description
Based on [this](https://github.com/demisto/content/pull/30124) change, the implementation:
Distinguish between a default value and a maximum value and preserve the behavior of the fetch function.

DEFAULT_LIMIT = 50 - is defined to be default limit in the azure-sentinel-list-incidents and the azure-sentinel-list-incident-comments commands.
FETCH_MAX_LIMIT = 20 - is defined to be default limit in the fetch and implemented as max limit in the fetch
COMMAND_MAX_LIMIT = 200 - is defined to be max limit in the azure-sentinel-list-incidents
MAX_INCIDENT_COMMENT_LIMIT = 50 is defined to be max limit in the azure-sentinel-list-incident-comments

## Must have
- [ ] Tests
- [ ] Documentation


eyalpalo (changes requested):
Please elaborate on how you choose the values for the limit vars.

sapirshuker:
DEFAULT_LIMIT = 50 - is defined to be default limit in the azure-sentinel-list-incidents and the azure-sentinel-list-incident-comments commands.
FETCH_MAX_LIMIT = 20 - is defined to be default limit in the fetch and implemented as max limit in the fetch
COMMAND_MAX_LIMIT = 200 - is defined to be max limit in the azure-sentinel-list-incidents
MAX_INCIDENT_COMMENT_LIMIT = 50 is defined to be max limit in the azure-sentinel-list-incident-comments 

The implementation:
Distinguish between a default value and a maximum value and preserve the behavior of the fetch function.

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -1368,7 +1467,7 @@ def fetch_incidents(client: AzureSentinelClient, last_run: dict, first_fetch_tim
 
     """
     # Get the last fetch details, if exist
-    limit = demisto.params().get("limit", DEFAULT_LIMIT)
+    limit = min(arg_to_number(demisto.params().get("limit")) or FETCH_MAX_LIMIT, FETCH_MAX_LIMIT)
```
eyalpalo, line 1470:
```suggestion
    limit = min(arg_to_number(demisto.params().get("limit")) , FETCH_MAX_LIMIT)
```
The param has default value

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
```diff
@@ -911,15 +978,19 @@ def list_incidents_command(client: AzureSentinelClient, args, is_fetch_incidents
         A CommandResult object with the array of incidents as output.
     """
     filter_expression = args.get("filter")
-    limit = min(DEFAULT_LIMIT, int(args.get("limit")))
     next_link = args.get("next_link", "")
+    limit = min(arg_to_number(args.get("limit")) or DEFAULT_LIMIT, COMMAND_MAX_LIMIT)
```
eyalpalo, line 982:
```suggestion
    limit = min(arg_to_number(args.get("limit")), COMMAND_MAX_LIMIT)
```
if we have the args from the command - we have default value in the yml. If from the fetch, we will always have value from the fetch_incidents method. Hence,

eyalpalo (changes requested):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
sapirshuker: This change will raise mypy error: 
Type "int | None" is not assignable to type "int" since min needs two int and arg_to_number return value is int|None

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
sapirshuker: This change will raise mypy error: 
Type "int | None" is not assignable to type "int" since min needs two int and arg_to_number return value is int|None

eyalpalo: So we can at list set the default value like this:
demisto.params().get("limit", FETCH_MAX_LIMIT)

eyalpalo (approved):

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
sapirshuker: The mypy error is because arg_to_number return None | int and min method want to get an int - so we leave it as is and didn't change it after a discussion.

Packs/AzureSentinel/Integrations/AzureSentinel/AzureSentinel.py
sapirshuker: The mypy error is because arg_to_number return None | int and min method want to get an int - so we leave it as is and didn't change it after a discussion.

PR #40144, opened 2025-05-30
Armis Release 1.2.1

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

#### Armis Event Collector

- Updated the API read timeout to 3 minutes to reduce overall request failures.
- Updated the value of `includeTotal` API parameter to `false` to improve performance.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/Armis/.pack-ignore
+++ b/Packs/Armis/.pack-ignore
@@ -30,4 +30,5 @@ p
 aql
 uuids
 unhandled
-
+linting
+programmatically
--- a/Packs/Armis/Integrations/Armis/Armis.py
+++ b/Packs/Armis/Integrations/Armis/Armis.py
@@ -1,6 +1,5 @@
 """IMPORTS"""
 
-
 import pytz
 import urllib3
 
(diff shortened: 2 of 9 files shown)


mmhw (commented):
Hi @crestdatasystems,

Could you please clarify what it means when `includeTotal` is set to `false`?
Specifically:
- Does this mean that less data will be returned to the user?
- Is there any impact on backward compatibility that we should be aware of?

Thanks again

crestdatasystems:
> Hi @crestdatasystems, Thank you for your contribution!
> 
> Could you please clarify what it means when `includeTotal` is set to `false`? Specifically:
> 
> * Does this mean that less data will be returned to the user?
> * Is there any impact on backward compatibility that we should be aware of?
> 
> Please feel free to reach out to me with any questions - I'm available here or on slack :) Thanks again

Hi @itssapir, Thanks for reviewing the PR.

The intention to set the `includeTotal` to `false` is to not return the total count of the records. If set to `true`, API computes to get the total count of the records which increases delay of the API call. Also, this doesn't have any impact on the code which is parsing the API response.

PR #40170, opened 2025-06-03
[Pack][Cyberwatch] Add Fetch-Incidents

tarraschk:

Done ✅ 

## Status
- [ ] In Progress
- [X] Ready

## Related Issues

None, add fetch-incidents for Cyberwatch

## Description

Add fetch incidents for Cyberwatch

## Must have
- [X] Tests
- [X] Documentation


tarraschk:
Need a few updates @AmineHazi, see you on Teams for the linter with `demisto-sdk pre-commit` and the Code Coverage percentage

tarraschk:
> Need a few updates @AmineHazi, see you on Teams for the linter with `demisto-sdk pre-commit` and the Code Coverage percentage

Ready for review

itssapir:
Hi @tarraschk,

I started looking at the fetch code, and I see that you are using client.get_assets() for the fetch, which defaults to fetching all the pages from the server, and only later do you filter out the older ones.
This sounds like it would result in a lot of unnecessary API calls and processing. Do the filters contain some option to give a minimum timestamp? This would be highly recommended to be used.
Alternatively, am I missing something here?

There also seems to be possible incident misses. The last run time is always being set to the current fetch time, but sometimes assets are skipped when we reach the max fetch value.
Does the assets response come sorted by time? If so, you can use the time of the last asset that is being pushed instead.

In any case, you would want to ensure the next fetch continues from where the previous one stopped, so you don't miss any assets and/or fetch duplicates.

Thanks again

tarraschk:
Hey @itssapir thank you for your message!

Unfortunately, the Cyberwatch API does not provide a direct route for new "detected CVEs on assets". It's either first you fetch all the assets, and then iterate over their CVEs, or first you fetch all present CVEs and then iterate over their assets.

Also, there is no way at the moment to filter or sort over the new CVEs on the assets "since date XXX" and that is why we have to loop over all items.

What we do usually when we have to make a list of "incidents" (1 CVE found on 1 asset) with other products than Cortex is we fetch all assets as it actually makes less API calls than fetching all present CVEs. And then we iterate over each asset CVEs.

So what we do here is:
- fetch all assets (usually people have less assets than CVEs lol)
- for each asset, look at its CVE entries and compare that to some filters that the user can use
- each CVE that matches the filters is then added to the incident list (1 incident = 1 CVE on 1 asset)

We have added `if len(incidents) >= max_fetch:` so that we also stop when we have reached max number of incidents instead of continuting the iteration over all CVEs on all assets. This is a little optimization here, not perfect but still better than nothing.


> In any case, you would want to ensure the next fetch continues from where the previous one stopped, so you don't miss any assets and/or fetch duplicates.

Actually regarding this, we talked with the customers and the risk of missing CVEs since last time is only supposed to really happen the first time you execute the script. They are okay to miss some entries on the first executions. Our choices here regarding the implementation was so that:
- first execution might miss CVEs, and it's okay
- then you are supposed to choose something like a period between 1 and 4 hours to execute the fetch_incident command (4 hours when you have a huge number of assets)

Please note also that the script has been developed with an actual customer instance as an example on our side, to be sure that everything works properly.

If you still want us to make some change in the code with these details, please tell us, we will adapt the PR :)

itssapir:
My main concern is regarding missing incidents from reaching the max fetch and then still marking the next run time as the current time. Doing it that way means all assets that were skipped by the max fetch, in every single fetch cycle, will be missed.

If you want to keep using the timestamp as the mechanism for filtering duplicates, you could perhaps sort the API response by time after you receive it, and then save the time of the last pushed asset to the "next run time" instead of the current time. This could be a rather compute intense operation if the asset list is very large though.

Alternatively, we could think of a different approach to filtering duplicates that would not miss entries when the max fetch is invoked. For instance, if you could cause/spot any type of consistent ordering in the API response, you could simply keep note of the last incident that was pushed and start from that point in the next fetch.

No duplicates and no missed events are two criteria we wish to maintain across all fetch-incident enabled integrations in the marketplace, so we do need to address the gap in this case.

If you're still not certain, we can also discuss this on the DFIR slack if you'd find it more convenient.
Thanks again :)

tarraschk:
Okay, I understand better your comment with the "No duplicates and no missed events are two criteria we wish to maintain across all fetch-incident enabled integrations in the marketplace, so we do need to address the gap in this case.".

Unfortunately the Cyberwatch API has a very huge limitation regarding this approach (basically new CVEs can be added to assets with no impact on the analyzed_at timestamps of the asset).

Thus, if there is really a requirement on "no duplicates and no missed events", the only real way to do so would be:

1. we fetch the list of all assets (total of API calls = total number of assets / 500)
2. we fetch details for all assets, including their CVEs (total of API calls = total number of assets)
3. we filter all CVEs based on user filters + `detection date > last_run`
4. we sort the results based on the detection_date with millisecond precision
5. we then process the results in the detection_date order ascending
6. if at some point when looping on the detection_dates we reach the max_fetch_incident, we set the last_run to this detection_date so that the next fetch-incident will continue with no missed incident

The only issue here is that if will always have to go through all assets.

Would that be okay for you?

itssapir:
Hi @tarraschk,
We discussed this a bit more. Also, there is no strict requirement for the incidents to be pushed sorted by time as long as no incidents are missed.

There are 2 different options to handle the fetching that I think would work for you:
1. Similar to what you suggested, you would fetch the information for all available assets, then create a list of all new CVE's sorted by creation time.
You can then take the first "Max fetch" entries to be pushed, while saving the remaining ones for later (using next run)
This means in the following fetch, you will first use your cached entries to be pushed, possibly not posting any API calls until you deplete the cache, then you would start a new fetch taking CVE's that are older than the oldest CVE that was pushed so far.

2. As we discussed in our call, you can do the round-robin approach, pushing CVE's in ascending asset id order, up to the max fetch. Then in the following fetch you continue fetching from the last run id.

I think both approaches should give similar average performance results (with the first approach taking longer for the first fetch, but shorter followups) with the first approach being a bit more complex but giving you ordering by creation date.

**its up to you which one to go with, or if you have another suggestion feel free to let me know.**

**Regardless, I strongly suggest you advise users to set a high enough fetch interval since we expect most fetches to post a lot of API requests.**

Again, feel free to reach out to me with any more questions/comments.
Thanks.

itssapir:
Hi @tarraschk,
Any updates on this PR?
Thanks

tarraschk:
Hello @itssapir yes we should complete the PR within 2 weeks

tarraschk:
Hey @itssapir we just sent an updated version of the code that should not miss any incident :)

Packs/Cyberwatch/Integrations/Cyberwatch/Cyberwatch.py
```diff
@@ -609,7 +662,179 @@ def iso8601_to_human(iso8601_str, default_value=""):
...
+                break
+        # Reset resume pointers after first asset is handled
+        resume_server_id = None
+        resume_cve_id = None
+
+        if hit_limit:
+            break
+
+    if hit_limit:
+        cursor = _build_cursor(
```
itssapir: This looks overly complicated. setLastRun and getLastRun support the use of saving dicts and lists, you should be able to simply create a dict with the needed data and save it as is instead of encoding and decoding it each time.

Its up to you if you want to try and change this though since your approach will work too, if not just mark this as resolved.

Packs/Cyberwatch/ReleaseNotes/1_0_3.md
```diff
@@ -0,0 +1,5 @@
+#### Integrations
+
+##### Cyberwatch
+- Add fetch-incidents.
```
itssapir:
```suggestion

- Added support for ***fetch-incidents*** to periodically fetch new CVEs on assets.
```
Add some verbosity to the RN. You can take the suggestion or rephrase it as you see fit.

itssapir:
Hi @tarraschk,
The changes look great overall, nice!

I had just a couple of minor comments, please see above.

Let me know if you have any questions/comments :)
Thanks.

AmineHazi:
Hi @itssapir,
I just made a mistake while pushing the last changes, the mistake is fixed on our fork, but i need to re-open the MR,
Sorry for the trouble !

tarraschk:
Hey @itssapir, there was a mistake when rebasing the previous code. Should we open a new PR to send the proper code, or could you re-open this PR?

Sorry for the inconvenience.

itssapir:
Hey @tarraschk,
For some reason the PR is marked as merged. You would need to open a new PR so we can continue.

Side-note: you don't need to go through the trouble of complicated rebases and force pushing. Just push your changes/fixes themselves, and if you need to merge with the content master just push a merge commit or use the built in github functionality.
This will keep the PR history clear and avoid issues.
The final merge of the PR squashes everything to a single commit anyway, so any rebase effort here is irrelevant.

PR #40200, opened 2025-06-05
Enhancements for Panorama integration commands

aneeshamore:

## Status
- [x] In Progress
- [ ] Ready

## Description
Updated pan-os-configuration-hygiene-check commands to include details about non-conforming profiles. Also updated the pan-os-platform-get-system-info command to add the global_protect_client_package_version to its output.

## Must have
- [x] Tests
- [x] Documentation


Packs/PAN-OS/ReleaseNotes/2_4_0.md
```diff
@@ -0,0 +1,11 @@
...
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Added support for *return_nonconforming_profiles* argument in the **pan-os-hygiene-check-vulnerability-profiles** command.
+- Added support for *return_nonconforming_profiles* argument in the **pan-os-hygiene-check-url-filtering-profiles** command.
+- Updated the **pan-os-hygiene-check-url-filtering-profiles** to check an extended best-practice list of URL categories to block: abused-drugs, adult, command-and-control, compromised-websites, gambling, grayware, hacking, malware, phishing, questionable, ransomware, scanning-activity, weapons
+- Added support for *return_nonconforming_profiles* argument in the **pan-os-hygiene-check-spyware-profiles** command.
+- Added support for **pan-os-hygiene-check-security-zones** to evaluate zones configured on both Panorama and connected firewalls (Panorama instances only)
+- Updated the Docker image to: *demisto/pan-os-python:1.0.0.3545654*.
```
itssapir:
```suggestion
- Added support for the *return_nonconforming_profiles* argument in the following commands:
    - ***pan-os-hygiene-check-vulnerability-profiles***.
    - ***pan-os-hygiene-check-url-filtering-profiles***.
    - ***pan-os-hygiene-check-spyware-profiles***.
- Updated the ***pan-os-hygiene-check-url-filtering-profiles*** command to check an extended best-practice list of URL categories to block: abused-drugs, adult, command-and-control, compromised-websites, gambling, grayware, hacking, malware, phishing, questionable, ransomware, scanning-activity, weapons
- Added support for ***pan-os-hygiene-check-security-zones*** to evaluate zones configured on both Panorama and connected firewalls (Panorama instances only)
- Updated the Docker image to: *demisto/pan-os-python:1.0.0.3545654*.
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -10701,6 +10724,24 @@ def check_vulnerability_profiles(
                 minimum_alert_severities=minimum_alert_severities,
```
itssapir: This loop logic seems to be identical across multiple methods, shouldn't it (or a part of it) be moved to a separate function?

Packs/PAN-OS/ReleaseNotes/2_4_0.json
```diff
@@ -0,0 +1,4 @@
+{
```
itssapir, line 1: These are not actually breaking changes, we can remove the BC note.

itssapir (changes requested):
Hi @aneeshamore,
Thank you for your contribution.

Please see my comments below.

A couple other things:
- Please add unit-tests for the new command argument and for the new check-security-zones logic.
- The pre-commit is failing on ruff formatting, you can fix this by running `demisto-sdk pre-commit -g` locally and committing the ruff changes that it creates.
- Regarding the updated best practice list, can you direct me to where we have this documented in PAN-OS?

Packs/PAN-OS/ReleaseNotes/2_4_0.json
mikejrizzo: @itssapir This change doesn't break existing usage of the command itself, however it's a more strict check than what existed before.  People using this command with the update may find their firewalls failing the hygiene check where it had passed before, so we figured a BC would help draw attention to this.

Do you still think we should remove it?

mikejrizzo:
@itssapir I've made improvements to the unit tests you mentioned above and re-run pre-commit on the files we've modified.  During the local pre-commit run it modified many other files we've not modified, such as playbooks and incident field files.  Since this work is not related to general formatting I've omitted those.

Please let me and @aneeshamore know if there's other adjustments needed to complete this PR.  Thanks!

Packs/PAN-OS/ReleaseNotes/2_4_0.json
itssapir: Its okay we can keep the BC note to improve the visibility of this change.

itssapir:
Looks great overall. Please have a look at the remaining comment regarding the duplicated logic.
There are some conflicts from the latest PAN-OS release that need to be resolved. Lets ensure those are resolved and everything passes in the automatic checks then we'll be able to move forward and set up time for a short demo.

richardbluestone:
Docs look good. Review done

mikejrizzo:
@itssapir We've merged in the new changes that were published to `master` and it looks like all the builds are succeeding.

As for the duplicated logic, please see my response to your in-line comment and let me know if we need to follow up further.

Thanks!

Packs/PAN-OS/Integrations/Panorama/Panorama.py
mikejrizzo: @itssapir This loop logic looks very similar in its three instances, but there are minor differences as to the type of profile it iterates over, and it also modifies the `issues` object that's defined elsewhere in the function and added to later.  Is it necessary to break this out as a modularized function?  If not, this seems simpler.

itssapir: While the issue is pre-existing, as the 3 functions already practically perform the same logic with slight variations, we don't want to keep adding more duplications.

All of the parameters appear to be simple enough to pass to a generic function, and it would make future changes/fixes to this logic easier to handle.

Side comment, what is the point of the `non_conforming_profiles` list? I don't see it being used anywhere other than being populated. It can probably be removed as well.

mikejrizzo: @itssapir I've consolidated the separate HygieneLookups class functions into one.  The resulting behavior matches what I showed in our demo, but now all three commands are ultimately handled by one function.

itssapir (approved):
Looks great! Thank you for taking the extra effort and consolidating the 3 functions entirely instead of only the new logic, it is much appreciated.

I will let you know once the internal PR has been merged.

PR #40226, opened 2025-06-10
DBOTFindSimilarAlerts unable to handle alerts with a list containing 1 element

screencoffee:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
This pull request fixes a small issue on XSIAM, where almost all alert fields that has been provided in the "similarTextFields" argument are skipped(see screenshot below). 
The problem seem to stem from the length check not handling datatype, as all of the fields that have been skipped are lists with only one element.

image

This was fixed in a copied script that, where an additional check for the fields datatype is added.
I must admit I am not quite sure what ramifications this has, but this does work on when testing locally:
image

It might be beneficial to perform the check on each of elements in the lists, but I think someone with more knowledge of the inner workings of this script should make that decision.
## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/Scripts/DBotFindSimilarIncidents/DBotFindSimilarIncidents.py
+++ b/Packs/Base/Scripts/DBotFindSimilarIncidents/DBotFindSimilarIncidents.py
@@ -478,7 +478,7 @@ class Model:
                     skip_reason = FIELD_SKIP_REASON_FALSY_VALUE.format(field=field, val=val)
                 elif valid_types and not isinstance(val, valid_types):
                     skip_reason = FIELD_SKIP_REASON_INVALID_TYPE.format(field=field, valid=valid_types, type=type(val))
-                elif len(val) < 2:
+                elif isinstance(val, str) and len(val) < 2:
                     skip_reason = FIELD_SKIP_REASON_TOO_SHORT.format(field=field, val=val, len=len(val))
                 elif isinstance(val, list) and all(not x for x in val):
                     skip_reason = FIELD_SKIP_REASON_LIST_OF_FALSY_VALS.format(field=field, val=val)


dantavori:
Hey @screencoffee, thank you for your contribution.
We are aware of the behavior you raised. Note that this was intentionally the original design and it's been like that for years, but we also understand your point.
Since this script is widely used, we will conduct an internal discussion about it and will update you when we have a conclusion.

screencoffee:
> Hey @screencoffee, thank you for your contribution. We are aware of the behavior you raised. Note that this was intentionally the original design and it's been like that for years, but we also understand your point. Since this script is widely used, we will conduct an internal discussion about it and will update you when we have a conclusion.

Thank you for the update, i understand this being intentional.
It unfortunately makes it more or less unusable on XSIAM. One example is the "Large Upload" playbook, as it is used in the first task which consequently will fail because most fields used are skipped.

I hope you find a solution, as we would love to make use of the script in our playbooks.

itssapir:
Hi @screencoffee,
After further internal discussion, it was decided that we need to address this problem directly. This issue will be tackled by our dev team to ensure the solution goes through our full development cycle and does not cause any issues with existing use-cases.

I'm closing this PR for now.
Thanks again for you contribution.

screencoffee:
> Hi @screencoffee, After further internal discussion, it was decided that we need to address this problem directly. This issue will be tackled by our dev team to ensure the solution goes through our full development cycle and does not cause any issues with existing use-cases.
> 
> I'm closing this PR for now. Thanks again for you contribution.

Thank you for the update! 
That is also my preferred outcome of this PR, as I was not able to fully test for issues.

PR #40239, opened 2025-06-11
Fixed an issue where duplicated incidents were created in searchalerts

eyalpalo:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-50047)

## Description
Fixed an issue where duplicated incidents were created in searchalerts

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
```diff
@@ -201,6 +205,16 @@ def search_incidents(args: Dict):  # pragma: no cover
         demisto.debug(f"before filtering {len(current_page_found_incidents)=} {args=} {page=}")
         more_pages = len(current_page_found_incidents) == page_size
 
+        # dedup incidents
+        demisto.debug(f"before dedup process with {len(current_page_found_incidents)=}")
+        current_page_found_incidents = [
+            incident for incident in current_page_found_incidents if incident.get("id") not in incidents_ids
```
ShahafBenYakir: Looking at this, it seems we are claiming that BE might return the same incident twice?

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2_test.py
```diff
@@ -412,3 +412,23 @@ def test_main_flow_with_limit(mocker, amount_of_mocked_incidents, args, expected
 
     assert return_results_mocker.called
     assert len(return_results_mocker.call_args[0][0].outputs) == expected_incidents_length
+
+
+def test_search_with_duplicates(mocker):
+    """
+    Given: Duplicated incidents from executing getIncidents.
+    When: Running the command
+    Then: Return incidents without duplications.
```
ShahafBenYakir: Let's make the given when then clearer - what exactly is given? What exactly is performed behind the scenes (3 pages search?) What do we expect to happen to what object (incident X that was returned twice should be removed from final list)

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
eyalpalo: Yes indeed, the problem here is an edge case where we call the API twice or more, and in between, new incidents are created... In this case, we request the second page, which is already different as we have new incidents in the system that push the other incident's order...

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2.py
```diff
@@ -194,6 +194,11 @@ def search_incidents(args: Dict):  # pragma: no cover
     demisto.debug(f"Amount of incidents after filtering = {len(all_found_incidents)} before pagination")
     page = STARTING_PAGE_NUMBER
 
+    if all_found_incidents:
+        first_incident = all_found_incidents[0]
+        args['todate'] = first_incident.get('created')
+        demisto.info(f"Setting todate argument to be {first_incident.get('created')} to avoid duplications")
+
```
ShahafBenYakir:
```suggestion
    if all_found_incidents and not args.get('todate') :
        first_incident = all_found_incidents[0]
        args['todate'] = first_incident.get('created')
        demisto.info(f"Setting todate argument to be {first_incident.get('created')} to avoid duplications")

```

ShahafBenYakir: Let's also add a comment saying that in case we don't have a todate, we add one to make sure the search will not change between two API requests.

Packs/CommonScripts/Scripts/SearchIncidentsV2/SearchIncidentsV2_test.py
```diff
@@ -439,3 +439,36 @@ def test_query_argument_with_unicode_escape(mocker):
...
+    page1_incidents = [{"created": f"2025-01-01T10:0{i}:00Z", "id": i} for i in range(5)]
+    # Page 2 returns fewer, triggering end-of-pagination
+    page2_incidents = [{"created": f"2025-01-01T09:59:00Z", "id": 101}]
+
+    # Mock execute_command behavior
+    execute_command_mocker = mocker.patch.object(SearchIncidentsV2, "execute_command", side_effect= [
+        # 1st call: initial getIncidents
+        [{"Contents": {"data": page1_incidents}}],
+        # 2nd call: getIncidents inside loop (page2)
+        {"data": page2_incidents},
```
ShahafBenYakir: How come execute_command returns one kind of object in the first run and another in the second?

eyalpalo: running execute_command without extract_contents in the second execution creates the difference...

PR #40308, opened 2025-06-19
New pan-os-get-certificate-info command

aneeshamore:
Added a new command to get the certificate information from Panorama or firewall as seen under Certificate Management:
  - ***pan-os-get-certificate-info***


Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
     )
 
 
+def pan_os_get_certificate_info_command(topology: Topology, args: Dict) -> CommandResults:
+    """
+    Get certificate information from PAN-OS device
+
+    Args:
```
rshunim, line 14826: Fix docs with the correct args of this function (topology and args instead of target)

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+    Returns:
+        CommandResults: Certificate information
+    """
+    CERT_DETAILS = []
+    SHOW_LOCAL_CERTS = "request certificate show"
+    SHOW_CONFIG_RUNNING = "show config running"
+    SHOW_CONFIG_PUSHED_TEMPLATE = "show config pushed-template"
+    PREDEFINED_CERTS_XPATH = "/config/predefined/certificate"
+
+    def expiration_status_check(cert_expiration: datetime) -> str:
```
rshunim: This function should be moved to the outer scope. Considering this function for further development in the future

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+        elif cert_expiration < now + timedelta(days=30):
+            return "Expiring in 30 days"
+        elif cert_expiration < now + timedelta(days=60):
+            return "Expiring in 60 days"
+        elif cert_expiration < now + timedelta(days=90):
+            return "Expiring in 90 days"
+        else:
+            return "Valid"
+
+    def consolidate_all_certificates(cert_list: List, cert_type: str, device: str, devices_using_certificate: Optional[List] = None):
```
rshunim: Here same comment as above

(This function should be moved to the outer scope. Considering this function for further development in the future)

In addition, I see the finction has a return value, so please add it to the function's signature

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+            subject_elem = cert.find("subject")
+            if not_valid_after is not None and not_valid_after.text is not None:
+                cert_expiration = datetime.strptime(not_valid_after.text, "%b %d %H:%M:%S %Y %Z")
+                expiration_status = expiration_status_check(cert_expiration)
+            else:
+                cert_expiration = None
+                expiration_status = None
+
+            cert_details_dict = {
+                "name": cert.get("name"),
```
rshunim: what does happen in case of "name" : None?
if it's ok so leave it with '.get' access.
in case of 'name' has to be valuable and 'cert' must contain this 'name' field it's better to use cert['name'], since we want to get a failure if somehow 'cert' does not contain the 'name' field

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+    """
+    Get certificate information from PAN-OS device
+
+    Args:
+        target: Command arguments
+
+    Returns:
+        CommandResults: Certificate information
+    """
+    CERT_DETAILS = []
```
rshunim: if CERT_DETAILS is just a list only for the 'pan_os_get_certificate_info_command' function scope' so let's change it to lower case since it's not a constant -> cert_details (maybe 'cert_details_final' or something like this)

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+    SHOW_LOCAL_CERTS = "request certificate show"
```
rshunim: Here regarding these constants (all the 4), please move them to the top of the file where the globals are defined.
for example: the string "show config pushed-template" already appears in line 10823,
so it's a good practice to define all constants (globals) at the top of the file so everyone can use it, and in case of changing, we need to change this string just in one place in our code

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+                # Process pushed certificates
+                if response_pushed is not None and hasattr(response_pushed, "get") and response_pushed.get("status") == "success":
+                    template_config = response_pushed.find(".//template")                    
+                    template_stack_config = response_pushed.find(".//template-stack")
+                    if template_config.find(".//certificate"):
+                        certificate = template_config.find(".//certificate")
+                        devices_using_certificate = [
+                            template.devices for template in templates if hasattr(template, "devices") if template.devices
+                        ]
+                    elif template_stack_config.find(".//certificate") is not None:
```
rshunim:
```suggestion
                    elif certificate := template_stack_config.find(".//certificate"):
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+                        certificate = template_stack_config.find(".//certificate")
```
rshunim: This line is redundant when you fix the condition as I suggest

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14598,172 @@ def pan_os_get_master_key_details_command() -> CommandResults:
     )
 
 
+def pan_os_get_certificate_info_command(topology: Topology, args: Dict) -> CommandResults:
```
rshunim, line 14822: This function is about 160 line, except of moving the nested 2 function to the outer scope
consider dividing this main function ("pan_os_get_certificate_info_command") to a few sub-functions which every sub function has its job, then this main function should call them in order.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -9828,6 +9828,39 @@ script:
...
       required: true
       isArray: false
+  - name: pan-os-get-certificate-info
+    description: 'Gathers name, expiration date and expiration status of certificates configured locally on a Firewall or pushed from Panorama, as seen under Certificate Management.'
+    arguments:
+    - description: Show only expired certificates.
+      name: show_expired_only
+      type: Boolean
+      required: false
+      isArray: false
```
rshunim: required and isArray by default are false so you can remove them

rshunim:
@jlevypaloalto @richardbluestone 
Please review too this PR
Thanks!

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aneeshamore: A certificate must have a name. But that is a validation in Panorama and Firewall itself. When we pull that information here and a certificate exists in the xml response, it will always have a name. We do not want a python dict error if it is None. If a certificate is present with a None name for some reason, the code should still continue to add it, in my opinion.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
aneeshamore: removed.

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aneeshamore: Moved to outer scope

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aneeshamore: renamed to consolidated_cert_details

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aneeshamore: Moved to outer scope

Packs/PAN-OS/Integrations/Panorama/Panorama.py
aneeshamore: Created a new function for processing certificates for each scenario (4 new functions added).

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14604,261 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+            "cert_type": cert_type,
+        }
+        if devices_using_certificate:
+            cert_details_dict.update({"devices_using_certificate": devices_using_certificate})
+        cert_details.append(cert_details_dict)
+
+    return cert_details
+
+
+def extract_certificates_from_running_config(device) -> Tuple[List, List]:
```
rshunim:
```suggestion
def extract_certificates_from_running_config(device: <add here the object's type>) -> Tuple[List, List]:
```

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14604,261 @@ def pan_os_get_master_key_details_command() -> CommandResults:
     )
 
 
+def expiration_status_check(cert_expiration: datetime) -> str:
```
rshunim, line 14604: please add a documentation for this function

Packs/PAN-OS/Integrations/Panorama/Panorama.py
```diff
@@ -14595,6 +14604,261 @@ def pan_os_get_master_key_details_command() -> CommandResults:
...
+        return "Expiring in 30 days"
+    elif cert_expiration < now + timedelta(days=60):
+        return "Expiring in 60 days"
+    elif cert_expiration < now + timedelta(days=90):
+        return "Expiring in 90 days"
+    else:
+        return "Valid"
+
+
+def compile_certificate_details(
```
rshunim, line 14632: please add a documentation for this function (too)

rshunim:
Hi @aneeshamore 
It's looking good, You just need to fix three little comments and we're on the right track.
Make sure the build is green 🟢 

@jlevypaloalto @richardbluestone please review it too, thanks

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -9828,6 +9828,37 @@ script:
...
+      description: Expiration date of the certificate.
+    - contextPath: Panorama.Certificate.expiration_status
+      type: String
+      description: Status of certificate - Expired, Expiring in 30 days, Expiring in 60 days, Expiring in 90 days or Valid.
+    - contextPath: Panorama.Certificate.location
+      type: String
+      description: Where this certificate was configured - Firewall or Panorama.
+    - contextPath: Panorama.Certificate.cert_type
+      type: String
+      description: If this certificate was Pushed from Panorama, Local to Firewall or Predefined on Panorama or Firewall.
```
richardbluestone: Whether this certificate was pushed from Panorama, Local to Firewall, or predefined on Panorama or a Firewall.

Packs/PAN-OS/Integrations/Panorama/README.md
```diff
@@ -9833,3 +9833,67 @@ There are no input arguments for this command.
...
+
+| **Path** | **Type** | **Description** |
+| --- | --- | --- |
+| Panorama.Certificate.name | String | Name of the certificate. |
+| Panorama.Certificate.device | String | Device where this certificate was found. |
+| Panorama.Certificate.subject | String | Subject of the certificate. |
+| Panorama.Certificate.expiration_date | String | Expiration date of the certificate. |
+| Panorama.Certificate.expiration_status | String | Status of certificate - Expired, Expiring in 30 days, Expiring in 60 days, Expiring in 90 days or Valid. |
+| Panorama.Certificate.location | String | Where this certificate was configured - Firewall or Panorama. |
+| Panorama.Certificate.cert_type | String | If this certificate was Pushed from Panorama, Local to Firewall or Predefined on Panorama or Firewall. |
```
richardbluestone: Whether this certificate was pushed from Panorama, Local to Firewall, or Predefined on Panorama or a Firewall.

Packs/PAN-OS/Integrations/Panorama/Panorama.yml
```diff
@@ -9828,6 +9828,37 @@ script:
       type: Number
       required: true
       isArray: false
+  - name: pan-os-get-certificate-info
+    description: 'Gathers name, expiration date and expiration status of certificates configured locally on a Firewall or pushed from Panorama, as seen under Certificate Management.'
```
richardbluestone: Gathers the name, expiration date, and expiration status of certificates configured locally on a Firewall or pushed from Panorama, as seen under Certificate Management.'

Packs/PAN-OS/Integrations/Panorama/README.md
```diff
@@ -9833,3 +9833,67 @@ There are no input arguments for this command.
 >| Auto-renew master key | Encrypted on HSM | Remind at | Expire at |
 >| --- | --- | --- | --- |
 >| 0 | no | 2024/11/27 04:26:05 | 2025/02/18 04:26:05 |
+
+### pan-os-get-certificate-info
+
+***
+Gathers name, expiration date and expiration status of certificates configured locally on a Firewall or pushed from Panorama, as seen under Certificate Management.
```
richardbluestone: Gathers the name, expiration date, and expiration status of certificates configured locally on a Firewall or pushed from Panorama, as seen under Certificate Management.

Packs/PAN-OS/ReleaseNotes/2_5_1.md
```diff
@@ -0,0 +1,6 @@
+#### Integrations
+
+##### Palo Alto Networks PAN-OS
+
+- Added a new command to get the certificate information from Panorama or firewall as seen under Certificate Management:
```
richardbluestone: Added a new command to get the certificate information from Panorama or a firewall, as seen under Certificate Management:

PR #40329, opened 2025-06-22
External login password spray DT fix

idovandijk:
## Status
Ready

## Description
DT fix in saving post login activity to evidence - could result in wrong verdict calculation


karinafishman:
Looks fine

PR #40333, opened 2025-06-23
Confense Triage Release 2.1.36

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

#### Integration - Cofense Triage v3

- Updated the logic to fetch missing reports. 

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/CofenseTriage/Integrations/CofenseTriagev3/CofenseTriagev3.py
+++ b/Packs/CofenseTriage/Integrations/CofenseTriagev3/CofenseTriagev3.py
@@ -1623,13 +1623,9 @@ def fetch_incidents(client: Client, last_run: dict, params: dict) -> tuple[dict,
 
     if results:
         next_run.pop("id", None)
+        next_run.pop("page_number", None)
         next_run["ids"] = ids
-        if len(results) < int(params.get("max_fetch")) or not next_run.get("updated_at"):  # type:ignore
-            next_run["updated_at"] = results[-1].get("attributes", {}).get("updated_at")
-            next_run["page_number"] = 1
-        else:
-            next_run["page_number"] = last_run.get("page_number", 1) + 1
-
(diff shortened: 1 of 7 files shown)


Packs/CofenseTriage/Integrations/CofenseTriagev3/CofenseTriagev3.py
```diff
@@ -1623,13 +1623,9 @@ def fetch_incidents(client: Client, last_run: dict, params: dict) -> tuple[dict,
...
         next_run.pop("id", None)
+        next_run.pop("page_number", None)
         next_run["ids"] = ids
-        if len(results) < int(params.get("max_fetch")) or not next_run.get("updated_at"):  # type:ignore
-            next_run["updated_at"] = results[-1].get("attributes", {}).get("updated_at")
-            next_run["page_number"] = 1
-        else:
-            next_run["page_number"] = last_run.get("page_number", 1) + 1
-
+        next_run["updated_at"] = results[-1].get("attributes", {}).get("updated_at")
```
ilappe, line 1628: Correct me if i'm wrong
what will happen here if there is a `max_fetch` events which happened at the same time (the same `updated_at`)? (it may be that this case are invalid in your product)
but if so, in the old implementation, it will stop at the end page, but in the new implementation it will fetch again the same events

crestdatasystems: The likelihood of this scenario occurring is very low due to the following reasons:

1. The `updated_at` parameter includes millisecond-level precision, making it highly unlikely that multiple reports will share the exact same timestamp.
2. Reports are often updated during processing—whether in XSOAR or through other systems. As a result, even if a rare case arises where reports share the same `updated_at` timestamp, the updated reports will be picked up in subsequent fetch cycles once their timestamp changes.

Therefore, this behavior is self-correcting and should not impact the reliability of report fetching.

PR #40370, opened 2025-06-25
QRadar_v3: Added sleep before initial check if query is complete.

mcvic1rj:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Adds in an additional sleep, plus debug log line to address an issue where the integration checks for a search being completed ~5-10 milliseconds after starting it. 
In testing, this reduced the amount of time it took to ingest 36 Offenses from ~5 minutes, down to 1 minute. 

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/QRadar/ReleaseNotes/2_5_21.md
+++ b/Packs/QRadar/ReleaseNotes/2_5_21.md
@@ -3,4 +3,4 @@
 
 ##### IBM QRadar v3
 
-- Updated to add a sleep to address a problem where integration would wait >60 seconds for a search that was completed.
+- Fixed an issue where the integration would wait for over 60 seconds after a search had already completed.
--- a/Packs/QRadar/ReleaseNotes/2_5_21.md
+++ b/Packs/QRadar/ReleaseNotes/2_5_21.md
@@ -3,5 +3,4 @@
 
 ##### IBM QRadar v3
 
-- Updated the Docker image to: *demisto/python3:3.12.8.3296088*.
 - Updated to add a sleep to address a problem where integration would wait >60 seconds for a search that was completed.
--- a/Packs/QRadar/ReleaseNotes/2_5_21.md
+++ b/Packs/QRadar/ReleaseNotes/2_5_21.md
@@ -1,6 +1,7 @@
 
 #### Integrations
 
-##### QRadar v3
-Updated to add a sleep to address a problem where integration would wait >60 seconds for a search that was completed.
+##### IBM QRadar v3
 
+- Updated the Docker image to: *demisto/python3:3.12.8.3296088*.
+- Updated to add a sleep to address a problem where integration would wait >60 seconds for a search that was completed.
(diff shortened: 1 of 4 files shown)


itssapir:
Hi @mcvic1rj,

You are raising a solid point. I tried to replicate the issue but unfortunately the searches didn't finish quick enough on our available instances for this to make a difference as there were always multiple retries regardless.

I have a couple of questions/comments:
- Did you try to use smaller sleep values? If the performance gain remains the same with a lower sleep time, it could be less disruptive to the current flow. If not, can you try lowering it and seeing if it still solves the problem?

- I personally feel the initial sleep is best added outside of the poll_offense_events_with_retry function, since we are waiting before starting the polling process. I would place the sleep here:
>         else:
>            time.sleep(FETCH_INITIAL_SLEEP)
>            events, failure_message = poll_offense_events_with_retry(client, search_id, int(offense_id))

mcvic1rj:
> * Did you try to use smaller sleep values? If the performance gain remains the same with a lower sleep time, it could be less disruptive to the current flow. If not, can you try lowering it and seeing if it still solves the problem?

Our initial tests showed between 400ms and 750ms for the initial searches to finish. We could lower the sleep, at the cost of 60 seconds for a miss. 

Worse case scenario: 
- Search is done in <10 milliseconds (Time between line [#2355](https://github.com/demisto/content/blob/c58349ec859b58fa8d1fccec3fab1b785691b1e7/Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py#L2355) and line [#2359](https://github.com/demisto/content/blob/c58349ec859b58fa8d1fccec3fab1b785691b1e7/Packs/QRadar/Integrations/QRadar_v3/QRadar_v3.py#L2359)) 
- adding ~1 second or ~10000%. This should be a very slim-to-none case.

When looking at other scenarios, we would be adding 1 second to ~60seconds, upto retries*60seconds, so between 1.7% and 0.17% to the execution time.

I see where you are suggesting the sleep be moved and I agree that it appears to be a better spot and will update the PR to move that.

mcvic1rj:
@itssapir PR has been updated with changes from your feedback.

Packs/QRadar/ReleaseNotes/2_5_21.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### IBM QRadar v3
+
+- Updated to add a sleep to address a problem where integration would wait >60 seconds for a search that was completed.
```
richardbluestone: Fixed an issue where the integration would wait for over 60 seconds after a search had already completed.

richardbluestone:
Doc review done

Packs/QRadar/ReleaseNotes/2_5_21.md
itssapir:
```suggestion
- Fixed an issue where the integration would wait for over 60 seconds after a search had already completed.
```

itssapir (approved):

PR #40409, opened 2025-06-30
Silent - Suspicious local user account creation

idovandijk:
# Status
Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-13622

## Description
New silent playbook - `Suspicious local user account creation`.

Suspicious_Local_User_Account_Creation_Mon_Jun_30_2025

### DIFF
--- a/Packs/CortexResponseAndRemediation/ReleaseNotes/1_1_90.md
+++ b/Packs/CortexResponseAndRemediation/ReleaseNotes/1_1_90.md
@@ -1,0 +1,1 @@
+## Documentation and metadata improvements.
--- a/Packs/CortexResponseAndRemediation/pack_metadata.json
+++ b/Packs/CortexResponseAndRemediation/pack_metadata.json
@@ -2,7 +2,7 @@
     "name": "Cortex Response And Remediation",
     "description": "The Cortex Response & Remediation Pack delivers a powerful collection of automated playbooks designed to streamline incident response and remediation processes. Built to support an Autonomous SOC vision.",
     "support": "xsoar",
-    "currentVersion": "1.1.89",
+    "currentVersion": "1.1.90",
     "author": "Cortex XSOAR",
     "url": "https://www.paloaltonetworks.com/cortex",
     "email": "",


Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "6":
+    id: "6"
+    taskid: 2fe4d3ae-4ad1-44e0-928c-997af07f4343
+    type: regular
+    task:
+      id: 2fe4d3ae-4ad1-44e0-928c-997af07f4343
+      version: -1
+      name: Search related alerts
```
ArikDay: The task name should describe the purpose of the investigation — for example, 'Search for signs of masquerading attempts,' or, if we want something more general, 'Search for additional suspicious behavior by the user.'

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "29":
+    id: "29"
+    taskid: 1ca5ab03-8c84-485b-801c-74f3d2d7b38e
+    type: regular
+    task:
+      id: 1ca5ab03-8c84-485b-801c-74f3d2d7b38e
+      version: -1
+      name: Save related alerts to evidence
```
ArikDay, line 771: Same - The task name should describe the purpose of the investigation.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "11":
+    id: "11"
+    taskid: ceabad22-9b7a-46d4-a02a-ad731ca6a548
+    type: regular
+    task:
+      id: ceabad22-9b7a-46d4-a02a-ad731ca6a548
+      version: -1
+      name: Retrieve local user password status
```
ArikDay, line 264: Either this task or the "set" task after should mention that we are checking if the user password will never expires

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+      closeReason:
+        complex:
+          root: CloseReason
+          transformers:
+          - operator: SetIfEmpty
+            args:
+              applyIfEmpty: {}
+              defaultValue:
+                value:
+                  simple: Resolved - False Positive
```
ArikDay: After a remediation, the alert will be closed and classified as "Resolved - False Positive"

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+    continueonerrortype: ""
+    view: |-
+      {
+        "position": {
+          "x": 786,
+          "y": 2804
+        }
+      }
+    note: false
+    timertriggers: []
```
ArikDay, line 2324: End the playbook with "Done" header

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+    isoversize: false
+    isautoswitchedtoquietmode: false
+  "44":
+    id: "44"
+    taskid: ce712cfe-c409-46bf-9d4a-9e1a54c57b96
+    type: condition
+    task:
+      id: ce712cfe-c409-46bf-9d4a-9e1a54c57b96
+      version: -1
+      name: Analyst review - disable suspicious user?
```
ArikDay, line 1255: There is a logic problem in this flow:
Playbook reaches to manual task -> Choosing yes to disable it -> the user type is domain but the integration does not exit. In this case, the user is indeed a domain user but will be blocked only locally. We should address that. I think that we need to check if the user is a domain user prior to the manual task, and suggest solutions accordingly.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+  "70":
+    id: "70"
+    taskid: 18e9f289-19ea-4b1b-b1e3-034b89c56d0f
+    type: regular
+    task:
+      id: 18e9f289-19ea-4b1b-b1e3-034b89c56d0f
+      version: -1
+      name: Undetermined user type - manual
+      description: |-
+        Active Directory was unavailable, and user details for "${Core.OriginalAlert.event.action_evtlog_data_fields.SubjectUserName}" could not be retrieved.
```
ArikDay, line 2185: Based on the conditional task, the user could not be retrieved because the integration does not exist. We should mention that.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+
+        The target (created) user: `${Core.OriginalAlert.event.evtlog_target_username}`
+
+        Involved Host: `${alert.hostname}`
+
+
+        ---
+
+        ### Related Alerts
+        Suspicious related alerts found: `${.=val.foundIncidents && val.foundIncidents.length > 0 ? "True" : "False"}`
```
ArikDay: We should provide the related alerts found. There is not much value for the analyst when we say "Related alerts: True". He should know that the same user has potentially performed "Suspicious disablement of the Windows Firewall" for example.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
```diff
@@ -0,0 +1,2395 @@
...
+
+        ### Host & User Risk Levels
+        Host risk level: `${Core.RiskyHost.risk_level}`
+
+        Creator user risk level: `${Core.RiskyHost.risk_level}`
+
+        ---
+
+        ### Suspicious command line executions
+        Created user: `${.=val.SuspiciousCommandlinesList || "N/A"}`
```
ArikDay, line 1295: We should provide the reasons that the command lines are suspicious. Should be under CommandLineAnalysis.findings

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: Due to text length limitation, changing to "Search suspicious user activity alerts"

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: Used for verdict, no further explanation needed for analyst IMO

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: fixing

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: it's using SetIfEmpty so it's OK, but since we don't want to close alerts I will modify to not close on FP/Inconclusive

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: done

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: I don't see a logical flaw. It will not disable the user locally if the integration is not available, because at that point it had already gone to the "Domain" path, so what will happen is what we expect:
If the user is a domain user, but the integration is unavailable - this means we cannot verify the source of the user. Therefore, remediation will not happen, and the alert will not be closed - these tasks will all become skipped.

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: improving

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: Since the new key used in the playbook (SuspiciousCommandlinesList) contains just the CMDs and not the reasons, changing this would require additional changes that I want to avoid at this stage as it was already tested in its current state

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: This is displayed in the message of the manual task where it's explained

Packs/CortexResponseAndRemediation/Playbooks/silent-playbook-Suspicious_Local_User_Account_Creation.yml
idovandijk: added under

ArikDay (approved):

PR #40431, opened 2025-07-01
moving clear-user-session,get-endpoint-data,block-external-ip to Aggr…

thefrieddan1:
…egatedScripts pack.

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-14014)

## Description
Moving aggregated commands from commonScripts to AggregatedScript


thefrieddan1:
Asked for force merge because Aggregated commands should be  separated from commonScripts pack

PR #40451, opened 2025-07-02
add agentix to corepacks

israelpoli:
Add suuported module `agentix` to corepacks


israelpoli:
Forced due to urgency in triggering a delayed upload
All validations and prepare stages passed successfully
The changes in the PR are solely in the pack metadata

PR #40462, opened 2025-07-02
move GR109 to warning

israelpoli:
move GR109 to warning


israelpoli:
force due to validation refactoring

PR #40478, opened 2025-07-03
bump version to all corepacks

israelpoli:
bump version to all corepacks


israelpoli:
Forced due to the RM108 validation, this is a validation that allows force merging
Aside from that, all prepare stages passed successfully, and the merge is highly urgent for the upcoming upload

PR #40554, opened 2025-07-08
improve implementation of RegexGroups

thefrieddan1:
## Related Issues

Will reduce "noise" in XSUP-52824

## Description
Fix the case where RegexGroups is run with groups args and no matches are found it throws an exception.

### DIFF
--- a/Packs/FiltersAndTransformers/ReleaseNotes/1_3_0.md
+++ b/Packs/FiltersAndTransformers/ReleaseNotes/1_3_0.md
@@ -1,0 +1,6 @@
+
+#### Scripts
+
+##### RegexGroups
+
+- Improved implementation of RegexGroups script.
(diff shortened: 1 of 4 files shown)


Packs/FiltersAndTransformers/ReleaseNotes/1_3_0.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+
+##### RegexGroups
+
+- Improved implementation of RegexGroups script.
```
JasBeilin:
```suggestion
- Fixed an issue where the regex has no matches.
```

Packs/FiltersAndTransformers/Scripts/RegexGroups/RegexGroups_test.py
```diff
@@ -48,3 +49,14 @@ def test_main__match(mocker):
     main()
     assert results_mock.call_count == 1
     assert results_mock.call_args[0][0] == ["test"]
+
+
+def test_get_regex_matches(mocker):
```
JasBeilin, line 54: Maybe we should consider adding some test cases.

thefrieddan1: Single case here as far as I understand which is  preventing the method throwing an exception in this edge case.

PR #40576, opened 2025-07-10
Set "value" argument to array type in DT script

kamalq97:
## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-53111)

## Description
Fixed an issue where performing operations on array values would return an incorrect result in the DT script.

### DIFF
--- a/Packs/FiltersAndTransformers/Scripts/DT/README.md
+++ b/Packs/FiltersAndTransformers/Scripts/DT/README.md
@@ -1,4 +1,4 @@
-Allows the usage of DT scripts within playbooks transformers.
+This automation allows the usage of DT scripts within playbooks transformers.
 
 ## Script Data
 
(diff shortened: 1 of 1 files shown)


Packs/FiltersAndTransformers/Scripts/DT/DT.yml
```diff
@@ -8,17 +8,18 @@ subtype: python3
 tags:
 - transformer
 - general
-comment: This automation allows the usage of DT scripts within playbooks transformers
+comment: This automation allows the usage of DT scripts within playbooks transformers.
 enabled: true
 args:
 - name: value
   required: true
   description: Value to perform the transformation on.
```
richardbluestone: The value to perform the transformation on.

richardbluestone:
Doc review done

Packs/FiltersAndTransformers/Scripts/DT/DT.yml
kamalq97: Fixed, thanks

PR #40610, opened 2025-07-15
Update parse-emails docker image

barryyosi-panw:
## Related Issues
fixes: [XSUP-53018](https://jira-dc.paloaltonetworks.com/browse/XSUP-53018), [XSUP-53425](https://jira-dc.paloaltonetworks.com/browse/XSUP-53425)

## Description
Update `ParseEmailFilesV2` docker image version with a fixed one that's able to cope with the `MIME entity text, ISO-8859 text, with CRLF line terminators` type & fallback when file decoding fails.


MosheEichler (approved):
Nice

PR #40615, opened 2025-07-16
Improve documentation of execute_polling_command function

kamalq97:
## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-13967)

## Description
Improve documentation of the `execute_polling_command` function in CommonServerPython to clarify that it should not be used under normal circumstances.

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -12717,7 +12717,8 @@ def execute_polling_command(
     default_polling_timeout=600,
 ):
     r"""
-    Continuously executes a specified command until the command indicates polling is done or a
+    **Only for use in special circumstances where polling is not supported. Avoid using otherwise!**
+    Continuously executes a specified command and sleeps until the command indicates polling is done or a
     timeout is reached.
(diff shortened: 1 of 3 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -12717,7 +12717,8 @@ def execute_polling_command(
     default_polling_timeout=600,
 ):
     r"""
-    Continuously executes a specified command until the command indicates polling is done or a
+    **Only for use in special circumstances where polling is not supported. Avoid using otherwise!**
```
jbabazadeh: ###########################################
#  DO NOT USE THIS UNLESS ABOLUTLY NECESSERY!  #
###########################################
This is intended only for special cases where polling is not supported.

jbabazadeh (approved):

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
kamalq97: Updated function docstring

kobymeir (approved):

PR #40651, opened 2025-07-21
Rubrik Release 1.5.2

crestdatasystems:

## Status
- [ ] In Progress
- [x] Ready

## Description
- Updated the test connection for integration to work regardless of RSC license.

## Must have
- [x] Tests
- [x] Documentation


crestdatasystems:
> Hey, The PR looks great, One thing that needs to be changed is the docker image in the yml file witch is out of date (see pre-commit failure), you also need to update the release notes for this change. Please send a short demo video so we see that the test button works. Also, I would like to get an explanation of why was this change needed? Thank you!

Hi @YaelShamai,
Thanks for reviewing the PR.

The reason of this release is to execute a GraphQL query which will work regardless of service account permissions.
- This will ensure that the end users won't need extra permissions just to pass the test connection.
- We have shared the demo recording over DFIR.

Regarding the Docker image update, we have tried to update it but we started receiving the errors related to the python packages.
image

- Here is the history of Dockerfile for Rubrik SDK: https://github.com/demisto/dockerfiles/commits/master/docker/rubrik-polaris-sdk-py3
- Can you investigate the issue from your side as well? As these updates were done by PAN XSOAR.

YaelShamai (approved):

PR #40745, opened 2025-07-28
SplunkPy: update code & doc

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-53385

## Description
Fixed an issue where the **splunk-submit-event-hec** command could fail with an 'index not found' error, even when a valid `index` was provided as an argument.


DeanArbel (approved):
Nicely done

PR #40751, opened 2025-07-28
 Added Advanced Hunting Process Ancestry Lookup

WildDogOne:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
This pull request is designed to add functionality to the MDE / ATP Integration for looking up the process ancestry by ProcessID and Hostname/HostId and FileName/Hash

Example Output:
image

**Remaining Questions:**
Ideas on how to structure the readable output for more user value.
Ideas on how to structure the JSON output for further use of the data in future commands.

I have added hashing of the inputs to make a "unique" process chain ID, which can be used for better searching the output.
However this of course adds a new import and thus a dependency, I don't know if that is OK or not.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
+++ b/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
@@ -4969,7 +4969,7 @@ def iterate_ancestry(client,     timeout = None,
     return response.get("Results"), query
 
 
-def process_ancestry_command(client, args):  # pragma: no cover
+def process_ancestry_command(client, args):
     timeout = arg_to_number(args.get("timeout", 10))
     time_range = args.pop("time_range", None)
     device_name = args.get("device_name")
--- a/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection_test.py
+++ b/Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection_test.py
@@ -30,6 +30,7 @@ from MicrosoftDefenderAdvancedThreatProtection import (
     put_live_response_file_action,
     run_live_response_script_action,
     run_polling_command,
+    process_ancestry_command
 )
 
 ARGS = {"id": "123", "limit": "2", "offset": "0"}
@@ -383,6 +384,15 @@ def test_check_limit_and_offset_values_limit_zero(mocker):
 
 
 """ API RAW RESULTS """
+
+ADVANCED_HUNTING_ANCESTRY_RESPONSE = {
+    "InitiatingProcessFileName": "explorer.exe",
+    "InitiatingProcessId": "10996",
+    "InitiatingProcessCommandLine": "Explorer.EXE",
+    "InitiatingProcessParentFileName": "userinit.exe",
+    "InitiatingProcessParentId": "10352",
+    "InitiatingProcessCreationTime": "23 Jul 2025 15:33:16",
+}
 
 FILE_DATA_API_RESPONSE = {
     "sha1": "123abc",
@@ -3807,3 +3817,19 @@ def test_list_auth_permissions_command(mocker):
     command_results = list_auth_permissions_command(client_mocker)
 
     assert command_results.readable_output == "### Permissions\nEvent.Write\nUser.Read"
+
+
+def test_process_ancestry_command(mocker):
+    """
+    Tests conversion of alert response
+
+    Given:
+        - alert response as json
+    When:
+        - calling for machine alerts
+    Then:
+        - return alert data dict
+    """
+    mocker.patch.object(client_mocker, "get_advanced_hunting", return_value=ADVANCED_HUNTING_ANCESTRY_RESPONSE)
+    results = process_ancestry_command(client_mocker, {"machine_id": "123abc"})
+    assert results.outputs[0] == MACHINE_ALERTS_OUTPUT


WildDogOne:
I have "improved" the outputs.
There is now a simple human readable form and an extended.
I don't think including both make sense. But I could add a user input argument to select which output form to use?
Also does anyone think I am missing some fields in output? There is a lot of data in MDE, I could push everything into the output, but quickly gets messy.

image

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    if not(md5 or sha1 or sha256 or file_name):
+        raise DemistoException(f"At least one hashtype or filename/process name must be specified.")
+    if not(device_name or device_id):
+        raise DemistoException(f"At least device name or device id must be specified.")
+    query = "DeviceProcessEvents"
+    if device_name:
+        query += f'| where DeviceName == "{device_name}"'
+    if device_id:
+        query += f'| where DeviceId == "{device_id}"'
+    query += f'| where ProcessId == "{file_pid}"'
```
YaelShamai, line 4950: This isn't in the 'if' case, this needs to be added anyways?

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+        query += f'| where MD5 == "{md5}"'
+    if file_name:
+        query += f'| where FileName == "{file_name}"'
+    if process_creation_time:
+        query += f'| extend timediff = datetime_diff("second", datetime({process_creation_time}), Timestamp)'
+        query += '| order by timediff asc'
+    query += '| limit 1'
+
+    # send request + return results and query used
+    response = client.get_advanced_hunting(query, timeout, time_range)
```
YaelShamai: The api call should be in a try except.

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    return response.get("Results"), query
+
+
+def process_ancestry_command(client, args):  # pragma: no cover
```
YaelShamai: Remove pragma no cover, unit tests need to reach the needed coverage.

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    timeout = int(args.pop("timeout", 10))
```
YaelShamai: Use arg_to_number(args.get("timeout", 10))

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    time_range = args.pop("time_range", None)
```
YaelShamai, line 4974: Is time range a number, change this same as timeout arg.

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    device_name = args.get("device_name")
+    file_name = args.get("file_name")
+    file_pid = args.get("file_pid")
+    sha1 = args.get("sha1")
+    sha256 = args.get("sha256")
+    md5 = args.get("md5")
+    device_id = args.get("device_id")
+    process_creation_time = args.get("process_creation_time")
```
YaelShamai: use arg_to_datetime

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
```diff
@@ -4924,6 +4926,158 @@ def process_details_command(client, args):  # pragma: no cover
...
+    show_query = argToBoolean(args.pop("show_query", False))
+    process_chain = None
+    process_json = []
+    fields_to_hash = f"{device_name}{file_name}{file_pid}{sha1}{sha256}{md5}{device_id}{process_creation_time}"
+    process_chain_id = hashlib.md5(fields_to_hash.encode()).hexdigest()
+    process_depth = 0
+    queries = "### Queries Used\n"
+    while True:
```
YaelShamai, line 4990: Explain why you need a while True loop, this needs to be handled carefully and it is better to avoid, lets think of a different way to implement this.
Also, is there a way to avoid so many api calls, maybe the limit shouldn't be 1?

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
```diff
@@ -5283,6 +5283,61 @@ script:
       description: The query results for process_excecution_powershell query_purpose.
       type: String
     polling: true
+  - arguments:
+    - description: Device name to look for. Device Name or Device ID must be entered.
+      name: device_name
+      defaultValue: ''
+    - description: Device ID to look for.Device Name or Device ID must be entered.
```
YaelShamai: - description: Device ID to look for. Device Name or Device ID must be entered.

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
```diff
@@ -5283,6 +5283,61 @@ script:
...
+    - contextPath: MicrosoftATP.HuntProcessDetails.Result.beaconing_evidence
+      description: The query results for beaconing_evidence query_purpose.
+      type: String
+    - contextPath: MicrosoftATP.HuntProcessDetails.Result.powershell_execution_unsigned_files
+      description: The query results for powershell_execution_unsigned_files query_purpose.
+      type: String
+    - contextPath: MicrosoftATP.HuntProcessDetails.Result.process_excecution_powershell
+      description: The query results for process_excecution_powershell query_purpose.
+      type: String
+    polling: true
```
YaelShamai: Why polling: true? I didn't see you implemented a polling mechanism, you implemented a pagination one, if this is confusing we can discuss thoroughly.

Packs/MicrosoftDefenderAdvancedThreatProtection/ReleaseNotes/1_20_4.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Microsoft Defender for Endpoint
+
+Added process ancestry hunting
```
YaelShamai: Added **microsoft-atp-advanced-hunting-process-ancestry** command.

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: Indeed, for this query to work well the PID is essential.
I have flagged it as mandatory in the yml
    - description: Process ID of the process ancestry entrypoint.
      required: true
      name: file_pid

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: No idea, I copied an existing command and that was just there :)
I can check the unittests for sure

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: fyi this is also then wrong in other commands in this integration, might want to change them.
I usually try to keep my code in line with the existing code

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: the existing integration uses that as some fancy word thing, for example, time range: 1 day
I have zero clue how it actually works, but it seems to work

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: I believe that could be dangerous, does arg_to_datetime change the datetime format?

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: Honestly, I don't have the brainpower to change the loop

basically what it does is:
Check the child of a given PID and Filename
then it has to rebuild another query with the new PID and Filename of that child as the now starting point of the query.

So if there are for example 4 parents, the query will have to run 4 times, no way around that imho

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
WildDogOne: Done, will be changed in next commit

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
WildDogOne: Artefact of copying another command, I don't know if it is needed or not.
I'll kick it out and test

Packs/MicrosoftDefenderAdvancedThreatProtection/ReleaseNotes/1_20_4.md
WildDogOne: Done, will be added in next commit

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: well it seems to not have broken anything, so I will keep it

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.yml
WildDogOne: made no difference when I removed it, so it's gone

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: done

Packs/MicrosoftDefenderAdvancedThreatProtection/Integrations/MicrosoftDefenderAdvancedThreatProtection/MicrosoftDefenderAdvancedThreatProtection.py
WildDogOne: Aright while adding unittests I have noticed that I get this error on all existing tests:
``fixture 'mocker' not found``

no idea where that fixture is declared

WildDogOne:
I will open a support request so that Palo Alto Support can take over on this issue

PR #40773, opened 2025-07-30
New Filters And Transformers Scripts: RemoveNullBytes

mandarnaik016:
## Status
- [x] Ready

## Description
Incidents data sometimes contains base64 encoded strings, when decoded using Base64Decode transformer leaves null bytes intact. This transformer script will remove null bytes.

## Must have
- [x] Tests
- [x] Documentation


mandarnaik016:
Hello,

The use case was when powershell encoded or obfuscated base64 scripts were executed/detected and incident was received for the same, the current transformer were unable to provide expected output, by removing null bytes present.

Adding the issue before and fix snapshot [here](https://drive.google.com/drive/folders/1VVSX8nHyS9Ir2nPdILKSjpn4MUYN8YW9)

merit-maita (approved):
the script looks great, thank you for your contribution

PR #41029, opened 2025-08-21
SplunkPy mirror-in: remove duplications

ilappe:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-54482

## Description
This change addresses a mirroring issue where Splunk's indexing delays could cause duplicate events to be processed.

The solution implements a caching mechanism to track events that have already been processed:

A 60-second look-behind is used in the Splunk query to ensure late-indexed events are not missed.
To prevent duplicates from this overlapping query window, a unique key (notable_id:timestamp) for each processed event is stored in the integration context.
In subsequent runs, the integration checks this cache and skips any event that has already been processed, ensuring each notable is handled only once.

### DIFF
--- a/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
+++ b/Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
@@ -1842,11 +1842,20 @@ def get_modified_remote_data_command(
     Returns:
         SplunkGetModifiedRemoteDataResponse: The response containing the list of notables changed
     """
-    modified_notables_map = {}
-    entries: list[dict] = []
     remote_args = GetModifiedRemoteDataArgs(args)
-    demisto.debug(f"Original last_update before 1 minute reduction: {get_last_update_in_splunk_time(remote_args.last_update)}")
-    # Subtract 1 minute (60 seconds) to allow indexing
+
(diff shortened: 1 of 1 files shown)


Packs/SplunkPy/Integrations/SplunkPy/SplunkPy.py
```diff
@@ -1880,6 +1908,10 @@ def get_modified_remote_data_command(
                 }
             )
 
+    # Persist the cache of events processed in this run for the next iteration.
+    integration_context["processed_mirror_in_events_cache"] = list(current_run_processed_events)
+    set_integration_context(integration_context)
```
Shellyber, line 1913: @ilappe What about the cache size? 
I think we should add a mechanism to clean it also after a while

ilappe: it will be deleted in the next iteration

Shellyber (approved):

ilappe:
Force merge reason:
Test playbook is [failing](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/28808019#L1734) on a command unrelated to the fix.
It seems to be an issue with the Splunk server.
It was tested manually on another Splunk server and works perfectly.

PR #41152, opened 2025-09-03
Add code/xdr list users to GetUserData

YuvHayun:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [CRTX-194707](https://jira-dc.paloaltonetworks.com/browse/CRTX-194707)

## Description

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
+++ b/Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
@@ -707,8 +707,7 @@ def get_core_and_xdr_data(
     brand_name: str,
     first_command: str,
     second_command: str,
-    first_arg_name: str,
-    first_arg_values: List[str],
+    user_names: List[str],
     additional_fields: bool,
     list_non_risky_users: bool = False,
     email_list: List[str] = [],
@@ -716,12 +715,12 @@ def get_core_and_xdr_data(
 ) -> tuple[list[CommandResults], list[dict[str, Any]]]:
     # prepare both commands
     first_commands = []
-    for user_name in first_arg_values:
+    for user_name in user_names:
         first_commands.append(
             Command(
                 brand=brand_name,
                 name=first_command,
-                args={first_arg_name: user_name, "using-brand": brand_name},
+                args={"user_id": user_name, "using-brand": brand_name},
             )
         )
         if auto_detect_indicator_type(user_name) == FeedIndicatorType.Email:
@@ -770,7 +769,6 @@ def main():
             raise ValueError("At least one of the following arguments must be specified: user_id, user_name or user_email.")
 
         command_results_list: list[CommandResults] = []
-        user_outputs_list: list[dict[str, Any]] = []
         users_outputs: list[dict] = []
         users_readables: list = []
 
@@ -1113,14 +1111,14 @@ def main():
             brand_name="Cortex XDR - IR",
             first_command="xdr-list-risky-users",
             second_command="xdr-list-users",
-            first_arg_name="user_id",
-            first_arg_values=users_names,
+            user_names=users_names,
             additional_fields=additional_fields,
             list_non_risky_users=list_non_risky_users,
             email_list=users_emails,
             outputs_key_field="PaloAltoNetworksXDR",
         )
-        if readable_output and outputs:
+        # we don't expect xdr-list-users to return readable outputs.
+        if readable_output or outputs:
             users_outputs.extend(outputs)
             users_readables.extend(readable_output)
 
@@ -1132,18 +1130,17 @@ def main():
             brand_name="Cortex Core - IR",
             first_command="core-list-risky-users",
             second_command="core-list-users",
-            first_arg_name="user_id",
-            first_arg_values=users_names,
+            user_names=users_names,
             additional_fields=additional_fields,
             list_non_risky_users=list_non_risky_users,
             email_list=users_emails,
             outputs_key_field="Core",
         )
+        # we don't expect core-list-users to return readable outputs.
         if readable_output or outputs:
             users_outputs.extend(outputs)
             users_readables.extend(readable_output)
 
-        demisto.debug(f"users list: {user_outputs_list}")
         command_results_list.append(
             CommandResults(
                 outputs_prefix="UserData",


Packs/AggregatedScripts/ReleaseNotes/1_1_10.md
```diff
@@ -0,0 +1,7 @@
+
+#### Scripts
+
+##### get-user-data
+
+- Added support for **list_non_risky_users** argument that decides whether to return only risky users from core/xdr brands or all given users. if set to true, the execution might take some time.
```
joeymizrahi, line 6: I was thinking about the argument name `list_non_risky_users` and realized it might be misinterpreted. Someone could read it as true listing only the non-risky users, instead of all of them.
What do you think?
Maybe the name should be something like: `list_all_users`, default will be false

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -604,6 +702,52 @@ def get_data(
     return [], []
 
 
+def get_core_and_xdr_data(
+    modules: Modules,
+    brand_name: str,
+    first_command: str,
+    second_command: str,
+    first_arg_name: str,
```
joeymizrahi: if this will always be user_id, why not hard code it?
more clear

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -604,6 +702,52 @@ def get_data(
...
+    first_arg_values: List[str],
```
joeymizrahi: if this will always be "user_names" why not call it user_names as opposed to first_arg_values

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -992,6 +1105,44 @@ def main():
...
+            first_command="core-list-risky-users",
+            second_command="core-list-users",
+            first_arg_name="user_id",
+            first_arg_values=users_names,
+            additional_fields=additional_fields,
+            list_non_risky_users=list_non_risky_users,
+            email_list=users_emails,
+            outputs_key_field="Core",
+        )
+        if readable_output or outputs:
```
joeymizrahi, line 1140: why is everywhere else "and" but the condition here is "or"?

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -992,6 +1105,44 @@ def main():
...
+            users_outputs.extend(outputs)
+            users_readables.extend(readable_output)
+
         demisto.debug(f"users list: {user_outputs_list}")
```
joeymizrahi: i think user_outputs_list is not even populated or used. 
should change debug to user_outputs maybe?

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -604,6 +702,52 @@ def get_data(
...
+    first_arg_name: str,
+    first_arg_values: List[str],
+    additional_fields: bool,
+    list_non_risky_users: bool = False,
+    email_list: List[str] = [],
+    outputs_key_field: str = "",
+) -> tuple[list[CommandResults], list[dict[str, Any]]]:
+    # prepare both commands
+    first_commands = []
+    for user_name in first_arg_values:
```
joeymizrahi: there is no support on these commands for sending list of users? :(

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -493,52 +493,150 @@ def gsuite_get_user(
...
+                outputs = output.get(output_key, [])
+            else:
+                outputs = []
+            demisto.debug(f"found {len(outputs)} users")
+            for output in outputs:  # type: ignore[assignment]
+                if (mail := output.get("user_email", "")) and mail in email_set:
+                    demisto.debug(f"found user with email: {mail}")
+                    email_set.remove(mail)
+                    found = False
+                    for user in users:
```
joeymizrahi, line 583: i think this can be optimized.
if you create a dict lookup by user email like: 
`users_by_email = {user['Email']: user for user in users if user.get('Email')}`

And then in the loop of `for output in outputs` do something like:

`
if (found_user := users_by_email.get(mail)):
            demisto.debug(f"found user with email in results from risky-users: {mail}")
            if additional_fields:
                output.pop("user_email")
                found_user.setdefault("AdditionalFields", {}).update(output)
        else:
            # ... logic to create a new user ...
`

So you wouldnt need to iterate the users list again

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -493,52 +493,150 @@ def gsuite_get_user(
...
-    command: Command,
+    Returns:
+        tuple[list[CommandResults], list[dict[str, Any]]]: the list of readable outputs and the list of users.
+    """
+    email_set = set(email_list)
+    risky_users_email_set = set()
+    for user in users:
+        if mail := user.get("Email"):
+            risky_users_email_set.add(mail)
+    if additional_fields:
```
joeymizrahi, line 566: so if user sets the arg: list_non_risky_users to true but additional_fields to False, then they wont get the non-risky-users list?
is that expected?

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
```diff
@@ -992,6 +1105,44 @@ def main():
...
+        if readable_output and outputs:
+            users_outputs.extend(outputs)
+            users_readables.extend(readable_output)
+
+        #################################
+        ### Running for Core ###
+        #################################
+        readable_output, outputs = get_core_and_xdr_data(  # type: ignore[assignment]
+            modules=modules,
+            brand_name="Cortex Core - IR",
```
joeymizrahi, line 1130: i see in the ticket: https://jira-dc.paloaltonetworks.com/browse/CRTX-194707
its mentioned "both commands will use same brand name" what are they referring to?

joeymizrahi (changes requested):
really nice code!

Packs/AggregatedScripts/ReleaseNotes/1_1_10.md
YuvHayun: I think it might confuse the users for not listing all by default.
The description explains the idea of the argument.

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: Because for list-users we don't return HR (discussed as a part of the design)

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: seems like remains from a previous enhancement..
removed it.

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: nope, already tested :/

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: Yes, because the results coming back from list-risky-users should be sufficient to populate all the main keys for the outputs.
If additional_fields is set to True, then there are other fields to add to the outputs.

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: list-risky-users + list-users

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
joeymizrahi: so i think there should be a comment that mentions that, and maybe a UT that ensures no one changes in future to AND

Packs/AggregatedScripts/Scripts/GetUserData/GetUserData.py
YuvHayun: The thing is, I've seen cases where the users come back without email address from the list risky executions, so we can't map by that..

PR #41216, opened 2025-09-08
[CrowdstrikeFalcon] Fix Detection duplicates

jlevypaloalto:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-55834)

## Description

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
+++ b/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
@@ -11697,6 +11697,10 @@ def remove_old_incidents_ids(found_incidents_ids, current_time, look_back):
     deletion_threshold_in_seconds = look_back_in_seconds * 2
 
     new_found_incidents_ids = {}
+
+    latest_incident = max(found_incidents_ids, key=lambda x: datetime.fromisoformat(found_incidents_ids.get(x)))
+    latest_incident_time = found_incidents_ids[latest_incident]
+
     for inc_id, addition_time in found_incidents_ids.items():
 
         if current_time - addition_time <= deletion_threshold_in_seconds:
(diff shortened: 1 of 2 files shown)


Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -11697,9 +11697,12 @@ def remove_old_incidents_ids(found_incidents_ids, current_time, look_back):
     deletion_threshold_in_seconds = look_back_in_seconds * 2
 
     new_found_incidents_ids = {}
+    latest_incident_time = max(found_incidents_ids.values() or [current_time])
+    demisto.debug('lb: latest_incident_time is {}'.format(latest_incident_time))
+
```
JudahSchwartz: explain this change?  current_time wont always be the biggest?

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -11697,9 +11697,12 @@ def remove_old_incidents_ids(found_incidents_ids, current_time, look_back):
     deletion_threshold_in_seconds = look_back_in_seconds * 2
 
     new_found_incidents_ids = {}
+    latest_incident_time = max(found_incidents_ids.values() or [current_time])
+    demisto.debug('lb: latest_incident_time is {}'.format(latest_incident_time))
+    
     for inc_id, addition_time in found_incidents_ids.items():
 
-        if current_time - addition_time <= deletion_threshold_in_seconds:
+        if current_time - addition_time <= deletion_threshold_in_seconds or addition_time == latest_incident_time:
```
JudahSchwartz: explain? add a comment explaining as well

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
jlevypaloalto: Not if there's an empty run.
That is the case we are fixing; when an empty run causes all the old IDs to be deleted, including the ID of the incident that gave us the "fetch_from" time we are still using. This means that we pull the incident again, but we don't have the ID stored in the LastRun, so we create a duplicate.
With this fix, we will always have that ID in the LastRun.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
jlevypaloalto: We need to keep all the latest ID as I explained in the last comment. I'll add a comment to the code as well.

AradCarmi (approved):

PR #41229, opened 2025-09-09
XSUP-55768 Executing the ResolveShortenedURL script results in Server Error (500)

BarGali:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-55768

## Description
Deprecate the service argument of ResolveShortenedURL

## Must have
- [ ] Tests
- [ ] Documentation


Packs/CommonScripts/Scripts/ResolveShortenedURL/ResolveShortenedURL.py
```diff
@@ -51,7 +51,6 @@ def to_context_dict(self) -> dict:
         data = {
             "OriginalURL": self.original_url,
             "ResolvedURL": self.resolved_url,
-            "ServiceName": self.service_name,
```
BEAdi: Even though it always has the same value, we will keep it - because removing it is also BC and it is not mandatory to remove it.

Packs/CommonScripts/Scripts/ResolveShortenedURL/ResolveShortenedURL_test.py
BEAdi, line 1: Add a test for the case you added of a link with no https at the beginning.

Packs/CommonScripts/ReleaseNotes/1_20_12.json
```diff
@@ -0,0 +1,4 @@
+{
+    "breakingChanges": true,
+    "breakingChangesNotes": "The service argument of the ResolveShortenedURL script is deprecated. This script no longer uses online services."
```
BEAdi: Add that only the "builtin" option is supported now.
Copy the same notes from here to the RN, and ask for docs-review.

BEAdi (commented):
Looks good! Great work on the research here

Packs/CommonScripts/ReleaseNotes/1_20_13.md
```diff
@@ -0,0 +1,8 @@
+
+#### Scripts
+
+##### ResolveShortenedURL
+- Updated the Docker image to: *demisto/python3:3.12.11.4819260*.
+
+
+- Deprecated the **service** argument. This script no longer uses online services, Only the 'Built-In' option is supported now."
```
BEAdi:
```suggestion

- Deprecated the **service** argument from the ***ResolveShortenedURL*** script. This script no longer uses online services, only the 'Built-In' option is supported."
```

talihaff:
@BarGali Doc review complete

BarGali:
Pre-commit is currently failing due to the deletion of some JSON files. These files were previously used in the services tests, which have now been removed as part of this PR. Since the files are no longer used anywhere in the codebase, their deletion is intentional and should not pose any issues.

PR #41293, opened 2025-09-14
Aggregated Command Module

almog2296:
Base module for Aggregated Command and Enrichment Commands

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes:[ link to the issue](https://jira-dc.paloaltonetworks.com/browse/CRTX-187731)

Base aggregated command and specially for enrichment commands for future usage for ip/url/cve/domain enrichments commands


Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.yml
```diff
@@ -0,0 +1,15 @@
...
+name: AggregatedCommandApiModule
+script: ''
+type: python
+subtype: python3
+comment: Common Aggregated Command API Module, provides generic Infrastructure for different kinds of aggregated commands.
+scripttarget: 0
+dependson: {}
+timeout: 0s
+dockerimage: demisto/python3:3.12.11.4819260
+fromversion: 5.0.0
```
Shellyber:
```suggestion
fromversion: 6.10.0
```

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+# --- Core Enumerations and Data Classes ---
+class Status(Enum):
+    """Enum for command status."""
+
+    SUCCESS = "Success"
+    FAILURE = "Failure"
+
+
+class IndicatorStatus(Enum):
+    """Enum for indicator status."""
```
Shellyber: Maybe let's explain how each one is being decided? 
Stale - more than seven days etc.

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+import demistomock as demisto
+
+# Type alias for complex dictionary structures to improve readability
+ContextResult = dict[str, Any]
+DBotScoreList = list[ContextResult]
+
+# Calculating time interval for indicators freshness
+STATUS_FRESHNESS_WINDOW = timedelta(weeks=1)
+
+DBOT_SCORE_TO_VERDICT = {
```
JudahSchwartz, line 21: isnt this in csp?

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+    """Enum for command status."""
+
+    SUCCESS = "Success"
+    FAILURE = "Failure"
+
+
+class IndicatorStatus(Enum):
+    """Enum for indicator status."""
+
+    FRESH = "Fresh"
```
JudahSchwartz, line 44: Is there a tim api module this should go to? Or should we consider adding to csp with the other tim stuff?

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+    type: str
+    value_field: str
+    context_path_prefix: str
+    context_output_mapping: dict[str, str]
+
+
+class CommandType(Enum):
+    """
+    Execution policy category:
+      - INTERNAL: runs when no brands specified, or if its brand is requested (e.g., wildfire-get-verdict).
```
Shellyber, line 104: This is when external_enrichment is False, right?

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+        """
+        final_args = self.args.copy() if self.args else {}
+        if brands_to_run and not self.ignore_using_brand:
+            final_args["using-brand"] = ",".join(brands_to_run)
+        return {self.name: final_args}
+
+    def execute(self) -> list[dict[str, Any]]:
+        """Executes the command and returns the results."""
+        demisto.debug(f"[Command.execute] Executing command {self.name} with args: {self.args}")
+        results = execute_command(self.name, self.args)
```
Shellyber: Currently execute assumes execute_command always succeeds and returns a list. If execute_command can raise or return None, you might want to guard it with try except

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1193 @@
...
+    ) -> list[list[list[tuple[ContextResult, str, str]]]]:
+        """
+        Execute batches in order. See `execute_batch` for inner tuple shape.
+
+        Args:
+            list_of_batches (list[list[Command]]): A list of batches to execute. each batch is list of commands.
+            brands_to_run (list[str]): A list of brands to run on.
+            verbose (bool): Whether to print verbose output.
+
+        Returns:
```
Shellyber, line 273: In our docs later, or even here, let's add an example how it looks
Probably like that:
```
    BatchesResults: Nested structure:
        [
          [  # batch 0
            [ (result, hr_output, error), ... ],  # command 0 results
            [ (result, hr_output, error), ... ],  # command 1 results
          ],
          [  # batch 1
            ...
          ]
        ]
```

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
Shellyber, line 21: Don't we have it in CommonServerPython already ?

Shellyber (commented):
Really nice work

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
almog2296: it doesn't rely on the external_enrichment, only on the brands. if brands is empty it will run, or if it specific brand was chosen
Example get-endpoint-data will always unless we specified brands without the Core

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
almog2296: We got where 1:"Good"
They wanted 1:"Benign" :\

Tests/demistomock/demistomock.py
```diff
@@ -708,6 +708,31 @@ def executeCommand(command, args):
 
     return ""
 
+def executeCommandBatch(commands_list):
+    """(Script only)
+    Execute list of commands in the following format {"command_name":args}
```
Shellyber, line 713:
```suggestion
    Execute list of commands in the following format {"command_name":args}
```
Let's give a better example of how it should run. More than one command

Packs/ApiModules/ReleaseNotes/2_3_0.md
```diff
@@ -0,0 +1,5 @@
+#### Scripts
+
+##### AggregatedCommandApiModule
+
+- Added AggregatedCommandApiModule.
```
Shellyber: Let's write better release notes
```suggestion
- Common Aggregated Command API Module provides a generic Infrastructure for different kinds of aggregated commands.
```

talihaff:
@almog2296 only checked Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.yml

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
```diff
@@ -0,0 +1,124 @@
...
+        "DetectionEngines": "DetectionEngines",
+        "PositiveDetections": "PositiveDetections",
+        "Score": "Score",
+        "Brand": "Brand",
+    }
+
+    domain_indicator = Indicator(
+        type="domain",
+        value_field="Name",
+        context_path_prefix="Domain(",
```
Shellyber: Why is the context_path_prefix is with "(" always at the end?

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
```diff
@@ -0,0 +1,124 @@
...
+        context_output_mapping=indicator_mapping,
+    )
+
+    # --- Batch 1: create indicators (BUILTIN) ---
```
Shellyber: Don't you want to add more debug logs here? at the start of each step? (probably true to all scripts)

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.yml
```diff
@@ -0,0 +1,330 @@
...
+  defaultValue: 'false'
+comment: This script enriches Domains data with information from multiple integrations and returns a "DomainEnrichment" object with consolidated information in the context output.
+commonfields:
+  id: domain-enrichment
+  version: -1
+enabled: false
+name: domain-enrichment
+outputs:
+# DomainEnrichment
+- contextPath: DomainEnrichment.Value
```
Shellyber, line 63: Didn't see the `reliability` field here, missed it?

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
Shellyber: XSOAR doesn't transform it automatically?

Shellyber (commented):
Awesome work!!

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
almog2296: There is no consistency on the Path of the Indicators context
Some like this:
URL(val.Data && val.Data == obj.Data)
Some like this URL(val.Data && val.Data == obj.Data && val.Vendor == obj.Vendor)
I tried to use "URL" as the prefix to extract the right context from each indicator but there is the integration URLhaus so her context contain URL(....) and  URLhaus(....), so i added the "(" and made it the same for all.
But it only relevant for the URL actually do you suggest to change it ?

Shellyber: Yeah, I think leave it to URL only with a comment why we add this, and the rest, let's use the standard

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
```diff
@@ -0,0 +1,140 @@
...
+    demisto.debug("Creating commands")
```
Shellyber: I was thinking maybe something like: "Domain Enrichment - creating new indicator
Same for the batch 2 and further, to make it more indicative

Shellyber (commented):
Great work. 
Let's do some final testing before merging to verify nothing broke. 
See my last comment as well

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1219 @@
...
+            commands (list[list[Command]]): List of batches of commands to run.
+        """
+        self.args = args
+        self.brands = brands
+        self.verbose = verbose
+        self.commands = commands or []
+        self.brand_manager = BrandManager(brands)
+
+    @cached_property
+    def unsupported_enrichment_brands(self) -> list[str]:
```
JudahSchwartz:
```suggestion
    def unsupported_brands(self) -> list[str]:
```
is this enrichment specific? We are in the base class here

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
```diff
@@ -0,0 +1,1219 @@
...
+    """
+    Executes commands in batch/batches and performs minimal, uniform result processing.
+    """
+
+    def process_results(
+        self,
+        results: list[list[ContextResult]],
+        commands_list: list[Command],
+        verbose: bool = False,
+    ) -> list[list[tuple[ContextResult, str, str]]]:
```
JudahSchwartz: I think we should type alias `list[list[tuple[ContextResult, str, str]]]` everywhere so that its clearer when we refer to this

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
```diff
@@ -0,0 +1,138 @@
...
+            args={"indicator_values": domain_list, "type": "Domain"},
+            command_type=CommandType.BUILTIN,
+            context_output_mapping=None,
+            ignore_using_brand=True,
+        )
+    ]
+
+    # --- Command Batch 2: internal analytics + external enrichment ---
+    command_batch2: list[Command] = []
+    if is_xsiam():
```
JudahSchwartz, line 60: can we use the same mechanism to check if core brand is enabled, instead of is_xsiam? We shouldnt treat as a special case if we dont need to. Same comment for IP

JudahSchwartz (approved):

JudahSchwartz:
lets make sure this is the repo we want to merge to fyi @Shellyber

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
almog2296: They wanted 1: Benign, in the CSP its 1:Good

Packs/ApiModules/Scripts/AggregatedCommandApiModule/AggregatedCommandApiModule.py
almog2296: I will move it to the Reputation it specific for unsupported enrichment brands and not internal brands

Packs/AggregatedScripts/Scripts/DomainEnrichment/DomainEnrichment.py
almog2296: Even if the integration is disabled i still want to add this command to the batch, in case the user will give this brand and the integration is not enabled i want the human readable will have a message saying this integration is not enabled.
And for xsoar i dont want to add the command because the integration is not exists on xsoar so i dont want to allways get a fail message

PR #41476, opened 2025-10-05
Pan os dynamic updates

rshunim:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-56913

## Description
Fixed an issue where the ***pan-os-check-dynamic-updates-status*** command was failing when there was no WildFire license on the Firewall.

## Must have
- [ ] Tests
- [ ] Documentation


rshunim:
@yuvalbenshalom 
ForceMerge is needed here due to an issue in the "BlockExternalIP - Test" TPB where it fails on the "block-external-ip" command in the build but locally it passes.
@noydavidi is trying to solve it

PR #42355, opened 2025-12-22
Add prisma cloud compute assets collector

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [CRTX-212383](https://jira-dc.paloaltonetworks.com/browse/CRTX-212383)

## Description:
Added a fetch assets mechanism for PrismaCloudCompute.
The issue is for one specific custoemr and therefore won't be merged.


Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +589,538 @@ def get_runtime_container_policy(self) -> dict:
...
+class AsyncClient:  # pragma: no cover
+    """
+    Base asynchronous client for interacting with APIs.
+    Handles session and authentication management.
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        headers: dict = {
```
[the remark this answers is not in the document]
JudahSchwartz: maskim

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +589,538 @@ def get_runtime_container_policy(self) -> dict:
...
+                zipped_data=zipped_data,
+                headers=headers,
+                num_of_attempts=num_of_attempts,
+                data_type=data_type,
+                error_handler=data_error_handler,
+                error_msg=header_msg,
+                is_json_response=True,
+            )
+
+        return chunk_size
```
JudahSchwartz: how will rest of fn execute if we return here?

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +589,538 @@ def get_runtime_container_policy(self) -> dict:
...
+            method="POST",
+            endpoint="/logs/v1/xsiam",
+            data=zipped_data,
+            headers=headers,
+            error_handler=error_handler,
+            ok_codes=ok_codes,
+        )
+        status_code = response.status
+        if status_code == 429:
+            time.sleep(1)
```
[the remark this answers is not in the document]
JudahSchwartz: nice catch

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +589,538 @@ def get_runtime_container_policy(self) -> dict:
...
+            the server's TLS certificate, or a string that represents a path to a CA bundle to use.
+        project: A projectID string, set in the integration parameters.
+            the projectID is saved under self._project
+        """
+
+        self._project = project
+        auth = (username + ":" + password).encode("utf-8")
+        self.auth_header = {"Authorization": "Basic " + b64encode(auth).decode()}
+
+        if verify in ["True", "False"]:
```
JudahSchwartz, line 1041: explain this block? something seems weird here

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2798,39 +3339,275 @@ def get_container_policy_list_command(client: PrismaCloudComputeClient, args: di
...
+            ctx_lock=ctx_lock,
+        )
+    else:
+        return AssetTypeRelatedData(
+            endpoint=endpoint, product=product, asset_type=asset_type, process_result_func=process_result_func, ctx_lock=ctx_lock
+        )
+
+
+async def preform_fetch_assets_main_loop_logic(client: PrismaCloudComputeAsyncClient):
+    ctx_lock = threading.Lock()
```
JudahSchwartz: writing this before i get into code - should this be threading lock, not asyncio lock?

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2798,39 +3341,272 @@ def get_container_policy_list_command(client: PrismaCloudComputeClient, args: di
...
+            ctx_lock=ctx_lock,
+        )
+    else:
+        return AssetTypeRelatedData(
+            endpoint=endpoint, product=product, asset_type=asset_type, process_result_func=process_result_func, ctx_lock=ctx_lock
+        )
+
+
+async def preform_fetch_assets_main_loop_logic(client: PrismaCloudComputeAsyncClient):
+    ctx_lock = asyncio.Lock()
```
JudahSchwartz: probably should be asyncio lock

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2798,39 +3341,272 @@ def get_container_policy_list_command(client: PrismaCloudComputeClient, args: di
...
+        collect_assets_and_send_to_xsiam(client, image_scan_related_data),
+    ]
+    demisto.debug("Initiated all asset collection tasks.")
+    await asyncio.gather(*tasks)
+    total = tas_droplets_related_data.total_count + host_scan_related_data.total_count + image_scan_related_data.total_count
+    demisto.debug(
+        f"Finished sending all assets to XSIAM. Fetch {tas_droplets_related_data.total_count} Tas Droplets,"
+        f"{host_scan_related_data.total_count} Hosts, and {image_scan_related_data.total_count} Runtime images."
+    )
+    demisto.updateModuleHealth({"assetsPulled": total})
```
JudahSchwartz: we should also update integration context here

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2798,39 +3341,272 @@ def get_container_policy_list_command(client: PrismaCloudComputeClient, args: di
...
+async def collect_assets_and_send_to_xsiam(client: PrismaCloudComputeAsyncClient, asset_type_related_data: AssetTypeRelatedData):
+    """Implement the main loop, sending requests to Prisma to fetch assets, process the assets and then send them to xsiam.
+    Saving the current advancement between intervals.
+
+    Args:
+        client (PrismaCloudComputeAsyncClient): The Prisma cloud compute client.
+        asset_type_related_data (AssetTypeRelatedData): The object that contains that data related to the specific asset type.
+    """
+    asset_type_related_data.write_debug_log(f"starting execution with snapshot = {asset_type_related_data.snapshot_id}.")
+    while True:
```
JudahSchwartz, line 3527: while True here is confusing, how do i know I've brought all the data? IMO obtain_asset_data_from_prisma should be a generator that goes until its done, and this loop should look like

```
async for assets in obtain_assets...:
      update assets to xsiam
```

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+
+    def next_page(self):
+        self.offset += self.limit
+
+    def write_debug_log(self, msg: str):
+        demisto.debug(f"[{self.asset_type.value}] {msg}")
+
+    async def safe_update_integration_context(self):
+        async with self.ctx_lock:
+            ctx = get_integration_context()
```
JudahSchwartz, line 1105: technically unneeded since none of these yield to the event loop, but i guess it feels safer to have

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+    process_result_func: Callable
+    offset: int = 0
+    limit: int = MAX_API_LIMIT
+    total_count: int = 0
+    vendor: str = "Prisma_Cloud_Compute"
+    snapshot_id: str = str(round(time.time() * 1000))
+
+    def next_page(self):
+        self.offset += self.limit
+
```
JudahSchwartz, line 1099:
```suggestion
    def next_page(self):
        offset= self.offset
        limit = self.limit
        self.offset += self.limit
        offset, limit
```

this is cleaner, make these vars private also
Current usage i need to remember to call these vars and to let this object know i need it to do next page, kind of messy and easy to forget when using

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+    def write_debug_log(self, msg: str):
+        demisto.debug(f"[{self.asset_type.value}] {msg}")
+
+    async def safe_update_integration_context(self):
+        async with self.ctx_lock:
```
JudahSchwartz, line 1104: self lock wont lock when other asset types are doing this manipulation, is this on purpose?

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+
+
+class AssetType(Enum):
+    TAS_DROPLET = "Tas Droplet"
+    RUNTIME_IMAGE = "Runtime Image"
+    HOST = "Host"
+
+
+@dataclass
+class AssetTypeRelatedData:
```
JudahSchwartz, line 1084: wouldnt refactor now, but i dont really like this kolboinick object that gets passed around. Maybe would be better to have an "AssetFetcher" class that knows how to get all the pages, internally paginate, and an XsiamSender class that knows to update integration context and do the work. This here doesnt seem to have a clear responsibility

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+
+        :type res: ``requests.Response``
+        :param res: Response from API after the request for which to check error type
+
+        :type error_handler ``callable``
+        :param error_handler: Given an error entry, the error handler outputs the
+            new formatted error message.
+        """
+        if error_handler:
+            error_handler(res)
```
JudahSchwartz, line 654: this is bad. Function shouldnt be "Passed in" Just override in base

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+            err_msg += f"\n{res.text}"
+            raise DemistoException(err_msg, res=res)
+
+    async def _http_request(
+        self,
+        method: str,
+        endpoint: str,
+        payload: dict = None,
+        headers: dict = None,
+        retryable_statuses: set = {500, 502, 503, 504, 429},
```
JudahSchwartz, line 681: let this field also be set on init, this might be wanted behavior that isnt unique to specific request

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+        params: dict = {},
+        timeout: int = 60,
```
JudahSchwartz, line 683: would add timeout also to init as a default

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +588,541 @@ def get_runtime_container_policy(self) -> dict:
...
+        auth=None,
+        error_handler=None,
+        num_of_retires=1,
```
JudahSchwartz: retries not retires

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -2798,39 +3341,272 @@ def get_container_policy_list_command(client: PrismaCloudComputeClient, args: di
...
+        asset_type_related_data (AssetTypeRelatedData): The object that contains that data related to the specific asset type.
+    """
+    asset_type_related_data.write_debug_log(f"starting execution with snapshot = {asset_type_related_data.snapshot_id}.")
+    while True:
+        try:
+            data = await obtain_asset_data_from_prisma(asset_type_related_data=asset_type_related_data, client=client)
+            if not data:
+                asset_type_related_data.write_debug_log("No more data to fetch, breaking.")
+                break
+            await process_asset_data_and_send_to_xsiam(data=data, asset_type_related_data=asset_type_related_data)
```
JudahSchwartz, line 3533: this await here means that this collcetor cannot go back to bring next page until this finishes sending to xsiam. would consider async generator like above (or otherwise maybe create task? dont remember here exact syntax)

JudahSchwartz (commented):
Looks good, theres a couple things here i would insist on if this wasnt a PR for a "once off", so let me know if theres some comments here you dont think is worth the investment

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: Since we're doing this job for one specific customer I don't think it's necessary.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: lock is being passed when the Object is initiated.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: Since we have 3 different types of assets/endpoints to fetch from, I believe this is not a big blocker to be dependent on as we can go fetch assets of type 2 when sending assets of type 1 to xsiam.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: I'm not sure what's the point of your suggestion

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: removed the return statement, we don't really need any info to return.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: Generally, I agree, in that case I wouldn't bother changing now, but the idea is that we use the `if not data` statement a few lines below to break.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: got it.
It was some sort of copy-paste from the BaseClient logic.
I wouldn't bother changing now, but got it.

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
YuvHayun: It's the same as the implementation in the basic Prisma client.

JudahSchwartz: so wont try to understand then since this is once off, but looks weird

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
JudahSchwartz: (Style, ignore if this already works)
But thw style point here is i think that this class is hard to use. I shouldnt need to worry about The fact that it needs to update itself internally from outside this class, so calling object.next_page, which doesnt do anything for me outside this class, is easy to forget and shouldnt be my concern

having this function also return the info i need, and also update internally automatically, makes it easier not to forget to call next_page (since i dont have to) - which i shouldnt have to do anyway since its not needed outside the object.

Basically, theres an encapsulation issue

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
JudahSchwartz: this is true for all of them - what could be a 6x parallel is now only 3. Your call if you think its worth it, seems like an easy win to me. If this wasnt a once off i would insist

Packs/PrismaCloudCompute/Integrations/PaloAltoNetworks_PrismaCloudCompute/PaloAltoNetworks_PrismaCloudCompute.py
```diff
@@ -580,6 +589,538 @@ def get_runtime_container_policy(self) -> dict:
...
+        else:
+            demisto.debug("Initializing an async client with a certificate.")
+            # verify points a path to certificate
+            super().__init__(
+                base_url=base_url,
+                verify=True,
+                proxy=proxy,
+                headers=headers,
+            )
+            self._verify = verify
```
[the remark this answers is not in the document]
JudahSchwartz: I think this needs more explanation. Dont just say "wrong logic", explain why its wrong

JudahSchwartz (approved):

JudahSchwartz:
@YuvHayun approved. Not "approving" since this shouldnt be merged

PR #44315, opened 2026-05-15
Miplabelingcommands

ymishrapalo1992:
This PR applies all reviewer feedback on the original sensitivity-label commands added to the `Packs/MicrosoftGraphFiles/` integration.

### Commands in the pack after this PR

| Command | What it does | Graph endpoint | Method |
|:--|:--|:--|:--|
| `msgraph-extract-sensitivity-label` | Reads the sensitivity label(s) currently assigned to a single drive item. | `/v1.0/{drive-prefix}/items/{itemId}/extractSensitivityLabels` | `POST` |
| `msgraph-apply-sensitivity-label` | Assigns a sensitivity label to a single drive item. Long-running operation. | `/v1.0/{drive-prefix}/items/{itemId}/assignSensitivityLabel` | `POST` |

`{drive-prefix}` is `drives/{id}` for `object_type=drives`, otherwise `{object_type}/{id}/drive` — same branching as every existing drive-item command in the pack.

### Changes addressed in this PR

**Blocking**

1. **Removed `msgraph-list-sensitivity-labels` entirely.** The `/beta/security/informationProtection/sensitivityLabels` endpoint is beta-only with no v1.0 GA equivalent. Pack policy forbids shipping beta-API commands. Deleted from `.py` (constants, client method, handler, dispatch), `.yml` (block + outputs), tests (2 tests + fixture), and release notes. README auto-regenerated.

2. **Fixed `msgraph-extract-sensitivity-label` response parsing.** Previously read the wrong key (`extractedSensitivityLabel`) and always returned empty. Now correctly parses the real MS Graph response shape:
    ```json
    { "value": { "labels": [ { "sensitivityLabelId": "...", "assignmentMethod": "..." } ] } }
    ```
   Added new additive output `MsGraphFiles.ExtractedSensitivityLabel.labels` (full array) while keeping the flat `sensitivityLabelId` / `assignmentMethod` fields populated from the first entry. Empty list still returns success.

3. **`msgraph-apply-sensitivity-label` — 202 Accepted + Location header.**
    - Added `202` to `ok_codes` (now `(200, 201, 202, 204)`) — the endpoint is a long-running operation that returns HTTP 202.
    - New output `MsGraphFiles.AppliedSensitivityLabel.operationLocation` exposes the `Location` response header so callers can poll the operation-status URL.
    - `result="SUCCESS"` semantics preserved on 202.

**Cleanup**

4. Removed unnecessary module-level constants (`EXTRACT_…_URL_SUFFIX`, `APPLY_…_URL_SUFFIX`, `DEFAULT_JUSTIFICATION_TEXT`, `DEFAULT_ASSIGNMENT_METHOD`); inlined literals at call sites.
5. Dropped `Calls the MS Graph endpoint: POST ...` lines from client-method docstrings.
6. Converted `_build_drive_item_uri`, `extract_sensitivity_label`, `apply_sensitivity_label` docstrings from `:param:` / `:return:` to Google-style `Args:` / `Returns:`.
7. Removed `INTEGRATION_NAME` prefix from `tableToMarkdown` headers and from `outputs_prefix` — both now use the plain literal `"MsGraphFiles.…"`, matching the pack convention used by the `SitePermission` commands.
8. Updated `sensitivity_label_id` arg description in the yml — now points users to the Microsoft Purview compliance portal or the PowerShell `Get-Label` cmdlet for obtaining label GUIDs (the list command was removed).
9. Trimmed [`ReleaseNotes/1_2_0.md`](Packs/MicrosoftGraphFiles/ReleaseNotes/1_2_0.md:1) to only mention the 2 remaining commands.

### Tests performed

10 unit tests (`pytest` + `requests_mock`, zero live API calls):

- `test_extract_sensitivity_label_command_happy_path` — single-label response, uses real `{"value": {"labels": [...]}}` shape.
- `test_extract_sensitivity_label_command_empty_label` — empty `labels` list returns empty `sensitivityLabelId` / `assignmentMethod` without raising.
- `test_extract_sensitivity_label_command_multiple_labels` — multiple labels preserved in `labels[]`, flat fields populated from first entry.
- `test_extract_sensitivity_label_uri_branching` (parametrised: `drives`, `users`, `sites`, `groups`) — URI is built correctly per `object_type`.
- `test_apply_sensitivity_label_command_happy_path` — mocks `status_code=202` + `Location` header, verifies `result=SUCCESS`, `httpStatusCode=202`, `operationLocation` populated, custom `assignment_method` / `justification_text` reach the request body.
- `test_apply_sensitivity_label_uri_branching` (parametrised over the 4 object types) — URI shape and JSON body verified for each.
- `test_apply_sensitivity_label_propagates_http_errors` (parametrised: 400, 403, 404, 503) — verifies `DemistoException` propagation with the raw Graph error message.
- `test_sensitivity_label_commands_missing_required_args` — required arg validation.

### Lifecycle commands

| Command | Result |
|:--|:--|
| `demisto-sdk format -i Packs/MicrosoftGraphFiles` | ✅ |
| `demisto-sdk validate -i Packs/MicrosoftGraphFiles` | ✅ All validations passed |
| `demisto-sdk pre-commit -i Packs/MicrosoftGraphFiles` | ✅ All hooks pass: `pylint`, `mypy`, `pytest`, `ruff`, `xsoar-lint`, `markdownlint`, `secrets`, `validate-content-paths` |
| `demisto-sdk generate-docs` | ✅ README regenerated; removed command section gone, new `labels` and `operationLocation` outputs present |

### Files changed

- [`Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py`](Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py:1) — 2 client methods, 2 command handlers, `main()` dispatch entries, shared URI helper, `ok_codes` updated.
- [`Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml`](Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml:1) — 2 command blocks with updated outputs (`labels`, `operationLocation`).
- [`Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles_test.py`](Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles_test.py:1) — 10 tests.
- [`Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/README.md`](Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/README.md:1) — regenerated.
- [`Packs/MicrosoftGraphFiles/ReleaseNotes/1_2_0.md`](Packs/MicrosoftGraphFiles/ReleaseNotes/1_2_0.md:1) — generic, vendor-focused notes for the 2 remaining commands.
- [`Packs/MicrosoftGraphFiles/pack_metadata.json`](Packs/MicrosoftGraphFiles/pack_metadata.json:1) — version `1.2.0` (minor bump for 2 net new commands).

### Backward compatibility

No existing command, argument, or `MsGraphFiles.*` context path modified. The new outputs (`labels`, `operationLocation`) are purely additive. All pre-existing tests continue to pass alongside the 10 new ones.


Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -11,6 +11,14 @@
 INTEGRATION_NAME = "MsGraphFiles"
 APP_NAME = "ms-graph-files"
 
+GRAPH_API_SENSITIVITY_LABELS_URL = "https://graph.microsoft.com/beta/security/informationProtection/sensitivityLabels"
```
noydavidi: This endpoint (/beta/security/informationProtection/sensitivityLabels) is beta-only, there is no v1.0 GA equivalent. We don't support beta endpoints as Microsoft can introduce breaking changes without notice. Please remove the msgraph-list-sensitivity-labels command entirely (including its client method, command function, YML definition, tests, README section, and release note entry).

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1146,170 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+
+    Returns:
+        CommandResults with the extracted label fields.
+    """
+    object_type = args["object_type"]
+    object_type_id = args["object_type_id"]
+    item_id = args["item_id"]
+
+    raw_response = client.extract_sensitivity_label(object_type, object_type_id, item_id)
+    extracted = raw_response.get("extractedSensitivityLabel") or []
```
noydavidi: According to the [MS Graph docs](https://learn.microsoft.com/en-us/graph/api/driveitem-extractsensitivitylabels?view=graph-rest-1.0), the response structure is 
`{ "value": { "labels": [ { "sensitivityLabelId": "...", "assignmentMethod": "..." } ] } }`
The key is labels inside value, not extractedSensitivityLabel. This means the command will always return empty results. Please fix the parsing to match the actual API response. The test mock data should also be updated to match the real response structure.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1146,170 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+    response = client.apply_sensitivity_label(
+        object_type=object_type,
+        object_type_id=object_type_id,
+        item_id=item_id,
+        sensitivity_label_id=sensitivity_label_id,
+        assignment_method=assignment_method,
+        justification_text=justification_text,
+    )
+
+    status_code = getattr(response, "status_code", 200)
```
noydavidi: The [MS Graph docs](https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-1.0) state that assignSensitivityLabel returns 202 Accepted (not 200 OK), and it's a long-running operation with a Location header for monitoring progress. Three issues:

1. The integration's ok_codes is (200, 204, 201) — 202 is missing, so the API module will raise an error on a successful 202 response. Add 202 to ok_codes.
2. The tests mock status_code=200 but the real API returns 202. Update the test mocks accordingly.
3. Please surface the Location header URL to the outputs so users can monitor the operation status.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,127 @@ script:
...
+        For resource type 'drive': To get a list of all drives in your site, use the msgraph-list-drives-in-site command.
+        For resource type 'group': To get a list of all groups that exists, configure the 'Entra ID Groups' integration and use the msgraph-groups-list-groups command.
+        For resource type 'sites': To get a list of all sites, use the msgraph-list-sharepoint-sites command.
+        For resource type 'users': To get a list of all users that exists, configure the 'Entra ID Users' integration and use the msgraph-user-list command.
+      name: object_type_id
+      required: true
+    - description: The ID of the drive item to assign the sensitivity label to.
+      name: item_id
+      required: true
+    - description: The GUID of the sensitivity label to assign. Use the msgraph-list-sensitivity-labels command to retrieve available label IDs.
```
noydavidi: Since the list command needs to be removed (beta endpoint), update this description to guide users on where to find label GUIDs, e.g., the Microsoft Purview compliance portal or PowerShell Get-Label.

Packs/MicrosoftGraphFiles/ReleaseNotes/1_2_0.md
```diff
@@ -0,0 +1,8 @@
+
+#### Integrations
+
+##### O365 File Management (Onedrive/Sharepoint/Teams)
+
+- Added the ***msgraph-list-sensitivity-labels*** command to retrieve sensitivity labels available in the tenant from Microsoft Information Protection.
```
noydavidi: After removing the msgraph-list-sensitivity-labels command, update the release notes to only mention the two remaining commands.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -11,6 +11,14 @@
 INTEGRATION_NAME = "MsGraphFiles"
 APP_NAME = "ms-graph-files"
 
+GRAPH_API_SENSITIVITY_LABELS_URL = "https://graph.microsoft.com/beta/security/informationProtection/sensitivityLabels"
+EXTRACT_SENSITIVITY_LABEL_URL_SUFFIX = "/extractSensitivityLabels"
+APPLY_SENSITIVITY_LABEL_URL_SUFFIX = "/assignSensitivityLabel"
```
noydavidi: We don't need to create constant variables for the endpoints, you can remove it and just use the values instead.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+            method="GET",
+            full_url=GRAPH_API_SENSITIVITY_LABELS_URL,
+        )
+
+    def extract_sensitivity_label(self, object_type: str, object_type_id: str, item_id: str) -> dict:
+        """
+        Retrieve the sensitivity label currently assigned to a drive item.
+
+        Calls the MS Graph endpoint:
+            POST /v1.0/{drive-prefix}/items/{itemId}/extractSensitivityLabels
```
noydavidi: You can remove the endpoint from the description

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+
+        :param object_type: ms graph resource.
+        :param object_type_id: ms graph resource id.
+        :param item_id: the drive item id.
+        :return: graph api raw response
```
noydavidi: Use the Args / Returns pattern instead of the param / return pattern.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+        Calls the MS Graph endpoint:
+            POST /v1.0/{drive-prefix}/items/{itemId}/assignSensitivityLabel
+
+        :param object_type: ms graph resource.
+        :param object_type_id: ms graph resource id.
+        :param item_id: the drive item id.
+        :param sensitivity_label_id: the GUID of the sensitivity label to assign.
+        :param assignment_method: one of 'standard', 'privileged', 'auto'.
+        :param justification_text: free-text justification recorded with the assignment.
+        :return: raw HTTP response object so the caller can inspect the status code.
```
noydavidi: Use the Args / Returns pattern instead of the param / return pattern.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+        item_id: str,
+        sensitivity_label_id: str,
+        assignment_method: str = DEFAULT_ASSIGNMENT_METHOD,
+        justification_text: str = DEFAULT_JUSTIFICATION_TEXT,
+    ) -> requests.Response:
+        """
+        Assign a sensitivity label to a drive item.
+
+        Calls the MS Graph endpoint:
+            POST /v1.0/{drive-prefix}/items/{itemId}/assignSensitivityLabel
```
noydavidi: You can remove the endpoint from the description

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1146,170 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+        first = extracted if isinstance(extracted, dict) else {}
+
+    outputs = {
+        "itemId": item_id,
+        "sensitivityLabelId": first.get("sensitivityLabelId", ""),
+        "assignmentMethod": first.get("assignmentMethod", ""),
+    }
+
+    readable_output = tableToMarkdown(
+        name=f"{INTEGRATION_NAME} - Extracted Sensitivity Label",
```
noydavidi: No need to add the integration name to the outputs header.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1146,170 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+    status_code = getattr(response, "status_code", 200)
+    outputs = {
+        "itemId": item_id,
+        "sensitivityLabelId": sensitivity_label_id,
+        "result": "SUCCESS",
+        "httpStatusCode": status_code,
+    }
+
+    readable_output = tableToMarkdown(
+        name=f"{INTEGRATION_NAME} - Applied Sensitivity Label",
```
noydavidi: No need to add the integration name to the outputs header.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -11,6 +11,14 @@
 INTEGRATION_NAME = "MsGraphFiles"
 APP_NAME = "ms-graph-files"
 
+GRAPH_API_SENSITIVITY_LABELS_URL = "https://graph.microsoft.com/beta/security/informationProtection/sensitivityLabels"
+EXTRACT_SENSITIVITY_LABEL_URL_SUFFIX = "/extractSensitivityLabels"
+APPLY_SENSITIVITY_LABEL_URL_SUFFIX = "/assignSensitivityLabel"
+
+DEFAULT_JUSTIFICATION_TEXT = "Applied via Cortex XSOAR"
+DEFAULT_ASSIGNMENT_METHOD = "standard"
+DEFAULT_LABELS_LIMIT = 50
```
noydavidi: Let's remove those constants as well and just use the values instead where needed.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+    def _build_drive_item_uri(object_type: str, object_type_id: str, item_id: str) -> str:
+        """
+        Build the drive-item URI prefix based on the MS Graph resource type.
+
+        Mirrors the existing branching used by other drive-item commands in this integration.
+
+        :param object_type: ms graph resource ('drives', 'groups', 'sites', 'users').
+        :param object_type_id: ms graph resource id.
+        :param item_id: the drive item id.
+        :return: relative URI of the drive item, with no trailing slash.
```
noydavidi: Use the Args / Returns pattern instead of the param / return pattern.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -527,6 +535,89 @@ def create_new_folder(self, object_type: str, object_type_id: str, parent_id: st
...
+        """
+        if object_type == "drives":
+            return f"drives/{object_type_id}/items/{item_id}"
+        if object_type in {"groups", "sites", "users"}:
+            return f"{object_type}/{object_type_id}/drive/items/{item_id}"
+        raise DemistoException(f"Invalid object_type '{object_type}'. Must be one of: drives, groups, sites, users.")
+
+    def list_sensitivity_labels(self) -> dict:
```
noydavidi: Don't forget to remove

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1146,170 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+        "description": label.get("description"),
+        "color": label.get("color"),
+        "isActive": label.get("isActive"),
+        "sensitivity": label.get("sensitivity"),
+        "tooltip": label.get("tooltip"),
+        "parent": {"id": parent.get("id")} if parent.get("id") else None,
+    }
+
+
+def list_sensitivity_labels_command(client: MsGraphClient, args: dict[str, str]) -> CommandResults:
```
noydavidi: Don't forget to remove

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1132,6 +1387,12 @@ def main():
             return_results(update_site_permissions_command(client, args))
         elif command == "msgraph-delete-site-permissions":
             return_results(delete_site_permission_command(client, args))
+        elif command == "msgraph-list-sensitivity-labels":
```
noydavidi: Don't forget to remove

noydavidi (changes requested):
Thanks for the contribution! The **extract** and **apply** sensitivity label commands are a great addition to this integration. However, there are a few blockers that need to be addressed:

**msgraph-list-sensitivity-labels** uses a **beta**-only endpoint — we can't ship beta API commands. Please remove it.
extract_sensitivity_label_command reads the wrong response key — the MS Graph API returns labels inside value, not extractedSensitivityLabel. This means the command will always return empty results.
assignSensitivityLabel returns 202 Accepted (long-running operation), but ok_codes doesn't include 202, so the command will error on success. Tests also mock 200 instead of 202.
Please fix these issues and my other comments and update the tests/docs accordingly. After you have changed the code don't forget to check again the commands :)

When you are done with the changes please ask for re-review from me, thanks :)

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+            justification_text (Optional): Free-text justification recorded with the assignment.
+
+    Returns:
+        CommandResults with the apply outcome.
+    """
+    object_type = args["object_type"]
+    object_type_id = args["object_type_id"]
+    item_id = args["item_id"]
+    sensitivity_label_id = args["sensitivity_label_id"]
+    assignment_method = args.get("assignment_method") or "standard"
```
noydavidi:
```suggestion
    assignment_method = args.get("assignment_method")
```

No need for default because this is not a required argument.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1132,6 +1314,10 @@ def main():
             return_results(update_site_permissions_command(client, args))
         elif command == "msgraph-delete-site-permissions":
             return_results(delete_site_permission_command(client, args))
+        elif command == "msgraph-extract-sensitivity-label":
+            return_results(extract_sensitivity_label_command(client, args))
+        elif command == "msgraph-apply-sensitivity-label":
```
noydavidi:
```suggestion
        elif command == "msgraph-assign-sensitivity-label":
```
Please apply to any relevant code line

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
noydavidi: Please remove the default we want this to pop in case of any API change.

```suggestion
    status_code = getattr(response, "status_code")
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+
+    Returns:
+        CommandResults with the apply outcome.
+    """
+    object_type = args["object_type"]
+    object_type_id = args["object_type_id"]
+    item_id = args["item_id"]
+    sensitivity_label_id = args["sensitivity_label_id"]
+    assignment_method = args.get("assignment_method") or "standard"
+    justification_text = args.get("justification_text") or "Applied via Cortex XSOAR"
```
noydavidi:
```suggestion
    justification_text = args.get("justification_text")
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+            item_id (Required): The drive item ID.
+
+    Returns:
+        CommandResults with the extracted label fields. The flat `sensitivityLabelId` and
+        `assignmentMethod` fields reflect the first entry in `labels` (or empty strings
+        when no label is assigned); `labels` contains the full array as returned by Graph.
+    """
+    object_type = args["object_type"]
+    object_type_id = args["object_type_id"]
+    item_id = args["item_id"]
```
noydavidi: Better to use dict.get("") even if this is required. Please apply anywhere relevant.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+
+    raw_response = client.extract_sensitivity_label(object_type, object_type_id, item_id)
+    labels = (raw_response.get("value") or {}).get("labels") or []
```
noydavidi:
```suggestion
    labels = (raw_response.get("value",{})).get("labels",[])
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+        "itemId": item_id,
+        "sensitivityLabelId": first.get("sensitivityLabelId", ""),
+        "assignmentMethod": first.get("assignmentMethod", ""),
+        "labels": labels,
+    }
+
+    readable_output = tableToMarkdown(
+        name="Extracted Sensitivity Label",
+        t=outputs,
+        headers=["itemId", "sensitivityLabelId", "assignmentMethod"],
```
noydavidi:
```suggestion
        headers=["itemId", "sensitivityLabelId", "assignmentMethod"],
        headerTransform=pascalToSpace,
```

noydavidi: This is for changing the titles from sensitivityLabelId to -> Sensitivity Label Id

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+            assignment_method (Optional): One of standard, privileged, auto. Default standard.
+            justification_text (Optional): Free-text justification recorded with the assignment.
+
+    Returns:
+        CommandResults with the apply outcome.
+    """
+    object_type = args["object_type"]
+    object_type_id = args["object_type_id"]
+    item_id = args["item_id"]
+    sensitivity_label_id = args["sensitivity_label_id"]
```
noydavidi: Better to use dict.get("") even if this is required. Please apply anywhere relevant.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+        "sensitivityLabelId": sensitivity_label_id,
+        "result": "SUCCESS",
+        "httpStatusCode": status_code,
+        "operationLocation": operation_location,
+    }
+
+    readable_output = tableToMarkdown(
+        name="Applied Sensitivity Label",
+        t=outputs,
+        headers=["itemId", "sensitivityLabelId", "result", "httpStatusCode", "operationLocation"],
```
noydavidi:
```suggestion
        headers=["itemId", "sensitivityLabelId", "result", "httpStatusCode", "operationLocation"],
        headerTransform=pascalToSpace
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.assignmentMethod
+      description: The assignment method recorded on the first label entry. One of standard, privileged, or auto. Empty string if no label is assigned.
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.labels
+      description: All sensitivity labels assigned to the drive item, as returned by Microsoft Graph.
+      type: Unknown
+  - arguments:
+    - auto: PREDEFINED
+      description: The MS Graph resource.
+      name: object_type
```
noydavidi, line 822:
```suggestion
      name: resource_type
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+      - sites
+      - users
+      required: true
+    - description: |-
+        MS Graph resource ID.
+        For resource type 'drive': To get a list of all drives in your site, use the msgraph-list-drives-in-site command.
+        For resource type 'group': To get a list of all groups that exists, configure the 'Entra ID Groups' integration and use the msgraph-groups-list-groups command.
+        For resource type 'sites': To get a list of all sites, use the msgraph-list-sharepoint-sites command.
+        For resource type 'users': To get a list of all users that exists, configure the 'Entra ID Users' integration and use the msgraph-user-list command.
+      name: object_type_id
```
noydavidi, line 835:
```suggestion
      name: resource_id
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+      required: true
+    - description: The ID of the drive item to assign the sensitivity label to.
+      name: item_id
+      required: true
+    - description: The GUID of the sensitivity label to assign. Retrieve label GUIDs from the Microsoft Purview compliance portal or via the PowerShell `Get-Label` cmdlet.
+      name: sensitivity_label_id
```
noydavidi, line 841: I don't think it should be required from Cortex side. As mentions in MS docs: 
`ID of the sensitivity label to be assigned, or empty string to remove the sensitivity label.`
So in our case we don't have to make this required so in case a customer wants to remove the label they can keep this arg empty and we will pop it by default empty string. Let's also add it to the arg description.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
   - description: Generate the login URL used for Authorization code flow.
     name: msgraph-files-generate-login-url
     arguments: []
+  - arguments:
+    - auto: PREDEFINED
+      description: The MS Graph resource.
+      name: object_type
```
noydavidi, line 786:
```suggestion
      name: resource_type
```
I think this name may be better, tell me what you think.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+      - sites
+      - users
+      required: true
+    - description: |-
+        MS Graph resource ID.
+        For resource type 'drive': To get a list of all drives in your site, use the msgraph-list-drives-in-site command.
+        For resource type 'group': To get a list of all groups that exists, configure the 'Entra ID Groups' integration and use the msgraph-groups-list-groups command.
+        For resource type 'sites': To get a list of all sites, use the msgraph-list-sharepoint-sites command.
+        For resource type 'users': To get a list of all users that exists, configure the 'Entra ID Users' integration and use the msgraph-user-list command.
+      name: object_type_id
```
noydavidi, line 799:
```suggestion
      name: resource_id
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.sensitivityLabelId
+      description: The currently assigned sensitivity label GUID, taken from the first entry of the labels array. Empty string if no label is assigned.
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.assignmentMethod
+      description: The assignment method recorded on the first label entry. One of standard, privileged, or auto. Empty string if no label is assigned.
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.labels
+      description: All sensitivity labels assigned to the drive item, as returned by Microsoft Graph.
+      type: Unknown
```
noydavidi: Let's stick to the outputs name as written in the docs 
https://learn.microsoft.com/en-us/graph/api/driveitem-extractsensitivitylabels?view=graph-rest-1.0&tabs=http

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+        auto: recorded as a system-driven assignment.
+      name: assignment_method
+      predefined:
+      - standard
+      - privileged
+      - auto
+      defaultValue: standard
+    - description: Free-text justification recorded with the assignment. Required by Microsoft Graph when downgrading or replacing a user-assigned label.
+      name: justification_text
+      defaultValue: Applied via Cortex XSOAR
```
noydavidi: No need for default values for non required args, you can remove. Also from the python code.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+        Assignment method recorded on Microsoft Graph.
+        standard: a user-driven assignment.
+        privileged: overrides existing user-applied labels.
+        auto: recorded as a system-driven assignment.
+      name: assignment_method
+      predefined:
+      - standard
+      - privileged
+      - auto
+      defaultValue: standard
```
noydavidi: No need for default values for non required args, you can remove. Also from the python code.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+    - description: Free-text justification recorded with the assignment. Required by Microsoft Graph when downgrading or replacing a user-assigned label.
+      name: justification_text
+      defaultValue: Applied via Cortex XSOAR
+    description: Assigns a sensitivity label to a drive item. On success, returns SUCCESS, the HTTP status code returned by Microsoft Graph, and the Location header URL for polling the long-running operation. HTTP error responses from Microsoft Graph are surfaced verbatim as a command error.
+    name: msgraph-apply-sensitivity-label
```
noydavidi:
```suggestion
    name: msgraph-assign-sensitivity-label
```

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,99 @@ script:
...
+      type: String
+    - contextPath: MsGraphFiles.AppliedSensitivityLabel.result
+      description: SUCCESS on a 2xx response from Microsoft Graph.
+      type: String
+    - contextPath: MsGraphFiles.AppliedSensitivityLabel.httpStatusCode
+      description: The HTTP status code returned by Microsoft Graph.
+      type: Number
+    - contextPath: MsGraphFiles.AppliedSensitivityLabel.operationLocation
+      description: URL returned in the Location header. Microsoft Graph treats assignSensitivityLabel as a long-running operation; poll this URL to monitor its completion status.
+      type: String
```
noydavidi, line 867: Same as here, lets stick with the original outputs as written in the docs
https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-1.0&tabs=http

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1124,119 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+    return CommandResults(
+        outputs_prefix="MsGraphFiles.ExtractedSensitivityLabel",
+        outputs_key_field="itemId",
+        outputs=outputs,
+        readable_output=readable_output,
+        raw_response=raw_response,
+    )
+
+
+def apply_sensitivity_label_command(client: MsGraphClient, args: dict[str, str]) -> CommandResults:
```
noydavidi:
```suggestion
def assign_sensitivity_label_command(client: MsGraphClient, args: dict[str, str]) -> CommandResults:
```

noydavidi (changes requested):
Nice Job! Added some more comments :) 
Please check the commands once you are done to check that everything is working as expected :) 
Let me know once you are done.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
ymishrapalo1992: This is what has been used in the graph api , I see conventions of object_id/object_type in the existing project

ymishrapalo1992:
Addressed the comments, apart object_type_id and object_id one, it follows the existing content pack conventions for microsoft

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py
```diff
@@ -1055,6 +1128,118 @@ def delete_site_permission_command(client: MsGraphClient, args: dict[str, str])
...
+
+    Returns:
+        CommandResults with the drive item ID, the assigned label GUID, and the
+        `Location` header URL returned by Microsoft Graph.
+    """
+    object_type = args.get("object_type", "")
+    object_type_id = args.get("object_type_id", "")
+    item_id = args.get("item_id", "")
+    sensitivity_label_id = args.get("sensitivity_label_id", "")
+    assignment_method = args.get("assignment_method", "") or "standard"
```
noydavidi:
```suggestion
    assignment_method = args.get("assignment_method", "")
```
We don't want to set a default when the arg is not required from MS side.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
noydavidi: Please remove

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
noydavidi: Still not aligned with the right outputs.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
```diff
@@ -780,6 +780,92 @@ script:
...
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.id
+      description: The GUID of the sensitivity label assigned to the drive item, or empty string when no label is assigned.
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.displayName
+      description: The human-readable display name of the assigned sensitivity label, or empty string when no label is assigned.
+      type: String
+    - contextPath: MsGraphFiles.ExtractedSensitivityLabel.protectionEnabled
+      description: True if the assigned label has encryption/protection settings; false for classification-only labels. Defaults to false when no label is assigned.
+      type: Boolean
```
noydavidi, line 818: The outputs are not aligned with MS outputs https://learn.microsoft.com/en-us/graph/api/driveitem-extractsensitivitylabels?view=graph-rest-1.0&tabs=http#:~:text=items/016GVDAP3RCQS5VBQHORFIVU2ZMOSBL25U/extractSensitivityLabels-,Response,-The%20following%20example

noydavidi (changes requested):
Great job! Hopefully last comments :) :

1. Make sure all context outputs are aligned with Microsoft API outputs.
2. Make sure you don't set any default value for any unrequited argument. We want to keep it empty and send it empty to Microsoft API. 

Once all changes are applied please make sure the commands are working and send me a quick demo video of the commands running and their outputs. Thanks.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
ymishrapalo1992: The AppliedSensitivityLabel.operationLocation field name no longer exists in the codebase — it was renamed earlier in this PR. The current outputs for msgraph-assign-sensitivity-label are:

- contextPath: MsGraphFiles.AssignedSensitivityLabel.itemId
  description: The ID of the drive item the label was assigned to.
  type: String
- contextPath: MsGraphFiles.AssignedSensitivityLabel.sensitivityLabelId
  description: The GUID of the sensitivity label that was assigned. Empty string indicates the existing label was removed.
  type: String
- contextPath: MsGraphFiles.AssignedSensitivityLabel.location
  description: URL returned in the Microsoft Graph Location response header. Microsoft Graph treats assignSensitivityLabel as a long-running operation; poll this URL to track the operation's completion status.
  type: String

These align with the [assignSensitivityLabel docs](https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-1.0&tabs=http) you linked:

location ← surfaced verbatim from the Location response header.
itemId and sensitivityLabelId ← echoes of the input args. The 202 Accepted response has no body (only the Location header), so without these echoes the output context would not be joinable back to which file/label the call targeted. Echoing input resource IDs into the output is the standard XSOAR convention used by every mutating command in this pack (and across the platform) — e.g. msgraph-create-new-folder echoes the parent ID, msgraph-replace-existing-file echoes the item ID, etc.
Could you confirm you're viewing the latest commit on this branch? If the diff still shows AppliedSensitivityLabel.operationLocation, the review view may be cached against an earlier commit.

Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.yml
ymishrapalo1992: The docs link in your comment (extractSensitivityLabels action endpoint) describes the endpoint we previously called — that's not what the current code does. The implementation was switched during this PR after live testing exposed an issue.

Current code ([MicrosoftGraphFiles.py:559](vscode-webview://0oo3fcc35mrl9fmnfkfeo5e53smgjmj5jb0qjduu71dsf10175ml/Packs/MicrosoftGraphFiles/Integrations/MicrosoftGraphFiles/MicrosoftGraphFiles.py:559)) calls:

GET /v1.0/{drive-prefix}/items/{item-id}?$select=sensitivityLabel

Docs for this endpoint:

GET DriveItem operation: https://learn.microsoft.com/en-us/graph/api/driveitem-get?view=graph-rest-1.0&tabs=http
sensitivityLabel property on the driveItem resource: https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0 (the sensitivityLabel property's value type carries id, displayName, and protectionEnabled)
OData $select parameter: https://learn.microsoft.com/en-us/graph/query-parameters?context=graph%2Fapi%2F1.0&view=graph-rest-1.0#select-parameter
The response shape from this endpoint is:

{
  "sensitivityLabel": {
    "id": "<guid>",
    "displayName": "<name>",
    "protectionEnabled": false
  }
}

Our outputs MsGraphFiles.ExtractedSensitivityLabel.{id, displayName, protectionEnabled} mirror this verbatim. Plus itemId as the echo field for context-path-joining (XSOAR convention).

Why the endpoint was changed:
Live Playground testing on the aperturesync tenant proved the original extractSensitivityLabels action endpoint returns empty labels[] for files with classification-only labels (those with protectionEnabled: false). Per the [extractSensitivityLabels docs](https://learn.microsoft.com/en-us/graph/api/driveitem-extractsensitivitylabels?view=graph-rest-1.0&tabs=http), that action extracts contents from encrypted labels — it's a content-extraction operation for forensic use cases, not a general classification lookup. For files without encryption protection (the majority of real-world labels), it correctly but uselessly returns empty.

Reproducible sequence on aperturesync:

Assigned label 08973045-... via msgraph-assign-sensitivity-label → 202 Accepted with Location header. ✅
Manual call: GET /v1.0/drives/{driveId}/items/{itemId}?$select=sensitivityLabel → returned {sensitivityLabel: {id: "08973045-...", displayName: "This label is created to test the MRB Activity", protectionEnabled: false}} ✅
Manual call: POST /v1.0/drives/{driveId}/items/{itemId}/extractSensitivityLabels on the same file → returned {value: {labels: []}} ❌ (silently empty, would hide the assignment from the caller)
The $select endpoint we now use is:

GA / v1.0 (not beta).
Non-metered (no HTTP 402 on tenants without Graph billing enrolment).
Returns the file's classification label regardless of protectionEnabled state.
Single object response, matching Microsoft's data model (each driveItem has at most one sensitivity label — there is no multi-label scenario for primary file classification).
Same Files.Read.All scope as the action endpoint

noydavidi (approved):
Great!

PR #39758, opened 2025-04-28
CRTX-168369: remove platform

masulin97:
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CRTX-168369)

## Description
A few sentences describing the overall goals of the pull request's commits.

Force merge required since we changed  the marketplaces.

## Must have
- [ ] Tests
- [ ] Documentation

PR #44678, opened 2026-06-16
fix code

tcarmeli1:


yedidyacohenpalo (approved):

tcarmeli1:
Require force merge since BA130 fail on:
Packs/CortexAttackSurfaceManagement: [BA130] - The content item has 'marketplacev2' but is missing 'platform' in its marketplaces and/or 'xsiam' in its supported modules (these can be set directly, inherited from the pack, or resolved from defaults).
but the platform is missing on purpose: https://github.com/demisto/content/pull/39758

jbabazadeh (approved):

PR #44825, opened 2026-06-24
fix(AbnormalSecurity): self-healing per-stream fetch checkpoints [INT-1959]

krdivesh:
## Summary

Fixes INT-1959: a single transient backend error (HTTP 500 / connection reset / timeout) mid `fetch-incidents` previously propagated out, so `setLastRun` was never called, `last_fetch` never advanced, and the identical window was retried and failed every cycle — wedging the fetch until a manual reset.

## Changes

- Replaced the single `last_fetch` scalar with three independent per-stream watermarks in `last_run` (`threats_last_fetch` / `campaigns_last_fetch` / `cases_last_fetch`); a missing per-stream key falls back to legacy `last_fetch`, then `first_fetch_time`, for backward compatibility with already-deployed instances.
- Each stream now computes its own start window, sorts its items ascending by the filter timestamp (`latestTimeRemediated` / `lastReportedTime` / `lastModifiedTime`, missing values treated as oldest), and advances its watermark only to the last successfully processed item.
- A non-skippable error (5xx / connection reset / 401/403/429) now stops only the affected stream — without raising and without advancing past the failed item — so `fetch_incidents` still returns and `main()` persists progress for all streams. Existing skippable-4xx skip-and-continue behavior is unchanged.
- Added `ReleaseNotes/2_4_7.md` and bumped `pack_metadata.json` to `2.4.7`.

## Test Plan

Unit tests run under the demisto-sdk venv (CI timezone is UTC):

```
$ TZ=UTC PYTHONPATH="Tests/demistomock:Packs/Base/Scripts/CommonServerPython:Packs/ApiModules/Scripts/DemistoClassApiModule:<integration dir>" \
    python -m pytest Packs/AbnormalSecurity/Integrations/AbnormalSecurity/AbnormalSecurity_test.py -q

72 passed, 1 warning in 0.17s
```

New focused tests cover:
- A non-skippable 500 on the 2nd of three threats: `fetch_incidents` returns (does not raise), emits the first threat, and `threats_last_fetch` advances to the last good threat but not past the failed one, while campaigns/cases still get watermarks.
- Backward compat: a `last_run` with only legacy `last_fetch` seeds the start for all streams.
- Ascending-sort safety: an out-of-order input where the newest item fails does not let the watermark skip the older unprocessed item.
- Clean run: all incidents emitted and all three watermarks advanced (no regression).

## Known limitation / follow-up

`setLastRun` only persists on a graceful return. If XSOAR hard-kills the Docker container (SIGKILL on the 300s wall-clock timeout) mid-fetch, nothing returns and progress within that run is still lost. Surviving a hard kill would require incremental `demisto.setIntegrationContext()` checkpointing during processing — tracked as follow-up **INT-1960**, out of scope for this PR. This PR fixes the dominant documented failure mode (the 500/timeout exception that propagates and wedges the window).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-17156

### DIFF
--- a/Packs/AbnormalSecurity/Integrations/AbnormalSecurity/AbnormalSecurity.py
+++ b/Packs/AbnormalSecurity/Integrations/AbnormalSecurity/AbnormalSecurity.py
@@ -65,6 +65,56 @@ def get_current_datetime() -> datetime:
     return datetime.utcnow().astimezone(timezone.utc)
 
 
+# Per-stream watermark keys persisted in last_run (see fetch_incidents).
+THREATS_LAST_FETCH = "threats_last_fetch"
+CAMPAIGNS_LAST_FETCH = "campaigns_last_fetch"
+CASES_LAST_FETCH = "cases_last_fetch"
+
+
+def _resolve_stream_watermark(last_run: dict[str, Any], stream_key: str, first_fetch_time: str) -> datetime:
+    """Resolve a stream's start watermark from last_run.
+
+    Precedence: the per-stream key, then the legacy ``last_fetch`` scalar (for instances
+    deployed before per-stream checkpoints existed), then ``first_fetch_time``.
+
+    Args:
+        last_run: The integration's last_run dict.
+        stream_key: One of THREATS_LAST_FETCH / CAMPAIGNS_LAST_FETCH / CASES_LAST_FETCH.
+        first_fetch_time: ISO-8601 fallback used on the very first fetch.
+
+    Returns:
+        The watermark as a tz-aware UTC datetime.
+    """
+    raw = last_run.get(stream_key) or last_run.get("last_fetch") or first_fetch_time
+    return _parse_iso_utc(raw)
+
+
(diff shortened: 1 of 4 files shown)


krdivesh:
It's still in draft. Please don't proceed with review for the time being :)

kamalq97:
Hi @krdivesh

Thank you once again for your contribution,

One additional question, is this PR related / dependent on the changes in https://github.com/demisto/content/pull/44853?

Should the changes be combined into one branch / PR?

kamalq97:
Hi @krdivesh 

Thank you once again for your contributions.

Due to the lack of response, we have decided to close this PR as well as https://github.com/demisto/content/pull/44853.

Please consider submitting a new contribution when ready ideally merging the changes from both PRs into a single branch before opening the PR.

In case you have additional questions, please do not hesitate to contact me via the DFIR Community Slack,
Many thanks,

PR #44971, opened 2026-07-07
XSUP-72364 - fix microsoft_windows_raw event_type task mapping

eepstain:
## Related Issues
https://jira-dc.paloaltonetworks.com/browse/XSUP-72364

## Description
Fixed an issue where `xdm.event.type` for the `microsoft_windows_raw` dataset could be set from an unreliable `task` (TaskCategory) value produced by the XDR Collector / Winlogbeat profile, which caused incorrect event types (for example a `4624` logon event showing as *Process Creation* or *Detailed File Share*).

Investigation on a live tenant showed the raw `task` column is wrong for a large share of Security-Auditing events (measured ~18% of `4624`, ~17% of `4634`, ~6% of `4688`), with the same generic task buckets spread across 17+ distinct event IDs. The underlying event (event_id, message, provider) is correct; only the `task` label is corrupted.

The modeling rule previously derived `event_type` via `coalesce(event_action, to_string(task), ...)`, so the corrupted `task` flowed straight into `xdm.event.type`. This change adds a `canonical_task` derived from `event_id` for standard `Microsoft-Windows-Security-*` events and prefers it in the `event_type` coalesce, so `xdm.event.type` is correct regardless of the corrupted raw value. Behavior for non-Security providers is unchanged (`canonical_task` is null and the previous coalesce order applies).

Pack `MicrosoftWindowsEvents` bumped 1.1.31 -> 1.1.32.

relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-17246


kamalq97:
Closing, no longe relevant

PR #45058, opened 2026-07-14
Revert "Auto RN: july-migrations-force"

tcarmeli1:
Reverts demisto/content#45011


israelpoli (approved):

israelpoli:
Force merge – We need to revert to the previous version to prevent this release from going out, so we can run a targeted upload for the pack to fix the bug for customers.

PR #45260, opened 2026-07-28
Xsup 73616 slackv3 incident entitlement regex

mayyagoldman:
## What & why

`SlackBlockBuilder` fails with `invalid_blocks_format` **only when run from a
Cortex Platform Case War Room**. The same command, with the same List, works in
the Playground and in Issue/Alert War Rooms. Reported in XSUP-73616.

### Root cause

`SlackBlockBuilder` builds an entitlement string `<GUID>@<investigation_id>` and
sends the Slack blocks **wrapped** together with that entitlement (plus reply /
expiry / default_response) into `SlackV3`'s `send-notification`.

`SlackV3` uses a single regex, `ENTITLEMENT_REGEX`, as its "is this an
entitlement?" test. Only when it matches does `SlackV3`:
1. **unwrap** the payload to pull out the real `blocks` list (send side), and
2. **route the user's reply** back to the right incident/task (receive side).

The regex accepted only a GUID or a purely numeric id after the `@`. A Case War
Room investigation id looks like `INCIDENT-<n>` (e.g. `INCIDENT-12`), which
matches neither branch. So in a Case War Room the regex fails, the payload is
**not** unwrapped, and the raw wrapper dict is sent to Slack as `blocks` — Slack
rejects it with `{'ok': False, 'error': 'invalid_blocks_format'}`.

Because the same regex also gates the **receive** path, `INCIDENT-<n>` silently
broke SlackAsk reply routing too (a latent second bug the customer hadn't hit
yet, only because the send fails first).

## The change

One-line, tightly-scoped widening of `ENTITLEMENT_REGEX` to also accept an
optional literal `INCIDENT-` prefix on the id segment:

```diff
-ENTITLEMENT_REGEX = rf"{GUID_REGEX}@(({GUID_REGEX})|(?:[\d_]+))_*(\|\S+)?\b"
+ENTITLEMENT_REGEX = rf"{GUID_REGEX}@(({GUID_REGEX})|(?:INCIDENT-)?(?:[\d_]+))_*(\|\S+)?\b"
Fixing this single regex fixes both the send and receive paths, consistent
with the original design (the regex is the one source of truth for entitlement
detection, used in all 4 places: 2 send, 2 receive).
Deliberately scoped decisions:
- Only the literal INCIDENT- prefix is allowed, not arbitrary [A-Za-z]+-, to
keep the match tight (e.g. FOO-12 is still rejected).
- Numeric and GUID ids are unchanged; extract_entitlement already handles
INCIDENT-12 correctly (it just split("@")), so no downstream changes needed.
How it was validated
Ran a diagnostic in each war-room context on a lab XSIAM tenant, logging the real
investigation().id and whether ENTITLEMENT_REGEX matches:
Context
Alert War Room
War Room
Case War Room
This reproduced the exact reported behavior and pinned it to the INCIDENT-<n>
id format. With the fix, INCIDENT-12 matches and the blocks are unwrapped.
Tests
- Extended test_check_entitlement (the existing entitlement-routing test) with
INCIDENT-12 and INCIDENT-12|43 cases, asserting the reply routes with
incident id INCIDENT-12. This exercises the receive path through the shared
regex.
- Regex verified against numeric / GUID / underscore / INCIDENT-<n> (accepted)
and FOO-12, name@mail (rejected).
Release / versioning
- Packs/Slack/ReleaseNotes/3_8_12.md
- pack_metadata.json → 3.8.12
Review notes / call-outs
- Blast radius: a single regex constant shared by 4 call sites. No behavior
change for existing id formats — only adds coverage for INCIDENT-<n>.
- The fix commit is unsigned ("Unverified") because it was committed from a
sandbox without my signing key; can re-sign if required.
- Local demisto-sdk validate passed. The Docker-backed unit-test hook couldn't
run in my sandbox (no Docker Hub egress) — relying on CI for that gate.
Fixes XSUP-73616



<!-- REMINDER: THIS IS A PUBLIC REPO DO NOT POST HERE SECRETS/SENSITIVE DATA -->
## Contributing to Cortex XSOAR Content
Make sure to register your contribution by filling the [contribution registration form](https://forms.gle/XDfxU4E61ZwEESSMA)

**The Pull Request will be reviewed only after the contribution registration form is filled.**

## Status
- [x] In Progress
- [ ] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
<!-- To link a Jira ticket use fixes/related: link to the issue, for example fixes: CIAC-XXXX -->

## Description
<!-- A few sentences describing the overall goals of the pull request's commits. -->

## Must have
- [ ] Tests
- [ ] Documentation


marketplace-ai-reviewer:
🤖 Analysis started. Please wait for results...

marketplace-ai-reviewer:
<details>
<summary><strong>🤖 AI Review Disclaimer</strong></summary>

This review was generated by an AI-powered tool and may contain inaccuracies. Please be advised, and we extend our sincere apologies for any inconvenience this may cause.
</details>

marketplace-ai-reviewer (commented):
Thank you for your contribution! No issues found. Great work! 👍

 @omerKarkKatz, @mayyagoldman please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

mmhw (approved):
Good catch!

PR #45272, opened 2026-07-29
CSF fetch assets improvments

MLainer1:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: XSUP-71944

## Description
The XSIAM dataset `crowdstrike_falcon_spotlight_vulnerabilities_raw` never became queryable for
the customer because the assets snapshot never sealed. A snapshot only seals when the declared
`total-items-count` matches the number of rows actually stored in BigQuery. We confirmed that even
a fully completed run failed to seal: **declared 85,585,027 vs stored 85,335,179, a −249,848
delta**.

Separately, on this tenant the **assets** dataset was not merely unsealed but permanently empty —
a distinct root cause from the vulnerability drift, fixed below.

This PR makes the counting exact, stops three classes of transient fault from aborting a severity
mid-fetch, repairs asset collection on multi-CID tenants, and adds an opt-in long-running execution
mode for tenants whose full Spotlight fetch cannot complete inside the assets fetch interval.

---

## Fix 1 — Counted-but-dropped sends (dominant cause of the −249,848 delta)

**Problem.** The fetch incremented its running total from the number of records it handed to the
sender, not from the number the sender confirmed stored. A send that failed after the count was
taken inflated `total-items-count` with records that never reached the dataset.

**Fix.** Accounting moved to per-chunk and post-success. A batch is compressed into N gzip chunks,
each sent as its own HTTP request with its own retries, so each chunk is independently atomic.
`send_batch_to_xsiam_and_save_context(count_stored=True)` returns the number of records whose chunk
was acknowledged, and only that number is added to the total.

A failed send is **not** fatal to its severity. The severity continues with the remaining pages;
unsent records are simply not counted, so the declared total stays consistent with what was stored,
and those records are picked up on the next full fetch cycle.

Two supporting corrections in the same path:

- **A rejected batch is now an error, not a silent success.** `xsiam_api_call_async` previously
  logged an HTTP failure (for example *502 Bad Gateway*) but neither raised nor set `status_code`,
  so the awaiting caller saw a clean return and counted the batch. It now raises after exhausting
  its retries.
- **The empty seal chunk can no longer inflate the count.** `send_data_to_xsiam_async` reports a
  chunk size of 1 for the non-list "seal" chunk. Empty batches are skipped before a task is
  created, and `records_stored` is clamped with `min(records_stored, items_attempted)`, so a chunk
  can never report more than was attempted.

## Fix 2 — Oversized-page JSON truncation

**Problem.** On very large pages the API response was truncated mid-JSON, raising
`JSONDecodeError` and aborting the whole severity.

**Fix.** `fetch_spotlight_page_with_shrink` re-requests the *same* page (same after token) down a
shrink ladder `[5000, 2500, 1000, 500]`, pausing `[2, 5, 15]` seconds before each retry, and
reverts to the default size on the next page. Only when the smallest limit still fails does the
error propagate.

## Fix 3 — Transient upstream errors aborting a severity

**Problem.** A transient CrowdStrike service error that exhausted the API client's own retry
budget aborted the entire severity.

**Fix.** The same recovery flow handles retryable statuses
(`{408, 413, 425, 429, 500, 502, 503, 504}`). Non-transient errors — expired cursor, `401`, `400` —
still propagate immediately, since shrinking cannot help them.

## Fix 4 — Assets never collected on multi-CID tenants

**Problem.** On Flight Control / MSSP parent-child deployments, Spotlight returns the host
identifier as the CID and device ID joined together. The Devices API accepts only the bare device
ID, so **every** enrichment request was rejected with `400 invalid device id` and the assets
dataset stayed empty. Vulnerability collection was unaffected.

**Fix.** `extract_device_id_from_aid` strips the CID prefix before the request is sent, keyed on
CID length rather than splitting on the separator (the separator varies, and non-sensor asset IDs
contain their own `_`).

**Scope.** The prefix rule holds for **100%** of AIDs measured across the full table (134,050,354
rows, 2026-08-27): zero rows where `aid` is not prefixed by `cid`, zero where nothing would remain
after stripping, and `cid` is always 32 lowercase hex characters. Stripping is also lossless —
361,756 distinct AIDs reduce to 361,756 distinct device IDs, so no two hosts can merge.

Four AID shapes exist. `aid[len(cid)+1:]` handles all four without needing to know which is which,
which is why the implementation keys on CID length rather than splitting on the separator (the
separator varies, and the base64url bodies contain their own `_`):

| `LENGTH(aid)` | rows | shape |
|---:|---:|---|
| 65 | 133,934,940 | `<cid>-<32-hex>` — standard sensor AID |
| 89 | 114,276 | `<cid>_<56-char base64url>` |
| 79 | 1,136 | `<cid>_<46-char base64url>` |
| 67 | 2 | `<cid>-n:<32-hex>` — colon inside the body |

The two base64url shapes are likely container or cloud-workload asset IDs rather than sensor AIDs.
Whether the Devices API accepts their stripped form is **untested** — they may need to be filtered
rather than enriched. They account for 115,412 of 134,050,354 vulnerability rows, but the relevant
figure is how many of the 361,756 *distinct* assets they represent, which has not been measured.
See Known gaps.

## Fix 5 — Cancellation aborting a severity

`asyncio.CancelledError` has not derived from `Exception` since Python 3.8, so it escaped the
`except Exception` in the send-task reaper and killed the severity. The reaper now catches
`(asyncio.CancelledError, Exception)` and treats a cancelled send as an uncounted send, like any
other failed send.

---

## New: long-running execution mode

On large tenants a full Spotlight fetch can exceed the Assets Fetch Interval, so the fetch is
repeatedly interrupted and restarted and never reaches the seal. This adds an opt-in mode that runs
the fetch in a dedicated long-running container instead of on the assets fetch schedule.

- New parameter **`longRunning`** ("Long running instance for Spotlight vulnerabilities"). When
  enabled, the integration runs `long_running_spotlight_execution()` via the
  `long-running-execution` command.
- **Fixed 24-hour period between cycle starts.** Deliberately not configurable, so it can never be
  set below the time a full fetch needs. `sleep = max(0, period - elapsed)`, so a cycle that
  overruns simply starts the next one immediately and cycles never overlap.
- **A failed cycle retries quickly** rather than idling out the remaining period. Delay starts at
  5 minutes and doubles per consecutive failure to a 60-minute cap (5, 10, 20, 40, 60, 60…),
  resetting after any success. Previously a cycle that died in seconds — bad credentials, a tripped
  circuit breaker — waited a further ~24 hours before retrying. The retry delay is also `min()`'d
  with the remaining interval, so the failure path can only ever *shorten* the wait, never extend
  it.
- **Spotlight only.** This mode fetches Spotlight vulnerabilities regardless of the *Asset types to
  fetch* selection. CNAPP Alerts are not collected here and continue to be ingested only by the
  regular assets fetch.
- **Do not enable both.** Only one of `longRunning` or the assets fetch should be active on an
  instance. This is documented in the YAML `additionalinfo`, the README and the release note, but
  is **not currently enforced in code** — see Known gaps.
- A cycle that raises is logged and does not kill the loop.

## Observability

Snapshot-lifecycle lines are emitted under a greppable `[LIFECYCLE]` prefix — run start, the seal
decision with its inputs (`SEAL-GATE`, `SEAL-SENT`, `SEAL-SKIPPED`), and long-running cycle
boundaries. These are emitted once per run or per cycle, never per record. Per-page pagination
metadata (after token, page size, cumulative counts) is logged so any future drift can be traced to
the exact page and severity.

The seal line also reports a snapshot-level count of records XSIAM did not store, so a snapshot
that seals *smaller* than what Falcon returned is visible rather than silent.

---

## Tests

363 test functions in the file, all green. New coverage targets the failure modes rather than the
happy path:

- **Counting:** empty-batch skip, no-results counts zero, clamp of an over-reported stored count,
  partial-chunk counting, a cancelled send being non-fatal, and a 5xx from XSIAM exhausting retries
  and then raising.
- **Page recovery:** recovery after truncation, recovery from a bare `ContentClientError` 500 and
  from `ContentClientRetryError`, failure at the smallest limit, backoff between attempts, and
  proof that non-transient statuses are *not* retried.
- **Seal gate:** seals with real withheld records, skips when no records were fetched, does not
  seal when severities are incomplete, and reports the snapshot-level lost-record total.
- **Composite AID:** parametrized across every AID shape observed in production, plus bare AIDs,
  mismatched CIDs, and degenerate input.
- **Long-running loop (11 tests):** fixed 24-hour period, elapsed time subtracted, a cycle longer
  than the period sleeps zero, a failing cycle does not stop the loop, CNAPP is not fetched, and
  the new retry behaviour — quick retry, exponential doubling to the cap, reset after success, and
  the never-longer-than-schedule clamp.

The retry tests were **mutation tested**: with the backoff disabled, 3 of the 5 fail. The other 2
assert deliberately unchanged behaviour, so they are correctly insensitive.

`demisto-sdk pre-commit` is green in **default** mode across all hooks on both docker images —
ruff-format, ruff, mypy, pylint, xsoar-lint, MDX, validate, secrets, markdownlint and pytest.
Default mode matters here: nightly mode skips `ruff-format`, `xsoar-lint` and `secrets`, and
running only nightly previously hid a real formatting failure.

## Must have
- [ ] Tests
- [ ] Documentation


MLainer1:
@marketplace-ai-reviewer start review

marketplace-ai-reviewer:

marketplace-ai-reviewer (commented):
Thanks for submitting your PR!
Before moving on with the detailed review, please take a moment to address the following general comments.

Once you've improved your PR according to the comments below, re-trigger the review by adding the `ready-for-ai-review` label or commenting `@marketplace-ai-reviewer start review`.


- The PR introduces a new long-running execution mode (`longRunning` parameter and `long_running_spotlight_execution` command) which is not mentioned in the PR description. Please update the description to reflect the addition of this new feature.

MLainer1:
@marketplace-ai-reviewer start review

marketplace-ai-reviewer:

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4697,6 +4798,53 @@ async def fetch_vulnerabilities_by_severity(
...
+                        "error",
+                    )
+            except BaseException as e:  # noqa: BLE001 - includes asyncio.CancelledError
+                # Whole batch rejected: nothing is counted, so declared == stored still holds.
+                # BaseException, not Exception: a cancelled task raises asyncio.CancelledError, which
+                # since Python 3.8 does not derive from Exception and would kill the severity.
+                lost_send_batches += 1
+                lost_send_records += items_attempted
+                first_send_error = first_send_error or e
+                log_falcon_assets(
```
marketplace-ai-reviewer, line 4929: Exception blocks that report or log an error MUST include the full traceback.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4697,6 +4798,53 @@ async def fetch_vulnerabilities_by_severity(
...
+                    f"[{severity}] Background vulnerability send task failed; "
+                    f"{items_attempted} records NOT counted and skipped (continuing with the next pages): {e}",
+                    "error",
+                )
 
     try:
         while True:
```
marketplace-ai-reviewer, line 4937: Never use `while True` (or any unbounded `while`) anywhere in the code.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -5267,6 +5432,51 @@ def fetch_assets_command():
...
+    Cycles are strictly sequential and start every LONG_RUNNING_ASSETS_INTERVAL_MINUTES
+    (24 hours, not configurable). A cycle that overruns the period simply starts the next one
+    immediately, so cycles never overlap.
+    """
+    log_falcon_assets(
+        f"Starting long-running Spotlight fetch loop (interval: {LONG_RUNNING_ASSETS_INTERVAL_MINUTES} minutes).",
+        "info",
+    )
+
+    while True:
```
marketplace-ai-reviewer, line 5577: Never use `while True` (or any unbounded `while`) anywhere in the code.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -5267,6 +5432,51 @@ def fetch_assets_command():
...
+        cycle_start = time.monotonic()
+        try:
+            asyncio.run(fetch_spotlight_assets())
+        except Exception as e:  # noqa: BLE001 - a single bad cycle must not kill the container
+            demisto.error(f"Long-running Spotlight fetch cycle failed: {e}")
```
marketplace-ai-reviewer: Exception blocks that report or log an error MUST include the full traceback.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
```diff
@@ -412,7 +412,23 @@ configuration:
   - xpanse
   supportedModules:
   - "xsiam"
-  - "exposure_management"
+  - "exposure_management"
```
marketplace-ai-reviewer: Remove trailing whitespace.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
```diff
@@ -412,7 +412,23 @@ configuration:
   - xpanse
   supportedModules:
   - "xsiam"
-  - "exposure_management"
+  - "exposure_management"  
+- additionalinfo: |-
+    Runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule.
```
marketplace-ai-reviewer: The `additionalinfo` for a boolean parameter must start with "Whether...".

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
```diff
@@ -412,7 +412,23 @@ configuration:
...
+    A fetch cycle starts every 24 hours. If a cycle runs longer than 24 hours, the next one starts as soon as it finishes.
+  display: Long running instance for Spotlight vulnerabilities
+  name: longRunning
+  type: 8
+  section: Collect
+  advanced: true
+  required: false
+  hidden:
+  - xsoar
+  - xpanse
```
marketplace-ai-reviewer, line 433: Consider adding `supportedModules` to match the `hidden` field and other asset fetch parameters.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/README.md
```diff
@@ -40,6 +40,7 @@ The CrowdStrike Falcon OAuth 2 API (formerly the Falcon Firehose API), enables f
...
 | Advanced: Time in minutes to look back when fetching events and detections | Use this parameter to determine the look-back period for searching for events that were created before the last run time and did not match the query when they were created. | False |
 | Assets Fetch Interval | The fetch interval for assets and vulnerabilities. It is recommended to set it to 1 hour. | False |
+| Long running instance for Spotlight vulnerabilities | Runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule. Use this when a full Spotlight fetch takes longer than the Assets Fetch Interval, so it is repeatedly interrupted. When enabled, do not enable the assets fetch as well - only one of the two should be active. This mode fetches Spotlight vulnerabilities ONLY, regardless of the Asset types to fetch selection. CNAPP Alerts are not collected in this mode; keep them on the regular assets fetch. A fetch cycle starts every 24 hours. If a cycle runs longer than 24 hours, the next one starts as soon as it finishes. | False |
```
marketplace-ai-reviewer: Consider refining the description for better clarity and to explicitly mention that separate instances are needed if both Spotlight and CNAPP alerts are required.

```suggestion
| Long running instance for Spotlight vulnerabilities | Runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule. Use this when a full Spotlight fetch takes longer than the Assets Fetch Interval and is repeatedly interrupted. When enabled, do not enable the assets fetch as well; only one of the two should be active per instance. This mode fetches Spotlight vulnerabilities ONLY, regardless of the **Asset types to fetch** selection. CNAPP Alerts are not collected in this mode; to collect them, configure a separate integration instance using the regular assets fetch. A fetch cycle starts every 24 hours. If a cycle runs longer than 24 hours, the next one starts as soon as it finishes. | False |
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_12_6.md
```diff
@@ -0,0 +1,11 @@
+
+#### Integrations
+
+##### CrowdStrike Falcon
+
+- Fixed an issue where the Spotlight vulnerabilities assets snapshot never sealed because records were counted before the asynchronous send to XSIAM, so failed sends were counted-but-dropped (declared > stored). Records are now counted only after XSIAM confirms storing them, counting each batch by the number of records actually stored. A failed send no longer aborts the severity: the fetch continues with the remaining pages and the records that were not stored are simply not counted, and are picked up on the next fetch cycle.
```
marketplace-ai-reviewer: Product names must include the "Cortex" prefix. Please change "XSIAM" to "Cortex XSIAM".

```suggestion
- Fixed an issue where the Spotlight vulnerabilities assets snapshot never sealed because records were counted before the asynchronous send to Cortex XSIAM, so failed sends were counted-but-dropped (declared > stored). Records are now counted only after Cortex XSIAM confirms storing them, counting each batch by the number of records actually stored. A failed send no longer aborts the severity: the fetch continues with the remaining pages and the records that were not stored are simply not counted, and are picked up on the next fetch cycle.
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_12_6.md
```diff
@@ -0,0 +1,11 @@
...
+- Fixed an issue where a page with no records left to send could still be sent to XSIAM and counted as one record that did not exist, causing the snapshot to declare more records than were stored.
```
marketplace-ai-reviewer: Product names must include the "Cortex" prefix. Please change "XSIAM" to "Cortex XSIAM".

```suggestion
- Fixed an issue where a page with no records left to send could still be sent to Cortex XSIAM and counted as one record that did not exist, causing the snapshot to declare more records than were stored.
```

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
Hi! Thanks for your contribution to the CrowdStrike Falcon pack. I've reviewed the changes and have a few notes to help polish things up:

* **Version Bump:** Since a new feature was added, please update the release notes to reflect a minor version bump (2.13.0) instead of a patch.
* **Python Code:** Avoid using unbounded `while True` loops and ensure all exception blocks include full tracebacks.
* **Docs & Config:** Ensure product names use the "Cortex" prefix (e.g., Cortex XSIAM) in the release notes, and address a few minor YAML formatting and README clarity tweaks.

Thanks again for your hard work on this!

- `Packs/CrowdStrikeFalcon/ReleaseNotes/2_12_6.md`: The version bump type does not match the nature of the changes. This release includes a new feature addition (`Added the *Long running instance for Spotlight vulnerabilities* parameter`), which requires a **minor** version bump, but the filename indicates a patch bump (`2.12.6`).

Please bump the version to `2.13.0` and rename this file to `2_13_0.md`. You can do this by running `demisto-sdk update-release-notes -i Packs/CrowdStrikeFalcon -u minor`. For more information, refer to the [Release Notes Documentation](https://xsoar.pan.dev/docs/documentation/release-notes#general).

 @marketplace-ai-reviewer, @github-code-quality[bot], @MLainer1 please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
MLainer1: No relevant for here, this is a long running command

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
MLainer1: No relevant for here, this is a long running command

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -840,6 +854,15 @@ def modify_detection_summaries_outputs(detection: dict):
     return detection
 
 
+def log_spotlight_verify(log_line: str, log_type: str = "info"):
```
Shellyber: Just a comment so we won't forget to remove this

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -5005,6 +5250,11 @@ async def finalize_severity_fetch(
                 data_type="assets",
```
nbensalm-palo, line 5291: Shouldn't we add here `count_stored=True` to close the single remaining gap where declared could diverge from stored?

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4362,11 +4399,22 @@ async def send_batch_to_xsiam_and_save_context(
...
                             Example: save_spotlight_state, save_cnapp_state, etc.
         data_type: Type of data being sent for XSIAM collector-type header. Defaults to "assets"
+        count_stored: When True, report exactly how many records XSIAM stored instead of failing the
+            whole batch on the first bad chunk. A batch is sent as independent chunks, each its own
+            request with its own retries, so a chunk either stores completely or not at all. The
+            default ``asyncio.gather(*tasks)`` raises on the first failing chunk and discards every
+            chunk's count, making a partially stored batch indistinguishable from a fully failed
+            one. With this flag the chunks are awaited with ``return_exceptions=True`` and only the
+            successful chunks' records are counted, so the caller can declare exactly what was
+            stored (``declared == stored``). A partial failure is logged and does not raise.
```
nbensalm-palo: This comment is very informative but its too long for function argument I think need to be shorten

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4663,6 +4732,95 @@ async def wait_for_background_tasks(pending_tasks: set[asyncio.Task], task_descr
...
+        "error",
+    )
+    log_falcon_assets(
+        f"[{severity}] Failed to fetch Spotlight page after exhausting the shrink ladder down to "
+        f"limit={SPOTLIGHT_PAGE_SIZE_SHRINK_LADDER[-1]}. Last error: {last_error}",
+        "error",
+    )
+    # The ladder is non-empty, so last_error is always set once the loop exits without returning.
+    assert last_error is not None
+    raise last_error
```
nbensalm-palo: Why does an `assert` is needed here?
It looks kind off in the code
Maybe something like:

```
if last_error:
    raise last_error
else:
    raise ContentClientError("fetch_spotlight_page_with_shrink failed")
```

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -93,8 +94,21 @@
 MAX_FETCH_SPOTLIGHT_ASSETS = 5000
```
nbensalm-palo, line 94: There is already a lot of constant, maybe adding here:

```
# region Spotlight Fetch Assets Constants
...
# endregion
```

nbensalm-palo (commented):
As usual looks great.

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
+
+#### Integrations
+
+##### CrowdStrike Falcon
+
+- Fixed an issue where the Spotlight vulnerabilities assets snapshot never sealed because records were counted before the asynchronous send to Cortex XSIAM, so failed sends were counted-but-dropped (declared > stored). Records are now counted only after Cortex XSIAM confirms storing them, counting each batch by the number of records actually stored. A failed send no longer aborts the severity: the fetch continues with the remaining pages and the records that were not stored are simply not counted, and are picked up on the next fetch cycle.
```
Shellyber: Regarding the release notes, I would rather we write something general about the number of records we send, rather than specifying exactly the causes and what was wrong. 
There were several reasons for that, lets summerize just the main problem that we've sent not the right amount, where there was a large scale,

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+- Fixed an issue where a transient CrowdStrike service error that exhausted the API client's own retries aborted an entire severity fetch. The same page is now retried using the same recovery flow, so a temporary upstream outage no longer prevents the snapshot from sealing.
+- Added the *Long running instance for Spotlight vulnerabilities* parameter, which runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule. Use it when a full Spotlight fetch takes longer than the assets fetch interval and is therefore repeatedly interrupted. A fetch cycle starts every 24 hours, and each cycle runs to completion before the next one starts, so a fetch may take any amount of time, including more than 24 hours. When this option is enabled, do not also enable the assets fetch on the same instance. This mode fetches Spotlight vulnerabilities only; to collect CNAPP Alerts, configure a separate integration instance using the regular assets fetch.
```
Shellyber: `When this option is enabled, do not also enable the assets fetch on the same instance.` - we should prevent this in the code if we didn't. Fetch assets or long running

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
```diff
@@ -413,6 +413,25 @@ configuration:
...
+  supportedModules:
```
Shellyber, line 434: What about Agentix?

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -840,6 +854,15 @@ def modify_detection_summaries_outputs(detection: dict):
     return detection
 
 
+def log_spotlight_verify(log_line: str, log_type: str = "info"):
+    """Emit a XSUP-71944 fix-verification line, greppable as ``[VERIFY]``.
```
talzich: Do we really need this function in prod? It served its purpose in the investigation phase, but I think we can lose it now

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4199,6 +4224,17 @@ async def xsiam_api_call_async(
         if status_code == 429:
             await asyncio.sleep(1)
         attempt_num += 1
+
+    if status_code != 200:
+        # XSUP-71944: this used to log the failure and return normally. The caller's
+        # asyncio.gather() then saw no exception and counted the batch as stored, pushing the
+        # declared total-items-count above the rows actually stored, so the snapshot never sealed.
+        # Raising keeps "counted" tied to "confirmed stored".
```
talzich:
```suggestion
        # Raising is necessary to keep "counted" tied to "confirmed stored". Failing gracefully will cause seal mismatch.
```

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4340,7 +4376,8 @@ async def send_batch_to_xsiam_and_save_context(
     state: ContentClientState,
     save_state_callback: Callable[[ContentClientContextStore, ContentClientState], None],
     data_type: str = "assets",
-) -> int:
+    count_stored: bool = False,
```
talzich, line 4374: Let's talk about this because the comment about this arg seems to suggest having it False is bug prone. I might be missing something, so lets discuss.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4663,13 +4732,105 @@ async def wait_for_background_tasks(pending_tasks: set[asyncio.Task], task_descr
...
+    Returns:
+        Tuple of (vulnerabilities_list, response_data) for the successfully fetched page.
+
+    Raises:
+        json.JSONDecodeError: If every page size in the ladder is still truncated.
+        ContentClientError: If a transient upstream fault persists across the whole ladder, or
+            immediately for any non-transient error that shrinking cannot resolve.
+    """
+    last_error: Exception | None = None
+    last_rung_index = len(SPOTLIGHT_PAGE_SIZE_SHRINK_LADDER) - 1
```
talzich:
```suggestion
    last_step_index = len(SPOTLIGHT_PAGE_SIZE_SHRINK_LADDER) - 1
```

talzich: Expanded my vocabulary today! Might be a bit obscure for most of us, though. Mind using this simpler alternative?

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4663,13 +4732,105 @@ async def wait_for_background_tasks(pending_tasks: set[asyncio.Task], task_descr
...
+
+    for rung_index, attempt_limit in enumerate(SPOTLIGHT_PAGE_SIZE_SHRINK_LADDER):
+        try:
+            page = await fetch_spotlight_vulnerabilities_page(
+                client=client, after_token=after_token, filter_query=filter_query, limit=attempt_limit
+            )
+            if rung_index:
+                # Only reachable when an earlier rung failed, so this is a genuine recovery.
+                log_spotlight_verify(
+                    f"FIX-2 SHRINK-RECOVERED [{severity}] succeeded at limit={attempt_limit} "
```
talzich: FIX-2? Makes sense in the context of solving the bug, but not in the context of production.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4826,6 +5047,34 @@ def update_last_saved(future, _pending=pending_tasks):
...
+        # Drain remaining send tasks so counting settles before the total is returned, guaranteeing
+        # total_fetched reflects only records XSIAM actually stored.
+        if pending_tasks:
+            log_falcon_assets(f"[{severity}] Draining {len(pending_tasks)} send tasks before finalizing count", "info")
+            await asyncio.gather(*pending_tasks, return_exceptions=True)
+            reap_completed_send_tasks(pending_tasks)
+            pending_tasks.clear()
+
+        log_spotlight_verify(
+            f"FIX-1 SEVERITY-DONE [{severity}] counted_total={total_fetched} "
```
talzich: Same

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
```diff
@@ -413,6 +413,25 @@ configuration:
...
+  - "xsiam"
+  - "exposure_management"
```
talzich, line 436: Do we need to add a default here?

talzich (changes requested):
B-E-A-utiful! Nice engineering with the progressively shrinking page size, you🪨

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4749,20 +4947,22 @@ async def fetch_vulnerabilities_by_severity(
             extract_unique_aids(vulnerabilities, unique_aids)
 
             # Send AIDs to asset handler for enrichment (async fire-and-forget)
-            batch_aids = {vuln.get("aid") for vuln in vulnerabilities if vuln.get("aid")}
+            batch_aids = {
+                device_id
+                for vuln in vulnerabilities
+                if (device_id := extract_device_id_from_aid(vuln.get("aid"), vuln.get("cid")))
+            }
```
nbensalm-palo: Lines 4950-49545 is duplicate to lines 4570-4572.
Maybe extract to function.

```
def device_ids_from_vulns(vulnerabilities: list) -> set[str]:
    """Return the set of bare device IDs for a batch, stripping composite CID prefixes."""
    return {
        device_id
        for vuln in vulnerabilities
        if (device_id := extract_device_id_from_aid(vuln.get("aid"), vuln.get("cid")))
    }
```

nbensalm-palo (approved):
Again, a big acknowledgement for your hard work.

Just a small thing I think can be added, except from that looking good.

talzich (approved):
🪨 ⭐

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
+
+#### Integrations
+
+##### CrowdStrike Falcon
+
+- Fixed an issue where Spotlight assets were never collected on tenants that span multiple CIDs (CrowdStrike Flight Control / MSSP parent-child deployments). On such tenants Spotlight returns the CID and the device ID joined together, which the CrowdStrike Devices API rejects. The device ID is now extracted before the request is sent. Vulnerability collection was not affected.
+- Fixed an issue where the Spotlight vulnerabilities assets snapshot never sealed because the declared record count did not match what was actually stored in Cortex XSIAM. Records were counted before being sent, an HTTP error returned while sending a batch (for example, *Bad Gateway*) was logged but otherwise ignored, and an empty page was counted as a record that did not exist. Records are now counted only after Cortex XSIAM confirms storing them, a rejected batch is treated as a failed send, and empty pages are no longer sent.
```
Shellyber: I don't think we should elaborate that much. 
Let's try to minimize the release notes as much as possible and keep them concise.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -888,18 +904,7 @@ def _get_process_memory_mb() -> str:
     else:
         peak_rss_mb = rusage.ru_maxrss / 1024
 
-    # Python tracked allocations via tracemalloc
-    # The gap between RSS and traced = arena fragmentation + non-Python allocations
-    traced_mb = 0.0
```
Shellyber, line 893: What's the change here? 
Not merged from master?

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4197,14 +4203,28 @@ async def xsiam_api_call_async(
...
-                        demisto.updateModuleHealth(header_msg + e.message, is_error=True)
+                        # Only logged here: a retry may still succeed, and reporting every attempt
+                        # to the health module turns a recovered blip into a red instance.
+                        last_error = e
+                        log_falcon_assets(f"Error sending {data_type} to XSIAM: {e.message}", "error")
+                        if attempt_num < num_of_attempts:
+                            await asyncio.sleep(XSIAM_SEND_RETRY_BACKOFF_SECONDS)
 
         log_falcon_assets(f"received status code: {status_code}")
         if status_code == 429:
```
Shellyber: This might not be related to your changes now, but the bottom-of-loop if status_code == 429 check is redundant with the in-exception handler and should be removed. The attempt_num += 1 inside the 429 handler should be moved after the continue guard to avoid double-incrementing if the continue is ever removed.

A 429 followed by a different error (502, 500) causes an unnecessary 1-second sleep on the wrong code path. More critically, the stale status_code can cause confusing log messages ("received status code: 429" when the actual response was 502). While the final outcome (raise on failure) is still correct, the retry budget is consumed incorrectly, and the backoff timing is wrong.

Suggested Fix:
```
except aiohttp.ClientResponseError as e:
    if ok_codes and e.status in ok_codes:
        status_code = e.status
        if e.status == 429:
            await asyncio.sleep(1)
        attempt_num += 1
        continue
    else:
        status_code = None  # Clear stale status from previous iterations
        last_error = e
        log_falcon_assets(f"Error sending {data_type} to XSIAM: {e.message}", "error")
        if attempt_num < num_of_attempts:
            await asyncio.sleep(XSIAM_SEND_RETRY_BACKOFF_SECONDS)

log_falcon_assets(f"received status code: {status_code}")
attempt_num += 1

```

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -4703,6 +4865,54 @@ async def fetch_vulnerabilities_by_severity(
...
+                # Whole batch rejected: nothing is counted, so declared == stored still holds.
+                # CancelledError is listed explicitly: since Python 3.8 it does not derive from
+                # Exception, and a cancelled send task would otherwise kill the whole severity.
+                lost_send_batches += 1
+                lost_send_records += items_attempted
+                first_send_error = first_send_error or e
+                log_falcon_assets(
+                    f"[{severity}] Background vulnerability send task failed; "
+                    f"{items_attempted} records NOT counted and skipped (continuing with the next pages): {e}\n"
+                    f"{traceback.format_exc()}",
```
Shellyber, line 4932:
```suggestion
                    f"{''.join(traceback.format_exception(type(e), e, e.__traceback__))}",
```
In reap_completed_send_tasks(), the except block calls traceback.format_exc(). However, this function is called when iterating over completed tasks and calling .result() on them. The exception from .result() is a re-raised exception from the task, not the current exception context. traceback.format_exc() returns the traceback of the current exception handler, which is the except (asyncio.CancelledError, Exception) as e — this will show the traceback of the .result() call, not the original task failure traceback.

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
```diff
@@ -5273,6 +5536,91 @@ def fetch_assets_command():
...
+                "info",
+            )
+        else:
+            # Cycle met or exceeded the period; start the next immediately instead of piling up.
+            log_falcon_assets(
+                f"Long-running Spotlight cycle took {elapsed / 3600:.1f} hours, meeting or exceeding "
+                f"the {LONG_RUNNING_ASSETS_INTERVAL_MINUTES}-minute interval; starting the next cycle immediately.",
+                "info",
+            )
+        time.sleep(sleep_seconds)
```
Shellyber: sleep_seconds can be up to 86400 seconds (24 hours). This blocks the entire thread for the sleep duration. In a long-running XSOAR container, this means the container cannot respond to any signals, health checks, or shutdown requests during the sleep.
Consider using a shorter sleep loop that checks for shutdown signals:

```
# Instead of:
time.sleep(sleep_seconds)

# Use:
remaining = sleep_seconds
while remaining > 0:
    chunk = min(remaining, 60)  # Wake every 60 seconds to check for signals
    time.sleep(chunk)
    remaining -= chunk

```

Shellyber (commented):
Nice.
I want docs team to review as well after the release notes changes please

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
+
+#### Integrations
+
+##### CrowdStrike Falcon
+
+- Fixed an issue where Spotlight assets were never collected on tenants that span multiple CIDs (CrowdStrike Flight Control / MSSP parent-child deployments). On such tenants Spotlight returns the CID and the device ID joined together, which the CrowdStrike Devices API rejects. The device ID is now extracted before the request is sent. Vulnerability collection was not affected.
```
julieschwartz18:
```suggestion
- Fixed an issue in multi-CID tenants (CrowdStrike Flight Control / MSSP parent-child deployments) where Spotlight assets were not collected because the device_id included the CID. Vulnerability collection was unaffected.
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
julieschwartz18:
```suggestion
- Fixed an issue where the Spotlight vulnerabilities assets snapshot failed to seal due to mismatched record counts in Cortex XSIAM.
```
@MLainer1 Is this just for Cortex XSIAM? Maybe use a more generic term

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+- Fixed an issue where the Spotlight vulnerabilities assets snapshot never sealed because the declared record count did not match what was actually stored in Cortex XSIAM. Records were counted before being sent, an HTTP error returned while sending a batch (for example, *Bad Gateway*) was logged but otherwise ignored, and an empty page was counted as a record that did not exist. Records are now counted only after Cortex XSIAM confirms storing them, a rejected batch is treated as a failed send, and empty pages are no longer sent.
+- Fixed an issue where a single failure aborted an entire severity fetch, so all remaining pages of that severity were lost. A failed or cancelled background send is now skipped like any other failed send, the remaining pages are still fetched, and the records that were not stored are picked up on the next fetch cycle.
```
julieschwartz18:
```suggestion
- Fixed an issue where a single failure aborted the fetch process for an entire severity level, causing remaining pages to be lost.
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+- Fixed an issue where a page that could not be fetched aborted an entire severity fetch, either because an oversized page was truncated and raised a JSON decode error, or because a transient CrowdStrike service error exhausted the API client's retries. The same page is now retried with progressively smaller page sizes and reverts to the default size on the next page, so a temporary upstream fault no longer prevents the snapshot from sealing.
```
julieschwartz18:
```suggestion
- Fixed an issue where an un-fetched page aborted the entire severity fetch due to page truncation JSON errors or exhausted CrowdStrike API retries.
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+- Fixed an issue where a transient Cortex XSIAM error while sending a batch (for example, *Bad Gateway*) marked the instance as errored even when a retry succeeded and no data was lost, and reported the reason twice in the error message. The error is now reported only after all retry attempts fail. Retries were also sent immediately one after another, so a brief error outlasted all of them; the integration now waits between attempts.
```
julieschwartz18:
```suggestion
- Fixed an issue where transient Cortex XSIAM errors that occurred while sending a batch falsely marked the instance as errored despite successful retries.
```
@MLainer1 Is this just for Cortex XSIAM? Maybe make it generic Cortex

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.py
MLainer1: I removed the tracemalloc usage, since it is consuming run-time memory and is not needed anymore

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+- Added the *Long running instance for Spotlight vulnerabilities* parameter, which runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule. Use it when a full Spotlight fetch takes longer than the assets fetch interval and is therefore repeatedly interrupted. A cycle starts every 24 hours and always runs to completion first, so a fetch may take longer than 24 hours. A failed cycle is retried after a short delay that grows with each consecutive failure.
```
julieschwartz18:
```suggestion
- Added the *Long running instance for Spotlight vulnerabilities* parameter, which runs the Spotlight vulnerabilities fetch in a long-running container instead of on the assets fetch schedule.
```

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
```diff
@@ -0,0 +1,12 @@
...
+NOTE: When this option is enabled, do not also enable the assets fetch on the same instance. This mode fetches Spotlight vulnerabilities only; to collect CNAPP Alerts, configure a separate instance using the regular assets fetch.
```
julieschwartz18:
```suggestion
NOTE: This mode fetches Spotlight vulnerabilities only; to collect CNAPP Alerts, configure a separate instance using the regular assets fetch.
```

julieschwartz18 (commented):
@MLainer1 Doc review complete. Please check the comments. In the meantime, adding the label docs-approved

Packs/CrowdStrikeFalcon/Integrations/CrowdStrikeFalcon/CrowdStrikeFalcon.yml
MLainer1: Agentix appears on nine parameters in this yml: all incident-fetch ones (isFetch, fetch_time, look_back, etc.).
Every assets parameter (isFetchAssets, fetch_assets_type, assetsFetchInterval) uses exactly xsiam + exposure_management, which is what longRunning has. Adding agentix to one assets parameter would make it inconsistent with the other three.

Packs/CrowdStrikeFalcon/ReleaseNotes/2_13_0.md
MLainer1: Added a validation for it in the code

PR #45577, opened 2026-08-19
Reco: fix fetch-incidents skipping alerts during bursts

yanivblumReco:
## Summary

During an alert flood, `fetch-incidents` could permanently skip the majority of alerts in a single time window. Separately, `fetch-incidents` was also ingesting alerts belonging to Policies in *Preview* state, which should never leave Reco.

**Fix 1 - burst pagination:**

**Root cause:** `get_alerts` fetched one page per cycle using the External API's default sort (`createdAt` descending), then `fetch_incidents` advanced the `lastRun` cursor to the newest timestamp seen in that page. When a burst exceeded `Max fetch` in one window, the returned page was the *newest* slice of the burst, so the cursor jumped past all the older alerts that didn't make it into that page — silently dropping them forever (reported: 669/1069 alerts lost in a one-minute burst, Iron Mountain support ticket).

- `get_alerts` now requests ascending `createdAt` order and forwards `start_index`.
- `fetch_incidents` only advances `lastRun` once a page comes back smaller than `max_fetch` (i.e. the window is fully drained). A full page instead advances a `startIndex` offset persisted in `last_run`, so a burst larger than `max_fetch` is paginated across multiple fetch cycles instead of being truncated.

**Fix 2 - exclude Preview-state alerts:**

**Root cause:** a Policy in *Preview* state is meant to keep its alerts out of the customer's SOAR while it's being evaluated. Reco's webhook and share-service delivery paths already enforce this, but `fetch-incidents` (which polls the External API directly) had no equivalent guard, so Preview alerts leaked into XSOAR.

- `get_alerts` now always adds `status ne "ALERT_STATUS_PREVIEW"` to its SCIM filter.

Version bumped 1.8.0 → 1.8.1.

## Test plan

- [x] Local: `pytest Packs/Reco/Integrations/Reco/Reco_test.py` — 72/72 pass (Docker unavailable locally, so `demisto-sdk pre-commit`'s ruff/xsoar-lint/pylint/mypy steps will run via CI on this PR rather than locally)
- [x] New unit tests: burst spanning two fetch cycles resumes via `startIndex` without skipping; `get_alerts` sends ascending sort + pagination params; `get_alerts` excludes `ALERT_STATUS_PREVIEW` alongside other filters
- [x] `status ne "ALERT_STATUS_PREVIEW"` verified against the live External API (confirmed a known Preview alert is excluded by this exact filter)

### DIFF
--- a/Packs/Reco/.pack-ignore
+++ b/Packs/Reco/.pack-ignore
@@ -1,3 +1,5 @@
 [known_words]
 SCIM
 backoff
+paginated
+resumable
--- a/Packs/Reco/Integrations/Reco/Reco.py
+++ b/Packs/Reco/Integrations/Reco/Reco.py
@@ -137,11 +137,17 @@ class RecoClient(BaseClient):
         filters: str = "",
         count: int = PAGE_SIZE,
         start_index: int = 0,
+        sort_by: str = "",
+        sort_order: str = "",
     ) -> dict[str, Any]:
-        """GET /external-api/{endpoint} with SCIM filter and pagination params."""
+        """GET /external-api/{endpoint} with SCIM filter, sort, and pagination params."""
         params: dict[str, Any] = {"count": count, "startIndex": start_index}
         if filters:
             params["filters"] = filters
+        if sort_by:
+            params["sortBy"] = sort_by
+        if sort_order:
+            params["sortOrder"] = sort_order
         return self._rate_limited_request(
             method="GET",
             url_suffix=f"{EXTERNAL_API_BASE}/{endpoint}",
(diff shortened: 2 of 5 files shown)


kamalq97 (approved):

kamalq97:
Hi @yanivblumReco 

Thanks for your contribution!

Could you please ensure the [Contributor License Agreement](https://cla-assistant.io/demisto/content?pullRequest=45577) is signed?

Thank you

yanivblumReco:
@kamalq97 Everything is signed, what else is missing from our side?

yanivblumReco:
@kamalq97 any update here? Our customers are waiting on this fix

kamalq97:
Hi @yanivblumReco 

Apologies for the delay in replying,

We will shortly be merging this PR.

PR #45685, opened 2026-08-30
Unit42Feed - Incremental Improvements

TheL0L:
## Related Issues
fixes: [CIAC-17491](https://jira-dc.paloaltonetworks.com/browse/CIAC-17491)

## Description
Introduce the following changes:
- Decrease the default value of *Maximum Indicators Per Fetch* from 100,000 to 20,000.
- Decrease the fetch interval from 12 hours to 1 hour.
- Update fetch logic to preserve progress state across runs.


marketplace-ai-reviewer:

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1080,71 +1142,70 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
+        # The limit was hit but the API has more pages - resume this unit on the next fetch
+        if next_page_token:
+            pending_units.append({"type": unit_type, "page_token": next_page_token})
+
+        demisto.debug(
+            f"UNIT42FEED: Fetched {fetched_count}/{type_limit} {unit_type}. "
+            f"Total: {total_fetched}, Unused quota: {remaining_quota}"
+        )
+
+    if pending_units:
```
marketplace-ai-reviewer, line 1200: There is a data loss bug here when a fetch cycle spans across multiple runs due to `pending_units`.

If Run 1 starts at T1 and hits the limit, it saves `pending_units`. The types that *did not* hit the limit (e.g., IPs) were fetched up to T1.
If Run 2 (at T2) and Run 3 (at T3) continue fetching the `pending_units` until they are exhausted, the `else` block will execute in Run 3 and set `last_successful_run` to T3 (`current_time`).
In Run 4, the next fetch will start from T3. This means any indicators created between T1 and T3 for the types that finished in Run 1 (and for the pending types, depending on how the API's page token works) will be completely missed.

To fix this, you must preserve the `current_time` of the *original* run that initiated the pending state, and use it as the `last_successful_run` when the pending units are finally exhausted:

```python
    if pending_units:
        # Preserve the original run time if we are already in a pending state
        original_run_time = last_run.get("original_run_time") or current_time.strftime(DATE_FORMAT)
        next_run = {
            "start_time": start_time, 
            "pending_units": pending_units,
            "original_run_time": original_run_time
        }
        demisto.info(
            f"UNIT42FEED: Fetch limit reached with more data available. "
            f"Next run will resume {len(pending_units)} pending type(s) from start_time {start_time}."
        )
    else:
        # Use the original run time if we just finished pending units, otherwise use current_time
        next_start = last_run.get("original_run_time") or current_time.strftime(DATE_FORMAT)
        next_run = {"last_successful_run": next_start}
```

*(Note: You will also need to update the mock `last_run` objects in the corresponding unit tests to include `original_run_time`)*

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1060,14 +1118,18 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
-    # Get start time
+    # Get start time - a pending query from an interrupted fetch takes precedence,
+    # so the resumed pages keep querying the same time window.
     default_start = (current_time - timedelta(hours=24)).strftime(DATE_FORMAT)
     last_run = demisto.getLastRun() or {}
-    start_time = last_run.get("last_successful_run", default_start)
+    if last_run.get("pending_units"):
+        start_time = last_run.get("start_time") or default_start
+    else:
+        start_time = last_run.get("last_successful_run", default_start)
```
marketplace-ai-reviewer: Logical flaw in state management causes missed indicators for types that finish early.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1080,71 +1142,70 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
+        )
+
+    if pending_units:
+        next_run = {"start_time": start_time, "pending_units": pending_units}
+        demisto.info(
+            f"UNIT42FEED: Fetch limit reached with more data available. "
+            f"Next run will resume {len(pending_units)} pending type(s) from start_time {start_time}."
+        )
+    else:
+        next_run = {"last_successful_run": current_time.strftime(DATE_FORMAT)}
```
marketplace-ai-reviewer: Update `next_run` to persist the `cycle_start_time` across resumed runs.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
```diff
@@ -57,15 +57,15 @@ configuration:
     4. URLs
     5. Files
     
-    Default (total for all types): 100,000. Maximum (total for all types): 100,000.
+    Default (total for all types): 20,000. Maximum (total for all types): 100,000.
```
marketplace-ai-reviewer:
```suggestion
    Maximum (total for all types): 100,000.
```
According to the [Cortex Content YAML Guidelines](https://xsoar.pan.dev/docs/integrations/yaml-file#description-quality), if a `defaultvalue` is set, the default value **must be removed from the description**. 

Additionally, please update line 45 to start with "The " (e.g., "The maximum number of indicators to fetch per type.") to comply with the descriptive phrasing guidelines.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
```diff
@@ -57,15 +57,15 @@ configuration:
...
   required: false
   section: Collect
   advanced: true
 - display: Feed Fetch Interval
-  additionalinfo: Don't set the feed fetch interval to less than 12 hours.
+  additionalinfo: Feed fetch interval is set to 1 hour and is not configurable.
```
marketplace-ai-reviewer:
```suggestion
  additionalinfo: The feed fetch interval, which is set to 1 hour and is not configurable.
```
If this parameter is truly not configurable, consider adding `hidden: true` to this parameter block so it doesn't appear as a configurable picker (`type: 19`) in the UI.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
```diff
@@ -4,15 +4,15 @@ Unit 42 Feed integration provides threat intelligence from Palo Alto Networks Un
 
 | **Parameter** | **Description** | **Required** |
 | --- | --- | --- |
-| Fetch indicators | Select this check box to fetch indicators \(default selected\). | True |
-| Feed Types | Choose the requested indicator feeds. Indicators feed and Threat Objects \(actors, malware, campaigns, techniques, etc.\) feed \(default is both\). | True |
+| Fetch indicators | Select this check box to fetch indicators \(default selected\). |  |
```
marketplace-ai-reviewer, line 7: The `Required` column for the `Fetch indicators` parameter is empty. According to the [Creating Documentation](https://xsoar.pan.dev/docs/documentation/readme_file#creating-documentation) guidelines, this column must contain either `True` or `False`.

Please update the `required` field for this parameter in the integration YAML file and regenerate the README using `demisto-sdk generate-docs`.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
```diff
@@ -4,15 +4,15 @@ Unit 42 Feed integration provides threat intelligence from Palo Alto Networks Un
...
+| Feed Types | Choose the requested indicator feeds. Indicators feed and Threat Objects \(actors, malware, campaigns, techniques, etc.\) feed \(default is both\). | False |
 | Indicator Types | Comma-separated list of indicator types to fetch \(File, IP, URL, Domain\). If not specified, all indicator types are fetched. | False |
-| Maximum Indicators Per Fetch | Maximum number of indicators to fetch per type. | False |
-| Source Reliability | Reliability of the source providing the intelligence context. | True |
+| Maximum Indicators Per Fetch | Maximum number of indicators to fetch per type. <br/>Example: limit=20,000 with 4 types selected = 80,000 total indicators.<br/><br/>The total \(limit × number of types\) is automatically capped at 100,000.<br/>If limit × types &gt; 100,000, the limit per type is adjusted.<br/>Example: limit=30,000 with 4 types selected = 120,000 total indicators<br/>adjusted to 25,000 per type \(100,000 / 4 = 25,000\).<br/><br/>Fetch priority order:<br/>1. Threat Objects<br/>2. IPs<br/>3. Domains<br/>4. URLs<br/>5. Files<br/><br/>Default \(total for all types\): 20,000. Maximum \(total for all types\): 100,000. | False |
```
marketplace-ai-reviewer: There appears to be a contradiction in the description regarding the default value. The text states `Default (total for all types): 20,000`, but earlier it says the parameter defines the `Maximum number of indicators to fetch per type`. If the default limit is 20,000 per type, the total for all types would be higher (e.g., 80,000 for 4 types).

If the default limit is applied per type, consider updating the text in the integration YAML's `additionalinfo` field to `Default (per type): 20,000` and regenerating the README.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
```diff
@@ -4,15 +4,15 @@ Unit 42 Feed integration provides threat intelligence from Palo Alto Networks Un
...
+| Indicator Reputation | Indicators from this integration instance will be marked with this reputation. | False |
+| Source Reliability | Reliability of the source providing the intelligence context. | False |
 | Tags | Supports CSV values. | False |
-| Traffic Light Protocol Color (TLP). | The Traffic Light Protocol \(TLP\) designation is to apply to indicators fetched from the feed. | False |
-| Indicator Reputation | Indicators from this integration instance will be marked with this reputation | True |
-| Feed Expiration Policy | The feed's expiration policy. | True |
+| Traffic Light Protocol Color (TLP). | The Traffic Light Protocol \(TLP\) designation applied to indicators fetched from the feed. | False |
+|  | The feed's expiration policy. | False |
```
marketplace-ai-reviewer, line 15: The `Parameter` column is empty for this row (previously `Feed Expiration Policy`). 

Please ensure the `display` field for this parameter is correctly set in the integration YAML file and regenerate the README using `demisto-sdk generate-docs`.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/ReleaseNotes/1_0_25.md
```diff
@@ -0,0 +1,9 @@
+
+#### Integrations
+
+##### Unit 42 Feed
+
+- Updated the Docker image to: *demisto/python3:3.12.14.12343672*.
+- Changed the default value of *Maximum Indicators Per Fetch* from 100,000 to 20,000 to improve fetch performance.
```
marketplace-ai-reviewer: The prefix "Changed the" is not in the approved list of prefixes for release notes.

```suggestion
- Updated the default value of *Maximum Indicators Per Fetch* from 100,000 to 20,000 to improve fetch performance.
```

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/ReleaseNotes/1_0_25.md
```diff
@@ -0,0 +1,9 @@
...
+- Changed the fetch interval from 12 hours to 1 hour.
```
marketplace-ai-reviewer: The prefix "Changed the" is not in the approved list of prefixes for release notes.

```suggestion
- Updated the fetch interval from 12 hours to 1 hour.
```

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
Hi! Thanks for your contribution to the Unit42Feed integration. 

I reviewed the code and found a critical data loss bug in the fetch logic where indicators can be missed if a fetch cycle spans multiple runs; you'll need to persist the original run time across these runs to resolve this. There are also a few minor documentation and YAML adjustments needed, such as fixing empty columns in the README, clarifying the fetch limit description, and using approved prefixes in the release notes. 

Let me know if you have any questions about the state management fix!

- `Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/ReleaseNotes/1_0_25.md`: The version bump type (patch) does not match the nature of the changes described.

 @TheL0L please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
TheL0L: Already set as `hidden: true`.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
TheL0L: Will not address.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
TheL0L: Will not fix.

julieschwartz18 (commented):
@TheL0L Doc review complete. No updates, well done! Adding the label docs-approved

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
TheL0L, line 1018: @MosheEichler Could you please review this section for potential dropped indicators when the API returns more than the specified limit?

TheL0L, line 944: @MosheEichler Could you please review this section for potential dropped indicators when the API returns more than the specified limit?

kamalq97 (approved):

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
MosheEichler: Yes — and here it happens with a compliant API, not just a misbehaving one.

A single threat object (may) expands during parsing (`map_threat_object` returns the object **plus** its derived location indicators, and `create_relationships` is on by default), so `len(page_objects) > len(data)`. But the limit is checked **after** the parse, against the expanded list — while `page_limit` was requested in API objects. So a page that the API returned correctly gets truncated, and the `next_page_token` read from that same response points past the discarded part. Those objects are never re-fetched.

It also cuts mid-object: locations are added before their parent, so we can persist Location indicators whose threat actor was dropped, leaving dangling relationships.

Suggested fix — count API objects, not derived indicators:

```suggestion
        # Push the whole page as one unit: a threat object and its derived location
        # indicators must not be split, or we persist locations whose parent is missing.
        push_indicators_in_batches(page_objects)

        # Count objects consumed from the API - not the indicators they expand into -
        # so the limit stays aligned with the page token used to resume.
        total_fetched += len(data)
```

(delete the `if ... > limit:` truncation block and change the counter below to `len(data)`)

MosheEichler: Yes, though narrower than the same check in `fetch_threat_objects_with_limit` (see that thread).

`parse_indicators` is 1:1, so this only fires if the API returns more than requested — but then the surplus is sliced off while `next_page_token` from that same response acknowledges it, so those indicators are never fetched again.

Suggested fix — push the full page instead. `createIndicators` is idempotent by value, so duplicates are free; the overshoot is capped at one page and the `while total_fetched < limit` guard still stops the loop:

```suggestion
        # Push the full page even if the API returned more than requested: the page token
        # below acknowledges the whole page, so dropping the surplus would lose it
        # permanently. Overshooting by one page is cheaper, and createIndicators dedups.
        push_indicators_in_batches(page_indicators)
```

(delete the `if ... > limit:` truncation block)

MosheEichler (approved):

johnnywilkes:
[TheL0L](https://github.com/TheL0L), tested and found a major concern.  Put information in [ticket](https://jira-dc.paloaltonetworks.com/browse/CIAC-17491)

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
```diff
@@ -40,32 +40,32 @@ configuration:
 - display: Maximum Indicators Per Fetch
   name: limit
   type: 0
-  defaultvalue: '100000'
+  defaultvalue: '20000'
```
johnnywilkes, line 43: :red_circle: **Critical (1/12) — the new default does not produce 20k indicators; it produces 100k**

The requirement asks for a 20k default. But `limit` is applied **per type**, and `calculate_limit_per_type()` only rejects the requested value when `limit × types > 100_000`:

```python
total_expected = limit * total_indicator_types
if total_expected > TOTAL_INDICATOR_LIMIT:   # 100,000
    return default_limit_per_type            # 100_000 // total_types
return int(limit)
```

With the default config (Threat Objects + IP/Domain/URL/File = 5 units), `20,000 × 5 = 100,000` does *not* exceed the cap, so each unit fetches 20k for **100k total per fetch**.

That is identical to the old behavior: under the previous default of `100000`, `100,000 × 5 = 500,000 > 100,000`, so it always silently fell back to `100_000 // 5 = 20,000` per type. **The effective per-type limit is unchanged.**

Combined with the interval going 12h → 1h, worst-case ingest rate increases **12×** (100k/12h → 100k/1h) — the opposite of the "improve fetch performance" intent stated in the release note.

If the intent is a 20k *total* budget, the limit needs dividing rather than multiplying:

```python
limit_per_type = min(requested_limit, TOTAL_INDICATOR_LIMIT) // total_types
```

If per-type really is intended, the release note and README should say so explicitly, since "from 100,000 to 20,000" reads as a 5× volume reduction that won't materialize.

johnnywilkes (commented):
## IDEX Code Review — Unit42Feed Incremental Improvements

_This review was generated as part of an **IDEX code review** and checked against the originally requested changes for CIAC-17491._

The incremental-fetch mechanism is well-built. The fetch-unit abstraction in `build_fetch_units()` cleanly collapses the previously duplicated threat-objects branch and indicator-types loop, the page truncation removal is a genuine data-loss bug catch, and the `cycle_start_time` handling shows real thought about cycles that span multiple runs. The 21 new tests cover the happy paths thoroughly.

My concerns concentrate in three areas, detailed inline below.

### Requirements traceability

| Requirement | Status |
|---|---|
| Keep `Maximum Indicators Per Fetch` configurable, default 20k | :warning: Configurable, but default yields **100k**, not 20k |
| Stop paging when limit is hit | :white_check_mark: |
| Fetch hourly, not configurable | :warning: Hidden field set correctly, but won't apply to existing instances |
| Store query + page_token in `setLastRun` when limit hit | :white_check_mark: |
| Store `{"last_successful_run": ...}` when limit not hit | :warning: Stores `cycle_start_time`, not `now` (deliberate, but differs from spec) |
| On new fetch, prefer stored query over `last_successful_run` | :white_check_mark: |

### Summary of findings

**Critical (3)** — the limit semantics don't deliver the requested 20k default; the cycle can never complete under sustained load; the new persisted state has no recovery path when a page token expires.

**High (5)** — resumed-run quota over-fetch; pending units replayed without re-validating config; `limit=0` silently disables the feed; threat-object limit no longer bounds indicators sent to TIM; existing instances won't pick up the 1-hour interval.

**Medium (4)** — unverified dedup claim in a comment; two stale `Returns:` docstrings; README parameter-table regressions; missing tests for the failure paths above.

### Two items without a natural inline anchor

**README parameter table regressions** — the regenerated `README.md` lost the `Feed Expiration Policy` display name (the row now has an empty name cell) and flipped several `Required` values from `True` to `False`/blank. Worth re-running `demisto-sdk generate-docs` to confirm this reflects the YAML rather than a generation artifact.

**Missing test coverage** — no tests for: page-token rejection/expiry, config change mid-cycle, `limit=0`, or the resumed-run quota-redistribution math. Findings 3, 5, and 6 below all live in those gaps.

---

Items 1, 2, and 3 are worth resolving before merge. Happy to discuss any of these — several involve intentional trade-offs where I may be missing context.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1080,71 +1142,72 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
+        if next_page_token:
+            pending_units.append({"type": unit_type, "page_token": next_page_token})
+
+        demisto.debug(
+            f"UNIT42FEED: Fetched {fetched_count}/{type_limit} {unit_type}. "
+            f"Total: {total_fetched}, Unused quota: {remaining_quota}"
+        )
+
+    if pending_units:
+        next_run = {"start_time": start_time, "pending_units": pending_units, "cycle_start_time": cycle_start_time}
```
johnnywilkes, line 1198: :red_circle: **Critical (2/12) — cycle can starve permanently; `last_successful_run` never advances**

```python
if pending_units:
    next_run = {"start_time": start_time, "pending_units": pending_units, "cycle_start_time": cycle_start_time}
else:
    next_run = {"last_successful_run": cycle_start_time}
```

Both `start_time` and `cycle_start_time` are frozen for the entire duration of a pending cycle. If the feed consistently returns more than the per-fetch budget, `pending_units` is never empty, the cycle never completes, and the time window **never advances**.

The integration then re-queries an ever-widening window forever and falls progressively further behind — each run does more work than the last, making completion *less* likely over time. There's no cycle-age cap or forced-advance escape hatch.

Given the 12× increase in fetch frequency, sustained high-volume periods are considerably more likely to trigger this than under the old 12h interval.

Consider bounding the cycle — force-complete after N resumed runs, or when `cycle_start_time` exceeds some age threshold — and emit a warning when it happens so the condition is visible.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -919,21 +929,21 @@ def fetch_indicator_type(
         # Parse response
         if not response or not isinstance(response, dict):
             demisto.debug(f"UNIT42FEED: Invalid response for {indicator_type}, stopping")
+            next_page_token = None
```
johnnywilkes, line 932: :red_circle: **Critical (3/12) — expired or invalid page token leaves the feed permanently stuck**

These guards clear the token only for *falsy or non-dict* responses:

```python
if not response or not isinstance(response, dict):
    next_page_token = None
    break
```

They do not cover the exception path. Page tokens are now persisted across runs — potentially many hours. If a stored token expires or is rejected by the API, `client.get_indicators()` raises, `main()` errors out, and `demisto.setLastRun()` is **never reached**.

The same bad token is therefore retried on every subsequent fetch, indefinitely. The feed silently stops ingesting with no self-recovery.

This is a new failure mode introduced by persisting tokens — previously a transient API error simply meant the next run started fresh.

Suggest wrapping the paged call so a token-rejection error drops `pending_units` and restarts the unit from `start_time`. Currently untested; a test asserting recovery from a rejected token would be valuable.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1080,71 +1142,72 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
+    for idx, fetch_unit in enumerate(fetch_units):
+        unit_type = fetch_unit.get("type")
+        page_token = fetch_unit.get("page_token")
+        is_last_type = idx == len(fetch_units) - 1
 
-        # Track unused quota
-        if fetched_count < limit_per_type:
-            remaining_quota += limit_per_type - fetched_count
+        # For the last type, add remaining quota
+        type_limit = limit_per_type + remaining_quota if is_last_type else limit_per_type
```
johnnywilkes, line 1159: :orange_circle: **High (4/12) — quota redistribution over-fetches on resumed runs**

```python
type_limit = limit_per_type + remaining_quota if is_last_type else limit_per_type
```

On a resumed run, `fetch_units` is the shorter `pending_units` list, but `limit_per_type` is still derived from the **full configured** `total_types`.

So a resume with 2 pending units receives a budget of `2 × limit_per_type` (plus redistribution), even though only 2 of 5 types are actually running. Each resumed run is budgeted as though it were a full cycle, so a multi-run cycle can ingest several times the intended ceiling.

Separately, even on a fresh run this means no single type is bounded by `limit` — the last unit can absorb all unused quota from the preceding four.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1036,7 +1050,41 @@ def test_module(client: Client) -> str:
...
+    Args:
+        feed_types: The configured feed types (Indicators / Threat Objects)
+        indicator_types: The configured indicator types
+        last_run: demisto.getLastRun()
+
+    Returns:
+        List of fetch units, each holding a "type" and its "page_token" (None when not resumed).
+    """
+    pending_units = last_run.get("pending_units") or []
+    if pending_units:
```
johnnywilkes, line 1069: :orange_circle: **High (5/12) — pending units are replayed without re-validating against current config**

```python
pending_units = last_run.get("pending_units") or []
if pending_units:
    demisto.debug(f"UNIT42FEED: Resuming pending fetch units: {pending_units}")
    return pending_units
```

The stored units are returned verbatim, bypassing `feed_types` and `indicator_types` entirely.

If an admin edits the instance configuration mid-cycle, the change is ignored until the cycle drains: deselected types keep being fetched, and newly selected types are not fetched at all. With a long-running pending cycle (see finding 2) this could persist indefinitely.

Suggest filtering `pending_units` against the current config before returning, and dropping the pending state entirely when the config no longer matches.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -950,10 +960,12 @@ def fetch_indicator_type(
             demisto.debug(f"UNIT42FEED: No more pages for {indicator_type}")
             break
 
-    return total_fetched
+    return total_fetched, next_page_token
```
johnnywilkes, line 963: :orange_circle: **High (6/12) — `limit=0` silently disables the feed forever**

`calculate_limit_per_type(0, n)` returns `0` — zero is neither `None` nor `< 0`, and `0 × n` never exceeds the cap.

Then `while total_fetched < limit` never executes even once. And because `next_page_token` is now the *function parameter* rather than a local initialized to `None`, this line returns the **incoming token unchanged**:

```python
return total_fetched, next_page_token
```

So `pending_units` gets rewritten identically on every run: a no-op loop that fetches zero indicators, raises no error, and never terminates. The feed appears healthy while ingesting nothing.

This parameter-shadowing subtlety is worth noting generally — the refactor changed `next_page_token` from a guaranteed-`None` local into a caller-supplied value, so every early-exit path now needs to be explicit about what it returns. The two invalid-response guards handle this correctly; this path does not.

Suggest validating `limit >= 1`, or treating `0` the same as `None`.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -987,25 +1000,26 @@ def fetch_threat_objects_with_limit(client: Client, limit: int, feed_tags: list,
...
-
-        # Push this page to the server immediately, in batches of 2000
+        # Push the whole page as one unit: a threat object and its derived location
+        # indicators must not be split, or we persist locations whose parent is missing.
         push_indicators_in_batches(page_objects)
 
-        total_fetched += len(page_objects)
+        # Count objects consumed from the API - not the indicators they expand into -
+        # so the limit stays aligned with the page token used to resume.
+        total_fetched += len(data)
```
johnnywilkes, line 1022: :orange_circle: **High (7/12) — threat-object limit no longer bounds indicators sent to TIM**

```python
total_fetched += len(data)
```

The rationale is sound — keeping the counter aligned with the page token is exactly right, and the accompanying comment explains it well.

The consequence, though, is that `parse_threat_objects()` can expand each API object into multiple location indicators, so the actual volume pushed for the Threat Objects unit is now unbounded relative to `limit`. A page of 100 objects might yield several hundred indicators, all counted as 100 against the budget.

Combined with the removed page truncation, a single fetch can materially exceed the advertised 100,000 ceiling. Worth documenting in the parameter help text, since "Maximum (total for all types): 100,000" is no longer strictly accurate.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
```diff
@@ -40,32 +40,32 @@ configuration:
...
   required: false
   section: Collect
   advanced: true
 - display: Feed Fetch Interval
-  additionalinfo: Don't set the feed fetch interval to less than 12 hours.
+  additionalinfo: The feed fetch interval, which is set to 1 hour and is not configurable.
   name: feedFetchInterval
   type: 19
-  defaultvalue: '720'
+  defaultvalue: '60'
```
johnnywilkes, line 68: :orange_circle: **High (8/12) — existing instances will not pick up the 1-hour interval**

Changing `defaultvalue` from `720` to `60` affects only **newly created** instances. Upgraded instances retain their stored `720` value.

Because the field is `hidden: true` and the `return_error` validation guard was removed from `main()`, nothing surfaces the discrepancy — existing customers stay on a 12-hour interval indefinitely with no indication.

That matters more than usual here: the incremental-fetch design assumes frequent runs to drain pending units. On a 12h interval, a cycle that needs 3 resumed runs takes 36 hours to complete.

The release note states "Updated the fetch interval from 12 hours to 1 hour" unconditionally, which won't be true for existing instances. Either document the manual step or handle the migration in code.

(Marking the field `hidden: true` to satisfy the "not configurable" requirement is the right call — that part looks good.)

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -919,21 +929,21 @@ def fetch_indicator_type(
...
         page_indicators = parse_indicators(data, feed_tags, tlp_color)
 
-        # Ensure we don't exceed the limit (safety check, in case the API returns more than requested)
-        if total_fetched + len(page_indicators) > limit:
-            page_indicators = page_indicators[: limit - total_fetched]
-
-        # Push this page to the server immediately, in batches of 2000
+        # Push the full page even if the API returned more than requested: the page token
+        # below acknowledges the whole page, so dropping the surplus would lose it
+        # permanently. Overshooting by one page is cheaper, and createIndicators dedups.
```
johnnywilkes, line 946: :yellow_circle: **Medium (9/12) — unverified claim in the justifying comment**

> `# permanently. Overshooting by one page is cheaper, and createIndicators dedups.`

The core reasoning is correct and this is a good catch — the page token acknowledges the whole page, so dropping the surplus would lose it permanently. That argument stands on its own.

The "createIndicators dedups" clause is the weak part: deduplication is a server-side indicator-merge behavior, not a documented guarantee of the `createIndicators` API. If someone later relies on that claim to justify a larger overshoot, it could mislead. Suggest dropping the clause.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -897,12 +906,13 @@ def fetch_indicator_type(
         start_time: Start time for fetching
         feed_tags: Tags to add to indicators
         tlp_color: TLP color
+        next_page_token: Page token to resume an interrupted fetch from
 
     Returns:
-        Number of indicators fetched and pushed (count <= limit)
+        Tuple of the number of indicators fetched and pushed (count <= limit), and the
```
johnnywilkes, line 912: :yellow_circle: **Medium (10/12) — `Returns:` docstring is now inaccurate**

> `Tuple of the number of indicators fetched and pushed (count <= limit), ...`

The `(count <= limit)` guarantee no longer holds — removing the truncation was precisely so that a full page gets pushed even when it overshoots. The count can now exceed `limit` by up to one page.

Same issue in `fetch_threat_objects_with_limit()` (line 985), where it's compounded by the object-vs-indicator counting change: that count can exceed `limit` by considerably more than one page's worth of indicators.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
```diff
@@ -4,15 +4,15 @@ Unit 42 Feed integration provides threat intelligence from Palo Alto Networks Un
...
```
johnnywilkes, line 15: :yellow_circle: **Medium (11/12) — README parameter table regressions**

This row lost its display name — it should read `Feed Expiration Policy`, but the first cell is now empty:

```
|  | The feed's expiration policy. | False |
```

There are related changes in the same table worth a second look:

- `Fetch indicators` (line 7) — `Required` went from `True` to blank rather than `True`/`False`
- `Feed Types`, `Source Reliability`, `Indicator Reputation` — all flipped `True` → `False`

The `Required` flips are plausibly legitimate, since the YAML does mark most of these `required: false`. The empty name cell is the clear defect.

Re-running `demisto-sdk generate-docs` should confirm whether this reflects the YAML or is a generation artifact. If the empty cell persists, the `feedExpirationPolicy` parameter may be missing a `display` field.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed_test.py
```diff
@@ -1837,3 +1837,716 @@ def test_parse_threat_objects_unknown_threat_class(mocker):
...
+        - Calling build_fetch_units
+    Then:
+        - No threat objects unit is included
+    """
+    units = build_fetch_units(feed_types=["Indicators"], indicator_types=["IP", "Domain"], last_run={})
+
+    assert [unit["type"] for unit in units] == ["IP", "Domain"]
+
+
+def test_build_fetch_units_resumes_pending_units():
```
johnnywilkes, line 2013: :yellow_circle: **Medium (12/12) — missing coverage for the new failure paths**

The 21 new tests cover the happy paths thoroughly — resume-from-token, token clearing, `cycle_start_time` propagation across runs, and the upgrade path from the old `lastRun` shape are all well covered. This test in particular is a good one.

The gap is that persisting state across runs introduces failure modes that aren't exercised. Each of these corresponds to a finding above:

- **Page-token rejection / expiry** (finding 3) — assert the feed recovers and restarts from `start_time` rather than wedging on a bad token
- **Config change mid-cycle** (finding 5) — set `pending_units` for a type no longer in `indicator_types`, assert it's dropped
- **`limit=0`** (finding 6) — assert this doesn't produce a silent no-op loop
- **Resumed-run quota math** (finding 4) — assert a 2-of-5-unit resume doesn't get a full-cycle budget

The first three are the ones I'd prioritize — each currently fails silently in a way that would be hard to diagnose from a running instance.

johnnywilkes (commented):
**IDEX code review — addendum.** Findings 11 and 12 were described in prose in my [previous review](https://github.com/demisto/content/pull/45685#pullrequestreview-5146334423) because I didn't initially anchor them to diff lines. Re-posting them inline so they're actionable alongside the rest.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
TheL0L: The suggested approach will yield fetches over the total limit, especially in the described scenario.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
TheL0L: As mentioned in the discussion above, this is to prevent permanent loss of indicators and de-referencing threat indicators from locations.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.yml
TheL0L: Any forced validation on this value will cause breaking changes.
This is a limitation we have to work with.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
TheL0L: The README was out of date and did not represent that actual yaml values.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "Unit 42 Threat Intelligence by Palo Alto Networks",
     "description": "Use the Unit 42 Threat Intelligence by Palo Alto Networks integrations to enrich indicators with threat intelligence data and fetch threat intelligence feeds.",
     "support": "xsoar",
-    "currentVersion": "1.0.25",
+    "currentVersion": "1.0.26",
     "author": "Cortex XSOAR",
```
TheL0L: bump to 1.1.0

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/README.md
```diff
@@ -4,78 +4,37 @@ Unit 42 Feed integration provides threat intelligence from Palo Alto Networks Un
...
 | Create relationships | Create relationships with other indicators. | False |
 | Bypass exclusion list | When selected, the exclusion list is ignored for indicators from this feed. This means that if an indicator from this feed is on the exclusion list, the indicator might still be added to the system. | False |
 | Use system proxy settings |  | False |
 | Trust any certificate (not secure) |  | False |
 
-## How the Maximum Indicators Per Fetch Parameter Works
+## How Fetching Works
 
-The **Maximum Indicators Per Fetch** parameter controls the maximum number of indicators fetched per type during each fetch cycle. The integration enforces a total limit of **100,000 indicators** across all types to ensure optimal performance.
+The integration fetches indicators and threat objects on a single shared budget per fetch. The total number of items fetched in a run is capped at **20,000** (across both threat objects and indicators combined).
```
TheL0L: Avoid using `shared budget`, use `shared limit` instead.

Packs/Unit42ThreatIntelligencebyPaloAltoNetworks/Integrations/Unit42Feed/Unit42Feed.py
```diff
@@ -1060,91 +1055,125 @@ def fetch_indicators(client: Client, params: dict, current_time: datetime) -> in
...
+        if next_page_token:
+            pending["threat_objects"] = next_page_token
+        else:
+            # TOs fully fetched this run -> reset the 24h window from now.
+            threat_objects_completed_at = current_time.strftime(DATE_FORMAT)
+        demisto.debug(f"UNIT42FEED: Fetched {fetched_count} threat objects. Total: {total_fetched}")
+
+    # Then all configured indicator types together in one combined query, using the
+    # remaining budget. On a resumed run only if indicators still had a pending token.
+    should_fetch_indicators = bool(indicator_types) and _should_fetch_indicators(
```
TheL0L: An empty list of types should fetch all types!

TheL0L:
Currently, the only failing validation is `IN124` caused by hiding the `limit` integration parameter.

PR #45721, opened 2026-09-01
Add multithreading + streaming

RosenbergYehuda:

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-17212
## Description
Add multithreading support for send-events-to-XSIAM to work with streaming.
Nightly:
https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/12695170

Screenshot 2026-09-06 at 10 35 46

Requesting a force merge due to a failing test playbook caused by a known server regression.

## Must have
- [ ] Tests
- [ ] Documentation


marketplace-ai-reviewer:

Packs/Base/ReleaseNotes/1_42_25.md
```diff
@@ -0,0 +1,6 @@
+
+#### Scripts
+
+##### CommonServerPython
+
+- Internal code improvements.
```
marketplace-ai-reviewer: Invalid release note prefix and version bump mismatch.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -14239,50 +14247,88 @@ def data_error_handler(res):
         # Streaming path: serialize+gzip one event at a time, freeing each as we go, so peak
         # memory stays ~flat. At the target chunk size we close the stream, POST it, and open a fresh one.
         target_chunk_size = min(chunk_size, XSIAM_EVENT_CHUNK_SIZE_LIMIT)
-        demisto.info("Sending events to xsiam with a single thread (streaming, free-as-you-go).")
 
-        def _post_zipped(zipped_data):
+        def _send_and_count(zipped_data, item_count):
```
marketplace-ai-reviewer: Please add type hints for the arguments and return type of this newly added function.

```python
        def _send_and_count(zipped_data: bytes, item_count: int) -> int:
```

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -14239,50 +14247,88 @@ def data_error_handler(res):
...
+        executor = None
+        all_futures = []  # type: list  # every submitted future - returned to the caller to tally counts
+        inflight = set()  # type: set  # subset not yet collected (bounds peak memory)
+        mode_desc = "multiple threads" if multiple_threads else "a single thread"
+        demisto.info("Sending events to xsiam with {} (streaming, free-as-you-go).".format(mode_desc))
+        if multiple_threads:
+            support_multithreading()
+            executor = concurrent.futures.ThreadPoolExecutor(max_workers=NUM_OF_WORKERS)
+
+        def _dispatch(zipped_data, item_count):
```
marketplace-ai-reviewer: Please add type hints for the arguments and return type of this newly added function.

```python
        def _dispatch(zipped_data: bytes, item_count: int) -> int:
```

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
Thanks for your contribution! I've reviewed your PR and found a few minor things to adjust. Please fix the release note prefix and version bump mismatch, and make sure to add type hints to the newly added functions in `CommonServerPython`. Let me know if you have any questions!

 @RosenbergYehuda please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

Packs/Base/Scripts/CommonServerPython/CommonServerPython.py
```diff
@@ -14239,50 +14247,88 @@ def data_error_handler(res):
...
+                    buf = _io.BytesIO()
+                    gz = gzip.GzipFile(fileobj=buf, mode='wb')
+                    chunk_uncompressed = 0
+                    chunk_items = 0
+
+            # flush the final (partial) chunk
+            gz.close()
+            if chunk_items:
+                data_size += _dispatch(buf.getvalue(), chunk_items)
+        except Exception:
```
DanielTal87, line 14321: already-delivered chunks become uncounted on a mid-stream failure.

When an exception fires here we `raise`, which means the `return all_futures` on L14331 never executes. But `all_futures` is exactly what the caller uses to tally delivered events:

```python
for future in concurrent.futures.as_completed(futures):
    data_size += future.result()
demisto.updateModuleHealth({'eventsPulled': data_size})
```

So any chunk that was **already successfully POSTed** before the failure is silently lost from the count. The events are genuinely in XSIAM, but the health module reports fewer (often zero). That's a data-integrity issue rather than a cosmetic one — support will chase a "collector not pulling" report for events that actually arrived.

I extracted this exact dispatch loop and ran it with 60 chunks, failing at chunk 5:

```
futures submitted before abort: 34
futures completed OK (payload already sent to XSIAM): 14
```

14 chunks delivered, 14 chunks uncounted.

**Separately:** the comment on L14322 says "cancel pending work", but `shutdown(wait=False)` does **not** cancel queued tasks — it only stops the pool accepting new ones. Verified: after `shutdown(wait=False)`, all 10 queued tasks still ran to completion. `cancel_futures=True` (3.9+) is what actually cancels.

Suggested:

```python
        except Exception as exc:
            # Don't leak worker threads. Note: shutdown(wait=False) alone does NOT cancel queued
            # work, so pass cancel_futures where the runtime supports it (3.9+).
            if executor is not None:
                try:
                    executor.shutdown(wait=False, cancel_futures=True)
                except TypeError:  # Python < 3.9
                    executor.shutdown(wait=False)
                # Chunks already POSTed are real events in XSIAM. Surface them so the caller can
                # still report an accurate count instead of silently losing them.
                exc.submitted_futures = all_futures  # type: ignore[attr-defined]
            raise
```

Could we also add a test covering this path? Right now the `except` block has no coverage at all.

Packs/Base/Scripts/CommonServerPython/CommonServerPython_test.py
```diff
@@ -10117,6 +10117,175 @@ def test_send_data_to_xsiam_streaming_skips_oversized_entry_like_legacy(self, mo
         assert len(streaming_lines) == 2
         assert all('blob' not in line for line in streaming_lines)
 
+    @pytest.mark.parametrize('chunk_size', [2 ** 20, 50])
+    def test_send_data_to_xsiam_streaming_threaded_matches_legacy(self, mocker, chunk_size):
```
DanielTal87, line 10121: Solid additions overall - the legacy payload-equivalence test is well designed. A few gaps before merge:

FIX:
1. `test_send_data_to_xsiam_streaming_threaded_bounds_inflight_chunks` is tautological. It measures
   *concurrent* `_http_request` calls, but `max_workers=NUM_OF_WORKERS` already caps that
   independently of `MAX_INFLIGHT_CHUNKS`. Verified: the assertion passes unchanged with the
   backpressure loop deleted. Peak *resident* payloads - the actual memory property - is 20 with the
   bound vs 60 without, and is never measured. Please assert on residency (count from submission
   until the future completes), and consider monkeypatching `MAX_INFLIGHT_CHUNKS` below
   `NUM_OF_WORKERS` so the gate isn't shadowed by the pool size.

ADD:
2. Mid-stream failure on the threaded path (highest priority). The `except` block has zero coverage.
   Assert the exception propagates, the executor is shut down, and chunks already delivered to XSIAM
   are still reportable to the caller.
3. Error surfacing from inside the backpressure loop - `finished.result()` re-raising a worker
   failure while blocked waiting for capacity.
4. `data_type='assets'` with `use_streaming_send=True` + `multiple_threads=True` - assert
   `snapshot-id` and `total-items-count` stay correct across concurrent workers.
5. `_dispatch`'s threaded return contract - returns `0`, `data_size` stays `0`, and the tally comes
   only from the returned futures.
6. All-oversized or empty input with `multiple_threads=True` - confirm the caller receives `[]`
   rather than `None`.

DanielTal87 (approved):

PR #45797, opened 2026-09-06
[CRTX-271524] - XDM Identity migration member 4

oatias:
…merge branch

The 2.11 modeling rule added for the xdm.*.identity.* migration uses a rule id and name that do not match the existing v1 rule:

  v1 : id=BeyondTrust_PRA_ModelingRule   name=BeyondTrust Privileged Remote Access Modeling Rule
  new: id=BeyondTrustPrivilegedRemoteAccess_ModelingRule
       name=BeyondTrustPrivilegedRemoteAccess Modeling Rule

A versioned modeling rule must keep the id and name of the version it replaces, otherwise the tenant treats it as a separate rule instead of a newer version of the same one. Because the id/name cannot be corrected without a dedicated review, the pack is parked on this branch for a force merge and reverted from crtx-271524-identity-pr1.

Relates to CRTX-271524.

## Status
- [ ] In Progress
- [x] Ready

## Must have
- [ ] Tests
- [ ] Documentation


oatias:
This PR is part of the identity.* XDM migration (CRTX-271524).
It caps the old modeling rule at toversion: 8.14.0 and adds a new _2_11 rule (fromversion: 8.15.0) mirroring xdm..user. → xdm..identity..

run-validations fails only on BC107 ("changing toversion requires a same-id successor"). This is unavoidable here: the affected packs have legacy rule ids that don't end in ModelingRule, so the new rule must be renamed to the conforming id (MR108) — which breaks the same-id chain BC107 checks. MR108 + BC105 (id immutable) + BC107 (same id) can't all be satisfied for legacy-id packs.

tmizrahiPA (approved):

PR #45802, opened 2026-09-07
 Mimecast no events ingestion bug fix

YaelShamai:

## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-75647

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
+++ b/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
@@ -791,12 +791,11 @@ async def fetch_siem_events(
     if not is_within_last_24_hours(start_date):
         demisto.info(f"{log_prefix} {start_date=} is older than 24 hours. Skipping forward to last 23 hours.")
         start_date = convert_to_siem_filter_format(UTC_NOW - timedelta(hours=23))
-        # The stored next_page cursor and last_fetched_ids belong to the old (out-of-range) window.
-        # They must be cleared as well; otherwise the stale next_page (which the API honors over the date range)
-        # keeps returning empty pages and the stale last_fetched_ids dedups real events away - a perpetual 0-events loop.
+        # The stored next_page cursor belongs to the old (out-of-range) window.
+        # It must be cleared, otherwise the stale next_page (which the API honors over the date range)
+        # keeps returning empty pages.
         demisto.info(f"{log_prefix} Resetting stale next_page and last_fetched_ids to match the new start date.")
         next_page = None
-        last_fetched_ids = []
 
     siem_events, new_next_page = await get_siem_events(
         client,
@@ -810,12 +809,12 @@ async def fetch_siem_events(
     # The Mimecast SIEM v2 endpoint is a checkpoint stream: even empty responses return a fresh @nextPage cursor
     # pointing at the current tip of each channel. We MUST persist that cursor (and advance the window) so the next
     # cycle resumes from the checkpoint. Otherwise we cold-start from "now - 1 minute" every run and only ever catch
-    # events that happen to land in that ~60s slice between fetches - the perpetual 0-events symptom seen in prod.
+    # events that happen to land in that ~60s slice between fetches.
     if not siem_events:
         siem_next_run = {
             START_DATE_KEY: start_date,
             LAST_FETCHED_IDS_KEY: last_fetched_ids,
-            NEXT_PAGE_KEY: new_next_page or next_page,
+            NEXT_PAGE_KEY: new_next_page if new_next_page is not None else next_page,
         }
         demisto.debug(f"{log_prefix} No new events found. Advancing cursor. Updating {siem_next_run=}.")
         return siem_next_run, []
--- a/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
+++ b/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
@@ -791,6 +791,12 @@ async def fetch_siem_events(
     if not is_within_last_24_hours(start_date):
         demisto.info(f"{log_prefix} {start_date=} is older than 24 hours. Skipping forward to last 23 hours.")
         start_date = convert_to_siem_filter_format(UTC_NOW - timedelta(hours=23))
+        # The stored next_page cursor and last_fetched_ids belong to the old (out-of-range) window.
+        # They must be cleared as well; otherwise the stale next_page (which the API honors over the date range)
+        # keeps returning empty pages and the stale last_fetched_ids dedups real events away - a perpetual 0-events loop.
+        demisto.info(f"{log_prefix} Resetting stale next_page and last_fetched_ids to match the new start date.")
+        next_page = None
+        last_fetched_ids = []
 
     siem_events, new_next_page = await get_siem_events(
         client,
@@ -800,10 +806,19 @@ async def fetch_siem_events(
         next_page=next_page,
     )
 
-    # Handle empty results
+    # Handle empty results.
+    # The Mimecast SIEM v2 endpoint is a checkpoint stream: even empty responses return a fresh @nextPage cursor
+    # pointing at the current tip of each channel. We MUST persist that cursor (and advance the window) so the next
+    # cycle resumes from the checkpoint. Otherwise we cold-start from "now - 1 minute" every run and only ever catch
+    # events that happen to land in that ~60s slice between fetches - the perpetual 0-events symptom seen in prod.
     if not siem_events:
-        demisto.debug(f"{log_prefix} No new events found. Keeping {siem_last_run=}.")
-        return siem_last_run, []
+        siem_next_run = {
+            START_DATE_KEY: start_date,
+            LAST_FETCHED_IDS_KEY: last_fetched_ids,
+            NEXT_PAGE_KEY: new_next_page or next_page,
+        }
+        demisto.debug(f"{log_prefix} No new events found. Advancing cursor. Updating {siem_next_run=}.")
+        return siem_next_run, []
 
     # Update state with newest events
     new_start_time, new_last_fetched_ids = get_siem_new_start_time_last_fetched_ids(siem_events)
--- a/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
+++ b/Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
@@ -1,4 +1,4 @@
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 import pytest
 from pytest_mock import MockerFixture
 from aiohttp import ClientResponseError, RequestInfo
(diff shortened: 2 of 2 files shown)


DanielTal87:
@marketplace-ai-reviewer start review

marketplace-ai-reviewer:

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
Thanks for submitting your PR!
Before moving on with the detailed review, please take a moment to address the following general comments.

Once you've improved your PR according to the comments below, re-trigger the review by adding the `ready-for-ai-review` label or commenting `@marketplace-ai-reviewer start review`.


- Your PR title and description are too generic ('fix bug'). Please update the title and description to mention the specific integration (MimecastEventCollectorV2) and briefly describe the actual issue being resolved (e.g., fixing pagination cursor persistence and stale start date handling).

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
DanielTal87: Use an explicit is not None 


```suggestion
            NEXT_PAGE_KEY: new_next_page if new_next_page is not None else next_page,
```

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
DanielTal87, line 730: Nice work adding both a good-path and a bad-path test for `fetch_siem_events()`. A few gaps remain before this is airtight:

1. `test_fetch_siem_events` - the parametrized case `id="No new events (keeps last run)"` still encodes the PRE-FIX contract. It passes only because `mock_next_page` is `None`, so `new_next_page or next_page` collapses to the old value. Please rename it and add a sibling case with a non-`None` `mock_next_page`.

2. Missing: normal (non-stale) path where the API returns `new_next_page=None` but `last_run` holds a non-`None` `next_page`. The `or` fallback resurrects the retired cursor here. Add a test asserting the intended behaviour.

3. Missing: `new_next_page=""` (falsy but not `None`). `assign_params` strips falsy values, so this needs an explicit assertion.

4. Missing: combined path - stale `start_date` AND the API returning a fresh cursor. `test_fetch_siem_events_stale_start_date_resets_cursor` hardcodes `([], None)`, so the reset branch and the empty-result branch are never exercised together.

5. Missing: a true regression test for the reported defect - two consecutive `fetch_siem_events()` calls proving run N+1 resumes from run N's cursor instead of cold-starting at `UTC_MINUTE_AGO`. This is the bug; please reproduce it.

6. Missing: coverage for the duplicate-ingestion consequence of `last_fetched_ids = []` on the stale path.

7. Nit: in `test_fetch_siem_events_stale_start_date_resets_cursor`, consider dropping the `is_within_last_24_hours` patch. The `2020-01-01` fixture is genuinely stale, so the real predicate will trip against the patched `UTC_NOW` and you gain a real assertion for free.

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
DanielTal87: The `last_fetched_ids = []` on the same line doesn't do what the comment says it does.

The comment claims the stale IDs "dedup real events away." Looking at `deduplicate_and_format_events()`,
it's an exact ID match - it only skips an event if that exact ID is already in the set. So an old ID can
only ever suppress the *same* event we already sent. It can't block a different event, because IDs don't
collide. That means stale IDs were never causing the 0-events symptom.

What clearing the list *does* do: after we jump the window back to `now - 23h`, that window overlaps
events we already sent last run. With the ID list emptied, we have nothing to filter them against and
we'll re-ingest them as duplicates.

Think of it like a guest list of "people we already let in" - binning it doesn't let new guests in
(they were never blocked), it just lets the people already inside walk in a second time.

Could you either:
  a) drop the `last_fetched_ids = []` line and keep only the `next_page` reset, or
  b) if there's concrete evidence from the ticket that the IDs really did cause loss, link it and I'll
     happily re-review?

Also worth noting: if the actual concern is the ID list growing unbounded over time, that's a fair worry
but a separate one - it'd be better handled with a size cap than an unconditional wipe on this path.
Either way the code comment needs updating, since as written it states something inaccurate about
`deduplicate_and_format_events()` and will mislead the next person who reads it.

marketplace-ai-reviewer:

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
```diff
@@ -791,6 +791,11 @@ async def fetch_siem_events(
     if not is_within_last_24_hours(start_date):
         demisto.info(f"{log_prefix} {start_date=} is older than 24 hours. Skipping forward to last 23 hours.")
         start_date = convert_to_siem_filter_format(UTC_NOW - timedelta(hours=23))
+        # The stored next_page cursor belongs to the old (out-of-range) window.
+        # It must be cleared, otherwise the stale next_page (which the API honors over the date range)
+        # keeps returning empty pages.
+        demisto.info(f"{log_prefix} Resetting stale next_page and last_fetched_ids to match the new start date.")
```
marketplace-ai-reviewer, line 797: The log message states that `last_fetched_ids` is being reset, but the code (and the corresponding unit test) intentionally preserves it for continued deduplication. Please update the log message to accurately reflect the code's behavior.

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
Hi there, thanks for your contribution to the Mimecast pack! I've reviewed your PR and found a few minor adjustments to make before we can merge. The main items include correcting an inaccurate log message in the integration, adding the vendor name to the metadata keywords, and properly formatting the release notes using the SDK. Thanks again for your work on this, and let me know if you have any questions!

- `Packs/Mimecast/pack_metadata.json`: The vendor name 'Mimecast' is missing from the `keywords` array.
- `Packs/Mimecast/ReleaseNotes/3_0_10.md`: The release notes file is currently empty. It must follow the required markdown hierarchy and use an approved template prefix.

For this PR, the hierarchy and template should look like this:
```markdown
#### Integrations
##### Mimecast Event Collector v2
- Fixed an issue where SIEM events could be missed when the API returned empty results.
```

Please run `demisto-sdk update-release-notes -i Packs/Mimecast -u revision` to auto-generate the release notes file with the correct hierarchy, and then update the placeholder with a description of the bug fix using an approved prefix (e.g., `Fixed an issue...`). For more details, refer to the [Release Notes Documentation](https://xsoar.pan.dev/docs/documentation/release-notes#general).

 @JasBeilin, @marketplace-ai-reviewer, @DanielTal87, @YaelShamai please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
```diff
@@ -800,10 +805,19 @@ async def fetch_siem_events(
         next_page=next_page,
     )
 
-    # Handle empty results
+    # Handle empty results.
+    # The Mimecast SIEM v2 endpoint is a checkpoint stream: even empty responses return a fresh @nextPage cursor
+    # pointing at the current tip of each channel. We MUST persist that cursor (and advance the window) so the next
+    # cycle resumes from the checkpoint. Otherwise we cold-start from "now - 1 minute" every run and only ever catch
+    # events that happen to land in that ~60s slice between fetches.
```
DanielTal87, line 812: The comment says "(and advance the window)" but start_date is persisted unchanged. For a tenant with no events, start_date freezes and goes stale after 24h — which then triggers the reset branch above and wipes the very cursor we just saved, plus re-fetches 23h against a ~100-ID dedup list. Please advance start_date when a resume cursor exists.
```suggestion
    if not siem_events:
        # The API returns a fresh @nextPage even on empty responses (checkpoint stream).
        # Persist it AND advance start_date to the fetch time, otherwise start_date rots
        # past the 24h window and the stale-reset branch above would discard the cursor.
        resume_cursor = new_next_page if new_next_page is not None else next_page
        siem_next_run = {
            START_DATE_KEY: convert_to_siem_filter_format(UTC_NOW) if resume_cursor else start_date,
            LAST_FETCHED_IDS_KEY: last_fetched_ids,
            NEXT_PAGE_KEY: resume_cursor,
        }
        demisto.debug(f"{log_prefix} No new events found. Advancing cursor. Updating {siem_next_run=}.")
        return siem_next_run, []
```

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
```diff
@@ -1,4 +1,4 @@
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
```
DanielTal87, line 1: Solid test work — the Given/When/Then docstrings and the `None`-vs-fresh-cursor
parametrization are genuinely good. A few gaps before this is ready:

1. MISSING (critical): A test for the 24-hour rot scenario. Because the empty-result
   branch persists `start_date` unchanged, a tenant with no events will eventually
   have `start_date` age past 24h, which triggers the `is_within_last_24_hours` reset
   branch and sets `next_page = None` — destroying the cursor this PR just saved.
   Please add a test that runs `fetch_siem_events` with an empty result, then feeds
   the resulting `next_run` back in with `UTC_NOW` advanced by >24h, and asserts the
   cursor is NOT discarded. This test should fail against the current implementation.

2. MISSING (bad path): No test covers `get_siem_events` raising. Please add coverage
   for `DemistoException` (the `error` key branch) and `ClientResponseError`, asserting
   that `last_run` is not silently corrupted on failure.

3. MISSING (boundary): No test pins `start_date` exactly at the 24-hour edge.
   `is_within_last_24_hours` uses `>=`, so please assert both sides of the boundary.

4. MISSING (integration): All new tests call `fetch_siem_events` directly. Please add
   one assertion through `fetch_events_command` confirming the corrected `next_run`
   is nested correctly under the `siem` key — persistence is where this bug lived.

5. REMOVE/REWORK: `test_fetch_siem_events_empty_result_empty_string_cursor_is_persisted`
   asserts `""` is "distinct from None". It is not — `assign_params` discards `""` and
   `None` identically before the request is built (verified). Either drop this test and
   normalize falsy cursors to `None`, or assert the normalization instead.

6. NIT: In `test_fetch_siem_events_resumes_from_previous_cursor`, the
   `assert call_kwargs["start_date"] != cold_start_date` line is implied by the
   `== run_n_start_date` assertion above it and can never fail independently.

Note: the suite requires Python 3.11+ (`from datetime import UTC`). Please confirm
green via `demisto-sdk pre-commit -i Packs/Mimecast/Integrations/MimecastEventCollectorV2/`.

DanielTal87 (approved):

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
```diff
@@ -784,39 +845,54 @@ async def fetch_siem_events(
...
 
-    # Ensure the start date within than 24 hours to avoid HTTP 400 (bad request) errors from SIEM API endpoint
-    # If no events were fetched within the last 24 hours, the start date may not be within the allowed API range
+    # Migration from the previous integration version, which persisted the cursor between fetch cycles:
+    # only reuse a persisted cursor if it is valid (offset is not 0-0 for all types). Either way, the cursor is
+    # no longer persisted going forward - it is only used within a single fetch cycle.
+    persisted_cursor = siem_last_run.get(NEXT_PAGE_KEY)
+    initial_next_page = persisted_cursor if is_valid_cursor(persisted_cursor) else None
+    if initial_next_page:
+        demisto.info(
```
DanielTal87: make sure only very important logs are in log level info, if no change to debug

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
```diff
@@ -13,11 +13,24 @@
...
     DEFAULT_BASE_URL,
+    NEXT_PAGE_KEY,
+    START_DATE_KEY,
+    LAST_FETCHED_IDS_KEY,
     EventTypes,
     AsyncClient,
 )
 from test_data.data import AUDIT_RAW_RESPONSE, SIEM_RAW_RESPONSE
 
+# A "bad" SIEM cursor: all event types have offset 0-0 (base64 of
```
DanielTal87, line 25: Solid test work overall - the parametrized `is_valid_cursor` suite and the migration-path coverage are appreciated. A few gaps that map directly to real failure modes:

MISSING TEST CASES for `is_valid_cursor()`:
- Unpadded base64 cursor (no trailing `=`). Verified locally: `base64.b64decode(..., validate=True)` raises `binascii.Error`, so the cursor is wrongly classified invalid.
- URL-safe base64 cursor containing `-` or `_`. Same wrong classification.
- Cursor with all `0-0` offsets but real pagination progress in the suffix, e.g. `[attachment_protect=0-0, av=0-0]:3:<real_token>`. Verified locally: currently returns `False`, which means we discard a working cursor and drop pages.
- Cursor with a malformed suffix (missing `:` segments) to confirm no `IndexError`/`ValueError`.

MISSING TEST CASES for `get_siem_events()`:
- Invalid cursor returned on page 2 (not page 1) - confirm pagination stops mid-run and events already collected are still returned sorted.
- Confirm that after an early break, the returned `next_page` is not a value the caller could mistakenly persist.

MISSING TEST CASES for `fetch_siem_events()`:
- Two consecutive fetch cycles where the first stops early on an invalid cursor: assert `last_fetched_ids` from cycle 1 prevents duplicate ingestion in cycle 2.
- Assert every returned event contains `_time` and that `_filter_time` has been removed.

CONTRADICTORY ASSERTION in `test_fetch_siem_events`:
- Line 855 includes `"next_page": None` in `expected_next_run`, while line 858 asserts `NEXT_PAGE_KEY not in next_run`. These are mutually exclusive - please remove `"next_page": None` from the expected dict.

ENVIRONMENT:
- The test module imports `UTC` from `datetime`, which requires Python 3.11+. Local `pytest` on 3.9 fails at collection. Please confirm CI runs via `demisto-sdk pre-commit` in the pinned Docker image.

Packs/Mimecast/pack_metadata.json
```diff
@@ -2,7 +2,7 @@
     "name": "Mimecast",
     "description": "Mimecast unified email management offers cloud email services for email security, continuity and archiving emails.",
     "support": "xsoar",
-    "currentVersion": "3.1.2",
+    "currentVersion": "3.1.3",
```
DanielTal87: Notice that the release notes are empty

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
```diff
@@ -622,6 +677,12 @@ async def get_siem_events(
             )
             break
 
+        # Validate the returned cursor before using it for the next page. A "bad" cursor (offset 0-0 for all types)
+        # would restart pagination from the beginning, so stop paginating instead.
+        if not is_valid_cursor(next_page):
+            demisto.info(f"{log_prefix} Received invalid cursor after {page_number=}. Stopping pagination for this fetch cycle.")
+            break
```
DanielTal87, line 685: We return a cursor we've just proven invalid, and the test asserts that as expected behaviour. Both callers discard it anyway. Please either set next_page = None before breaking, or drop it from the return signature entirely so nobody can misuse it later.

```suggestion
                # Validate the returned cursor before using it for the next page. A degenerate cursor would restart
        # pagination from the beginning, so discard it and stop paginating.
        if not is_valid_cursor(next_page):
            demisto.info(f"{log_prefix} Received invalid cursor after {page_number=}. Stopping pagination for this fetch cycle.")
            next_page = None
            break
```

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2_test.py
YaelShamai: Added relevant tests

marketplace-ai-reviewer:

DanielTal87 (approved):

Packs/Mimecast/Integrations/MimecastEventCollectorV2/MimecastEventCollectorV2.py
```diff
@@ -784,39 +852,54 @@ async def fetch_siem_events(
...
+    # date is not kept. The cursor is never persisted between fetch cycles.
     if not siem_events:
-        demisto.debug(f"{log_prefix} No new events found. Keeping {siem_last_run=}.")
-        return siem_last_run, []
+        siem_next_run = {START_DATE_KEY: start_date, LAST_FETCHED_IDS_KEY: last_fetched_ids}
+        demisto.debug(f"{log_prefix} No new events found. Persisting {siem_next_run=}.")
+        return siem_next_run, []
 
-    # Update state with newest events
+    # Update state with newest events. The cursor is intentionally NOT persisted - only the latest event
```
marketplace-ai-reviewer, line 893: Consider using `event.pop(FILTER_TIME_KEY, None)` to avoid a potential `KeyError`.

marketplace-ai-reviewer:
<details>

</details>

marketplace-ai-reviewer (commented):
#### 🔄 Re-reviewing updates...
> [!IMPORTANT]
> To maintain review integrity, threads marked as **Resolved** without accompanying changes or explanations will be automatically **unresolved** for further verification.

Hi! Thanks for your contribution to the Mimecast Event Collector integration. I've taken a look at the code and just have one quick suggestion regarding dictionary handling. Specifically, using `.pop()` with a default value of `None` will help prevent potential `KeyError` exceptions when removing keys from events. Great work overall!

 @marketplace-ai-reviewer, @YaelShamai, @DanielTal87, @JasBeilin please review and approve the results generated by the AI Reviewer by responding 👍 on this comment.

Packs/Mimecast/ReleaseNotes/3_1_4.md
```diff
@@ -0,0 +1,6 @@
+
+#### Integrations
+
+##### Mimecast Event Collector v2
+
+- Fixed an issue where the collector permanently stopped returning SIEM events when the API returned an invalid pagination cursor.
```
JuliaH-TW, line 6:
```suggestion
Fixed an issue where the collector permanently stopped returning SIEM events when the API returned an invalid pagination cursor.
```

JuliaH-TW (commented):
@YaelShamai 
Doc review complete. Please check the comment. In the meantime, I'm adding the docs-approved label.

